PR #26, opened 2023-08-22
Windows compilation

### LINKED ISSUE #25: Zig's std.os.getenv is linux-only and does not compile on windows.
### Description

This program cannot compile on Windows because `std.os.getenv` is not cross-platform. It should be replaced with `std.process.getEnvVarOwned`.

### How to Reproduce the Bug

- Run `zig build` on windows.

### Additional Context

```
D:\zig\lib\std\os.zig:1910:9: error: std.os.getenv is unavailable for Windows because environment string is in WTF-16 format. See std.process.getEnvVarOwned for cross-platform API or std.os.getenvW for Windows-specific API.
        @compileError("std.os.getenv is unavailable for Windows because environment string is in WTF-16 format. See std.process.getEnvVarOwned for cross-platform API or std.os.getenvW for Windows-specific API.");
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
referenced by:
    getOrInitIdentity: src\gossip\cmd.zig:19:28
    initSpy: src\gossip\cluster_info.zig:62:30
    remaining reference traces hidden; use '-freference-trace' to see all reference traces
```

sarahkittyy:
Fixes #25

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -1,5 +1,4 @@
 const cmd = @import("cmd/cmd.zig");
-const std = @import("std");
 
 pub fn main() !void {
     try cmd.run();
--- a/src/main.zig
+++ b/src/main.zig
@@ -1,4 +1,5 @@
 const cmd = @import("cmd/cmd.zig");
+const std = @import("std");
 
 pub fn main() !void {
     try cmd.run();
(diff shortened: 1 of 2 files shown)


src/main.zig
```diff
@@ -1,4 +1,5 @@
 const cmd = @import("cmd/cmd.zig");
+const std = @import("std");
```
ultd, line 2: Can you remove this `std` import as it's not needed?

ultd (changes requested):

ultd (approved):

PR #31, opened 2023-09-05
core: add ShredVersion calculation and HardForks to Core

anoushk1234:
Summary of Changes:
- added a method that computes the shred version in the exact same way as done in the Solana labs client using a 32 byte hash
- added HardForks imlementation to core


build.zig
```diff
@@ -83,20 +83,32 @@ pub fn build(b: *std.Build) void {
         .filter = if (b.args) |args| args[0] else null, // filter tests like so: zig build test -- "<FILTER>"
     });
 
+    const hard_fork_tests = b.addTest(.{
```
0xNineteen, line 86: you shouldnt need to modify the build file -- should remove these changes -- after adding the file path to `lib.zig`, running `zig build test` should run the tests you added

src/core/hard_forks.zig
```diff
@@ -0,0 +1,56 @@
...
+    }
+    pub fn deinit(self: *Self) void {
+        self.hard_forks.deinit();
+    }
+    pub fn get_forks(self: *Self) []HardFork {
+        return self.hard_forks.items;
+    }
+};
+
+test "hard_forks" {
```
0xNineteen, line 38: can we change this to "core.hard_forks: test hardforks" to follow the test naming standards

src/core/shred.zig
```diff
@@ -11,4 +13,24 @@ pub const ShredVersion = struct {
...
+            for (chunk, accum[0..chunk.len]) |ch, *acc| {
+                acc.* ^= ch;
+            }
+        }
+        var version = (@as(u16, accum[0]) << 8) | accum[1];
+        return version +| 1;
+    }
 };
+
+test "computes version from raw hash" {
```
0xNineteen, line 32: can we also prepend "core.shred: ..." to the test name here

src/core/shred.zig
```diff
@@ -11,4 +13,24 @@ pub const ShredVersion = struct {
...
         return Self{ .value = version };
     }
+
+    pub fn version_from_hash(hash: *Hash) u16 {
+        const hash_bytes = hash.data;
+        var accum: [2]u8 = .{ 0, 0 };
+        var chunks = std.mem.window(u8, &hash_bytes, 2, 2);
+
+        while (chunks.next()) |chunk| {
+            for (chunk, accum[0..chunk.len]) |ch, *acc| {
```
0xNineteen, line 23: can we unwrap this for loop so we just do `acc[0] ^= chunk[0]` for indexs 0 and 1 to be more clear

src/core/hard_forks.zig
```diff
@@ -0,0 +1,56 @@
...
+
+pub const HardFork = struct { Slot, usize };
+pub const HardForks = struct {
+    hard_forks: ArrayList(HardFork),
+
+    const Self = @This();
+    pub fn default(alloc: allocator) Self {
+        return .{ .hard_forks = ArrayList(HardFork).init(alloc) };
+    }
+    pub fn register(self: *Self, new_slot: Slot) !void {
```
0xNineteen, line 14: can you run `zig fmt src/` for the linter

0xNineteen (changes requested):

src/core/hard_forks.zig
anoushk1234: Yea will do and would be good to add these to contributing.md

src/core/hard_forks.zig
```diff
@@ -0,0 +1,85 @@
+const std = @import("std");
+const ArrayList = std.ArrayList;
+const allocator = std.mem.Allocator;
+const Slot = @import("slot.zig").Slot;
+pub const HardFork = struct { Slot, usize };
```
0xNineteen, line 5: can you add names to this struct and then reference the fields instead of [0] (better for readability)

src/core/hard_forks.zig
```diff
@@ -0,0 +1,85 @@
...
+    pub fn default(alloc: allocator) Self {
+        return .{ .hard_forks = ArrayList(HardFork).init(alloc) };
+    }
+    pub fn register(self: *Self, new_slot: Slot) !void {
+        const index = for (self.hard_forks.items, 0..) |hf, i| {
+            if (hf[0].value == new_slot.value) break i;
+        } else null;
+
+        if (index != null) {
+            try self.hard_forks.append(.{ new_slot, self.hard_forks.items[index.?][1] +| 1 });
```
0xNineteen, line 19: this shouldnt be append - it should be `self.hard_forks[index] =`

src/core/hard_forks.zig
```diff
@@ -0,0 +1,85 @@
...
+            var current_fork_slot = hard_fork[0].value;
+            var current_fork_count = hard_fork[1];
+
+            if (parent_slot.value < current_fork_slot and slot.value >= current_fork_slot) {
+                fork_count += current_fork_count;
+            }
+        }
+
+        if (fork_count > 0) {
+            var buf = [_]u8{0} ** 8;
```
0xNineteen, line 47: better to do `var buf: [8]u8 = undefined` so we dont waste time zeroing

src/core/hard_forks.zig
```diff
@@ -0,0 +1,85 @@
...
+            std.mem.writeIntLittle(u64, &buf, @as(u64, fork_count));
```
0xNineteen, line 48: if we are casting `@as` here maybe we just set fork_count as a `u64`

src/core/shred.zig
```diff
@@ -11,4 +11,51 @@ pub const ShredVersion = struct {
...
+        var chunks = std.mem.window(u8, &hash_bytes, 2, 2);
+
+        while (chunks.next()) |chunk| {
+            accum[0] ^= chunk[0];
+            accum[1] ^= chunk[1];
+        }
+        var version = (@as(u16, accum[0]) << 8) | accum[1];
+        return version +| 1;
+    }
+    pub fn compute_shred_version(genesis_hash: Hash, hard_forks: ?HardForks, alloc: Allocator) u16 {
```
0xNineteen, line 27: can we add spaces between the functions

src/core/hard_forks.zig
```diff
@@ -0,0 +1,85 @@
...
+pub const HardFork = struct { Slot, usize };
+pub const HardForks = struct {
+    hard_forks: ArrayList(HardFork),
+
+    const Self = @This();
+    pub fn default(alloc: allocator) Self {
+        return .{ .hard_forks = ArrayList(HardFork).init(alloc) };
+    }
+    pub fn register(self: *Self, new_slot: Slot) !void {
+        const index = for (self.hard_forks.items, 0..) |hf, i| {
```
0xNineteen, line 14: rename `i` to `index` to be more explicit - same with `hf`

src/core/hard_forks.zig
```diff
@@ -0,0 +1,85 @@
...
+const Slot = @import("slot.zig").Slot;
+pub const HardFork = struct { Slot, usize };
+pub const HardForks = struct {
+    hard_forks: ArrayList(HardFork),
+
+    const Self = @This();
+    pub fn default(alloc: allocator) Self {
+        return .{ .hard_forks = ArrayList(HardFork).init(alloc) };
+    }
+    pub fn register(self: *Self, new_slot: Slot) !void {
```
0xNineteen, line 13: kinda a bit out of scope for the PR but could you change the `Slot` struct to just be `pub const Slot = u64` - should clean up the code a lot by removing `.value` checks

0xNineteen (changes requested):

src/core/hard_forks.zig
anoushk1234: @0xNineteen this is major change, it would affect other parts of the code where slot is used, probably not a good idea to do it in this PR itself

src/core/shred.zig
anoushk1234: is there any way we can configure zig fmt similar to rust fmt to follow this?

src/core/hard_forks.zig
0xNineteen: @anoushk1234  I think its only used in `crds.zig` rn as default values? - shouldnt be too bad

anoushk1234: Just to clarify you want to remove Slot as a struct altogether since you wrote `pub const Slot = u64`
```zig
pub const Slot = struct {
    value: u64,

    const Self = @This();

    pub fn init(slot: u64) Self {
        return Self{
            .value = slot,
        };
    }

    /// returns slot with 0 as value
    pub fn default() Self {
        return Self{ .value = 0 };
    }
};
```

0xNineteen: @anoushk1234 yeah exactly

anoushk1234: @0xNineteen Zig doesnt support this yet, using an enum is an option though although it's not the clean

anoushk1234: @0xNineteen done

src/core/hard_forks.zig
```diff
@@ -0,0 +1,85 @@
...
+    hard_forks: ArrayList(HardFork),
+
+    const Self = @This();
+
+    pub fn default(alloc: allocator) Self {
+        return .{ .hard_forks = ArrayList(HardFork).init(alloc) };
+    }
+
+    pub fn register(self: *Self, new_slot: Slot) !void {
+        const index = for (self.hard_forks.items, 0..) |hard_fork, index| {
```
0xNineteen, line 16: can you change this to `maybe_index` and then instead of `!= null` check - use a `if (maybe_index) |index| {` - ill also add this to contributing.md

src/core/hash.zig
```diff
@@ -46,4 +49,13 @@ pub const Hash = struct {
         }
         return CompareResult.Equal;
     }
+
+    pub fn extend_and_hash(id: Hash, val: []u8, alloc: Allocator) !Self {
+        var hash_data = std.ArrayList(u8).init(alloc);
+        defer _ = hash_data.deinit();
```
0xNineteen, line 55: this should just be `defer hash_data.deinit()` - deinit is void

also can you init the array `withCapacity` since we know the length (then well only have one alloc) - also should use a fcn like insertSliceWithCapacity

src/core/shred.zig
```diff
@@ -11,4 +12,52 @@ pub const ShredVersion = struct {
...
+
+        while (chunks.next()) |chunk| {
+            accum[0] ^= chunk[0];
+            accum[1] ^= chunk[1];
+        }
+        var version = (@as(u16, accum[0]) << 8) | accum[1];
+        return version +| 1;
+    }
+
+    pub fn compute_shred_version(genesis_hash: Hash, hard_forks: ?HardForks, alloc: Allocator) u16 {
```
0xNineteen, line 29: change naming to maybe_hard_forks - and follow the `if (maybe_x) |x|` format

0xNineteen (changes requested):

src/core/hash.zig
```diff
@@ -46,4 +49,13 @@ pub const Hash = struct {
         }
         return CompareResult.Equal;
     }
+
+    pub fn extend_and_hash(id: Hash, val: []u8, alloc: Allocator) !Self {
+        var hash_data = try std.ArrayList(u8).initCapacity(alloc, val.len + id.data.len);
+        defer hash_data.deinit();
+        hash_data.appendSliceAssumeCapacity(id.data[0..]);
+        try hash_data.appendSlice(val);
```
0xNineteen, line 57: this should also be `appendSliceAssumeCapacity`

src/core/hash.zig
```diff
@@ -46,4 +49,13 @@ pub const Hash = struct {
         }
         return CompareResult.Equal;
     }
+
+    pub fn extend_and_hash(id: Hash, val: []u8, alloc: Allocator) !Self {
```
0xNineteen, line 53: can you make the errors explicit - ie, this fcn should be `error { OutOfMemory }` since thats the only error it returns

src/core/shred.zig
```diff
@@ -11,4 +12,52 @@ pub const ShredVersion = struct {
...
+    }
+
+    pub fn compute_shred_version(genesis_hash: Hash, maybe_hard_forks: ?HardForks, alloc: Allocator) u16 {
+        var hash = genesis_hash;
+        if (maybe_hard_forks) |hard_forks| {
+            for (hard_forks.get_forks()) |hard_fork| {
+                var buf: [16]u8 = undefined;
+                std.mem.writeIntLittle(u64, buf[0..8], hard_fork.slot);
+                std.mem.writeIntLittle(u64, buf[8..], @as(u64, hard_fork.count));
+                hash = Hash.extend_and_hash(genesis_hash, buf[0..], alloc) catch genesis_hash;
```
0xNineteen, line 36: when passing slices can we use the `&buf` syntax instead of `buf[0..]`

src/core/shred.zig
```diff
@@ -11,4 +12,52 @@ pub const ShredVersion = struct {
...
+        }
+        var version = (@as(u16, accum[0]) << 8) | accum[1];
+        return version +| 1;
+    }
+
+    pub fn compute_shred_version(genesis_hash: Hash, maybe_hard_forks: ?HardForks, alloc: Allocator) u16 {
+        var hash = genesis_hash;
+        if (maybe_hard_forks) |hard_forks| {
+            for (hard_forks.get_forks()) |hard_fork| {
+                var buf: [16]u8 = undefined;
```
0xNineteen, line 33: can we define this outside the loop -- 
```
var buf: [16]u8 = undefined;
for (hard_forks.get_forks()) |hard_fork| {
```

src/core/shred.zig
```diff
@@ -11,4 +12,52 @@ pub const ShredVersion = struct {
...
+
+        while (chunks.next()) |chunk| {
+            accum[0] ^= chunk[0];
+            accum[1] ^= chunk[1];
+        }
+        var version = (@as(u16, accum[0]) << 8) | accum[1];
+        return version +| 1;
+    }
+
+    pub fn compute_shred_version(genesis_hash: Hash, maybe_hard_forks: ?HardForks, alloc: Allocator) u16 {
```
0xNineteen, line 29: (small nit) if `Allocator` is a fcn parameter, can we make it the first parameter

src/core/hard_forks.zig
```diff
@@ -0,0 +1,85 @@
...
+
+    pub fn deinit(self: *Self) void {
+        self.hard_forks.deinit();
+    }
+
+    pub fn get_forks(self: *const Self) []HardFork {
+        return self.hard_forks.items;
+    }
+
+    pub fn get_hash_data(self: *Self, slot: Slot, parent_slot: Slot) ?[8]u8 {
```
0xNineteen, line 40: can we change this to `*const Self` - all fcns which dont modify the ptr should be const

0xNineteen (changes requested):

src/core/shred.zig
```diff
@@ -11,4 +12,56 @@ pub const ShredVersion = struct {
...
+    }
+
+    pub fn compute_shred_version(alloc: Allocator, genesis_hash: Hash, maybe_hard_forks: ?HardForks) u16 {
+        var hash = genesis_hash;
+        if (maybe_hard_forks) |hard_forks| {
+            var buf: [16]u8 = undefined;
+            for (hard_forks.get_forks()) |hard_fork| {
+                std.mem.writeIntLittle(u64, buf[0..8], hard_fork.slot);
+                std.mem.writeIntLittle(u64, buf[8..], @as(u64, hard_fork.count));
+                hash = Hash.extend_and_hash(genesis_hash, &buf, alloc) catch genesis_hash;
```
0xNineteen, line 36: the only way this fcn fails (and catch is called) is if it runs out of memory, right? - in this case, we should propagate the out of memory error instead of catching

also, the first parameter should be `hash`

src/core/hash.zig
```diff
@@ -46,4 +49,13 @@ pub const Hash = struct {
         }
         return CompareResult.Equal;
     }
+
+    pub fn extend_and_hash(id: Hash, val: []u8, alloc: Allocator) Allocator.Error!Self {
```
0xNineteen, line 53: (nit) alloc should be first here

src/core/shred.zig
```diff
@@ -11,4 +12,56 @@ pub const ShredVersion = struct {
     pub fn init_manually_set(version: u16) Self {
         return Self{ .value = version };
     }
+
+    pub fn version_from_hash(hash: *Hash) u16 {
```
0xNineteen, line 16: should be `*const`

src/core/shred.zig
anoushk1234: I will add more tests to make sure the calculation is correct

0xNineteen (approved):

ultd (approved):

PR #32, opened 2023-09-07
gossip: add network entrypoints

0xNineteen:
- add contributing.md file 
- fix `Bloom` serialization to match rust 
- add entrypoints cli args 
- send full crds filters to a random entrypoints if dont have their contact info
- add code to handle pong messages (and fix ping_cache)
- ipv4 address parsing
- remove unused code


src/gossip/ping_pong.zig
```diff
@@ -146,11 +144,13 @@ pub const PingCache = struct {
...
         var peer_and_addr = newPubkeyAndSocketAddr(pong.from, socket);
-        if (self.pending_cache.peek(pong.hash)) |val| {
-            if (val == peer_and_addr) {
-                self.pings.pop(&peer_and_addr);
-                self.pongs.put(peer_and_addr, now);
-                self.pending_cache.pop(&pong.hash);
+        if (self.pending_cache.peek(pong.hash)) |pubkey_and_addr| {
+            const pubkey: Pubkey = pubkey_and_addr[0];
+            const addr: SocketAddr = pubkey_and_addr[1];
+            if (pubkey.equals(&pong.from) and addr.eql(&socket)) {
```
0xNineteen, line 150: std.meta.eql

src/gossip/net.zig
```diff
@@ -16,6 +16,64 @@ pub const SocketAddr = union(enum(u8)) {
         },
     };
 
+    pub fn parse(bytes: []const u8) !Self {
+        // TODO: parse v6 if v4 fails
+        return parse_v4(bytes);
```
0xNineteen, line 21: ` network.Address.IPv4.parse(`

src/gossip/ping_pong.zig
0xNineteen: - may lead to incorrect result (ie, if one pubkey has `cache_str = null` and another has `cache_str` set it will fail

src/gossip/net.zig
0xNineteen: leads to unexpected results: 
```zig 
        var addr = "127.0.0:1123";
        var result = SocketAddr.parse_v4(addr); // succeeds with IP{127, 0, 0, 0} when should fail
```

ultd (approved):

PR #36, opened 2023-09-11
core: add benchmarking code

0xNineteen:
- adds benchmarking logic using `zig build benchmark -- {filter}`
- add style guide info in `contributing.md`
- main logic in `src/benchmarks.zig` binary
- clean up packet building logic in gossip_service - `crds_values_to_packets`
  - use new logic in fuzz testing - (which is also used in benchmarking code) 
- add benchmarks for: sockets and gossip service (in corresponding files) 
- `src/trace/log.zig`: add generics for sink types (add null sink - used in benchmarking cod) + remove dead `consume_entry` fcn 
---
- change method/fcn names to camelCase
- add channel benchmark 
- change benchmarking code to use milliseconds

### DIFF
--- a/src/trace/log.zig
+++ b/src/trace/log.zig
@@ -125,7 +125,6 @@ pub const Logger = union(enum) {
 
 pub const StandardErrLogger = struct {
     allocator: std.mem.Allocator,
-    arena: std.heap.ArenaAllocator,
     default_level: Level,
     exit_sig: AtomicBool,
     handle: ?std.Thread,
@@ -135,11 +134,8 @@ pub const StandardErrLogger = struct {
 
     pub fn init(allocator: std.mem.Allocator, default_level: Level) *Self {
         var self = allocator.create(Self) catch @panic("could not allocator.create Logger");
-        var arena = std.heap.ArenaAllocator.init(allocator);
-
         self.* = .{
             .allocator = allocator,
-            .arena = arena,
             .default_level = default_level,
             .exit_sig = AtomicBool.init(false),
             .handle = null,
@@ -159,7 +155,6 @@ pub const StandardErrLogger = struct {
             handle.join();
         }
         self.channel.deinit();
-        self.arena.deinit();
         self.allocator.destroy(self);
     }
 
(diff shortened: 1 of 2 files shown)


src/trace/log.zig
```diff
@@ -17,19 +17,31 @@ pub const Logger = struct {
...
+    entry_sink: EntrySink,
 
     const Self = @This();
 
-    pub fn init(allocator: std.mem.Allocator, default_level: Level) *Self {
+    pub fn init(allocator: std.mem.Allocator, default_level: Level, entry_sink: ?EntrySink) *Self {
         var self = allocator.create(Self) catch @panic("could not allocator.create Logger");
         var arena = std.heap.ArenaAllocator.init(allocator);
+
+        var sink = blk: {
```
0xNineteen, line 28: change to if comptime var on info/debug/... calls

src/trace/log.zig
```diff
@@ -107,10 +117,25 @@ pub const Logger = struct {
     }
 };
 
-const BasicStdErrSink = struct {
+const EntrySink = struct {
```
0xNineteen, line 120: change to Sink

src/gossip/socket_utils.zig
```diff
@@ -18,79 +21,157 @@ pub fn read_socket(
...
     }
-    std.debug.print("send_socket loop closed\n", .{});
+    logger.debugf("send_socket loop closed\n", .{});
 }
 
-// // TODO: fix
-// test "gossip.socket_utils: sending a packet" {
-//     var allocator = std.testing.allocator;
-//     var addr = SocketAddr.init_ipv4(.{ 127, 0, 0, 1 }, 9999);
+pub const benchmark_packet_processing = struct {
```
0xNineteen, line 73: benchmark channel

src/benchmarks.zig
```diff
@@ -0,0 +1,226 @@
...
+
+    if (std.mem.startsWith(u8, "gossip", filter)) {
+        try benchmark(@import("gossip/gossip_service.zig").benchmark_message_processing);
+    }
+}
+
+// src: https://github.com/Hejsil/zig-bench
+pub fn benchmark(comptime B: type) !void {
+    const args = if (@hasDecl(B, "args")) B.args else [_]void{{}};
+    const arg_names = if (@hasDecl(B, "arg_names")) B.arg_names else [_]u8{};
```
0xNineteen, line 45: format to ms

src/trace/log.zig
0xNineteen: - seems you cant change comptime values on structs (eg, example below doesnt compile -- error: `value stored in comptime field does not match the default value of the field`) 
```zig 
const std = @import("std");

const Foo = struct { 
  comptime do: bool = true,

  pub fn foo(self: @This()) void { 
    if (self.do) std.debug.print("hi", .{}); 
  }

  pub fn doNothing(self: *@This()) void { 
    self.do = false;
  }
};

pub fn main() void {
  var foo = Foo{}; 
  foo.foo(); 
  foo.doNothing(); 
  foo.foo(); 
}

```

src/bincode/bincode.zig
```diff
@@ -123,7 +123,7 @@ pub fn Deserializer(comptime Reader: type) type {
             return try visitor.visitBool(ally, De, value);
         }
 
-        const custom_dser = struct {
+        const customDeser = struct {
```
ultd, line 126: Should be Pascal case `CustomDeser`

src/bincode/bincode.zig
```diff
@@ -417,7 +417,7 @@ pub fn Serializer(
             }
         }
 
-        const custom_ser = struct {
+        const customSer = struct {
```
ultd, line 420: Should be Pascal case `CustomDeser`

ultd (changes requested):

0xNineteen:
waiting for endpoint PR to merge first - then this should be ok to merge

ultd (approved):

src/benchmarks.zig
```diff
@@ -0,0 +1,272 @@
...
+            max_time_per_bench,
+            TimeUnits.milliseconds,
+        );
+    }
+
+    if (std.mem.startsWith(u8, "gossip", filter)) {
+        try benchmark(
+            @import("gossip/gossip_service.zig").BenchmarkMessageProcessing,
+            max_time_per_bench,
+            TimeUnits.milliseconds,
```
jbuckmccready, line 46: I think it would be nice to have the benchmarking auto select the time units (by repeated dividing by 1000 to determine magnitude of nanoseconds) rather than explicitly passing it in - not a big deal either way but might be nice as the number of benchmarks grow.

PR #39, opened 2023-10-09
gossip: optimize logic

0xNineteen:
`lru.zig`
- remove inner mux from lru - PingCache is already wrapped in Mux() so its ok
- fix memleak in pop()

`crds_table.zig`
- add threadpool to table (used to find old labels to prune)

`fuzz.zig`
- change to indep binary which targets a specific port (can fuzz both rust + sig) 

`gossip_service.zig`
- change channels to Arraylist(Packet) vs Packet (batched)
- add threadpool 
  - used in packet verification, processing pull requests
- add runSpy (which just processes incoming messages) 
- batch the message processing
    - change methods to process them in batches too (handleBatchPullResponses)
- change `std.Arraylist` to `Arraylist`
- add logic to filter values based on shred version (`filterCrdsValuesBasedOnShredVersion`)

`socket_utils.zig`
- update logic of reading/sending socket data to do so in batches 

`ping_pong.zig`
- change filter to return a list of `usize` indexs of valid peers instead of copying contact-info 

`threadpool.zig`
- add threadpool logic


src/common/lru.zig
```diff
@@ -14,7 +14,6 @@ pub fn LruCache(comptime K: type, comptime V: type) type {
         dbl_link_list: TailQueue(LruEntry),
         max_items: usize,
         len: usize = 0,
-        mux: std.Thread.Mutex,
```
0xNineteen, line 17: change comment of 'thread-safe' LRU

snawaz: Hi @0xNineteen. I just stumbled upon this project few days back. It's an interesting project and Zig is a greattt language!

Just wondering...  why don't you use policy-based design to implement `LruCache` both as thread-safe and thread-unsafe? It's zero-cost to have this flexibility.

I mean, in this way:

```zig
const NullMutex = struct {
    const Self = @This();
    pub fn lock(_: *Self) void {}
    pub fn unlock(_: *Self) void {}
};

/// A thread-unsafe LRU Cache
pub fn LruCache(comptime K: type, comptime V: type) type {
    return LruCacheImpl(NullMutex, K, V);
}

/// A thread-safe LRU Cache
pub fn SafeLruCache(comptime K: type, comptime V: type) type {
    return LruCacheImpl(std.Thread.Mutex, K, V);
}

// private impl... 
fn LruCacheImpl(comptime Mutex: type, comptime K: type, comptime V: type) type {
      // the actual impl that uses Mutex as policy.
}
```

For the unsafe LruCache, the `NullMutex` lock/unlock calls would be optimized away (and removed) by the compiler.

snawaz:
@0xNineteen. Posted a comment. Not sure if you got the notification for that, as the comment shows _resolved_.

src/common/lru.zig
0xNineteen: hey @snawaz - thanks for the comment! - rn we only use the LRU in the ping-cache which is already wrapped in a Mux so each LRU in the cache is already thread-safe using this approach

0xNineteen:
@snawaz thanks for the tag - missed the comment originally - answered it in the comment

src/common/lru.zig
snawaz: Yes, you have mentioned that in the description itself, which I understood. I was just saying you get both thread-safe and thread-unsafe versions without paying any extra cost. It _could_ be useful in other scenarios (in the future) where you need thread-safety. Anyway, it's up to you. The code is otherwise great and I loved the project.

0xNineteen: yup agreed its another good soln which could be useful in future scenarios - thanks for the complements and the comment! :)

src/gossip/fuzz.zig
```diff
@@ -122,7 +126,7 @@ pub fn randomPushMessage(rng: std.rand.Random, keypair: *const KeyPair, to_addr:
         &to_addr,
         ChunkType.PushMessage,
     );
-    return packets;
+    return packets.?;
```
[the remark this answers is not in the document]
0xNineteen: in this case, it will never be null bc crds_values is not empty (only case it returns null) - but yeah i can see the case for using an empty array - should probs decide what the approach should be going forward

src/gossip/gossip_service.zig
```diff
@@ -250,254 +296,350 @@ pub const GossipService = struct {
...
+                push_messages.clearRetainingCapacity();
+            }
 
-                    var crds_table: *CrdsTable = crds_table_lock.mut();
-                    crds_table.attemptTrim(CRDS_UNIQUE_PUBKEY_CAPACITY) catch |err| {
-                        self.logger.warnf("error trimming crds table: {s}", .{@errorName(err)});
-                    };
-                }
+            if (prune_messages.items.len > 0) {
+                var x_timer = std.time.Timer.start() catch unreachable;
```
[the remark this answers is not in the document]
0xNineteen: yeah we typically avoid unreachable unless its guaranteed to not fail (in the reasonable case) - in this can we should probs change it to a try

src/gossip/gossip_service.zig
```diff
@@ -600,8 +732,13 @@ pub const GossipService = struct {
...
             break :blk result;
         };
-        var valid_gossip_peers = ping_cache_result.valid_peers;
-        defer valid_gossip_peers.deinit();
+        var valid_gossip_indexs = ping_cache_result.valid_peers;
+        defer valid_gossip_indexs.deinit();
+
+        var valid_gossip_peers: [NUM_ACTIVE_SET_ENTRIES]LegacyContactInfo = undefined;
+        for (valid_gossip_indexs.items) |i| {
+            valid_gossip_peers[i] = gossip_peers[i];
```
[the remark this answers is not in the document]
0xNineteen: nice catch - fixed

src/sync/channel.zig
```diff
@@ -37,18 +39,34 @@ pub fn Channel(comptime T: type) type {
...
                 return error.ChannelClosed;
             }
-            var buffer = self.buffer.lock();
-            defer buffer.unlock();
-            try buffer.mut().append(value);
-            self.hasValue.signal();
+            var buffer_lock = self.buffer.lock();
+            defer buffer_lock.unlock();
+
+            var buffer: *std.ArrayList(T) = buffer_lock.mut();
```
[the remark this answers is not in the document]
0xNineteen: readability + you dont get autocomplete for just .mut() since it doesnt figure out the type - in this case its not a big deal since its just `append()`, but in other situations the autocomplete is really nice to have - so we just keep it consistent here

src/gossip/gossip_service.zig
```diff
@@ -250,254 +296,350 @@ pub const GossipService = struct {
...
+                        });
                     },
                 }
+            }
 
-                {
-                    var crds_table_lock = self.crds_table_rw.write();
-                    defer crds_table_lock.unlock();
+            // handle batch messages
+            if (push_messages.items.len > 0) {
```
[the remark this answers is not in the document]
0xNineteen: yeah i think its ok for now

src/gossip/crds_table.zig
```diff
@@ -565,25 +581,35 @@ pub const CrdsTable = struct {
...
+        // standard
+        task: Task = .{ .callback = callback },
+        done: std.atomic.Atomic(bool) = std.atomic.Atomic(bool).init(false),
+
+        pub fn deinit(self: *GetOldLabelsTask) void {
+            self.old_labels.deinit();
+        }
+
+        pub fn callback(task: *Task) void {
+            var self = @fieldParentPtr(@This(), "task", task);
```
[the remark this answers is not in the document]
0xNineteen: yeah i tried it out - the intertwined-ness of it made it less clear what was going on compared to the pros of saving a few lines - would probs be good to re-visit if our tasks become more complex

dnut (approved):

src/gossip/crds_table.zig
```diff
@@ -592,15 +618,60 @@ pub const CrdsTable = struct {
...
+            tasks[i] = GetOldLabelsTask{
+                .key = key,
+                .crds_table = self,
+                .cutoff_timestamp = cutoff_timestamp,
+                .old_labels = old_labels,
+            };
+
+            // run it
+            const batch = Batch.from(&tasks[i].task);
+            ThreadPool.schedule(self.thread_pool, batch);
```
0xNineteen, line 656: change this to use `self.thread_pool`

src/gossip/crds_table.zig
```diff
@@ -592,15 +618,60 @@ pub const CrdsTable = struct {
...
+        }
+
+        // wait for them to be done to release the lock
+        var output_length: u64 = 0;
+        for (tasks) |*task| {
+            while (!task.done.load(std.atomic.Ordering.Acquire)) {
```
0xNineteen, line 662: mutex + cv approach

ultd (approved):

PR #62, opened 2023-12-22
cli options for spy node and log level

dnut:
- Add cli option to enable spy node so the bool doesn't need to be edited in the code
- Add cli option for log level to customize the granularity of log messages to log
- Change some log messages that indicate errors to log at the error instead of debug level
- Remove redundant newlines from some log messages in gossip

## Spy node

I made it default to *not* run as a spy node. You'll need to specify the flag to run as a spy node. This is contrary to the current behavior, which runs as a spy node (unless you edit the source code). I did this because it lent itself to the most intuitive command-line interface. It seemed a little confusing to have a CLI flag that only served to *disable* spy_node mode. It's not clear to me what name to give to a normal gossip node that isn't spying.

It's also not entirely clear to my why you would want to run a spy node. Whenever I run gossip, I set spy_node to false. As a spy node, you never send outgoing packets unless someone else connects to you first and you are responding to them. No one will know you're in the network unless you use some external channel to communicate your IP address. It just seems like an unlikely use case, but maybe I'm missing something.

If you'd like me to change the default to run as a spy node, please help me come up with an intuitive name for the cli flag that enables all outgoing communication.
- `--not-spy-node`
- `--full-node`
- ...?

## Logs

To implement the log filter, I added log level checks in both Logger and Entry. Checking the level only in Entry would be sufficient to filter all messages correctly, but the performance is not optimal. Checking in Logger is faster, but does not cover all cases. So I added both checks.

1. When calling a Logger's log method, it can check if it needs to log anything, and exit if not. This is optimal because it can avoid allocating an Entry.
2. Another approach to logging is by calling Logger.field(), which returns an Entry, and then calling a log method on the entry. When calling Logger.field, this method does not know what level you intend to log at, so it needs to allocate the Entry no matter what. Then when you call the log method in Entry, the Entry finally knows what log level you intend to log at, and it can exit without logging.

### DIFF
--- a/src/cmd/cmd.zig
+++ b/src/cmd/cmd.zig
@@ -31,6 +31,15 @@ var gossip_entrypoints_option = cli.Option{
     .value_name = "Entrypoints",
 };
 
+var gossip_spy_node_option = cli.Option{
+    .long_name = "spy-node",
+    .help = "run as a gossip spy node (minimize outgoing packets)",
+    .short_alias = 's',
+    .value = cli.OptionValue{ .bool = false },
+    .required = false,
+    .value_name = "Spy Node",
+};
+
 var app = &cli.App{
     .name = "sig",
     .description = "Sig is a Solana client implementation written in Zig.\nThis is still a WIP, PRs welcome.",
@@ -52,6 +61,7 @@ var app = &cli.App{
         , .action = gossip, .options = &.{
             &gossip_port_option,
             &gossip_entrypoints_option,
+            &gossip_spy_node_option,
         } },
     },
 };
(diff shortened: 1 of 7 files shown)


src/cmd/cmd.zig
```diff
@@ -31,11 +33,30 @@ var gossip_entrypoints_option = cli.Option{
     .value_name = "Entrypoints",
 };
 
+var gossip_spy_node_option = cli.Option{
+    .long_name = "spy-node",
+    .help = "run as a gossip spy node (minimize outgoing packets)",
+    .short_alias = 's',
```
ultd: remove

ultd (approved):

PR #64, opened 2024-01-03
fix(gossip): Add `extensions` field to ContactInfo

### LINKED ISSUE #61: Running in gossip mode on mainnet results in failed to verify protocol message signature InvalidPullRequest
### Description

Building from tip for tree and running in gossipmode against mainnet results in:

`time=2023-12-21T17:11:34Z level=debug  failed to verify protocol message signature InvalidPullResponse `

Not sure if this is expected based on current state of codebase or not hence filing issue

### How to Reproduce the Bug

1. Clone 949eb15dbecc1cbd091ccf8344aa5c65b16278a8
2. Follow instructions from new (much better!) readme.md to run against mainnet
3. Result: above error message.

### Additional Context

_No response_

dnut:
fixes #61

https://github.com/solana-labs/solana/pull/32309 added a new field to ContactInfo called `extensions`. The field must be supported for serialization because it adds an extra byte to the serialization format to represent the length of the vec. The type contained in the vec is 0-sized, so that's the only data it adds.

When a pull response is received from another node, the message is deserialized to extract the signature and the pubkey, and then it is re-serialized to validate the signature+pubkey against the serialized version of the message. 

Since we're missing the `extensions` field, when re-serializing the ContactInfo, it is missing that last byte. Since the signature was created for a ContactInfo including that byte, our signature verification process fails. This was causing the bug reported in #61.

This change adds the `extensions` field as an ArrayList(void) so that the struct de/serializes correctly. I've selected `void` as the type to keep it simple. `void` is the unit type in zig, and the `Extensions` enum in rust is also a unit type since it has no variants. If any extension variants do get added in rust, we'll need a change to address that regardless.

This change also makes the log message more descriptive when a signature verification fails

### DIFF
--- a/src/gossip/node.zig
+++ b/src/gossip/node.zig
@@ -51,11 +51,13 @@ pub const ContactInfo = struct {
     version: Version,
     addrs: ArrayList(IpAddr),
     sockets: ArrayList(SocketEntry),
+    extensions: ArrayList(void) = ArrayList(void).init(undefined),
     cache: [SOCKET_CACHE_SIZE]SocketAddr = socket_addrs_unspecified(),
 
     pub const @"!bincode-config:cache" = bincode.FieldConfig([SOCKET_CACHE_SIZE]SocketAddr){ .skip = true };
     pub const @"!bincode-config:addrs" = ShortVecArrayListConfig(IpAddr);
     pub const @"!bincode-config:sockets" = ShortVecArrayListConfig(SocketEntry);
+    pub const @"!bincode-config:extensions" = ShortVecArrayListConfig(void);
     pub const @"!bincode-config:wallclock" = var_int_config_u64;
 
     const Self = @This();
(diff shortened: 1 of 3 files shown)


src/gossip/node.zig
```diff
@@ -51,11 +51,13 @@ pub const ContactInfo = struct {
     version: Version,
     addrs: ArrayList(IpAddr),
     sockets: ArrayList(SocketEntry),
+    extensions: ArrayList(void) = ArrayList(void).init(undefined),
```
0xNineteen, line 54: would probably be better to have the same empty enum in rust client so its more clear

0xNineteen (commented):
another thing to note - rn crds_table only stores **legacy** contact_infos - if the new `ContactInfo` structs are being pushed - well probs need to update some of the crds_table logic in the future to store/retrieve these new values (added a task on the board)

src/gossip/node.zig
dnut: Sounds good. In my mind, an empty enum or struct is basically a newtype for void, since they are both unit types. But I can see void being confusing since it doesn't explain anything to a reader.

The only problem is getting this to compile. `const Extension = union(enum) {};` triggers a compilation error that I can't make sense of and may be a compiler bug. `const Extension = enum {};` and `const Extension = union {};` seem to reveal limitations in the serialization logic. With `const Extension = void;` it compiles no problem, and I can add a doc explaining its purpose.

0xNineteen (approved):

ultd (approved):

Issue #65, opened 2024-01-04
Support New ContactInfo Structs


0xNineteen:
### Problem

rn crds_table only stores legacy contact_infos - but recently, new ContactInfo structs are being pushed through gossip which we arent looking at

### Proposal

well need to update some of the crds_table logic to store/retrieve these new values 
- [ ] store new `ContactInfo` struct in crds table 
- [ ] move `node.zig` into `crds.zig`
- [ ] decide if we should also depreciate `LegacyContactInfos`

PR #74, opened 2024-02-27
feat(gossip): minimal ContactInfo support

dnut:
Supports ContactInfo messaging but continues to use LegacyContactInfo internally

fixes https://github.com/Syndica/sig/issues/65

Previously, ContactInfo was not properly supported. It could be deserialized from incoming messages, stored in the CrdsTable, and sent out to other nodes. However, Sig did not send its own contact info as ContactInfo. It only used LegacyContactInfo for this purpose. Also, when identifying other gossip nodes, Sig only ever looked for LegacyContactInfo.

With this change, Sig stores its own contact info internally as a ContactInfo, and publishes it as both LegacyContactInfo and ContactInfo. When identifying other gossip nodes from the CrdsTable, it looks for both ContactInfo and LegacyContactInfo.

The logic that interprets these contact infos continues to use LegacyContactInfo, which means all ContactInfo instances are converted to LegacyContactInfo for the internal logic.


ultd:
#72 solves this

PR #93, opened 2024-03-21
feat(accounts-db): snapshot download from gossip

0xNineteen:
this uses gossip data to find peers which have a snapshot and trys to download it from them 

this also moves the accounts-db logic into the validator command 

and adds a new cli argument `download_snapshot` (which is self-descriptive)

--- 

another change is to reduce the rate of pull requests - from 200ms (on the right) to every 5seconds (on the left) - we see an improvement in how fast we recv all the msgs we need (likely we were sending too many pull requests with redundant data) 

image


src/accountsdb/download.zig
```diff
@@ -0,0 +1,365 @@
...
+                }
+
+                // dont try to download from a slow peer
+                for (slow_peer_pubkeys.items) |slow_peer| {
+                    if (slow_peer.equals(&ci.pubkey)) {
+                        is_slow_count += 1;
+                        continue :search_loop;
+                    }
+                }
+
```
0xNineteen, line 104: add entrypoints + trusted validators cli args

src/accountsdb/download.zig
```diff
@@ -0,0 +1,652 @@
...
+        var max_inc_hash: ?SlotAndHash = null;
+        for (trusted_hashes.incremental) |inc_hash| {
+            if (max_inc_hash == null or inc_hash.slot > max_inc_hash.?.slot) {
+                max_inc_hash = inc_hash;
+            }
+        }
+
+        // track the full and incremental
+        var r = try trusted_snapshot_hashes.getOrPut(trusted_hashes.full);
```
dnut, line 126: This inserts an undefined value into the hashmap. If there are no incremental snapshots, it remains undefined.

src/accountsdb/download.zig
```diff
@@ -0,0 +1,652 @@
...
+            .inc_snapshot = max_inc_hash,
+        });
+    }
+    result.is_valid = valid_peers.items.len;
+
+    return result;
+}
+
+/// uses swap remove to remove peers that do not have a hash which is trusted
+pub fn removeUnTrustedSnapshotPeers(
```
dnut, line 99: It probably doesn't matter, but my instinct would be to have this function directly return the trusted_snapshot_hashes, and then pass that into findPeersToDownloadFromAssumeCapacity, where it would be used as a whitelist. Then you wouldn't need to iterate over the list after the fact and remove items.

src/accountsdb/download.zig
```diff
@@ -0,0 +1,652 @@
...
+
+    // get the hashes of trusted validators
+    var trusted_count: usize = 0;
+    for (trusted_validators.items) |trusted_validator| {
+        const gossip_data = table.get(.{ .SnapshotHashes = trusted_validator }) orelse continue;
+        const trusted_hashes = gossip_data.value.data.SnapshotHashes;
+        trusted_count += 1;
+
+        var max_inc_hash: ?SlotAndHash = null;
+        for (trusted_hashes.incremental) |inc_hash| {
```
dnut, line 119: Why not insert all of these into the set of incremental hashes?

Alternatively, why not abandon the set entirely, and restrict each full snapshot to be associated with only a single max incremental across all trusted validators?

The current behavior is like a medium between the two extremes and I'm curious what the goal is of finding per-validator maxes.

src/accountsdb/download.zig
```diff
@@ -0,0 +1,652 @@
...
+
+    // remove peers that do not have a trusted hash
+    var peer_index: usize = 0;
+    for (0..available_snapshot_peers.items.len) |_| {
+        const peer = available_snapshot_peers.items[peer_index];
+        var should_remove = false;
+        // full snapshot must be trusted
+        if (trusted_snapshot_hashes.getEntry(peer.full_snapshot)) |entry| {
+            // if we have an incremental snapshot
+            if (peer.inc_snapshot) |inc_snapshot| {
```
dnut, line 148: This trusts the peer to decide whether it needs to bring along an incremental snapshot. It only restricts the peer's incremental snapshot if it offers one. What if all the trusted validators specify incremental snapshots, but this peer specifies none?

src/accountsdb/download.zig
```diff
@@ -0,0 +1,652 @@
...
+    try easy.setWritedata(&download_progress);
+    try easy.setWritefunction(DownloadProgress.bufferWriteCallback);
+
+    var resp = easy.perform() catch return false;
+    defer resp.deinit();
+
+    const full_download = download_progress.file_memory_index == download_size;
+    if (!full_download) {
+        // too slow = early exit
+        return false;
```
dnut, line 461: This feels like an error. Is it conceivable that this would be acceptable? For example, maybe the download can be resumed, or the partial file could be useful somehow.

src/accountsdb/download.zig
```diff
@@ -0,0 +1,652 @@
...
+        };
+    }
+
+    pub fn bufferWriteCallback(ptr: [*c]c_char, size: c_uint, nmemb: c_uint, user_data: *anyopaque) callconv(.C) c_uint {
+        const len = size * nmemb;
+        var self: *Self = @alignCast(@ptrCast(user_data));
+        var typed_data: [*]u8 = @ptrCast(ptr);
+        var buf = typed_data[0..len];
+
+        @memcpy(self.mmap[self.file_memory_index..][0..len], buf);
```
dnut, line 353: This looks a little weird to me. Seems fine, it's just not what I'm used to. Normally I would write it like this:
```suggestion
        @memcpy(self.mmap[self.file_memory_index .. self.file_memory_index + len], buf);
```
Do you know if it makes a difference? I'm not sure what the compiler optimizes away. My assumption is that it would be slightly faster to do the addition rather than slicing twice.

src/cmd/cmd.zig
```diff
@@ -223,28 +268,41 @@ var app = &cli.App{
...
+                &disk_index_path_option,
+                &force_unpack_snapshot_option,
+                &min_snapshot_download_speed_mb_option,
+                &force_new_snapshot_download_option,
             },
         },
         &cli.Command{
-            .name = "accounts_db",
-            .help = "run accounts_db",
+            .name = "download_snapshot",
```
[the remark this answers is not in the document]
0xNineteen: f06580e

src/cmd/cmd.zig
```diff
@@ -558,29 +760,54 @@ fn accountsDb(_: []const []const u8) !void {
...
+            trusted_validators,
+            gossip_service orelse return error.SnapshotsNotFoundAndNoGossipService,
+            snapshot_dir_str,
+            @intCast(min_mb_per_sec),
+        );
+        maybe_snapshot_files = try SnapshotFiles.find(allocator, snapshot_dir_str);
+    }
+
+    // SAFE: this unwrap is safe because we either found the snapshot or downloaded it
+    var snapshot_files = maybe_snapshot_files.?;
```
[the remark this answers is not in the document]
0xNineteen: great point - c46360f

src/accountsdb/download.zig
```diff
@@ -0,0 +1,652 @@
...
+                    inc_snapshot_url,
+                    output_dir,
+                    inc_snapshot_filename,
+                    // NOTE: no min limit (we already downloaded the full snapshot at a good speed so this should be ok)
+                    null,
+                ) catch |err| {
+                    // failure here is ok (for now?)
+                    logger.infof("failed to download inc_snapshot: {s}", .{@errorName(err)});
+                    return;
+                };
```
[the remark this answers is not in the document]
0xNineteen: i feel like deduplicating this logic would add more complexity than the benefit you would gain from keeping it separate - its pretty clear the steps are 1 - alloc the filename, 2 - alloc the url, 3 - download the file

src/accountsdb/download.zig
0xNineteen: not sure the difference but the double-slice approach seems to be the way stdlib does it

src/accountsdb/download.zig
0xNineteen: incremental snapshots are always associated with a base_slot which corresponds to a full snapshot - so we cant have a single incremental across all validators (different base_slots wouldnt be valid)

it does make sense to include all the incrementals though (which the rust client isnt doing) - eg, with [full_slot, inc_slot] notion: we could have a trusted validator with [19, {19, 12}] and a valid validator download could be [19, 12] 💯 - will make the change

0xNineteen: 4c588e4

src/accountsdb/download.zig
0xNineteen: if this peer specifies none then we wont attempt to download it from them so its ok -- see `// no incremental snapshot, thats ok` comment

src/accountsdb/download.zig
0xNineteen: 4c588e4

src/accountsdb/download.zig
0xNineteen: agreed error is more accurate - 507e95b

src/cmd/cmd.zig
```diff
@@ -335,6 +402,101 @@ fn validator(_: []const []const u8) !void {
...
+    var status_cache = try StatusCache.init(gpa_allocator, status_cache_path);
+    defer status_cache.deinit();
+
+    var slot_history = try accounts_db.getSlotHistory();
+    defer slot_history.deinit(accounts_db.allocator);
+
+    const bank_slot = snapshot.bank_fields.slot;
+    try status_cache.validate(gpa_allocator, bank_slot, &slot_history);
+
+    logger.infof("accounts-db setup done...", .{});
```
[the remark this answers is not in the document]
0xNineteen: agreed - not a big fan of the global gpa & moving things out when it gets too much makes sense 

2fe0382 - not sure how to do it better so open to ideas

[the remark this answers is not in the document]
dnut: Nice. I only had a simple change in mind, which was throwing the whole highlighted region into a single function. But I like your change better, it's more organized.

I'm trying to be careful with the overall length of the validator function, and this file as a whole. During rapid development, main files/functions can get extremely long, with the code doing all kinds of things in random places, and this can be a problem for readability and maintainability. It's OK for now though.

src/cmd/cmd.zig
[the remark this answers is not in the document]
dnut: FYI, In zig 0.12 it becomes a hard error to make something a var when it can be const. This is going to impact many lines of code when we upgrade. I don't think you need to change anything here but I'm just mentioning it because I see you used var for maybe_snapshot_files.

I used to feel that curly braces should *always* be used with if statements, to provide clarity and to avoid mistakes. But auto-formatters and linters can do almost the same thing. Also there are some cases in zig where omitting braces has readability advantages, like avoiding the whole block labeling + `break` pattern.

src/accountsdb/download.zig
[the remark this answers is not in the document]
dnut: makes sense

src/gossip/service.zig
```diff
@@ -56,8 +56,9 @@ const Gauge = @import("../prometheus/gauge.zig").Gauge;
 const PacketBatch = ArrayList(Packet);
 const GossipMessageWithEndpoint = struct { from_endpoint: EndPoint, message: GossipMessage };
 
-pub const GOSSIP_PULL_TIMEOUT_MS: u64 = 15000;
-pub const GOSSIP_PUSH_MSG_TIMEOUT_MS: u64 = 30000;
+pub const GOSSIP_PULL_RATE_MS: u64 = 5 * std.time.ms_per_s;
```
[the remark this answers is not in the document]
dnut: My mistake, I read this as 5 milliseconds. 5 seconds seems pretty infrequent. Are you sure that's sufficient? There are ways to reduce logging noise without changing how gossip works.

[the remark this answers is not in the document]
0xNineteen: great point - updated the description with extra validation on this change

dnut (approved):

PR #105, opened 2024-03-23
feat(gossip): resolve entrypoint(s) dns records

ultd:
Previously, we needed to resolve entrypoint DNS manually and pass IPs to the startup command like so: 

```sh
sig gossip --entrypoint 34.83.231.102:8001 \
    --entrypoint 145.40.67.83:8001 \
    --entrypoint 147.75.38.117:8001 \
    --entrypoint 145.40.93.177:8001 \
    --entrypoint 86.109.15.59:8001
```

Now, we can pass the domains directly and Sig will resolve DNS:

```sh
sig gossip --entrypoint entrypoint.mainnet-beta.solana.com:8001 \
    --entrypoint entrypoint2.mainnet-beta.solana.com:8001 \
    --entrypoint entrypoint3.mainnet-beta.solana.com:8001 \
    --entrypoint entrypoint4.mainnet-beta.solana.com:8001 \
    --entrypoint entrypoint5.mainnet-beta.solana.com:8001
```

This PR adds new dependency on [`zigdig`](https://github.com/lun-4/zigdig/) which is only used to resolve DNS. I've reviewed the code and its safe.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -71,15 +71,15 @@ To run the debugger, you need a run configuration. This launch.json includes an 
       "args": [
         "gossip",
         "--entrypoint",
-        "34.83.231.102:8001",
+        "entrypoint.mainnet-beta.solana.com:8001",
         "--entrypoint",
-        "145.40.67.83:8001",
+        "entrypoint2.mainnet-beta.solana.com:8001",
         "--entrypoint",
-        "147.75.38.117:8001",
+        "entrypoint3.mainnet-beta.solana.com:8001",
         "--entrypoint",
-        "145.40.93.177:8001",
+        "entrypoint4.mainnet-beta.solana.com:8001",
         "--entrypoint",
-        "86.109.15.59:8001"
+        "entrypoint5.mainnet-beta.solana.com:8001"
       ],
       "cwd": "${workspaceFolder}",
       "preLaunchTask": "zig build"
--- a/readme.md
+++ b/readme.md
@@ -21,6 +21,7 @@ _Sig_ is a Solana validator client implemented in Zig. Read the [introductory bl
 ## 📋 Setup
 
 ### Build Dependencies
+
 - Zig 0.11.0 - Choose one:
   - [Binary Releases](https://ziglang.org/download/) (extract and add to PATH)
   - [Install with a package manager](https://github.com/ziglang/zig/wiki/Install-Zig-from-a-Package-Manager)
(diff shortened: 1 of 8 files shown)


src/net/echo.zig
```diff
@@ -259,10 +259,10 @@ pub fn handleRequest(
 
         // convert a u32 to Ipv4
         var socket_addr = SocketAddr.initIpv4(.{
-            @as(u8, @intCast(addr.sa.addr >> 24 & 0xFF)),
-            @as(u8, @intCast(addr.sa.addr >> 16 & 0xFF)),
-            @as(u8, @intCast(addr.sa.addr >> 8 & 0xFF)),
             @as(u8, @intCast(addr.sa.addr & 0xFF)),
```
dnut, line 262: I wonder how the test below has been working with this reversed.

It could be good to extract out this logic into a conversion function for Address->SocketAddr, and write a test for it.

0xNineteen (dismissed):

src/cmd/cmd.zig
```diff
@@ -146,11 +147,37 @@ fn gossip(_: []const []const u8) !void {
...
+                        @as(u8, @intCast(ipv4_addr & 0xFF)),
+                        @as(u8, @intCast(ipv4_addr >> 8 & 0xFF)),
+                        @as(u8, @intCast(ipv4_addr >> 16 & 0xFF)),
+                        @as(u8, @intCast(ipv4_addr >> 24 & 0xFF)),
+                    }, port);
+                };
+                break :brk value;
             };
-            try entrypoints.append(value);
+
```
dnut, line 179: I think you can simplify this by removing the nesting of `value` within `socket_addr`.

```suggestion
            var socket_addr = SocketAddr.parse(entrypoint) catch brk: {
                // if we couldn't parse as IpV4, we attempt to resolve DNS and get IP
                var domain_and_port = std.mem.splitScalar(u8, entrypoint, ':');
                const domain_str = domain_and_port.next() orelse return error.EntrypointDomainMissing;
                const port_str = domain_and_port.next() orelse return error.EntrypointPortMissing;

                // get dns address lists
                var addr_list = try dns.helpers.getAddressList(domain_str, gpa_allocator);
                defer addr_list.deinit();
                if (addr_list.addrs.len == 0) {
                    return error.EntrypointDnsResolutionFailure;
                }

                // use first A record address
                var ipv4_addr: u32 = addr_list.addrs[0].in.sa.addr;

                // parse port from string
                var port = std.fmt.parseInt(u16, port_str, 10) catch return error.EntrypointPortNotValid;

                break :brk SocketAddr.initIpv4(.{
                    @as(u8, @intCast(ipv4_addr & 0xFF)),
                    @as(u8, @intCast(ipv4_addr >> 8 & 0xFF)),
                    @as(u8, @intCast(ipv4_addr >> 16 & 0xFF)),
                    @as(u8, @intCast(ipv4_addr >> 24 & 0xFF)),
                }, port);
            };

```

dnut (commented):
Overall looks good. Just a couple minor points for possible improvement. If we're in a hurry, it should be fine to move forward with this and come back to it.

The DNS library prints out a lot of debug logs. Is there a way to disable that?

readme.md
```diff
@@ -88,24 +61,36 @@ To run the debugger, you need a run configuration. This launch.json includes an
...
+  "configurations": [
+    {
+      "type": "lldb",
+      "request": "launch",
+      "name": "Debug Gossip Mainnet",
+      "program": "${workspaceFolder}/zig-out/bin/sig",
+      "args": [
+        "gossip",
+        "--entrypoint",
+        "34.83.231.102:8001",
```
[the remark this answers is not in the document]
ultd: Good point

src/cmd/cmd.zig
```diff
@@ -146,11 +147,37 @@ fn gossip(_: []const []const u8) !void {
...
+                    // if we couldn't parse as IpV4, we attempt to resolve DNS and get IP
+                    var domain_and_port = std.mem.splitScalar(u8, entrypoint, ':');
+                    const domain_str = domain_and_port.next() orelse return error.EntrypointDomainMissing;
+                    const port_str = domain_and_port.next() orelse return error.EntrypointPortMissing;
+
+                    // get dns address lists
+                    var addr_list = try dns.helpers.getAddressList(domain_str, gpa_allocator);
+                    defer addr_list.deinit();
+                    if (addr_list.addrs.len == 0) {
+                        return error.EntrypointDnsResolutionFailure;
```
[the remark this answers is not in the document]
ultd: Makes sense

src/net/echo.zig
ultd: Well it wasn't actually testing the ip address being that in order to assert the ip, you'd need to know the ip which is a circular dependency haha

src/cmd/cmd.zig
ultd: agreed

src/net/echo.zig
ultd: On second thought, I can at the very least assert localhost ip addr - done

PR #111, opened 2024-03-24
perf(gossip): speed up handleBatchPullMessages

dnut:
## Problem

The processMessages loop typically runs pretty fast, many times per second, until a certain point when it gets bottlenecked by an overwhelming number of push messages and is unable to catch up. The iteration duration grows exponentially until it takes minutes to deal with tens of thousands of messages. This continues for several minutes, until sig reaches a degenerate state where it is no longer receiving messages. The pull request thread is unable to identify any peers. This continues indefinitely. I have observed it continue for hours.

@0xNineteen confirmed that push messages are taking significantly longer to process than other messages:

image

## Cause

I timed individual pieces of handleBatchPushMessages and found that just the log messages alone accounted for around 70% of the execution time. Next in line was GossipTable.insertValues, which uses about 80-90% of the remaining time.

I measured the time of individual insertions, and it would rapidly increase to around 200-300 μs within the first minute of starting sig. At the time of a bottleneck, I'm seeing insertions take upwards of 500 μs.

In GossipTable.insertValues, there are ArrayList allocations in that function to return the result of the insertion. Completely eliminating that return and all of its allocations reduced the execution time to 70 μs. This is not an acceptable solution but it identifies the main contributing factor to its duration.

## Solution

First, I removed the log message. This eliminated the fatal bottlenecks. Instead it would reach recoverable bottlenecks. Each batch's processing time would usually be on the order of several seconds instead of exponentially growing numbers of minutes. After a couple minutes it would recover and go back to processing multiple batches per second.

To optimize GossipTable.insertValues, I pre-allocate an ArrayList to re-use for the failures every time it is called, and pass it in as a pointer, rather than allocating a new list for each insertion. This cut down the insertion time down to around over 120 μs.

After both changes there are no longer significant bottlenecks. processMessages always completes its processing in well under 1s.

I also moved the prune messages from info to debug so there are no per-message info-level logging. I replaced them with a summary info message. It still clutters the logs a lot, but it's an improvement.
```
time=2024-03-24T19:08:21Z level=info  gossip: recv 63 messages: 0 ping, 0 pong, 62 push, 1 pull request, 0 pull response, 0 prune
```

## Caveats

- Most of my initial benchmarking was done in Debug mode, but I realized my mistake and re-ran some of it in ReleaseFast. I confirmed there is still a problem there, and my change still resolves the issue.
- After running for tens of minutes, I sometimes see a long-lasting slowdown in the processMessages loop. Instead of looping hundreds of times per second, it may get to a state where it's only running a few times per second. However it does not appear to ever reach a runaway bottleneck where it simply can't keep up. That issue is resolved. It always works its way through its messages at least as fast as they are coming in.
- This does not fully resolve the NoPeers error when generating pull requests. It makes an improvement, but that error still comes up occasionally. Somehow we're not getting enough ping responses from other nodes. Maybe we need to send more pings.

### DIFF
--- a/src/gossip/service.zig
+++ b/src/gossip/service.zig
@@ -694,6 +694,19 @@ pub const GossipService = struct {
             gossip_packets_processed += prune_messages.items.len;
             self.stats.gossip_packets_processed.add(gossip_packets_processed);
 
+            self.logger.infof(
+                "gossip: recv {} messages: {} ping, {} pong, {} push, {} pull request, {} pull response, {} prune",
+                .{
+                    messages.len,
+                    ping_messages.items.len,
+                    pong_messages.items.len,
+                    push_messages.items.len,
+                    pull_requests.items.len,
+                    pull_responses.items.len,
+                    prune_messages.items.len,
+                },
+            );
+
             // handle batch messages
             if (push_messages.items.len > 0) {
                 var x_timer = std.time.Timer.start() catch unreachable;
@@ -1585,11 +1598,19 @@ pub const GossipService = struct {
             pubkey_to_endpoint.deinit();
         }
 
+        // pre-allocate memory to track insertion failures
+        var max_inserts_per_push: usize = 0;
+        for (batch_push_messages.items) |*push_message| {
+            max_inserts_per_push = @max(max_inserts_per_push, push_message.gossip_values.len);
+        }
+        var failed_insert_indexs = try std.ArrayList(usize)
+            .initCapacity(self.allocator, max_inserts_per_push);
+        defer failed_insert_indexs.deinit();
+
         // insert values and track the failed origins per pubkey
         {
             var gossip_table_lock = self.gossip_table_rw.write();
             defer gossip_table_lock.unlock();
-
             for (batch_push_messages.items) |*push_message| {
                 var gossip_table: *GossipTable = gossip_table_lock.mut();
                 const valid_len = self.filterBasedOnShredVersion(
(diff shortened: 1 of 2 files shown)


src/gossip/service.zig
```diff
@@ -1585,33 +1598,38 @@ pub const GossipService = struct {
...
                 );
-                const failed_insert_indexs = result.failed.?;
-                defer failed_insert_indexs.deinit();
 
-                self.logger
-                    .field("n_values", valid_len)
-                    .field("from_addr", &push_message.from_pubkey.string())
-                    .field("n_failed_inserts", failed_insert_indexs.items.len)
-                    .info("gossip: recv push_message");
+                // logging this message takes too long and causes a bottleneck
```
0xNineteen, line 1627: instead of commenting this out - maybe we create a new logging level which is not the default? - likely because the other recv logs will also be slowing it down

might also be ok to leave it commented out for now but thinking more long-term - is the solution:
- fix the logging bottleneck to support this large freq of messages - (this is probs the long term soln)
- dont log too much (hard to say how much value these bring tbh) - (probs keep this in the back of our minds now that we know its a bottleneck - _70%!_)
- modify the log level to only print these if the user wants - (probs a good quick fix)

0xNineteen (approved):

src/gossip/service.zig
dnut: All three sound like good ideas.

> modify the log level to only print these if the user wants - (probs a good quick fix)

70% of the time spent logging is what I saw when I was already filtering out the messages by setting the log level to warn.

Calling `field` results in an allocation regardless of the log level. We may want to change the logging so you specify the level somehow before calling field. Also it could be useful to introduce a compile-time max log level.

dnut: I'm running the benchmark some more and seeing some counter-intuitive behavior.

With debug logs **disabled**, the timer still reports about 70% of the time is used during the single log statement.

With debug logs **enabled**, I would expect the logger to consume even more time, but it's actually reporting a bit less, more like 50-60% of the time the function is running.

src/gossip/table.zig
```diff
@@ -321,6 +321,41 @@ pub const GossipTable = struct {
...
+    ///
+    /// It accepts an arraylist of failures instead of returning an InsertResults, so it
+    /// can reuse the arraylist from a previous execution rather than allocating a new one.
+    ///
+    /// For simplicity and performance, only tracks failures without `inserted` and `timeouts`,
+    pub fn insertValuesMinAllocs(
+        self: *Self,
+        values: []SignedGossipData,
+        timeout: u64,
+        failed: *std.ArrayList(usize),
```
0xNineteen, line 337: can we change this to `failed_indexes`

ultd (approved):

PR #117, opened 2024-04-04
fix(gossip): update prometheus metric collection + dashboards

0xNineteen:
this changes metric collection to be a continuously increasing counter with the logging included in the struct which is more inline with what prometheus expects

this also improves the grafana dashboards

image


src/gossip/service.zig
```diff
@@ -2031,6 +1892,120 @@ pub const GossipService = struct {
...
+    push_messages_time_to_insert: *Counter,
+    push_messages_time_build_prune: *Counter,
+
+    table_n_values: *Counter,
+    table_n_pubkeys: *Counter,
+
+    // logging details
+    _logging_fields: struct {
+        logger: Logger,
+        log_interval_micros: i64 = 10_000_000,
```
0xNineteen, line 1939: `std.time.micros_per_sec * 10`

ultd (approved):

PR #123, opened 2024-04-08
feat(accountsdb): use Logger instead of std.debug.print

dnut:
Using the logger instead of calling `print` is a form of dependency inversion and has these benefits:

- standard log format, including timestamps
- easier to customize behavior, such as filtering log messages
- prevent unnecessary logging noise during tests
- integrate with log monitoring (long term)
- explicit dependencies make it more transparent to a calling context what a struct or function needs to do

Logging during tests was my primary motivation for working on this now. I find it distracting if successful tests result in "error" messages, which is what happens if the code calls `std.debug.print` during the tests. Too much noise from successful tests means we may not even recognize if there's a problem at all, and makes it harder to track down the cause when there is a problem.

### DIFF
--- a/src/accountsdb/db.zig
+++ b/src/accountsdb/db.zig
@@ -321,7 +321,7 @@ pub const AccountsDB = struct {
             file_map.putAssumeCapacityNoClobber(file_id_u32, accounts_file);
 
             if (file_count % 100 == 0 or (file_names.len - file_count) < 100) {
-                printTimeEstimate(&timer, file_names.len, file_count, "reading account files", null);
+                printTimeEstimate(self.logger, &timer, file_names.len, file_count, "reading account files", null);
             }
         }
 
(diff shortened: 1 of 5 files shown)


0xNineteen:
the main place we print instead of log is `printTimeEstimate` which uses `\r` when it prints so we only use one line per progress report - using the logging library, it doesnt work and you just get spammed with new lines

dnut:
Good point. So, this change is not what we want.

Reprinting the same line is nice when accountsdb is running on its own. But I think it will break whenever there are other threads writing to stderr or stdout.

Would it still be useful to have infrequent summary messages instead of sharing constant updates?

It could be customizable somehow so you have an updating line when that is desired, and just a few summary messages when that is desired.

0xNineteen:
yeah that might be a good approach - something like print the update stats every 10% or every 10seconds

0xNineteen:
one thing we should probably think about going forward -- with a lot of periodic print statements, its hard to know what stage the validator is in / what its doing 

for example, during snapshot download, gossip is running to get contact infos for peers to download from -- this means well have log statements for both snapshot download and gossip printing out - but from the logs it should be clear the 'stage' is 'downloading the snapshot'

how best to represent this 🤔

0xNineteen:
closing because was fixed in #131

PR #126, opened 2024-04-12
feat(gossip): new data types + abi hash script

0xNineteen:
## abi watching
adds a script which watches abi changes on agave's master branch and compares that to stable/what we support 

plan is to run this check with a cron job 

## new gossip types 
two new gossip types were added/supported too: 
- RestartHeaviestFork
- RestartLastVotedForkSlots

this required a change to support u8 bitsets (prev we only supported u64) - changed the code to use a different stdlib implementation`DynamicArrayBitSet` to support different block types with runtime sizes (stdlib only supports comptime size or a single blocktype -- not both)


ultd (approved):
🚀

PR #128, opened 2024-04-17
build script refactor, suggested improvements, and fixes

InKryption:
Hi, I'm beginning to familiarize myself with the codebase by diving in and applying improvements to organization & bug fixes. I started by hoisting the "Hash" type into a file struct, allowing more convenient import (`@import("Hash.zig")` instead of `@import("hash.zig").Hash`).

I also took the liberty of moving the inline namespace in `lib` to a `core/core.zig` "index" file, which puts the set of "exports" closer to the actual module, making it less likely for it to ever go out of sync with the existing modules.

It should be noted that the continued use of usingnamespace is potentially dubious, as there is more than a decent amount of energy in the zig project to possibly remove it; there is no official issue to do so at the time of writing this, however discussions alluding to such have become common in the compiler-devel channel in the Zig Programming Language Discord server, and I will cite the following issues/pull requests which are manifestations of this sentiment:
* https://github.com/ziglang/zig/pull/19565
* https://github.com/ziglang/zig/pull/18774
* https://github.com/ziglang/zig/issues/19352
* https://github.com/ziglang/zig/pull/19214

Potential rationale for its removal, as I understand it, is related to the complexity it adds to the implementation of incremental compilation.
Given the above reasoning, I would like to create additional follow-up PRs performing similar refactors.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -3,19 +3,8 @@ const std = @import("std");
 const package_name = "sig";
 const package_path = "src/lib.zig";
 
-// Although this function looks imperative, note that its job is to
-// declaratively construct a build graph that will be executed by an external
-// runner.
 pub fn build(b: *std.Build) void {
-    // Standard target options allows the person running `zig build` to choose
-    // what target to build for. Here we do not override the defaults, which
-    // means any target is allowed, and the default is native. Other options
-    // for restricting supported target set are available.
     const target = b.standardTargetOptions(.{});
-
-    // Standard optimization options allow the person running `zig build` to select
-    // between Debug, ReleaseSafe, ReleaseFast, and ReleaseSmall. Here we do not
-    // set a preferred release mode, allowing the user to decide how to optimize.
     const optimize = b.standardOptimizeOption(.{});
 
     const opts = .{ .target = target, .optimize = optimize };
@@ -29,15 +18,6 @@ pub fn build(b: *std.Build) void {
     const zstd_dep = b.dependency("zstd", opts);
     const zstd_mod = zstd_dep.module("zstd");
     const zstd_c_lib = zstd_dep.artifact("zstd");
-
-    const lib = b.addStaticLibrary(.{
-        .name = "sig",
-        // In this case the main source file is merely a path, however, in more
-        // complicated build scripts, this could be a generated file.
-        .root_source_file = .{ .path = "src/lib.zig" },
-        .target = target,
-        .optimize = optimize,
-    });
 
     // expose Sig as a module
     _ = b.addModule(package_name, .{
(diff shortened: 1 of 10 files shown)


build.zig
```diff
@@ -30,15 +19,6 @@ pub fn build(b: *std.Build) void {
     const zstd_mod = zstd_dep.module("zstd");
     const zstd_c_lib = zstd_dep.artifact("zstd");
 
-    const lib = b.addStaticLibrary(.{
```
0xNineteen, line 33: why did you remove this?

src/core/Hash.zig
```diff
@@ -0,0 +1,48 @@
+const std = @import("std");
+const Sha256 = std.crypto.hash.sha2.Sha256;
+
+const base58 = @import("base58-zig");
+
+const Hash = @This();
```
0xNineteen, line 6: we're not a big fan of turning files into structs - its a very zig-only thing which will just be confusing for new devs and just another 'thing' to understand

src/core/core.zig
```diff
@@ -0,0 +1,8 @@
+pub usingnamespace @import("pubkey.zig");
```
0xNineteen, line 1: the single lib with all namespaces is the simplest approach rn - moving these across files just adds another layer of indirection

while the prs you mentioned do bring up removing it - i think making changes once it is removed (and not before) will be the best approach

src/core/Hash.zig
```diff
@@ -0,0 +1,48 @@
...
+
+pub fn generateSha256Hash(bytes: []const u8) Hash {
+    var data: [SIZE]u8 = undefined;
+    Sha256.hash(bytes, &data, .{});
+    return .{ .data = data };
+}
+
+pub fn extendAndHash(id: Hash, val: []const u8) Hash {
+    var hasher = Sha256.init(.{});
+    hasher.update(&id.data);
```
0xNineteen, line 23: this is a nice change - would be good to keep the removing allocator changes

0xNineteen (commented):
general idea is some of the changes dont add enough value for the added complexity tradeoff *yet*

the hash function improvements are nice though and would be happy to merge those

build.zig
InKryption: It doesn't appear to serve any purpose. I don't see a single instance of any `export`s in the source code, and the actual outputted binary is functionally empty.

src/core/Hash.zig
InKryption: In my opinion that's an unfortunate tradeoff to make, but I understand.

src/core/hash.zig
```diff
@@ -1,69 +1,48 @@
...
 
-    pub fn cmp(a: *const Self, b: *const Self) CompareOperator {
-        for (0..HASH_SIZE) |i| {
-            if (a.data[i] > b.data[i]) {
-                return .gt;
-            } else if (a.data[i] < b.data[i]) {
-                return .lt;
-            }
+    pub fn order(a: *const Hash, b: *const Hash) std.math.Order {
+        for (a.data, b.data, 0..HASH_SIZE) |a_byte, b_byte, _| {
```
0xNineteen, line 28: can you remove this unused `0..HASH_SIZE`

InKryption: sure thing, come up right now

0xNineteen (approved):

0xNineteen:
should be merged soon - just waiting on final confirmation

thanks for the pr!

PR #131, opened 2024-04-23
fix(logs): move std.debug.print usage to use a Logger

0xNineteen:
prev we would print progress on large tasks using "\r" and std.debug.print - however this wasnt nice for running tests 

this pr changes progress to print with the logger (using \n) every few seconds

for more details see #123


src/accountsdb/db.zig
```diff
@@ -363,8 +372,16 @@ pub const AccountsDB = struct {
...
+            if (print_progress and progress_timer.read() > DB_PROGRESS_UPDATES_NS) {
+                printTimeEstimate(
+                    self.logger,
+                    &timer,
+                    file_names.len,
+                    file_count,
+                    "loading account files",
+                    "thread0",
+                );
+                progress_timer.reset();
```
dnut, line 384: This logic looks good, but it could be written more concisely using `progress_timer.lap()`, because then you won't need to call `progress_timer.reset()` at the end.

src/accountsdb/db.zig
```diff
@@ -390,9 +407,17 @@ pub const AccountsDB = struct {
...
+            if (print_progress and progress_timer.read() > DB_PROGRESS_UPDATES_NS) {
+                printTimeEstimate(
+                    self.logger,
+                    &timer,
+                    refs_ptr.items.len,
+                    ref_count,
+                    "building index",
+                    null,
+                );
+                progress_timer.reset();
```
dnut, line 420: could be `lap` (see above)

src/accountsdb/db.zig
```diff
@@ -40,6 +40,7 @@ const RefMemoryLinkedList = _accounts_index.RefMemoryLinkedList;
 const AccountRef = _accounts_index.AccountRef;
 const DiskMemoryAllocator = _accounts_index.DiskMemoryAllocator;
 
+pub const DB_PROGRESS_UPDATES_NS = 5 * std.time.ns_per_s;
```
dnut, line 43: I think you once mentioned a concern about clogging up the logs with too many periodic messages. We could expose log intervals for this kind of thing with a centrally managed configuration, for example CLI options. For now this looks good, just something to think about.

src/accountsdb/db.zig
```diff
@@ -724,7 +764,17 @@ pub const AccountsDB = struct {
                 local_total_lamports += lamports;
             }
 
-            printTimeEstimate(&timer, thread_bins.len, count, "gathering account hashes", null);
+            if (print_progress and progress_timer.read() > DB_PROGRESS_UPDATES_NS) {
```
dnut, line 767: could be `lap` (see above)

src/gossip/service.zig
```diff
@@ -2616,10 +2616,8 @@ test "gossip.gossip_service: test packet verification" {
     const id = Pubkey.fromPublicKey(&keypair.public_key);
     const contact_info = try localhostTestContactInfo(id);
 
-    var logger = Logger.init(std.testing.allocator, Logger.TEST_DEFAULT_LEVEL);
-    defer logger.deinit();
-    logger.spawn();
-
+    // noop for this case because this tests error failed verification
```
dnut, line 2619: I'm a little confused by this comment. Does this mean that error messages are expected because this is a negative test? If so, I agree with making the logger a noop.

src/utils/tar.zig
```diff
@@ -113,8 +120,18 @@ pub fn parallelUntarToFileSystem(
                     try dir.makePath(dir_name);
                 }
 
-                if (n_files_estimate) |n_files| {
-                    printTimeEstimate(&timer, n_files, file_count, "untar_files", null);
+                if (n_files_estimate) |total_n_files| {
+                    if (progress_timer.read() > TAR_PROGRESS_UPDATES_NS) {
```
dnut, line 124: could be `lap` (see comment in db.zig)

src/accountsdb/db.zig
```diff
@@ -488,7 +516,16 @@ pub const AccountsDB = struct {
...
+                printTimeEstimate(
+                    logger,
+                    &timer,
+                    total_bins,
+                    iteration_count,
+                    "combining thread indexes",
+                    "thread0",
+                );
+            }
         }
```
dnut, line 529: I don't see reset being called over here, is that intentional? Might be another place to use `lap` (as mentioned above)

dnut (commented):
Looks good overall, but I see one thing that looks like it might be a bug. Also in a few places I have a suggestion to use `lap`, which could help avoid that kind of bug.

src/gossip/service.zig
0xNineteen: yup exactly

src/accountsdb/db.zig
0xNineteen: yeah agreed - would be nice to move these to something more easily configurable

src/accountsdb/db.zig
0xNineteen: the problem with `lap()` is if the loop is faster than the threshold, the print code block will never be hit

src/accountsdb/db.zig
0xNineteen: 5a79870

src/accountsdb/db.zig
dnut: oh oops. it seems so obvious now, what was i thinking

dnut (approved):

PR #133, opened 2024-04-25
refactor(sync): simplify bitwise operations in bounded channel

dnut:
This aims to make the bitwise operations in the bounded channel implementation easier to read and understand. The channel logic is basically the same, but it uses the bits in the head, tail, and stamp differently. These usize values still encode the same information, but in a different format.

## Existing approach

The current logic dynamically divides the usize into three segments: lap, disconnect, and index. Head, tail, and stamp are each divided in the same way.

The least significant bits are used to represent an index within the buffer. Just enough bits are used to represent the largest possible index. For example, if the buffer has a length of 7, the three least significant bits are used to represent the index.

The next largest bit is a flag to indicate whether the channel is disconnected. This bit is set in the tail when the channel is disconnected.

All larger bits represent the number of laps that have been circled around the ring buffer. If the buffer is 7 items in size, and we've sent/received a total of 15 items, that represents two complete laps plus one more item. Since the 4 least significant bits are used for the disconnect and the index, the fifth and so on bits will represent the number 2 for 2 laps. In other words, this is the number 2 left shifted by 4 bits, or 0b0010_0000. The struct has a field `one_lap` that represents the fifth bit.

The usize is also used to indicate whether a slot is occupied. There is no individual bit allocated for this purpose. Instead, a slot's stamp has a +1 added to it. We can identify that the slot is occupied when we see that it is one larger than we would expect it to be for that particular slot.

## New approach

With the new approach, the regions are no longer dynamic. I use the two most significant bits in the usize as flags. Like before: head, tail, and stamp are each divided in the same way as each other.

The largest bit is the disconnect bit. If this bit is set in the tail, the channel is disconnected

The second largest bit is the occupied bit. If this is set in a slot's stamp, that means the slot is occupied.

All remaining bits represent the cumulative index. This is a monotonically increasing number. If 100 items have been sent, the tail will equal 100. The actual index in the buffer is `cumulative_index % capacity`. There is no need to calculate the number of laps, but you could do this with: `cumulative_index / capacity`.

The field `one_lap` is now just the capacity of the buffer. `self.buffer.len` could actually be used instead.

### easier to make sense of at first glance

In the current code, `+1` or `... + 1 == ... + one_lap` is used in various places throughout the code, but it's not entirely clear why. The actual reason for this code is often to check or set a slot as occupied, but you won't know this until you read and understand the entire channel implementation. In the new approach, `| occupied_bit` is used for this purpose. `occupied_bit` is documented, so the meaning is clear, even if you haven't read code from other functions yet.

### simpler logic

When moving the head or tail forward, the existing code checks whether the index will overflow, so it can restart at zero and increment the lap. With the new code, this is unnecessary. We can just increment the cumulative index with a single operation. The modulo operator handles wraparound when getting the index.

The existing code needs to explicitly separate the index and lap, and then work with them in separate steps. The new code still extracts the index, but doesn't need to extract the lap. The lap is no longer needed for anything.


src/sync/bounded.zig
```diff
@@ -100,36 +102,34 @@ pub fn Bounded(comptime T: type) type {
...
             while (true) {
                 // Check if the channel is disconnected.
-                if (tail & self.disconnect_bit != 0) {
+                if (tail & disconnect_bit != 0) {
                     return error.disconnected;
                 }
 
                 // Deconstruct the tail.
-                const index = tail & (self.disconnect_bit - 1);
-                const lap = tail & ~(self.one_lap_bit - 1);
```
dnut, line 109: Previously the lap was only used to determine how to increment the tail. This is not needed in the new implementation because the tail can always be incremented with +1.

dnut (commented):
Added some comments to explain the things that were changed: how they worked before, and how they work now.

0xNineteen:
this pr has been stale for a while now (merge conflicts will likely be expensive to solve) - gonna close it and move it to this ticket which contains more context and a ref to this pr:

https://github.com/orgs/Syndica/projects/2/views/1?pane=issue&itemId=71252009

PR #143, opened 2024-05-23
fix(gossip,accountsdb): removed leaks in error path, constified things, pass slices instead of ArrayList

InKryption:
* Switch from `std.ArrayList` to `std.ArrayListUnmanaged` to save memory in some cases.

* Improve `spawnThreadTasks` to output into a caller-provided arraylist, allowing them the ability to properly clean up even in the error path. Also document more precisely.

* Use tuple destructuring feature now available in zig 0.12.

* Constify certain things not caught by zig's ast checks.

* Fix leaks.

Supercedes #142.

### DIFF
--- a/src/cmd/cmd.zig
+++ b/src/cmd/cmd.zig
@@ -370,15 +370,13 @@ fn gossip() !void {
         logger,
         my_keypair,
         &exit,
-        entrypoints,
+        entrypoints.items,
         my_data.shred_version,
         my_data.ip,
         &.{},
     );
     defer gossip_service.deinit();
-
-    var handle = try spawnGossip(&gossip_service);
-    handle.join();
+    try runGossip(&gossip_service);
 }
 
 /// entrypoint to run a full solana validator
@@ -402,7 +400,7 @@ fn validator() !void {
         logger,
         my_keypair,
         &exit,
-        entrypoints,
+        entrypoints.items,
         ip_echo_data.shred_version, // TODO atomic owned at top level? or owned by gossip is good?
         ip_echo_data.ip,
         &.{.{ .tag = socket_tag.REPAIR, .port = repair_port }},
@@ -515,7 +513,7 @@ fn initGossip(
     logger: Logger,
     my_keypair: KeyPair,
     exit: *Atomic(bool),
-    entrypoints: std.ArrayList(SocketAddr),
+    entrypoints: []const SocketAddr,
     shred_version: u16,
     gossip_host_ip: IpAddr,
     sockets: []const struct { tag: u8, port: u16 },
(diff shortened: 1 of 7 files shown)


src/cmd/cmd.zig
```diff
@@ -572,14 +570,14 @@ fn initRepair(
     };
 }
 
+fn runGossip(gossip_service: *GossipService) !void {
```
0xNineteen, line 573: few things:
- can we rename this to `runGossipWithConfigValues` or something more descriptive? 
- with this change we can remove all `spawnGossip` calls to just spawn the thread directly (vs single-line fn call)
- lastly, can we change `current` variable name to `gossip_config`

src/gossip/fuzz.zig
```diff
@@ -306,7 +303,7 @@ pub fn run() !void {
         allocator,
         fuzz_contact_info,
         fuzz_keypair,
-        entrypoints,
+        (&entrypoint)[0..1],
```
0xNineteen, line 306: never seen that syntax before - thats cool

src/utils/thread.zig
```diff
@@ -4,39 +4,41 @@ const ThreadPool = @import("../sync/thread_pool.zig").ThreadPool;
...
-    var end_index: usize = 0;
     for (0..n_threads) |thread_id| {
-        if (thread_id == (n_threads - 1)) {
-            end_index = data_len;
-        } else {
-            end_index = start_index + chunk_size;
-        }
-        const handle = try std.Thread.spawn(.{}, f, params ++ .{ start_index, end_index, thread_id });
+        const end_index = if (thread_id == n_threads - 1) data_len else (start_index + chunk_size);
+        // NOTE(trevor): instead of just `try`ing, we could fill an optional diagnostic struct
```
0xNineteen, line 34: yeah thats an interesting idea - rn we only care about full coverage but could be useful in the future

0xNineteen (changes requested):
small naming changes - otherwise lgtm

src/cmd/cmd.zig
InKryption: can do

src/gossip/fuzz.zig
InKryption: Yeah, it's semi-recent I believe. I believe it was added in 0.12.

src/utils/thread.zig
InKryption: Yeah, I didn't see anywhere currently where it'd be useful, but it's not impossible - being able to continue execution with accurate information in spite of a system hiccup or lack of available memory seems valuable enough that I'd want to keep the thought salient through this TODO.

0xNineteen (approved):
lgtm - i think we should wait for drew to approve also before merging

dnut (approved):

PR #144, opened 2024-05-23
refactor(build.zig): updates for latest zig & improvements

InKryption:
* Removes superfluous manual linking of curl & zstd libraries, which are already linked by their respective build functions.

* Add `filters` instead of supplying one test filter via `b.args[0]`.

* Use `b.path` instead of the now deprecated `path` field in `LazyPath`.

* Unroll the loop declaring the `fuzz` and `benchmark` steps. This also resolves the TODO regarding not needing all dependencies for all of the executables.

* Group top level step declarations together at the beginning of the build function.

* Don't add `zig-cli` import to the sig executable twice.

* Install all compile step artifacts for convenient debugging.

* Use more consistent naming for dependency declarations.

* Use `addImport` instead of declaring the dependencies for the sig module up front; it's a lot more concise, and looks neater.

* Inline the `package_name` and `package_path` constants.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -1,166 +1,120 @@
 const std = @import("std");
+const Build = std.Build;
 
-const package_name = "sig";
-const package_path = "src/lib.zig";
-
-pub fn build(b: *std.Build) void {
+pub fn build(b: *Build) void {
+    // CLI options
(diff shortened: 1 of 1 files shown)


0xNineteen (approved):
lgtm

Issue #120, opened 2024-04-08
Utils: Threadpool Error Handling


0xNineteen:
right now thread pool tasks' return type is `void` which means we dont return any errors (we just `catch` rn)

would be good to figure out a good way to communicate/handle these errors better

Issue #119, opened 2024-04-08
Gossip: Improve thread management


0xNineteen:
rn we just spawn threads using hard-coded fixed values - we need to pull these values into cli args + set reasonable default values depending on the number of threads available along with good log values to tell the user where the threads are being used

PR #145, opened 2024-05-24
improvement(gossip,tar): Work towards improving thread management

InKryption:
Starts work towards #119 & #120.
More details in the commits.

### DIFF
--- a/src/gossip/service.zig
+++ b/src/gossip/service.zig
@@ -168,7 +168,7 @@ pub const GossipService = struct {
         const gossip_address = my_contact_info.getSocket(socket_tag.GOSSIP) orelse return error.GossipAddrUnspecified;
         var gossip_socket = UdpSocket.create(.ipv4, .udp) catch return error.SocketCreateFailed;
         gossip_socket.bindToPort(gossip_address.port()) catch return error.SocketBindFailed;
-        gossip_socket.setReadTimeout(socket_utils.SOCKET_TIMEOUT) catch return error.SocketSetTimeoutFailed; // 1 second
+        gossip_socket.setReadTimeout(socket_utils.SOCKET_TIMEOUT_US) catch return error.SocketSetTimeoutFailed; // 1 second
 
         const failed_pull_hashes = HashTimeQueue.init(allocator);
         const push_msg_q = ArrayList(SignedGossipData).init(allocator);
--- a/src/gossip/service.zig
+++ b/src/gossip/service.zig
@@ -1,8 +1,10 @@
 const std = @import("std");
+const builtin = @import("builtin");
 const network = @import("zig-network");
 const EndPoint = network.EndPoint;
 const Packet = @import("../net/packet.zig").Packet;
 const PACKET_DATA_SIZE = @import("../net/packet.zig").PACKET_DATA_SIZE;
+const ThreadPoolTask = @import("../utils/thread.zig").ThreadPoolTask;
 const ThreadPool = @import("../sync/thread_pool.zig").ThreadPool;
 const Task = ThreadPool.Task;
 const Batch = ThreadPool.Batch;
@@ -87,6 +89,8 @@ const DEFAULT_EPOCH_DURATION: u64 = 172800000;
 
 pub const PUB_GOSSIP_STATS_INTERVAL_MS = 2 * std.time.ms_per_s;
 pub const GOSSIP_TRIM_INTERVAL_MS = 10 * std.time.ms_per_s;
+
+pub const GOSSIP_VERIFY_PACKET_PARALLEL_TASKS = 4;
 
 pub const GossipService = struct {
     allocator: std.mem.Allocator,
@@ -260,7 +264,7 @@ pub const GossipService = struct {
 
     /// these threads should run forever - so if they join - somethings wrong
     /// and we should shutdown
-    fn joinAndExit(self: *Self, handle: *std.Thread) void {
+    fn joinAndExit(self: *const Self, handle: std.Thread) void {
         handle.join();
         self.exit.store(true, .unordered);
     }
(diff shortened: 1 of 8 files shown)


src/net/socket_utils.zig
```diff
@@ -17,53 +17,42 @@ pub fn readSocket(
...
         // recv packets into batch
         while (true) {
-            const n_packets_read = recvMmsg(socket, packet_batch.items[count..capacity], exit) catch |err| {
-                if (count > 0 and err == error.WouldBlock) {
-                    if (timer.read() > MAX_WAIT_NS) {
-                        break;
-                    }
-                }
+            const old_count = packet_batch.items.len;
+            packet_batch.items.len += recvMmsg(socket, packet_batch.unusedCapacitySlice(), exit) catch |err| {
```
0xNineteen, line 38: re-looking at this logic - i think we can remove a lot of unnecessary complexity - just set the timeout on the socket to something reasonable (given that we only want to break to check for exit signals) and simplify recvMmsg to fill up the batch until its full or it would block and loop again - wdyt - can huddle on it if you want

src/gossip/service.zig
```diff
@@ -280,61 +284,64 @@ pub const GossipService = struct {
...
 
-        var maybe_build_messages_handle = if (!spy_node) try Thread.spawn(.{}, buildMessages, .{self}) else null;
-        defer {
-            if (maybe_build_messages_handle) |*handle| {
-                self.joinAndExit(handle);
-            }
-        }
+        const task_allocator = switch (builtin.os.tag) {
+            .windows => heap_task_allocator.allocator(),
+            else => gp_task_allocator.allocator(),
```
0xNineteen, line 305: in processesMessages loop we use `self.allocator` to shallow free - so using a different allocator here would break that rn (which you can see if u try to run gossip) - fyi should run most of the commands before opening pr (im not great at this too - ci/cd which runs gossip for a few seconds and then shuts down is on the todo list)

src/gossip/service.zig
```diff
@@ -417,26 +410,23 @@ pub const GossipService = struct {
...
-                task_ptr.packet_batch = packet_batch;
+                const acquired_task_idx = VerifyMessageTask.awaitAndAcquireFirstAvailableTask(tasks, task_search_start_idx);
+                task_search_start_idx = (acquired_task_idx + 1) % tasks.len;
+
+                const task_ptr = &tasks[acquired_task_idx];
+                task_ptr.entry.packet_batch = packet_batch;
+                task_ptr.result catch |err| self.logger.errf("VerifyMessageTask encountered error: {s}", .{@errorName(err)});
 
                 const batch = Batch.from(&task_ptr.task);
                 self.thread_pool.schedule(batch);
```
0xNineteen, line 423: i feel like these two are always used together - can we throw them in a helper fcn of `ThreadPoolTask`

src/utils/tar.zig
```diff
@@ -87,8 +88,10 @@ pub fn parallelUntarToFileSystem(
     var file_count: usize = 0;
     const strip_components: u32 = 0;
     loop: while (true) {
-        var header_buf = try allocator.alloc(u8, 512);
-        _ = try reader.readAtLeast(header_buf, 512);
+        const header_buf = try allocator.alloc(u8, 512);
+        if (header_buf.len != try reader.readAtLeast(header_buf, 512)) {
```
0xNineteen, line 92: can we change `header_buf.len` to hardcoded `512` in this `==` check -- just easier to read

src/utils/tar.zig
```diff
@@ -136,19 +139,25 @@ pub fn parallelUntarToFileSystem(
                 file_count += 1;
 
                 const contents = try allocator.alloc(u8, file_size);
-                _ = try reader.readAtLeast(contents, file_size);
+                const actual_contents_len = try reader.readAtLeast(contents, file_size);
+                if (contents.len != actual_contents_len) {
```
0xNineteen, line 143: same thing here -- just easier to understand if file_size is explicit in the comparison

0xNineteen (changes requested):

src/gossip/service.zig
```diff
@@ -417,26 +410,23 @@ pub const GossipService = struct {
...
-                // find a free task
-                var task_ptr = &tasks[task_i];
-                while (!task_ptr.done.load(.acquire)) {
-                    task_i = (task_i + 1) % n_tasks;
-                    task_ptr = &tasks[task_i];
-                }
-                // schedule it
-                task_ptr.done.store(false, .release);
-                task_ptr.packet_batch = packet_batch;
+                const acquired_task_idx = VerifyMessageTask.awaitAndAcquireFirstAvailableTask(tasks, task_search_start_idx);
```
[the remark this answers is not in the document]
InKryption: I've tried this a few different ways, but I personally couldn't find a way that I liked it. If it's alright something to this effect could be left as a future improvement.

src/net/socket_utils.zig
InKryption: To clarify on this, is this to mean that `recvMmsg` should become blocking?

src/gossip/service.zig
InKryption: I would prefer to keep the `thread_pool.schedule` invocation top-level, not too hidden down in helper functions. I think it would be reasonable to have a method on ThreadPoolTask to make a batch out of it though, making it `self.thread_pool.schedule(task_ptr.toBatch())`.

src/net/socket_utils.zig
0xNineteen: yeah 'blocking' by using a reasonable timeout (maybe 1second) - not full blocking

InKryption: I'd like to huddle on this, because I'm not exactly sure what the correct approach is.

src/gossip/service.zig
```diff
@@ -125,16 +129,14 @@ pub const GossipService = struct {
 
     const Entrypoint = struct { addr: SocketAddr, info: ?ContactInfo = null };
 
-    const Self = @This();
```
0xNineteen, line 128: we actually want to keep everything as `Self` throughout the codebase for consistency & style sake

src/gossip/service.zig
```diff
@@ -258,12 +260,40 @@ pub const GossipService = struct {
...
+        packet_verifier_thread: std.Thread,
+        message_processor_thread: std.Thread,
+        message_builder_thread: ?std.Thread,
+        responder_thread: std.Thread,
+        dumper_thread: ?std.Thread,
+
+        /// If any of the threads join, all other threads will be signalled to join.
+        pub fn joinAndExit(handles: RunHandles) void {
+            inline for (@typeInfo(RunHandles).Struct.fields, 0..) |field, i| cont: {
+                comptime if (@field(std.meta.FieldEnum(RunHandles), field.name) == .exit) {
```
0xNineteen, line 275: is this `.exit` equality comparison equal to an "exit" string comparison ?

src/gossip/service.zig
```diff
@@ -258,12 +260,40 @@ pub const GossipService = struct {
...
+        exit: *AtomicBool,
+        receiver_thread: std.Thread,
+        packet_verifier_thread: std.Thread,
+        message_processor_thread: std.Thread,
+        message_builder_thread: ?std.Thread,
+        responder_thread: std.Thread,
+        dumper_thread: ?std.Thread,
+
+        /// If any of the threads join, all other threads will be signalled to join.
+        pub fn joinAndExit(handles: RunHandles) void {
```
0xNineteen, line 273: can you change this to follow the `Self` convention

0xNineteen (changes requested):
looking like a really solid pr nice work - just a few small things

src/gossip/service.zig
InKryption: It isn't a string comparison, it's an enum comparison, which is the same as an integer comparison. To be clear though, this isn't a runtime optimisation, as it's all happening at comptime. If we use `std.mem.eql`, the eval branch quota consumption is `fields.len * (sum_of_all_field_name_lengths)`; `std.meta.FieldEnum` consumes the eval branch quota only once to generate the enum (and it only consumes `fields.len` of the quota) thanks to memoization, and then the comparison itself consumes no eval branch quota, because it's a simple enum comparison (same as comparing an integer).

src/gossip/service.zig
```diff
@@ -258,12 +260,40 @@ pub const GossipService = struct {
...
+                };
+                const maybe_thread: ?std.Thread = @field(handles, field.name);
+                const thread = maybe_thread orelse break :cont;
+                thread.join(); // if we end up joining, something's gone wrong, so signal exit
+                if (i == 0) handles.exit.store(true, .unordered);
+            }
+        }
+    };
+
+    pub const RunThreadsParams = struct {
```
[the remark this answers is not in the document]
InKryption: At the relevant callsites, I kept getting confused on as to the order of `spy_node` and `dump` as parameters, so I decided to assign them names. Another solution would be to use an enum instead of a boolean, I just opted for the path of least resistance.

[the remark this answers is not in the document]
0xNineteen: gotcha - yeah that makes sense - thinking if we should adopt this going forward 🤔  - i like the explicit names for reading too

InKryption:
I conducted many more runs than presented here, but those presented are representative of the general trends I observed, where performance is either equivalent or minimally better, excluding outliers where variance exceeded more than 100 in either (which occurred twice with the old code, and once with the new code).

Side note: brief testing indicated that ReleaseFast demonstrates near identical trends.

Before socket_utils changes:
```
❯ zig build -Doptimize=ReleaseSafe benchmark -- socket_utils  
filtering benchmarks with prefix: socket_utils

Benchmark                      Iterations    Min(ms)    Max(ms)   Variance   Mean(ms)
-------------------------------------------------------------------------------------
benchmarkReadSocket(100k_msgs)          3       1739       1755         43       1747



❯ zig build -Doptimize=ReleaseSafe benchmark -- socket_utils
filtering benchmarks with prefix: socket_utils

Benchmark                      Iterations    Min(ms)    Max(ms)   Variance   Mean(ms)
-------------------------------------------------------------------------------------
benchmarkReadSocket(100k_msgs)          3       1737       1755         56       1745



❯ zig build -Doptimize=ReleaseSafe benchmark -- socket_utils
filtering benchmarks with prefix: socket_utils

Benchmark                      Iterations    Min(ms)    Max(ms)   Variance   Mean(ms)
-------------------------------------------------------------------------------------
benchmarkReadSocket(100k_msgs)          3       1743       1751         12       1747



❯ zig build -Doptimize=ReleaseSafe benchmark -- socket_utils
filtering benchmarks with prefix: socket_utils

Benchmark                      Iterations    Min(ms)    Max(ms)   Variance   Mean(ms)
-------------------------------------------------------------------------------------
benchmarkReadSocket(100k_msgs)          3       1739       1742          2       1740



❯ zig build -Doptimize=ReleaseSafe benchmark -- socket_utils
filtering benchmarks with prefix: socket_utils

Benchmark                      Iterations    Min(ms)    Max(ms)   Variance   Mean(ms)
-------------------------------------------------------------------------------------
benchmarkReadSocket(100k_msgs)          3       1740       1747          8       1743
```

After socket_utils changes:
```
❯ zig build -Doptimize=ReleaseSafe benchmark -- socket_utils
filtering benchmarks with prefix: socket_utils

Benchmark                      Iterations    Min(ms)    Max(ms)   Variance   Mean(ms)
-------------------------------------------------------------------------------------
benchmarkReadSocket(100k_msgs)          3       1719       1735         44       1727



❯ zig build -Doptimize=ReleaseSafe benchmark -- socket_utils
filtering benchmarks with prefix: socket_utils

Benchmark                      Iterations    Min(ms)    Max(ms)   Variance   Mean(ms)
-------------------------------------------------------------------------------------
benchmarkReadSocket(100k_msgs)          3       1729       1734          4       1732



❯ zig build -Doptimize=ReleaseSafe benchmark -- socket_utils
filtering benchmarks with prefix: socket_utils

Benchmark                      Iterations    Min(ms)    Max(ms)   Variance   Mean(ms)
-------------------------------------------------------------------------------------
benchmarkReadSocket(100k_msgs)          3       1730       1733          1       1731



❯ zig build -Doptimize=ReleaseSafe benchmark -- socket_utils
filtering benchmarks with prefix: socket_utils

Benchmark                      Iterations    Min(ms)    Max(ms)   Variance   Mean(ms)
-------------------------------------------------------------------------------------
benchmarkReadSocket(100k_msgs)          3       1728       1738         17       1732



❯ zig build -Doptimize=ReleaseSafe benchmark -- socket_utils
filtering benchmarks with prefix: socket_utils

Benchmark                      Iterations    Min(ms)    Max(ms)   Variance   Mean(ms)
-------------------------------------------------------------------------------------
benchmarkReadSocket(100k_msgs)          3       1721       1738         51       1728
```

src/gossip/service.zig
InKryption: Small problem: using `const Self = @This()` both here and in the enclosing `GossipService` struct namespace means using the self convention here is an ambiguous reference. This is one of the big disadvantages of strictly following the `Self` convention, instead of directly using the type name. Either we make an exception here, or I move the `RunHandles` struct outside of `GossipService`. Which do you prefer?

Also note: this same issue exists for `VerifyMessageEntry`, which was also breaking this rule before the PR.

src/gossip/service.zig
InKryption: See my reply to the comment on `RunHandles`.

src/gossip/service.zig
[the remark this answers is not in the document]
InKryption: Yeah, struct param for named parameters, and descriptive enums are both conventions I would like to see used in the codebase, unnamed booleans and integers can be pretty hard to follow otherwise.

src/gossip/service.zig
0xNineteen: ah i see - ok keeping `RunHandles`/`VerifyMessageEntry` as-is is ok then

InKryption: Understood, sending changes in a minute.

0xNineteen (approved):
lgtm

PR #149, opened 2024-05-30
fix(snapshot-download): deinit after download & delete failed downloads

0xNineteen:


InKryption (approved):
LGTM. I might've made the deinit method take a constant pointer to self since it doesn't actually mutate, but the type is always going to be mutable, so it's a non-issue.

Issue #152, opened 2024-06-03
invalid free inside bincode from gossip when using mainnet entrypoints


InKryption:
### Description

As described by the title.

### How to Reproduce the Bug

```
❯ zig build run -- -l err gossip -e 86.109.15.59:8001
thread 455251 panic: Invalid free
/Users/trevor/ziglang/zig/zig-out/lib/zig/std/heap/general_purpose_allocator.zig:643:21: 0x100cee563 in freeLarge (sig)
                    @panic("Invalid free");
                    ^
/Users/trevor/ziglang/zig/zig-out/lib/zig/std/heap/general_purpose_allocator.zig:854:31: 0x100c39f0b in free (sig)
                self.freeLarge(old_mem, log2_old_align, ret_addr);
                              ^
/Users/trevor/ziglang/zig/zig-out/lib/zig/std/mem/Allocator.zig:98:28: 0x100c3eb8f in free__anon_4904 (sig)
    return self.vtable.free(self.ptr, buf, log2_buf_align, ret_addr);
                           ^
/Users/trevor/projects/zig/sig/src/bincode/bincode.zig:391:35: 0x100e434c3 in free__anon_29827 (sig)
                    allocator.free(value);
                                  ^
/Users/trevor/projects/zig/sig/src/bincode/bincode.zig:369:33: 0x100f5812b in free__anon_33419 (sig)
                    bincode.free(allocator, @field(value, field.name));
                                ^
/Users/trevor/projects/zig/sig/src/bincode/bincode.zig:381:40: 0x100ef0ceb in free__anon_32470 (sig)
                    return bincode.free(allocator, @field(value, field.name));
                                       ^
/Users/trevor/projects/zig/sig/src/bincode/bincode.zig:389:37: 0x100e433df in free__anon_29803 (sig)
                        bincode.free(allocator, item);
                                    ^
/Users/trevor/projects/zig/sig/src/bincode/bincode.zig:369:33: 0x100de998f in free__anon_29018 (sig)
                    bincode.free(allocator, @field(value, field.name));
                                ^
/Users/trevor/projects/zig/sig/src/bincode/bincode.zig:369:33: 0x100d8bbdf in free__anon_27829 (sig)
                    bincode.free(allocator, @field(value, field.name));
                                ^
/Users/trevor/projects/zig/sig/src/bincode/bincode.zig:381:40: 0x100d29d7f in free__anon_26513 (sig)
                    return bincode.free(allocator, @field(value, field.name));
                                       ^
[...]

PR #153, opened 2024-06-03
fix(gossip): Fix invalid free

InKryption:
This involves passing the "message allocator" as a more general "table allocator" at the initialization site of gossip service, to allow the gossip table to properly free the data with the right allocator.

Merging this should be high priority, because currently there is a panic in a currently very reachable set of codepaths, which only manifested when testing on mainnet.

### DIFF
--- a/src/cmd/cmd.zig
+++ b/src/cmd/cmd.zig
@@ -47,8 +47,14 @@ const config = @import("config.zig");
 const ACCOUNT_INDEX_BINS = @import("../accountsdb/db.zig").ACCOUNT_INDEX_BINS;
 const socket_tag = @import("../gossip/data.zig").socket_tag;
 
+// TODO: use better allocator, unless GPA becomes more performant.
+
 var gpa = std.heap.GeneralPurposeAllocator(.{}){};
 const gpa_allocator = gpa.allocator();
+
+var gossip_value_gpa: std.heap.GeneralPurposeAllocator(.{}) = .{};
+const gossip_value_gpa_allocator = gossip_value_gpa.allocator();
+
 const base58Encoder = base58.Encoder.init(.{});
 
 const gossip_host = struct {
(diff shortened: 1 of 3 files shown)


InKryption:
Closes #152.

src/gossip/service.zig
```diff
@@ -132,13 +133,27 @@ pub const GossipService = struct {
     const Entrypoint = struct { addr: SocketAddr, info: ?ContactInfo = null };
 
     pub fn init(
-        allocator: std.mem.Allocator,
-        my_contact_info: ContactInfo,
-        my_keypair: KeyPair,
-        entrypoints: ?[]const SocketAddr,
-        exit: *AtomicBool,
-        logger: Logger,
+        params: struct {
```
0xNineteen, line 136: can we keep the params as params for now - not sure how we feel about making them structs for rn - likely something we should discuss more

src/cmd/cmd.zig
```diff
@@ -47,8 +47,14 @@ const config = @import("config.zig");
 const ACCOUNT_INDEX_BINS = @import("../accountsdb/db.zig").ACCOUNT_INDEX_BINS;
 const socket_tag = @import("../gossip/data.zig").socket_tag;
 
+// TODO: use better allocator, unless GPA becomes more performant.
+
 var gpa = std.heap.GeneralPurposeAllocator(.{}){};
 const gpa_allocator = gpa.allocator();
+
+var table_gpa: std.heap.GeneralPurposeAllocator(.{}) = .{};
+const table_gpa_allocator = table_gpa.allocator();
```
0xNineteen, line 56: can we change `table` to `value` so its `value_allocator` since it allocates gossip values

src/gossip/service.zig
InKryption: Sure, that's fine. I did it here because having two unnamed allocator parameters would be confusing, but that's less important to me than getting this merged.

0xNineteen (approved):

PR #159, opened 2024-06-05
improvement(accountsdb): Make `FileId` a typed alias, & enhance `AccountFileInfo.validate`.

InKryption:
This provides stronger type safety guarantees than the typedef, and makes the association between code that works with `FileId`s clearer.

Also enhances `AccountFileInfo.validate` by annotating it with an explicit error set, and adding a check to ensure the file id is within the range described by the comment.

### DIFF
--- a/src/accountsdb/accounts_file.zig
+++ b/src/accountsdb/accounts_file.zig
@@ -10,7 +10,27 @@ const Pubkey = @import("../core/pubkey.zig").Pubkey;
 
 const AccountFileInfo = @import("snapshots.zig").AccountFileInfo;
 
-pub const FileId = u32;
+/// Simple strictly-typed alias for an integer, used to represent a file ID.
+pub const FileId = enum(u32) {
+    _,
+
+    pub inline fn fromInt(int: u32) FileId {
+        return @enumFromInt(int);
+    }
+
+    pub inline fn toInt(file_id: FileId) u32 {
+        return @intFromEnum(file_id);
+    }
+
+    pub fn format(
+        _: FileId,
+        comptime _: []const u8,
+        _: std.fmt.FormatOptions,
+        _: anytype,
+    ) !void {
+        @compileError("Should not print " ++ @typeName(FileId) ++ " directly");
+    }
+};
 
 // an account thats stored in an AccountFile
 pub const AccountInFile = struct {
--- a/src/accountsdb/db.zig
+++ b/src/accountsdb/db.zig
@@ -373,8 +373,8 @@ pub const AccountsDB = struct {
                 std.debug.panic("failed to *sanitize* AccountsFile: {d}.{d}: {s}\n", .{ accounts_file.slot, accounts_file.id, @errorName(err) });
             };
 
-            const file_id_u32: u32 = @intCast(accounts_file_id);
-            file_map.putAssumeCapacityNoClobber(file_id_u32, accounts_file);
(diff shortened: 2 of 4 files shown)


src/accountsdb/accounts_file.zig
```diff
@@ -10,7 +10,27 @@ const Pubkey = @import("../core/pubkey.zig").Pubkey;
 
 const AccountFileInfo = @import("snapshots.zig").AccountFileInfo;
 
-pub const FileId = u32;
+/// Simple strictly-typed alias for an integer, used to represent a file ID.
+pub const FileId = enum(u32) {
+    _,
+
+    pub inline fn fromInt(int: u32) FileId {
```
dnut, line 17: I am often tempted to inline functions when it seems like a good fit (as in this case). I usually choose not to because I've heard it suggested many times by people more knowledgeable than me about compilers that the compiler is better at making these decisions. I'm curious about your thoughts.

Example: https://ziggit.dev/t/inlining-functions/1341/3

src/accountsdb/accounts_file.zig
```diff
@@ -10,7 +10,27 @@ const Pubkey = @import("../core/pubkey.zig").Pubkey;
...
+        return @intFromEnum(file_id);
+    }
+
+    pub fn format(
+        _: FileId,
+        comptime _: []const u8,
+        _: std.fmt.FormatOptions,
+        _: anytype,
+    ) !void {
+        @compileError("Should not print " ++ @typeName(FileId) ++ " directly");
```
dnut, line 31: why not?

src/accountsdb/accounts_file.zig
InKryption: My reasoning here is that these functions ought to compile down to their exact function bodies; there are no smart optimizations the compiler can make here beyond simply doing exactly what I've already done, which is to inline the function into the builtin call. This work could be left to the optimizer, but then that requires that the optimizer commit to more work than is strictly necessary. I already know there is nothing else it can do, so by marking it inline I am giving it one less thing to do.

This isn't performance critical, and it won't come up in any measurements, but it expresses my intent for these functions to be nothing more than direct wrappers over these builtins. If they became more complicated, or I suspected the compiler could intuit more than me, I would then also heed Andrew's advice.

As well, this will not have any effect on any stack traces, because these functions cannot panic or return an error, and thus their inline-ness will have no impact on the stack frames.

Lmk what you think. I am open to un-inlining these if we want to adhere to a strict policy of only `inline`ing when strictly necessary or measurably beneficial.

src/accountsdb/accounts_file.zig
InKryption: I forget which, but I saw a PR in which `FieldId` were being printed as part of a filename somewhere, so I set this up to catch any unintended behavior.

0xNineteen (approved):
lgtm - should wait till drew also approves before merging

src/accountsdb/accounts_file.zig
dnut: I'm not convinced that we need to strictly enforce a policy like that. And I don't see any harm of inlining this function. Mainly I just wanted to hear your thoughts. I'm fine with keeping it as is.

src/accountsdb/accounts_file.zig
dnut: makes sense

dnut (approved):

PR #161, opened 2024-06-05
feat(ci/cd): add running gossip + benchmarks

0xNineteen:
- running gossip for N seconds and then shutting down [x]
- run benchmarks [x]
- unused imports check [x]

the CI will now run only on review request AND every review submitted (instead on every commit of a pull request) - due to the increased time it will run i didnt want to run it for long

sample run: https://github.com/Syndica/sig/actions/runs/9390446592/job/25860335747

---

_note:_ decided to not include accounts-db because of the uncertainty of how long to let it run to download a snapshot and didnt want to introduce too many changes in CI in one go


dnut:
What is the downside of running it on every commit? It's nice to get feedback for each commit.

src/net/socket_utils.zig
```diff
@@ -7,7 +7,7 @@ const Channel = @import("../sync/channel.zig").Channel;
 const std = @import("std");
 const Logger = @import("../trace/log.zig").Logger;
 
-pub const SOCKET_TIMEOUT_US: usize = 1 * std.time.us_per_s;
+pub const SOCKET_TIMEOUT_MS: usize = 1 * std.time.ms_per_s;
```
InKryption, line 10: Why the change to milliseconds? All the usage sites still interpret this value as being in microseconds.

0xNineteen:
> What is the downside of running it on every commit? It's nice to get feedback for each commit.

spinning up gossip to interact with the network and send packets seems wasteful imo - feels like we would only want to run these checks before we have it reviewed (to save reviewer time) or before we merge to main (to ensure theres no bugs) - if you want it every commit, you can always run it locally?

dnut:
> spinning up gossip to interact with the network and send packets seems wasteful

What resource are we wasting though? Is there an actual cost?

I feel that it is useful to understand whether my recent change broke anything. While I could run the checks locally, I might forget, or I might want to use my computer for something else that it interferes with. Also, the historical record of runs provides an audit trail that can be useful to diagnose issues that might have been overlooked.

0xNineteen:
good point - if you find it useful we can try it

InKryption (approved):

InKryption:
If there are no other outstanding contentions, I'd like for this to get merged.

PR #162, opened 2024-06-05
fix(core,accountsdb): Guarantee the layout of some struct types

InKryption:
by making them `extern struct`s.

### DIFF
--- a/src/accountsdb/accounts_file.zig
+++ b/src/accountsdb/accounts_file.zig
@@ -45,14 +45,14 @@ pub const AccountInFile = struct {
     len: usize = 0,
 
     /// info about the account stored
-    pub const StorageInfo = struct {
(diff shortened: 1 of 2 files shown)


0xNineteen:
lgtm - but can you confirm snapshot loading + validation works as expected with this change (testnet snapshots is fine)

InKryption:
I've not been able to get the validator to fully download and load a snapshot locally on my machine, but I did do some basic checks and verified that the compiler wasn't doing anything too strange with the layout of these types in release, so this change should have absolutely no effect other than ensuring it continues to behave the same as it has up until now.

If it's imperative to confirm this change is safe, it has to be done on a machine other than mine, because I consistently run into "download speed is too slow (2 MB/s) -- disconnecting".

0xNineteen (approved):

PR #164, opened 2024-06-06
fix(cmd): DNS improvements

InKryption:
Couldn't identify any issues with just using zig stdlib's version of `getAddressList`, it appears to work better internationally, and using it means we have one less dependency (we get rid of zigdig).

I also took the time to replace the string buliding with a formatting function for the log, and used an adapted ArrayHashMap(void, void) hash map to deduplicate entrypoints (on that note, should we issue a warning for duplicate domains or anything like that?).

### DIFF
--- a/src/cmd/cmd.zig
+++ b/src/cmd/cmd.zig
@@ -617,12 +617,12 @@ fn getEntrypoints(logger: Logger) !std.ArrayList(SocketAddr) {
             };
             const domain_str = entrypoint[0..domain_port_sep];
             if (domain_str.len == 0) {
-                logger.field("entrypoint", entrypoint).err("entrypoint domain not valid");
+                logger.errf("'{s}': entrypoint domain not valid", .{entrypoint});
                 return error.EntrypointDomainNotValid;
             }
             // parse port from string
             const port = std.fmt.parseInt(u16, entrypoint[domain_port_sep + 1 ..], 10) catch {
-                logger.field("entrypoint", entrypoint).err("entrypoint port not valid");
+                logger.errf("'{s}': entrypoint port not valid", .{entrypoint});
                 return error.EntrypointPortNotValid;
             };
 
@@ -631,7 +631,7 @@ fn getEntrypoints(logger: Logger) !std.ArrayList(SocketAddr) {
             defer addr_list.deinit();
 
             if (addr_list.addrs.len == 0) {
-                logger.field("entrypoint", entrypoint).err("entrypoint resolve dns failed (no records found)");
+                logger.errf("'{s}': entrypoint resolve dns failed (no records found)", .{entrypoint});
                 return error.EntrypointDnsResolutionFailure;
             }
 
--- a/src/cmd/cmd.zig
+++ b/src/cmd/cmd.zig
@@ -584,23 +584,53 @@ fn getMyDataFromIpEcho(
 }
 
 fn getEntrypoints(logger: Logger) !std.ArrayList(SocketAddr) {
+    const EntrypointSet = std.ArrayHashMap(void, void, struct {
+        // zig fmt: off
+        pub fn hash(_: @This(), _: void) u32 { unreachable; }
+        pub fn eql(_: @This(), _: void, _: void, _: usize) bool { unreachable; }
+        // zig fmt: on
+    }, true);
+    const EntrypointCtx = struct {
+        entrypoints: []const SocketAddr,
+        pub fn hash(_: @This(), entrypoint: SocketAddr) u32 {
+            const array, const len = entrypoint.toString();
+            return std.array_hash_map.hashString(array[0..len]);
+        }
+        pub fn eql(ctx: @This(), a: SocketAddr, _: void, b_index: usize) bool {
+            return a.eql(&ctx.entrypoints[b_index]);
+        }
+    };
+
     var entrypoints = std.ArrayList(SocketAddr).init(gpa_allocator);
+    errdefer entrypoints.deinit();
+
+    var entrypoint_set = EntrypointSet.init(gpa_allocator);
+    defer entrypoint_set.deinit();
+
+    try entrypoint_set.ensureTotalCapacity(config.current.gossip.entrypoints.len);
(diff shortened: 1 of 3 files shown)


src/cmd/cmd.zig
```diff
@@ -618,23 +617,53 @@ fn getMyDataFromIpEcho(
...
+    };
+
     var entrypoints = std.ArrayList(SocketAddr).init(gpa_allocator);
+    errdefer entrypoints.deinit();
+
+    var entrypoint_set = EntrypointSet.init(gpa_allocator);
+    defer entrypoint_set.deinit();
+
+    try entrypoint_set.ensureTotalCapacity(config.current.gossip.entrypoints.len);
+    try entrypoints.ensureTotalCapacityPrecise(config.current.gossip.entrypoints.len);
```
0xNineteen, line 644: why do you need both of these?

src/cmd/cmd.zig
```diff
@@ -618,23 +617,53 @@ fn getMyDataFromIpEcho(
...
+    const EntrypointCtx = struct {
+        entrypoints: []const SocketAddr,
+        pub fn hash(_: @This(), entrypoint: SocketAddr) u32 {
+            const array, const len = entrypoint.toString();
+            return std.array_hash_map.hashString(array[0..len]);
+        }
+        pub fn eql(ctx: @This(), a: SocketAddr, _: void, b_index: usize) bool {
+            return a.eql(&ctx.entrypoints[b_index]);
+        }
+    };
```
0xNineteen, line 635: this seems like overkill for something thats not performance critical - can you just use key to void approach with autocontext?

src/cmd/cmd.zig
```diff
@@ -618,23 +617,53 @@ fn getMyDataFromIpEcho(
...
-                return error.EntrypointDomainMissing;
-            };
-            const port_str = domain_and_port.next() orelse {
+            const domain_port_sep = std.mem.indexOfScalar(u8, entrypoint, ':') orelse {
                 logger.field("entrypoint", entrypoint).err("entrypoint port missing");
                 return error.EntrypointPortMissing;
             };
+            const domain_str = entrypoint[0..domain_port_sep];
+            if (domain_str.len == 0) {
+                logger.field("entrypoint", entrypoint).err("entrypoint domain not valid");
```
0xNineteen, line 654: i know you didnt write this code but were avoiding using the field approach because of performance issues  - can you use fmt instead

src/cmd/cmd.zig
```diff
@@ -643,30 +672,34 @@ fn getEntrypoints(logger: Logger) !std.ArrayList(SocketAddr) {
...
+            comptime fmt_str: []const u8,
+            fmt_options: std.fmt.FormatOptions,
+            writer: anytype,
+        ) !void {
+            for (0.., entrypoints_fmt.entrypoints) |i, entrypoint| {
+                if (i != 0) try writer.writeAll(", ");
+                try entrypoint.toAddress().format(fmt_str, fmt_options, writer);
+            }
+        }
+    };
```
0xNineteen, line 701: can you pull this struct out of the function? struct definitions inside isnt functions isnt standard and we cant reuse the logic elsewhere

0xNineteen (changes requested):

src/cmd/cmd.zig
InKryption: I think this is a fairly reasonable approach, it's a simple context that's re-using the items already stored in the existing array. Why duplicate state?

src/cmd/cmd.zig
InKryption: Is there somewhere you'd want to re-use this? Unscoping this single-use formatter doesn't seem terribly useful compared to just defining it exactly where it's needed.

Also, to be clear, by "standard", do you just mean according to the sig style guide?

src/cmd/cmd.zig
InKryption: If you mean the variables, `entrypoint_set` is a set of entrypoints that have already been set. I ensure capacity for both so that neither have to reallocate during the loop.

0xNineteen: oh sorry didnt notice the different variable names

src/cmd/cmd.zig
InKryption: f761151

src/cmd/cmd.zig
InKryption: f761151

src/cmd/cmd.zig
0xNineteen: the main problem is it adds unnecessary complexity which reduces the code readability - our first priority is simplicity and easy to ready code

the code block is only run once on startup so a small amount of state duplication is fine

InKryption: Alright I suppose, do you want to keep the context for possible future use, or should I remove it, and we can add it back on a needs-to-be basis?

src/cmd/cmd.zig
0xNineteen: does this approach still call toAddress on the entrypoints before formatting?

InKryption: It doesn't, but it also doesn't seem to change the behavior, because the `format` method is implemented directly on `SocketAddr`. Is there some substantially different behavior in some edge case?

src/cmd/cmd.zig
InKryption: 7430655d36eca21cf19986cf29dd5aaaf8e811ee

src/cmd/cmd.zig
0xNineteen: yeah great point there is no difference - can you remove all the formatting code and just log entrypoints.items directly

InKryption: well, that would make it different, the default format for a list is `{ item1, item2 }`

0xNineteen: that's ok the main goal is seeing the ip and ports - a extra {} is ok

InKryption: f8e0fbddaa2ca457bf1bbda52607a144072707b9

0xNineteen (approved):

0xNineteen:
looks great - great to remove a dep 🚀

PR #168, opened 2024-06-10
refactor: remove unused imports

dnut:
This removes any top-level private consts that are not used. This should be mostly uncontroversial, except there's this one line that is not an import:
```zig
const MAX_REQ_HEADER_SIZE = 8192;
```
It's not used, so I'm still inclined to remove it, but open to feedback.


InKryption:
According to git blame, MAX_REQ_HEADER_SIZE was added when IpEchoServer was first implemented, back when we were using `std.http`, which had the max request header size parameterized, wherein we passed this constant in as the value. Eventually we switched over to httpz, which overall has a different structure, and doesn't accept this.

I think it'd be fine to remove it, since it doesn't actually serve any purpose in the current implementation - even if we're moving away from httpz anytime soon, I think it'd be better to come up with a value based on whatever new constraints or advantages we end up facing, instead of keeping and using this outdated one.

PR #171, opened 2024-06-10
fix(shred-collector): SlotUnderflow bug, and generally improve error handling in shred processor

dnut:
I've been seeing SlotUnderflow get logged occasionally as an unhandled error in the shred processor. Service manager ends up restarting it, so it's not fatal, but it's not ideal to rely on this, for multiple reasons, such as:
- this underflow is not a problem and should not disrupt or log anything
- any shreds that have already been drained from the channel, but not processed, are discarded without being processed
- the error log message is overly generic
- restarting the loop has a performance penalty

This bug was introduced because I added `tracker.skipSlots` which was looking backwards from the shred's slot in order to skip the slots behind it. it's running into the underflow error when the slot has already been skipped. In that case it should just move on without complaining, because the job is already done. So the solution is just to ignore this error.

The error is now ignored, as you can see here: https://github.com/Syndica/sig/blob/4bd9aa1de5f494b1e2e978c125dcb97e6cc22cb0/src/shred_collector/shred_processor.zig#L68-L69

I also went a bit further to make the error handling more robust in general. The logic to process a single shred is moved into a separate function. Any error coming out of that function is handled in one place. This guarantees that any problem with a single shred does not interrupt the processing of other shreds. Also these errors are always logged. This should be fine because the shreds have already been verified. Any errors here means that the actual leader is misbehaving, or we have a bug in our code.

Also I made changes in a few places to add explicit error return types.

Another point worth mentioning is that I've implemented a workaround here to deal with the fact that zig does not support error payloads. I made the error context a function parameter that is typically ignored both inside and outside the function, unless there is an error.

### DIFF
--- a/src/shred_collector/service.zig
+++ b/src/shred_collector/service.zig
@@ -103,7 +103,7 @@ pub fn start(
     try service_manager.spawn(
         .{ .name = "Shred Processor" },
         shred_collector.shred_processor.runShredProcessor,
-        .{ deps.allocator, verified_shred_channel, shred_tracker },
+        .{ deps.allocator, deps.logger, verified_shred_channel, shred_tracker },
     );
 
     // repair (thread)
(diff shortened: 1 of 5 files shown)


InKryption (approved):

0xNineteen (approved):
error context is v nice

PR #177, opened 2024-06-18
docs(gossip): add the name of the rust equivalent type in the comments

aang114:
Currently - for some types, only a permalink to its Rust equivalent is added as the comment.

In this PR, the name of the Rust equivalent type is also added in the comment to improve searchability.

Note: This PR resolves the issue mentioned in https://github.com/orgs/Syndica/projects/2/views/5?pane=issue&itemId=60466783.


0xNineteen (approved):
thanks for the pr 🔥

PR #178, opened 2024-06-19
feat(net): initial go at IPv6 parsing

mathaou:
Replacing this with a similar approach outlined in the IPv4 parsing would be good for performance reasons, but given that this is a fallback in the first place maybe this approach is fine for now. If this way is a non-starter, I could take another go to try and make it cleaner / faster. Tried following contributions document and the example of the rest of the code in the codebase.

Essentially, I saw a TODO and thought this might be someplace where someone without much experience with the codebase as a whole could contribute something of relative substance.

### DIFF
--- a/src/net/net.zig
+++ b/src/net/net.zig
@@ -32,8 +32,41 @@ pub const SocketAddr = union(enum(u8)) {
     }
 
     pub fn parse(bytes: []const u8) !Self {
-        // TODO: parse v6 if v4 fails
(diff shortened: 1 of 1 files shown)


0xNineteen (approved):
looks great - thanks for the pr!

mathaou:
Missed the clause on the PR that said commits must be signed. I force pushed a new one containing all the others with a proper GPG key.

PR #180, opened 2024-06-21
docs(accountsdb): add the rust equivalent type in the comments

aang114:
In this PR, the name and permalink of the Rust equivalent type of a Zig type is added as a comment. This is done to improve searchability and to understand the similarities between Sig and the [Rust client](https://github.com/anza-xyz/agave) and also to understand what modifications were made in Sig.


0xNineteen (approved):

0xNineteen:
thanks for the pr - the doc comments add a ton of value 🚀

PR #184, opened 2024-06-25
refactor+fix(gossip):  update naming in active set and fix table trim condition

yewman:
**Overview**
1. Rename pruned_peers to peers in ActiveSet
2. Fix shouldTrim method in GossipTable so we trim when 90% capacity is reached, not 110%
3. Address TODO in src/bincode/bincode.zig
4. Update imports in modified files to fit new import style guide (draft here):

**Import Style Guide (Draft)**
Declare imports in the following groups, separated by a newline.

1. group @import's
2. group namespace aliases
3. group struct aliases
4. group function aliases 
5. group constant aliases

If it improves clarity, split group's into Sig and External imports / aliases.

**Namespaces and Aliases**
In my mind, it would be useful to establish some sort of convention for how we alias namespaces, structs, methods etc. For example in service.zig we access structs and methods from src/gossip/data.zig using a mix of approaches. Whilst this is not a big issue, in my opinion it does impact readability so is something to think about. 

One approach could be to aim to import structs, methods, constants, etc into the local namespace, and only use namespace aliases when there is a conflict or it improves readability by providing context. For example, importing ContactInfo into the local namespace should never be a problem, but perhaps when using Counter from prometheus in a seperate module prometheus.Counter may be more informative.

Anyway these are just some thoughts which popped up while updating the imports!

### DIFF
--- a/src/accountsdb/snapshots.zig
+++ b/src/accountsdb/snapshots.zig
@@ -1,25 +1,28 @@
 const std = @import("std");
+const zstd = @import("zstd");
+const sig = @import("../lib.zig");
+
+const bincode = sig.bincode;
+
 const ArrayList = std.ArrayList;
 const HashMap = std.AutoHashMap;
+const ZstdReader = zstd.Reader;
 
-const _genesis_config = @import("genesis_config.zig");
-const UnixTimestamp = _genesis_config.UnixTimestamp;
-const FeeRateGovernor = _genesis_config.FeeRateGovernor;
-const EpochSchedule = _genesis_config.EpochSchedule;
-const Rent = _genesis_config.Rent;
-const Inflation = _genesis_config.Inflation;
+const Account = sig.core.account.Account;
+const Hash = sig.core.hash.Hash;
+const Slot = sig.core.time.Slot;
+const Epoch = sig.core.time.Epoch;
+const Pubkey = sig.core.pubkey.Pubkey;
+const UnixTimestamp = sig.accounts_db.genesis_config.UnixTimestamp;
+const FeeRateGovernor = sig.accounts_db.genesis_config.FeeRateGovernor;
+const EpochSchedule = sig.accounts_db.genesis_config.EpochSchedule;
+const Rent = sig.accounts_db.genesis_config.Rent;
+const Inflation = sig.accounts_db.genesis_config.Inflation;
+const SlotHistory = sig.accounts_db.sysvars.SlotHistory;
 
-const Account = @import("../core/account.zig").Account;
-const Hash = @import("../core/hash.zig").Hash;
-const Slot = @import("../core/time.zig").Slot;
-const Epoch = @import("../core/time.zig").Epoch;
-const Pubkey = @import("../core/pubkey.zig").Pubkey;
-const bincode = @import("../bincode/bincode.zig");
-const defaultArrayListOnEOFConfig = @import("../utils/arraylist.zig").defaultArrayListOnEOFConfig;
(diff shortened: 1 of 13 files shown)


src/gossip/active_set.zig
```diff
@@ -22,40 +22,40 @@ const BLOOM_MAX_BITS: usize = 1024 * 8 * 4;
 
 pub const ActiveSet = struct {
     // store pubkeys as keys in gossip table bc the data can change
-    pruned_peers: std.AutoHashMap(Pubkey, Bloom),
+    peers: std.AutoHashMap(Pubkey, Bloom),
```
0xNineteen, line 25: can you add a comment mentioning how the bloom tracks origins which have been pruned from peers that we push to above this field?

0xNineteen (changes requested):
one small thing - otherwise lgtm

src/gossip/active_set.zig
yewman: Yeah cool ill add that, on another note I'm not quite sure what the "// store pubkeys..." comment is getting at? Could you fill me in on that?

0xNineteen: i think a previous approach would store the full contact info's in the hashmap to reduce looking up the value in the gossip table - but since the values in the gossip table can change, we only store the pubkey and do a lookup when we need to read their contact info

src/accountsdb/snapshots.zig
```diff
@@ -15,7 +15,7 @@ const Slot = @import("../core/time.zig").Slot;
 const Epoch = @import("../core/time.zig").Epoch;
 const Pubkey = @import("../core/pubkey.zig").Pubkey;
 const bincode = @import("../bincode/bincode.zig");
-const defaultArrayListOnEOFConfig = @import("../utils/arraylist.zig").defaultArrayListOnEOFConfig;
```
0xNineteen, line 18: for any new imports were following the convention below (see `cmd.zig`)
```
const sig = @import("../lib.zig")
const X = sig.bincode.arraylist.X;
```
we only recently decided this and are slowly moving over imports to follow this - could you change these imports to follow this format too?

src/gossip/table.zig
```diff
@@ -698,11 +698,10 @@ pub const GossipTable = struct {
         bincode.free(self.allocator, versioned_value.value.data);
     }
 
+    // Trim when over 90% of max capacity
```
0xNineteen, line 701: method comments need to have three - ie, '///' 

https://ziglang.org/documentation/master/#Doc-Comments

0xNineteen (changes requested):
few small things - then should be good

src/bloom/bloom.zig
```diff
@@ -1,16 +1,17 @@
...
+const bincode = sig.bincode;
 
 const RndGen = std.rand.DefaultPrng;
+const ArrayList = std.ArrayList;
+
+const DynamicArrayBitSet = sig.bloom.bit_set.DynamicArrayBitSet;
+const BitVec = sig.bloom.bit_vec.BitVec;
+const BitVecConfig = sig.bloom.bit_vec.BitVecConfig;
+const ArrayListConfig = bincode.arraylist.ArrayListConfig;
+const FnvHasher = @import("../crypto/fnv.zig").FnvHasher; // SHOULD CRYPTO BE ADDED TO LIB
```
0xNineteen, line 14: yes it should - can you add it?

0xNineteen:
just responding to your comment - then should be good for merge

0xNineteen (approved):
lgtm - great first pr 🔥

PR #185, opened 2024-06-27
fix(cli): amend CLI details & fix crash

InKryption:
The validator was silently crashing most of the time due to the non-existence of the accountsb/snapshot directory. Refactored a couple other adjacent things.

### DIFF
--- a/src/cmd/cmd.zig
+++ b/src/cmd/cmd.zig
@@ -198,7 +198,7 @@ var force_new_snapshot_download_option = cli.Option{
 
 var snapshot_dir_option = cli.Option{
     .long_name = "snapshot-dir",
-    .help = "path to snapshot directory (where snapshots are downloaded and/or unpacked to/from) - default: test_data/",
+    .help = "path to snapshot directory (where snapshots are downloaded and/or unpacked to/from) - default: ledger/accounts_db",
     .short_alias = 's',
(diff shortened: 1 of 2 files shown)


src/cmd/cmd.zig
```diff
@@ -435,13 +438,11 @@ fn validator() !void {
...
-    // cli parsing
-    const snapshot_dir_str = config.current.accounts_db.snapshot_dir;
-    const n_cpus = @as(u32, @truncate(try std.Thread.getCpuCount()));
-    var n_threads_snapshot_load: u32 = @intCast(config.current.accounts_db.num_threads_snapshot_load);
-    if (n_threads_snapshot_load == 0) {
-        n_threads_snapshot_load = n_cpus;
-    }
+    const n_threads_snapshot_load: u32 = blk: {
+        const n_threads_snapshot_load: u32 = config.current.accounts_db.num_threads_snapshot_load;
+        if (n_threads_snapshot_load != 0) break :blk n_threads_snapshot_load;
```
0xNineteen, line 443: nit: can we change this to be symmetric and have if statement check if == 0 -- its a lot easier to reason about things in the true case instead of the negative (ie, == 0 means only one value of x runs the branch, != 0 is inf values run the branch)

0xNineteen (changes requested):

PR #188, opened 2024-07-03
improvement(accountsdb): add mapping in DiskMemoryAllocator that maps a file's mmap-ed memory address to its filename

aang114:
In this PR, a HashMap field has been added to the `DiskMemoryAllocator` that maps the mmap-ed memory address of a file to its filename so that the file can be deleted from storage when `DiskMemoryAllocator::free()` is called.

Note: This PR attempts to resolve part of the issue mentioned in https://github.com/orgs/Syndica/projects/2?pane=issue&itemId=56221388: 
> rn `free(mem_ptr)` does nothing - all the cleanup happens on `deinit()` - if we keep a `map(mem_ptr, file_path)` then we can properly delete disk memory which we no longer need while the process is running

### DIFF
--- a/src/accountsdb/db.zig
+++ b/src/accountsdb/db.zig
@@ -122,7 +122,7 @@ pub const AccountsDB = struct {
                 logger.infof("using disk index in {s}", .{disk_file_suffix});
 
                 const ptr = try allocator.create(DiskMemoryAllocator);
-                ptr.* = DiskMemoryAllocator.init(disk_file_suffix, allocator);
+                ptr.* = DiskMemoryAllocator.init(allocator, disk_file_suffix);
 
                 break :blk .{ ptr, ptr.allocator() };
             } else {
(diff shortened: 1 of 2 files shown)


src/accountsdb/db.zig
```diff
@@ -121,8 +121,9 @@ pub const AccountsDB = struct {
                 const disk_file_suffix = try std.fmt.allocPrint(allocator, "{s}/bin", .{disk_dir});
                 logger.infof("using disk index in {s}", .{disk_file_suffix});
 
+                var gpa = std.heap.GeneralPurposeAllocator(.{}){};
                 const ptr = try allocator.create(DiskMemoryAllocator);
-                ptr.* = DiskMemoryAllocator.init(disk_file_suffix);
+                ptr.* = DiskMemoryAllocator.init(disk_file_suffix, gpa.allocator());
```
InKryption, line 126: This is a UAF error.

src/accountsdb/index.zig
```diff
@@ -925,11 +925,15 @@ pub const DiskMemoryAllocator = struct {
     count: usize = 0,
     mux: std.Thread.Mutex = .{},
 
+    /// HashMap that maps the Mmap-ed memory address of a file to the file's index number
+    hashmap: std.AutoHashMap([*]u8, usize),
```
InKryption, line 929: I'm not a huge fan of the fact that this introduces a dependency on another allocator which has to potentially also be thread safe. I would like to instead see an approach where the metadata is placed in (disk) memory before or after the user-facing buffer.

Issue #157, opened 2024-06-05
gossip: Random test failure in gossip test "test filtering values works"


InKryption:
### Description

At random, this line in the referenced test will fail:

https://github.com/Syndica/sig/blob/a5d1f34d8a93f0d101e58e2a5aa2a8a5126a632f/src/gossip/pull_response.zig#L143

I currently do not have a stacktrace demonstrating this behavior, because I do not know the circumstances under which this failure occurs.

### How to Reproduce the Bug

run `zig build test -Dfilter="gossip.pull_response: test filtering values works"` and pray

### Additional Context

It should be noted that the function being tested, named `filterSignedGossipDatas`, currently instantiates and uses a PRNG seeded with `std.time.milliTimeStamp()`, which could explain its seemingly random and unreproducible behavior.

What would be ideal is capturing a stack trace of this occurring, as well as the exact or approximate timestamp at which this failure occurs, to diagnose what random values can cause such a failure, and identify what circumstances they lead to which must be guarded against.

At the time of writing I plan on pushing a change which would help in the above goals, without needing to actively search for this mythical seed.

InKryption:
Sorry, I'd like to keep this open until we actually have it solved. The code for helping us solve it is in place, it just needs to randomly occur at some point, or the code needs to fundamentally change such that we can guarantee it won't.

0xNineteen:
just hit it - The failing seed is: '1718031575764'

InKryption:
Can confirm that using '1718031575764' as the seed consistently reproduces the failure state.

dnut:
another one: 1718158590375

InKryption:
Can't reproduce the behavior using '1718158590375'.

0xNineteen:
yeah pulling out the random seeds to deterministic would be good - we can create a ticket - kind of low priority right now

InKryption:
New failing seed caught in CI:
https://github.com/Syndica/sig/actions/runs/9811929838/job/27094994026?pr=189
Seed number is '1720202203195'.

InKryption:
https://github.com/Syndica/sig/actions/runs/9897406056/job/27341819918#step:4:1

We got another one: '1720726415143'

dnut:
This is still a problem

PR #189, opened 2024-07-05
refactor(gossip) Add methods, data structures, and edits to improve memory safety in gossip 

yewman:
**Memory Leaks**

- Found a few more memory leaks while thinking about clone / deinit implementation. 
- Implemented basic fix - call bincode.free 
- Ideally should add test cases for unhappy paths

**ThreadSafeContactInfo**
- In a number of places, ContactInfo was being passed around outside of lock boundaries for purposes not requiring access to the heap memory contained within. Whilst these usages were not causing any errors, creating and using an explicitly thread safe type seems like a good idea as makes the code more robust, simpler to understand, and easier to debug. 

**Clone**
- Implement clone methods for instances where we need access to the heap allocated memory of gossip values outside of a table lock scope.

**Other**
- Add saturating subs for timestamp operations
- Remove parameter to processMessages - no need to parse a parameter which is contained in self


src/gossip/ping_pong.zig
```diff
@@ -129,12 +130,12 @@ pub const PingCache = struct {
...
     ) error{OutOfMemory}!Self {
-        assert(rate_limit_delay_ns <= ttl_ns / 2);
+        std.debug.assert(rate_limit_delay_ns <= ttl_ns / 2);
         return Self{
             .ttl_ns = ttl_ns,
             .rate_limit_delay_ns = rate_limit_delay_ns,
-            .pings = try LruCache(.non_locking, PubkeyAndSocketAddr, Instant).init(allocator, cache_capacity),
-            .pongs = try LruCache(.non_locking, PubkeyAndSocketAddr, Instant).init(allocator, cache_capacity),
+            .pings = try LruCache(.non_locking, PubkeyAndSocketAddr, std.time.Instant).init(allocator, cache_capacity),
+            .pongs = try LruCache(.non_locking, PubkeyAndSocketAddr, std.time.Instant).init(allocator, cache_capacity),
```
InKryption, line 138: In contrast but not opposition to my other comment, I think this kind of thing instead warrants extracting the type into a named local declaration somewhere to avoid this kind of duplication, ie:
```zig
const Pings = LruCache(.non_locking, PubkeyAndSocketAddr, std.time.Instant);
const Pongs = LruCache(.non_locking, PubkeyAndSocketAddr, std.time.Instant);
```
maybe only one type like `PingsOrPongs` if the type of these fields are "intertwined" (as in, probably never going to differ).

InKryption (commented):
Just a few notes, otherwise LGTM.

InKryption:
Th failing CI is due to #157.

src/gossip/ping_pong.zig
```diff
@@ -112,9 +113,9 @@ pub const PingCache = struct {
     rate_limit_delay_ns: u64,
     // Timestamp of last ping message sent to a remote node.
     // Used to rate limit pings to remote nodes.
-    pings: LruCache(.non_locking, PubkeyAndSocketAddr, Instant),
+    pings: LruCache(.non_locking, PubkeyAndSocketAddr, std.time.Instant),
     // Verified pong responses from remote nodes.
-    pongs: LruCache(.non_locking, PubkeyAndSocketAddr, Instant),
+    pongs: LruCache(.non_locking, PubkeyAndSocketAddr, std.time.Instant),
```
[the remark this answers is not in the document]
yewman: I generally prefer it as well, I noticed that the rest of the gossip lib tends to use the fully-qualified approach so I thought I'd update it for consistency.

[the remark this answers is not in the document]
InKryption: There is some inconsistency, some code fully-qualifies, other code does stuff like `std.ArrayList` and similar.

src/gossip/service.zig
```diff
@@ -1208,10 +1211,9 @@ pub const GossipService = struct {
             defer gossip_table_lock.unlock();
             var gossip_table: *GossipTable = gossip_table_lock.mut();
 
-            for (pull_requests.items) |*req| {
-                const caller = req.value.id();
+            for (pull_requests.items) |req| {
```
0xNineteen, line 1214: should keep this as a pointer in this for loop (`*pull_request`) - without the pointer it will likely lead to a stack copy which we dont need since we only access the fields

0xNineteen (changes requested):

src/gossip/service.zig
```diff
@@ -906,7 +907,7 @@ pub const GossipService = struct {
             defer gossip_table_lock.unlock();
 
             const gossip_table: *const GossipTable = gossip_table_lock.get();
-            break :blk gossip_table.getEntriesWithCursor(&buf, push_cursor);
+            break :blk gossip_table.getEntriesWithCursor(&buf, push_cursor); // MEM1: CLONE HERE; OR
```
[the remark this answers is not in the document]
yewman: Yeah I left this one because I actually wanted to briefly discuss the best approach here. We can either move the logic responsible for building messages into the scope of the read lock, or clone the entries we pull out of the gossip table. This is an instance where we cant use thread safe contact info because we are serialising the heap allocated data when constructing the messages. I don't have a strong intuition over which approach would be better performance wise, so unless one of the two above options jumps out at you as the clear choice perhaps Ill need to figure out a way to benchmark

src/gossip/service.zig
```diff
@@ -769,7 +770,7 @@ pub const GossipService = struct {
         while (!self.exit.load(.unordered)) {
             const top_of_loop_ts = getWallclockMs();
```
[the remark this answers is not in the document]
yewman: Quick question, there is not default implementation for starting a std.time.Timer at an effective 'zero' state, thus timing variables which are referenced before their non default assignment won't fit the Timer well. I think it makes sense to just create a function to create this 'zero' state for the std.Timer but thought I'd check before actually implementing it

0xNineteen (commented):
lgtm - should be good to merge after you add the docs

0xNineteen (approved):
lgtm - should get @dnut approval too since he has a lot of context around this problem

src/gossip/table.zig
```diff
@@ -398,22 +399,23 @@ pub const GossipTable = struct {
...
+            return ThreadSafeContactInfo.fromContactInfo(v.value.data.ContactInfo);
         } else {
-            return self.converted_contact_infos.get(pubkey);
+            return ThreadSafeContactInfo.fromContactInfo(self.converted_contact_infos.get(pubkey) orelse return null);
         }
     }
 
-    pub fn genericGetWithCursor(
+    pub fn genericGetEntriesWithCursor(
+        allocator: ?std.mem.Allocator,
```
dnut, line 413: In all of the usages where null is passed here, the return contains a pointer to data owned by the table.

- getVotesWithCursor
- getEpochSlotsWithCursor
- getDuplicateShredsWithCursor

So it would be unsafe to use the data returned by those methods across a lock boundary.

Those methods are not actually used anywhere. My first instinct is to remove them. Not sure why we have unused methods. I'm guessing it's for feature parity with the agave implementation. In that case, I would probably just change them to accept an allocator. It seems preferable to default to safety and require close inspection to be fast, rather than defaulting to speed and requiring close inspection to be safe.

Also I would probably make this method private and document that it may be unsafe to use the return value across lock boundaries if the allocator is not set.

src/gossip/service.zig
```diff
@@ -759,18 +754,17 @@ pub const GossipService = struct {
     /// this includes sending push messages, pull requests, and triming old
     /// gossip data (in the gossip_table, active_set, and failed_pull_hashes).
     fn buildMessages(self: *Self) !void {
-        var last_push_ts: u64 = 0;
-        var last_stats_publish_ts: u64 = 0;
-        var last_pull_req_ts: u64 = 0;
+        var loop_timer = sig.time.Timer.start() catch unreachable;
+        var push_timer = sig.time.Timer.start() catch unreachable;
+        var pull_req_timer = sig.time.Timer.start() catch unreachable;
+        var stats_publish_timer = sig.time.Timer.start() catch unreachable;
```
dnut, line 760: Is there a disadvantage to using `try` here? Returning an error seems preferable to `unreachable` when performance is not a concern.

src/gossip/service.zig
```diff
@@ -828,20 +823,18 @@ pub const GossipService = struct {
...
-                last_stats_publish_ts = getWallclockMs();
             }
 
             // sleep
-            const elapsed_ts = getWallclockMs() - top_of_loop_ts;
-            if (elapsed_ts < GOSSIP_SLEEP_MILLIS) {
-                const time_left_ms = GOSSIP_SLEEP_MILLIS - elapsed_ts;
+            if (loop_timer.read().as_millis() < GOSSIP_SLEEP_MILLIS) {
+                defer loop_timer.reset();
+                const time_left_ms = GOSSIP_SLEEP_MILLIS -| loop_timer.read().as_millis();
```
dnut, line 837: We could probably start replacing all of this stuff with `Duration` at some point. Doesn't need to be in this PR, just thinking out loud.

src/gossip/service.zig
```diff
@@ -1220,23 +1215,18 @@ pub const GossipService = struct {
             defer ping_cache_lock.unlock();
             var ping_cache: *PingCache = ping_cache_lock.mut();
 
-            var peers = try ArrayList(ContactInfo).initCapacity(self.allocator, pull_requests.items.len);
+            var peers = try ArrayList(ThreadSafeContactInfo).initCapacity(self.allocator, pull_requests.items.len);
             defer peers.deinit();
             var arena = std.heap.ArenaAllocator.init(self.allocator);
             defer arena.deinit();
-            {
-                var gossip_table_lock = self.gossip_table_rw.read();
```
dnut, line 1228: nice

dnut (commented):
Looks good, I just have a few minor concerns.

I like the approach of just returning a lean contact info with the bare minimum data, instead of requiring more allocations. It actually moves us back away from the new ContactInfo, closer towards something resembling LegacyContactInfo. If we go all-in on this approach, at some point we could probably get rid of the contact info mapping that I added to facilitate the migration to ContactInfo.
```
converted_contact_infos: AutoArrayHashMap(Pubkey, ContactInfo),
```
I can imagine some approaches where a caller could specify exactly the minimum amount of data they need about a contact. Something that feels like graphql.

src/bincode/bincode.zig
```diff
@@ -149,6 +149,15 @@ pub fn read(allocator: std.mem.Allocator, comptime U: type, reader: anytype, par
...
             }
+
+            const init_field = "!bincode-config:init";
+            if (@hasDecl(T, init_field)) {
+                const field_config = @field(T, init_field);
+                if (field_config.init_fn) |init| {
+                    init(&data);
+                }
+            }
+
```
InKryption, line 160: This doesn't actually respect the added field in `FieldConfig`, making either one or the other redundant, or the former is missing code that actually uses it.

But beyond that, both this and the added field in `FieldConfig` is what I was wanting to avoid. It's this kind of functionality that should, in my view, be added to a field via a custom `serialize` function, leaving bincode logic simple.

InKryption (commented):
Need to go over some details concerning bincode.

src/bincode/bincode.zig
0xNineteen: for more context - the approach ink's talking about is similar to [this](https://github.com/Syndica/sig/blob/20a6d3dd4a2f47ceea50ba2faad9e622a60ecc54/src/bincode/arraylist.zig#L17) - where after the serialization, you populate the cache in the same `serialize` function

InKryption: We spoke about this in a huddle, we'll proceed with this as is, but I have a plan to improve the situation, which is out of scope for this PR.

src/gossip/service.zig
0xNineteen: created a ticket - https://github.com/orgs/Syndica/projects/2/views/1?pane=issue&itemId=71010101

0xNineteen (approved):

0xNineteen (approved):

PR #192, opened 2024-07-10
feat(docs) Add import style guide

yewman:
Implement style guide roughly defined here https://github.com/Syndica/sig/pull/184

Note:

Consideration of using local imports within modules so that lib files expose only the public interface of a module
Could use the following approach for imports  (src/gossip/service.zig)

const sig = https://github.com/import("../lib.zig");
const lib = struct {
const table = https://github.com/import("table.zig");
const active_set = https://github.com/import("active_set.zig");
};

const SocketAddr = sig.net.SocketAddr;
const ServiceManager = sig.utils.service_manager.ServiceManager;
const GossipTable = lib.table.GossipTable;
const HashTimeQueue = lib.table.HashTimeQueue;
const AutoArrayHashSet = lib.table.AutoArrayHashSet;

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -10,6 +10,53 @@ See the [readme](../readme.md#-setup).
 
 ## Style Guide
 
+### Imports 
+
+#### Sig Dependencies
+By convention, all internal dependencies should be defined as aliases of fully qualified paths from the root module. For example, within 'src/gossip/message.zig' we should import types from 'src/gossip/data.zig' in the following manner:
+```zig
+const sig = @import("../lib.zig");
+
+const GossipData = sig.gossip.data.GossipData;
+```
+
+#### Grouping
+Group import statements and alias definitions into the following categories, separated by a newline:
+
+1. @import statements
+2. namespace aliases
+3. struct aliases
+4. function aliases
+5. constant aliases
+
+If it improves clarity, split groups into external and sig imports/aliases, otherwise list external imports/aliases before sig imports/aliases. Within groups, try to follow alphabetical order with respect to fully qualified namespace i.e.
+```zig
+// Import statements
+const std = @import("std");
+const sig = @import("../lib.zig");
+
+// Namespace aliases
+const bincode = sig.bincode;
+const pull_request = sig.gossip.pull_request;
+const pull_response = sig.gossip.pull_response;
+
+// External aliases
+const EndPoint = network.EndPoint;
+const UdpSocket = network.Socket;
+const ArrayList = std.ArrayList;
+const AtomicBool = std.atomic.Value(bool);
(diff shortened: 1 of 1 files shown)


docs/CONTRIBUTING.md
```diff
@@ -10,6 +10,60 @@ See the [readme](../readme.md#-setup).
...
+#### Sig Dependencies
+By convention, all internal dependencies should be defined as aliases of fully qualified paths from the root module. For example, within 'src/gossip/message.zig' we should import types from 'src/gossip/data.zig' in the following manner:
+```zig
+const sig = @import("../lib.zig");
+
+const GossipData = sig.gossip.data.GossipData;
+```
+
+NOTE:
+- I've got my reservations about this approach within a lib, because it renders pub useless in terms of defining a modules public interface as we need to mark every import in lib.zig as pub
```
InKryption, line 24: What renders it useless? `pub` acts the same way in between all files, so I'm not understanding what tradeoff is being referred to here.

yewman: Oh right give me a second sorry

yewman: yeah i guess to achieve what I mean you would need all module imports to use the @import statement

yewman: Thoughts on this

const sig = @import("../lib.zig");
const lib = struct {
    const table = @import("table.zig");
    const active_set = @import("active_set.zig");
};

InKryption: I don't know if that would really improve clarity. More than anything I think it would create confusion when you reach for part of the public API and it's not there. But I'm not strongly against it either.

yewman: Yeah thats fair

yewman: Do you think just having everything public makes the most sense? I don't have a strong opinion either

InKryption: Well, inside every `lib.zig` yeah, since that is the public facing API - it's at that layer we can decide what should and shouldn't be exposed. If one lib needs something from another lib that isn't accessible, to me that would indicate that it should be accessible, otherwise we wind up in a weird situation where the sub-libraries are coupled but separate - separation of concerns and all that.

Right now, we don't need anything too complicated, and hopefully it stays that way. But if we are to ever need something like an "internal component" in a given lib, we needn't have a hard rule about needing to import that through the public API, a direct `@import` would make sense.

yewman: I actually find it more explicit as I would never expect to be accessing a dependency which is contained in the same module via the modules public interface. This is probably a reflection of the fact that most of my experience stems from languages which do not facilitate, or discourage circular imports.

0xNineteen (approved):
nice

0xNineteen (approved):

PR #194, opened 2024-07-11
feat(blockstore): blockstore database

dnut:
This implements a bare Blockstore database.

There is an abstract `Database` type in `database.zig` which defines the concept of a key-value store. A database actually is composed of not just one store, but multiple separate k-v stores called "column families". Each column family is typed, storing a specific key type and value. In that way, a column family behaves sort of like a hashmap. A database may contain an arbitrary number of these maps. In the code, a column family is defined by instantiating the comptime struct `ColumnFamily` with a few types (key, value, and serializers). Column Family definitions are passed as a parameter to `Database` at comptime (for runtime performance).

`Database` must be implemented with a backend, so I have implemented two backends.
- `SharedHashMapDB`: A basic `Database` POC that just uses hashmaps under the hood
- `RocksDB`: A `Database` that is implemented by calling the zig wrapper around the C++ library: https://github.com/Syndica/rocksdb-zig

The "Blockstore database" is called `BlockstoreDB`. `BlockstoreDB` is just an instance of `Database` that includes a specification of 20 column families (such as slot metadata, coding shred, data shred) that are relevant for the blockstore. These column families are defined in `schema.zig` and they store data types that are defined in `meta.zig`. BlockstoreDB can be used with any Database implementation.


src/utils/types.zig
```diff
@@ -43,3 +43,21 @@ pub fn ParamsTuple(comptime function: anytype) type {
...
+}
+
+pub fn fieldNames(comptime Struct: type) [@typeInfo(Struct).Struct.fields.len][:0]const u8 {
+    const fields = @typeInfo(Struct).Struct.fields;
+    var names: [fields.len][:0]const u8 = undefined;
+    for (fields, 0..) |field, i| {
+        names[i] = field.name;
+    }
+    return names;
+}
```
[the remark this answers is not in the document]
dnut: yes cbb230408992569d1a041c867650173f6a4cd1a9

src/blockstore/hashmap_db.zig
```diff
@@ -0,0 +1,254 @@
...
+        self: *Self,
+        comptime cf: ColumnFamily,
+        comptime cf_index: usize,
+        key: cf.Key,
+        value: cf.Value,
+    ) !void {
+        return try self.instructions.append(self.allocator, .{ .put = .{
+            cf_index,
+            try cf.key().serializeAlloc(self.allocator, key),
+            try cf.value().serializeAlloc(self.allocator, value),
```
InKryption, line 182: Careful with these kinds of nested expressions, this is hiding a leak in the error path. If the first try succeeds, but the second fails, the first is leaked. If both of these trys succeed, but the try on the append fails, that's two leaks. Best to split these kind of things up into assignments with respective `errdefer free()`s.

src/blockstore/hashmap_db.zig
```diff
@@ -0,0 +1,254 @@
...
+        return try self.instructions.append(self.allocator, .{ .delete = .{
+            cf_index,
+            try cf.key().serializeAlloc(self.allocator, key),
+        } });
+    }
+};
+
+const SharedHashMap = struct {
+    allocator: Allocator,
+    map: std.StringHashMap([]const u8),
```
InKryption, line 201: Might as well make the map unmanaged if we're carrying its allocator anyways.

src/utils/collections.zig
```diff
@@ -62,3 +62,265 @@ pub fn RecyclingList(
...
+    try expect(eql(u8, &.{1}, set.range(1, 2)));
+    try expect(eql(u8, &.{}, set.range(1, 1)));
+    try expect(eql(u8, &.{ 3, 5 }, set.range(2, 6)));
+    try expect(eql(u8, &.{ 3, 5 }, set.range(3, 6)));
+    try expect(eql(u8, &.{5}, set.range(4, 6)));
+    try expect(eql(u8, &.{5}, set.range(5, 6)));
+    try expect(eql(u8, &.{3}, set.range(3, 4)));
+    try expect(eql(u8, &.{}, set.range(3, 3)));
+    try expect(eql(u8, &.{}, set.range(2, 3)));
+    try expect(eql(u8, &.{}, set.range(2, 2)));
```
InKryption, line 308: Similar comment as in the insert_shreds PR, we get better error messages if we use the more specific `std.testing.expect*` functions, in this case `std.testing.expectEqualSlices`.

src/utils/collections.zig
```diff
@@ -62,3 +62,265 @@ pub fn RecyclingList(
...
+        try expect(find(u8, &items, 4, w).after == 2);
+        try expect(find(u8, &items, 5, w).found == 3);
+        try expect(find(u8, &items, 6, w) == .greater);
+    }
+    expect(find(u8, &items, 3, .any).found == 1) catch {
+        try expect(find(u8, &items, 3, .any).found == 2);
+    };
+    try expect(find(u8, &items, 3, .first).found == 1);
+    try expect(find(u8, &items, 3, .last).found == 2);
+}
```
InKryption, line 326: Same as other test comment, `expectEqual` here please.

src/blockstore/hashmap_db.zig
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    pub fn deinit(self: Self) void {
+        self.instructions.deinit(self.allocator);
+    }
+
+    pub fn put(
+        self: *Self,
+        comptime cf: ColumnFamily,
+        comptime cf_index: usize,
```
InKryption, line 175: Is there a reason this is `comptime`? Does the interface expect it to be, or is this part of some future work?

src/blockstore/hashmap_db.zig
```diff
@@ -0,0 +1,254 @@
...
+        return .{
+            .allocator = self.allocator,
+            .data = ret,
+        };
+    }
+
+    pub fn delete(
+        self: *Self,
+        comptime cf: ColumnFamily,
+        comptime cf_index: usize,
```
InKryption, line 108: Same question here, and in general, what's the meaning of these `comptime cf_index`s which appear to only be used at runtime?

InKryption (changes requested):
A couple off details, improvable tests, and a question, but very solid work.

src/blockstore/hashmap_db.zig
dnut: Yeah, this doesn't need to be comptime here. I should probably change this.

The idea is that I want `Database.put` to calculate the index at comptime. I must have put this here as an assertion that Database is actually doing this. But I can see how it is redundant, confusing, and doesn't help with separation of concerns.

dnut: 25f37dbdd8f25ac29873e4df59da491984b22ed1

src/utils/collections.zig
dnut: 6a82fd1125d3a723db15420c757c8dde63634711

src/utils/collections.zig
dnut: 6a82fd1125d3a723db15420c757c8dde63634711

src/blockstore/hashmap_db.zig
dnut: cf54686a55196c62abae1a50badfb538e79a052e

src/blockstore/hashmap_db.zig
dnut: 0aa78ac7f94501f2203178748849b47fe16ec7be

I'm afraid we may be missing errdefers throughout the codebase. It's easy to overlook these negative cases.

InKryption: Yeah, it's likely we are - there are ways to snuff out edge case memory leaks we could/should look to utilizing, namely `std.testing.checkAllAllocationFailures`. We could create both general and targeted fuzz tests for that.
But it's still a win when we can catch them early in review like this.

src/blockstore/rocksdb.zig
```diff
@@ -0,0 +1,198 @@
...
+        };
+    }
+
+    pub fn delete(
+        self: *Self,
+        comptime cf: ColumnFamily,
+        cf_index: usize,
+        key: cf.Key,
+    ) !void {
+        const key_bytes = try cf.key().serializeToRef(self.allocator, key);
```
0xNineteen, line 135: theres a lot of abstract types (key: Value, Serializer, KeySerializer) which i get are to support different types of serialization/types of dbs but it gets extremely confusing very quickly - for an initial implementation we should only use these types of abstractions only when necessary

if we want to keep this - we should add more docs (both doc comments and more of a post explaning how the abstractions work)

src/blockstore/transaction_status.zig
```diff
@@ -0,0 +1,245 @@
...
+    /// Transaction exceeded max loaded accounts data size cap
+    // #[error("Transaction exceeded max loaded accounts data size cap")]
+    MaxLoadedAccountsDataSizeExceeded,
+
+    /// LoadedAccountsDataSizeLimit set for transaction must be greater than 0.
+    // #[error("LoadedAccountsDataSizeLimit set for transaction must be greater than 0.")]
+    InvalidLoadedAccountsDataSizeLimit,
+
+    /// Sanitized transaction differed before/after feature activiation. Needs to be resanitized.
+    // #[error("ResanitizationNeeded")]
```
0xNineteen, line 231: can we remove these \#\[errors\]

0xNineteen (changes requested):
tldr; the abstractions to support different dbs/serializations are pretty confusing for an initial implementation - would ideally want to simplify it greatly

we should also add a significant amount of docs to this pr explaining the code layout for blockstore + design decisions before merging

src/utils/collections.zig
```diff
@@ -62,3 +62,266 @@ pub fn RecyclingList(
...
+                }
+            }{ .items = self.map.unmanaged.entries.slice() });
+            self.is_sorted = true;
+        }
+    };
+}
+
+/// binary search that is very specific about the outcome.
+/// only works with numbers
+fn find(
```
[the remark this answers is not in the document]
dnut: I'm not sure what you mean by "just one round". This function does not return until either the item is found or until it concludes that the item is not in the list.

My thought when naming this was that "find" is what is relevant to callers. Callers want to find the item in the list. The fact that it's a binary search is just an implementation detail. But I'm ok renaming this to binarySearch since its the obvious approach to find an item in a sorted list.

src/blockstore/database.zig
```diff
@@ -0,0 +1,248 @@
...
+            const was_deleted = try db.delete(cf1, 123);
+            _ = was_deleted;
+            // try std.testing.expect(was_deleted);
+            const not_now = try db.get(cf1, 123);
+            try std.testing.expect(null == not_now);
+        }
+    };
+}
+
+pub fn BatchImpl(
```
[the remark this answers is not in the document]
dnut: Regarding the separation between Batch and Database, they inevitably need to be two separate structs, regardless of whether the abstraction layer exists. In rocksdb itself, a write batch is a distinct struct with separate put and delete methods from the database.

Once I start working on the performance of the blockstore, I was planning to swap rocksdb out with alternatives. lmdb seems like it could be better. I don't think this is very far away. I'm not sure what you mean about it being high risk. I do see vendor lock-in as a risk though.

I understand that abstractions like this have a cost of additional click-throughs. I guess I just see it as a small cost in exchange for a large benefit.

That said, I have some vague ideas about a way to refactor this that might address your concerns, without abandoning the abstraction. Basically it would mean pushing the abstraction to the side, instead of using it as a middleman that obscures control flow. Initially it didn't seem possible, but maybe it's doable. I'll mess with the code a bit and see if I can make it work.

build.zig.zon
```diff
@@ -31,5 +31,10 @@
             .url = "https://github.com/jiacai2050/zig-curl/archive/8a3f45798a80a5de4c11c6fa44dab8785c421d27.tar.gz",
             .hash = "1220f70ac854b59315a8512861e039648d677feb4f9677bd873d6b9b7074a5906485",
         },
+        .rocksdb = .{
+            .url = "https://github.com/Syndica/rocksdb-zig/archive/87fbb125b71bf9dd230dbe600f31876282517e40.tar.gz",
+            .hash = "12205ed01a05d21837321e385ab9584e9813b017853cab7df93fdad9cdd662a6b4e5",
+            // .path = "../rocksdb-zig",
```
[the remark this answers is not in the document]
dnut: 8b17a9dc2ec74ef6c62c25f2174bc5aecec408a5

src/utils/collections.zig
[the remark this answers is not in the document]
dnut: e17030db295a9ccb6661df2701f84aaac27c9d52

src/shred_collector/shred.zig
```diff
@@ -128,10 +164,10 @@ pub const DataShred = struct {
     const Self = @This();
     const consts = data_shred;
 
-    fn sanitize(self: *const Self) !void {
+    pub fn sanitize(self: *const Self) !void {
```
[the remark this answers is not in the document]
dnut: > looking back at some prev code - the GenericShred approach seems a bit confusing - it would probs be a lot easier to understand the way agave did it [here](https://github.com/anza-xyz/agave/blob/b0df3945e8c9a2f85093fb720d349b536040a8d4/ledger/src/shred/legacy.rs#L36) by putting the common_fields  onto the struct itself

I don't like that there are so many layers either. I've been trying to come up with a better alternative but haven't come up with anything yet. In my opinion, the agave approach is even more confusing.

The issue with putting the fields directly in CodingShred and DataShred is that you need to copy identical duplicates for all the generic methods into both of those structs.

In agave, they accomplished this with macros, like `impl_shred_common` and `impl_merkle_shred`. Personally I find it easier to make sense of a nested struct rather than wading through macro definitions.

The other approach is just having duplicate code. One disadvantage to this is that you lose the idea that these methods apply to all shreds. Another thing to keep in mind is that we are going to need a lot more of them. The insert shreds pr doubles the number of generic methods on shreds. Maintaining duplicates of each is extra work and will almost inevitably lead to accidental divergence at some point.

I can spend some time looking for an approach that flattens some nesting without resorting to confusing metaprogramming or excessive code duplication. Some repetition may be inevitable, but I'd like to avoid duplicating every single detail. I think it will make more sense to do this on top of the changes in the insert shreds pr, first of all because GenericShred isn't introduced by the current pr, and also the insert shreds pr has more concrete methods that are very relevant to this. I'd like to ensure any change to the structure here is compatible with all of those methods, and avoid too many complicated merges.

src/blockstore/hashmap_db.zig
```diff
@@ -0,0 +1,258 @@
...
+
+    pub fn open(
+        allocator: Allocator,
+        _: Logger,
+        _: []const u8,
+        column_families: []const ColumnFamily,
+    ) !SharedHashMapDB {
+        var maps = try allocator.alloc(SharedHashMap, column_families.len);
+        errdefer allocator.free(maps);
+        errdefer for (maps) |*m| m.deinit();
```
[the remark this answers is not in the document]
dnut: 2cb69e9c1b0c22856525d860310729e8f6944225

src/blockstore/transaction_status.zig
dnut: 4d870529c9c26d3dd546c97032bd7f6dde057bb4

src/blockstore/meta.zig
```diff
@@ -0,0 +1,346 @@
...
+    //
+    // CONNECTED is explicitly the first bit to ensure backwards compatibility
+    // with the boolean field that ConnectedFlags replaced in SlotMeta.
+    connected = 0b0000_0001,
+    // PARENT_CONNECTED IS INTENTIIONALLY UNUSED FOR NOW
+    parent_connected = 0b1000_0000,
+});
+
+pub const DuplicateSlotProof = struct {
+    // #[serde(with = "serde_bytes")]
```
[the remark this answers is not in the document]
dnut: 4d870529c9c26d3dd546c97032bd7f6dde057bb4

src/blockstore/rocksdb.zig
dnut: I thought the custom serializers were necessary because agave uses different serializers. Agave uses traits for this, but it's basically the same idea.

But actually bincode is able to do most of it just fine, so I switched to using the bincode serializer for almost everything. The only exceptions are shreds, which need to pass through untouched. There are also some cases where the endianness is big instead of little, but maybe I don't need to worry about that since we're not targeting ledger binary compatibility with agave.

I'll see if I can come up with a more concrete solution for serializers.

src/blockstore/hashmap_db.zig
```diff
@@ -0,0 +1,258 @@
...
+        try self.map.put(self.allocator, key, value);
+    }
+
+    /// Only call this while holding the lock
+    pub fn getPreLocked(self: *Self, key: []const u8) ?[]const u8 {
+        return self.map.get(key) orelse return null;
+    }
+
+    pub fn delete(self: *Self, liberator: Allocator, key_: []const u8) void {
+        const key, const value = lock: {
```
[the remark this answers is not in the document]
dnut: I used a different label here because I wanted to be very clear that the sole purpose of this block is to define the scope of the lock. No code should be put in here unless it needs to run while the lock is held. Is that a good enough reason for a non-standard name? Can you think of a better name for that purpose?

src/blockstore/hashmap_db.zig
```diff
@@ -0,0 +1,258 @@
...
+        errdefer allocator.free(maps);
+        errdefer for (maps) |*m| m.deinit();
+        inline for (0..column_families.len) |i| {
+            maps[i] = try SharedHashMap.init(allocator);
+        }
+        return .{ .allocator = allocator, .maps = maps };
+    }
+
+    pub fn deinit(self_: Self) void {
+        var self = self_;
```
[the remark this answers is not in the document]
dnut: It might have been for compatibility with Database.deinit. I'm going to mess with that abstraction a bit and see if I can remove the need for weirdness like this.

src/utils/types.zig
```diff
@@ -33,6 +33,15 @@ pub fn ParamsTuple(comptime function: anytype) type {
     } });
 }
 
+/// Gets the return type of a function or function pointer
+pub fn Return(comptime FnPtr: type) type {
```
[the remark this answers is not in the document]
dnut: a84b24c0c52950229278f41f324ff73d413f749d

src/shred_collector/shred.zig
```diff
@@ -128,10 +164,10 @@ pub const DataShred = struct {
...
         try self.fields.sanitize();
         const flags = self.fields.custom.flags;
-        if (flags.intersects(.last_shred_in_slot) and
+        if (flags.intersects(.last_shred_in_slot) and // TODO redundant
```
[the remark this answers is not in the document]
dnut: 4f178921d26f02c4584818fef9c1cae841f82486

src/utils/types.zig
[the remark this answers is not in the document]
0xNineteen: theres still a few `const Return = sig.utils.types.ReturnType;`

src/shred_collector/shred.zig
```diff
@@ -165,12 +200,25 @@ pub const DataShred = struct {
...
 };
 
 /// Analogous to [Shred trait](https://github.com/anza-xyz/agave/blob/8c5a33a81a0504fd25d0465bed35d153ff84819f/ledger/src/shred/traits.rs#L6)
 pub fn GenericShred(
     comptime CustomHeader: type,
-    constants: ShredConstants,
+    constants_: ShredConstants,
 ) type {
     return struct {
         common: ShredCommonHeader,
```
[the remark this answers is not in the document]
dnut: c24c3a432a0cfa03eccb12e5d1a4a33cec280bc6

src/utils/types.zig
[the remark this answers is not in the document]
dnut: 7febee9fd0a7dbea728df2708107f4d2119db7bb

src/shred_collector/shred.zig
```diff
@@ -317,6 +401,15 @@ fn proofOffset(constants: ShredConstants, variant: ShredVariant) !usize {
         if (variant.chained) SIZE_OF_MERKLE_ROOT else 0;
 }
 
+/// agave: get_chained_merkle_root_offset
```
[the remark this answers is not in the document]
dnut: abcd0108d204ced445abef0594a49af6dd8e796f

src/shred_collector/shred.zig
```diff
@@ -85,12 +114,19 @@ pub const CodingShred = struct {
     const Self = @This();
     const consts = coding_shred;
 
-    fn sanitize(self: *const Self) error{InvalidNumCodingShreds}!void {
+    pub fn firstCodingIndex(self: *const Self) ?u32 {
```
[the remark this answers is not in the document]
dnut: Is this what you're looking for? 9be5d821a6487c20131c09f2bc8854d732f5ccfd

src/shred_collector/shred.zig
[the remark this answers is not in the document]
dnut: I created a ticket to track this concern, where we can continue the discussion.

https://github.com/Syndica/sig/issues/212

src/blockstore/database.zig
```diff
@@ -0,0 +1,248 @@
...
+
+        pub fn initBatch(self: *Self) !Batch {
+            return .{ .impl = self.impl.initBatch() };
+        }
+
+        pub fn commit(self: *Self, batch: Batch) !void {
+            return self.impl.commit(batch.impl);
+        }
+
+        pub fn runTest() !void {
```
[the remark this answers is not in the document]
dnut: I've done some refactoring in response to your other comments and as a part of that I moved this out of the struct into a `tests` namespace which can hold all database tests (I intend to add more). 

https://github.com/Syndica/sig/blob/9be5d821a6487c20131c09f2bc8854d732f5ccfd/src/blockstore/database.zig#L163-L169

And all tests can be run by calling this function in a zig `test {}`:

https://github.com/Syndica/sig/blob/9be5d821a6487c20131c09f2bc8854d732f5ccfd/src/blockstore/database.zig#L13-L19

src/blockstore/database.zig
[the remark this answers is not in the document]
dnut: I have refactored the code to address what I believe are your primary concerns:
1. Inconsistent naming and excessive type aliasing makes it confusing to use write batches
2. It would be easier to follow the blockstore control flow if the blockstore could directly use RocksDB instead of needing to go through an abstract Database struct as an intermediary.

1  was straightforward. I moved the batch struct definitions into the database struct definitions, instead of using a type alias. Also I consolidated everything onto the name "WriteBatch" so there is no more `Batch` `BatchImpl` etc.

For 2, I'm actually not sure if I've properly understood your concern. But anyway I came up with an approach here so that the RocksDB struct can be directly used as the database, without needing to declare `Database` as the actual database implementation. The key change is that we have this code now:

https://github.com/Syndica/sig/blob/9be5d821a6487c20131c09f2bc8854d732f5ccfd/src/blockstore/blockstore.zig#L3

instead of this:

```zig
pub fn BlockstoreDB(comptime DB: type) type {
    return sig.blockstore.database
        .Database(DB, &sig.blockstore.schema.list);
}
```

Database itself still exists, and can be used as a RocksDB wrapper. But now it's only used for tests. 

I'm actually not convinced this approach is better. I prefer to depend on abstractions rather than concrete implementations in cases like this. But I'm OK with it because I can see the advantage, and I found a way to retain the most valuable aspects of the abstraction.

Basically the change I made was to ensure that Database, RocksDB, and SharedHashMapDB all have truly identical interfaces. Also I made it so each database impl stands alone, as in, I stripped out whatever minimal logic was in Database, and moved it to the impls (such as finding the cf index). That way you can actually swap out each of these structs for one another. Any of them could work as the blockstore backend.

I also implemented some validations to guarantee that the structs are (and will remain) fully interchangeable. You can see that code here:
https://github.com/Syndica/sig/blob/9be5d821a6487c20131c09f2bc8854d732f5ccfd/src/utils/interface.zig#L3

It's used here. And this function is called when running the database tests for a particular implementation.
https://github.com/Syndica/sig/blob/9be5d821a6487c20131c09f2bc8854d732f5ccfd/src/blockstore/database.zig#L8-L12

src/blockstore/hashmap_db.zig
```diff
@@ -0,0 +1,258 @@
...
+    pub fn deinit(self: *Self) void {
+        var iter = self.map.iterator();
+        while (iter.next()) |entry| {
+            self.allocator.free(entry.key_ptr.*);
+            self.allocator.free(entry.value_ptr.*);
+        }
+        self.map.deinit(self.allocator);
+    }
+
+    pub fn free(self: Self, bytes: []const u8) void {
```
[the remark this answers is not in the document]
dnut: I have since removed this in one of my larger commits while addressing other comments.

src/blockstore/hashmap_db.zig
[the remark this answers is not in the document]
dnut: This has been addressed:

https://github.com/Syndica/sig/blob/9be5d821a6487c20131c09f2bc8854d732f5ccfd/src/blockstore/hashmap_db.zig#L200

src/blockstore/database.zig
```diff
@@ -0,0 +1,202 @@
...
+            .data = item,
+        } else .{
+            .allocator = allocator,
+            .data = serializeAlloc(allocator, item),
+        };
+    }
+
+    /// Returned data is owned by the caller. Free with `allocator.free`.
+    pub fn deserialize(comptime T: type, allocator: Allocator, bytes: []const u8) !T {
+        return try sig.bincode.readFromSlice(allocator, T, bytes, .{});
```
0xNineteen, line 150: should just call `bincode.readFromSlice` instead of this wrapper function

src/blockstore/database.zig
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    /// Returned data may or may not be owned by the caller.
+    /// Do both:
+    ///  - Assume the data is owned by the scope where `item` originated,
+    ///    so finish using the slice before returning from the caller (do not store slice as-is)
+    ///  - Call BytesRef.deinit before returning from the caller (as if you own it).
+    ///
+    /// Use this if the database backend accepts a pointer and immediately calls memcpy.
+    pub fn serializeToRef(allocator: Allocator, item: anytype) !BytesRef {
```
0xNineteen, line 138: this can probably be a bincode helper function - could see it being useful elsewhere too

src/blockstore/rocksdb.zig
```diff
@@ -0,0 +1,193 @@
...
+                .inner = rocks.WriteBatch.init(),
+                .cf_handles = self.cf_handles,
+            };
+        }
+
+        pub fn commit(self: *Self, batch: WriteBatch) Error!void {
+            return callRocks(self.logger, self.db.write, .{batch.inner});
+        }
+
+        pub const WriteBatch = struct {
```
0xNineteen, line 140: can you add a comment explaining more context on why we have this WriteBatch interface -- the code seems to be the exact same as the outer struct (`WriteBatch.put` and `RocksDB.put`)

src/blockstore/hashmap_db.zig
```diff
@@ -0,0 +1,237 @@
...
+        }
+
+        pub fn put(
+            self: *Self,
+            comptime cf: ColumnFamily,
+            key: cf.Key,
+            value: cf.Value,
+        ) Allocator.Error!void {
+            const key_bytes = try serializeAlloc(self.allocator, key);
+            errdefer self.allocator.free(key_bytes);
```
0xNineteen, line 54: can we add some spaces between after the `errdefer`s ? hard to read whats going on

src/blockstore/rocksdb.zig
```diff
@@ -0,0 +1,193 @@
...
+            const key_bytes = try serializeToRef(self.allocator, key);
+            defer key_bytes.deinit();
+            return try callRocks(
+                self.logger,
+                rocks.DB.delete,
+                .{ &self.db, self.cf_handles[cf.find(column_families)], key_bytes.data },
+            );
+        }
+
+        pub fn writeBatch(self: *Self) Error!WriteBatch {
```
0xNineteen, line 129: rename to `initWriteBatch` ?

src/blockstore/hashmap_db.zig
```diff
@@ -0,0 +1,237 @@
...
+    pub fn getPreLocked(self: *Self, key: []const u8) ?[]const u8 {
+        return self.map.get(key) orelse return null;
+    }
+
+    pub fn delete(self: *Self, liberator: Allocator, key_: []const u8) void {
+        const key, const value = lock: {
+            self.lock.lock();
+            defer self.lock.unlock();
+            const entry = self.map.getEntry(key_) orelse return;
+            const key = entry.key_ptr.*;
```
0xNineteen, line 225: why do we need this? - cant we just return `key_` ? -- would also help with naming since we could remove the suffix _

src/blockstore/meta.zig
```diff
@@ -0,0 +1,344 @@
...
+
+const BitFlags = sig.utils.bitflags.BitFlags;
+const Shred = sig.shred_collector.shred.Shred;
+const CodingShred = sig.shred_collector.shred.CodingShred;
+const Slot = sig.core.Slot;
+const SortedSet = sig.utils.collections.SortedSet;
+
+/// The Meta column family
+pub const SlotMeta = struct {
+    // allocator: Allocator,
```
0xNineteen, line 14: should delete the allocator mentions that are commented out if we arent using them (are throughout the file too)

src/blockstore/meta.zig
```diff
@@ -0,0 +1,344 @@
...
+            .next_slots = std.ArrayList(Slot).init(allocator),
+            .completed_data_indexes = SortedSet(u32).init(allocator),
+        };
+    }
+
+    pub fn deinit(self: Self) void {
+        self.next_slots.deinit();
+    }
+
+    // TODO copy on write
```
0xNineteen, line 70: this TODO still valid?

src/blockstore/schema.zig
```diff
@@ -0,0 +1,132 @@
...
+};
+
+pub const list: [decls.len]ColumnFamily = l: {
+    var ret: [decls.len]ColumnFamily = undefined;
+    for (decls, 0..) |decl, i| {
+        ret[i] = @field(schema, decl.name);
+    }
+    break :l ret;
+};
+const decls = @typeInfo(schema).Struct.decls;
```
0xNineteen, line 132: can we move `decls` to be be declared above list?

src/shred_collector/shred.zig
```diff
@@ -236,9 +292,34 @@ pub fn GenericShred(
...
+            return signature.verify(signer, &signed_data.data);
+        }
+
+        fn signedData(self: Self) !Hash {
+            return getMerkleRoot(self.payload, constants, self.common.shred_variant);
+        }
+
+        pub fn merkleRoot(self: Self) !Hash {
+            return getMerkleRoot(self.payload, constants, self.common.shred_variant);
+        }
```
0xNineteen, line 308: why do we have two methods that do the same thing?

src/shred_collector/shred.zig
```diff
@@ -71,26 +76,58 @@ pub const Shred = union(ShredType) {
...
+/// Analogous to [ShredCode](https://github.com/anza-xyz/agave/blob/7a9317fe25621c211fe4ab5491b88a4757d4b6d4/ledger/src/shred/merkle.rs#L74)
 pub const CodingShred = struct {
     fields: Fields,
     const Fields = GenericShred(CodingShredHeader, coding_shred);
 
     const Self = @This();
     const consts = coding_shred;
 
-    fn sanitize(self: *const Self) error{InvalidNumCodingShreds}!void {
+    pub fn firstCodingIndex(self: *const Self) ?u32 {
```
0xNineteen, line 118: a bit out of scope but theres a method `first_coding_index` which returns an error instead of optional - probs should choose one or the other and use a `catch` or and `orelse`

src/shred_collector/shred.zig
```diff
@@ -47,33 +47,62 @@ pub const Shred = union(ShredType) {
...
     }
 
-    pub fn fromPayload(allocator: Allocator, payload: []const u8) !Self {
-        const variant = layout.getShredVariant(payload) orelse return error.InvalidShredVariant;
+    pub fn fromPayload(allocator: Allocator, payload_: []const u8) !Self {
+        const variant = layout.getShredVariant(payload_) orelse return error.InvalidShredVariant;
         return switch (variant.shred_type) {
-            .Code => .{ .Code = .{ .fields = try CodingShred.Fields.fromPayload(allocator, payload) } },
-            .Data => .{ .Data = .{ .fields = try DataShred.Fields.fromPayload(allocator, payload) } },
+            .Code => .{ .Code = .{ .fields = try CodingShred.Fields.fromPayload(allocator, payload_) } },
```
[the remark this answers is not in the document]
0xNineteen: im saying its not clear that `fromPayload` would have sanitize logic from its name - one reason is because its similar to `bincode.fromSlice`, which has no sanitize logic

a method name which describes that it will read and sanitize would be better imo eg, `DeserializeAndSanitize` or even `DeserializeAndValidate` (not a big fan of `sanitize` - was more so just following agave's approach) or something along those lines

if you find the function names in bincode confusing, we can always change them to something more clear so we can get consistency across the codebase - not in this PR, but create a ticket with some other ideas and we can come up with something better

0xNineteen (changes requested):
few notes - would also also be nice to have a blockstore/readme.md with docs on the important info along with any contextual information but for the sake of time we can create a ticket, and tackle it another time

https://github.com/orgs/Syndica/projects/2/views/1?pane=issue&itemId=74031612

src/shred_collector/shred.zig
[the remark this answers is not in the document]
0xNineteen: for the sake of time - ill resolve this, create a ticket, and we can come back to it later

src/blockstore/meta.zig
dnut: f88f48dce43ee03831e536c2cc67e810451e25ff

src/blockstore/meta.zig
dnut: I think I put it there because it was relevant for a specific usage of clone being particularly wasteful. But I must have removed that usage.

cd8719524603b05de51a16d5f57f8a7d77a6b21e

src/shred_collector/shred.zig
dnut: d29e2fd0edccdaee741be8b3cfceca78468eb542

src/shred_collector/shred.zig
dnut: d9bd45a3b71e1178bb9ca9c8c395d22c8f17e38a

src/blockstore/hashmap_db.zig
dnut: 8a89d582b72043deaec934200e7150fd61c14086

src/blockstore/rocksdb.zig
dnut: 8b93fa92ed066a8cb70bc80745f33a2951cbfc3c

0xNineteen (approved):
:shipit:

PR #196, opened 2024-07-15
refactor(net,cmd): Add explicit error sets, refactor cli option variables

InKryption:
Moving the cli option variables into the `run` function caught the fact that the `number_of_index_bins_option` variable wasn't being exposed in the relevant subcommands.


0xNineteen (approved):
lgtm - great catch 🚀

PR #200, opened 2024-07-19
revert: remove .vscode folder and add to .gitignore

dnut:
I don't know if this was added on purpose or by accident. I don't think we should include files in the repo that interfere with local IDE configuration.


0xNineteen:
yeah agreed - would be nice to have these files still defined and referenced in the documentation somewhere for others to know how to setup sig + vscode before we remove them

0xNineteen:
fixed this in #201

PR #202, opened 2024-07-23
feat(gossip): improved prom metrics

0xNineteen:
- updated grafana dashboards to use histograms (see img below)
- updated gossip table stats (number of values + number of pubkeys) to use gauge
- cap gossip table from 1M -> 100_000 records (much more accurate to prod)
- add additional clean up metrics (pubkeys removed + old labels removed)
- change gossip to use `sig.time.Timer` and our new `Duration` 
- add additional channel metrics 

image

_note:_ in some cases we recv a lot of packets at once which bottlenecks the channel - and could eventually lead to OOM if we do nothing (inf channel growth is a problem)

image


src/gossip/service.zig
```diff
@@ -1877,22 +1934,32 @@ pub const GossipStats = struct {
...
-    handle_batch_pull_req_time: *Counter,
-    handle_batch_pull_resp_time: *Counter,
-    handle_batch_prune_time: *Counter,
-    handle_trim_table_time: *Counter,
-
     push_message_n_values: *Counter,
     push_message_n_failed_inserts: *Counter,
     push_message_n_invalid_values: *Counter,
+
+    // TODO(x19): these should be histograms
```
0xNineteen, line 1941: remove

InKryption (changes requested):
Just a couple details.

InKryption:
Also, something I just noticed while reviewing (but can't add in a review because it's too far away from the changes for github to let me review it):
the `sync.toConst` method can potentially return a pointer to a local variable, if the parameter isn't a pointer. This will cause illegal behavior.
It currently isn't used anywhere other than tests - would be good to either remove it now or fix it, potentially in this PR which seems on its way to being merged.

src/gossip/service.zig
```diff
@@ -58,43 +59,41 @@ const getWallclockMs = sig.gossip.data.getWallclockMs;
...
+pub const MAX_NUM_VALUES_PER_PULL_RESPONSE = 20; // TODO: this is approx the rust one -- should tune
+pub const NUM_ACTIVE_SET_ENTRIES: usize = 25;
 /// Maximum number of origin nodes that a PruneData may contain, such that the
 /// serialized size of the PruneMessage stays below PACKET_DATA_SIZE.
 pub const MAX_PRUNE_DATA_NODES: usize = 32;
-pub const NUM_ACTIVE_SET_ENTRIES: usize = 25;
+
+pub const PING_CACHE_CAPACITY: usize = 65_536;
+pub const PING_CACHE_TTL_NS: u64 = std.time.ns_per_s * 1280;
+pub const PING_CACHE_RATE_LIMIT_DELAY_NS: u64 = std.time.ns_per_s * (1280 / 64);
```
yewman, line 91: Is there a reason we are not using Duration here as well?

src/gossip/service.zig
```diff
@@ -657,116 +665,137 @@ pub const GossipService = struct {
...
-                self.stats.handle_batch_pong_time.add(elapsed);
+                const elapsed = x_timer.read().asMillis();
+                self.stats.handle_batch_pong_time.observe(@floatFromInt(elapsed));
 
                 pong_messages.clearRetainingCapacity();
             }
 
             // TRIM gossip-table
-            if (trim_table_timer.read().asMillis() > GOSSIP_TRIM_INTERVAL_MS) {
+            if (trim_table_timer.read().asMillis() > TABLE_TRIM_RATE.asMillis()) {
```
yewman, line 730: Minor preference for using Duration.gt(other) for comparison. The gt method compares the underlying u64 rather than converting both and then comparing

yewman (commented):
A couple of minor comments, otherwise lgtm

src/gossip/service.zig
```diff
@@ -1923,14 +1990,37 @@ pub const GossipStats = struct {
...
+        50,
+        100,
+        250,
+        500,
+        1000,
+        2500,
+        5000,
+        10000,
+    };
+
```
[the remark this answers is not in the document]
0xNineteen: not sure what you mean - do you have an example?

src/gossip/service.zig
0xNineteen: great point 

cc1ed30

src/gossip/service.zig
0xNineteen: mmm not sure whats the best way 

you bring up a good point on the conversion and comparison - i like the '>' operator because its extremely clear what it does - the `gt()` method is easy to understand but theres always the mental energy of reading a new method name and wondering if it does what you expect or something more (ptsd from other codebase methods doing random things)

will change to use `asNanos()` so theres no conversion - hopefully thats ok but curious on your thoughts too

bb22c36

src/gossip/table.zig
```diff
@@ -717,8 +715,8 @@ pub const GossipTable = struct {
     /// returns true if any values were removed.
     ///
     /// NOTE: the `now` parameter is used to populate the purged field with the timestamp of the removal.
-    pub fn attemptTrim(self: *Self, now: u64, max_pubkey_capacity: usize) error{OutOfMemory}!bool {
-        if (!self.shouldTrim(max_pubkey_capacity)) return false;
+    pub fn attemptTrim(self: *Self, now: u64, max_pubkey_capacity: usize) error{OutOfMemory}!u64 {
```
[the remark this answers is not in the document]
0xNineteen: 38bcfd4

0xNineteen:
> Also, something I just noticed while reviewing (but can't add in a review because it's too far away from the changes for github to let me review it): the `sync.toConst` method can potentially return a pointer to a local variable, if the parameter isn't a pointer. This will cause illegal behavior. It currently isn't used anywhere other than tests - would be good to either remove it now or fix it, potentially in this PR which seems on its way to being merged.

not sure what you mean - should be a quick approve in a separate pr if you want to tackle it or can add it to a ticket with an example solution

src/gossip/service.zig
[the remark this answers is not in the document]
InKryption: Zig fmt accepts this, for example:
```suggestion
    const HANDLE_TIME_BUCKETS_MS: [10]f64 = .{
        10,   25,
        50,   100,
        250,  500,
        1000, 2500,
        5000, 10000,
    };
```
could use 3, 4, or 5 columns - you could even essentially make it horizontal by making it have as many columns as it does elements.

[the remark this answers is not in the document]
0xNineteen: ah interesting

src/gossip/service.zig
yewman: Yeah, I agree with your preference for using the operators as it is clearer. I think using asNanos as you suggested is the best way to go

PR #209, opened 2024-07-25
feat(geyser): support account streaming

0xNineteen:
this is the initial implementation for a geyser-like plugin to support account streaming out of the validator 

see `src/geyser/readme.md` for description on how it works (and a link to an example geyser-reader)

benchmark results on a macbook pro
```
read mb/sec: 2011.97
```


InKryption (commented):
A couple basic comments on `RecycleFBA`.

src/utils/allocators.zig
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    /// creates a new file with size aligned to page_size and returns a pointer to it
+    pub fn alloc(ctx: *anyopaque, n: usize, log2_align: u8, return_address: usize) ?[*]u8 {
+        _ = log2_align;
+        _ = return_address;
+        const self: *Self = @ptrCast(@alignCast(ctx));
+
+        if (n > self.alloc_allocator.buffer.len) {
+            @panic("RecycleFBA.alloc: requested size too large, make the buffer larger");
```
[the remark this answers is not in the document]
0xNineteen: a common pattern with the RecycleFBA is to try to alloc, if it fails, wait for some amount of time and then try again (hoping something was freed): this is how we use it - if you try to alloc a slice greater than whats available using this approach, you will inf loop

src/utils/allocators.zig
```diff
@@ -0,0 +1,116 @@
...
+                .alloc = alloc,
+                .resize = resize,
+                .free = free,
+            },
+        };
+    }
+
+    /// creates a new file with size aligned to page_size and returns a pointer to it
+    pub fn alloc(ctx: *anyopaque, n: usize, log2_align: u8, return_address: usize) ?[*]u8 {
+        _ = log2_align;
```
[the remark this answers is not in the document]
0xNineteen: how would you respect it? tbh im not sure what that variable means

src/accountsdb/db.zig
```diff
@@ -422,17 +430,8 @@ pub const AccountsDB = struct {
...
-        }
-        counting_alloc_ptr.* = .{
-            .self_allocator = self.allocator,
-            .references = references,
-            .alloc_count = 0,
-        };
+        const counting_alloc = try FreeCounterAllocator.init(self.allocator, references);
+        defer counting_alloc.deinitIfSafe();
 
         const ref_timer = try std.time.Timer.start();
```
yewman, line 436: Could update this timer as well?

src/accountsdb/db.zig
```diff
@@ -1858,7 +1904,16 @@ pub const AccountsDB = struct {
...
             n_accounts,
         );
-        try self.account_index.validateAccountFile(account_file, bin_counts, &references);
+
+        try indexAndValidateAccountFile(
+            account_file,
+            self.account_index.pubkey_bin_calculator,
+            bin_counts,
+            &references,
+            // NOTE: this method should only be called in tests/benchmarks so we dont need
```
yewman, line 1913: Could you name the method which is only called in tests/benchmarks? Or perhaps add doc comment on the method which should only be called for tests/benchmarks

src/accountsdb/db.zig
```diff
@@ -2114,16 +2171,129 @@ pub const AccountsDB = struct {
...
+}
+
+/// allocator which counts the number of times free is called. when count
+/// reaches 0, it will deinit the full arraylist. useful for when you want
+/// to allocate a large Arraylist and split it across multiple different
+/// ArrayLists -- alloc and resize are not implemented.
+///
+/// see `loadAndVerifyAccountsFiles` for an example of how to use this allocator
+const FreeCounterAllocator = struct {
     /// optional heap allocator to deinit the ptr on deinit
```
yewman, line 2279: Optional? 

Also on a side note, I have a slight preference to using 
`pub fn intoAllocator(self: *Self) std.mem.Allocator`  
or something similar over
`pub fn allocator(self: *Self) std.mem.Allocator` 
and thus avoiding the naming conflict. 

Given that struct fields have no accessibility modifiers I think that if we have a naming conflict between a struct field and a method it is likely that the method name could be more informative, the method must be doing 'something' otherwidse we would just access the field directly?

yewman: Never mind, I see this is conforming to a pattern used by other allocators. Instead of `self_allocator` I think `backing_allocator` as used in RecycleFBA is better. On that note, should we move this allocator into `'utils/allocators.zig'`

yewman:
lgtm, some vary minor comments which can go unaddressed if time is of concern

src/utils/allocators.zig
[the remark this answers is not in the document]
InKryption: I would respect it using the `rawAlloc` function; could also do what that function would logically do and find a sub-buffer which is aligned enough manually, if necessary.

src/accountsdb/db.zig
0xNineteen: c99a691

src/accountsdb/db.zig
0xNineteen: f367f54

src/accountsdb/db.zig
0xNineteen: was originally thinking that too but the struct is so niche to accounts-db (eg, `references: ArrayList(AccountRef)`) i feel like its better here unless things change

src/utils/allocators.zig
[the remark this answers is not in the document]
0xNineteen: f1e14e4

src/utils/allocators.zig
```diff
@@ -0,0 +1,114 @@
...
+    backing_allocator: std.mem.Allocator,
+    // this does the data allocations (data is returned from alloc)
+    alloc_allocator: std.heap.FixedBufferAllocator,
+    // recycling depot
+    records: std.ArrayList(Record),
+
+    // for thread safety
+    mux: std.Thread.Mutex = .{},
+
+    const Record = struct { is_free: bool, buf: [*]u8, len: u64, log2_align: u8 };
```
InKryption, line 15: We don't need to store the alignment as a value, we can simply check whether the pointer is aligned with `std.mem.isAlignedLog2(@intFromPtr(ptr), log2_align)`.

0xNineteen: ooo cool 170c652

InKryption (approved):
LGTM

src/accountsdb/download.zig
```diff
@@ -201,7 +200,15 @@ pub fn downloadSnapshotsFromGossip(
...
-            logger.infof("searched for peers to downlod from: {any}", .{result});
+            var write_buf: [512]u8 = undefined;
+            var i: usize = 0;
+            inline for (@typeInfo(PeerSearchResult).Struct.fields) |field| {
+                if (@field(result, field.name) != 0) {
+                    const r = try std.fmt.bufPrint(write_buf[i..], "{s}: {d} ", .{ field.name, @field(result, field.name) });
+                    i += r.len;
+                }
+            }
+            logger.infof("searched for snapshot peers: {s}", .{write_buf[0..i]});
```
dnut, line 211: How is this different from `any`? Is it to exclude the type name and braces?

src/core/account.zig
```diff
@@ -1,6 +1,18 @@
...
+const sig = @import("../lib.zig");
+const bincode = sig.bincode;
+
+const Pubkey = sig.core.Pubkey;
+const Epoch = sig.core.Epoch;
+const AccountInFile = sig.accounts_db.accounts_file.AccountInFile;
+
+const U8SliceConfig = bincode.int.U8SliceConfig;
+
+pub const AccountWithoutData = struct {
```
dnut, line 10: I probably would have called this AccountMetadata, but we have this already and it's called AccountInfo. Actually this struct isn't even used so I would just delete it.

src/accountsdb/db.zig
```diff
@@ -2011,6 +2056,8 @@ pub const AccountsDB = struct {
         );
     }
 
+    /// index and validate an account file.
+    /// NOTE: should only be called in tests/benchmarks
```
dnut, line 2060: What do you think about adding an assertion?

src/geyser/core.zig
```diff
@@ -0,0 +1,628 @@
...
+    payload: VersionedAccountPayload,
+
+    const Self = @This();
+
+    pub fn deinit(self: *Self, allocator: std.mem.Allocator) void {
+        self.payload.deinit(allocator);
+    }
+};
+
+pub const VersionedAccountPayload = union(enum(u8)) {
```
dnut, line 36: Why does this need to be versioned? Do you expect to have multiple versions going over the same stream?

src/geyser/core.zig
```diff
@@ -0,0 +1,628 @@
...
+                std.debug.print("Failed to create pipe: {}\n", .{err});
+                return error.FailedToCreatePipe;
+            },
+        }
+    }
+
+    const file = try std.fs.cwd().openFile(pipe_path, .{ .mode = .read_write });
+
+    // set to nonblocking
+    // TODO(x19): use F_GETFL | NON_BLOCK
```
dnut, line 384: should this be addressed?

src/utils/allocators.zig
```diff
@@ -0,0 +1,114 @@
...
+    // this allocates the underlying memory + dynamic expansions
+    // (only used on init/deinit + arraylist expansion)
+    backing_allocator: std.mem.Allocator,
+    // this does the data allocations (data is returned from alloc)
+    alloc_allocator: std.heap.FixedBufferAllocator,
+    // recycling depot
+    records: std.ArrayList(Record),
+
+    // for thread safety
+    mux: std.Thread.Mutex = .{},
```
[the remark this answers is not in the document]
0xNineteen: in the scenario where if its safe to use without locking, they can use it without having an extra lock in the code

src/utils/allocators.zig
```diff
@@ -0,0 +1,114 @@
...
+            if (item.is_free and item.len >= n and std.mem.isAlignedLog2(@intFromPtr(item.buf), log2_align)) {
+                item.is_free = false;
+                return item.buf;
+            }
+        }
+
+        // otherwise, allocate a new one
+        const buf = self.alloc_allocator.allocator().rawAlloc(n, log2_align, return_address) orelse {
+            // std.debug.print("RecycleFBA alloc error: {}\n", .{ err });
+            return null;
```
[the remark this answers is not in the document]
0xNineteen: yeah it would get stuck, should we panic instead of returning null?

src/utils/allocators.zig
```diff
@@ -0,0 +1,114 @@
...
+
+        return buf;
+    }
+
+    pub fn free(ctx: *anyopaque, buf: []u8, log2_align: u8, return_address: usize) void {
+        _ = log2_align;
+        _ = return_address;
+        const self: *Self = @ptrCast(@alignCast(ctx));
+
+        for (self.records.items) |*item| {
```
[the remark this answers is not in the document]
0xNineteen: great idea - will leave for future work since its more of a nice to have/perf improvement - the disk_based_allocator could also use this improvement too

43ca99a

src/utils/allocators.zig
```diff
@@ -0,0 +1,114 @@
...
+        _ = log2_align;
+        _ = return_address;
+
+        const self: *Self = @ptrCast(@alignCast(ctx));
+        for (self.records.items) |*item| {
+            if (item.buf == buf.ptr) {
+                if (item.len >= new_size) {
+                    return true;
+                } else {
+                    return false;
```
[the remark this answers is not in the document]
0xNineteen: d6d965f

src/accountsdb/db.zig
0xNineteen: i dont think we have a way to tell if code is in a benchmark rn?

src/geyser/core.zig
0xNineteen: rn it doesnt need to be versioned but in the future when we roll out improvements / other payload types, this allows us to be backwards compatible

src/geyser/core.zig
0xNineteen: its fine for now since we dont expect any preset flags on the pipe - the idea in the TODO was to use `F_GETFL` to get the current fcn flags and then OR in the NON_BLOCK flag so we dont reset any preset flags

src/bincode/int.zig
```diff
@@ -30,3 +30,65 @@ pub fn defaultOnEof(comptime T: type, comptime eof_value: T) bincode.FieldConfig
         .serializer = S.serializer,
     };
 }
+
+pub fn U8SliceConfig() bincode.FieldConfig([]u8) {
```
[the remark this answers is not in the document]
0xNineteen: f411ef8

src/accountsdb/db.zig
InKryption: A simple (hacky) way to tell that would be to `comptime assert(@import("root") == @import("../benchmarks.zig"));`.

src/geyser/core.zig
dnut: What I mean is, when you do an upgrade, why is this approach to versioning necessary? I would only expect it is necessary when multiple versions need to coexist in memory simultaneously. Since geyser streams are ephemeral, not persistent, seems that you can just make the upgrade by switching to the new type and forgetting about the old one.

The advantage I see to versioning is that it presents a clear signal to readers when they have the wrong version and they need to upgrade. But that could be implemented with just an int field within a struct, doesn't need to be an enum.

src/utils/allocators.zig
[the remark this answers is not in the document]
dnut: I think null is correct here. You'd have to add more logic to detect that you're in a degraded state. Even at that point, I don't think a panic is what you want. I'm just wondering if there's a different approach that could avoid this issue altogether. But it's probably not necessary to figure out at this point

src/utils/allocators.zig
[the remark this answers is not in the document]
dnut: It just feels like it breaks the allocator abstraction. Normally you can make some assumptions about an allocator, but you can't do that here. There is this extra special knowledge you need to have to ensure you're using it correctly. It feels like it could be easy to misuse. It's not a big deal in this case because you're using it correctly in the one case it was designed for. But I feel like it would be cleaner to integrate this into the allocator somehow. For example you could have some way to optionally specify whether or not you want a thread safe allocator when you instantiate it. The thread safe one would always call the locks internally, and the non-thread safe one wouldn't.

src/utils/allocators.zig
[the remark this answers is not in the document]
dnut: It's weird though for this code to panic based on some assumption about how some far away code is going to use an allocator. In the context that you're writing code that might become an infinite loop, you already have the special knowledge about the allocator. That's why the loop is there. So I think it makes sense to check for this case in the loop.

src/utils/allocators.zig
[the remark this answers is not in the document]
0xNineteen: yeah thats a great idea - will add a comment to add a config struct which specifies thread-safe version or not

src/accountsdb/db.zig
0xNineteen: gonna leave it for now with an open ticket: https://github.com/orgs/Syndica/projects/2/views/1?pane=issue&itemId=75090925

src/geyser/core.zig
0xNineteen: good point - yeah enums are not neccesary in this case and the same thing could be accomplished by ints - its just the simplest approach and what agave does rn

though the memory layout of enums will be wasteful since you need to alloc the largest struct, bincode serialization will basically be what you talk about (a single int to define which enum)

im not really hard stuck on either approach, but in terms of dev time required to add a new version, it would probably be easier to add a new enum

src/utils/allocators.zig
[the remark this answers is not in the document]
0xNineteen: yeah probably good enough for now and when we come across a situation when it breaks we can figure it out

thinking over it more i changed it to panic because if it ever hits that situation (which should never happen) it will get stuck and will be a massive pain to debug whats going wrong

dnut (approved):

src/utils/allocators.zig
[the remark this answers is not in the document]
0xNineteen: the problem i see is assuming special knowledge about the allocator is dangerous and easy to miss/forget about, being such a killer-case (will hard stuck and be difficult to debug) might as well be panic and give a clear message to the dev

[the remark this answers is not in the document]
dnut: > assuming special knowledge about the allocator is dangerous

Yeah, that's my point.

There is a loop that would normally be expected to have a high risk of running infinitely. To mitigate that risk, the loop relies on an assumption that the unique properties of this particular allocator will break out of the loop.

At that point, the loop no longer depends on just any old `Allocator`. It depends on this specific allocator. So it's only thinly disguised behind the `Allocator` interface. Plus, the allocator itself is written in a way that is specifically designed to deal with this particular looping pattern. Each piece of code is tightly coupled to specific details about how the other piece of code is written. The abstraction is broken in both directions and it might be better to implement it without the `Allocator` interface at all, so the special behavior that's already required is made more obvious.

I think it would be better for the loop itself to be written to be safe, rather than having an inherently dangerous loop that can only be saved by an allocator that's written with this particular loop in mind.

PR #211, opened 2024-07-26
fix(leader-schedule): Fix random number generator to match rust implementation and filter vote accounts with zero stake

yewman:


src/rand/rand.zig
```diff
@@ -114,6 +114,66 @@ pub fn BlockRng(
...
+    while (true) {
+        const v = r.int(T);
+        const m = std.math.mulWide(T, v, less_than);
+        const lo: T = @truncate(m);
+        if (lo <= zone) {
+            return @truncate(m >> bits);
+        }
+    }
+
+    unreachable;
```
InKryption, line 137: I don't think this `unreachable` is necessary, the compiler can already prove that a `break`less `while (const_condition)` never reaches this.

src/rand/rand.zig
```diff
@@ -114,6 +114,66 @@ pub fn BlockRng(
...
+        150245030477280661, 213624595435805867, 98539226549693351,  33801042929801430,  203768025145824241, 85942546992148859,  75920971919241723,  178282653246684682, 107180956559443505, 5383407800161210,   25924734343585423,  113505140769538087, 158294491487555208, 202058441263127158, 23754109538483632,  164332741174520337, 177386067656078943, 63092035208918209,  105193165892558811, 90842259062651642,  200087190942123447, 118864219926511175, 178914468731322740, 123929446491278238, 83233452188148132,  130678632301769656, 148140195893539773, 189717099709818244, 3866162913691558,   165276931946696165, 195669972620483101, 220740178411808744, 218597519430516114, 181499314628684163, 56281591694254078,  112584340323866656, 73160071773275256,  216001312005955998, 184061143539899962, 132782144182150573, 40566523540155677,  171562338978082216, 151758301674069632, 173398251098334249, 63757156897079302,  81032160488073578,  173298561206951917, 154421360720272045, 15981567533895039,  88393403508710277,
+    };
```
InKryption, line 164: Might be a good idea to move this to its own file, seems a bit out of place in the middle of everything. Ie something like `@import("test_data.zig").rust_chacha_uint_less_than_sequence`.
Rationale:
 * If it's horizontally longer, it's quicker to scroll past, but it causes the editor side scroll to be huge, which is awkward.
 * If it's vertically longer, it takes a long time to move past and breaks the flow of code.
 * Always best to keep test data separate from tests if it's meaningful and specific like this.

Another option would be to generate a file for this in build.zig that could then be obtained with ie `@embedFile("rust_chacha_uint_less_than_sequence")` - I leave that decision to you, though feel free to reach out if you want to do that but are unsure how.

src/rand/rand.zig
```diff
@@ -114,6 +114,66 @@ pub fn BlockRng(
...
+
+    var seed: [32]u8 = .{0} ** 32;
+    std.mem.writeInt(u64, seed[0..8], epoch, .little);
+    var rng = ChaChaRng(20).fromSeed(seed);
+    const random = rng.random();
+
+    for (rust_sequence) |expected| {
+        const actual = uintLessThanRust(random, u64, max);
+        try std.testing.expectEqual(expected, actual);
+    }
```
InKryption, line 174: :pray: thank you for using the specific `expect*` function. But one thing that might be desirable here would be to compare in bulk using `std.testing.expectEqualSlices`:
```suggestion
    var actual_sequence: [rust_sequence.len]u8 = undefined;
    for (&actual_sequence) |*actual| actual.* = uintLessThanRust(random, u64, max);
    try std.testing.expectEqualSlices(u64, &rust_sequence, &actual);
```
This way, if it were to fail, it would indicate both the mismatched values and their index.

InKryption (commented):
Good stuff, just a couple of code hygiene comments.

0xNineteen (approved):

0xNineteen (commented):
lgtm after addressing inks comments

src/rand/rand.zig
yewman: I got a compile error without it, thats why I chucked it in their, its not very nice though I agree

yewman: [060375](https://github.com/Syndica/sig/pull/211/commits/060375c68e4b0d08b9f6b52c6a2047f9ae00765c)

src/rand/rand.zig
yewman: I agree, for now I have just moved it to the base of the file and adjusted width. I think this is something that we should come up with a plan of attack for the project as a whole and then be consistent.

[060375](https://github.com/Syndica/sig/pull/211/commits/060375c68e4b0d08b9f6b52c6a2047f9ae00765c)

src/rand/rand.zig
yewman: [060375](https://github.com/Syndica/sig/pull/211/commits/060375c68e4b0d08b9f6b52c6a2047f9ae00765c)

InKryption (approved):
LGTM

PR #214, opened 2024-08-07
feat(accountsdb): incremental snapshot generation

InKryption:
Incremental snapshots as a whole still aren't ready for the manager loop, but this does set up the basic infrastructure required to flesh them out later, giving us a working base for generating incremental snapshots.

This also includes improvements to full snapshot generation (both in service of incremental snapshot generation and in general).

At the time of writing, I still need to check that it hasn't broken fuzz testing with full snapshots on; this removes the explicit fuzz testing in the fuzz loop, but it does generate full snapshots in the manager loop.

### DIFF
--- a/src/accountsdb/db.zig
+++ b/src/accountsdb/db.zig
@@ -1096,7 +1096,7 @@ pub const AccountsDB = struct {
         // TODO: get rid of this once `makeFullSnapshotGenerationPackage` can actually
         // derive this data correctly by itself.
         var rand = std.Random.DefaultPrng.init(1234);
-        const random_bank_fields = try BankFields.random(self.allocator, rand.random(), 128);
+        var random_bank_fields = try BankFields.random(self.allocator, rand.random(), 128);
         defer random_bank_fields.deinit(self.allocator);
 
         while (!exit.load(.monotonic)) {
@@ -1139,7 +1139,7 @@ pub const AccountsDB = struct {
             const must_flush_slots = flush_slots.items.len > 0;
 
             if (must_flush_slots) {
-                self.logger.debugf("flushing slots: {any}", .{flush_slots.items});
+                // self.logger.debugf("flushing slots: {any}", .{flush_slots.items});
                 defer {
                     flush_slots.clearRetainingCapacity();
                     unclean_account_files.clearRetainingCapacity();
@@ -1170,7 +1170,7 @@ pub const AccountsDB = struct {
             if (largest_flushed_slot - latest_full_snapshot_slot >= slots_per_full_snapshot) {
                 var snapshot_gen_pkg, const snapshot_gen_info = try self.makeFullSnapshotGenerationPackage(
                     largest_flushed_slot,
-                    random_bank_fields,
+                    &random_bank_fields,
                     rand.random().int(u64),
                     0,
                 );
@@ -2231,7 +2231,7 @@ pub const AccountsDB = struct {
         /// Must exist explicitly in the database.
         target_slot: Slot,
         /// Temporary: See above TODO
-        bank_fields: BankFields,
+        bank_fields: *BankFields,
         lamports_per_signature: u64,
         /// For tests against older snapshots. Should just be 0 during normal operation.
         deprecated_stored_meta_write_version: u64,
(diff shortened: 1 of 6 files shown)


src/accountsdb/db.zig
```diff
@@ -2011,68 +2164,367 @@ pub const AccountsDB = struct {
...
+            .IncrementalAccountHash = .{
+                .min_slot = full_snapshot_info.slot,
+                .max_slot = max_flushed_slot,
+            },
+        });
 
-                .stored_meta_write_version = stored_meta_write_version,
+        // TODO: figure out what the delta hash actually means - this happens to work in some cases,
+        // but sometimes fails with it being an empty range.
+        const delta_hash, _ = self.computeAccountHashesAndLamports(.{
```
0xNineteen, line 2485: can remove this too

src/accountsdb/db.zig
```diff
@@ -2254,35 +2708,121 @@ fn testWriteSnapshotFull(
...
+    };
+    defer archive_file.close();
+
+    var actual_snapshot_dir = try tmp_dir.makeOpenPath("output", .{ .iterate = true });
+    defer actual_snapshot_dir.close();
+
+    try archive_file.seekTo(0);
+    try std.tar.pipeToFileSystem(actual_snapshot_dir, archive_file.reader(), .{});
+
+    {
```
0xNineteen, line 2756: we should probably also include validating from the snapshot in the unit tests too

src/accountsdb/db.zig
```diff
@@ -1072,24 +1148,67 @@ pub const AccountsDB = struct {
...
                 for (flush_slots.items) |flush_slot| {
-                    self.flushSlot(flush_slot, &unclean_account_files) catch |err| {
+                    const unclean_file_id = self.flushSlot(flush_slot) catch |err| {
                         // flush fail = loss of account data on slot -- should never happen
                         self.logger.errf("flushing slot {d} error: {s}", .{ flush_slot, @errorName(err) });
+                        continue;
                     };
+                    unclean_account_files.appendAssumeCapacity(unclean_file_id);
                 }
+                self.largest_flushed_slot.store(root_slot, .seq_cst);
```
0xNineteen, line 1160: probs safter to store the max slot value in `flush_slots.items`

src/accountsdb/db.zig
```diff
@@ -1072,24 +1148,67 @@ pub const AccountsDB = struct {
...
+                const archive_file_name = archive_file_name_bounded.constSlice();
+                const archive_file = try self.snapshot_dir.createFile(archive_file_name, .{ .read = true });
+                defer archive_file.close();
+
+                const zstd_write_ctx = zstd.writerCtx(archive_file.writer(), &zstd_compressor, zstd_buffer);
+
+                try snapshot_gen_pkg.write(zstd_write_ctx.writer());
+                try zstd_write_ctx.finish();
+
+                try self.commitFullSnapshotInfo(snapshot_gen_info, .delete_old);
```
0xNineteen, line 1193: would you want to put this logic into a method? seems its used both here and in tests

src/accountsdb/db.zig
InKryption: I'm not sure I understand what you mean? Could you elaborate?

src/accountsdb/db.zig
```diff
@@ -2510,8 +2510,6 @@ pub const AccountsDB = struct {
         errdefer file_map_rw.unlock();
         const file_map = file_map_rw.get();
 
-        const max_flushed_slot = self.largest_flushed_slot.load(.monotonic);
```
InKryption: Good catch, unfortunate copy+paste followed by code desync error.

src/accountsdb/db.zig
0xNineteen: for (flush_slots) |slot| { max_flushed_slot = max(slot, max_flushed_slot) }
self.largest_flushed_slot = max_flushed_slot

InKryption: dacb0ec60e0b84d06474dc121019f053807858ee

src/accountsdb/db.zig
InKryption: The requirements of the code in the unit test and here are subtly different; in the unit test, we retain the archive file, whilst here, we want to close it as soon as we're finished writing the snapshot.

src/accountsdb/db.zig
InKryption: we'll have to leave this as future work for when we revive the unit test.

InKryption: 68730af34847e389077fa52359b4a4b3f6b0f34f

0xNineteen (approved):
lgtm :shipit:

PR #217, opened 2024-08-12
feat(blockstore): read and write methods

dnut:


src/blockstore/reader.zig
```diff
@@ -0,0 +1,2037 @@
...
+            while (iter.next()) |entry| {
+                entry.value_ptr.deinit();
+            }
+            map.deinit();
+        }
+        for (slots) |slot| {
+            if (try self.db.get(schema.slot_meta, slot)) |meta| {
+                errdefer meta.next_slots.deinit();
+                var cdi = meta.completed_data_indexes;
+                cdi.deinit();
```
0xNineteen, line 1137: why do this?

dnut: `meta` as a whole cannot be deinitialized because next_slots moving to the map. But when meta goes out of scope, completed_data_indexes is going to leak unless its backing memory is freed. It's assigned to a var because hashmaps need a mutable pointer to deinit.

PR #218, opened 2024-08-12
Update fnv.zig

chikoYEAT:
fixed wrong multiplier in `self.hash` in fn update()

### DIFF
--- a/src/crypto/fnv.zig
+++ b/src/crypto/fnv.zig
@@ -24,7 +24,7 @@ pub const FnvHasher = struct {
     pub fn update(self: *Self, input: []const u8) void {
         for (input) |byte| {
             self.hashh ^= byte;
-            self.hashh *%= prime;
+            self.hashh *= prime;
         }
     }
 


InKryption:
Based on the [wikipedia](https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function), wouldn't the more correct thing here be `self.hash = @mulWithOverflow(self.hash, prime)[0];`?

chikoYEAT:
FNV-1 hash algorithm works by:

1. Initializing the hash value with a predefined offset.
2. Iteratively multiplying the hash by a prime 
3. XOR-ing the hash with each byte of the input data.

algorithm uses a prime multiplier (0x100000001b3 for 64-bit FNV-1) in the multiplication step.

InKryption:
Then what exactly is wrong about using `*%=`? At least according to the wikipedia article:
> The [multiply](https://en.wikipedia.org/wiki/Multiply) returns the lower 64-[bits](https://en.wikipedia.org/wiki/Bit) of the [product](https://en.wikipedia.org/wiki/Product_(mathematics)).

And it would seem the zig stdlib's own implementation of fnv does exactly what we currently do. So this change needs better justification.

InKryption:
Justification does not make sense. `*%=` is equivalent to `*=` aside from its behavior on overflow, where the former performs modulo arithmetic, and the latter overflowing is illegal behavior.

PR #220, opened 2024-08-14
fix(accountsdb): file_map thread-safety

InKryption:
Before: we copy the RwMux of a given file in a bunch of places, resulting in not actually respecting the original RwMux in the file map.

Infeasible naive solution: lock all the account files in the file map after locking it.

Solution applied in this PR: make the entries in the file map an unwrapped AccountFile struct, and add a RwLock guard field that is documented as protecting access to all account file entries in the file map.

Possible future solution that may be necessary if this is too slow: shard the account file entries in the file map, such that reasonable groups of them are locked at a time instead of each individual file map, or all at the same time.

Edit: this also includes some changes to swiss map, such as the addition of an unmanaged variant. This started as part of the work to fix to the rwmux usage in account index, but as stated in a comment, I decided to do that in another PR - the changes are good though, and can tag along in this one.


src/accountsdb/db.zig
```diff
@@ -117,11 +117,21 @@ pub const AccountsDB = struct {
...
     pub const AccountCache = std.AutoHashMap(Slot, PubkeysAndAccounts);
     pub const DeadAccountsCounter = std.AutoArrayHashMap(Slot, u64);
     pub const BankHashStatsMap = std.AutoArrayHashMapUnmanaged(Slot, BankHashStats);
+    pub const FileMap = std.AutoArrayHashMapUnmanaged(FileId, AccountFile);
+
+    pub const DiskAccounts = struct {
+        /// Any code simply inserting account files into `map` may ignore this field,
+        /// it simply needs to acquire a write lock from the `map` rwmux.
+        /// Any code needing to read and/or write account file data, or remove account
+        /// files, must first acquire this lock, and then the lock on the `map` rwmux.
```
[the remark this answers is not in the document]
InKryption: I decided to expand on this, to more clearly delineate the three actual MOs of this thread-safety mechanism: ecc2bb3ccf16ac876bb5d410153191c4ea4a4d0b

src/accountsdb/db.zig
```diff
@@ -996,7 +1009,10 @@ pub const AccountsDB = struct {
                         const account, var lock_guard = try self.getAccountFromRefWithReadLock(max_slot_ref);
                         defer lock_guard.unlock();
 
-                        account_hash = account.hash(&key);
+                        account_hash = switch (account) {
+                            .file => |in_file| in_file.hash().*, // TODO: is this the same as the `.hash` function on `Account`?
```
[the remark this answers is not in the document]
InKryption: :facepalm: silly mistake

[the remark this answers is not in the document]
InKryption: c864c267c9cef9066442c2399a67c136a60baa20

InKryption:
I'm marking this ready for review because this is at a point where feedback is welcomed, but I still want to try and deal with the RwMux bug in `AccountIndex.gerReference` in this PR.

InKryption:
Current notable fuzz test benchmark:
```
hyperfine \
-w 3
-L exe ./zig-out/bin/fuzz,./zig-out-old/bin/fuzz \
-L seed '2220023635' \
-L actions '100000' \
'{exe} accountsdb {seed} {actions}'
```
ksnip_20240815-154507
Diff between accountsdb benchmark before and after the file map changes:
```diff
 Benchmark                                                              Iterations    Min(ns)    Max(ns)   Variance   Mean(ns)
 -----------------------------------------------------------------------------------------------------------------------------
-WRITE: 20913000
-readWriteAccounts(100k accounts (1_slot - ram index - ram accounts))            1   20220833   20220833          0   20220833
-WRITE: 35175292
-readWriteAccounts(100k accounts (1_slot - disk index - ram accounts))           1   21557583   21557583          0   21557583
-WRITE: 21198792
-readWriteAccounts(100k accounts (1_slot - ram index - disk accounts))           1   19876625   19876625          0   19876625
-WRITE: 25146334
-readWriteAccounts(100k accounts (1_slot - disk index - disk accounts))          1   20081208   20081208          0   20081208
+WRITE: 20975208
+readWriteAccounts(100k accounts (1_slot - ram index - ram accounts))            1   20614666   20614666          0   20614666
+WRITE: 35163625
+readWriteAccounts(100k accounts (1_slot - disk index - ram accounts))           1   21112625   21112625          0   21112625
+WRITE: 20002625
+readWriteAccounts(100k accounts (1_slot - ram index - disk accounts))           1   20335125   20335125          0   20335125
+WRITE: 24199584
+readWriteAccounts(100k accounts (1_slot - disk index - disk accounts))          1   20517583   20517583          0   20517583
```
These numbers indicate a slight regression in performance, however it does not appear to be to an untenable degree. There may be further performance improvements we can apply to this design before needing to resort to the more complex sharding solution.

InKryption:
> I'm marking this ready for review because this is at a point where feedback is welcomed, but I still want to try and deal with the RwMux bug in AccountIndex.gerReference in this PR.

I've decided this should maybe be done in a separate PR. The changes are related, but not interdependent.

0xNineteen:
we should also run the `accounts_db_snapshot` benchmark - you will need to download a testnet snapshot and put it in the `test_data/bench_snapshot/` dir

InKryption:
# `accounts_db_snapshot` benchmarks:
*Disclaimer: the snapshot used here is a snapshot I downloaded a month ago while working on accountsdb snapshot loading, and I don't remember whether it was a testnet, devnet, or even mainnet snapshot. I believe it was from testnet.*

Both binaries were compiled using `-Doptimize=RelaseSafe`.
The results presented here are a sample of various runs.

## Before
image

## After
image

### Observations
Minimal differences in between most runs, the timing differences observed in the above two runs vary between a relatively small range, each outpacing the other by a small margin in different measurements at random.
This matches my expectations, as snapshot loading largely involves creating new account files and inserting them into the file map, which does not require locking the account files for reading or writing, meaning performance should not be measurably different.

CC: @0xNineteen

src/accountsdb/db.zig
```diff
@@ -1645,15 +1692,15 @@ pub const AccountsDB = struct {
 
         var total_accounts_deleted: u64 = 0;
         for (shrink_account_files) |shrink_file_id| {
-            var shrink_account_file_rw = blk: {
-                const file_map, var file_map_lg = self.file_map.readWithLock();
+            self.file_map_safe.fd_keep_open_rw.lock();
```
0xNineteen, line 1695: this operation is what im worried about - holding this lock for too long ....

talked offline: should move this to queue the shrink_account_file into delete_account files

src/accountsdb/db.zig
```diff
@@ -1434,18 +1484,17 @@ pub const AccountsDB = struct {
...
-            const account_file, var account_file_lg = account_file_rw.readWithLock();
-            defer account_file_lg.unlock();
+            self.file_map_safe.fd_keep_open_rw.lockShared();
+            defer self.file_map_safe.fd_keep_open_rw.unlockShared();
+
+            const account_file = blk: {
+                const file_map, var file_map_lg = self.file_map_safe.file_map.readWithLock();
+                defer file_map_lg.unlock();
+                break :blk file_map.get(file_id).?;
+            };
```
0xNineteen, line 1497: this is a pretty common pattern which i feel like we should be able to contain in a method on self.file_map_safe - wdyt

my main concern is i dont like how many field accesses we need everywhere (ie, `self.file_map_safe.fd_keep_open_rw`) - should be a better way

src/accountsdb/db.zig
```diff
@@ -88,7 +88,7 @@ pub const AccountsDB = struct {
 
     /// track per-slot for purge/flush
     account_cache: RwMux(AccountCache),
-    file_map: RwMux(FileMap),
+    file_map_safe: ThreadSafeFileMap = .{},
```
0xNineteen, line 91: can probs come up with a better name than this, but this is what i have for now -- maybe `accountsbd.file_map.inner` is more accurate and we abstract most of it åway with methods?

0xNineteen (changes requested):

src/accountsdb/index/swiss_map.zig
```diff
@@ -0,0 +1,492 @@
+//! custom hashmap used for the index's map
```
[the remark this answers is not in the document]
InKryption: * 533ca01baa20b8adf1f0b68d0e8b54ac185c3135
* 9d9d609db92e131781d3c2eb68effbfda7dd7079

src/accountsdb/db.zig
InKryption: f175b944bceaefa3099b363f7b097a71d1f392f1

0xNineteen (approved):

src/accountsdb/db.zig
```diff
@@ -1667,37 +1674,45 @@ pub const AccountsDB = struct {
...
         defer {
-            const number_of_files = delete_account_files.len;
             self.stats.number_files_deleted.add(number_of_files);
         }
 
-        // hold the lock for the entire duration of delete to
-        // reduce how much contention exists with other threads.
-        self.file_map_safe.fd_keep_open_rw.lock();
-        defer self.file_map_safe.fd_keep_open_rw.unlock();
+        var account_file_infos = try std.ArrayList(struct { AccountFile, FileId }).initCapacity(
```
InKryption: Any reason to store the `FileId` again when the account file already carries its file id? Seems like we should just use that @0xNineteen

0xNineteen: great catch 68d5af6

0xNineteen (approved):

PR #222, opened 2024-08-16
feat(cli): improvement to fuzz cli filter 

dadepo:
Two improvements:

- Moved error handling/reporting to same place
- Include the supported fuzz filters in error message.

### DIFF
--- a/src/fuzz.zig
+++ b/src/fuzz.zig
@@ -10,6 +10,15 @@ const logger = lib.trace.log;
 // where seeds are saved (in case of too many logs)
 const SEED_FILE_PATH = "test_data/fuzz_seeds.txt";
 
+// Supported fuzz filters.
+// NOTE: changing these enum variants will require a change to the fuzz/kcov in `scripts/`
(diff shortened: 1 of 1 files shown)


InKryption (dismissed):
Nice use of enums for string matching

src/fuzz.zig
```diff
@@ -22,8 +31,12 @@ pub fn main() !void {
     const maybe_filter = cli_args.next();
     const filter = blk: {
```
InKryption, line 32: A worthwhile safety improvement that just occurred to me here would be that it'd be nice to move `maybe_filter` into the `blk` block here, as the return value of `.next()` is only valid up until its next call - technically on posix systems this ultimately has no impact since its actually a pointer to `std.os.argv` I believe, but if we're using the iterator API we might as well use it right. Same idea wrt `maybe_seed`.

InKryption (approved):
LGTM

InKryption:
CI failure unrelated to this PR.

PR #225, opened 2024-08-16
feat(docs) add a `docs` command for generating documentation

Rexicon226:
In this PR I've added a `zig build docs` step for generating and installing the autodoc into `zig-out/docs`. That can be then served via something like `python -m http.server`.

I had to rename the root module of the library from `lib.zig` to `sig.zig` because of how the autodoc system works. When looking for the "root" source file, the autodoc will go in this order:
- Name of the folder in the `sources.tar.gz`, so in this case `sig(.zig)`
- A `root.zig`
- The first entry in the tar file.

The name `lib.zig` won't work with it unfortunately and without a very hacky solution, there is no way to change this behavior.
 I'm happy to discuss whether it should be named `sig.zig` or `root.zig`, or if autodoc is worth renaming from `lib.zig`!

(First PR btw, the start to a fun journey!)


0xNineteen (approved):
:shipit:

PR #227, opened 2024-08-19
blockstore cleanup suggestions

dnut:
These are some suggested changes to the blockstore cleanup pr. I implemented them instead of leaving comments because I thought this would be faster.


dnut:
closing because i set the wrong base branch

PR #229, opened 2024-08-19
refactor(blockstore): rename blockstore to ledger and move shred

dnut:
Rename the "blockstore" folder to "ledger" since the purpose of the blockstore is to store solana's ledger. Also this is the name used in agave.

Move shred.zig into ledger since the ledger is composed of shreds. Originally this was in the shred collector because it had nowhere else to go, but now we have an entire folder dedicated to the ledger.


0xNineteen (approved):
:shipit:

PR #234, opened 2024-08-19
fix(thread): remove race on `deinit`

Rexicon226:
`Thread.unregister` would decrement `sync` until it was the last thread, at which point it would set the `join_event` to, in theory, wake up the `ThreadPool.join`. An optimization was made to skip the `join_event.wait`, however, this meant that the `sync` could be decremented and the `wake` skipped before `join_event.set` was called.

I think that this should fix the CI leaks.

### DIFF
--- a/src/sync/thread_pool.zig
+++ b/src/sync/thread_pool.zig
@@ -378,7 +378,7 @@ pub const ThreadPool = struct {
             self.run_queue.push(list);
         }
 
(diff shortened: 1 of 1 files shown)


InKryption (approved):
LGTM but would be good to get approval from either @dnut or @0xNineteen.

0xNineteen (approved):

PR #235, opened 2024-08-20
fix(accountsdb): AccountRefHead thread-safety

InKryption:
Follow-up to #220.

Before this change set code was copying RwMux(AccountReferenceHead)s, rendering it entirely thread-unsafe, since the actual rwlocks were not being respected.
After this change set, each account reference head is simply stored as is, and is considered to be under the same lock as the bin in which it lives - this is a similar fix as was applied to AccountsDB's file map.

This also comes with a myriad of improvements to AccountIndex which simplify the code dealing with the linked-list component of account references.

### DIFF
--- a/src/accountsdb/db.zig
+++ b/src/accountsdb/db.zig
@@ -678,7 +678,7 @@ pub const AccountsDB = struct {
         while (slot_iter.next()) |slot| {
             const refs = reference_memory.get(slot.*).?;
             for (refs.items) |*ref| {
-                _ = self.account_index.indexRefIfNotDuplicateSlot(ref);
+                _ = self.account_index.indexRefIfNotDuplicateSlotAssumeCapacity(ref);
                 ref_count += 1;
             }
 
@@ -779,7 +779,7 @@ pub const AccountsDB = struct {
     pub fn combineThreadIndexesMultiThread(
         logger: Logger,
         index: *AccountIndex,
-        thread_dbs: []AccountsDB,
+        thread_dbs: []const AccountsDB,
         // task specific
         bin_start_index: usize,
         bin_end_index: usize,
@@ -791,40 +791,33 @@ pub const AccountsDB = struct {
         const print_progress = thread_id == 0;
 
         for (bin_start_index..bin_end_index, 1..) |bin_index, iteration_count| {
-            const index_bin_rw = index.getBin(bin_index);
-
             // sum size across threads
             var bin_n_accounts: usize = 0;
             for (thread_dbs) |*thread_db| {
-                var bin_rw = thread_db.account_index.getBin(bin_index);
-                const bin, var bin_lg = bin_rw.readWithLock();
+                const bin, var bin_lg = thread_db.account_index.getBin(bin_index).readWithLock();
                 defer bin_lg.unlock();
 
                 bin_n_accounts += bin.count();
             }
             // prealloc
             if (bin_n_accounts > 0) {
-                const index_bin, var index_bin_lg = index_bin_rw.writeWithLock();
+                const index_bin, var index_bin_lg = index.getBin(bin_index).writeWithLock();
                 defer index_bin_lg.unlock();
 
                 try index_bin.ensureTotalCapacity(bin_n_accounts);
             }
 
             for (thread_dbs) |*thread_db| {
-                const bin_rw = thread_db.account_index.getBin(bin_index);
-                const bin, var bin_lg = bin_rw.readWithLock();
+                const bin, var bin_lg = thread_db.account_index.getBin(bin_index).readWithLock();
                 defer bin_lg.unlock();
 
                 // insert all of the thread entries into the main index
                 var iter = bin.iterator();
                 while (iter.next()) |thread_entry| {
-                    var thread_head_ref_rw = thread_entry.value_ptr.*;
-                    const thread_head_ref, var thread_head_ref_lg = thread_head_ref_rw.readWithLock();
-                    defer thread_head_ref_lg.unlock();
-
+                    const thread_head_ref = thread_entry.value_ptr.*;
                     // NOTE: we dont have to check for duplicates because the duplicate
                     // slots have already been handled in the prev step
-                    index.indexRef(thread_head_ref.ref_ptr);
+                    index.indexRefAssumeCapacity(thread_head_ref.ref_ptr);
                 }
             }
 
@@ -1044,9 +1037,8 @@ pub const AccountsDB = struct {
         }
         try hashes.ensureTotalCapacity(hashes_allocator, total_n_pubkeys);
 
-        // well reuse this over time so this is ok (even if 1k is an under estimate)
-        var keys = try self.allocator.alloc(Pubkey, 1_000);
-        defer self.allocator.free(keys);
+        var keys_buf = try std.ArrayList(Pubkey).initCapacity(self.allocator, 1000);
+        defer keys_buf.deinit();
 
         var local_total_lamports: u64 = 0;
         var timer = try sig.time.Timer.start();
@@ -1058,27 +1050,18 @@ pub const AccountsDB = struct {
             defer bin_lg.unlock();
 
             const n_pubkeys_in_bin = bin.count();
-            if (n_pubkeys_in_bin == 0) {
-                continue;
-            }
-            if (n_pubkeys_in_bin > keys.len) {
-                if (!self.allocator.resize(keys, n_pubkeys_in_bin)) {
-                    self.allocator.free(keys);
-                    const new_keys = try self.allocator.alloc(Pubkey, n_pubkeys_in_bin);
-                    keys.ptr = new_keys.ptr;
-                    keys.len = new_keys.len;
-                } else {
-                    keys.len = n_pubkeys_in_bin;
-                }
-            }
+            if (n_pubkeys_in_bin == 0) continue;
 
-            var i: usize = 0;
-            var key_iter = bin.iterator();
-            while (key_iter.next()) |entry| {
-                keys[i] = entry.key_ptr.*;
-                i += 1;
-            }
-            const bin_pubkeys = keys[0..n_pubkeys_in_bin];
+            try keys_buf.ensureTotalCapacity(n_pubkeys_in_bin);
+            keys_buf.clearRetainingCapacity();
+
+            const bin_pubkeys: []Pubkey = blk: {
+                var key_iter = bin.iterator();
+                while (key_iter.next()) |entry| {
+                    keys_buf.appendAssumeCapacity(entry.key_ptr.*);
+                }
+                break :blk keys_buf.items;
+            };
 
             std.mem.sort(Pubkey, bin_pubkeys, {}, struct {
                 fn lessThan(_: void, lhs: Pubkey, rhs: Pubkey) bool {
@@ -1088,9 +1071,7 @@ pub const AccountsDB = struct {
 
             // get the hashes
             for (bin_pubkeys) |key| {
-                const ref_head_rw = bin.getPtr(key).?;
-                const ref_head, var ref_head_lg = ref_head_rw.readWithLock();
-                defer ref_head_lg.unlock();
+                const ref_head = bin.getPtr(key).?;
 
                 // get the most recent state of the account
                 const ref_ptr = ref_head.ref_ptr;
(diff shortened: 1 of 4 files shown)


InKryption:
NOTE: initial measurements indicate this has regressed snapshot loading, further work on this PR will involve addressing this.

src/accountsdb/db.zig
```diff
@@ -1893,49 +1871,33 @@ pub const AccountsDB = struct {
...
-        const pubkeys, const accounts = blk: {
-            const account_cache, var account_cache_lg = self.account_cache.readWithLock();
+        const pubkeys: []const Pubkey, //
+        const accounts: []const Account //
+        = blk: {
+            const account_cache, var account_cache_lg = self.account_cache.writeWithLock();
             defer account_cache_lg.unlock();
 
-            const pubkeys, const accounts = account_cache.get(slot) orelse {
+            const removed = account_cache.fetchRemove(slot) orelse {
```
0xNineteen, line 1880: rename `removed_entry`

src/accountsdb/index.zig
```diff
@@ -39,7 +38,32 @@ pub const AccountReferenceHead = RwMux(struct {
...
+        null,
+        head,
+        inner: *?*AccountRef,
+    };
+    /// Returns a pointer to the field which is a pointer to the
+    /// account reference pointer with a field `.slot` == `slot`.
+    /// Returns `.null` if no account reference has said slot value.
+    /// Returns `.head` if `head_ref.ref_ptr.slot == slot`.
+    /// Returns `.inner = ptr` if `ptr.*.?.*.slot == slot`.
+    pub inline fn getPtrToFieldThatIsPtrToRefWithSlot(
```
0xNineteen, line 52: same with this name

src/accountsdb/index.zig
```diff
@@ -39,7 +38,32 @@ pub const AccountReferenceHead = RwMux(struct {
 
         return .{ rooted_ref_count, ref_slot_max };
     }
-});
+
+    pub const PtrToFieldThatIsPtrToRef = union(enum) {
+        null,
+        head,
```
0xNineteen, line 44: why even have head - why not just return the pointer directly?

src/accountsdb/index.zig
```diff
@@ -39,7 +38,32 @@ pub const AccountReferenceHead = RwMux(struct {
...
+    /// account reference pointer with a field `.slot` == `slot`.
+    /// Returns `.null` if no account reference has said slot value.
+    /// Returns `.head` if `head_ref.ref_ptr.slot == slot`.
+    /// Returns `.inner = ptr` if `ptr.*.?.*.slot == slot`.
+    pub inline fn getPtrToFieldThatIsPtrToRefWithSlot(
+        head_ref: *const AccountReferenceHead,
+        slot: Slot,
+    ) PtrToFieldThatIsPtrToRef {
+        if (head_ref.ref_ptr.slot == slot) return .head;
+        var curr_ref_ptr_ptr: *?*AccountRef = &head_ref.ref_ptr.next_ptr;
```
0xNineteen, line 57: same with this naming

src/accountsdb/index.zig
```diff
@@ -146,214 +170,170 @@ pub const AccountIndex = struct {
...
+    /// Get a write-safe account reference head, and its associated lock guard.
+    /// If access to many different account reference heads which are potentially in the same bin is
+    /// required, prefer instead to use `getBinFromPubkey(pubkey).write*(){.get(pubkey)}` directly.
+    pub fn getReferenceHeadWrite(self: *Self, pubkey: *const Pubkey) ?struct { AccountReferenceHead, RwMux(RefMap).WLockGuard } {
+        const bin, const bin_lg = self.getBinFromPubkey(pubkey).writeWithLock();
+        const ref_head = bin.get(pubkey.*) orelse return null;
+        return .{ ref_head, bin_lg };
+    }
+
+    pub const GetAccountRefError = error{ SlotNotFound, PubkeyNotFound };
```
0xNineteen, line 198: move to top of struct

src/accountsdb/index.zig
```diff
@@ -146,214 +170,170 @@ pub const AccountIndex = struct {
...
+        // to observe the `next_ptr` value, because they won't have the lock.
+        slot_ref_copy.next_ptr = null;
+        return slot_ref_copy;
     }
 
     pub fn exists(self: *Self, pubkey: *const Pubkey, slot: Slot) bool {
-        var head_reference_rw = self.getReference(pubkey) orelse return false;
-        const head_ref, var head_reference_lg = head_reference_rw.readWithLock();
-        defer head_reference_lg.unlock();
+        const head_ref, var head_ref_lg = self.getReferenceHeadRead(pubkey) orelse return false;
```
0xNineteen, line 239: should be bin_lg

src/accountsdb/index.zig
```diff
@@ -146,214 +170,170 @@ pub const AccountIndex = struct {
...
-                    // keep traversing
-                    prev_ref = curr_ref;
-                    curr_ref = curr_ref.next_ptr orelse return error.SlotNotFound;
-                }
-            }
+        var curr_ref = head_ref.ref_ptr;
+        while (curr_ref.next_ptr) |next_ref| {
+            curr_ref = next_ref;
         }
+        std.debug.assert(curr_ref.next_ptr == null);
```
0xNineteen, line 307: dont need this

src/accountsdb/swiss_map.zig
```diff
@@ -273,7 +277,22 @@ pub fn SwissMapUnmanaged(
...
+        }
+
+        fn removeImpl(
+            self: *@This(),
+            key: Key,
+            comptime ret: enum { void, value },
+        ) error{KeyNotFound}!switch (ret) {
+            .void => void,
+            .value => Value,
+        } {
```
0xNineteen, line 295: the less comptime zig specific things we can do the better - should just have fetchRemove named as remove, and we can ignore the value if we dont care about it

src/accountsdb/swiss_map.zig
```diff
@@ -62,6 +62,10 @@ pub fn SwissMapManaged(
             return @call(.always_inline, Unmanaged.remove, .{ &self.unmanaged, key });
```
0xNineteen, line 62: out of scope but `SwissMapManaged` should probs be just `SwissMap` to follow the way zig does it

src/utils/allocators.zig
```diff
@@ -320,3 +320,59 @@ test "disk allocator on arraylists" {
     };
     try std.testing.expect(did_error);
 }
+
+pub const failing = struct {
```
0xNineteen, line 324: dead code?

0xNineteen (changes requested):
few things on naming but overall looks good imo

src/accountsdb/swiss_map.zig
InKryption: In the future, zig may likely be switching to unmanaged by default, so in this case I decided to just qualify them both to avoid this going out of sync with that at all.

src/utils/allocators.zig
InKryption: It's dead, but potentially very useful code - it's (imo) a more useful version of `std.testing.failing_allocator`, which would even be reasonable to use outside of tests for the purposes of asserting that an allocating path is not reached.
I can remove it if you want, and just save it until such a time where we need it.

src/accountsdb/swiss_map.zig
InKryption: 9a7da9775ad6d8213c3d81aad892d5e7def41ef3

src/accountsdb/index.zig
InKryption: Good catch, part of old debug code.

InKryption: e1ad9751ee4f971c198b238203cf67d3b9449586

src/accountsdb/index.zig
InKryption: a4672c0d5bc5605471eb012f128b133d61d2ec99

src/accountsdb/index.zig
InKryption: 599503fa293d0b9e5ede22c54afa525937c810f4

src/accountsdb/index.zig
InKryption: Yeah sure I'm fine with returning the `*AccountRef` parent instead, it preserves the same capabilities, and doesn't actually add any extra bytes.

src/accountsdb/db.zig
InKryption: 0745b4b6290b777c764788678cf9834d20df8a7e

src/accountsdb/db.zig
```diff
@@ -1831,24 +1811,22 @@ pub const AccountsDB = struct {
...
                     const pubkey = account.pubkey();
-                    // SAFE: we know the pubkey exists in the index because its alive
-                    const old_ref = self.account_index.getReferenceSlot(pubkey, slot).?;
 
-                    // copy + update the values
-                    var new_ref = old_ref;
-                    new_ref.location.File.offset = offsets.items[offset_index];
-                    new_ref.location.File.file_id = new_file_id;
-                    offset_index += 1;
+                    const ref_ptr_ptr, var ref_lg = self.account_index.getReferencePtrPtrWrite(pubkey, slot) catch |err| switch (err) {
```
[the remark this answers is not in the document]
InKryption: 0745b4b6290b777c764788678cf9834d20df8a7e

src/accountsdb/index.zig
```diff
@@ -39,7 +38,32 @@ pub const AccountReferenceHead = RwMux(struct {
 
         return .{ rooted_ref_count, ref_slot_max };
     }
-});
+
+    pub const PtrToFieldThatIsPtrToRef = union(enum) {
```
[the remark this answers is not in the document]
InKryption: 0745b4b6290b777c764788678cf9834d20df8a7e

src/accountsdb/index.zig
InKryption: 0745b4b6290b777c764788678cf9834d20df8a7e

src/accountsdb/index.zig
InKryption: Just a note: the `getReferencePtrPtrWrite` can't be changed as easily in the same way, because it's intended to be agnostic to the `head` case. Could perhaps change it to also return the `PtrToRefField` union instead of a pointer to the parent, but lmk what you think.

src/utils/allocators.zig
0xNineteen: in that case - maybe add a comment explaining context around how its useful and how it could be used

src/accountsdb/swiss_map.zig
0xNineteen: makes sense but still not a huge fan of coding for future situations - leads to small things getting out of sync which add up

src/accountsdb/index.zig
InKryption: 73c6d1e56828fee1917852265734de353ab380f7

src/utils/allocators.zig
InKryption: Well, there is a doc comment, this is just the namespace in which it lives. This seemed better than having to prefix everything related to it with `FailingAllocator`:
* `utils.allocators.FailingAllocatorConfig` vs `utils.allocators.failing.Config`
* `utils.allocators.FailingAllocatorMode` vs `utils.allocators.failing.Mode`
* `utils.allocators.failingAllocator` vs `utils.allocators.failing.allocator`

src/accountsdb/swiss_map.zig
InKryption: Well the point in this case is that it can't go out of sync, since it's not adhering to any convention, it's simply denoting all the relevant information necessary despite context.
I understand the desire to avoid trying to predict the future, but zig is an evolving language, in which it would be unwise to write code based on patterns & conventions which can go out of sync in the near future, leading to communicating the wrong implicit information.

0xNineteen: its out of sync with the convention of the rest of the codebase rn

src/utils/allocators.zig
0xNineteen: it makes sense but is different from everywhere else in the codebase which leads to confusion

the config/mode isnt that big of deal since they wont be imported much -- you can just do `FailingAllocator.allocator(.{ .alloc = .noop_or_fail, ... })`

would rather have it follow how everywhere else does it (including how zig's GPA does it)

src/accountsdb/index.zig
0xNineteen: lgtm - we would want to modify more of the code to properly support it though -- `getReferencePtrPtrWrite` should be `getParentRefWrite` which any modifications we want to make can be made to `ref_field_ptr = parent_ref.next_ptr`

so we would go from `ref_ptr_ptr.* = new_ref_ptr;` to `parent_ref.next_ptr.* = new_ref_ptr`

InKryption: Like I said, we can't do that exactly, because there is the possibility of the `ref_ptr` field of the account ref head being the queried reference, in which case we can't return an `*AccountRef`.

src/utils/allocators.zig
InKryption: The stdlib does it that way for gpa because gpa isn't a stateless allocator - a better comparison would be to `c_allocator` or `page_allocator`: there is no `cAllocator(){}.allocator()`, you simply access the global constant for the `std.mem.Allocator` interface. The same occurs here, it is simply a parameterized global.

As for the namespace grouping, that is a very common convention throughout both the ecosystem and in the stdlib itself, and is a very useful way to organize code. It also facilitates being able to pull in an entire group of related declarations without having to reference each individual one.

I don't see any scenario where this could realistically lead to any real confusion: it's the same as if it were in another file and we exported the namespace here.

src/accountsdb/swiss_map.zig
InKryption: As far as I looked, I couldn't really find any other instances where we are implementing any data structures with both a managed and unmanaged version, could you explain how this breaks any currently established convention in the codebase?

InKryption: 28f90e9a55d6b8c2d203fe1b612b3f8ca6cc4d1c

src/accountsdb/index.zig
InKryption: a627c2a543b07517153ff28ca0c6c96de64b9ecc

src/utils/allocators.zig
InKryption: 43f2956f225d009a3583fd69c6e8502ae1912df6

0xNineteen (dismissed):

src/accountsdb/index.zig
InKryption: 027ace7ccaf01ed6a01049148aeac8719873b689

0xNineteen (approved):

PR #236, opened 2024-08-20
fix(ci): switch to using `mlugg@setup-zig` for better caching of `rocksdb`

Rexicon226:
Mlugg's `setup-zig` will cache the global directory for us, where the `rocksdb` dependency resides. This will negate the increased CI times from 3c8129697a6eb3e7e69366047547244c46a8b0cc.
I also took the opportunity to remove the `rocksdb` fix, since I don't believe it's needed anymore.


dadepo (approved):
LGTM

PR #237, opened 2024-08-20
feat(ci) Set timeout for CI

dadepo:


.github/workflows/check.yml
```diff
@@ -39,6 +39,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
     runs-on: ${{matrix.os}}
+    timeout-minutes: 60
```
dadepo, line 42: Not all that arbitrary...checked most of the run time in the past and it falls under ~30mins. Glancing around on agave and only a handful of runs push pass 1hr mark.

0xNineteen (approved):

PR #238, opened 2024-08-20
fix(accountsdb): move defer outside of if block, improve logs

InKryption:
The defer for the list clearings was being done in the if block, leading to the lists becoming cleared before the lists are actually passed as function arguments, meaning no slots were being cleaned, flushed, shrunk, or deleted.

### DIFF
--- a/src/accountsdb/db.zig
+++ b/src/accountsdb/db.zig
@@ -1180,15 +1180,16 @@ pub const AccountsDB = struct {
             }
 
             const must_flush_slots = flush_slots.items.len > 0;
+            defer if (must_flush_slots) {
+                flush_slots.clearRetainingCapacity();
+                unclean_account_files.clearRetainingCapacity();
+                shrink_account_files.clearRetainingCapacity();
+                delete_account_files.clearRetainingCapacity();
+            };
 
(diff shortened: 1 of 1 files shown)


src/accountsdb/db.zig
```diff
@@ -1446,7 +1447,7 @@ pub const AccountsDB = struct {
             const account_file, var account_file_lg = account_file_rw.readWithLock();
             defer account_file_lg.unlock();
 
-            self.logger.infof("cleaning slot: {}...", .{account_file.slot});
+            // self.logger.debugf("cleaning slot: {}...", .{account_file.slot});
```
yewman, line 1450: Is this intended to be commented out?

yewman (commented):
just the one question otherwise lgtm

src/accountsdb/db.zig
InKryption: Yeah, printing cleaned slots takes up a lot of real estate and isn't hugely useful compared to most of the other logs. If this information becomes valuable during a debugging session we can uncomment it, but otherwise I'd like to leave this commented.

yewman (approved):
lgtm

PR #240, opened 2024-08-21
fix(accounts-db): random fixes

0xNineteen:
- adds a new cli-flag `-n` which allows you to use pre-defined entrypoints (eg, `gossip -n testnet` will run gossip with the default testnet entrypoints) 
- adds a new `buildFullSnapshot` method on accounts-db to create a snapshot with many reasonable defaults 
- improve snapshot downloading log statements + error handling
- fix bug in `SnapshotFields.collapse` (src/accountsdb/snapshots.zig)
- add a few new commands to cli - `snapshot-create` which creates a snapshot, `print-manifest` which loads and prints some info on a manifest

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ data/
 index_storage/
 kcov-output/
 .vscode/
+alt_ledger/
 
 /gossip-dumps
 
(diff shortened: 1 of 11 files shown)


0xNineteen:
@InKryption  feel free to push status-cache fixes directly into this branch

InKryption (approved):
The PR LGTM, I added in some basic status cache fixes as requested, which I'll need someone else to check that it works on a testnet snapshot with:
```
1: ./zig-out/bin/sig snapshot-create -g genesis-files/testnet-genesis.bin -s <your ledger>
2: ./zig-out/bin/sig snapshot-validate -g genesis-files/testnet-genesis.bin -s alt_ledger
```
Ensure there is a full snapshot present and no incremental snapshot present in the ledger directory you use.
Currently trying to validate the status cache on an incremental snapshot appears to fail, at least with the small test_data snapshots we have; if someone else can also repro that with a testnet snapshot, or if they can't, that would be good to know.

PR #241, opened 2024-08-22
fix(ledger): shred recovery

dnut:
This PR includes a bug fix described in **Problem** as well as some related unit tests and refactoring changes that are described in **Other Changes**.

# Problem

On main, there is a bug in shred recovery where we see error.InvalidMerkleProof. It's thrown here:
https://github.com/Syndica/sig/blob/09bc1c9b5e27659dd545ebaef717f36886909aed/src/ledger/recovery.zig#L330-L331

This prevents most shreds from being recovered. Occasionally some shreds are recovered successfully, but usually we see this error preventing recovery.

## Background

Shred recovery uses erasure coding to recover shreds that were not received. For example, an erasure set may include 5 data shreds and 10 code shreds. The data shreds contain the information we care about. The code shreds only contain redundant data that allows us to recover the data shreds. From those 15 shreds, we could receive any 5 of them, and the reed solomon algorithm allows us to recover all 15 of the shreds.

Each erasure set also has a merkle tree calculated for it. Each shred contains its merkle proof.

During shred recovery, we calculate the merkle proof for every shred in the erasure set. For newly recovered shreds, the calculation is used to set the actual merkle proof in the shred, since that data is external to the erasure coding, and is not recovered by reed solomon.

We also calculate the merkle proof for every shred that we started with and already have the proof for. This is just a validation to ensure we are able to reproduce the same merkle proof that we already know is correct for that shred. This check ensures the integrity of the entire erasure set.

## Root cause

The error that we're seeing indicates that the wrong merkle proof was being calculated for existing shreds. The merkle proof was wrong because code shreds had the wrong indices being set. But the index field and the position in the common and code headers were being set incorrectly. Normally, each code shred should have a different index, representing its location in the erasure set and the slot. Instead, every code shred had an index that indicated it was the first shred in the erasure set.

I already added this logic to increment the index forward for each coding shred. As you can see it is assigned for this_common_header and this_coding_header. The bug is that those newly created headers with incremented indices are not actually being used.
https://github.com/Syndica/sig/blob/8637c2122d0242331c59de37c5102329bd97dc14/src/ledger/shredder.zig#L201-L209

## Fix

The fix is to use the headers with the incremented indices.
https://github.com/Syndica/sig/blob/09bc1c9b5e27659dd545ebaef717f36886909aed/src/ledger/recovery.zig#L279-L287

# Other changes

I made some other changes in the process of this investigation and fix.

## Shred recovery refactoring + tests
- move from shredder.zig into recovery.zig
- break apart massive `recover` function into several smaller functions
- unit test each function used for recovery. This also includes functions in shred.zig and reed_solomon.zig

## Rename "coding" shred names to "code"
Previously there was a mix of items named "code shred" and "coding shred". This change settles on "code shred" as the name to use everywhere. They are interchangeable so consistency is the priority here.

I also renamed variables and fields called "shred_variant" to "variant" when they are in the scope of a shred

## base58 refactoring
Hash, Pubkey, and Signature each need the same base58 logic for encoding statically sized arrays. I extracted out the common logic and put it into crypto/base58.zig

## eql function
I fleshed out the sig.utils.types.eql function to cover most of the cases I was looking for. This differs from std.meta.eql in that the std version means "are these actually the same exact item" and the sig version means "are these items equivalent". I believe the sig version is much more useful in most cases, whereas the std version is only useful in rare cases when you want to do things like comparing pointer equality. Specifically, the sig function:
https://github.com/Syndica/sig/blob/09bc1c9b5e27659dd545ebaef717f36886909aed/src/utils/types.zig#L280-L283


src/core/hash.zig
```diff
@@ -1,12 +1,17 @@
 const std = @import("std");
+const sig = @import("../sig.zig");
+
 const Sha256 = std.crypto.hash.sha2.Sha256;
-const base58 = @import("base58-zig");
 const Allocator = std.mem.Allocator;
 
-pub const HASH_SIZE: usize = 32;
+pub const HASH_SIZE: usize = Hash.size;
```
Rexicon226, line 7: I don't see why `HASH_SIZE` is better than `Hash.size`.

src/core/signature.zig
```diff
@@ -1,43 +1,64 @@
...
+const core = @import("lib.zig");
+
 const Ed25519 = std.crypto.sign.Ed25519;
 const Verifier = std.crypto.sign.Ed25519.Verifier;
 const e = std.crypto.errors;
 
-pub const SIGNATURE_LENGTH: usize = 64;
+const Pubkey = core.Pubkey;
+
+pub const SIGNATURE_LENGTH: usize = Signature.size;
```
Rexicon226, line 11: Ditto here, why not just use `Signature.size`?

src/ledger/recovery.zig
```diff
@@ -0,0 +1,500 @@
...
+    // Obtain erasure encoded shards from shreds.
+    const input_shreds = try allocator.alloc(?Shred, num_shards);
+    errdefer allocator.free(input_shreds);
+    const shards = try allocator.alloc(?[]const u8, num_shards);
+    errdefer {
+        for (shards) |shard| if (shard) |s| allocator.free(s);
+        allocator.free(shards);
+    }
+
+    for (input_shreds) |*s| s.* = null;
```
Rexicon226, line 188:
```suggestion
    @memset(input_shreds, null);
```

src/ledger/recovery.zig
```diff
@@ -0,0 +1,500 @@
...
+    for (shards) |*s| s.* = null;
```
Rexicon226, line 189:
```suggestion
    @memset(shards, null);
```

src/ledger/recovery.zig
```diff
@@ -0,0 +1,500 @@
...
+    shreds: RecoveryShreds,
+) ![]const Shred {
+    // Reconstruct code and data shreds from erasure encoded shards.
+    const all_shreds = try allocator.alloc(Shred, shreds.input_shreds.len);
+    var num_recovered_so_far: usize = 0;
+    errdefer {
+        // The shreds that are created below need to be freed if there is an error.
+        // If no error, ownership is transfered to calling scope.
+        // Existing shreds do not need to be freed because they were already owned by calling scope.
+        for (all_shreds, shreds.mask, 0..num_recovered_so_far) |shred, was_present, _| {
```
Rexicon226, line 232: multi-for loops in Zig assert that all elements are the same length. The `0..num_recovered_so_far` doesn't do anything.

src/ledger/recovery.zig
```diff
@@ -0,0 +1,500 @@
...
+    return meta;
+}
+
+fn organizeShredsForRecovery(
+    allocator: Allocator,
+    shreds: []const Shred,
+    meta: RecoveryMetadata,
+) !RecoveryShreds {
+    const num_shards: usize =
+        @intCast(meta.code_header.num_data_shreds + meta.code_header.num_code_shreds);
```
Rexicon226, line 177: Don't think you need this `@intCast`, `num_data_shreds` and `num_code_shreds` are `u16`s.

src/ledger/recovery.zig
```diff
@@ -0,0 +1,500 @@
...
+    var tree = try std.ArrayList(Hash).initCapacity(allocator, all_shreds.len);
+    defer tree.deinit();
+    for (all_shreds) |shred| {
+        tree.appendAssumeCapacity(try shred.merkleNode());
+    }
+    try makeMerkleTree(&tree);
+
+    // set the merkle proof on the recovered shreds.
+    const num_shards: usize =
+        @intCast(meta.code_header.num_data_shreds + meta.code_header.num_code_shreds);
```
Rexicon226, line 312: Don't think you need this `@intCast`, `num_data_shreds` and `num_code_shreds` are `u16`s.

src/ledger/recovery.zig
```diff
@@ -0,0 +1,500 @@
...
+            var i: usize = 0;
+            while (expected_proof_iterator.next()) |expected_entry| : (i += 1) {
+                const actual_entry = proof.get(i) orelse return error.InvalidMerkleProof;
+                if (!std.mem.eql(u8, expected_entry, &actual_entry)) {
+                    return error.InvalidMerkleProof;
+                }
+            }
+        } else {
+            try shred.setMerkleProof(proof);
+            std.debug.assert(if (shred.sanitize()) |_| true else |_| false);
```
Rexicon226, line 336:
```suggestion
            std.debug.assert(std.meta.isError(shred.sanitize()));
```

src/utils/types.zig
```diff
@@ -324,14 +386,36 @@ fn sliceEql(comptime T: type, a: []const T, b: []const T) bool {
         .stage2_riscv64 => false,
         else => true,
     };
-    if (!@inComptime() and std.meta.hasUniqueRepresentation(T) and backend_can_use_eql_bytes)
+    if (!@inComptime() and
+        std.meta.hasUniqueRepresentation(T) and
+        backend_can_use_eql_bytes and
```
Rexicon226, line 391: These checks happen in `std.mem.eql`, not need to do it here.

Rexicon226 (dismissed):
I just did a quick lookover, don't know much about code correctness, but there are a few changes I'd consider :)

src/core/hash.zig
dnut: HASH_SIZE is already used throughout the codebase and I wasn't trying to change every usage in this pr.

src/core/signature.zig
dnut: same as above. I'm not opposed to changing it, just didn't feel in scope at the time.

src/ledger/recovery.zig
dnut: > The 0..num_recovered_so_far doesn't do anything.

It does do something. It triggers a panic if an error happens during shred initialization. Which is obviously not what we want. I assumed the shortest one would constrain the loop, since unbounded ranges are always constrained by a bounded range. Guess I didn't know how loops work.

73b26d10c514dfb80ae8b65f34076638c2b022ca

dnut: 35763cb991c8f8894e5f915d6cecf3c8741e79fe

src/ledger/recovery.zig
```diff
@@ -0,0 +1,500 @@
...
+        for (self.mask, self.input_shreds) |was_present, shred| {
+            if (!was_present) {
+                if (shred) |s| s.deinit();
+            }
+        }
+    }
+};
+
+/// agave: merkle::recover
+pub fn recover(
```
0xNineteen, line 103: can we change this to include permalink + standard analgous format

src/ledger/recovery.zig
Rexicon226: I don't agree with the `ArrayList` solution, it's unnecessary overhead. If you need to constrain to the shortest one do something to the effect of:
```zig
const min_length = @min(all_shreds.len, shreds.mask.len, num_recovered_so_far);
for (all_shreds[0..min_length], shreds.mask[0..min_length]) ...
```
This 1. better describes what the code is trying to do and 2. Is just more efficient than an `ArrayList`.

src/ledger/shredder.zig
```diff
@@ -3,59 +3,7 @@ const sig = @import("../sig.zig");
...
-        const acquired = rs.acquire();
-        std.debug.assert(acquired);
-        const old = self.cache.put(.{ .data = data_shards, .parity = parity_shards }, rs);
-        std.debug.assert(old == null);
-        return rs;
-    }
-};
 
 /// Combines all shreds to recreate the original buffer
 pub fn deshred(allocator: Allocator, shreds: []const DataShred) !std.ArrayList(u8) {
```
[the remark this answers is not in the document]
dnut: Shredding and deshredding are not a part of recovery. It's all related to the erasure coding of shreds, but shred recovery is only one aspect. If I was going to combine it all into a single file, "shredder" might actually a better name than "recovery".

This shredder file still needs more logic added to it. Currently it only supports conversions in one direction, but it needs to support both directions eventually, also it needs to work with entries and not just bytes.

Recovery alone already takes up a few hundred lines of code which I think justifies its own file. Generally I prefer to keep files small and focused. I find it difficult to make sense of files with a thousand or more lines of non-test code.

src/ledger/recovery.zig
dnut: d29e2d6799fd58bae113708d61d31b4eead8f694

src/ledger/recovery.zig
dnut: How is this more efficient? Are you talking about the extra logic in `ensureTotalCapacityPrecise`?

I knew I could accomplish this without a list. The sole reason I chose to use a list is because I feel that it actually better describes what the code is trying to do.

When you iterate over the list, you're only going to access items that were added to the list. That's ultimately the goal here. It's also the first assumption I would expect to have about how a list would behave here. It's nice when the behavior of code is intuitive and simple to describe. It's also slightly more concise.

With the slice allocation there is this situation where the slice length actually includes undefined regions of memory and you need to carefully track it with a separate variable to avoid undefined behavior. It's a footgun that may not be obvious to every reader. The code to deal with it doesn't really help understand what the function is trying to accomplish. I find it cleaner to use a data structure that doesn't expose the footgun in the first place.

Ultimately it's not really a big deal, but I just wanted to express why I switched to using a list. But if there's a meaningful performance penalty then I could see removing the list.

src/ledger/recovery.zig
dnut: 9d24462d596fb2aae458c3483045e2240155f468

src/ledger/recovery.zig
dnut: 9d24462d596fb2aae458c3483045e2240155f468

src/utils/types.zig
dnut: I intentionally copied these checks directly from `std.mem.eql`.

Let's say we call `std.mem.eql` here without doing these checks first, and for some type, one of these checks are false. std.mem.eql will use `==` to compare each element. What I actually wanted is to use my `eql` function in that case. These filters ensure the desired behavior. Basically all I wanted was a way to call `eqlBytes` in the same situation when `std.meta.eql` would, just with an added filter to avoid comparing pointers. Since `eqlBytes` is private, the easiest way to do that is by indirectly calling it through `std.meta.eql`.

I added a comment to explain this. Also I improved the pointer filter to actually check if there are pointers contained, rather than blindly skipping the optimization.

c2a717961e08d700c05435cb35677725c2e6eb27

Rexicon226: Ah, you're right, thanks for explaining it to me. I missed that bit of control flow. :+1:

src/ledger/recovery.zig
Rexicon226: I understand why you're using a list here. Those seem like good reasons to me, and I'm all about showing code intent. I just need to think of the code as "accessing how many items were added" and not "accessing the smallest range possible", and it makes more sense.

> But if there's a meaningful performance penalty then I could see removing the list.

Probably not, but without a good reason (before you explained it to me), I would have rather just used the faster option.


Anyways, makes sense, I'm happy with it.

src/core/hash.zig
dnut: ef6543db2bce46db98fefc69cddcfc48d37d4439

src/core/signature.zig
dnut: ef6543db2bce46db98fefc69cddcfc48d37d4439

0xNineteen (approved):

PR #242, opened 2024-08-22
feat(accountsdb): improved metrics

Sobeston:
https://github.com/orgs/Syndica/projects/2?pane=issue&itemId=68540818

Dashboard preview:
image

### DIFF
--- a/src/accountsdb/db.zig
+++ b/src/accountsdb/db.zig
@@ -1441,7 +1441,7 @@ pub const AccountsDB = struct {
     /// as the data field ([]u8) for each account.
     /// Returns the unclean file id.
     pub fn flushSlot(self: *Self, slot: Slot) !FileId {
-        var timer = std.time.Timer.start() catch @panic("Timer unsupported");
+        var timer = try sig.time.Timer.start();
 
         defer self.stats.number_files_flushed.inc();
 
@@ -1587,8 +1587,6 @@ pub const AccountsDB = struct {
                 break :blk file_map.get(file_id).?;
             };
 
-            self.logger.debugf("cleaning slot: {}...", .{account_file.slot});
-
             var account_iter = account_file.iterator();
             while (account_iter.next()) |account| {
                 const pubkey = account.pubkey().*;
@@ -1684,7 +1682,7 @@ pub const AccountsDB = struct {
             references_to_delete.clearRetainingCapacity();
             self.stats.clean_references_deleted.set(references_to_delete.items.len);
             self.logger.debugf(
-                "cleaned slot {} -  old_state: {}, zero_lamports: {}",
+                "cleaned slot {} - old_state: {}, zero_lamports: {}",
                 .{ account_file.slot, num_old_states, num_zero_lamports },
             );
         }
--- a/src/accountsdb/db.zig
+++ b/src/accountsdb/db.zig
@@ -104,10 +104,10 @@ pub const AccountsDBStats = struct {
                     @field(self, field.name) = field_counter;
                 },
                 *Histogram => {
-                    const histogram_kind = std.meta.stringToEnum(
+                    const histogram_kind = comptime std.meta.stringToEnum(
                         HistogramKind,
                         field.name,
-                    ) orelse @panic("no matching HistogramKind for AccountsDBStats *Histogram field");
+                    ) orelse @compileError("no matching HistogramKind for AccountsDBStats *Histogram field");
 
                     const buckets = histogram_kind.buckets();
                     const field_histogram: *Histogram = try registry.getOrCreateHistogram(field.name, buckets);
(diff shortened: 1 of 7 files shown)


src/accountsdb/db.zig
```diff
@@ -1447,7 +1509,7 @@ pub const AccountsDB = struct {
             const account_file, var account_file_lg = account_file_rw.readWithLock();
             defer account_file_lg.unlock();
 
-            // self.logger.debugf("cleaning slot: {}...", .{account_file.slot});
```
Sobeston, line 1450: Was this commented out for a reason?

src/fuzz.zig
```diff
@@ -1,11 +1,14 @@
 const std = @import("std");
 const lib = @import("./lib.zig");
+const config = @import("./cmd/config.zig");
```
Sobeston, line 3: Or should we configure the metrics port another way?

src/accountsdb/db.zig
0xNineteen: yeah it was too much logs

src/accountsdb/db.zig
```diff
@@ -1313,6 +1362,8 @@ pub const AccountsDB = struct {
     /// as the data field ([]u8) for each account.
     /// Returns the unclean file id.
     pub fn flushSlot(self: *Self, slot: Slot) !FileId {
+        var timer = std.time.Timer.start() catch @panic("Timer unsupported");
```
0xNineteen, line 1365: should use `sig.time.Timer` here with a `try` ... - maybe we can move the `catch @panic` into the sig.time.Timer struct to remove the try in another pr

src/accountsdb/db.zig
```diff
@@ -1548,8 +1610,19 @@ pub const AccountsDB = struct {
                 }
             }
             references_to_delete.clearRetainingCapacity();
+            self.stats.clean_references_deleted.set(references_to_delete.items.len);
+            self.logger.debugf(
+                "cleaned slot {} -  old_state: {}, zero_lamports: {}",
```
0xNineteen, line 1615: extra space here

src/accountsdb/db.zig
```diff
@@ -1548,8 +1610,19 @@ pub const AccountsDB = struct {
...
+                .{ account_file.slot, num_old_states, num_zero_lamports },
+            );
         }
 
+        self.stats.clean_files_queued_deletion.set(delete_account_files.count());
+        self.stats.clean_files_queued_shrink.set(delete_account_files.count());
+        self.stats.clean_slot_old_state.set(num_old_states);
+        self.stats.clean_slot_zero_lamports.set(num_zero_lamports);
+
+        self.logger.debugf("cleanAccountFiles time elapsed: {}ns", .{timer.read()});
```
0xNineteen, line 1625: should probably move elapsed times into a bucket -- see how we handle it in gossip service `processMessages`

src/fuzz.zig
0xNineteen: i think this is ok for now

src/fuzz.zig
```diff
@@ -19,6 +22,13 @@ pub const FuzzFilter = enum {
     gossip_table,
 };
 
+fn spawnMetrics(gpa_allocator: std.mem.Allocator) !std.Thread {
+    const metrics_port: u16 = config.current.metrics_port;
+    logger.default_logger.infof("metrics port: {d}", .{metrics_port});
+    const registry = globalRegistry();
+    return std.Thread.spawn(.{}, servePrometheus, .{ gpa_allocator, registry, metrics_port });
+}
```
0xNineteen, line 30: can we move this into the prometheus file and call it from both here and `cmd.zig` (shouldnt need `spawnMetrics`) - seems like code duplication rn

src/accountsdb/db.zig
```diff
@@ -1313,6 +1362,8 @@ pub const AccountsDB = struct {
     /// as the data field ([]u8) for each account.
     /// Returns the unclean file id.
     pub fn flushSlot(self: *Self, slot: Slot) !FileId {
+        var timer = std.time.Timer.start() catch @panic("Timer unsupported");
+
```
0xNineteen, line 1366: outside of scope for a comment but i think with these metric changes we can delete the result returns on clean and shrink in the `runManagerLoop` and just return void instead (eg, change from `const clean_result = try self.cleanAccountFiles(` to `try self.cleanAccountFiles(` and delete `self.logger.debugf("clean_result: {any}", .{clean_result});` for both clean and shrink)

src/accountsdb/fuzz.zig
```diff
@@ -39,6 +39,8 @@ pub const TrackedAccount = struct {
 };
 
 pub fn run(seed: u64, args: *std.process.ArgIterator) !void {
+    _ = try sig.accounts_db.db.AccountsDBStats.init();
```
0xNineteen, line 42: why do we need this? this is included in AccountsDB.init

0xNineteen (changes requested):

src/accountsdb/db.zig
Sobeston: I'm thinking to make this change after removing the error union from Timer.start(). Happy to "resolve" this for now and pick it up later?

yewman (approved):
Looks good 👍

0xNineteen (approved):
:shipit:

PR #249, opened 2024-08-25
feat(trace): improve trace

dadepo:
Key points:
- [x] Replace allocation with RecycleFBA.
- [x] Add scope
  - [x] Be able to switch scope. 
- [x]  Be able to assert log output in tests
- [x] Complete implantation (add method for other log levels)
- [x] Try out interfaces
   - [ ] Tried this out but had issues with comptime. Illustrated here https://gist.github.com/dadepo/a0c47dda6fc87aaf737260f1a5961a55
- [ ] ~Make formatter pluggable~ For later
- [ ] ~Make output (writer) pluggable~ For later

Output currently

```bash
zig build test -Dfilter="trace_ng"
test
└─ run test stderr
[trace_ng.log.test.trace_ng: scope switch.Stuff] time=2024-09-03T09:05:32Z level=info doing stuff
[trace_ng.log.test.trace_ng: scope switch.StuffChild] time=2024-09-03T09:05:32Z level=info doing stuff details
time=2024-09-03T09:05:32Z level=info Logging with log
time=2024-09-03T09:05:32Z level=info Logging with logf
time=2024-09-03T09:05:32Z level=info f_agent=Firefox f_version=2.0 Logging with logWithFields
time=2024-09-03T09:05:32Z level=info f_agent=Firefox f_version=120 f_local=en f_stock=nvidia Logging with logfWithFields
```


src/trace_ng/log.zig
```diff
@@ -0,0 +1,273 @@
...
+    fba_bytes: u64,
+};
+
+const INITIAL_LOG_CHANNEL_SIZE: usize = 1024;
+
+const UnscopedLogger = StandardLogger(null);
+pub fn StandardLogger(comptime scope: ?type) type {
+    return struct {
+        const Self = @This();
+        level: Level,
```
dnut, line 29: What do you think about putting the log-level filtering in the interface instead of the implementation? We're probably going to want the same logic for every implementation right? "Do nothing if this log level is turned off."

Here's the tradeoffs I can think of:

pros:
- reduce redundant code
- improved performance since it preempts the need for dynamic dispatch. calling a function pointer is costly, compared to normal function calls

cons:
- unable to put special logic in the implementation that allows you to give some special treatment for excluded logs

Overall I don't think there's a huge difference which way we go.

The performance benefit could be meaningful. Ideally we shouldn't be afraid of improving the debug-level observability to the code. At some point maybe we should add a trace log-level along with scope-filtering for when we want a flood of data for some specific component.

Another idea for performance is comptime log filtering. There could be a be build parameter that results in certain calls to the logger being completely optimized out of the binary. Debug logs would truly have zero cost in that scenario. But this adds another dimension to the configuration which may not be worth the complexity.

The main issue with putting the filter in the interface is that the con is very open-ended. My guess is that it's a non-issue, but maybe in the future we'll suddenly realize that we care about it a lot? I guess we can make changes at that point if we need to.

src/trace_ng/log.zig
```diff
@@ -0,0 +1,273 @@
...
+            };
+            free_fba.mux.unlock();
+            errdefer {
+                free_fba.mux.lock();
+                free_fba.allocator().free(buf);
+                free_fba.mux.unlock();
+            }
+            var fmt_message = std.io.fixedBufferStream(buf);
+            const writer = fmt_message.writer();
+
```
dnut, line 133: Doesn't passing `self.fba_bytes` to this function mean that the logger will request an allocation for the entire backing buffer behind the fba? At that point is there any reason to use an allocator instead of passing the entire buffer?

src/trace_ng/log.zig
```diff
@@ -0,0 +1,273 @@
...
+            self.channel.deinit();
+            self.free_fba.deinit();
+        }
+
+        pub fn run(self: *Self) void {
+            while (!self.exit_sig.load(.seq_cst)) {
+                std.time.sleep(std.time.ns_per_ms * 5);
+                const messages = self.channel.drain() orelse {
+                    // channel is closed
+                    return;
```
dnut, line 92: You can use `std.fmt.count` to determine the number of bytes you need to allocate.

src/trace_ng/log.zig
```diff
@@ -0,0 +1,273 @@
...
+            free_fba: *RecycleFBA,
+            total_len: u64,
+            level: Level,
+            maybe_scope: ?[]const u8,
+            maybe_msg: ?[]const u8,
+            maybe_fields: anytype,
+            comptime maybe_fmt: ?[]const u8,
+            args: anytype,
+        ) logfmt.LogMsg {
+            // obtain a memory to write to
```
dnut, line 113: This should realloc the buffer to a smaller size.

When you call free in the other thread, it's going to pass in a len that is smaller than the actual size of the allocation. For the RecycleFBA it does not actually make a difference. But with other allocators it may cause a problem. The GPA will complain about this.

```zig
test "invalid free" {
    const x = try std.testing.allocator.alloc(u8, 100);
    std.testing.allocator.free(x[0..50]);
}
```

src/trace_ng/log.zig
dadepo: Used here https://github.com/Syndica/sig/pull/249/files#diff-a5f543f6d9dfe65e2e547cdf7bd2cd49d8e85d4afccc81a3c51568b0208dfeb7R29

src/trace_ng/log.zig
dadepo: For the formatted message, used  `std.fmt.count` to determine how much allocation to ask from the allocator...https://github.com/Syndica/sig/pull/249/files#diff-a5f543f6d9dfe65e2e547cdf7bd2cd49d8e85d4afccc81a3c51568b0208dfeb7R29

For the k/v fields on the other hand, I have not found how to determine the size. Currently hardcoding it to 512 here https://github.com/Syndica/sig/pull/249/files#diff-a5f543f6d9dfe65e2e547cdf7bd2cd49d8e85d4afccc81a3c51568b0208dfeb7R108

src/trace_ng/log.zig
dadepo: I could not get the interface/vtable approach to work. Ran into issues which I illustrated here https://gist.github.com/dadepo/a0c47dda6fc87aaf737260f1a5961a55 in other not to be blocked, I moved on with the the tagged union approach, which "in theory" should also be performant since it’s a more static type of dispatch.

For most of the re-usable logic, that was moved to the `logfmt` which give specific implementation ability to have custom logic while being able to re-use the ones that are re-usable.

src/trace_ng/log.zig
```diff
@@ -0,0 +1,600 @@
...
+                    return;
+                };
+                defer self.channel.allocator.free(messages);
+
+                for (messages) |message| {
+                    const writer = std.io.getStdErr().writer();
+                    logfmt.writeLog(writer, message) catch @panic("logging failed");
+                }
+                self.recycle_fba.mux.lock();
+                self.recycle_fba.alloc_allocator.reset();
```
dadepo, line 89: Not sure if this is the idiomatic way to do this, but without this it seems the space allocated is not reclaimed and given multiple allocation the recycle_fba get stuck when trying to allocate.

src/trace_ng/log.zig
```diff
@@ -0,0 +1,600 @@
...
+            const maybe_scope = blk: {
+                if (scope) |s| {
+                    break :blk @typeName(s);
+                } else {
+                    break :blk null;
+                }
+            };
+
+            const logMessage = logfmt.createLogMessage(&self.recycle_fba, self.max_buffer, level, maybe_scope, message, null, null, null);
+            self.channel.send(logMessage) catch @panic("could not send to channel");
```
dadepo, line 108: `@panic` is used a lot in the codebase but not sure if this is idiomatic/recommended

src/trace_ng/log.zig
```diff
@@ -0,0 +1,672 @@
...
+const INITIAL_LOG_CHANNEL_SIZE: usize = 1024;
+
+const LogKind = enum {
+    standard,
+    testing,
+    noop,
+};
+
+const UnscopedLogger = StandardLogger(null);
+pub fn StandardLogger(comptime scope: ?type) type {
```
dadepo, line 30: consider changing the scope to a string.

src/trace_ng/log.zig
```diff
@@ -0,0 +1,672 @@
...
+                .max_buffer = config.max_buffer,
+                .max_level = config.max_level,
+                .exit_sig = config.exit_sig,
+                .channel = Channel(logfmt.LogMsg).init(config.allocator, INITIAL_LOG_CHANNEL_SIZE),
+                .handle = null,
+            };
+        }
+
+        fn unscoped(self: *Self) *UnscopedLogger {
+            return @ptrCast(self);
```
dadepo, line 54: Suspect! Investigate.

src/trace_ng/log.zig
```diff
@@ -0,0 +1,672 @@
...
+        }
+
+        fn withScope(self: *Self, comptime new_scope: anytype) *StandardLogger(new_scope) {
+            return @ptrCast(self);
+        }
```
dnut, line 59: Instead of this you could have pub functions called `logger` and `scopedLogger` that return the interface by value (not by pointer).
```zig
fn logger(self: *Self) Logger {
    return .{ .standard = self };
}
fn scopedLogger(self: *Self, comptime scope: []const u8) ScopedLogger(scope) {
    return .{ .standard = self };
}
```

src/trace_ng/log.zig
```diff
@@ -0,0 +1,672 @@
...
+            );
+            self.channel.send(log_message) catch @panic("could not send to channel");
+        }
+    };
+
+    const TestingLogger = struct {
+        const Self = @This();
+        max_level: Level,
+        allocator: Allocator,
+        recycle_fba: RecycleFBA,
```
dnut, line 203: You don't need an allocator in this logger. You can format strings on the stack since they're not being sent over a channel

src/trace_ng/log.zig
```diff
@@ -0,0 +1,672 @@
...
+                null,
+                null,
+                fmt,
+                args,
+            );
+            const writer = self.log_msg.?.writer();
+            logfmt.writeLog(writer, log_message) catch @panic("Failed to write log");
+        }
+
+        pub fn logfWithFields(self: *Self, level: Level, comptime fmt: []const u8, args: anytype, fields: anytype) void {
```
dnut, line 313: There is something weird about having two anytypes back to back. I'm not sure what is best. What would you think about putting `fields` before `fmt`? This way the code that calls this method would actually look more similar to how usages of `.field()` currently looks.

dnut (changes requested):
Nice work. I have a few suggestions that, from my perspective, I think could help make the code safer, more readable, or more modular. Let me know what you think.

Also what do you think about removing the existing trace module and swapping out all usages with this logger?

src/trace_ng/log.zig
dnut:
```zig
logger.withScope(@typeName(@This()));
```

doesn't seem too bad

src/trace_ng/log.zig
dnut: You could remove all calls to ptrCast if you make some adjustments to passing things by ptr vs value. I mention this in the comments in my review.

src/trace_ng/log.zig
dnut: It's probably best not to have any panics in the logger. I would only explicitly panic in the code when the following conditions are met:
- the panicking branch can only be triggered if a bug has been introduced in the code
- the error is likely to corrupt the entire application and make it unusable or unsafe to use
- the error cannot be recovered from without editing the code and recompiling a new binary

src/trace_ng/log.zig
dnut: This could corrupt any allocations that have not been freed yet, and any future allocations that use the same region of memory.

src/trace_ng/log.zig
```diff
@@ -0,0 +1,672 @@
...
+                        .max_buffer = config.max_buffer,
+                    }) };
+                },
+            }
+        }
+
+        pub fn deinit(self: *Self) void {
+            switch (self.*) {
+                .standard => |logger| {
+                    var standard = logger;
```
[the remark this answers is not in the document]
dnut: `|*logger|` would solve this, but actually I think you can just remove this entire function.

dadepo:
> Nice work. I have a few suggestions that, from my perspective, I think could help make the code safer, more readable, or more modular. Let me know what you think.

Thanks! Will be going through the reviews now and processing them.

> Also what do you think about removing the existing trace module and swapping out all usages with this logger?

I was thinking of keeping that to a follow up PR so as to make this PR focused (removing the existing trace would lead to a lot of diffs that might be noise to the actual reviewing the new trace module)

src/trace_ng/log.zig
```diff
@@ -0,0 +1,672 @@
...
+        fn unscoped(self: *Self) *UnscopedLogger {
+            return @ptrCast(self);
+        }
+
+        fn withScope(self: *Self, comptime new_scope: anytype) *StandardLogger(new_scope) {
+            return @ptrCast(self);
+        }
+
+        pub fn spawn(self: *Self) void {
+            self.handle = std.Thread.spawn(.{}, Self.run, .{self}) catch @panic("could not spawn Logger");
```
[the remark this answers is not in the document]
dadepo: Copy _pasta_ from the current module https://github.com/Syndica/sig/blob/e765ca10d701a4bb2e94ff314e2d3f09687dd085/src/trace/log.zig#L175

Will improve in here to return an error.

src/trace_ng/log.zig
```diff
@@ -0,0 +1,672 @@
...
+
+const LogKind = enum {
+    standard,
+    testing,
+    noop,
+};
+
+const UnscopedLogger = StandardLogger(null);
+pub fn StandardLogger(comptime scope: ?type) type {
+    const StanardErrLogger = struct {
```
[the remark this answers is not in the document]
dadepo: > For more flexibility (and so people like me can understand it more easily) you could define these types outside the StandardLogger function. That way they can be directly initialized in other scopes. And it decouples the interface from the implementations. It keeps the interface simpler.

I remember trying this but ran into a snag because they need to capture the `comptime scope` scope. hence they need to be within the type function. 

I'll cycle back to this after processing the other comments

src/trace_ng/log.zig
```diff
@@ -0,0 +1,672 @@
...
+
+        pub fn unscoped(self: *Self) *UnscopedLogger {
+            return @ptrCast(self);
+        }
+
+        pub fn withScope(self: *Self, comptime new_scope: anytype) *StandardLogger(new_scope) {
+            return @ptrCast(self);
+        }
+
+        pub fn log(self: *Self, level: Level, message: []const u8) void {
```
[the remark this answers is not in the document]
dadepo: I think in this case it is fine if we deviate from the existing usages (as we have actually done already in some part of the API)...exposing public methods for each log level would blow up the method counts (4 to 16) and it would mostly be repetitive.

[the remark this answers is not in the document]
dnut: I think it would be more ergonomic and intuitive to have separate method for each log level. It's how all the popular loggers works in every language I've used.

I understand that large amounts of repetitive code can be seen as clutter, making it difficult to read and maintain code. But I think we can avoid it causing any actual problems, as long as the code is organized and the pattern is clear and simple with minimal boilerplate.

You could have a single file dedicate to defining only the logger interface. The file would just contain a single struct definition. That struct would only need 20 methods. 4 are the `log*` methods you currently have. The 16 additional methods would each have only a single line of code. The pattern would be clear and it wouldn't get in the way of anything else if that's the only thing in the file.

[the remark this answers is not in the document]
dadepo: > It's how all the popular loggers work in every language I've used

Then you have not used the tracing crate? :) https://docs.rs/tracing/latest/tracing/ (I think it also supports the individual methods but having the level as a param is also idiomatic as that is what is used in the examples)

Reason I was hesitant before is exactly what you also echoed "large amounts of repetitive code can be seen as clutter,"  - But I can update it, if that is indeed the requirements.

src/trace_ng/log.zig
```diff
@@ -0,0 +1,672 @@
...
+                &self.recycle_fba,
+                self.max_buffer,
+                level,
+                maybe_scope,
+                message,
+                null,
+                null,
+                null,
+            );
+            self.channel.send(log_message) catch @panic("could not send to channel");
```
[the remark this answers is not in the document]
dadepo: > ignores errors in the log* functions

> > I agree that's an improvement we can have in the new one.

Although this means if the logger runs into an error condition, it might be confounding to figure out why logs are not showing up (and no errors either) - an alternative may be to default to std.debug.err in this cases?

I won't do that now, and only ignore, but something to keep in mind if such a scenario pops up...

src/trace_ng/log.zig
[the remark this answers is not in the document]
dnut: I'm not opposed to having public methods that can be used as `log(.info`. In some cases I think those are actually more useful than `info` style methods, when the level needs to be dynamic. I'm advocating for publicly exposing *both* approaches in the api, like the tracing crate.

Let's say you're a consumer of the library and both options are available. You want to log an info message, and you have the choice between calling `log(.info` and `info`. Which one would you use? Personally I would use the info method in most cases, since it's easier to read and write. For me, that feels like a compelling reason to include it in the api.

> But I can update it, if that is indeed the requirements.

The requirements are for us to decide. I hope I didn't give the impression that it needs to strictly be whatever I say. I just wanted to share my reasoning in hope that you might agree with my conclusion. If we can't reach a consensus, we should loop in some others to get a fresh perspective.

src/trace_ng/log.zig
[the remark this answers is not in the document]
dnut: > Actually it panics in more places than not. See

Sorry, I think I was confusing this with `std.debug.print`, which just ignores errors and silently fails to print. I should have checked the actual code.

I don't think it's necessary to ignore the errors. I see your point about silent failures being confusing. I'm just concerned about putting explicit panics in the code. I feel like panics should only be used very rarely in the specific situation I described here: https://github.com/Syndica/sig/pull/249#discussion_r1742294348. I don't think these panics fit those criteria. I personally would prefer to return an error instead of panic.

src/trace_ng/log.zig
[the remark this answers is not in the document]
dadepo: > I'm advocating for publicly exposing both approaches in the api, like the tracing crate.

Yup. I definitely agree with this. My initial hesitation was that there would be a blow up of the methods...but at the end of the day it turned out to be not that bad :)

> I hope I didn't give the impression that it needs to strictly be whatever I say. I just wanted to share my reasoning in hope that you might agree with my conclusion.

Nope. It did not come across like that. Although I'll say I also don't mind differing requirements spec to older contributors, given I am just start contributing, hence I don't have a lot of the context that may influence certain decisions.

src/trace_ng/log.zig
dadepo: So revising this now, and for example we currently have

```
   logger.errfWithFields(
        "{s}",
        .{"Logging with logfWithFields"},
        .{
            .f_agent = "Firefox",
            .f_version = 120,
            .f_local = "en",
            .f_stock = "nvidia",
        },
    );
```

I can move the fields to to be the first argument,  which then becomes:

```
   logger.errfWithFields(.{
            .f_agent = "Firefox",
            .f_version = 120,
            .f_local = "en",
            .f_stock = "nvidia",
        },
        "{s}",
        .{"Logging with logfWithFields"},
    );
```

This is without renaming. For renaming it would then be something like

```
   logger.fieldsErrf(.{
            .f_agent = "Firefox",
            .f_version = 120,
            .f_local = "en",
            .f_stock = "nvidia",
        },
        "{s}",
        .{"Logging with logfWithFields"},
    );
```

only that this won't be consistent with the other naming `err`, `errf` where the `err` (or `info`, or `debug` etc) comes first.

What do you think? We move the fields to the first but keep the naming?

src/trace_ng/log.zig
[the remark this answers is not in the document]
InKryption: Just a note: easy way to get around this would be to simply parameterize the implementations with the scope parameter as well.

src/trace_ng/log.zig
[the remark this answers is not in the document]
dadepo: I also agree that `fallback to std.debug.print` is probably the best option here. I have updated to ignoring error, but I i'll update in a bit to fallback to std.debug.print instead.

src/trace_ng/log.zig
```diff
@@ -0,0 +1,1048 @@
...
+                .channel = Channel(logfmt.LogMsg).init(config.allocator, INITIAL_LOG_CHANNEL_SIZE),
+                .handle = try std.Thread.spawn(.{}, run, .{self}),
+            };
+            return self;
+        }
+
+        pub fn deinit(self: *Self) void {
+            self.channel.close();
+            if (self.handle) |*handle| {
+                std.time.sleep(std.time.ns_per_ms * 5);
```
dadepo, line 240: Not 100% sure if this is the best way to fix this, but the issue I found was that in the test case where the scope of the logger was switched (ie essentially having two loggers...the one used by the parent struct and the child struct) it often happens that the log from the child does not make it to the std err before the process dies. It seems the handle.join() sometimes does not work...but if I introduce a `std.debug.print` (or the sleep above) it consistently shows up.

src/trace_ng/log.zig
```diff
@@ -0,0 +1,1048 @@
...
+
+    const logger = Logger.init(.{
+        .allocator = allocator,
+        .max_level = Level.info,
+        .max_buffer = 2048,
+    }) catch @panic("Logger init failed");
+    defer logger.deinit();
+
+    var stuff = Stuff.init(&logger);
+    defer stuff.deinit();
```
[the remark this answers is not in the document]
dnut: I think the typical usage pattern would be that you initialize StdErrLogger somewhere near the main function of your application. You defer deinit in that scope only. Then you convert it to a Logger, which contains a pointer to StdErrLogger. The logger gets passed around everywhere by value, but it never needs to be deinited because it's just a pointer to the logger owned by main, which will be deinited when main returns.

src/trace_ng/log.zig
```diff
@@ -0,0 +1,1048 @@
...
+        max_level: Level,
+        exit_sig: std.atomic.Value(bool),
+        allocator: Allocator,
+        log_allocator: Allocator,
+        log_allocator_state: *RecycleFBA(.{}),
+        max_buffer: u64,
+        channel: *Channel(logfmt.LogMsg),
+        handle: ?std.Thread,
+
+        pub fn init(config: Config) !*Self {
```
dnut, line 219: Is there a particular reason why this this returns a pointer to self? It could just return `Self` instead of `*Self`, which would simplify the function. Then the calling scope has the choice of how to allocate it. Currently it's forced into allocating the logger with the same allocator that's used for log messages, which may not always be a desirable constraint.

src/trace_ng/log.zig
```diff
@@ -0,0 +1,1048 @@
...
+            return .{
+                .allocator = self.allocator,
+                .recycle_fba = self.log_allocator_state,
+                .max_buffer = self.max_buffer,
+                .max_level = self.max_level,
+                .exit_sig = self.exit_sig,
+                .channel = self.channel,
+                .handle = self.handle,
+            };
+        }
```
dnut, line 272: These functions would fail to compile if they were actually used. `ScoppedLogger` is a union that holds only a pointer to the logger implementation.

For these to work it just needs to create the union from the pointer to self. Also I might change the function names to clarify that it's returning a different type. This is consistent with the interface pattern used in zig std.

```suggestion
        pub fn logger(self: *Self) Logger {
            return .{ .standard = self };
        }

        pub fn scopedLogger(self: Self, comptime new_scope: anytype) ScoppedLogger(new_scope) {
            return .{ .standard = self };
        }
```

src/trace_ng/log.zig
```diff
@@ -0,0 +1,1048 @@
...
+const LogKind = enum {
+    standard,
+    testing,
+    noop,
+};
+
+/// A ScopedLogger could either be:
+/// - A StandardErrLogger
+/// - A TestingLogger
+pub fn ScoppedLogger(comptime scope: ?[]const u8) type {
```
dnut, line 31:
```suggestion
pub fn ScopedLogger(comptime scope: ?[]const u8) type {
```

src/trace_ng/log.zig
```diff
@@ -0,0 +1,1048 @@
...
+                .standard => |*logger| {
+                    logger.*.deinit();
+                },
+                .testing => |*logger| {
+                    logger.*.deinit();
+                },
+                .noop => {},
+            }
+        }
+
```
dnut, line 67: Do we need these functions? The caller could instead initialize whatever logger implementation they want, and convert it to the interface.

```zig
const std_logger = StdErrLogger.init(...);
const logger = std_logger.logger();
```

This approach is more flexible because it doesn't couple all the logger `init` functions together to have a unified set of dependencies.

src/trace_ng/log.zig
```diff
@@ -0,0 +1,1048 @@
...
+                        .allocator = logger.*.allocator,
+                        .kind = LogKind.testing,
+                    }) catch @panic("message: []const u8");
+                },
+                .noop => {
+                    @panic("Cannot scope noop");
+                },
+            }
+        }
+
```
dnut, line 109: Why not just use the existing logger instead of creating a new one? Also what's wrong with scoping noop?

src/trace_ng/log.zig
```diff
@@ -0,0 +1,1048 @@
...
+                inline else => |impl| impl.logfWithFields(level, fmt, args, fields),
+            }
+        }
+    };
+}
+
+pub const Logger = ScoppedLogger(null);
+
+/// An instance of `ScopedLogger` that logs to the standard err.
+pub fn StandardErrLogger(comptime scope: ?[]const u8) type {
```
dnut, line 207: You could make `scope` a parameter for the log functions instead of being a parameter for the logger implementation types. The logger interface would still need it, but the implementations of the interface do not need it.

Personally I feel that it simplifies the code because this can be defined as a normal struct instead of a generic struct. It cleans up an unnecessary layer of abstraction

dadepo: Not sure if/how that would work since the implementation contains logic that refers to the scope. For example:

```
    pub fn log(self: *Self, level: Level, message: []const u8) void {
        if (@intFromEnum(level) > @intFromEnum(self.max_level)) {
            // noop
            return;
        }

        const maybe_scope = if (scope) |s| s else null; // <---- The scope needed here
        const log_msg = logfmt.LogMsg{
            .level = level,
            .maybe_scope = maybe_scope,
            .maybe_msg = message,
            .maybe_fields = null,
            .maybe_fmt = null,
        };

        self.channel.send(log_msg) catch |err| {
            std.debug.print("Send msg through channel failed with err: {any}", .{err});
            return;
        };
    }
```

src/trace_ng/log.zig
dadepo: The idea behind the current approach is to have the caller call one init function (specifying the kind of logger they want) and they good to go ie:

```
const std_logger =  Logger.init(.{.. .kind = LogKind.standard,})
```

Your approach would be two calls, but it might indeed be more flexible since it removes the init from the interface. I'll play around with it and see.

src/trace_ng/log.zig
dadepo: > Is there a particular reason why this this returns a pointer to self?

If the `union` holds pointers, ie

```
pub fn ScopedLogger(comptime scope: ?[]const u8) type {
    return union(LogKind) {
       ...
        standard: *StandardErrLogger(scope),
        testing: *TestingLogger(scope),
        ...
}
```

And `ScopedLogger.init(...)` needs to be able to create one of the union variants, then it needs to return a a pointer to self.

With the suggestion here https://github.com/Syndica/sig/pull/249#discussion_r1751166855 might be possible to remove them.

src/trace_ng/log.zig
dadepo: Unless the method on the implementation be modified to take scope like this

```zig
pub fn log(self: *Self, comptime scope: ?[]const u8, level: Level, message: []const u8) void
```

and then modify the interface to supply the scope when calling the implementation.

dadepo: Done in [460cf27](https://github.com/Syndica/sig/pull/249/commits/460cf2779a39f19fe1a5ff26c15eccda90796ea1)

src/trace_ng/log.zig
dadepo: > Why not just use the existing logger instead of creating a new one? 

That is due to the pattern of passing in a logger, and allowing the struct use that to create a new one that is scoped.

```zig
    const StuffChild = struct {
        const StuffChild = @This();
        logger: ScopedLogger(@typeName(StuffChild)), // <- Scoped logger needed by struct

        pub fn init(logger: *const Logger) StuffChild {
            return .{ .logger = logger.withScope(@typeName(StuffChild)) catch { // <- New scoped logger created
                @panic("Init logger failed");
            } };
        }
```

Also if this is not done this way, ie, if the struct uses existing one, won't there be the risk of changing the scope externally?

> Also what's wrong with scoping noop?

Nothing. It could technically be scoped, just did not pay much attention to it yet and defaulted to panic. But can update.

src/trace_ng/log.zig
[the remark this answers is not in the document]
dadepo: For scoped logger, does this not bring back the possibility for error where a scoped logger is passed and used in a different scope than was intended?

I felt having the logger be part of the state of the struct, ie

```zig

const Foo = struct {
  logger: ScopedLogger(Foo Scope)

  self.logger.info(..)
} 
```

Would ensure that logging from `Foo` would always be scoped to `Foo Scope`. 

And I am not sure that can be accomplished with the approach you described.

src/trace_ng/log.zig
dnut: The instance of the underlying logger implementation can remain the same while changing the scope.

```zig
        pub fn unscoped(self: Self) Logger {
            return switch (self) {
                .standard => |logger| .{ .standard = logger },
                .testing => |logger| .{ .testing = logger },
                .noop => .noop,
            };
        }

        pub fn withScope(self: Self, comptime new_scope: []const u8) !ScoppedLogger(new_scope) {
            return switch (self) {
                .standard => |logger| .{ .standard = logger },
                .testing => |logger| .{ .testing = logger },
                .noop => .noop,
            };
        }
```

src/trace_ng/log.zig
dnut: > the scope of the logger was switched (ie essentially having two loggers

This wouldn't be an issue if you use only a single instance of the underlying logger.

src/trace_ng/log.zig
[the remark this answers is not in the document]
dnut: Scope is not a part of the state. It's an immutable part of the type. Moving the logger to another place and giving a new scope in that place is not going to make any changes to the state of the original logger. The original logger will continue using the scope as specified in its type.

src/trace_ng/log.zig
```diff
@@ -0,0 +1,1028 @@
...
+        while (!self.exit_sig.load(.seq_cst)) {
+            std.time.sleep(std.time.ns_per_ms * 5);
+            const messages = self.channel.drain() orelse {
+                // channel is closed
+                return;
+            };
+            defer self.channel.allocator.free(messages);
+            for (messages) |message| {
+                const writer = std.io.getStdErr().writer();
+                std.debug.lockStdErr();
```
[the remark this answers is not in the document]
dadepo: I think the base case is to acquire the lock for each message written. This is what the std lib log does (alongside using buffered writer), and also debug.print. And given that we are offloading the writing out of the log message off to another thread, hopefully the computation needed to acquire/release the std err does not easily affect other working threads.

Also giving that the stderr is a shared resource, i'll err on the side of not holding on to it too much. Even though the likelihood of that happening is low in this case, if there is a bug in the logger, where the loop runs for more time than needed, there is the possibility of hogging unto the shared resource for more than needed.

So maybe acquire the lock for each message (as it is currently) but if in practice, this proofs to have a noticeable performance impact, then it can be revisited?

src/trace_ng/log.zig
```diff
@@ -0,0 +1,1028 @@
...
+    pub fn run(self: *Self) void {
+        while (!self.exit_sig.load(.seq_cst)) {
+            std.time.sleep(std.time.ns_per_ms * 5);
+            const messages = self.channel.drain() orelse {
+                // channel is closed
+                return;
+            };
+            defer self.channel.allocator.free(messages);
+            for (messages) |message| {
+                const writer = std.io.getStdErr().writer();
```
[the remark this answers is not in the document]
dadepo: I think same comment here https://github.com/Syndica/sig/pull/249#discussion_r1752387168 applies here too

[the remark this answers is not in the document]
dadepo: Actually misread the comment to be referring to the lock acquisition.

> What's the advantage of instantiating the writer inside the loop?

I don't think there is any. Removed in [45c4535](https://github.com/Syndica/sig/pull/249/commits/45c4535a884f770df32eec36ca17dc5265526d0b)

0xNineteen (commented):
lgtm - we'll want this fully integrated (production ready) before we merge this to main - to make reviewing easier (like you mentioned previously), could you create another PR which branches off this PR which integrates the new logger?

once integration is reviewed and approved well merge both into main

PR #250, opened 2024-08-26
chore: Ensure test utility are only used in testing context

dadepo:
One idea was to move all test only methods to a utility struct in the module being tested.

Ensuring test utility are only used in testing context is the other idea which is not as intrusive as the first idea.


0xNineteen:
some of the random/default methods will be used in fuzz testing and benchmarking code which would break -- its a good idea, but probably best to not restrict these for now

PR #251, opened 2024-08-26
test(ledger)

dnut:


src/bincode/bincode.zig
```diff
@@ -797,6 +808,21 @@ const ShredType = enum(u8) {
     pub const @"!bincode-config" = ShredTypeConfig();
 };
 
+pub fn testRoundTrip(comptime T: type, test_data: anytype) !void {
+    const allocator = std.testing.allocator;
+    const expected_bytes = test_data.bincode_serialized_bytes;
+    const expected_struct = try test_data.asStruct(allocator);
```
0xNineteen, line 814: why not just pass `expected_struct` and `expected_byte` as paramters instead of enforcing them as fields/methods on an anytype?

src/core/transaction.zig
```diff
@@ -360,3 +451,117 @@ test "core.transaction: Message sanitize fails if account index is out of bounds
...
+        2,   85,  213, 101, 44,  13,  181, 13,  65,  128, 17,  94,  229, 31,  215, 47,  49,  72,
+        57,  158, 144, 193, 224, 205, 241, 120, 78,  5,   1,   3,   5,   7,   90,  0,
+    };
+};
+
+fn allocopy(allocator: std.mem.Allocator, comptime T: type, source: []const T) ![]T {
+    const new_slice = try allocator.alloc(T, source.len);
+    @memcpy(new_slice, source);
+    return new_slice;
+}
```
0xNineteen, line 567: isnt this `allocator.dupe` ?

src/utils/types.zig
```diff
@@ -433,6 +443,69 @@ pub fn containsPointer(comptime T: type) ?bool {
     };
 }
 
+/// This is an exact copy of std.mem.eqlBytes because it is private
```
0xNineteen, line 446: just a side note, this is another example of why i push for all our methods to be public

src/bincode/bincode.zig
```diff
@@ -797,6 +808,21 @@ const ShredType = enum(u8) {
     pub const @"!bincode-config" = ShredTypeConfig();
 };
 
+pub fn testRoundTrip(comptime T: type, test_data: anytype) !void {
```
yewman, line 811: I think some documentation briefly explaining this method and the constraints on test_data would be useful

src/ledger/rocksdb.zig
```diff
@@ -70,14 +70,9 @@ pub fn RocksDB(comptime column_families: []const ColumnFamily) type {
             };
         }
 
-        pub fn deinit(self: *Self, delete_disk: bool) void {
+        pub fn deinit(self: *Self) void {
```
0xNineteen, line 73: any reason you removed `delete_disk` ?

src/ledger/tests.zig
```diff
@@ -44,3 +244,175 @@ pub fn freshDir(path: []const u8) !void {
...
+            self.db.deinit();
+            self.registry.deinit();
+            std.testing.allocator.destroy(self._leak_check);
+            _allocator.destroy(self);
+            _ = gpa.detectLeaks();
+        }
+    };
+}
+
+pub fn TestDB(scope: []const u8) type {
```
0xNineteen, line 404: why not just have `scope` as a parameter to `init` ? or have the scope prefix as a global constant in the file so you have `.init(TEST_DB_SCOPE + "blockstoreTest")` - want to minimize the layers of abstraction wherever possible imo

src/ledger/insert_shred.zig
```diff
@@ -1861,77 +1861,72 @@ fn newlinesToSpaces(comptime fmt: []const u8) [fmt.len]u8 {
 
 const test_shreds = @import("test_shreds.zig");
 const comptimePrint = std.fmt.comptimePrint;
+const TestDB = ledger.tests.TestDB("insert_shred");
```
0xNineteen, line 1864: not used?

src/ledger/insert_shred.zig
```diff
@@ -1861,77 +1861,72 @@ fn newlinesToSpaces(comptime fmt: []const u8) [fmt.len]u8 {
...
-    _leak_check: []const u8,
-
-    // var gpa = std.heap.GeneralPurposeAllocator(.{ .stack_trace_frames = 100 }){};
-    // pub const allocator = gpa.allocator();
-
-    pub const allocator = std.testing.allocator;
-
-    pub fn init(comptime test_name: []const u8) !TestState {
+    pub fn init(comptime test_name: []const u8) !ShredInserterTestState {
         return initWithLogger(test_name, (sig.trace.TestLogger{}).logger());
```
0xNineteen, line 1879: it makes sense but is strange to use this way to create a logger when everwhere else we do it a different way ... 

the way it works, we could probably just use it as a global constant in the logger file? `sig.trace.test_logger`

src/core/transaction.zig
```diff
@@ -360,3 +451,117 @@ test "core.transaction: Message sanitize fails if account index is out of bounds
...
+        83,  42,  196, 46,  189, 136, 194, 90,  249, 14,  154, 144, 141, 234, 253, 148, 146, 168,
+        110, 10,  237, 82,  157, 190, 248, 20,  215, 105, 1,   100, 2,   1,   3,   32,  104, 232,
+        42,  254, 46,  48,  104, 89,  101, 211, 253, 161, 65,  155, 204, 89,  126, 187, 180, 191,
+        60,  59,  88,  119, 106, 20,  194, 80,  11,  200, 76,  0,   1,   8,   65,  203, 149, 184,
+        2,   85,  213, 101, 44,  13,  181, 13,  65,  128, 17,  94,  229, 31,  215, 47,  49,  72,
+        57,  158, 144, 193, 224, 205, 241, 120, 78,  5,   1,   3,   5,   7,   90,  0,
+    };
+};
+
+fn allocopy(allocator: std.mem.Allocator, comptime T: type, source: []const T) ![]T {
```
yewman, line 563: Is this different to `allocator.dupe(comptime T: type, source: []const T)`

0xNineteen (changes requested):

src/ledger/tests.zig
```diff
@@ -44,3 +244,175 @@ pub fn freshDir(path: []const u8) !void {
...
+///         let payload = shred.payload();
+///         file.write(&payload.len().to_le_bytes()).unwrap();
+///         file.write(payload).unwrap();
+///     }
+/// }
+/// ```
+pub fn loadShredsFromFile(allocator: Allocator, path: []const u8) ![]const Shred {
+    const file = try std.fs.cwd().openFile(path, .{});
+    const reader = file.reader();
+    var shreds = std.ArrayList(Shred).init(allocator);
```
yewman, line 272: Does `errdefer shreds.deinit();` make sense here? Not really an issue though since I believe this is only used in testing anyway

src/ledger/tests.zig
```diff
@@ -44,3 +244,175 @@ pub fn freshDir(path: []const u8) !void {
...
+    const num_size_bytes_read = try reader.readAll(&size_bytes);
+    if (num_size_bytes_read == 0) {
+        return null;
+    }
+    if (num_size_bytes_read != 8) {
+        return error.IncompleteSize;
+    }
+    const size = std.mem.readInt(u64, &size_bytes, .little);
+
+    const chunk = try allocator.alloc(u8, @intCast(size));
```
yewman, line 296: `errdefer chunk.deinit();`?

src/ledger/tests.zig
```diff
@@ -44,3 +244,175 @@ pub fn freshDir(path: []const u8) !void {
...
+    for (shreds) |shred| shred.deinit();
+    allocator.free(shreds);
+}
+
+/// Read entries from binary file structured like this:
+/// [entry0_len: u64(little endian)][entry0_bincode][entry1_len...
+pub fn loadEntriesFromFile(allocator: Allocator, path: []const u8) ![]const Entry {
+    const file = try std.fs.cwd().openFile(path, .{});
+    const reader = file.reader();
+    var entries = std.ArrayList(Entry).init(allocator);
```
yewman, line 322: `errdefer entries.deinit();`?

yewman (commented):
Couple of minor comments otherwise lgtm

src/bincode/bincode.zig
dnut: 87ae4a59b84503a0adc38e2b59ccd0fbe2f59322

src/bincode/bincode.zig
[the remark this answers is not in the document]
dnut: 87ae4a59b84503a0adc38e2b59ccd0fbe2f59322

dnut: 87ae4a59b84503a0adc38e2b59ccd0fbe2f59322

src/core/transaction.zig
dnut: ec46d34e5a63dcc4218b8fa67c162e63555c0da0

src/core/transaction.zig
dnut: ec46d34e5a63dcc4218b8fa67c162e63555c0da0

src/bincode/bincode.zig
```diff
@@ -708,11 +715,15 @@ pub fn HashMapConfig(comptime hm_info: sig.utils.types.HashMapInfo) type {
...
-        const type_config = @field(struct_type, bincode_field);
-        return type_config;
+pub fn getConfig(comptime T: type) ?FieldConfig(T) {
+    const config_field_name = "!bincode-config";
+    switch (@typeInfo(T)) {
+        .Struct, .Enum, .Union, .Opaque => {
+            if (@hasDecl(T, config_field_name)) {
+                return @field(T, config_field_name);
+            }
+        },
```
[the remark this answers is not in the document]
dnut: 821d5783ecdf1a255d7f9d67492b32b4fb91324d

src/utils/io.zig
```diff
@@ -0,0 +1,52 @@
...
+
+        pub fn read(self: *Self, dest: []u8) Error!usize {
+            if (dest.len == 0) return 0;
+            if (self.next_byte) |byte| {
+                dest[0] = byte;
+                self.next_byte = null;
+                if (dest.len == 1) return 1;
+                return 1 + try self.backing_reader.read(dest[1..]);
+            }
+            return self.backing_reader.read(dest);
```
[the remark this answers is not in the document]
dnut: `if (dest.len == 1) return 1;` is necessary to avoid index out of bounds when dest.len is 1.

d619020a9eaaa2827569208536d4096c319bdd74

src/utils/types.zig
dnut: It's easy to copy and paste the code. It's just a question of whose responsibility it is to maintain compatibility.

src/ledger/insert_shred.zig
dnut: 58ad7851e9220723184c5eac0ec394dff0fcccc0

src/ledger/rocksdb.zig
dnut: Whenever I see a bool as a function parameter, I ask myself if it would be better to actually have two separate functions instead of one function with a bool. In this case, I think it would be cleaner to have two separate functions.

There is no need currently to delete the data from the disk so I didn't add the other function.

Also the recommended way to delete rocksdb from the disk is by using the `DestroyDB` function. Today I added this function to rocksdb-zig and created a new branch in sig that upgrades to use this version of the library.

src/ledger/tests.zig
dnut: c0319bfec52b84410ef3feb0c60c78a1254745c8

src/ledger/tests.zig
dnut: c0319bfec52b84410ef3feb0c60c78a1254745c8

src/ledger/tests.zig
dnut: c0319bfec52b84410ef3feb0c60c78a1254745c8

src/ledger/tests.zig
0xNineteen: ok that makes sense - thanks for explaining - we can keep it as is

src/ledger/insert_shred.zig
dnut: >  everwhere else we do it a different way

Do you mean like this?

```zig
const logger = Logger.init(allocator, .warn);
defer logger.deinit();
const some_service = SomeService.init(logger);
```

I think there's a better approach, which is the interface pattern that's used throughout std for things like Allocator, Reader, Writer, and Random.

```zig
const stderr_logger = StandardErrLogger.init(allocator, .warn);
defer stderr_logger.deinit();
const some_service = SomeService.init(stderr_logger.logger());
```

This approach is more flexible, is clearer about the underlying state, and streamlines the interface struct definition.


<br>

> global constant

I can make it a constant, but I'd like to do it in a way that makes it clear that the log level is configurable, and that the constant is only the default. I think it might be slightly confusing if it's at the top level scope of log.zig labelled as the canonical "test_logger". What do you think about this?

```zig
initWithLogger(test_name, sig.trace.TestLogger.default.logger());
```

cf6db5baece72c0a885702d11cb7952b3731273d

PR #252, opened 2024-08-26
feat(channel): replace with a faster implementation

Rexicon226:
Status: All integration is done and benchmarks are showing good signs, I will write up a description tomorrow :)

- Replace the channel with a faster, and more importantly actually correct mpmc channel.
- Remove the major concept of `PacketBatch`s in favor of sending individual packets. Since sending in the channel is (kinda) lock-free, this is viable and batching just isn't needed anymore.
- Re-organized and re-did a lot of the memory orderings in the codebase to actually be correct. `unordered` should *never* be used by us. Exits actually have the happens-before edge of acquire-release pairs now.

TODO checklist:
- [x] After some thinking, I don't really see a point to implementing this. This can easily be done as a follow up, or even better when we come back to the channel being the bottleneck. It would increase the complexity of the algorithm quite a bit.
~~Implement `initial_capacity` on the new channel by wrapping against `LAP` and linking blocks but not updating the index head. Unsure if this is possible yet with this design. I don't think it would matter much since allocation overhead is already at a minimum here.~~ 
- [x] Audit all call sites of `close` and see whether we have any races there.
- [x] This is a blocker now, there are CI failures that were missed because of us missing a message in Debug mode but not in ReleaseSafe
We do signaling inside the threads of the gossip server in order to not lose packets in the case where `sendSocket` sees `exit` before something else does, causing us to lose packets in edge-cases. The best solution here is to create a shutdown event chain between critical threads in order to ensure processing order is fullfilled.


src/sync/Backoff.zig
```diff
@@ -0,0 +1,29 @@
+const std = @import("std");
+const builtin = @import("builtin");
+const Backoff = @This();
```
0xNineteen, line 3: can we change this to use the struct { } approach we use throughout the codebase - were not following zig's approach - would be nice to also add this note to the contributing.md - will also want to change the file to lowercase (backoff.zig)

src/sync/channel.zig
```diff
@@ -166,105 +168,519 @@ pub fn Channel(comptime T: type) type {
...
-const BlockPointerChannel = Channel(*Block);
-
 const logger = std.log.scoped(.sync_channel_tests);
 
-fn testReceiver(chan: *BlockChannel, recv_count: *Atomic(usize), id: u8) void {
-    _ = id;
-    while (chan.receive()) |v| {
-        _ = v;
-        _ = recv_count.fetchAdd(1, .seq_cst);
+fn testUsizeReceiver(chan: anytype, recv_count: usize) void {
```
0xNineteen, line 468: @dnut prev found some channel bugs with the test in this commit: https://github.com/Syndica/sig/commit/bac636ecacc5ac7f21df62d276f87a64f05fdad7

its pretty simple - probs worth it to port it over or have similar assertions to make sure everything works expected

src/sync/Backoff.zig
Rexicon226: Got it!

src/sync/channel.zig
```diff
@@ -1,170 +1,426 @@
...
+                    block = channel.tail.block.load(.acquire);
+                    backoff.spin();
+                } else {
+                    if (offset + 1 == BLOCK_CAP) {
+                        const next_index = new_tail +% (1 << SHIFT);
+                        channel.tail.block.store(next_block, .release);
+                        channel.tail.index.store(next_index, .release);
+                        block.?.next.store(next_block, .release);
+                    } else if (next_block) |b| {
+                        channel.allocator.destroy(b);
```
dnut, line 149:
```suggestion
                        // In a previous iteration, we created a buffer when we thought we needed to
                        // create the next block. There was contention and another thread created it
                        // instead. The slot we've actually succeeded to acquire is no longer the 
                        // first slot in a new block, so we can destroy the buffer.
                        channel.allocator.destroy(b);
```

src/sync/channel.zig
```diff
@@ -1,170 +1,426 @@
...
+                }
+
+                const new_tail = tail + (1 << SHIFT);
+
+                if (channel.tail.index.cmpxchgWeak(tail, new_tail, .seq_cst, .acquire)) |t| {
+                    tail = t;
+                    block = channel.tail.block.load(.acquire);
+                    backoff.spin();
+                } else {
+                    if (offset + 1 == BLOCK_CAP) {
```
dnut, line 143:
```suggestion
                    // successfully acquired a slot
                    if (offset + 1 == BLOCK_CAP) {
```

src/sync/channel.zig
```diff
@@ -1,170 +1,426 @@
...
+                        tail = channel.tail.index.load(.acquire);
+                        block = channel.tail.block.load(.acquire);
+                        continue;
+                    }
+                }
+
+                const new_tail = tail + (1 << SHIFT);
+
+                if (channel.tail.index.cmpxchgWeak(tail, new_tail, .seq_cst, .acquire)) |t| {
+                    tail = t;
```
dnut, line 139:
```suggestion
                    // contention: another thread has already acquired the slot and updated the tail
                    tail = t;
```

src/sync/channel.zig
```diff
@@ -1,170 +1,426 @@
...
+                        next_block = new;
+                        tail = channel.tail.index.load(.acquire);
+                        block = channel.tail.block.load(.acquire);
+                        continue;
+                    }
+                }
+
+                const new_tail = tail + (1 << SHIFT);
+
+                if (channel.tail.index.cmpxchgWeak(tail, new_tail, .seq_cst, .acquire)) |t| {
```
dnut, line 138:
```suggestion
                // try to increment the tail by one to acquire an exclusive write lock to that slot
                if (channel.tail.index.cmpxchgWeak(tail, new_tail, .seq_cst, .acquire)) |t| {
```

src/sync/channel.zig
```diff
@@ -1,170 +1,426 @@
...
+                    tail = t;
+                    block = channel.tail.block.load(.acquire);
+                    backoff.spin();
+                } else {
+                    if (offset + 1 == BLOCK_CAP) {
+                        const next_index = new_tail +% (1 << SHIFT);
```
dnut, line 144:
```suggestion
                        // The acquired slot is in a new block. We already eagerly initialized
                        // that block. Now we need to update the state to point to it.
                        const next_index = new_tail +% (1 << SHIFT);
```

src/sync/channel.zig
```diff
@@ -1,170 +1,426 @@
...
+                    tail = channel.tail.index.load(.acquire);
+                    block = channel.tail.block.load(.acquire);
+                    continue;
+                }
+
+                if (offset + 1 == BLOCK_CAP and next_block == null) {
+                    next_block = try Buffer.create(channel.allocator);
+                }
+
+                if (block == null) {
```
dnut, line 122:
```suggestion
                // Check if this is the first send to see if we should create the first block.
                if (block == null) {
```

src/sync/channel.zig
```diff
@@ -1,170 +1,426 @@
...
+                    const new = try Buffer.create(channel.allocator);
```
dnut, line 123: Why not eagerly allocate the first block in Channel.init? Then this won't need to be nullable, and you can remove the null-handling logic from send and recv. It feels like a simplification that would help the readability/maintainability of the code.

src/sync/channel.zig
```diff
@@ -1,170 +1,426 @@
...
+        const READ: usize = 0b10;
+        const DESTROY: usize = 0b100;
+
+        const HAS_NEXT: usize = 0b01;
+
+        const Position = struct {
+            index: Atomic(usize),
+            block: Atomic(?*Buffer),
+
+            fn init() Position {
```
dnut, line 30: Is this function necessary? Do you think default field values are harmful?

src/sync/channel.zig
```diff
@@ -1,170 +1,426 @@
...
+                const offset = (tail >> SHIFT) % LAP;
+                if (offset == BLOCK_CAP) {
+                    backoff.snooze();
+                    tail = channel.tail.index.load(.acquire);
+                    block = channel.tail.block.load(.acquire);
+                    continue;
+                }
+
+                if (offset + 1 == BLOCK_CAP and next_block == null) {
+                    next_block = try Buffer.create(channel.allocator);
```
dnut, line 119: This allocation will be unnecessary if there is contention on the tail and another thread creates the next buffer and updates the tail before we do. Ideally we would only create this if we really know 100% that the current thread needs to provide the next block. But waiting on an allocation while holding an exclusive write lock seems even worse. So maybe this is the best way.

src/sync/channel.zig
```diff
@@ -1,170 +1,426 @@
...
         const Self = @This();
+        const BLOCK_CAP = 31;
+        const SHIFT = 1;
+        const LAP = 32;
 
-        pub fn init(allocator: std.mem.Allocator, init_capacity: usize) *Self {
-            const self = allocator.create(Self) catch unreachable;
-            self.* = .{
-                .buffer = Mux(std.ArrayList(T)).init(std.ArrayList(T).initCapacity(allocator, init_capacity) catch unreachable),
+        const WRITE: usize = 0b01;
```
dnut, line 20: This name is confusing. The name seems to imply there is currently a write in progress. But really this flag indicates that a write has already completed, and it is occupied with data that is safe to read. I think "WRITTEN", "OCCUPIED", or "READABLE" would be a better name.

src/sync/channel.zig
```diff
@@ -1,170 +1,426 @@
...
+        const READ: usize = 0b10;
```
dnut, line 21: Same as above. I might call this something like "FINISHED_READING"

dnut (changes requested):
Overall looks great but I have a few questions/concerns.

Also I suggested several comments. You don't need to take my suggestions if you don't like the wording, but code like this can require a bit of thinking to figure out the purpose. Would be nice to have a little explanation about the general idea for all the branches. I mainly commented on send but would like to see comments in recv as well.

src/sync/backoff.zig
```diff
@@ -0,0 +1,29 @@
...
+
+    const SPIN_LIMIT = 6;
+
+    pub fn init() Backoff {
+        return .{ .step = 0 };
+    }
+
+    pub fn snooze(_: *Backoff) void {
+        switch (builtin.cpu.arch) {
+            .aarch64 => asm volatile ("wfe" ::: "memory"),
```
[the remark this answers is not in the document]
Rexicon226: `wfe` is better than `isb` for snoozing because it waits longer. Ideally it waits until the thread's quota is up until waking up.

src/sync/backoff.zig
```diff
@@ -0,0 +1,29 @@
+const std = @import("std");
+const builtin = @import("builtin");
+
+pub const Backoff = struct {
+    step: u32,
+
+    const SPIN_LIMIT = 6;
+
+    pub fn init() Backoff {
+        return .{ .step = 0 };
```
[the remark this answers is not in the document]
Rexicon226: No reason, good idea!

src/sync/channel.zig
Rexicon226: I added logic for creating the first block in the `init` function, good idea.

src/sync/channel.zig
Rexicon226: Not at all, removed! It used to do something else before internal refactor.

src/sync/channel.zig
Rexicon226: I added a comment explaining why this allocation should happen before the CAS and not after.

src/sync/channel.zig
Rexicon226: Renamed to `WRITTEN_TO`.

src/sync/channel.zig
Rexicon226: Renamed to `READ_FROM`.

Rexicon226:
Fixed up the review, thanks a lot!

src/gossip/service.zig
```diff
@@ -497,62 +518,26 @@ pub const GossipService = struct {
...
-            msg_count += messages.len;
-
-            for (messages) |*message| {
+        // keep waiting for new data until,
+        // - `exit` isn't set,
+        // - there isn't any data to process in the input channel, in order to block the join until we've finished
+        while (!(self.exit.load(.acquire) and self.verified_incoming_channel.len() == 0)) {
+            var msg_count: usize = 0;
+            while (self.verified_incoming_channel.receive()) |message| {
+                msg_count += 1;
```
0xNineteen, line 530: can we have a max msg_count in this loop so it wont get get stuck draining the channel

src/gossip/service.zig
```diff
@@ -1528,7 +1533,7 @@ pub const GossipService = struct {
 
         // pre-allocate memory to track insertion failures
         var max_inserts_per_push: usize = 0;
-        for (batch_push_messages.items) |*push_message| {
+        for (batch_push_messages.items) |push_message| {
```
0xNineteen, line 1536: why dont we want a pointer here? isnt the no-pointer approach an unnecessary stack-copy

src/gossip/service.zig
```diff
@@ -1882,36 +1880,17 @@ pub const GossipService = struct {
...
                 else => {
                     // only allow values where both the sender and origin match our shred version
-                    if (!sender_matches or
-                        !gossip_table.checkMatchingShredVersion(gossip_value.id(), my_shred_version))
-                    {
-                        const removed_value = gossip_values_array.swapRemove(i);
-                        bincode.free(self.gossip_value_allocator, removed_value);
-                        continue; // do not incrememnt `i`. it has a new value we need to inspect.
+                    if (!gossip_table.checkMatchingShredVersion(gossip_value.id(), my_shred_version)) {
+                        continue; // do not increment `i`, this wasn't valid.
```
0xNineteen, line 1884: dont think this logic is correct anymore - we still need to swapRemove to make sure the values we insert have a matching shred version

src/sync/channel.zig
```diff
@@ -1,170 +1,455 @@
...
+            state: Atomic(usize),
+
+            const uninit: Slot = .{
+                .value = undefined,
+                .state = Atomic(usize).init(0),
+            };
+        };
+
+        pub fn init(allocator: Allocator, initial_capacity: usize) !Self {
+            _ = initial_capacity; // TODO: do something with this
```
0xNineteen, line 82: do we want to fix this todo?

src/sync/channel.zig
```diff
@@ -1,170 +1,455 @@
...
+                .block = Atomic(?*Buffer).init(first_block),
+            };
+            return .{
+                .head = first_position,
+                .tail = first_position,
+                .allocator = allocator,
+            };
+        }
+
+        pub fn create(allocator: Allocator, initial_capacity: usize) !*Self {
```
0xNineteen, line 95: great pattern - `create` vs `init`

0xNineteen:
also generally curious - why do you want to make deinit() calls take a const pointer? - it works but doesnt make sense to me, were destroying state, so it should be mutable

docs/CONTRIBUTING.md
```diff
@@ -10,6 +10,13 @@ See the [readme](../readme.md#-setup).
 
 ## Style Guide
 
+### Identifiers
+
+We follow the [Zig Style](https://ziglang.org/documentation/master/#Style-Guide) guide as is reasonable.
+
+- Do not use identifier names that start with a `_`, i.e (`_foo` or `__bar`). These are the bane of all codebases and the product 
+of not enough time spent thinking about naming. Every decl is unique in some way, use that to your advantage when naming.
```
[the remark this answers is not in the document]
Rexicon226: Yep, will do.

src/gossip/service.zig
```diff
@@ -347,45 +377,46 @@ pub const GossipService = struct {
     const VerifyMessageTask = ThreadPoolTask(VerifyMessageEntry);
     const VerifyMessageEntry = struct {
         gossip_value_allocator: std.mem.Allocator,
-        packet_batch: ArrayList(Packet),
+        packet: Packet,
         verified_incoming_channel: *Channel(GossipMessageWithEndpoint),
         logger: Logger,
 
+        var x: u32 = 0;
```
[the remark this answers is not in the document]
Rexicon226: To be honest, I have literally no idea. I have no recollection of writing such code. I guess removing it until I remember what I was thinking.

src/gossip/service.zig
Rexicon226: Fair point!

src/gossip/service.zig
Rexicon226: There will be no stack copy here, the optimizer can easily see that `batch_push_messages.items` or `push_message`  is not modified inside the loop and will index into `batch_push_messages.items`.

The only cases where you don't modify the value of the capture, but need a pointer would be either
- You actually require the pointer itself.
- You're modifying the location where that pointer is inside of the loop, meaning `push_message` loads cannot be memorized as easily.

src/gossip/service.zig
Rexicon226: You may be right, I will look into this.

src/sync/channel.zig
Rexicon226: Hmm, I think I'd like to leave it for now. It's not necessary for any real performance reasons for now, however, it might be useful to add if we realize that the channel is slowing us down at some point.

src/sync/channel.zig
Rexicon226: Agreed. `create` should allocate and return a pointer to `@This()`, `init` should return `@This()` as a value.

Rexicon226:
> also generally curious - why do you want to make deinit() calls take a const pointer? - it works but doesnt make sense to me, were destroying state, so it should be mutable

I don't see how state has any requirement to be mutable. Take for example:
```zig
const list = std.ArrayList(u8).initCapacity(allocator, 20);
```

The `ArrayList(u8)` value exists on the stack, containing the length, slice, and allocator reference. It will become invalid and the memory re-used when the frame returns. The memory that was created on the heap is owned by us, so we need to call `deinit` to free it. No part of that process required `list` to be mutable.

src/sync/channel.zig
0xNineteen: ok - maybe we just remove the parameter for now and add a comment saying we could add support for initial capacity if performance is slow

src/gossip/service.zig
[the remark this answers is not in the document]
0xNineteen: lmfao 😆

dnut:
> also generally curious - why do you want to make deinit() calls take a const pointer? - it works but doesnt make sense to me, were destroying state, so it should be mutable

> I don't see how state has any requirement to be mutable.

If it works with a const pointer, why not just pass self by value?

It's hard for me to imagine why a typical deinit function actually needs a pointer to self at all. Accepting self by value in most cases seems that it should be fine. Calling deinit implies that the item in the calling scope is no longer usable. So, it's not meaningful to mutate the item within the calling scope.

The only exception I can think of is when `init` allocated the pointer to self, and self was only ever passed around as the exact same pointer. Likewise, `deinit` needs to free that pointer. In this case, it may be better to use the names "create" and "destroy" to make those semantics more obvious.

HashMap is a curious example from std that takes a pointer to self in its deinit function. It imposes a seemingly pointless restriction that can easily be worked around. The main difference I see is that it replaces invalidated pointers with `undefined` and zeros some other fields. My guess is that the purpose of this is to make memory bugs easier to identify. For example, misusing the map is more likely to result in a segfault (fail fast) instead of corrupting memory that has been recycled.

Rexicon226:
> If it works with a const pointer, why not just pass self by value?

No reason! This is just a preference, to be honest. I like using `*const T` more than `T`, just from an inherent fear of an implicit pass-by-ref foot gun. Tigerbeetle does the same thing if I remember.

> HashMap is a curious example from std that takes a pointer to self in its deinit function.

Yeah, this was debated as an inconsistency. The main argument was that if you have a hashmap onto which you're putting, it'd already be mutable, so having the `map.* = undefined` is fine in the `deinit`.

dnut:
> like using `*const T` more than `T`, just from an inherent fear of an implicit pass-by-ref foot gun. Tigerbeetle does the same thing if I remember.

Could you elaborate on the footgun?

My understanding was that it's usually better to pass by value instead of const pointer because it allows the compiler to make the optimal choice. If it would actually be faster to pass by reference, it would compile to code that passes by reference, and that wouldn't actually change the behavior since parameters are immutable. But if you write the code to pass a pointer, then the compiled binary will pass a pointer regardless of what is optimal. Is this right?

Rexicon226:
> My understanding was that it's usually better to pass by value instead of const pointer because it allows the compiler to make the optimal choice. 

This is true, however it's only optimal in specific cases. Nothing we're doing requires this optimization and I believe it's best for us to follow the rule of "if it's larger than `[2]usize` and we don't explicitly need a copy, use `*const`". This prevents *other* bugs that can happen, where an implicit copy wasn't intended.

dnut:
> This prevents _other_ bugs that can happen, where an implicit copy wasn't intended.

I'm very curious about this because I'm drawing a blank trying to think of an example that could happen in zig. Can you describe an example for me please?

Rexicon226:
> > This prevents _other_ bugs that can happen, where an implicit copy wasn't intended.
> 
> I'm very curious about this because I'm drawing a blank trying to think of an example that could happen in zig. Can you describe an example for me please?

I've only heard this second hand, but it *is* possible to hit stack overflows when passing unexpectedly large values by value in an argument. This is what happened to tigerbeetle, and why they chose to add using explicitly `*const T` syntax to tiger style. Again, it's a mostly unnecessary precaution, but I think it makes it clearer that you're not making a copy and *could* prevent bugs.

src/gossip/service.zig
Rexicon226: So yeah, I mistook the purpose of the code originally. I've now implemented a much more efficient Hoare's partitioning scheme to sort these. It's `O(N)` instead of `O(N^2)` like the other one was.

0xNineteen (approved):
~overall lgtm - if you think its good to go, would be great to get it on main~

src/gossip/service.zig
```diff
@@ -1856,62 +1888,102 @@ pub const GossipService = struct {
...
-        from_pubkey: Pubkey,
+        sender_pubkey: Pubkey,
     ) usize {
-        // we use swap remove which just reorders the array
-        // (order dm), so we just track the new len -- ie, no allocations/frees
-        const my_shred_version = self.my_shred_version.load(.monotonic);
-        if (my_shred_version == 0) {
-            return gossip_values.len;
-        }
+        const S = struct {
```
0xNineteen, line 1899: would ideally want this logic to be as simple as possible unless we need it faster - can we return to how it was originally wrote? - running this on testnet and the table gets zero values inserted (guessing somethings going wrong here)

0xNineteen (changes requested):

src/gossip/service.zig
Rexicon226: interesting. I am quite against putting back the old logic since its time complexity is large for no reason. can you instead share how you tested it and I'll just fix the issue?

0xNineteen: makes sense - time complexity is important but only when the code is a hot path and needs optimization (not convinced this code needs it) - our first priority is simplicity and performance second which is why i prefer the original approach

would prefer to keep this pr changes restricted to the channel too - happy to discuss further in a separate pr

run it on testnet and youll see values arent inserted

Rexicon226: > makes sense - time complexity is important but only when the code is a hot path and needs optimization (not convinced this code needs it) - our first priority is simplicity and performance second which is why i prefer the original approach

yeah good point. i will revert it and look into it later.

Rexicon226: done: 45915bc69ff3f713167694db02a03127d7dc4e83

Rexicon226:
I've been looking over the code again, just to double-check everything, and I'm seeing some weird behavior in the gossip packet processing. Give me a couple of hours to figure this out, I'll update this message when I fix whatever this is.

Rexicon226:
After running on testnet and devnet for a bit, I'm still hitting double frees! Time to untangle some more memory ownership and then fuzz overnight.

src/accountsdb/db.zig
```diff
@@ -325,7 +326,7 @@ pub const AccountsDB = struct {
         const load_duration = try self.loadFromSnapshot(
             snapshot_fields.accounts_db_fields,
             n_threads,
-            std.heap.page_allocator,
+            allocator,
```
0xNineteen, line 329: we want the page_alloc here because we want a thread-safe and lock-free allocations for fast parallel loading

maybe add a comment to the method parameter mentioning this if were not hard coding it anymore

Rexicon226: the page_allocator is far from lock-free under contention. memory page pool contention is a large issue that happens when 1. concurrently page allocation and 2. concurrently mmaping. using the provided allocator, that hopefully being the c allocator, will *greatly* increase the speed.

0xNineteen: nice - it would still be good to include a comment with the given context

Rexicon226: totally agree with the comment, done

0xNineteen (approved):
lgtm

src/shred_collector/shred_receiver.zig
```diff
@@ -80,76 +80,69 @@ pub const ShredReceiver = struct {
...
         reply_packet.addr = packet.addr;
+        return reply_packet;
     }
 };
 
 fn shouldDiscardShred(
     packet: *const Packet,
     root: Slot,
-    shred_version: u16,
+    shred_version: *const Atomic(u16),
```
dnut, line 145: Another advantage of the packet arrays is that you can reduce the number of atomic loads of this value.

src/shred_collector/shred_receiver.zig
```diff
@@ -80,76 +80,69 @@ pub const ShredReceiver = struct {
...
+            if (shouldDiscardShred(packet, self.root_slot, self.shred_version, max_slot)) {
                 packet.flags.set(.discard);
             }
         }
     }
 
-    /// Handle a ping message and return
-    fn handlePing(self: *Self, packet: *const Packet, responses: *ArrayList(Packet)) !void {
+    /// Handle a ping message and returns the repair message.
+    fn handlePing(self: *const Self, packet: *const Packet) !?Packet {
```
dnut, line 122: Do you expect inlining or RVO to guarantee that the packet will never be copied to the calling context?

src/shred_collector/shred_verifier.zig
```diff
@@ -15,30 +14,23 @@ const Packet = sig.net.Packet;
...
             // TODO parallelize this once it's actually verifying signatures
-            for (packet_batch.items) |*packet| {
-                if (!verifyShred(packet, leader_schedule)) {
-                    packet.flags.set(.discard);
-                } else {
-                    verified_count += 1;
-                }
+            var our_packet = packet;
+            if (!verifyShred(&packet, leader_schedule)) {
+                our_packet.flags.set(.discard);
```
dnut, line 28: Without using packet batches, this line could just be `continue`

dnut (changes requested):

src/ledger/insert_shred.zig
```diff
@@ -462,7 +462,7 @@ pub const ShredInserter = struct {
                 return false;
             }
 
-            if (!shouldInsertCodeShred(&shred, self.max_root.load(.unordered))) {
+            if (!shouldInsertCodeShred(&shred, self.max_root.load(.acquire))) {
```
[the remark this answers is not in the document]
Rexicon226: Why use `unordered`? Understand that `unordered` loads have no memory order and the value will be "unspecified" during a data race. This is not a valid solution for this use case.

Here's a basic example that shows the difference: https://godbolt.org/z/YnYxPd8sv. LLVM is allowed to memoize and elide unordered loads.

src/shred_collector/repair_service.zig
```diff
@@ -101,7 +98,7 @@ pub const RepairService = struct {
         var waiting_for_peers = false;
         var timer = try std.time.Timer.start();
         var last_iteration: u64 = 0;
-        while (!self.exit.load(.unordered)) {
+        while (!self.exit.load(.acquire)) {
```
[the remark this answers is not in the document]
Rexicon226: Why shouldn't the loop's contents be executed when `exit` is true? The function is run in a thread, which is joined before deinitialization happens. The join blocks the thread that's deiniting, waiting for the loop to see the `exit` flag and exit. 

It's necessary for the `exit` to be an `acquire` load in order to see modifications the parent thread does before releasing the exit flag being set.

src/shred_collector/shred_receiver.zig
Rexicon226: Doing more atomic loads is not something we should be worrying about. Not only are they the cheapest atomic memory operation, but on architectures like x86 it's the default order of all loads.

src/shred_collector/shred_receiver.zig
```diff
@@ -80,76 +80,69 @@ pub const ShredReceiver = struct {
     /// Returns when exit is set to true.
     fn runPacketHandler(
         self: *Self,
-        receivers: []const *Channel(ArrayList(Packet)),
-        response_sender: *Channel(ArrayList(Packet)),
+        receivers: []const *Channel(Packet),
+        response_sender: *Channel(Packet),
```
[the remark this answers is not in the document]
Rexicon226: The `ArrayList` of packets doesn't make much sense, since the channel already stores its items in a managed state. When we "send" a packet into the channel you can think of it as equivalent to appending to an `ArrayList`. The only difference is that we're using less memory by not storing the extra array list metadata in every item we send.

src/shred_collector/shred_receiver.zig
```diff
@@ -80,76 +80,69 @@ pub const ShredReceiver = struct {
...
     ) !void {
         if (packet.size == REPAIR_RESPONSE_SERIALIZED_PING_BYTES) {
-            try self.handlePing(packet, responses);
-            packet.flags.set(.discard);
+            if (try self.handlePing(packet)) |p| try responses.append(p);
         } else {
             const max_slot = std.math.maxInt(Slot); // TODO agave uses BankForks for this
-            if (shouldDiscardShred(packet, self.root_slot, shred_version, max_slot)) {
+            if (shouldDiscardShred(packet, self.root_slot, self.shred_version, max_slot)) {
                 packet.flags.set(.discard);
```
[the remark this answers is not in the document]
Rexicon226: Makes sense!

src/shred_collector/shred_verifier.zig
Rexicon226: Gotcha.

src/ledger/insert_shred.zig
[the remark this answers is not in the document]
dnut: > Why use unordered?

Because it is more efficient. I would use it any time there is no need for a particular ordering of operations within the current thread on the value in question.

> This is not a valid solution for this use case.

How is it invalid? I don't see why we care about a data race here. The atomic is not being used for synchronization. It just needs to get a value that max_root has been set to. It's OK to read a slightly stale value.

src/shred_collector/repair_service.zig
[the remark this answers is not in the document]
dnut: > It's necessary for the exit to be an acquire load in order to see modifications the parent thread does before releasing the exit flag being set.

I don't understand what you mean by "releasing the exit flag being set."

All that the "acquire" does is make sure that the atomic operation completes before the current thread executes the lines of code below the load.

Actually since this code branches based on the output of the load, I believe acquire semantics are already implied. I expect there is no difference between unordered and acquire in this case.

src/shred_collector/shred_receiver.zig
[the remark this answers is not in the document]
dnut: > The only difference is that we're using less memory by not storing the extra array list metadata in every item we send.

There is another difference which I explained in my first comment. The arraylist approach reduces the amount of memory that needs to be copied.

Let's say we just received 100 packets. We have two options:

### The "arraylist" approach

This doesn't necessarily need to be an arraylist. I'll refer to it as a "pointer" to be more generic.

socket reader thread:
1. Allocate memory large enough to hold 100 packets
2. Write each packet into that region of memory
3. Write the pointer into a channel

shred receiver thread:

4. Read the pointer from the channel
5. Basic validation of each packet
6. Write the pointer into a channel

shred verifier thread:

7. Read the pointer from the channel
8. Verify the thread of each signature
9. Write the pointer into a channel

shred processor thread:

10. Read the pointer from the channel
11. Write each shreds into the blockstore

With this approach, steps 2 and 11 each copy 120 KB. The other steps all copy very little data in comparison.


### The pass-each-packet-by-value approach

<skip steps 1-2>

shred verifier thread:

3. Write each packet into the channel

shred receiver thread

4. Read each packet from the channel
5. Basic validation of each packet
6. Write the pointer to the buffer into a channel

shred verifier thread:

7. Read the pointer from the channel
8. Verify the thread of each signature
9. Write the pointer to the buffer into a channel

shred processor thread:

10. Read the pointer from the channel
11. Write the shreds into the blockstore

With this approach, steps 3, 4, 6, 7, 9, 10, and 11 each copy 120 KB. At best with copy elision, you might see something like only steps 3, 6, 9, and 11 each copy 120 KB. So the amount of data copied is 2-3.5 times as much as the arraylist approach.

src/shred_collector/repair_service.zig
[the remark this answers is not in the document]
Rexicon226: > All that the "acquire" does is make sure that the atomic operation completes before the current thread executes the lines of code below the load.

This is, not what `acquire` means. A memory operation with `acquire` semantics creates a relationship with a pair `release`, on the same memory object. 

This means that if thread A does a release store to memory object M, and thread B sees that store using an acquire memory operation, in this specific case that would be the acquire load, it is guaranteed that all memory modifications done by thread A before that release store will be visible to thread B after the acquire load. 

This is critical to our exit flag, since it removes the possibility of some modification done by thread A being missed by thread B after it sees the exit flag "released", or set.


> Actually since this code branches based on the output of the load, I believe acquire semantics are already implied. I expect there is no difference between unordered and acquire in this case.

It is not implied anywhere, the path the code goes in has nothing to do with the memory operations. Atomics deals with the guaranteed ordering of visibility of memory between and inside of threads.

Please see my above godbolt link that shows exactly why an `unordered` load here would create an infinite loop, as the compiler will move the load outside of the while loop and it will never be updated.

src/ledger/insert_shred.zig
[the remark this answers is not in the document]
Rexicon226: > Because it is more efficient. 

It's not. There will be no difference at all.

> I would use it any time there is no need for a particular ordering of operations within the current thread on the value in question.

This is what `monotonic` does. Think of `unordered` as telling the compiler that that we don't care about the value read and only care that there is no "technical" data race. It changes the value read from undefined (any non-atomic operation on memory that was atomically operated on is considered UB), to unspecified (where there is nothing that dictates what the value will be). 

> How is it invalid? I don't see why we care about a data race here. The atomic is not being used for synchronization. 

By definition, atomics cannot race. However, we care about the validity of the value under a data-race. Understand that a data-race in this context is not a bug, but a contention between two threads on the same memory object. 

As an absolute rule, `unordered` should never be used in our codebase unless it's very clear why. It's a smell to use, it makes any code around it very suspect, and it is not more efficient in any practical sense.

src/shred_collector/shred_receiver.zig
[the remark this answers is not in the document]
Rexicon226: While I can't disagree that there will be more copies,

1. These copies are tens of times faster than the heap allocation that the array list would be doing.
2. We have much more important optimization to do before we need to worry about doing too many copies.

It may seem like copying the packets is bad for performance, however, when the alternative is our current excessive usage of array lists and heap allocation, it wins.

docs/CONTRIBUTING.md
```diff
@@ -10,6 +10,17 @@ See the [readme](../readme.md#-setup).
 
 ## Style Guide
 
+### Identifiers
+
+We follow the [Zig Style](https://ziglang.org/documentation/master/#Style-Guide) guide as is reasonable.
+
+- Do not use identifier names that start with a `_`, i.e (`_foo` or `__bar`). These are the bane of all codebases and the product
```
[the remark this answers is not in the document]
0xNineteen: i missed this - i agree underscore prefixes in the correct context are ok; not enough to be permabanned

src/shred_collector/repair_service.zig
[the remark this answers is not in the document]
dnut: > This means that if thread A does a release store to memory object M, and thread B sees that store using an acquire memory operation, in this specific case that would be the acquire load, it is guaranteed that all memory modifications done by thread A before that release store will be visible to thread B after the acquire load.

This is consistent with my understanding. This is basically what I was getting at here:

> > All that the "acquire" does is make sure that the atomic operation completes before the current thread executes the lines of code below the load.
> 
> This is, not what acquire means.

It is what acquire means in the context of the thread executing the acquire operation, and it's compatible with your definition. I just didn't describe the full context of how it relates to a release, because I was only talking about the operations in the current thread that we're looking at now.

> This is critical to our exit flag, since it removes the possibility of some modification done by thread A being missed by thread B after it sees the exit flag "released", or set.

In the thread that sets exit to true, are there operations that need to happen before any other threads exit? I'm simply not aware of any scenarios where this is true, so I'm curious to hear about them.

In my previous comment, I was trying to imagine some scenarios where the ordering might matter. It was very generic/hypothetical, so I'm not sure if it applies, but the general idea was this:

Let's say thread A is going to trigger an exit with a store, and then thread B is going to load the exit variable and then exit after seeing it is true. If there are operations in thread A which need to complete before thread B exits, this can be handled by release/acquire semantics. But you might also want to ensure that some operations in thread B complete before thread A exits. This seems like it is potentially equally important, since exits can be triggered from anywhere, and it's not helped by the simple release/acquire ordering on exit store/loads.

> Please see my above godbolt link that shows exactly why an `unordered` load here would create an infinite loop, as the compiler will move the load outside of the while loop and it will never be updated.

This is a good argument against using unordered. I expect that monotonic would solve this specific problem. But actually it would be a more convincing argument if the code branched based on the output of the load, since that's how our code works. Something like this: https://godbolt.org/z/7e5vT3rW4. Since `cmp dword ptr [rdi], 0` is outside of `.LBB0_1:`, I think this confirms your argument about the infinite loop.

> > Actually since this code branches based on the output of the load, I believe acquire semantics are already implied. I expect there is no difference between unordered and acquire in this case.
>
> It is not implied anywhere, the path the code goes in has nothing to do with the memory operations. Atomics deals with the guaranteed ordering of visibility of memory between and inside of threads.

For some reason, I thought that the compiler would automatically use at least acquire ordering when the load is directly used in conditional branching logic, regardless of what the user specifies. But I think the godbolt example that I just shared proves me wrong. I looked around and could only find this on the topic: 

>When you use atomic variable directly in `if` or `switch` expression, implicit conversion happens, which is equivalent to `.load()` without parameters, and the memory order is implicitly `memory_order_seq_cst`. It is the safest memory order, so will not introduce any new issues.
>
>https://stackoverflow.com/questions/78298260/is-branching-on-a-stdatomic-variable-a-possible-source-of-bugs. 

Here they actually suggest sequential consistency is used, but again, this contradicts the example in godbolt, so I guess they are wrong.

src/shred_collector/shred_receiver.zig
Rexicon226: This will “RVO” not just because of Zig, but because of every real calling convention ever passing non-floats larger than 2 * usize through a return pointer.

As mentioned above, this doing extra copies due to missed elisions is fine, and much better/faster than the alternative of heap allocating.

A more technical argument is that the function will be a bit deeper in the stack than the caller, meaning a new cache set will be used for storing the out buffer. This means no cache pressure is put on the calling function.

docs/CONTRIBUTING.md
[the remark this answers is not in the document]
Rexicon226: yeah fair enough. my bad for not highlighting this as important to review. i also phrased it a bit harshly, it was late.

no strong argument from me, if you guys want to allow them for something like private fields, that’s reasonable.

src/shred_collector/repair_service.zig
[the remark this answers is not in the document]
Rexicon226: Sorry, limited to my phone for a couple of hours, so I’ll respond to the easier part. We can discuss the thread part later in our sync.

> For some reason, I thought that the compiler would automatically use at least acquire ordering when the load is directly used in conditional branching logic, regardless of what the user specifies.
…
Here they actually suggest sequential consistency is used, but again, this contradicts the example in godbolt, so I guess they are wrong. Maybe it's specific to C++.

This is specifically a C++ thing. `std::atomic` is a wrapper for a type, which like the more familiar `Arc<T>` in Rust, forwards use/deref to the inner type. In the C++ case, the use *must* load, so it does with `seq_cst` to ensure total order for edge case types.


Thankfully we don’t have terrible things like implicit total order in Zig.

src/shred_collector/shred_receiver.zig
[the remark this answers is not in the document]
dnut: Fair point about the allocations. One way to deal with this is by recycling them.

The shred collector had sufficiently good performance before this change to process many more packets than mainnet produces. Obviously we'd like this to be as fast as possible. I assume that you have improved the performance with this change. But since there's a tradeoff, it's hard to be sure without some tests or benchmarks to demonstrate it.

src/ledger/insert_shred.zig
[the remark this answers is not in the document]
dnut: > It's not. There will be no difference at all.

What is the point of having unordered at all, if monotonic always has the same exact performance characteristics? Why would anyone ever use unordered? In another comment you gave examples of optimizations that may be performed by the compiler for unordered which would not be performed for monotonic. Do those optimizations not make the code more efficient?

> > I would use it any time there is no need for a particular ordering of operations within the current thread on the value in question.
> 
> This is what monotonic does. 

Not exactly. Monotonic ordering guarantees that atomic operations on the same value in the same thread are executed in the correct order relative to each other. So there is a minimal amount of ordering that is guaranteed by monotonic.

> As an absolute rule, unordered should never be used in our codebase unless it's very clear why

I'm fine with this. I just wanted to understand the reasoning a bit better, and your godbolt example about eliding the loads out of the loop is satisfying.

src/shred_collector/shred_receiver.zig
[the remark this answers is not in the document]
dnut: I'm satisfied with the allocation argument, so I would only want it to pass a basic sanity check in terms of performance. The shred collector has never been formally benchmarked. The criteria for "good enough" so far has only been "Can I run it on my t3.xlarge instance in ec2, and it catches up to mainnet from at least 100 slots behind, and doesn't fall behind after running for a few minutes?"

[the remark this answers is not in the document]
Rexicon226: I'm happy to do all of these, could you explain the steps? I agree with @0xNineteen that we should try to get this PR through soon, it's quite large and fixes a lot of bugs.


I'm also happy to do formal benchmarking of the shred collector, but after I get through the snapshot benchmarking, continuous benchmarking with the ec2, and getting telemetry for historical data. Just have a bit on the plate already :).

src/shred_collector/shred_verifier.zig
Rexicon226: b47ab63

src/shred_collector/shred_receiver.zig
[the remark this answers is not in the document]
Rexicon226: b47ab63

docs/CONTRIBUTING.md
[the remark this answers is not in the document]
Rexicon226: b47ab63, reworded it. how does it sound?

src/shred_collector/repair_service.zig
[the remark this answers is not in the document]
dnut: Commenting for posterity, to close the loop regarding my main concern:

> In the thread that sets exit to true, are there any specific modifications that actually need to happen before any of the other threads exit? I'm simply not aware of any scenarios where this is true, so I'm curious to hear about them.

No, there are no known scenarios where we actually need this. But it could conceivably be useful in the future, and there is no downside to including it at this point.

dnut:
I'm seeing some errors testing this branch on mainnet:


```
zig build run -Doptimize=ReleaseSafe -- -l info shred-collector --test-repair-for-slot $(solana -um slot) --leader-schedule -- --entrypoint entrypoint.mainnet-beta.solana.com:8001 --entrypoint entrypoint2.mainnet-beta.solana.com:8001 --entrypoint entrypoint3.mainnet-beta.solana.com:8001 --entrypoint entrypoint4.mainnet-beta.solana.com:8001 --entrypoint entrypoint5.mainnet-beta.solana.com:8001

time=2024-09-26T15:06:29Z level=info  metrics port: 12345
time=2024-09-26T15:06:29Z level=info  entrypoints: { 34.83.231.102:8001, 147.28.180.247:8001, 145.40.64.255:8001, 147.28.134.81:8001, 147.28.198.87:8001 }
time=2024-09-26T15:06:30Z level=info  shred version: 50093 - from entrypoint ip echo: 34.83.231.102:8001
time=2024-09-26T15:06:30Z level=info  my ip: 18.221.215.219
time=2024-09-26T15:06:30Z level=info  gossip host: 18.221.215.219
time=2024-09-26T15:06:30Z level=info  gossip port: 8001
time=2024-09-26T15:06:30Z level=info  Starting gossip readSocket
time=2024-09-26T15:06:30Z level=info  Starting gossip verifyPackets
time=2024-09-26T15:06:30Z level=info  Starting gossip processMessages
time=2024-09-26T15:06:30Z level=info  Starting gossip buildMessages
time=2024-09-26T15:06:30Z level=info  Starting gossip sendSocket
time=2024-09-26T15:06:31Z level=info  gossip: recv 1: 1 ping, 0 pong, 0 push, 0 pull request, 0 pull response, 0 prune
time=2024-09-26T15:06:31Z level=info  gossip: sent: 0 ping, 0 pong, 0 push, 0 pull request, 0 pull response, 0 prune
time=2024-09-26T15:06:41Z level=info  gossip: recv 14: 14 ping, 0 pong, 0 push, 0 pull request, 0 pull response, 0 prune
time=2024-09-26T15:06:41Z level=info  gossip: sent: 0 ping, 14 pong, 0 push, 2 pull request, 0 pull response, 0 prune
time=2024-09-26T15:06:51Z level=info  gossip: recv 27882: 3906 ping, 1402 pong, 16356 push, 2706 pull request, 3503 pull response, 1 prune
time=2024-09-26T15:06:51Z level=info  gossip: sent: 1645 ping, 3907 pong, 0 push, 3 pull request, 22 pull response, 70 prune
thread 29657 panic: switch on corrupt value
/home/ubuntu/sig/src/gossip/service.zig:0:0: 0x159ff7f in buildPushMessages (sig)
/home/ubuntu/sig/src/gossip/service.zig:844:62: 0x15a2e89 in runService__anon_48039 (sig)
            const maybe_push_packets = self.buildPushMessages(&push_cursor) catch |e| blk: {
                                                             ^
/home/ubuntu/zig-linux-x86_64-0.13.0/lib/std/Thread.zig:429:13: 0x14e5f7b in entryFn (sig)
            @call(.auto, f, args) catch |err| {
            ^
???:?:?: 0x77c985494ac2 in ??? (libc.so.6)
Unwind information for `libc.so.6:0x77c985494ac2` was not available, trace may be incomplete

run
└─ run sig failure
error: the following command terminated unexpectedly:
/home/ubuntu/sig/.zig-cache/o/0ffdc7ce91ed414658608f2b4566efb4/sig -l info shred-collector --test-repair-for-slot 292136186 --leader-schedule -- --entrypoint entrypoint.mainnet-beta.solana.com:8001 --entrypoint entrypoint2.mainnet-beta.solana.com:8001 --entrypoint entrypoint3.mainnet-beta.solana.com:8001 --entrypoint entrypoint4.mainnet-beta.solana.com:8001 --entrypoint entrypoint5.mainnet-beta.solana.com:8001
Build Summary: 15/17 steps succeeded; 1 failed (disable with --summary none)
run transitive failure
└─ run sig failure
error: the following build command failed with exit code 1:
/home/ubuntu/sig/.zig-cache/o/1793159a57b3fd420970e1e515c1a7b2/build /home/ubuntu/zig-linux-x86_64-0.13.0/zig /home/ubuntu/sig /home/ubuntu/sig/.zig-cache /home/ubuntu/.cache/zig --seed 0xfe09fa4d -Z6b0d69921d0949d0 run -Doptimize=ReleaseSafe -- -l info shred-collector --test-repair-for-slot 292136186 --leader-schedule -- --entrypoint entrypoint.mainnet-beta.solana.com:8001 --entrypoint entrypoint2.mainnet-beta.solana.com:8001 --entrypoint entrypoint3.mainnet-beta.solana.com:8001 --entrypoint entrypoint4.mainnet-beta.solana.com:8001 --entrypoint entrypoint5.mainnet-beta.solana.com:8001
```

```
zig build run -Doptimize=ReleaseSafe -- -l info gossip --entrypoint entrypoint.mainnet-beta.solana.com:8001 --entrypoint entrypoint2.mainnet-beta.solana.com:8001 --entrypoint entrypoint3.mainnet-beta.solana.com:8001 --entrypoint entrypoint4.mainnet-beta.solana.com:8001 --entrypoint entrypoint5.mainnet-beta.solana.com:8001

time=2024-09-26T15:07:10Z level=info  metrics port: 12345
time=2024-09-26T15:07:10Z level=info  entrypoints: { 34.83.231.102:8001, 147.28.180.247:8001, 145.40.64.255:8001, 147.28.134.81:8001, 147.28.198.87:8001 }
time=2024-09-26T15:07:10Z level=info  shred version: 50093 - from entrypoint ip echo: 34.83.231.102:8001
time=2024-09-26T15:07:10Z level=info  my ip: 18.221.215.219
time=2024-09-26T15:07:10Z level=info  gossip host: 18.221.215.219
time=2024-09-26T15:07:10Z level=info  gossip port: 8001
time=2024-09-26T15:07:10Z level=info  Starting gossip verifyPackets
time=2024-09-26T15:07:10Z level=info  Starting gossip processMessages
time=2024-09-26T15:07:10Z level=info  Starting gossip readSocket
time=2024-09-26T15:07:10Z level=info  Starting gossip buildMessages
time=2024-09-26T15:07:10Z level=info  Starting gossip sendSocket
time=2024-09-26T15:07:10Z level=info  gossip: recv 5: 0 ping, 0 pong, 0 push, 1 pull request, 0 pull response, 0 prune
time=2024-09-26T15:07:10Z level=info  gossip: sent: 0 ping, 0 pong, 0 push, 0 pull request, 0 pull response, 0 prune
time=2024-09-26T15:07:20Z level=info  gossip: recv 47794: 22 ping, 962 pong, 44132 push, 2678 pull request, 0 pull response, 0 prune
time=2024-09-26T15:07:20Z level=info  gossip: sent: 1594 ping, 22 pong, 0 push, 1 pull request, 4 pull response, 37 prune
double free or corruption (!prev)
run
└─ run sig failure
error: the following command terminated unexpectedly:
/home/ubuntu/sig/.zig-cache/o/0ffdc7ce91ed414658608f2b4566efb4/sig -l info gossip --entrypoint entrypoint.mainnet-beta.solana.com:8001 --entrypoint entrypoint2.mainnet-beta.solana.com:8001 --entrypoint entrypoint3.mainnet-beta.solana.com:8001 --entrypoint entrypoint4.mainnet-beta.solana.com:8001 --entrypoint entrypoint5.mainnet-beta.solana.com:8001
Build Summary: 15/17 steps succeeded; 1 failed (disable with --summary none)
run transitive failure
└─ run sig failure
error: the following build command failed with exit code 1:
/home/ubuntu/sig/.zig-cache/o/1793159a57b3fd420970e1e515c1a7b2/build /home/ubuntu/zig-linux-x86_64-0.13.0/zig /home/ubuntu/sig /home/ubuntu/sig/.zig-cache /home/ubuntu/.cache/zig --seed 0x4525f724 -Z2e90ab92e869cb68 run -Doptimize=ReleaseSafe -- -l info gossip --entrypoint entrypoint.mainnet-beta.solana.com:8001 --entrypoint entrypoint2.mainnet-beta.solana.com:8001 --entrypoint entrypoint3.mainnet-beta.solana.com:8001 --entrypoint entrypoint4.mainnet-beta.solana.com:8001 --entrypoint entrypoint5.mainnet-beta.solana.com:8001
```

```
zig build run -Doptimize=ReleaseSafe -- -l info gossip --entrypoint entrypoint.mainnet-beta.solana.com:8001 --entrypoint entrypoint2.mainnet-beta.solana.com:8001 --entrypoint entrypoint3.mainnet-beta.solana.com:8001 --entrypoint entrypoint4.mainnet-beta.solana.com:8001 --entrypoint entrypoint5.mainnet-beta.solana.com:8001

time=2024-09-26T15:07:59Z level=info  metrics port: 12345
time=2024-09-26T15:07:59Z level=info  entrypoints: { 34.83.231.102:8001, 147.28.180.247:8001, 145.40.64.255:8001, 147.28.134.81:8001, 147.28.198.87:8001 }
time=2024-09-26T15:07:59Z level=info  shred version: 50093 - from entrypoint ip echo: 34.83.231.102:8001
time=2024-09-26T15:07:59Z level=info  my ip: 18.221.215.219
time=2024-09-26T15:07:59Z level=info  gossip host: 18.221.215.219
time=2024-09-26T15:07:59Z level=info  gossip port: 8001
time=2024-09-26T15:07:59Z level=info  Starting gossip readSocket
time=2024-09-26T15:07:59Z level=info  Starting gossip verifyPackets
time=2024-09-26T15:07:59Z level=info  Starting gossip processMessages
time=2024-09-26T15:07:59Z level=info  Starting gossip sendSocket
time=2024-09-26T15:07:59Z level=info  Starting gossip buildMessages
time=2024-09-26T15:08:02Z level=info  gossip: recv 1: 1 ping, 0 pong, 0 push, 0 pull request, 0 pull response, 0 prune
time=2024-09-26T15:08:02Z level=info  gossip: sent: 0 ping, 0 pong, 0 push, 0 pull request, 0 pull response, 0 prune
time=2024-09-26T15:08:15Z level=info  gossip: recv 5: 4 ping, 0 pong, 0 push, 0 pull request, 0 pull response, 0 prune
time=2024-09-26T15:08:15Z level=info  gossip: sent: 0 ping, 5 pong, 0 push, 3 pull request, 0 pull response, 0 prune
time=2024-09-26T15:08:25Z level=info  gossip: recv 45259: 3830 ping, 180 pong, 39732 push, 1512 pull request, 0 pull response, 0 prune
time=2024-09-26T15:08:25Z level=info  gossip: sent: 1131 ping, 3830 pong, 0 push, 2 pull request, 0 pull response, 1640 prune
time=2024-09-26T15:08:27Z level=error  failed to generate push messages: error.OutOfMemory
thread 29853 panic: switch on corrupt value
/home/ubuntu/sig/src/gossip/service.zig:0:0: 0x159ff7f in buildPushMessages (sig)
/home/ubuntu/sig/src/gossip/service.zig:844:62: 0x15a2e89 in runService__anon_48039 (sig)
            const maybe_push_packets = self.buildPushMessages(&push_cursor) catch |e| blk: {
                                                             ^
/home/ubuntu/zig-linux-x86_64-0.13.0/lib/std/Thread.zig:429:13: 0x14e5f7b in entryFn (sig)
            @call(.auto, f, args) catch |err| {
            ^
???:?:?: 0x746294494ac2 in ??? (libc.so.6)
Unwind information for `libc.so.6:0x746294494ac2` was not available, trace may be incomplete

run
└─ run sig failure
error: the following command terminated unexpectedly:
/home/ubuntu/sig/.zig-cache/o/0ffdc7ce91ed414658608f2b4566efb4/sig -l info gossip --entrypoint entrypoint.mainnet-beta.solana.com:8001 --entrypoint entrypoint2.mainnet-beta.solana.com:8001 --entrypoint entrypoint3.mainnet-beta.solana.com:8001 --entrypoint entrypoint4.mainnet-beta.solana.com:8001 --entrypoint entrypoint5.mainnet-beta.solana.com:8001
Build Summary: 15/17 steps succeeded; 1 failed (disable with --summary none)
run transitive failure
└─ run sig failure
error: the following build command failed with exit code 1:
/home/ubuntu/sig/.zig-cache/o/1793159a57b3fd420970e1e515c1a7b2/build /home/ubuntu/zig-linux-x86_64-0.13.0/zig /home/ubuntu/sig /home/ubuntu/sig/.zig-cache /home/ubuntu/.cache/zig --seed 0x37e06834 -Z662a2f288dace029 run -Doptimize=ReleaseSafe -- -l info gossip --entrypoint entrypoint.mainnet-beta.solana.com:8001 --entrypoint entrypoint2.mainnet-beta.solana.com:8001 --entrypoint entrypoint3.mainnet-beta.solana.com:8001 --entrypoint entrypoint4.mainnet-beta.solana.com:8001 --entrypoint entrypoint5.mainnet-beta.solana.com:8001
```

0xNineteen:
interestingly enough, ive mainly been testing on testnet which we never seemed to receive any push messages from -- will likely need to run gossip across all networks for CI/CD

Rexicon226:
Huh, I didn't realize that there is a difference! Thanks for the crashes, I will now look into fixing them.

Rexicon226:
OK so, [60d49d5](https://github.com/Syndica/sig/pull/252/commits/60d49d52787d1b9c8de65ba3fd12d13450d94823) is the fix for this. It passes a couple of hours of running on `mainnet`. Not sure how else to test it, but I'm pretty certain this was the issue.

0xNineteen (approved):
runs good for me too :shipit:

0xNineteen (changes requested):
nvm just hit this on debug build: `./zig-out/bin/sig gossip -n mainnet`, 

```
thread 13629069 panic: Invalid free
/Users/x19/Documents/zig-macos-aarch64-0.13.0/lib/std/heap/general_purpose_allocator.zig:652:21: 0x103ff372f in freeLarge (sig)
                    @panic("Invalid free");
                    ^
/Users/x19/Documents/zig-macos-aarch64-0.13.0/lib/std/heap/general_purpose_allocator.zig:865:31: 0x103f3fe27 in free (sig)
                self.freeLarge(old_mem, log2_old_align, ret_addr);
                              ^
/Users/x19/Documents/zig-macos-aarch64-0.13.0/lib/std/mem/Allocator.zig:98:28: 0x103f44613 in free__anon_4878 (sig)
    return self.vtable.free(self.ptr, buf, log2_buf_align, ret_addr);
                           ^
/Users/x19/Documents/sig/src/bincode/bincode.zig:591:35: 0x104043673 in free__anon_39642 (sig)
                    allocator.free(value);
                                  ^
/Users/x19/Documents/sig/src/bincode/bincode.zig:569:33: 0x1045262b3 in free__anon_54792 (sig)
                    bincode.free(allocator, @field(value, field.name));
                                ^
/Users/x19/Documents/sig/src/bincode/bincode.zig:581:40: 0x10444cd2f in free__anon_52653 (sig)
                    return bincode.free(allocator, @field(value, field.name));
                                       ^
/Users/x19/Documents/sig/src/bincode/bincode.zig:589:37: 0x104335267 in free__anon_48361 (sig)
                        bincode.free(allocator, item);
                                    ^
/Users/x19/Documents/sig/src/bincode/bincode.zig:569:33: 0x104529ae3 in free__anon_54831 (sig)
                    bincode.free(allocator, @field(value, field.name));
                                ^
/Users/x19/Documents/sig/src/bincode/bincode.zig:569:33: 0x104450e77 in free__anon_52694 (sig)
                    bincode.free(allocator, @field(value, field.name));
                                ^
/Users/x19/Documents/sig/src/bincode/bincode.zig:581:40: 0x1043658ab in free__anon_49142 (sig)
                    return bincode.free(allocator, @field(value, field.name));
                                       ^
/Users/x19/Documents/sig/src/bincode/bincode.zig:569:33: 0x104365e07 in free__anon_49730 (sig)
                    bincode.free(allocator, @field(value, field.name));
                                ^
/Users/x19/Documents/sig/src/gossip/service.zig:1245:33: 0x104368997 in callback (sig)
                    bincode.free(self.allocator, val);
                                ^
/Users/x19/Documents/sig/src/sync/thread_pool.zig:669:36: 0x104545d5f in run (sig)
                    (task.callback)(task);
```

0xNineteen (dismissed):
:shipit:

src/shred_collector/shred_receiver.zig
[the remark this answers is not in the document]
Rexicon226: @dnut would you like me to do some benchmarking, or can this be done as a follow up? Need this convo resolved to merge 😄.

dnut (approved):

PR #258, opened 2024-08-29
refactor: remove misc folder

dnut:
This pr is a request for a discussion about a recent change. I overlooked a change in #255, which I would have liked to discuss before merging. The change was to introduce the `misc` folder to the repo root.

My assumption is that the folder was added to keep the root clean and high-level. I think that's a good goal, but I don't think adding a misc folder is the best approach to achieve that. It's unclear what belongs in a misc folder. What are the criteria for inclusion? A misc folder becomes a catchall bucket for people to put things when they don't know where else to put it. It hides the problem of information being disorganized, without actually solving it, which makes it easier for things to become less organized over time. If you're ever trying to find something that is in misc, it won't be obvious that you should look for it there. It's an extra step when you're looking for something.

I think it's ok to have a few more folders in the repo root as long as we take care to ensure they are reasonably high level and easy to navigate. In this PR I am not proposing that we should restore the previous structure. Instead I have tried to make things cleaner and more high-level, in an attempt to accomplish the purpose of the misc folder.

Another thing is the distinction between underscores `_` and dashes `-` in filesystem paths. Personally I think dashes are better because they are easier to type (don't need to hold shift). It's not a big deal but it's a minor pain point when navigating folders using the keyboard. Also I find dashes more readable (easier to ignore them as if they're spaces) but that probably varies from person to person.

### before #255
- genesis-files
- prometheus-grafana
   - grafana
   - prometheus
- scripts
- test_data

### after #255
- misc
  - genesis_files
  - prometheus_grafana
    - grafana
    - prometheus
  - scripts
  - test_data

### this pr's proposal
- data
  - genesis-files
  - test-data
- metrics
  - grafana
  - prometheus
- scripts


yewman (approved):
Thumbs up from me. I like your reasoning and tend to agree with the dashes over underscores in file paths.

dnut:
Thanks for the feedback @yewman.

Before merging, I'd like to hear from @0xNineteen and address any of his concerns first, since I'm proposing something different from what he intended.

0xNineteen:
i think your organization changes make sense 

i dont have a strong preference for underscore over dashes - i noticed 'src/shred_collector' was underscored and just wanted consistency so i made everything underscore - wdyt?

dnut:
>  i noticed 'src/shred_collector' was underscored and just wanted consistency so i made everything underscore

Yeah there are actually a ton of examples of this in our source code, mainly it's the filenames though.

Many programming languages have a strict requirement that source code paths are valid identifiers in the language, which usually excludes anything other than alphanumeric characters plus underscores. It's not required in zig but it seems to be a standard in zig anyway.

Even in languages that require underscores for source paths, packages are still named with dashes instead of underscores. Not all, there are many exceptions, but I dashes are more common. The name may be written in source code with an underscore, but then when the same exact name is used in a package manager, or for the binary in the filesystem, dashes are used instead. It seems there is a strong enough preference for dashes that it overrides the desire for universal consistency in most cases. This pattern has become so normalized that it almost feels "correct" to me in a way, despite it being totally arbitrary. The dash/underscore inconsistency can be seen as a sort of demarcation between source code and everything else.

With zig, there is no requirement that the actual filesystem paths have underscores. We could use dashes for the filenames. But that would be *very* unusual. I haven't seen any other projects use dashes for this. Alternatively we could use underscores for everything. That would be less weird. I'm not totally opposed to it. I just have a slight personal preference for using dashes for any non-source files. It's internally inconsistent, but it's externally consistent with a commonly used pattern.

0xNineteen:
ok makes sense - thanks for clarifying

0xNineteen (approved):

PR #259, opened 2024-08-29
improve(allocators): improve DiskMemoryAllocator

InKryption:
Makes the allocator a bit smarter, able to resize allocations, and the ability to delete the associated file when freeing an allocation. Also removes the `init` method in favor of simply initializing the struct; as well, removes the `deinit` method, since there's nothing left to cleanup now that the files are deleted after properly freeing allocations.
Also adds a couple tests.

### DIFF
--- a/src/accountsdb/readme.md
+++ b/src/accountsdb/readme.md
@@ -178,9 +178,12 @@ disk allocator which creates memory from mmap-ing files stored on disk.
 // incremented by one for each new allocation.
 var dma_dir = try std.fs.cwd().makeOpenPath("data/test-data");
 defer dma_dir.close();
-var dma_state: DiskMemoryAllocator = try DiskMemoryAllocator.init(dma_dir, logger);
+var dma_state: DiskMemoryAllocator = .{};
 const dma = dma_state.allocator();
 ```
+
+Unlike a simpler page allocator, it stores certain metadata after the user-facing
+buffer which tracks the associated file, and the true mmap'd size to allow for resizes.
 
 ### background-threads
 
--- a/src/accountsdb/readme.md
+++ b/src/accountsdb/readme.md
@@ -178,7 +178,7 @@ disk allocator which creates memory from mmap-ing files stored on disk.
 // incremented by one for each new allocation.
 var dma_dir = try std.fs.cwd().makeOpenPath("data/test-data");
 defer dma_dir.close();
-var dma_state = try DiskMemoryAllocator.init(dma_dir, logger);
+var dma_state: DiskMemoryAllocator = try DiskMemoryAllocator.init(dma_dir, logger);
 const dma = dma_state.allocator();
 ```
 
--- a/src/accountsdb/readme.md
+++ b/src/accountsdb/readme.md
@@ -174,10 +174,12 @@ to support disk-based account references, we created a general purpose
 disk allocator which creates memory from mmap-ing files stored on disk.
 
 ```zig
-// files are created using `data/test-data/tmp_{i}` format where `i` is 
-// incremented by one for each alloc call.
-var allocator = try DiskMemoryAllocator.init("data/test-data/tmp");
-defer allocator.deinit(null);
+// files are created using `data/test-data/bin_{i}` format where `i` is 
+// incremented by one for each new allocation.
+var dma_dir = try std.fs.cwd().makeOpenPath("data/test-data");
+defer dma_dir.close();
+var dma_state = try DiskMemoryAllocator.init(dma_dir, logger);
+const dma = dma_state.allocator();
 ```
 
 ### background-threads
--- a/src/utils/allocators.zig
+++ b/src/utils/allocators.zig
@@ -140,36 +140,20 @@ pub fn RecycleFBA(config: struct {
 pub const DiskMemoryAllocator = struct {
     dir: std.fs.Dir,
     logger: sig.trace.Logger,
-    /// The address space mmap'd to a particular file will be at least
-    /// `(file_size * (1000 + mmap_ratio)) / 1000` in integer terms.
-    /// With a value of 0, the mmap'd size will be equal to the file size.
-    /// With a value of 1000, the mmap'd size will be double the file size.
-    /// NOTE: not intended to be changed after initialization.
-    mmap_ratio: MmapRatio = 0,
+    /// The amount of memory mmap'd for a particular allocation will be `file_size * mmap_ratio`.
+    /// See `alignedFileSize` and its usages to understand the relationship between an allocation
+    /// size, and the size of a file, and by extension, how this then relates to the allocated
+    /// address space.
+    mmap_ratio: MmapRatio = 1,
     count: std.atomic.Value(u32) = std.atomic.Value(u32).init(0),
     const Self = @This();
 
-    pub const MmapRatio = u32;
+    pub const MmapRatio = u16;
 
     /// Metadata stored at the end of each allocation.
     const Metadata = extern struct {
         file_index: u32,
-        /// In an extern struct only ABI-sized types are allowed, so this is a `std.math.Log2Int(usize)`,
-        /// wrapped in a packed struct to define the padding bits.
-        mmap_size: packed struct(u8) {
-            /// The log2 of the mmap size - the actual mmap size can be acquired via `1 << log2`. See `get`.
-            /// This is done this way to minimize the amount of redundantly stored information: the mmap size
-            /// will always be a power of two, and as such can be represented minimally using its log2.
-            log2: UsizeLog2,
-            _padding: enum(PaddingInt) { unset = 0 } = .unset,
-
-            const UsizeLog2 = std.math.Log2Int(usize);
-            const PaddingInt = std.meta.Int(.unsigned, @bitSizeOf(u8) - @bitSizeOf(UsizeLog2));
-
-            pub inline fn get(self: @This()) usize {
-                return @as(usize, 1) << self.log2;
-            }
-        },
+        mmap_size: usize align(4),
     };
 
     /// Returns the aligned size with enough space for `size` and `Metadata` at the end.
@@ -177,14 +161,12 @@ pub const DiskMemoryAllocator = struct {
         return std.mem.alignForward(usize, size + @sizeOf(Metadata), std.mem.page_size);
     }
 
-    /// Returns the aligned size to mmap relative to `alignedFileSize(size)`.
     inline fn alignedMmapSize(
         /// Must be `= alignedFileSize(size)`.
         aligned_file_size: usize,
         mmap_ratio: MmapRatio,
     ) usize {
-        const min_mmap_size = (aligned_file_size * (1000 + mmap_ratio)) / 1000;
-        return std.mem.alignForward(usize, min_mmap_size, std.mem.page_size);
+        return aligned_file_size *| mmap_ratio;
     }
 
     pub inline fn allocator(self: *Self) std.mem.Allocator {
@@ -206,6 +188,7 @@ pub const DiskMemoryAllocator = struct {
     fn alloc(ctx: *anyopaque, size: usize, log2_align: u8, return_address: usize) ?[*]u8 {
         _ = return_address;
         const self: *Self = @ptrCast(@alignCast(ctx));
+        std.debug.assert(self.mmap_ratio != 0);
 
         const alignment = @as(usize, 1) << @intCast(log2_align);
         std.debug.assert(alignment <= std.mem.page_size); // the allocator interface shouldn't allow this (aside from the *Raw methods).
@@ -244,11 +227,12 @@ pub const DiskMemoryAllocator = struct {
         std.debug.assert(size <= file_aligned_size - @sizeOf(Metadata)); // sanity check
         std.mem.bytesAsValue(Metadata, full_alloc[size..][0..@sizeOf(Metadata)]).* = .{
             .file_index = file_index,
-            .mmap_size = .{ .log2 = std.math.log2_int(usize, aligned_mmap_size) },
+            .mmap_size = aligned_mmap_size,
         };
         return full_alloc.ptr;
     }
 
+    /// Resizes the allocation within the bounds of the mmap'd address space if possible.
     fn resize(
         ctx: *anyopaque,
         buf: []u8,
@@ -258,6 +242,7 @@ pub const DiskMemoryAllocator = struct {
     ) bool {
         _ = return_address;
         const self: *Self = @ptrCast(@alignCast(ctx));
+        std.debug.assert(self.mmap_ratio != 0);
 
         const alignment = @as(usize, 1) << @intCast(log2_align);
         std.debug.assert(alignment <= std.mem.page_size); // the allocator interface shouldn't allow this (aside from the *Raw methods).
@@ -272,7 +257,7 @@ pub const DiskMemoryAllocator = struct {
         const buf_ptr: [*]align(std.mem.page_size) u8 = @alignCast(buf.ptr);
         const metadata: Metadata = @bitCast(buf_ptr[old_file_aligned_size - @sizeOf(Metadata) ..][0..@sizeOf(Metadata)].*);
 
-        if (new_file_aligned_size > metadata.mmap_size.get()) {
+        if (new_file_aligned_size > metadata.mmap_size) {
             return false;
         }
 
@@ -288,10 +273,7 @@ pub const DiskMemoryAllocator = struct {
         file.setEndPos(new_file_aligned_size) catch return false;
 
         std.debug.assert(new_size <= new_file_aligned_size - @sizeOf(Metadata)); // sanity check
-        std.mem.bytesAsValue(Metadata, buf_ptr[new_size..][0..@sizeOf(Metadata)]).* = .{
-            .file_index = metadata.file_index,
-            .mmap_size = metadata.mmap_size,
-        };
+        std.mem.bytesAsValue(Metadata, buf_ptr[new_size..][0..@sizeOf(Metadata)]).* = metadata;
 
         return true;
     }
@@ -300,7 +282,7 @@ pub const DiskMemoryAllocator = struct {
     fn free(ctx: *anyopaque, buf: []u8, log2_align: u8, return_address: usize) void {
         _ = return_address;
         const self: *Self = @ptrCast(@alignCast(ctx));
-
+        std.debug.assert(self.mmap_ratio != 0);
         std.debug.assert(buf.len != 0); // should be ensured by the allocator interface
 
         const alignment = @as(usize, 1) << @intCast(log2_align);
@@ -428,13 +410,13 @@ test "disk allocator stdlib test" {
     const tmp_dir = tmp_dir_root.dir;
 
     for ([_]DiskMemoryAllocator.MmapRatio{
-        0,    1,    2,    3,  4,  5,
-        10,   11,   12,   13, 14, 15,
-        20,   21,   22,   23, 24, 25,
-        30,   31,   32,   33, 34, 35,
-        40,   41,   42,   43, 44, 45,
-        50,   51,   52,   53, 54, 55,
-        1000, 2000, 3000,
+        1,  2,  3,  4,  5,
+        10, 11, 12, 13, 14,
+        15, 20, 21, 22, 23,
+        24, 25, 30, 31, 32,
+        33, 34, 35, 40, 41,
+        42, 43, 44, 45, 50,
+        51, 52, 53, 54, 55,
     }) |ratio| {
         var dma_state: DiskMemoryAllocator = .{
             .dir = tmp_dir,
@@ -530,4 +512,3 @@ test "disk allocator large realloc" {
 
     page1[page1.len - 1] = 10;
 }
-
--- a/src/utils/allocators.zig
+++ b/src/utils/allocators.zig
@@ -170,16 +170,6 @@ pub const DiskMemoryAllocator = struct {
     count: std.atomic.Value(u32) = std.atomic.Value(u32).init(0),
     const Self = @This();
 
-    pub inline fn init(
-        index_dir: std.fs.Dir,
-        logger: sig.trace.Logger,
-    ) Self {
-        return .{
-            .dir = index_dir,
-            .logger = logger,
-        };
-    }
-
     /// Metadata stored at the end of each allocation.
     const Metadata = extern struct {
         file_index: u32,
@@ -319,7 +309,10 @@ test "disk allocator on hashmaps" {
     defer tmp_dir_root.cleanup();
     const tmp_dir = tmp_dir_root.dir;
 
-    var allocator = DiskMemoryAllocator.init(tmp_dir, .noop);
+    var allocator: DiskMemoryAllocator = .{
+        .dir = tmp_dir,
+        .logger = .noop,
+    };
 
     var refs = std.AutoHashMap(u8, u8).init(allocator.allocator());
     defer refs.deinit();
@@ -337,7 +330,10 @@ test "disk allocator on arraylists" {
     defer tmp_dir_root.cleanup();
     const tmp_dir = tmp_dir_root.dir;
 
-    var dma_state = DiskMemoryAllocator.init(tmp_dir, .noop);
+    var dma_state: DiskMemoryAllocator = .{
+        .dir = tmp_dir,
+        .logger = .noop,
+    };
     const dma = dma_state.allocator();
 
     {
--- a/src/utils/allocators.zig
+++ b/src/utils/allocators.zig
@@ -204,11 +204,12 @@ pub const DiskMemoryAllocator = struct {
         const alignment = @as(usize, 1) << @intCast(log2_align);
         std.debug.assert(alignment <= std.mem.page_size); // the allocator interface shouldn't allow this (aside from the *Raw methods).
 
-        const aligned_size = fullMmapSize(size);
+        const aligned_size = alignedFileSize(size);
 
         const file_index = self.count.fetchAdd(1, .monotonic);
         const file_name_bounded = fileNameBounded(file_index);
         const file_name = file_name_bounded.constSlice();
+
         const file = self.dir.createFile(file_name, .{ .read = true, .truncate = true }) catch |err| {
             self.logFailure(err, file_name);
             return null;
@@ -252,10 +253,10 @@ pub const DiskMemoryAllocator = struct {
         const alignment = @as(usize, 1) << @intCast(log2_align);
         std.debug.assert(alignment <= std.mem.page_size); // the allocator interface shouldn't allow this (aside from the *Raw methods).
 
-        const buf_ptr: [*]align(std.mem.page_size) u8 = @alignCast(buf.ptr);
-        const aligned_size = fullMmapSize(buf.len);
-        const new_aligned_size = fullMmapSize(new_size);
+        const aligned_size = alignedFileSize(buf.len);
+        const new_aligned_size = alignedFileSize(new_size);
 
+        const buf_ptr: [*]align(std.mem.page_size) u8 = @alignCast(buf.ptr);
         const metadata: Metadata = @bitCast(buf_ptr[buf.len..][0..@sizeOf(Metadata)].*);
         const file_name_bounded = fileNameBounded(metadata.file_index);
         const file_name = file_name_bounded.constSlice();
@@ -300,10 +301,11 @@ pub const DiskMemoryAllocator = struct {
         const alignment = @as(usize, 1) << @intCast(log2_align);
         std.debug.assert(alignment <= std.mem.page_size); // the allocator interface shouldn't allow this (aside from the *Raw methods).
 
-        const buf_ptr: [*]align(std.mem.page_size) u8 = @alignCast(buf.ptr);
-        const aligned_size = fullMmapSize(buf.len);
+        const aligned_size = alignedFileSize(buf.len);
 
+        const buf_ptr: [*]align(std.mem.page_size) u8 = @alignCast(buf.ptr);
         const metadata: Metadata = @bitCast(buf_ptr[buf.len..][0..@sizeOf(Metadata)].*);
+
         const file_name_bounded = fileNameBounded(metadata.file_index);
         const file_name = file_name_bounded.constSlice();
 
@@ -314,7 +316,7 @@ pub const DiskMemoryAllocator = struct {
     }
 
     /// Returns the aligned size with enough space for `size` and `Metadata` at the end.
-    inline fn fullMmapSize(size: usize) usize {
+    inline fn alignedFileSize(size: usize) usize {
         return std.mem.alignForward(usize, size + @sizeOf(Metadata), std.mem.page_size);
     }
 
--- a/src/utils/allocators.zig
+++ b/src/utils/allocators.zig
@@ -247,22 +247,24 @@ pub const DiskMemoryAllocator = struct {
         const aligned_size = alignedFileSize(buf.len);
         const new_aligned_size = alignedFileSize(new_size);
 
-        const buf_ptr: [*]align(std.mem.page_size) u8 = @alignCast(buf.ptr);
-        const metadata: Metadata = @bitCast(buf_ptr[buf.len..][0..@sizeOf(Metadata)].*);
-
         if (new_aligned_size == aligned_size) {
             return true;
         }
 
-        if (new_aligned_size < aligned_size) {
-            std.posix.munmap(@alignCast(buf_ptr[new_aligned_size..aligned_size]));
-            std.mem.bytesAsValue(Metadata, buf_ptr[new_size..][0..@sizeOf(Metadata)]).* = .{
-                .file_index = metadata.file_index,
-            };
-            return true;
+        if (new_aligned_size > aligned_size) {
+            return false;
         }
 
-        return false;
+        std.debug.assert(new_aligned_size < aligned_size);
+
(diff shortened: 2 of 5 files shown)


src/utils/allocators.zig
```diff
@@ -178,120 +192,154 @@ pub const DiskMemoryAllocator = struct {
...
-        // TODO: build a mapping from ptr to file so we can delete the corresponding file on free
-        const buf_aligned_len = std.mem.alignForward(usize, buf.len, std.mem.page_size);
-        std.posix.munmap(@alignCast(buf.ptr[0..buf_aligned_len]));
+        std.mem.bytesAsValue(Metadata, full_alloc[size..][0..@sizeOf(Metadata)]).* = .{
+            .file_index = file_index,
+        };
+        return full_alloc.ptr;
     }
 
     /// not supported rn
```
0xNineteen, line 237: probs want to remove this now

src/accountsdb/readme.md
```diff
@@ -174,10 +174,12 @@ to support disk-based account references, we created a general purpose
...
-// files are created using `data/test-data/tmp_{i}` format where `i` is 
-// incremented by one for each alloc call.
-var allocator = try DiskMemoryAllocator.init("data/test-data/tmp");
-defer allocator.deinit(null);
+// files are created using `data/test-data/bin_{i}` format where `i` is 
+// incremented by one for each new allocation.
+var dma_dir = try std.fs.cwd().makeOpenPath("data/test-data");
+defer dma_dir.close();
+var dma_state = try DiskMemoryAllocator.init(dma_dir, logger);
+const dma = dma_state.allocator();
```
0xNineteen, line 182: can you add a sentence or two on how the disk allocator works (ie, storing metdata at the end of the buffer)

src/utils/allocators.zig
```diff
@@ -178,120 +192,154 @@ pub const DiskMemoryAllocator = struct {
...
+        const alignment = @as(usize, 1) << @intCast(log2_align);
+        std.debug.assert(alignment <= std.mem.page_size); // the allocator interface shouldn't allow this (aside from the *Raw methods).
+
+        const buf_ptr: [*]align(std.mem.page_size) u8 = @alignCast(buf.ptr);
+        const aligned_size = fullMmapSize(buf.len);
+        const new_aligned_size = fullMmapSize(new_size);
+
+        const metadata: Metadata = @bitCast(buf_ptr[buf.len..][0..@sizeOf(Metadata)].*);
+        const file_name_bounded = fileNameBounded(metadata.file_index);
+        const file_name = file_name_bounded.constSlice();
```
0xNineteen, line 257: these var are only used in the second if block -- can we move them there

src/utils/allocators.zig
```diff
@@ -178,120 +192,154 @@ pub const DiskMemoryAllocator = struct {
...
+        const self: *Self = @ptrCast(@alignCast(ctx));
+
+        const alignment = @as(usize, 1) << @intCast(log2_align);
+        std.debug.assert(alignment <= std.mem.page_size); // the allocator interface shouldn't allow this (aside from the *Raw methods).
+
+        const buf_ptr: [*]align(std.mem.page_size) u8 = @alignCast(buf.ptr);
+        const aligned_size = fullMmapSize(buf.len);
+        const new_aligned_size = fullMmapSize(new_size);
+
+        const metadata: Metadata = @bitCast(buf_ptr[buf.len..][0..@sizeOf(Metadata)].*);
```
0xNineteen, line 255: similarily, this should be after the alignment if check

src/utils/allocators.zig
```diff
@@ -178,120 +192,154 @@ pub const DiskMemoryAllocator = struct {
...
+        const file_name_bounded = fileNameBounded(metadata.file_index);
+        const file_name = file_name_bounded.constSlice();
 
-    // diff result with larger slice
-    const bytes4 = try allocator.allocator().alloc(u8, 200);
-    try std.testing.expect(ptr != bytes4.ptr);
-    allocator.allocator().free(bytes4);
-}
+        std.posix.munmap(buf_ptr[0..aligned_size]);
+        self.dir.deleteFile(file_name) catch |err| {
```
0xNineteen, line 307: why do we want to delete the file?

src/utils/allocators.zig
```diff
@@ -178,120 +192,154 @@ pub const DiskMemoryAllocator = struct {
...
+            return true;
+        }
     }
-};
 
-test "recycle allocator" {
-    const backing_allocator = std.testing.allocator;
-    var allocator = try RecycleFBA(.{}).init(backing_allocator, 1024);
-    defer allocator.deinit();
+    /// unmaps the memory (file still exists and is removed on deinit())
```
0xNineteen, line 291: there is no deinit()

src/utils/allocators.zig
```diff
@@ -178,120 +192,154 @@ pub const DiskMemoryAllocator = struct {
...
-    try std.testing.expect(ptr != bytes4.ptr);
-    allocator.allocator().free(bytes4);
-}
+        std.posix.munmap(buf_ptr[0..aligned_size]);
+        self.dir.deleteFile(file_name) catch |err| {
+            self.logFailure(err, file_name);
+        };
+    }
+
+    const Metadata = extern struct {
```
0xNineteen, line 312: can you define this underneath the fields -- also can u update the contributing .md to describe this 

we want it ordered like `struct { fields, self = @This(), structs, init, deinit, methods }`

src/utils/allocators.zig
```diff
@@ -178,120 +192,154 @@ pub const DiskMemoryAllocator = struct {
...
+    const file_name_max_len = sig.utils.fmt.boundedLenValue("bin_{d}", .{std.math.maxInt(usize)});
+    inline fn fileNameBounded(file_index: usize) std.BoundedArray(u8, file_name_max_len) {
+        return sig.utils.fmt.boundedFmt("bin_{d}", .{file_index});
+    }
+};
 
 test "disk allocator on hashmaps" {
-    var allocator = DiskMemoryAllocator.init(sig.TEST_DATA_DIR ++ "tmp");
-    defer allocator.deinit(null);
+    var tmp_dir_root = std.testing.tmpDir(.{});
```
0xNineteen, line 334: even though the probability of two random dir being the same name is close to zero, would pref to have it a static string value -- see how we do blockstore tests

0xNineteen (changes requested):

src/utils/allocators.zig
```diff
@@ -178,120 +192,154 @@ pub const DiskMemoryAllocator = struct {
...
-    allocator.allocator().free(bytes4);
-}
+        std.posix.munmap(buf_ptr[0..aligned_size]);
+        self.dir.deleteFile(file_name) catch |err| {
+            self.logFailure(err, file_name);
+        };
+    }
+
+    const Metadata = extern struct {
+        file_index: usize,
```
dnut, line 313: Is there an advantage of usize over u64 or u32?

src/utils/allocators.zig
0xNineteen: thinking about it more ... probably good to delete the file for now - dont want to modify this struct for future functionality (fastboot) which may never come about

src/utils/allocators.zig
InKryption: Mostly to express the link to the address space - but thinking about it, each increment is at least a whole page size, meaning there's effectively no situation in which we'd even come close to the address space limit, so u32 would make more sense.

src/utils/allocators.zig
InKryption: What's the advantage? Static dirs have a bigger weakness imo: programmers can accidentally cause them to race by doing something as simple as copy-pasting code.

Also, to be clear, not only are the chances of two dirs having the same randomly having the same name essentially zero, but the temporary directory tree is deleted on cleanup, meaning the only time two tests could race is if they are running in parallel, in which case the likelihood of the crypto rng generating the same sequence twice is practically indistinguishable from zero.

src/utils/allocators.zig
InKryption: Yeah, I wanted to discuss about that, if we ever do go forward with fastboot, I don't think this is the right layer of abstraction to do that; to properly implement fastboot, the allocator would need to expose where the memory for the data structures it's allocating is actually located, which just isn't expressible through the allocator interface.

src/utils/allocators.zig
```diff
@@ -178,120 +192,154 @@ pub const DiskMemoryAllocator = struct {
...
+        if (aligned_size == new_aligned_size) return true;
+
+        if (new_aligned_size < aligned_size) {
+            std.posix.munmap(@alignCast(buf_ptr[new_aligned_size..aligned_size]));
+            std.mem.bytesAsValue(Metadata, buf_ptr[new_size..][0..@sizeOf(Metadata)]).* = .{
+                .file_index = metadata.file_index,
+            };
+            return true;
+        } else {
+            const file = self.dir.openFile(file_name, .{ .mode = .read_write }) catch |err| {
```
dnut, line 268: Doesn't the file need to be resized? Won't you get a SIGBUS when you try to access mmapped memory that is out of the file bounds?

> SIGBUS   Attempted access to a page of the buffer that lies beyond the end of the mapped file.

https://man7.org/linux/man-pages/man2/mmap.2.html

src/utils/allocators.zig
```diff
@@ -178,120 +192,154 @@ pub const DiskMemoryAllocator = struct {
...
+                new_aligned_size,
+                std.posix.PROT.READ | std.posix.PROT.WRITE,
+                std.posix.MAP{ .TYPE = .SHARED },
+                file.handle,
+                0,
+            ) catch |err| {
+                self.logFailure(err, file_name);
+                return false;
+            };
+            std.debug.assert(mapped.ptr == buf_ptr);
```
dnut, line 285: Is this guaranteed by mmap?

src/utils/allocators.zig
```diff
@@ -178,120 +192,154 @@ pub const DiskMemoryAllocator = struct {
...
+            };
+            return true;
+        } else {
+            const file = self.dir.openFile(file_name, .{ .mode = .read_write }) catch |err| {
+                self.logFailure(err, file_name);
+                return false;
+            };
+            defer file.close();
+
+            const mapped = std.posix.mmap(
```
dnut, line 274: Could this call to mmap be avoided if the initial mmap is sufficiently large? The initial alloc could create a small file while mmapping a much larger address space. For example, what about this:

1. alloc is called requesting less than one page of memory
2. alloc creates a file sized as 1 page
3. alloc mmaps 10 pages of address space, backed by the newly created file
4. alloc returns a slice with a len for the requested size (within the first page, which is valid memory)
5. resize is called requesting a size between 2 and 3 pages
6. the file is resized to be 3 pages in size
7. the same pointer can be reused without another call to mmap

The downside is the complexity of separately keeping track of the mmap size that was used during the prior call to alloc or resize, perhaps by storing it in the metadata.

src/utils/allocators.zig
```diff
@@ -178,120 +192,154 @@ pub const DiskMemoryAllocator = struct {
...
+        const aligned_size = fullMmapSize(buf.len);
+        const new_aligned_size = fullMmapSize(new_size);
+
+        const metadata: Metadata = @bitCast(buf_ptr[buf.len..][0..@sizeOf(Metadata)].*);
+        const file_name_bounded = fileNameBounded(metadata.file_index);
+        const file_name = file_name_bounded.constSlice();
+
+        if (aligned_size == new_aligned_size) return true;
+
+        if (new_aligned_size < aligned_size) {
```
dnut, line 261: What about when new_aligned_size == aligned_size? Shouldn't that just return true without doing anything else?

src/utils/allocators.zig
0xNineteen: ah thats a good point - static name dirs do have a chance of overlapping if the programmer makes a mistake - random names makes more sense in that case

@dnut wdyt about changing the blockstore test code to use tmpDirs too?

src/utils/allocators.zig
0xNineteen: that makes sense - maybe we can add a comment how its a general rule and isnt strict - adding the example you mentioned

src/utils/allocators.zig
InKryption: That is what happens just a couple of lines above, is something unclear? Guess maybe I should reorder that comparison.

src/utils/allocators.zig
InKryption: Nope, I thought it was but I was mistaken.

src/utils/allocators.zig
InKryption: This is an interesting plan, and I think it could actually done without any extra state tracking, ie by using a fixed algorithm which calculates a requested length into a number of pages/page-size-compatible length.
Though this altogether also comes with the drawback of using more (disk) memory. A different tradeoff we could make would be to instead use DMA as the backing allocator of something like an arena allocator, or some other intelligent allocator which allocates larger amounts of memory at a time, to reduce the extent to which DMA has to do complicated tracking of this kind of state. Lmk your thoughts.

src/utils/allocators.zig
InKryption: The main reason I put the Metadata struct where it is is because it's more of an implementation detail to the allocator, there's no real way for a user of this struct to interface with this part of the DMA. I'm fine to move it up if you're set on that, but imo it's mostly clutter as opposed to actually useful information.

src/utils/allocators.zig
InKryption: What do you mean here exactly? This is happening after the alignment assertion.

src/utils/allocators.zig
dnut: > comes with the drawback of using more (disk) memory

My suggestion was to use the same amount of disk memory, not to increase it. I think you're talking about something different. The file size and mmap length are two distinct dimensions. Here are three strategies for managing those numbers:
- _the current implementation:_ the file size and the mmap length are the same number. both are the minimum number of pages to satisfy the alloc
- _what I think you're talking about when you say "more disk memory":_ the file size and the mmap length are the same number. this is an arbitrarily large number containing extra pages that are not needed now but can be used later.
- _what I was suggesting:_ the file and the mmap have different sizes from each other. the file size is the minimum number of pages to satisfy the alloc. the mmap length is a larger number that could support more pages being added to the file.

So, in my suggestion, the file would have the same size as your current approach. The only difference I was suggesting was to pass an arbitrarily large `length` argument into mmap. That means mmap would assign a larger address space to the pointer, but it wouldn't actually consume any extra memory in disk or ram. The full address space would not be usable at first, but that's ok because we only need a subset of that address space to satisfy the current alloc request.

The goal was just to avoid extra calls to mmap. You would still need to increase the size of the file for every resize. As long as the actual size of the file doesn't exceed the size of the address space created from the prior call to mmap, you could continue using the same pointer without needing to call mmap again.

> A different tradeoff we could make would be to instead use DMA as the backing allocator of something like an arena allocator, or some other intelligent allocator which allocates larger amounts of memory at a time, to reduce the extent to which DMA has to do complicated tracking of this kind of state. Lmk your thoughts.

This sounds like a good idea.

InKryption: Ah, I had a partial misunderstanding. I understood you intended for the file size and address space sizes to be different, but I actually didn't realize that mmapping more memory than exists doesn't consume disk space - in retrospect that's obvious, my bad.

As for the DMA as a backing allocator, mayhaps we can make the DMA configurable: a mode where it acts like it currently does, and a mode where it uses the slightly more intelligent strategy you've described that I now better comprehend the implications of.
I'll go ahead and do it this way, and you can tell me how you feel about each version, and if we keep both or just one.

dnut: If you want to make it configurable, the configuration could be an integer parameter that specifies the ratio of mmap length to file size. Then it would essentially always be in the same mode of `mmap length >= file size`.  `mmap length == file size` would be a special case of the data, but would use the same logic.

src/utils/allocators.zig
dnut: haha woops, I posted this without thinking

src/utils/allocators.zig
InKryption: 3a87651d249056617ab0fc24687f56a16bcfd4ee

src/utils/allocators.zig
dnut: I tried to address that issue, and I've been having some more ideas to reduce the risk even further. But I can't think of a totally foolproof solution for that problem, unless you ensure a unique path is generated on every call, which tmpdir is good for.

My original goal was to leave around the state from the test in a place that's easy to find. That way I could inspect it if something goes wrong in the test. But actually, it's pretty rare that I need to do that. Usually I just re-run the entire test and debug it. Also, if I need to ad hoc inspect the state for failed tests, it's not that hard to change the test to use a known path. So I'm fine with using tmp dirs if that's what you guys prefer.

src/utils/allocators.zig
0xNineteen: its currently 
```
        const metadata: Metadata = @bitCast(buf_ptr[buf.len..][0..@sizeOf(Metadata)].*);
        if (aligned_size == new_aligned_size) return true;
```
whereas it should be 
```
        if (aligned_size == new_aligned_size) return true;
        const metadata: Metadata = @bitCast(buf_ptr[buf.len..][0..@sizeOf(Metadata)].*);
```

src/utils/allocators.zig
0xNineteen: talked about this offline - for consistency with the rest of the codebase were gonna move Metadata to the top

src/utils/allocators.zig
0xNineteen: yeah would pref that, whenever we can, pretty low priority rn - created a ticket : https://github.com/orgs/Syndica/projects/2/views/1?pane=issue&itemId=78811926

src/utils/allocators.zig
InKryption: Oh that's what you mean. Sure, good point.

src/utils/allocators.zig
InKryption: 6c5d2322b3903eb4e646722255223fe931a3e14c

src/utils/allocators.zig
InKryption: 6c5d2322b3903eb4e646722255223fe931a3e14c

src/utils/allocators.zig
InKryption: 6a43f7ef85ac570547240b9619acf95fe0ff9b9f

src/utils/allocators.zig
InKryption: 9afa8d5e369f503a9716c34dd3813bbb88af27bc

src/utils/allocators.zig
InKryption: I actually attempted to do this, but I ran into an error where it appears that resizing the file invalidates the mapped pointer? It leads segmentation faults and bus errors when trying to store the metadata.
I'm going to eschew the more advanced version of this.

InKryption: ae191ee9c6db6640e068f82dda492c766ddf8b80

InKryption:
Sporadic kcov_test failure before re-running the job: https://github.com/Syndica/sig/actions/runs/10743493043/job/29798352658?pr=259

src/utils/allocators.zig
InKryption: * Fixed the bug: 9afa8d5e369f503a9716c34dd3813bbb88af27bc
* Improvement: c3dc473a41c1d900ff434a7a4e3c9c375f386b25

src/accountsdb/readme.md
InKryption: I'll get to updating this as soon as drew's review is resolved and the final details are confirmed.

src/utils/allocators.zig
```diff
@@ -177,121 +209,191 @@ pub const DiskMemoryAllocator = struct {
...
+        const alignment = @as(usize, 1) << @intCast(log2_align);
+        std.debug.assert(alignment <= std.mem.page_size); // the allocator interface shouldn't allow this (aside from the *Raw methods).
+
+        const old_file_aligned_size = alignedFileSize(buf.len);
+        const old_mmap_aligned_size = alignedMmapSize(old_file_aligned_size, self.mmap_ratio);
+
+        const new_file_aligned_size = alignedFileSize(new_size);
+        const new_mmap_aligned_size = alignedMmapSize(new_file_aligned_size, self.mmap_ratio);
+
+        if (new_mmap_aligned_size > old_mmap_aligned_size) {
```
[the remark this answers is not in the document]
InKryption: This is equivalent to the file size comparison only when mmap_ratio = 0; when it isn't, the mmap'd size can be larger than the file size. This check is to return false when the resize request would require mmapping more space, which isn't possible.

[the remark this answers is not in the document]
dnut: > This check is to return false when the resize request would require mmapping more space, which isn't possible.

The new mmap size will always be larger than the old mmap size as long as the new file size is larger than the old file size. But you don't actually need to mmap additional space as long as the file size is less than the mmap size.

src/utils/allocators.zig
```diff
@@ -177,121 +209,191 @@ pub const DiskMemoryAllocator = struct {
...
-        _ = new_size;
         _ = return_address;
-        return false;
+        const self: *Self = @ptrCast(@alignCast(ctx));
+
+        const alignment = @as(usize, 1) << @intCast(log2_align);
+        std.debug.assert(alignment <= std.mem.page_size); // the allocator interface shouldn't allow this (aside from the *Raw methods).
+
+        const old_file_aligned_size = alignedFileSize(buf.len);
+        const old_mmap_aligned_size = alignedMmapSize(old_file_aligned_size, self.mmap_ratio);
```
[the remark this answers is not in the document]
InKryption: It doesn't assume it's the first resize, it's assuming there is a concrete mapping from the allocation size to the file size, and a concrete mapping from the file size to the mmap size - which there is.
That is to say, there is a one to many relationship between the mmap size & file size, and a one to many relationship between the file size & allocation size.
The logic is such that it succeeds only if the mmap size remains the same or lower.
To demonstrate how this holds up, observe the compile log output of this code:
```zig
comptime {
    @setEvalBranchQuota(1_000_000);
    for (.{
        0,

        std.mem.page_size * 1 - @sizeOf(Metadata),
        std.mem.page_size * 1 - 1,
        std.mem.page_size * 1,
        std.mem.page_size * 1 + 1,

        std.mem.page_size * 2 - @sizeOf(Metadata),
        std.mem.page_size * 2 - 1,
        std.mem.page_size * 2,
        std.mem.page_size * 2 + 1,
    }) |size| {
        const file_size = alignedFileSize(size);
        const mmap_size = alignedMmapSize(file_size, 100);
        @compileLog(size, file_size, mmap_size);
    }
}
```
```zig
@as(comptime_int, 0), @as(usize, 16384), @as(usize, 32768)
@as(comptime_int, 16380), @as(usize, 16384), @as(usize, 32768)
@as(comptime_int, 16383), @as(usize, 32768), @as(usize, 49152)
@as(comptime_int, 16384), @as(usize, 32768), @as(usize, 49152)
@as(comptime_int, 16385), @as(usize, 32768), @as(usize, 49152)
@as(comptime_int, 32764), @as(usize, 32768), @as(usize, 49152)
@as(comptime_int, 32767), @as(usize, 49152), @as(usize, 65536)
@as(comptime_int, 32768), @as(usize, 49152), @as(usize, 65536)
@as(comptime_int, 32769), @as(usize, 49152), @as(usize, 65536)
```

src/utils/allocators.zig
```diff
@@ -177,121 +209,191 @@ pub const DiskMemoryAllocator = struct {
...
+
+        const file = self.dir.openFile(file_name, .{ .mode = .read_write }) catch |err| {
+            self.logFailure(err, file_name);
+            return false;
+        };
+        defer file.close();
+
+        file.setEndPos(new_file_aligned_size) catch return false;
+        if (new_mmap_aligned_size < old_mmap_aligned_size) {
+            // we'll never be able to recover this anyway, because the `alignedFileSize` will
```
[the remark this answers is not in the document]
InKryption: The way I refactored it after our conversation was to mmap more memory based on the mmap_ratio in the `alloc` call, which then grants more available but uncommitted memory.

If we allow shrinking the allocation such that the mmap size we derive from it becomes smaller, the rest of the address space becomes unobtainable, since any resizes which would require growing the mmap size is rejected. So keeping the memory around with this design would simply be a potential waste, and possibly even hide out of bounds accesses on many item ptrs by causing the uncommitted memory to be committed vs causing a segmentation fault.

I'm open to adding the real size as metadata to avoid this path if you want, that just comes with the tradeoff of committing more pages earlier the bigger an allocation becomes (`@sizeOf(Metadata)` becomes bigger) - not by a huge amount obviously, but still a tradeoff.

src/utils/allocators.zig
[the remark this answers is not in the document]
InKryption: Ah I see what you mean, good point - originally I had it in such a way that the mmap ratio was being applied to the alignment boundary, but then that was very drastic and would make it only reasonable to have the ratio ranged from 0 to 7.
With that understanding I guess the only reasonable way to do this is to store the available size in the metadata.

src/utils/allocators.zig
```diff
@@ -136,38 +136,70 @@ pub fn RecycleFBA(config: struct {
...
-    count: std.atomic.Value(u64) = std.atomic.Value(u64).init(0),
-
+    dir: std.fs.Dir,
+    logger: sig.trace.Logger,
+    /// The address space mmap'd to a particular file will be at least
+    /// `(file_size * (1000 + mmap_ratio)) / 1000` in integer terms.
+    /// With a value of 0, the mmap'd size will be equal to the file size.
+    /// With a value of 1000, the mmap'd size will be double the file size.
+    /// NOTE: not intended to be changed after initialization.
+    mmap_ratio: MmapRatio = 0,
```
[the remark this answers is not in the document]
InKryption: I think protecting against the 0 footgun is more important; if it's misunderstood, the worst that can happen is that you end up mmapping a bit more than expected, within a relatively low magnitude.
If you think it's confusing, we can rename it to something more descriptive, ie `mmap_ratio_exp_offset`.
Mostly I allowed for thousands of decimal precision to facilitate a granular level of control. I can't predict what range of use cases one may have, this just allows us a bit of fine tuning, and costs relatively little to have - having it be a simple multiple was fairly limiting in the sizes that could be expressed; having the mmap size be more than double or triple the file size just doesn't seem like it would be very common, compared to wanting just a slightly larger margin for resizes.

src/utils/allocators.zig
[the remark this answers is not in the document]
InKryption: 1308c9ef293afb5b3428b1fa8f3887d1ac827318

src/utils/allocators.zig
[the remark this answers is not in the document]
InKryption: 1308c9ef293afb5b3428b1fa8f3887d1ac827318 (see other comment)

src/utils/allocators.zig
```diff
@@ -136,38 +136,78 @@ pub fn RecycleFBA(config: struct {
     };
 }
 
+test "recycle allocator" {
```
0xNineteen, line 139: can we keep all tests on the bottom of the file

src/utils/allocators.zig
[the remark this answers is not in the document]
0xNineteen: if we dont use it, it shouldnt be in the codebase because it will just make things more confusing - would rather remove and will be easy to add back if we need it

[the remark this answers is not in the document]
dnut: I don't have a strong opinion about this being configurable. But I would not use a ratio of 1 (i.e. mmap_ratio = 0). With a ratio of 1, the resize function would only return true if you're resizing to a size that's within 1 page of the original allocation.

I don't think there's much cost of making the number very large. I would probably use a ratio of at least 8. 1024 or larger may also be fine. (i.e. mmap_ratio = 7000 or 1023000)

src/utils/allocators.zig
```diff
@@ -136,38 +136,78 @@ pub fn RecycleFBA(config: struct {
...
-    pub fn init(filepath: []const u8) Self {
-        return Self{
-            .filepath = filepath,
-        };
+    pub const MmapRatio = u32;
+
+    /// Metadata stored at the end of each allocation.
+    const Metadata = extern struct {
+        file_index: u32,
+        mmap_size: packed struct(u8) {
```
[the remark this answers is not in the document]
InKryption: Well, it saves 7 bytes which can instead be used for actual data in the user-facing allocation. The more memory used by the metadata, the sooner a second page is needed.

Also: the page size is always a power of two on all realistic platforms. I'm not sure there's a scenario where we are going to be running on a platform that's exotic enough for that to not be true.
You can technically request to mmap a size that's not aligned to the page size, but it will still just allocate a whole page - the kernel isn't in the business of tracking the size of your allocations at a granularity smaller than a page. So this is imo an entirely reasonable property/assumption to take advantage of.

The main reason this is a packed struct wrapping the integer is because the exotically sized integer isn't allowed in an extern struct. Would use a packed struct, but that makes it act bit-packed and backed by an integer, which has less intuitive properties on the size of how much space it takes up than using an extern struct.

src/utils/allocators.zig
```diff
@@ -136,38 +136,78 @@ pub fn RecycleFBA(config: struct {
...
+            log2: std.math.Log2Int(usize),
+            _padding: enum(u2) { unset = 0 } = .unset,
```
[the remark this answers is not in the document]
InKryption: Yeah currently - it isn't too complicated to make this generic over the size of `usize` though.

src/utils/allocators.zig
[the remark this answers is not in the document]
InKryption: Well it would be good to have a conclusive decision on this: I'm fine with:
* Status quo: no resize is supported.
* Current change: configurable ratio to allow a proportional range of resizability - perhaps better documented and simplified.
* Possible change: hard-code a ratio which we assume to be logical for our use cases.

If you have anything specific in mind, lmk, otherwise I'll just go with the first route, maintaining status quo with no resize support.

src/utils/allocators.zig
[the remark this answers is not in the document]
InKryption: 671c11ff0692942fdffdbf4ca786e9a688aa511a

src/utils/allocators.zig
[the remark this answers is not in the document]
dnut: > Well, it saves 3 bytes which can instead be used for actual data in the user-facing allocation. The more memory used by the metadata, the sooner a second page is needed,

I understand this, it just seems negligible though. Pages are measured in thousands of bytes. The performance gain of this approach is less than 0.1%. It's hard to justify the code complexity. Actually this approach requires extra cpu operations to calculate logs/shifts, so it might even have a net negative impact on performance.

> the page size is always a power of two on all realistic platforms

That doesn't mean the mapped size is a power of two. Regardless of page size, 3 pages has 3 as a prime factor, which means it's not a power of two. To ensure the mapped size is a power of two, the number of pages per mapping would also need to be a power of two.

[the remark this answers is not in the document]
InKryption: Ah, a fair point, I was conflating my alignment math and power²/log₂ math.

src/utils/allocators.zig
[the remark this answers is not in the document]
dnut: > - Current change: configurable ratio to allow a proportional range of resizability - perhaps better documented and simplified.
> - Possible change: hard-code a ratio which we assume to be logical for our use cases.

I think we should go with one of these. Resize is useful, and you already have an implementation that looks good to me. Among these options, I'm fine deferring to either of your preferences @InKryption  and @0xNineteen.

If I have to make a choice, I would select simplified configurability, as in 
`mmap_size == mmap_ratio * file_size`. For me, this is the simplest/easiest to understand. But I see the advantages to other approaches too, so I'm not going to demand a specific approach here.

src/utils/allocators.zig
dnut: I almost left the same comment on this pr. But it grew on me a little. There is something nice about the test being immediately adjacent to the struct being tested. It's a little unexpected though since the rest of the project does not do this.

src/utils/allocators.zig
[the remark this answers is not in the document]
0xNineteen: `mmap_ratio * file_size` with a default ratio of 1 sounds good to me

src/utils/allocators.zig
0xNineteen: it has not grown on me hahaha

InKryption: e5e987370512681ddbfe300f4d18a9edcd52f626

src/utils/allocators.zig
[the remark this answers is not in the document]
InKryption: fc14c0746a8dca4c94fac28ab548c7a4795a3df0

src/utils/allocators.zig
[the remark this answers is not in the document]
InKryption: fc14c0746a8dca4c94fac28ab548c7a4795a3df0

InKryption:
Hit an interesting path in the kcov job. Re-running it to get this green, but here is the link to the log: https://github.com/Syndica/sig/actions/runs/11058573690/job/30724902400?pr=259

dnut (approved):

src/accountsdb/readme.md
InKryption: 308ae941d92dc508b041b7af413cecd25b4b5ac1

0xNineteen (approved):
lgtm

PR #262, opened 2024-08-30
feat(allocators): RecycleFBA: configurable inherent thread safety

dnut:
This implements thread safety within RecycleFBA by acquiring and releasing the lock in the alloc, free, and resize functions. Code that uses the allocator no longer needs to directly access the mutex in the RecycleFBA

Thread safety is configurable using the same approach as the GPA. The type takes a comptime config parameter. Thread safety is on by default if multithreading is supported by the platform.

pinging @dadepo for visibility since he is also using this allocator in his branch.

### DIFF
--- a/src/geyser/core.zig
+++ b/src/geyser/core.zig
@@ -21,7 +21,7 @@ const RecycleFBA = sig.utils.allocators.RecycleFBA;
 const PIPE_MAX_SIZE_PATH = "/proc/sys/fs/pipe-max-size";
 
 pub const AccountPayload = struct {
-    // used to know how much to allocate to read the full data slice
+    /// used to know how much to allocate to read the full data slice
     len: u64,
     payload: VersionedAccountPayload,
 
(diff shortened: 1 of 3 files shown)


src/utils/allocators.zig
```diff
@@ -1,123 +1,140 @@
...
-    mux: std.Thread.Mutex = .{},
-
-    const Record = struct { is_free: bool, buf: [*]u8, len: u64 };
-    const Self = @This();
-
-    pub fn init(backing_allocator: std.mem.Allocator, n_bytes: u64) !Self {
-        const buf = try backing_allocator.alloc(u8, n_bytes);
-        const alloc_allocator = std.heap.FixedBufferAllocator.init(buf);
-        const records = std.ArrayList(Record).init(backing_allocator);
+pub fn RecycleFBA(config: struct {
```
dadepo, line 5: More of a question: Any reason to have this as a type function? given that it does not take any "generic" parameter

src/geyser/core.zig
```diff
@@ -170,25 +173,19 @@ pub const GeyserWriter = struct {
         const total_len = bincode.sizeOf(payload, .{});
 
         // obtain a memory to write to
-        self.io_free_fba.mux.lock();
```
dadepo, line 173: +1 to improving this. I found it a bit clunky when I made the initial switch to the recycle allocator

src/utils/allocators.zig
dadepo: > given that it does not take any "generic" parameter

type function does not need to take "generic" parameter. They can take parameter of `type` or any compile time known value.

0xNineteen (approved):

PR #264, opened 2024-09-04
refactor(ledger)

dnut:


src/ledger/hashmap_db.zig
```diff
@@ -174,15 +186,30 @@ pub fn SharedHashMapDB(comptime column_families: []const ColumnFamily) type {
         pub const WriteBatch = struct {
             allocator: Allocator,
             instructions: std.ArrayListUnmanaged(Instruction),
+            executed: *bool,
```
yewman, line 189: How come this guy is a pointer to a bool instead of just a bool?

src/ledger/reader.zig
```diff
@@ -515,7 +520,11 @@ pub const BlockstoreReader = struct {
             const signature = transaction.signatures[0];
             txns_with_statuses.appendAssumeCapacity(.{
                 .transaction = transaction,
-                .meta = try self.db.get(schema.transaction_status, .{ signature, slot }) orelse
+                .meta = try self.db.get(
+                    schema.transaction_status,
+                    .{ signature, slot },
+                    self.allocator,
```
yewman, line 526: With these methods, it might be best to update to the sig convention of passing allocators as the first non-self argument to function calls unless there is a compelling reason not to

src/ledger/meta.zig
```diff
@@ -60,7 +60,7 @@ pub const SlotMeta = struct {
         };
     }
 
-    pub fn deinit(self: *Self) void {
+    pub fn deinit(self: Self) void {
```
yewman, line 63: What is the reasoning behind passing by value rather than by reference? Honestly just curious here as my (probably naive) understanding tells me that for larger structs, passing by reference is usually preferred

src/ledger/rocksdb.zig
```diff
@@ -98,10 +112,15 @@ pub fn RocksDB(comptime column_families: []const ColumnFamily) type {
             );
         }
 
-        pub fn get(self: *Self, comptime cf: ColumnFamily, key: cf.Key) anyerror!?cf.Value {
+        pub fn get(
```
yewman, line 115:
```suggestion
        pub fn get(
            self: *Self,
            allocator: Allocator,
            comptime cf: ColumnFamily,
            key: cf.Key,
        ) anyerror!?cf.Value {
```

yewman (changes requested):
Couple of minor questions / suggestions otherwise looks good :)

src/ledger/meta.zig
dnut: The main thing I was trying to do here was avoid needing a *mutable* pointer. That was an annoying constraint the caused issues in calling code that's unnecessary.

I passed by value instead of const pointer because the core zig developers recommend not to pass const pointers in cases like this, and instead you should pass by value allow the compiler to optimize this. For example: https://github.com/ziglang/zig/pull/19770

After our the recent discussions with the sig team though, I'm not really convinced by their arguments any more, and I think both approaches have their merits.

0xNineteen (approved):
lgtm after yews comments

src/ledger/hashmap_db.zig
dnut: The deinit logic of write batch depends on whether or not it has been executed. If you use the wrong logic, there may be memory errors.

The goal of using a pointer was to make it safe to call deinit from the context where it is initialized, even if it is passed by value into the context where it is executed.

src/ledger/rocksdb.zig
dnut: I'm fine with changing this, but just want to explain my reasoning first.

I prefer the column family to come first because I think it communicates the meaning of the code more clearly.

When written like this:
```zig
db.get(schema.slot_meta, slot, self.allocator);
```
It's almost like the column family is part of the method name, which is reasonable since it determines the types of the parameters and return of the function.
```zig
db.getSlotMeta(slot, self.allocator);
```

The allocator argument tells you less about the meaning of the code. It's more of a minor detail, which is why it feels cleaner to me at the end.

src/ledger/reader.zig
dnut: f59c2103b90928234a562f530f0164ab957886e0

src/ledger/rocksdb.zig
dnut: f59c2103b90928234a562f530f0164ab957886e0

0xNineteen (approved):
lgtm - saw you addressed yews small comments and since hes sick i resolved them - should be good to merge

PR #265, opened 2024-09-04
refactor(bincode): extract hashmap and arraylist code to field configs

dnut:
This takes all the custom logic from bincode.read and bincode.write that was written for hashmaps and arraylists, and it moves it to functions that return it as a FieldConfig in bincode/hashmap.zig and bincode/arraylist.zig. This `hashmap` field in FieldConfig is no longer needed, since this can be specified directly instead.

The `getConfig` function automatically applies the hashmap and arraylist FieldConfig for compatible types that don't have another config already. Long term there may be a better or more flexible way to include standard configs (maybe in Params?) but this was the easiest place to put it for now.

### DIFF
--- a/src/accountsdb/snapshots.zig
+++ b/src/accountsdb/snapshots.zig
@@ -1030,11 +1030,9 @@ pub const AccountsDbFields = struct {
     rooted_slots: std.ArrayListUnmanaged(Slot),
     rooted_slot_hashes: std.ArrayListUnmanaged(SlotAndHash),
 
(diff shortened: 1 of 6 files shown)


InKryption (approved):
No notes. Will leave it to you if you want to rename `FieldConfig` in this PR.

dnut:
> No notes. Will leave it to you if you want to rename `FieldConfig` in this PR.

I decided to merge it as is, since I prefer iterating rapidly with smaller commits, once they are good to go. Also I wasn't sure what would be the best name. We can open another pr to address the rename.

PR #266, opened 2024-09-05
feat(gossip): data source metrics

### LINKED ISSUE #121: Gossip: Data Source Metrics
right now we just insert new gossip data directly into the table

- would be nice to have metrics to know where our data is being updated from: either from push messages, pull responses, or local 

- would also be nice to know how many crds values are being produced for each pull request received

this would be a new prometheus metric

sonicfromnewyoke:
attempt to close #121

### DIFF
--- a/src/gossip/service.zig
+++ b/src/gossip/service.zig
@@ -1251,6 +1251,7 @@ pub const GossipService = struct {
 
             for (pull_requests.items) |*req| {
                 _ = gossip_table.insert(req.value, now) catch {};
+                self.stats.table_n_updates_pull_request.inc();
                 gossip_table.updateRecordTimestamp(req.value.id(), now);
             }
         }
@@ -1441,6 +1442,7 @@ pub const GossipService = struct {
                         pull_message.gossip_values[index],
                         now,
                     ) catch {};
+                    self.stats.table_n_updates_pull_response.inc();
                 }
 
                 // update the contactInfo timestamps of the successful inserts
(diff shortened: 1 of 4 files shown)


Sobeston:
Looks all good so far :+1:

src/gossip/service.zig
```diff
@@ -1253,6 +1253,7 @@ pub const GossipService = struct {
                 _ = gossip_table.insert(req.value, now) catch {};
                 gossip_table.updateRecordTimestamp(req.value.id(), now);
             }
+            self.stats.table_n_updates_pull_req.add(pull_requests.items.len);
```
0xNineteen, line 1256: `gossip_table.insert` will fail with an error if the value was not inserted which is not accounted for in this case -- would need to add logic to the catch block

src/gossip/service.zig
```diff
@@ -1442,6 +1443,7 @@ pub const GossipService = struct {
                         now,
                     ) catch {};
                 }
+                self.stats.table_n_updates_pull_resp.add(timeout_indexs.items.len);
```
0xNineteen, line 1446: same here -- but we need to look at `insertValues` in the lines above

src/gossip/service.zig
```diff
@@ -1769,6 +1771,7 @@ pub const GossipService = struct {
 
         while (push_msg_queue.popOrNull()) |gossip_value| {
             _ = gossip_table.insert(gossip_value, now) catch {};
+            self.stats.table_n_updates_push_msg.inc();
```
0xNineteen, line 1774: same here

0xNineteen (changes requested):
the main thing we want to track is when values are successfully **inserted** and **updated** in two separate metrics

0xNineteen:
closed for #276

PR #269, opened 2024-09-06
ci(scripts): better remove_unused.py output

dnut:
This adds `--check` and `--verbose` flags to remove_unused.py, and uses `--check` in CI.

- `--check`: don't actually edit the files, and instead provide a more helpful output including the line number and the actual line of code that should be removed.
- `--verbose`: log filenames that had 0 unused imports (previously this was the default behavior, but now it's disabled by default)

```bash
python scripts/remove_unused.py --check src
```
```
Found 2 unused import(s) in src/core/pubkey.zig:
75 | const ChaChaRng = sig.rand.ChaChaRng;
76 | const Epoch = sig.core.Epoch;

Found 3 unused import(s) in src/core/entry.zig:
4 | const Pubkey = sig.core.Pubkey;
5 | const Epoch = sig.core.Epoch;
6 | const AccountInFile = sig.accounts_db.accounts_file.AccountInFile;

total unused found: 5
```
---------------
```bash
python scripts/remove_unused.py  src
```
```
src/core/pubkey.zig 2
src/core/entry.zig 3
src/core/entry.zig
removed this iteration: 5

removed this iteration: 0

total unused found: 5
```

### DIFF
--- a/scripts/remove_unused.py
+++ b/scripts/remove_unused.py
@@ -78,7 +78,7 @@ while lines_removed_this_time > 0:
                         for i, line in enumerate(orig_lines)
                         if i not in lines_to_drop
                     )
-                print(path, num_lines_to_remove)
+                print(f"Removed {num_lines_to_remove} unused import(s) in {path}")
                 os.system(f"zig fmt {path}")
         elif args.verbose:
             print(path, num_lines_to_remove)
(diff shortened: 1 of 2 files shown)


scripts/remove_unused.py
```diff
@@ -34,40 +38,60 @@
...
-    print()
+    if args.check:
+        break
+    else:
+        print("removed this iteration:", lines_removed_this_time)
+        print()
 
-print("total lines removed:", total_num_lines_removed)
-if (total_num_lines_removed > 0):
+print("total unused found:", total_num_lines_removed)
```
dadepo, line 95: If the --check flag is not passed, the output would be something like

```
src/trace/log.zig 1
removed this iteration: 1

removed this iteration: 0

total unused found: 1
```

Might be nice to have the message say Removed, instead of Found. And something about the files where the unused lines has been removed.

Something like:

`Removed x unused import(s) in src/trace/log.zig:`

dadepo (approved):
LGTM

scripts/remove_unused.py
dnut: 4d21abecd72dcdf48bbd79eb95680da047435f22

yewman (approved):
lgtm

PR #270, opened 2024-09-10
fix(blockstore): remove stray blockstore file

dnut:
The `blockstore` folder has already been renamed to `ledger`. Somehow, `blockstore/blockstore.zig` snuck its way back in the repo. It's no longer in use, since `ledger/blockstore.zig` has replaced it. So this file should simply be deleted.


sonicfromnewyoke (approved):
LGTM

PR #272, opened 2024-09-10
feat(core): implement LeaderScheduleCache

yewman:
- The current leader schedule implementation does not provide functionality to handle epoch transitions. 
- LeaderScheduleCache keeps a map of Epoch -> LeaderSchedule wrapped in a RwMux and provides methods to both access and compute LeaderSchedule's for new Epoch's
- If a caller has the required information to compute a LeaderSchedule (BankFields or RpcClient) they can use one of the following methods to fetch a slot leader, computing the LeaderSchedule if required:
    - `getSlotLeaderMaybeCompute(self: *LeaderScheduleCache, slot: Slot, bank: *const BankFields) !sig.core.Pubkey`
    - `getSlotLeaderMaybeComputeRpc(self: *LeaderScheduleCache, slot: Slot, rpc_client: *RpcClient) !sig.core.Pubkey`
- Otherwise, a caller can use
    - `getSlotLeaderNoCompute(self: *LeaderScheduleCache, slot: Slot) ?Pubkey`
- The LeaderScheduleCache keeps an EpochSchedule for computing the Epoch and slot index for a given Slot
    - as far as I know, this object will not change during runtime and so won't need any sync primitives

Thoughts on Abstractions: 
- Drew made some good comments about potential abstractions
- In my mind, implementation of these suggestions should be based around some sort of team discussion about how we approach abstractions and interfacing in sig
- They can improve decoupling and make it easier to mock complex functionality for testing, however, they can also make code more verbose and difficult follow
- I think for now should just run with the basic implementation and implement the abstractions in the future if they:
    - become required for testing
    - simplify the existing code
    - or form part of a sig wide design philosophy


dnut:
> If a caller has the required information to compute a LeaderSchedule (BankFields or RpcClient) they can use one of the following methods to fetch a slot leader, computing the LeaderSchedule if required: 

Why does the caller need to provide these things? Why not include one of those dependencies as a field in LeaderScheduleCache itself?

yewman:
> > If a caller has the required information to compute a LeaderSchedule (BankFields or RpcClient) they can use one of the following methods to fetch a slot leader, computing the LeaderSchedule if required:
> 
> Why does the caller need to provide these things? Why not include one of those dependencies as a field in LeaderScheduleCache itself?
> 
> Do you think something like this could work?
> 
> ```zig
> const BankFieldsLeaderScheduleProvider = struct { 
>      cache: LeaderScheduleCache,
>      bank_fields_provider: struct {
>           fn getBankFields(Epoch) BankFields
>      },
> 
>      pub fn provider() LeaderScheduleProvider
> };
> 
> const RpcLeaderScheduleProvider = struct {
>      cache: LeaderScheduleCache,
>      rpc: RpcClient,
> 
>      pub fn provider() LeaderScheduleProvider
> };
> 
> /// Just contains the leader schedule for an epoch
> const LeaderSchedule = struct {
>      slot_leaders: []Pubkey
> 
>      pub fn provider() LeaderScheduleProvider
> };
> 
> /// only used internally to the first two structs above
> /// does not need any methods that calculate the leader schedule
> /// can just be something like a hashmap or lru cache
> const LeaderScheduleCache = Lru(Epoch, LeaderSchedule);
> 
> const LeaderScheduleProvider = struct {
>     state: *anyopaque,
>     slotLeaderFn: *const fn slotLeader(*anyopaque, Slot) Pubkey,
> }
> ```
> 
> The idea would be that you can instantiate one of the provider structs in cmd.zig and then any code that needs to calculate a leader schedule only depends on LeaderScheduleProvider, which only needs you to provide a slot to get its leader.

yewman:
> > If a caller has the required information to compute a LeaderSchedule (BankFields or RpcClient) they can use one of the following methods to fetch a slot leader, computing the LeaderSchedule if required:
> 
> Why does the caller need to provide these things? Why not include one of those dependencies as a field in LeaderScheduleCache itself?
> 
> Do you think something like this could work?
> 
> ```zig
> const BankFieldsLeaderScheduleProvider = struct { 
>      cache: LeaderScheduleCache,
>      bank_fields_provider: struct {
>           fn getBankFields(Epoch) BankFields
>      },
> 
>      pub fn provider() LeaderScheduleProvider
> };
> 
> const RpcLeaderScheduleProvider = struct {
>      cache: LeaderScheduleCache,
>      rpc: RpcClient,
> 
>      pub fn provider() LeaderScheduleProvider
> };
> 
> /// Just contains the leader schedule for an epoch
> const LeaderSchedule = struct {
>      slot_leaders: []Pubkey
> 
>      pub fn provider() LeaderScheduleProvider
> };
> 
> /// only used internally to the first two structs above
> /// does not need any methods that calculate the leader schedule
> /// can just be something like a hashmap or lru cache
> const LeaderScheduleCache = Lru(Epoch, LeaderSchedule);
> 
> const LeaderScheduleProvider = struct {
>     state: *anyopaque,
>     slotLeaderFn: *const fn slotLeader(*anyopaque, Slot) Pubkey,
> }
> ```
> 
> The idea would be that you can instantiate one of the provider structs in cmd.zig and then any code that needs to calculate a leader schedule only depends on LeaderScheduleProvider, which only needs you to provide a slot to get its leader.

No reason in particular, I like your suggested approach though.
I am just wondering if the following circumstance could ever cause a issue.
Consider using a BankFieldsLeaderScheduleProvider, which is dependent on some bank provider `B`, now the LeaderScheduler provider for all users will be dependent on the state of the bank provider `B`. If there exists some other process who has a bank which the bank provider `B` does not have yet, and this process wants to obtain a valid result from the leader schedule, it will have to wait an unknown amount of time until the bank provider `B` has the same data. I am not sure if this would ever actually present a problem, just a thought. I'll have a think about it a little more, if it is an issue there might be a nice way to modify the above slightly to manage such instances

src/cmd/cmd.zig
```diff
@@ -639,10 +638,16 @@ fn validator() !void {
...
-    // leader schedule
-    var leader_schedule = try getLeaderScheduleFromCli(allocator) orelse
-        try leaderScheduleFromBank(allocator, snapshot.bank.bank_fields);
-    const leader_provider = leader_schedule.provider();
+    // leader schedule cache
+    var leader_schedule_cache = LeaderScheduleCache.init(allocator, snapshot.bank.bank_fields.epoch_schedule);
+    if (try getLeaderScheduleFromCli(allocator) orelse null) |leader_schedule| {
+        try leader_schedule_cache.insertLeaderSchedule(snapshot.bank.bank_fields.epoch, leader_schedule);
+    } else {
+        _ = try leader_schedule_cache.getSlotLeaderMaybeCompute(snapshot.bank.bank_fields.slot, snapshot.bank.bank_fields);
```
0xNineteen, line 646: the double usage of this method is a little confusing - maybe create a helper function `insertLeaderScheduleFromBank` which calls `LeaderSchedule.fromBank(` so its more clear

src/core/leader_schedule.zig
```diff
@@ -9,205 +9,318 @@ const Pubkey = sig.core.Pubkey;
...
-        .slot_leaders = slot_leaders,
-        .start_slot = epoch_start_slot,
-    };
-}
+    pub fn getSlotLeaderNoCompute(self: *LeaderScheduleCache, slot: Slot) ?Pubkey {
+        const epoch, const slot_index = self.epoch_schedule.getEpochAndSlotIndex(slot);
+        return self.getSlotLeader(epoch, slot_index);
+    }
+
+    pub fn getSlotLeaderMaybeCompute(self: *LeaderScheduleCache, slot: Slot, bank: *const BankFields) !sig.core.Pubkey {
```
0xNineteen, line 73: maybe better name is `getOrComputeSlotLeader`

src/core/leader_schedule.zig
```diff
@@ -9,205 +9,318 @@ const Pubkey = sig.core.Pubkey;
...
-
-    _, const slot_index = bank_fields.epoch_schedule.getEpochAndSlotIndex(bank_fields.slot);
-    const epoch_start_slot = bank_fields.slot - slot_index;
-    return SingleEpochLeaderSchedule{
-        .allocator = allocator,
-        .slot_leaders = slot_leaders,
-        .start_slot = epoch_start_slot,
-    };
-}
+    pub fn getSlotLeaderNoCompute(self: *LeaderScheduleCache, slot: Slot) ?Pubkey {
```
0xNineteen, line 68: can then change this to just `getSlotLeader`

src/core/leader_schedule.zig
```diff
@@ -9,205 +9,318 @@ const Pubkey = sig.core.Pubkey;
...
+        const epoch, const slot_index = self.epoch_schedule.getEpochAndSlotIndex(slot);
+        return self.getSlotLeader(epoch, slot_index);
+    }
+
+    pub fn getSlotLeaderMaybeCompute(self: *LeaderScheduleCache, slot: Slot, bank: *const BankFields) !sig.core.Pubkey {
+        const epoch, const slot_index = self.epoch_schedule.getEpochAndSlotIndex(slot);
```
0xNineteen, line 74: can move to just call `getSlotLeaderNoCompute`

src/cmd/cmd.zig
```diff
@@ -739,8 +744,16 @@ fn shredCollector() !void {
 
     // leader schedule
     // NOTE: leader schedule is needed for the shred collector because we skip accounts-db setup
-    var leader_schedule = try getLeaderScheduleFromCli(allocator) orelse @panic("No leader schedule found");
-    const leader_provider = leader_schedule.provider();
+    var leader_schedule_cache = LeaderScheduleCache.init(allocator, try EpochSchedule.default());
+
+    // This is a sort of hack to get the epoch of the leader schedule and then insert into the cache
```
0xNineteen, line 749: do we want to implement this before merging to main?

src/core/leader_schedule.zig
```diff
@@ -9,205 +9,318 @@ const Pubkey = sig.core.Pubkey;
...
+                bank,
+            );
+        }
+        return entry.value_ptr.slot_leaders[slot_index];
     }
 
-    const Record = struct { slot: Slot, key: Pubkey };
+    pub fn getSlotLeaderMaybeComputeRpc(self: *LeaderScheduleCache, slot: Slot, rpc_client: *RpcClient) !sig.core.Pubkey {
+        const epoch, const slot_index = self.epoch_schedule.getEpochAndSlotIndex(slot);
+        if (self.getSlotLeader(epoch, slot_index)) |leader| return leader;
```
0xNineteen, line 96: same here

src/core/leader_schedule.zig
```diff
@@ -9,205 +9,318 @@ const Pubkey = sig.core.Pubkey;
...
+        defer leader_schedule_lg.unlock();
 
-    var rpc_leader_iter = rpc_leader_schedule.iterator();
-    while (rpc_leader_iter.next()) |entry| {
-        const key = try Pubkey.fromString(entry.key_ptr.*);
-        for (entry.value_ptr.*) |slot| {
-            leaders[leaders_index] = .{ .slot = slot, .key = key };
-            leaders_index += 1;
+        const entry = try leader_schedules.getOrPut(epoch);
+        if (!entry.found_existing) {
```
0xNineteen, line 102: maybe want to change this to `assert !entry.found_existing` since it should never exist if it didnt exist calling noCompute

src/core/leader_schedule.zig
```diff
@@ -9,205 +9,318 @@ const Pubkey = sig.core.Pubkey;
...
+        const leader_schedules, var leader_schedule_lg = self.leader_schedules_rw.writeWithLock();
+        defer leader_schedule_lg.unlock();
 
-    var rpc_leader_iter = rpc_leader_schedule.iterator();
-    while (rpc_leader_iter.next()) |entry| {
-        const key = try Pubkey.fromString(entry.key_ptr.*);
-        for (entry.value_ptr.*) |slot| {
-            leaders[leaders_index] = .{ .slot = slot, .key = key };
-            leaders_index += 1;
+        const entry = try leader_schedules.getOrPut(epoch);
```
0xNineteen, line 101: maybe want to have an `insertLeaderSchedule` method call here

src/core/leader_schedule.zig
```diff
@@ -9,205 +9,318 @@ const Pubkey = sig.core.Pubkey;
...
-        .slot_leaders = leader_pubkeys,
-        .start_slot = start_slot,
-    };
-}
-
-pub const StakedNode = struct { id: Pubkey, stake: u64 };
+    /// This method exists purely to allow using leader schedule from cli, manually inserting it into the
+    /// cache with its associated epoch.
+    pub fn insertLeaderSchedule(self: *LeaderScheduleCache, epoch: Epoch, leader_schedule: LeaderSchedule) !void {
+        const leader_schedules, var leader_schedule_lg = self.leader_schedules_rw.writeWithLock();
```
0xNineteen, line 129: should we have a `if (leader_schedules.count() == MAX_CACHED_LEADER_SCHEDULES) {` check in this method?

src/core/leader_schedule.zig
```diff
@@ -9,205 +9,318 @@ const Pubkey = sig.core.Pubkey;
...
 
-    std.mem.sortUnstable(Record, leaders, {}, struct {
-        fn gt(_: void, lhs: Record, rhs: Record) bool {
-            return switch (std.math.order(lhs.slot, rhs.slot)) {
-                .gt => false,
-                else => true,
-            };
-        }
-    }.gt);
+    fn getSlotLeader(self: *LeaderScheduleCache, epoch: Epoch, slot_index: usize) ?Pubkey {
```
0xNineteen, line 115: probs dont even need this method - can move it into getSlotLeaderNoCompute

src/core/leader_schedule.zig
```diff
@@ -9,205 +9,318 @@ const Pubkey = sig.core.Pubkey;
...
+
+        return .{
+            .allocator = allocator,
+            .slot_leaders = slot_leaders,
+        };
     }
 
-    return slot_leaders;
-}
+    pub fn computeSlotLeaders(
```
0xNineteen, line 218: maybe `fromStakedNodes` is better name?

src/core/leader_schedule.zig
```diff
@@ -9,205 +9,318 @@ const Pubkey = sig.core.Pubkey;
...
-        return if (index >= self.slot_leaders.len) null else self.slot_leaders[index];
+    pub fn getUniqueLeaders(self: *LeaderScheduleCache) !std.AutoArrayHashMap(Pubkey, void) {
+        const leader_schedules, var leader_schedules_lg = self.leader_schedules_rw.readWithLock();
+        defer leader_schedules_lg.unlock();
+
+        var unique_leaders = std.AutoArrayHashMap(Pubkey, void).init(self.allocator);
+
+        for (leader_schedules.values()) |leader_schedule| {
+            for (leader_schedule.slot_leaders) |leader| {
+                try unique_leaders.put(leader, void{});
```
0xNineteen, line 57:
```suggestion
                try unique_leaders.put(leader, {});
```

0xNineteen (changes requested):
overall lgtm - just small naming + reorg things to clean up

0xNineteen:
just to add my 5cents to the abstraction talk - not a big fan of abstractions, makes the code more confusing to follow - would prefer to stick with the method approach rn

src/core/leader_schedule.zig
yewman: 9295eea8

src/core/leader_schedule.zig
yewman: 9295eea8

src/core/leader_schedule.zig
yewman: 9295eea8

src/core/leader_schedule.zig
yewman: 265ed12

src/core/leader_schedule.zig
yewman: 265ed12

src/core/leader_schedule.zig
yewman: 265ed12

src/core/leader_schedule.zig
yewman: 265ed12

src/core/leader_schedule.zig
yewman: 265ed12

src/core/leader_schedule.zig
yewman: 265ed12

src/core/leader_schedule.zig
yewman: 265ed12

src/core/leader_schedule.zig
yewman: 265ed12

src/core/leader_schedule.zig
yewman: 265ed12

src/cmd/cmd.zig
yewman: I think the comment was outdated, the 'hack' is referring to how we get the epoch to insert the leader schedule and I think its fine for now

dnut (changes requested):
My initial comment was sort of a combination of two things:
- layering on unnecessary abstractions
- some guiding principles about structuring dependencies

Even if we forget about the abstractions, I still have opinions about how this should be structured. I started writing up an explanation but it's a lot of disorganized rambling paragraphs. I decided to just implement my suggestions instead. I'm curious what you think about this code. I'm happy to refine the paragraphs and post it, or hop on a huddle to explain my thought process.

https://github.com/Syndica/sig/pull/296

src/cmd/cmd.zig
```diff
@@ -639,10 +638,16 @@ fn validator() !void {
         geyser_writer,
     );
 
-    // leader schedule
-    var leader_schedule = try getLeaderScheduleFromCli(allocator) orelse
-        try leaderScheduleFromBank(allocator, snapshot.bank.bank_fields);
-    const leader_provider = leader_schedule.provider();
+    // leader schedule cache
+    var leader_schedule_cache = LeaderScheduleCache.init(allocator, snapshot.bank.bank_fields.epoch_schedule);
+    if (try getLeaderScheduleFromCli(allocator) orelse null) |leader_schedule| {
```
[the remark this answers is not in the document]
yewman: d094aea

src/cmd/cmd.zig
yewman: d094aea

0xNineteen:
looked over both prs - they are very similar and can see the pros/cons of both - if we want a tie breaker - i think drews changes make the most sense - caches dont usually have logic to populate itself in them and overall it makes that easy to see/understand

0xNineteen (approved):
one small change - otherwise lgtm

src/core/leader_schedule.zig
```diff
@@ -8,206 +8,241 @@ const Epoch = sig.core.Epoch;
...
-        return if (index >= self.slot_leaders.len) null else self.slot_leaders[index];
+    pub fn slotLeaderProvider(self: *Self) SlotLeaderProvider {
+        return SlotLeaderProvider.init(self, LeaderScheduleCache.slotLeader);
     }
 
-    pub fn provider(self: *Self) SlotLeaderProvider {
-        return SlotLeaderProvider.init(self, Self.getLeader);
+    pub fn put(self: *Self, epoch: Epoch, leader_schedule: LeaderSchedule) !void {
+        self.rwlock.lock();
+        defer self.rwlock.unlock();
```
0xNineteen, line 46: do we want to respect `MAX_CACHED_LEADER_SCHEDULES` here?

src/core/leader_schedule.zig
```diff
@@ -8,206 +8,241 @@ const Epoch = sig.core.Epoch;
...
-    start_slot: Slot,
+/// LeaderScheduleCache is a cache of leader schedules for each epoch.
+/// Leader schedules are expensive to compute, so this cache is used to avoid
+/// recomputing leader schedules for the same epoch.
+/// LeaderScheduleCache also keeps a copy of the epoch_schedule so that it can
+/// compute epoch and slot index from a slot.
+pub const LeaderScheduleCache = struct {
+    allocator: std.mem.Allocator,
+    epoch_schedule: EpochSchedule,
+    rwlock: std.Thread.RwLock,
```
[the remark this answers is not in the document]
yewman: yeah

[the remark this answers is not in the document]
yewman: 9ef544

src/core/leader_schedule.zig
yewman: d055f2

yewman: c916da

0xNineteen (approved):
:shipit:

PR #273, opened 2024-09-12
comments(blockstore): various questions

dadepo:
Intention of this PR is not to merge (hence why it opened as draft) but to surface questions. 

Some answers to the questions may require making actual changes, which can then be done in a different PR

### DIFF
--- a/src/ledger/insert_shred.zig
+++ b/src/ledger/insert_shred.zig
@@ -733,6 +733,8 @@ pub const ShredInserter = struct {
                     "Received *last* shred index {} less than previous shred index {}, and slot {} is not full, marking slot dead",
                     .{ shred_index, slot_meta.received, slot },
                 );
+                // QUESTION: Should the function not return at this point?
+                // Given we have a dead slots and we have saved that.
                 try write_batch.put(schema.dead_slots, slot, true);
             }
 
--- a/src/ledger/insert_shred.zig
+++ b/src/ledger/insert_shred.zig
@@ -1,5 +1,5 @@
 const std = @import("std");
-const sig = @import("../sig.zig");
+const sig = @import("sig.zig");
 
 const ledger = sig.ledger;
 const meta = ledger.meta;
@@ -874,7 +874,7 @@ pub const ShredInserter = struct {
     ) bool {
         const shred_index: u64 = @intCast(shred.fields.common.index);
         // QUESTION: is the "shred_index < slot_meta.consecutive_received_from_0" check needed?
-        // Ie would there ever be a case where data_index.contains(shred_index) is true but 
+        // Ie would there ever be a case where data_index.contains(shred_index) is true but
         // shred_index < slot_meta.consecutive_received_from_0 is false? If not then is data_index.contains(shred_index)
         // not enough check?
         // Also if shred_index > slot_meta.consecutive_received_from_0 is true and also data_index.contains(shred_index) is true
@@ -898,6 +898,9 @@ pub const ShredInserter = struct {
         const shred_index_u32 = shred.fields.common.index;
         const shred_index_u64: u64 = @intCast(shred_index_u32);
         const is_last_in_slot = shred.isLastInSlot();
+        // QUESTION/SUGGESTION rename this to is_valid (or similar) instead of sanitize as santize gives the
+        // impression this is fixing up things and updating the shred, when it is only running a bunch of
+        // validation.
         assertOk(shred.sanitize());
 
         // Check that we do not receive a shred with either:
--- a/src/ledger/insert_shred.zig
+++ b/src/ledger/insert_shred.zig
@@ -690,6 +690,7 @@ pub const ShredInserter = struct {
 
         const index_meta_working_set_entry =
             try self.getIndexMetaEntry(self.allocator, slot, index_working_set, index_meta_time_us);
+
         const index_meta = &index_meta_working_set_entry.index;
         const slot_meta_entry = try self.getSlotMetaEntry(
             slot_meta_working_set,
@@ -805,6 +806,7 @@ pub const ShredInserter = struct {
             if (try self.db.get(schema.index, slot)) |item| {
                 entry.value_ptr.* = .{ .index = item };
             } else {
+                // QUESTION: Find out at what point this get persisted?
                 entry.value_ptr.* = IndexMetaWorkingSetEntry.init(allocator, slot);
             }
         }
@@ -869,6 +871,12 @@ pub const ShredInserter = struct {
         data_index: *meta.ShredIndex,
     ) bool {
         const shred_index: u64 = @intCast(shred.fields.common.index);
+        // QUESTION: is the "shred_index < slot_meta.consecutive_received_from_0" check needed?
+        // Ie would there ever be a case where data_index.contains(shred_index) is true but 
+        // shred_index < slot_meta.consecutive_received_from_0 is false? If not then is data_index.contains(shred_index)
+        // not enough check?
+        // Also if shred_index > slot_meta.consecutive_received_from_0 is true and also data_index.contains(shred_index) is true
+        // is it a valid case case for data_index.contains(shred_index) to be true but shred_index > slot_meta.consecutive_received_from_0.
         return shred_index < slot_meta.consecutive_received_from_0 or
             data_index.contains(shred_index);
     }
--- a/src/ledger/insert_shred.zig
+++ b/src/ledger/insert_shred.zig
@@ -50,6 +50,7 @@ pub const ShredInserter = struct {
     logger: sig.trace.Logger,
     db: BlockstoreDB,
     lock: Mutex,
+    // QUESTION: What is the max_root?
     max_root: Atomic(u64), // TODO shared
     metrics: BlockstoreInsertionMetrics,
 
@@ -81,6 +82,7 @@ pub const ShredInserter = struct {
     ///
     /// This function updates the following column families:
     ///   - [`schema.dead_slots`]: mark a shred as "dead" if its meta-data indicates
+    ///     QUESTION: What does replying a shred mean?
     ///     there is no need to replay this shred.  Specifically when both the
     ///     following conditions satisfy,
     ///     - We get a new shred N marked as the last shred in the slot S,
@@ -89,6 +91,7 @@ pub const ShredInserter = struct {
     ///     - The slot is not currently full
     ///     It means there's an alternate version of this slot. See
     ///     `check_insert_data_shred` for more details.
+    ///    // QUESTION: How are these various family corrolated with one another together?
     ///   - [`schema.shred_data`]: stores data shreds (in check_insert_data_shreds).
     ///   - [`schema.shred_code`]: stores code shreds (in check_insert_code_shreds).
     ///   - [`schema.slot_meta`]: the SlotMeta of the input `shreds` and their related
@@ -149,11 +152,15 @@ pub const ShredInserter = struct {
         const allocator = self.allocator;
         var reed_solomon_cache = try ReedSolomonCache.init(allocator);
         defer reed_solomon_cache.deinit();
+        // QUESTION: In scenario where shreds.len > 0 and not same as is_repaired.len
+        // this would fail but this would have been recorded in self.metrics.num_shreds.add(shreds.len);
+        // Is this fine?
         std.debug.assert(shreds.len == is_repaired.len);
         var total_timer = try Timer.start();
         var write_batch = try self.db.initWriteBatch();
 
         var just_inserted_shreds = AutoHashMap(ShredId, Shred).init(allocator); // TODO capacity = shreds.len
+        // QUESTION. Would ErasureSetId.fec_set_index always === ErasureMeta.fec_set_index
         var erasure_metas = SortedMap(ErasureSetId, WorkingEntry(ErasureMeta)).init(allocator);
         var merkle_root_metas = AutoHashMap(ErasureSetId, WorkingEntry(MerkleRootMeta)).init(allocator);
         var slot_meta_working_set = AutoHashMap(u64, SlotMetaWorkingSetEntry).init(allocator);
@@ -176,6 +183,7 @@ pub const ShredInserter = struct {
         var newly_completed_data_sets = ArrayList(CompletedDataSetInfo).init(allocator);
         defer newly_completed_data_sets.deinit();
         for (shreds, is_repaired) |shred, is_repair| {
+            // QUESTION: What guarantees that contents of shred/is_repair are ordered same way?
             const shred_source: ShredSource = if (is_repair) .repaired else .turbine;
             switch (shred) {
                 .data => |data_shred| {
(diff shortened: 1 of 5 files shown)


src/ledger/insert_shred.zig
```diff
@@ -617,6 +625,47 @@ pub const ShredInserter = struct {
         };
     }
 
+    /// QUESTION: Okay to reuse this documentation from Agave?
```
sonicfromnewyoke, line 628: I assume this documentation was used during the migration of logic, and it was forgotten to actualize

src/ledger/insert_shred.zig
```diff
@@ -149,11 +152,15 @@ pub const ShredInserter = struct {
         const allocator = self.allocator;
         var reed_solomon_cache = try ReedSolomonCache.init(allocator);
         defer reed_solomon_cache.deinit();
+        // QUESTION: In scenario where shreds.len > 0 and not same as is_repaired.len
```
sonicfromnewyoke, line 155: `shreds.len > 0 and not same as is_repaired.len` this is illegal behavior. Maybe it is worth turning `debug` assert into real check with `panic`

src/ledger/insert_shred.zig
```diff
@@ -81,6 +82,7 @@ pub const ShredInserter = struct {
     ///
     /// This function updates the following column families:
     ///   - [`schema.dead_slots`]: mark a shred as "dead" if its meta-data indicates
+    ///     QUESTION: What does replying a shred mean?
```
sonicfromnewyoke, line 85: repl**a**y stage is responsible for the verification/execution of blocks from leaders (and voting/consensus).

src/ledger/insert_shred.zig
```diff
@@ -176,6 +183,7 @@ pub const ShredInserter = struct {
         var newly_completed_data_sets = ArrayList(CompletedDataSetInfo).init(allocator);
         defer newly_completed_data_sets.deinit();
         for (shreds, is_repaired) |shred, is_repair| {
+            // QUESTION: What guarantees that contents of shred/is_repair are ordered same way?
```
sonicfromnewyoke, line 186: should be. In another way, it doesn't make sense

sonicfromnewyoke (commented):
Answered to a couple of questions. Hope it'll shed some light

src/ledger/insert_shred.zig
```diff
@@ -89,6 +91,7 @@ pub const ShredInserter = struct {
     ///     - The slot is not currently full
     ///     It means there's an alternate version of this slot. See
     ///     `check_insert_data_shred` for more details.
+    ///    // QUESTION: How are these various family corrolated with one another together?
```
0xNineteen, line 94: data shreds store the data, coding shreds are used to recover missing data shreds 

https://github.com/solana-foundation/specs/blob/main/p2p/shred.md

src/ledger/insert_shred.zig
```diff
@@ -635,21 +684,28 @@ pub const ShredInserter = struct {
     ) !ArrayList(CompletedDataSetInfo) {
         const slot = shred.fields.common.slot;
         const shred_index: u64 = @intCast(shred.fields.common.index);
+        // QUESTION: Why set this back to a union when we are in a function that deals with
```
0xNineteen, line 687: duplicate shreds can be either coding or data shreds, so we convert it to a union to add it to the arraylist (should probably moved to be defined right above where it is used)

src/ledger/insert_shred.zig
```diff
@@ -829,6 +898,9 @@ pub const ShredInserter = struct {
         const shred_index_u32 = shred.fields.common.index;
         const shred_index_u64: u64 = @intCast(shred_index_u32);
         const is_last_in_slot = shred.isLastInSlot();
+        // QUESTION/SUGGESTION rename this to is_valid (or similar) instead of sanitize as santize gives the
```
0xNineteen, line 901: sanitize is what is used throughout agave so originally we started to copy their approach which has led to many sanitize method calls everywhere (mainly gossip since it was the first module we worked on) -- imo i agree we should stop using sanitize, ideally change it to `validate` (see `AccountInFile.validate`)

src/ledger/meta.zig
```diff
@@ -145,6 +145,8 @@ pub const ConnectedFlags = BitFlags(enum(u8) {
     // 2) S's parent is connected AND S is full (S's complete block present)
     //
     // 1) is a straightfoward case, roots are finalized blocks on the main fork
+    // // QUESTION: How come there can be many roots?
```
0xNineteen, line 148: if a slot is rooted then its parent is also rooted

ie with chain A -> B -> C -- A and B are rooted but C still doesnt have enough votes to be considered rooted

src/ledger/meta.zig
```diff
@@ -258,12 +260,18 @@ pub const ErasureConfig = struct {
 /// Index recording presence/absence of shreds
 pub const Index = struct {
     slot: Slot,
+    // QUESTION: maybe name these data_index and code_index?
```
0xNineteen, line 263: 👍

src/ledger/rocksdb.zig
```diff
@@ -38,18 +38,38 @@ pub fn RocksDB(comptime column_families: []const ColumnFamily) type {
...
+            //         .{ .create_if_missing = true, .create_missing_column_families = true },
+            //         column_family_descriptions,
+            //     },
+            // );
+            // defer allocator.free(cfs);
+
+            // QUESTION:
+            // What's the benefit of passing rocks.DB.openCf into callRocks to be called using @call
+            // over calling it directly as below?
+            // PROBABLE ANSWER: Reading more of the code to avoid repeating the error handling for
```
0xNineteen, line 58: 👍

src/ledger/rocksdb.zig
```diff
@@ -70,6 +90,8 @@ pub fn RocksDB(comptime column_families: []const ColumnFamily) type {
             };
         }
 
+        // QUESTION/NIT: What about naming this close? To be symmetric with open.
```
0xNineteen, line 93: agreed - init() would be better to be consistent -- can add a doc comment explaining how it opens the db

src/ledger/shred.zig
```diff
@@ -529,6 +529,8 @@ pub fn GenericShred(
 pub const ShredId = struct {
     slot: Slot,
     index: u32,
+    // Question: Why have shred_type as part of the ShredId?
```
0xNineteen, line 532: its possible to have a data and code shred of the same slot and the same index, which without shred_type, would not be a unique struct -- to make the struct unique, we need the shred_type

src/ledger/insert_shred.zig
dnut: It's expected that the caller structures them in the same order. If not, that's a bug, and you should not be surprised to see problems.

This is only used for metrics and logging. The only thing a bug here would do is mess with observability.

src/ledger/insert_shred.zig
dnut: It's expected that the caller makes them the same length. If not, that's a bug, and you should not be surprised to see problems.

I think an assert here is fine since it's the kind of bug that would probably show up pretty quickly in testing. Also we typically use ReleaseSafe, so it is effectively a panic. If we really expect this to be a malformed input under some scenarios, I would have it return an error, not panic.

src/ledger/insert_shred.zig
```diff
@@ -635,21 +684,28 @@ pub const ShredInserter = struct {
...
+            // the self.getIndexMetaEntry would have been a waste
             try shred.parent(),
         );
         const slot_meta = &slot_meta_entry.new_slot_meta;
 
         const erasure_set_id = shred.fields.common.erasureSetId();
         // TODO: redundant get or put pattern
         if (!merkle_root_metas.contains(erasure_set_id)) {
+            // QUESTION: Why do some variables have underscore as a suffix?
+            // In this case meta_
```
dnut, line 708: I was uncreative with naming, and zig prohibits shadowing.

src/ledger/insert_shred.zig
```diff
@@ -677,6 +733,8 @@ pub const ShredInserter = struct {
                     "Received *last* shred index {} less than previous shred index {}, and slot {} is not full, marking slot dead",
                     .{ shred_index, slot_meta.received, slot },
                 );
+                // QUESTION: Should the function not return at this point?
+                // Given we have a dead slots and we have saved that.
```
dnut, line 737: That's a good point. It would probably help performance to return immediately. 

My main hesitation is a fear of diverging from agave's behavior, since agave continues to insert the shred at this point. I'm not sure whether or not it matters. I'd prefer to diverge only if I'm sure if it matters. It's probably a good idea to record ideas like this in a permanent document.

Maybe it could be beneficial to go ahead and insert the shred anyway. For example, maybe we do get all the shreds for the current slot, and consensus decides to actually use the larger version of the block. Then we would want to make sure we're not discarding shreds from that block. I'm not sure how likely this scenario is.

src/ledger/insert_shred.zig
```diff
@@ -739,6 +797,8 @@ pub const ShredInserter = struct {
         allocator: std.mem.Allocator,
         slot: Slot,
         working_set: *AutoHashMap(Slot, IndexMetaWorkingSetEntry),
+        // QUESTION: Is there another way apart from passing this down? What about passing self.metrics instead?
+        // See index_meta_time_us was not initially sure why this? But this is being passed all the way down for timing purposes
```
dnut, line 801: You're right. This is messy. How about this: https://github.com/Syndica/sig/commit/c4c0a754adf778812609b47d1f1bf30699185668

src/ledger/insert_shred.zig
```diff
@@ -748,6 +808,7 @@ pub const ShredInserter = struct {
             if (try self.db.get(schema.index, slot)) |item| {
                 entry.value_ptr.* = .{ .index = item };
             } else {
+                // QUESTION: Find out at what point this get persisted?
```
dnut, line 811: Near the bottom of insertShreds you'll find
```zig
try write_batch.put(schema.index, entry.key_ptr.*, working_entry.index);
```

src/ledger/insert_shred.zig
```diff
@@ -762,6 +823,8 @@ pub const ShredInserter = struct {
         slot: Slot,
         parent_slot: Slot,
     ) !*SlotMetaWorkingSetEntry {
+        // QUESTION: Any reason why no metrics is being recorded here, but it is
+        // for getIndexMetaEntry with index_meta_time_us
```
dnut, line 827: There's not a very good reason. I just copied agave. There are actually a few similar ones also that could be measured, but are not.

src/ledger/insert_shred.zig
```diff
@@ -810,6 +873,12 @@ pub const ShredInserter = struct {
...
     ) bool {
         const shred_index: u64 = @intCast(shred.fields.common.index);
+        // QUESTION: is the "shred_index < slot_meta.consecutive_received_from_0" check needed?
+        // Ie would there ever be a case where data_index.contains(shred_index) is true but
+        // shred_index < slot_meta.consecutive_received_from_0 is false? If not then is data_index.contains(shred_index)
+        // not enough check?
+        // Also if shred_index > slot_meta.consecutive_received_from_0 is true and also data_index.contains(shred_index) is true
+        // is it a valid case case for data_index.contains(shred_index) to be true but shred_index > slot_meta.consecutive_received_from_0.
         return shred_index < slot_meta.consecutive_received_from_0 or
             data_index.contains(shred_index);
```
dnut, line 883: You made a great observation here. But there is actually a specific performance reason for this redundancy.

The idea is that `<` is faster than `.contains()`. We run the fast check first. If true, that's a definitive answer. But if false, we can't know for sure. This comparison could have false negatives. So in that case, we run the slower check to get a definitive answer.

dnut, line 883: You made a great observation here. But there is actually a specific performance reason for this redundancy.

`data_index.contains(shred_index)` tells us the correct answer. This could be the only thing in the function, and it would return the correct answer.

`shred_index < slot_meta.consecutive_received_from_0`, on the other hand, is not perfect. If this is true, then `data_index.contains(shred_index)` will definitely be true. But If this is false, `data_index.contains(shred_index)` might be either true or false. This is because shreds do not always come in order.

So, why bother with the `<` comparison at all? The idea is that `<` is faster than `.contains()`. We run the fast check first. If true, that's a definitive answer. But if false, we can't know for sure, so we run the slower check to get a definitive answer.

src/ledger/insert_shred.zig
```diff
@@ -1742,6 +1814,7 @@ const ShredSource = enum {
     recovered,
 };
 
+// QUESTION: Is this guaranteed to not have gaps ie start_index..end_index always?
```
dnut, line 1817: Yes, this struct is intended to indicate that we've received every single shred in the range. The logic is in updateCompletedDataIndexes.

src/ledger/insert_shred.zig
```diff
@@ -1820,6 +1893,7 @@ pub const BlockstoreInsertionMetrics = struct {
     }
 };
 
+/// Question: Is it possible to use WorkingEntry(SlotMeta) here?
```
dnut, line 1896: It's possible but would represent a different usage pattern. Normally WorkingEntry is not mutated and just represents whether the item is new or existing. It could probably be adapted for this use case though, where it needs to allow mutations, and it flips from clean to dirty as soon as a shred is inserted. I can't be sure which way is better without seeing it implemented.

src/ledger/meta.zig
```diff
@@ -258,12 +260,18 @@ pub const ErasureConfig = struct {
...
+    // but it is just the index.
     data: ShredIndex,
     code: ShredIndex,
 
     pub fn init(allocator: std.mem.Allocator, slot: Slot) Index {
         return .{
             .slot = slot,
+            // QUESTION: So the slot is passed in here, but it is not used to construct
+            // the data and code index? Would have expected some relationship here between
+            // the slot being passed and the data/code index?
```
dnut, line 274: The indexes tell you which shreds have been received for a particular slot. The slot number is an input here just to include a field that specifies which slot this index is for. But the slot number itself doesn't tell you which shreds have been received for that slot.

Maybe this struct doesn't even need the slot field at all. Current usages would require some rework to get the data from somewhere else though.

src/ledger/meta.zig
dnut: bce4ce1fdb23d4ea7ff3f8a567eedb21d49148d6

src/ledger/shred.zig
```diff
@@ -799,6 +802,7 @@ pub const CommonHeader = struct {
     slot: Slot,
     index: u32,
     version: u16,
+    // Question: Is this also set when it is a data shred
```
dnut, line 805: Yes, each erasure set includes data shreds and coding shreds. fec_set_index represents the erasure set within a slot. "fec set" and "erasure set" are synonyms as far as I'm concerned.

I think the fec_set_index might actually be based on a data shred's index? Not 100% sure about this though

src/ledger/reader.zig
```diff
@@ -153,6 +153,8 @@ pub const BlockstoreReader = struct {
     }
 
     /// Analogous to [get_data_shred](https://github.com/anza-xyz/agave/blob/15dbe7fb0fc07e11aaad89de1576016412c7eb9e/ledger/src/blockstore.rs#L2220)
+    // QUESTION: Agave does a check if the shred retrieved is of the right variant. Ie if it is not
+    // LegacyCode or MerkleCode. Is the implementation of the retrieval already preventing that here?
     pub fn getDataShred(self: *Self, slot: Slot, index: u64) !?BytesRef {
```
dnut, line 158: More or less, yes. Since we're not supporting legacy shreds, we don't need to do the fallible conversion, so there's no error to handle.

src/ledger/meta.zig
```diff
@@ -359,6 +367,8 @@ pub const MerkleRootMeta = struct {
             // shred that contains a proper merkle root would constitute
             // a valid duplicate shred proof.
             .merkle_root = shred.fields.merkleRoot() catch null,
+            // QUESTION: This does not look like the _first_ recieved shred index,
+            // but the index of _this_ shred we used to construct the MerkleRootMeta
```
dnut, line 371: That's a great point. This code is actually correct because it is only used for the first shred. But it's not clear from the naming.
41878cb670dd4837ddbb722172ff3d45488c651c

PR #276, opened 2024-09-25
feat(gossip): improved metrics

0xNineteen:
new gossip metrics include: 
```
    // inserting push messages stats
    push_message_n_invalid_shred_version: *Counter,
    push_message_n_new_inserts: *Counter,
    push_message_n_overwrite_existing: *Counter,
    push_message_n_old_value: *Counter,
    push_message_n_duplicate_value: *Counter,
    push_message_n_timeouts: *Counter,

    // inserting pull response stats
    pull_response_n_invalid_shred_version: *Counter,
    pull_response_n_new_inserts: *Counter,
    pull_response_n_overwrite_existing: *Counter,
    pull_response_n_old_value: *Counter,
    pull_response_n_duplicate_value: *Counter,
    pull_response_n_timeouts: *Counter,
```

this gives new insight into whats happening to the messages we receive (mainly pull responses, since the gossip client doesnt seem to get many push messages)

Screenshot 2024-09-25 at 3 40 32 PM

this pr also includes a few bug fixes: 
- grafana: now uses the correct delta on metrics (was prev using incorrect idelta)
- table: prev a full table was not allowing updating of existing labels, this changes it 
- table: when gossip data was updated, we were not free-ing the old overwritten data, this fixes it 
- cmd: fixes the predefined mainnet entrypoints
- pull_requests: dont send requests to peers which have a different shred version than us

### DIFF
--- a/src/gossip/table.zig
+++ b/src/gossip/table.zig
@@ -33,24 +33,6 @@ pub fn AutoArrayHashSet(comptime T: type) type {
     return AutoArrayHashMap(T, void);
 }
 
-pub const InsertResults = struct {
-    inserted: ?std.ArrayList(usize),
-    timeouts: ?std.ArrayList(usize),
-    failed: ?std.ArrayList(usize),
-
-    pub fn deinit(self: InsertResults) void {
-        if (self.inserted) |inserted| {
-            inserted.deinit();
-        }
-        if (self.timeouts) |timeouts| {
-            timeouts.deinit();
-        }
-        if (self.failed) |failed| {
-            failed.deinit();
-        }
-    }
-};
-
 /// Cluster Replicated Data Store: stores gossip data
 /// the self.store uses an AutoArrayHashMap which is a HashMap that also allows for
 /// indexing values (value = arrayhashmap[0]). This allows us to insert data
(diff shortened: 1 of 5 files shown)


src/gossip/table.zig
```diff
@@ -339,23 +339,37 @@ pub const GossipTable = struct {
...
         for (values, 0..) |value, index| {
             const value_time = value.wallclock();
             const is_too_new = value_time > now +| timeout;
             const is_too_old = value_time < now -| timeout;
             if (is_too_new or is_too_old) {
+                timeout_count += 1;
                 continue;
             }
 
             const did_insert = self.insert(value, now) catch false;
```
sonicfromnewyoke, line 361: I'm not sure about the design of the `insert` function.

this function returns only `true` or an error. 
`false` - is unreachable due to (line 274)
```zig
if (old_entry.value_hash.order(&versioned_value.value_hash) != .eq) {
    // if hash isnt the same and override() is false then msg is old
    try self.purged.insert(old_entry.value_hash, now);
    return InsertionError.OldValue;
} else {
    // hash is the same then its a duplicate
    return InsertionError.DuplicateValue;
}
```


maybe, it makes sense to return something like an enum which will say what, actually, was done inside the `insert`, either a fresh insertion or overwrite of the previously stored one.

what to you think?

InKryption: I think it would make more sense to simply return void instead of bool if this would only ever return false on error; using an enum can be elegant, but it then also then misses out on the type safety that the compiler provides in not allowing the caller to ignore errors in the same way as a simple enum value.

sonicfromnewyoke: sorry, I wasn't clear enough, I meant something like `!enum`

0xNineteen: nice thats a great idea - [ebfee79](https://github.com/Syndica/sig/pull/276/commits/ebfee79d6f7f4179d00d0be32bfcb66791af80e0)

sonicfromnewyoke: TBH, I thought about something simpler, like [file_ensure_content](https://github.com/tigerbeetle/tigerbeetle/blob/main/src/shell.zig#L275) function in TigerBeetle's codebase.

But your solution I like more because it gives more flexibility and an easy way to track and clean up old/duplicate values 💪

sonicfromnewyoke (approved):
LGTM

InKryption (approved):

PR #291, opened 2024-10-01
improve(grafana): use rate for the packet info counter

### LINKED ISSUE #288: feat(metrics): grafana packet rate instead of count

Rexicon226:
closes #288

This gives us a 'packets/second' metric that should be more useful. Any lags in processing messages will be reflected in the channel length rather than the slope of the packets counter.

The new counter looks like this:
image

If there are any delays in processing, the length of the channel will go up:
image

### DIFF
--- a/src/gossip/service.zig
+++ b/src/gossip/service.zig
@@ -479,7 +479,7 @@ pub const GossipService = struct {
             // PERF: investigate CPU pinning
             var task_search_start_idx: usize = 0;
             while (self.packet_incoming_channel.receive()) |packet| {
-                defer self.stats.gossip_packets_received.inc();
+                defer self.stats.gossip_packets_received_total.inc();
 
                 const acquired_task_idx = VerifyMessageTask.awaitAndAcquireFirstAvailableTask(tasks, task_search_start_idx);
                 task_search_start_idx = (acquired_task_idx + 1) % tasks.len;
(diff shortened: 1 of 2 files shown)


src/gossip/service.zig
```diff
@@ -1942,9 +1942,9 @@ pub const GossipService = struct {
 
 /// stats that we publish to prometheus
 pub const GossipStats = struct {
-    gossip_packets_received: *Counter,
-    gossip_packets_verified: *Counter,
-    gossip_packets_processed: *Counter,
+    gossip_packets_received_total: *Counter,
```
dnut, line 1945: Isn't "total" implied by it being a counter? I feel like you could add "total" to the name of every counter and it would be just as accurate.

Rexicon226: > Isn't "total" implied by it being a counter? I feel like you could add "total" to the name of every counter and it would be just as accurate.

I agree, but I did this because Grafana was complaining about using a variable not ending in `_total` for `rate(`.

Rexicon226: image

sonicfromnewyoke: `_total` ending is recommended by the [Prometheus team](https://prometheus.io/docs/practices/naming) itself

dnut: Fair enough. I can see the reasoning for using this name when exposed to prometheus and grafana, since the http API does not include type information.

Maybe we should open another pr to rename all counters, and make this a standard practice.

0xNineteen (approved):

PR #293, opened 2024-10-01
improve(gossip): optimize packet filtering

### LINKED ISSUE #287: perf: optimize packet filtering

Rexicon226:
closes #287

Implements a more optimized version of the filtering. Using [Hoare's Partitioning Scheme](https://en.wikipedia.org/wiki/Quicksort#Hoare_partition_scheme), as it's the simplest and provides the best time complexity of O(N).

Benchmarks look identical to main, this is a small-ish optimization that will become more apparent as other parts of the validator get faster.

### DIFF
--- a/src/gossip/service.zig
+++ b/src/gossip/service.zig
@@ -1904,39 +1904,87 @@ pub const GossipService = struct {
         gossip_values: []SignedGossipData,
         sender_pubkey: Pubkey,
     ) usize {
-        // we use swap remove which just reorders the array
-        // (order dm), so we just track the new len -- ie, no allocations/frees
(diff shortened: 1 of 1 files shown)


0xNineteen:
if benchmarks look the same and its not a clear bottleneck, would prefer to stick to the swapRemove approach just for simplicity-sake (even if its slower) -- a new algo to learn just adds too much complexity to the code with little benefit

Rexicon226:
> a new algo to learn just adds too much complexity to the code with little benefit

Fair enough, we can keep this in our pocket until the time comes :)
I will just close the PR, and if needed it's here waiting for us.

Issue #281, opened 2024-09-30
AccountsDB: Accounts Cache


dnut:
there are a few hot accounts which are written to most blocks so ideally we want to keep their state in RAM to reduce DISK reads - this would likely be an accounts cache for hot accounts

PR #297, opened 2024-10-02
feat(accountsdb): implement account cache

Sobeston:
aims to close https://github.com/Syndica/sig/issues/281


src/accountsdb/db.zig
```diff
@@ -58,18 +59,334 @@ pub const ACCOUNT_INDEX_BINS: usize = 8192;
...
+
+//     try accounts_lru.putAccount(slot, pubkey, account);
+
+//     // Purge the slot
+//     try accounts_lru.purgeSlot(slot);
+
+//     const result = accounts_lru.get(slot, pubkey);
+//     try std.testing.expect(result == null);
+// }
+
```
Sobeston, line 371: Should fix and re-enable these tests before merging. There's a memory leak somewhere

src/accountsdb/db.zig
```diff
@@ -58,18 +59,334 @@ pub const ACCOUNT_INDEX_BINS: usize = 8192;
...
+        if (self.highest_slot == null or slot > self.highest_slot.?) {
+            self.highest_slot = slot;
+        }
+
+        const new_cached_account = try self.slot_lrus.allocator.create(CachedAccount);
+        errdefer self.slot_lrus.allocator.destroy(new_cached_account);
+        new_cached_account.* = try CachedAccount.init(self.slot_lrus.allocator, account);
+        errdefer new_cached_account.deinit(self.slot_lrus.allocator);
+
+        if (slot_lru.put(pubkey, new_cached_account) != null) return error.InvalidPut;
```
0xNineteen:
```suggestion
        if (slot_lru.put(pubkey, new_cached_account) != null) return error.AlreadyExistsInCache;
```

src/accountsdb/db.zig
```diff
@@ -58,18 +59,334 @@ pub const ACCOUNT_INDEX_BINS: usize = 8192;
...
+
+    /// remove slot lru, decreasing ref_counts on CachedAccounts (optionally removing)
+    fn purgeSlot(self: *AccountsLRU, slot: Slot) void {
+        const slot_lru = self.slot_lrus.get(slot) orelse return;
+        slot_lru.deinit();
+        _ = self.slot_lrus.remove(slot);
+    }
+
+    /// bring slot lru forward to new slot, increasing ref_counts
+    fn copySlot(self: *AccountsLRU, old_slot: Slot, new_slot: Slot) !void {
```
0xNineteen, line 184:
```suggestion
    fn shiftLRUForward(self: *AccountsLRU, old_slot: Slot, new_slot: Slot) !void {
```

src/accountsdb/db.zig
```diff
@@ -58,18 +59,334 @@ pub const ACCOUNT_INDEX_BINS: usize = 8192;
...
+
+                _ = new_slot_lru.put(node.data.key, node.data.value.copyRef());
+            }
+        }
+
+        self.correctSlotCount();
+
+        try self.slot_lrus.put(new_slot, new_slot_lru);
+
+        self.highest_slot = new_slot;
```
0xNineteen, line 206: is this always true?

0xNineteen (changes requested):
awesome work - it took a bit to understand the architecture but its very well designed and once it clicked, very cool too - most comments are related to organizing the code to how we have it elsewhere in the codebase and naming

[for anyone else doing a PR review - this diagram may help:

Screenshot 2024-10-10 at 9 57 17 PM

src/accountsdb/db.zig
Sobeston: Good point, it might not be

src/accountsdb/cache.zig
```diff
@@ -0,0 +1,331 @@
...
+//     defer accounts_cache.deinit();
+
+//     const pubkey = Pubkey.random(rng);
+//     const slot = 1;
+
+//     try accounts_cache.put(slot, pubkey, account);
+
+//     // Trying to insert the same account again should fail
+//     try std.testing.expectEqual(error.AlreadyExistsInCache, accounts_cache.put(slot, pubkey, account));
+// }
```
Sobeston, line 311: There's a memory leak hiding here somewhere, will re-enable this test when found

Sobeston: Fixed - slot_lru.put was overwriting the old value in cases where the key already existed, causing the value to be dangling and not freed. Fixed by introducing putNoClobber.

0xNineteen (dismissed):
just pushed a commit to fix some naming update changes - lgtm :shipit: 🔥

should be good to merge once davids/inks review is done

src/accountsdb/cache.zig
```diff
@@ -0,0 +1,331 @@
...
+        return .{
+            .slot_lrus = .{},
+            .allocator = allocator,
+            .max_items = max_items,
+            .max_slots = max_slots,
+            .highest_slot = null,
+        };
+    }
+
+    pub fn get(self: *const Self, slot: Slot, pubkey: Pubkey) ?CachedAccount {
```
InKryption, line 86: What's the point of returning a `CachedAccount` here, if not as a a pointer, and without `copyRef`? This seems like a potential source of race conditions. Please correct me if I'm off the mark here.

src/accountsdb/cache.zig
```diff
@@ -0,0 +1,331 @@
...
+        }
+
+        pub fn copyRef(self: *CachedAccount) *CachedAccount {
+            _ = self.ref_count.fetchAdd(1, .acq_rel);
+            return self;
+        }
+    };
+
+    pub const LRU = LruCacheCustom(
+        .locking,
```
InKryption, line 56: Could you generally comment on the multi-threaded interaction with the AccountsCache itself, and each LRU entry, and/or any future plans regarding this?

Within AccountsDB the whole account cache is under a read-write lock, so I'm struggling to see when having each individual LRU entry be thread-safe in and of itself would be necessary.

src/accountsdb/cache.zig
```diff
@@ -0,0 +1,331 @@
...
+            slot_lru_ptr.deinit();
+        }
+        self.slot_lrus.deinit(self.allocator);
+    }
+};
+
+test "CachedAccount ref_count" {
+    const allocator = std.testing.allocator;
+    var random = std.rand.DefaultPrng.init(19);
+    const rng = random.random();
```
InKryption, line 188: To this and all other instances of this in the other tests: I think it would be worth falling in line with #304, which looks like it's going to be merged.

InKryption (changes requested):
A couple of minor nits, and a couple of questions about the design related to multi-threaded interactions I'd like to understand and potentially document.

I overall quite like the design, I think it's a clear improvement.

src/accountsdb/cache.zig
Sobeston: Down the line I'd like to get rid of the mutex wrapping the cache and make it internally locking.

src/accountsdb/cache.zig
```diff
@@ -0,0 +1,331 @@
...
+const Pubkey = sig.core.pubkey.Pubkey;
+
+const LruCacheCustom = sig.common.lru.LruCacheCustom;
+
+/// Stores read-only in-memory copies of commonly used accounts
+pub const AccountsCache = struct {
+    /// Atomically refcounted account
+    pub const CachedAccount = struct {
+        account: Account,
+        // this account has since been mutated, and will not progress to newer slots
```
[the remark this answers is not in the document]
Sobeston: https://github.com/Syndica/sig/pull/297/commits/ae36aa27390e13793a692a6c0f2c1356439b4e27#diff-7f62328c8d0fd51e206b91ff2b72aef34de0fae176ba46106a0c768d29f306f8L17-L19

src/accountsdb/cache.zig
Sobeston: https://github.com/Syndica/sig/pull/297/commits/ae36aa27390e13793a692a6c0f2c1356439b4e27#diff-7f62328c8d0fd51e206b91ff2b72aef34de0fae176ba46106a0c768d29f306f8L86-R104

InKryption (approved):

src/accountsdb/cache.zig
```diff
@@ -0,0 +1,365 @@
...
+        }
+
+        pub fn releaseOrDestroy(self: *CachedAccount, allocator: std.mem.Allocator) void {
+            const current_count = self.ref_count.load(.acquire);
+            if (current_count == 1) {
+                self.deinit(allocator);
+                allocator.destroy(self);
+            } else {
+                _ = self.ref_count.fetchSub(1, .acq_rel);
+            }
```
Rexicon226, line 41: There are multiple races here. I'm not going to list them, but here's how it should look:
```suggestion
            const current_count = self.ref_count.fetchSub(1, .release);
            if (current_count == 1) {
                _ = self.ref_count.load(.acquire);
                self.deinit(allocator);
                allocator.destroy(self);
            }
```

src/accountsdb/cache.zig
```diff
@@ -0,0 +1,365 @@
...
+        }
+
+        // satisfies the type LruCacheCustom expects
+        pub fn releaseOrDestroyDoublePtr(self: **CachedAccount, allocator: std.mem.Allocator) void {
+            releaseOrDestroy(self.*, allocator);
+        }
+
+        pub fn copyRef(self: *CachedAccount) *CachedAccount {
+            _ = self.ref_count.fetchAdd(1, .acq_rel);
```
Rexicon226, line 50: This should be `monotonic`.

src/accountsdb/cache.zig
```diff
@@ -0,0 +1,365 @@
...
+            return self;
+        }
+
+        pub fn getMeta(self: *CachedAccount) CachedAccountMeta {
+            return .{
+                .is_dirty = self.is_dirty.load(.acquire),
+                .ref_count = self.ref_count.load(.acquire),
```
Rexicon226, line 57: These should be `monotonic`.

src/accountsdb/cache.zig
```diff
@@ -0,0 +1,365 @@
...
+            .max_items = max_items,
+            .max_slots = max_slots,
+            .highest_slot = null,
+        };
+    }
+
+    /// user is expected to release the returned CachedAccount
+    pub fn get(self: *const Self, slot: Slot, pubkey: Pubkey) ?*CachedAccount {
+        const slot_lru = self.slot_lrus.getPtr(slot) orelse return null;
+        return (slot_lru.get(pubkey) orelse return null).copyRef();
```
Rexicon226, line 102: Maybe split this into a 
```zig
const account = slot_lru.get(pubkey) orelse return null;
return account.copyRef();
```
I feel like that would look cleaner. Same for the one below.

src/accountsdb/cache.zig
```diff
@@ -0,0 +1,365 @@
...
+        var new_slot_lru = try LRU.initWithContext(self.allocator, self.max_items, self.allocator);
+
+        // copy from old to new slot lru
+        {
+            old_slot_lru.mux.lock();
+            defer old_slot_lru.mux.unlock();
+
+            var it = old_slot_lru.dbl_link_list.first;
+            while (it) |node| : (it = node.next) {
+                if (node.data.value.is_dirty.load(.acquire)) continue; // do not copy forward modified accounts
```
Rexicon226, line 174: This should be `monotonic`.

src/accountsdb/cache.zig
```diff
@@ -0,0 +1,365 @@
...
+        }
+    }
+
+    /// removes lowest slot when exceeding .max_slots
+    pub fn enforceMaxSlotCount(self: *Self) void {
+        if (self.slot_lrus.count() > self.max_slots) {
+            var lowest_slot: Slot = std.math.maxInt(usize);
+            var iter = self.slot_lrus.iterator();
+            while (iter.next()) |entry| {
+                if (entry.key_ptr.* < lowest_slot) lowest_slot = entry.key_ptr.*;
```
Rexicon226, line 195: Not a blocker, just a suggestion:
``` suggestion
                lowest_slot = @min(entry.key_ptr.*, lowest_slot);
```
Maybe it's clearer? Curious to hear your opinion.

Rexicon226 (changes requested):

src/accountsdb/cache.zig
```diff
@@ -0,0 +1,365 @@
...
+const LruCacheCustom = sig.common.lru.LruCacheCustom;
+
+/// Stores read-only in-memory copies of commonly used accounts
+pub const AccountsCache = struct {
+    /// Atomically refcounted account
+    pub const CachedAccount = struct {
+        account: Account,
+        /// represents whether this account has been mutated, and shouldn't progress to newer slots
+        // TODO: when we start mutating accounts, make sure to set this field
+        is_dirty: std.atomic.Value(bool) = .{ .raw = false },
```
dnut, line 19: Every time you want to mark an account dirty, you need to get it from the Lru. This simultaneously marks the account as the most recently used item, and also as a stale item at the same time. Something feels off about that.

dnut (changes requested):
There are some potential races with the reference counter. I suggest using `sig.sync.ReferenceCounter` if you want a robust way to count references. That counter is a bit more paranoid than necessary for this scenario, since you already have deinits guarded with an exclusive write lock, but I think it's better to share a common rc rather than reinvent the wheel in various different places.

Also I'm still not sure about about indexing starting from the Slot. Seems like it brings unnecessary overhead.

My comments go into more detail.

src/accountsdb/cache.zig
```diff
@@ -0,0 +1,365 @@
...
+        return (slot_lru.get(pubkey) orelse return null).getMeta();
+    }
+
+    /// does not influence underlying LRU's cache ordering
+    pub fn contains(self: *const Self, slot: Slot, pubkey: Pubkey) bool {
+        const slot_lru = self.slot_lrus.getPtr(slot) orelse return false;
+        return slot_lru.contains(pubkey);
+    }
+
+    /// should only be called iff account is not already present
```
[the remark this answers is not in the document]
Sobeston: I think I worded it too strongly, it's fine if the caller calls it on an existing key, it will just not modify the state and return an error.

src/accountsdb/cache.zig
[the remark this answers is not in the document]
Rexicon226: Potential solution: assert the `ref_count` isn't 0

Rexicon226:
> There are some potential races with the reference counter. I suggest using sig.sync.ReferenceCounter if you want a robust way to count references. That counter is a bit more paranoid than necessary for this scenario, since you already have deinits guarded with an exclusive write lock, but I think it's better to share a common rc rather than reinvent the wheel in various different places.

I do want to note that `ReferenceCounter` is not correctly implemented as of now. I'll open a PR to fix it later. Just use it for now, it won't affect too much.

src/accountsdb/cache.zig
```diff
@@ -0,0 +1,365 @@
...
+
+    /// remove slot lru, decreasing ref_counts on CachedAccounts (optionally removing)
+    pub fn purgeSlot(self: *Self, slot: Slot) void {
+        const slot_lru = self.slot_lrus.getPtr(slot) orelse return;
+        slot_lru.deinit();
+        _ = self.slot_lrus.remove(slot);
+    }
+
+    /// bring slot lru forward to new slot, increasing ref_counts
+    pub fn shiftLRUForward(self: *Self, old_slot: Slot, new_slot: Slot) !void {
```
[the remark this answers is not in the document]
Sobeston: Removed per-slot behaviour in https://github.com/Syndica/sig/pull/297/commits/f039a412be0a120833bb2e60244d6b958c526707

src/accountsdb/cache.zig
```diff
@@ -0,0 +1,365 @@
...
+    pub const SlotLRUs = std.AutoHashMapUnmanaged(Slot, LRU);
+
+    const Self = @This();
+
+    slot_lrus: SlotLRUs,
+    allocator: std.mem.Allocator,
+    max_items: usize,
+    max_slots: usize,
+    highest_slot: ?Slot,
+
```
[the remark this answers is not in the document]
Sobeston: changed in https://github.com/Syndica/sig/pull/297/commits/ee1d3869d31b9a1d9ffd9b4bec8a186a5d3a4fd6#diff-7f62328c8d0fd51e206b91ff2b72aef34de0fae176ba46106a0c768d29f306f8L11-R20

src/accountsdb/db.zig
```diff
@@ -1855,37 +1877,52 @@ pub const AccountsDB = struct {
     pub fn getAccountFromRef(self: *Self, account_ref: *const AccountRef) !Account {
         switch (account_ref.location) {
             .File => |ref_info| {
-                return try self.getAccountInFile(
-                    self.allocator,
-                    ref_info.file_id,
-                    ref_info.offset,
-                );
+                const account_in_cache = blk: {
+                    const accounts_cache, var accounts_cache_lg = self.accounts_cache.writeWithLock();
```
0xNineteen, line 1881: read lock?

0xNineteen (dismissed):
lgtm

src/accountsdb/db.zig
Sobeston: The .get requires a mutable pointer, as it can modify the lru state (previously the lru state was behind a pointer).

src/accountsdb/cache.zig
```diff
@@ -0,0 +1,209 @@
...
+        // satisfies the type LruCacheCustom expects
+        pub fn releaseOrDestroyDoublePtr(self: **CachedAccount, allocator: std.mem.Allocator) void {
+            releaseOrDestroy(self.*, allocator);
+        }
+
+        pub fn copyRef(self: *CachedAccount) *CachedAccount {
+            if (self.ref_count.acquire()) {
+                return self;
+            } else {
+                @panic("Attempted to acquire reference to destroyed CachedAccount");
```
Rexicon226, line 49: This panic doesn't make sense to have, `ref_count` will also be destroyed with the last release.

src/accountsdb/cache.zig
```diff
@@ -0,0 +1,209 @@
...
+
+        pub fn releaseOrDestroy(self: *CachedAccount, allocator: std.mem.Allocator) void {
+            if (self.ref_count.release()) {
+                self.deinit(allocator);
+                allocator.destroy(self);
+            }
+        }
+
+        // satisfies the type LruCacheCustom expects
+        pub fn releaseOrDestroyDoublePtr(self: **CachedAccount, allocator: std.mem.Allocator) void {
```
Rexicon226, line 41: This is a hack to fix something that we ourselves wrote. `LruCacheCustom` should be patched so as not to need this.

src/accountsdb/cache.zig
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    /// Does not influence underlying LRU's cache ordering
+    pub fn contains(self: *const Self, pubkey: Pubkey) bool {
+        return self.lru.contains(pubkey);
+    }
+
+    /// Replaces the previous entry, if one existed
+    pub fn put(self: *Self, pubkey: Pubkey, account: Account) error{OutOfMemory}!void {
+        const new_entry = try self.allocator.create(CachedAccount);
```
Rexicon226, line 104: This allocation seems superfluous to me. The LRU already allocates new keys when they are added. Can we rewrite something somewhere to not need this second allocation on every `put` instead of on the first one?

Rexicon226 (changes requested):
It looks great. From a maintainability point of view, I have a couple of nits that I would like to address before merging.

src/accountsdb/cache.zig
```diff
@@ -0,0 +1,209 @@
...
+        }
+    };
+
+    /// A copy of a CachedAccout's fields, without the account
+    const CachedAccountMeta = struct {
+        ref_count: ReferenceCounter,
+    };
+
+    pub const LRU = LruCacheCustom(
+        .locking,
```
dnut, line 66: This is redundant with the RwMux. I get that you want internal locking for this structure eventually, but specifying `locking` here isn't good enough for that, which is why you need the RwMux. It's not clear to me how a `locking` Lru could be beneficial here, so I would just make this `non_locking` for now. If you want to start moving towards internal locking, you can support it fully by integrating a RwLock in AccountsCache, and get rid of the RwMux wrapper.

dnut (changes requested):
Looks good, I just have a few more thoughts

src/accountsdb/cache.zig
dnut: Might be a complicated patch to retain compatibility with the situation where you're storing values instead of pointers in the cache. If your other suggestion is implemented, I don't think we need this function any more

> This allocation seems superfluous to me. The LRU already allocates new keys when they are added. Can we rewrite something somewhere to not need this second allocation on every put instead of on the first one?

src/accountsdb/cache.zig
```diff
@@ -0,0 +1,209 @@
...
+        const new_entry = try self.allocator.create(CachedAccount);
+        new_entry.* = try CachedAccount.init(self.allocator, account);
+
+        if (self.lru.put(pubkey, new_entry)) |old_entry| {
+            old_entry.releaseOrDestroy(self.allocator);
+        }
+    }
+
+    pub fn deinit(self: *Self) void {
+        self.lru.deinit();
```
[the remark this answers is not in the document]
Sobeston: The LRU calls that whenever it deinits a node so that's already handled

src/accountsdb/cache.zig
dnut: I agree that the only case where this could theoretically be false is when the pointer itself is already invalidated and unsafe to use. If the function is called at that point, it means you already have a bug somewhere else, from mismanaging the pointer, and the behavior of calling this function is undefined.

So you could just have `_ = self.ref_count.acquire()` here. Essentially you use it as a basic reference counter, and don't bother to check for cases that are actually impossible to accurately detect.

That being said, maybe there is an advantage to having a panic or unreachable here. Nothing is guaranteed, since any code that could trigger this would have undefined behavior, but at least you get a probabilistic tilt towards failing fast instead of failing silently.

src/accountsdb/cache.zig
Sobeston: The reasoning is that the LRU doesn't necessarily "own" the account, we don't necessarily want the entry to be deallocated when it's evicted; references to it could still be held. With that being said we could remove one allocation per put by allocating a CachedAccount and its .account field in the same block, but that feels like a bit of a hack.

dnut: I think the core issue here is that the reference counter must be allocated in some memory that is not owned by the lru. It's fine to copy the CachedAccount struct into the Lru but it needs to respect the same reference counter that's being used by other copies of the CachedAccount.

So you need an allocation outside the Lru regardless, it's just a question of whether to allocate the CachedAccount as a whole, or to only allocate its reference counter.

src/accountsdb/db.zig
```diff
@@ -1855,37 +1877,52 @@ pub const AccountsDB = struct {
...
+                    defer accounts_cache_lg.unlock();
+
+                    const cached_account = accounts_cache.get(account_ref.pubkey) orelse break :blk null;
```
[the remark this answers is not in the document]
Sobeston: Addressed in https://github.com/Syndica/sig/pull/297/commits/de424828b6dbad6b49ebf2b09db33a614f237941

src/accountsdb/cache.zig
Sobeston: I went ahead and changed this to an unreachable, I think conveys intent a bit better

src/accountsdb/cache.zig
Sobeston: I went ahead and replaced it with an unreachable, I think that shows intent a bit better

src/accountsdb/cache.zig
Sobeston: Simplified with https://github.com/Syndica/sig/pull/297/commits/4ad2cb1e2a5f853c1041c03d46dc013536737e76

src/common/lru.zig
```diff
@@ -32,18 +32,26 @@ pub fn LruCacheCustom(
...
+
+        fn (V, DeinitContext) void => struct {
+            fn f(v: *V, ctx: DeinitContext) void {
+                deinitFn_(v.*, ctx);
+            }
+        }.f,
+
         fn (V) void => struct {
             fn f(v: V, _: DeinitContext) void {
                 V.deinit(v);
```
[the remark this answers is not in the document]
Sobeston: Yeah that sounds right to me. Might as well fix the other spot too, let me get that.

dnut (approved):
@Rexicon226 have all your concerns been addressed?

src/accountsdb/cache.zig
```diff
@@ -0,0 +1,228 @@
...
+        const account: *CachedAccount = self.lru.peek(pubkey) orelse return null;
+        if (account.slot == slot) {
+            return account.getMeta();
+        } else {
+            return null;
+        }
+    }
+
+    /// Does not influence underlying LRU's cache ordering
+    pub fn contains(self: *const Self, pubkey: Pubkey) bool {
```
Rexicon226, line 105: This function isn't actually used anywhere, plus users can just call `self.lru.contains` without much hassle.

Rexicon226 (approved):
LGTM

Rexicon226 (approved):
LGTM 2.0

PR #300, opened 2024-10-04
refactor(thread): Make `spawnThreadTasks` use a wait group & replace task params with `TaskParams` struct

InKryption:
Extracted from some of my experimentation; doesn't appear to cause any measurable performance impact, reduces allocations (the total amount allocated will likely be similar since it's adding more metadata to the thread metadata), and overall seems to simplify and improve the code quality in my opinion.

Also resolves a TODO with a fairly simple solution.

### DIFF
--- a/src/accountsdb/db.zig
+++ b/src/accountsdb/db.zig
@@ -440,11 +440,16 @@ pub const AccountsDB = struct {
         {
             var wg: std.Thread.WaitGroup = .{};
             defer wg.wait();
-            try spawnThreadTasks(&wg, n_account_files, n_parse_threads, null, loadAndVerifyAccountsFilesMultiThread, .{
-                loading_threads.items,
-                accounts_dir,
-                snapshot_manifest.file_map,
-                accounts_per_file_estimate,
+            try spawnThreadTasks(loadAndVerifyAccountsFilesMultiThread, .{
+                .wg = &wg,
+                .data_len = n_account_files,
+                .max_threads = n_parse_threads,
+                .params = .{
+                    loading_threads.items,
+                    accounts_dir,
+                    snapshot_manifest.file_map,
+                    accounts_per_file_estimate,
+                },
             });
         }
         self.logger.infof("total time: {s}", .{timer.read()});
@@ -698,10 +703,15 @@ pub const AccountsDB = struct {
     ) !void {
         var combine_indexes_wg: std.Thread.WaitGroup = .{};
         defer combine_indexes_wg.wait();
-        try spawnThreadTasks(&combine_indexes_wg, self.account_index.numberOfBins(), n_threads, null, combineThreadIndexesMultiThread, .{
-            self.logger,
-            &self.account_index,
-            thread_dbs,
+        try spawnThreadTasks(combineThreadIndexesMultiThread, .{
+            .wg = &combine_indexes_wg,
+            .data_len = self.account_index.numberOfBins(),
+            .max_threads = n_threads,
+            .params = .{
+                self.logger,
+                &self.account_index,
+                thread_dbs,
+            },
         });
 
         // ensure enough capacity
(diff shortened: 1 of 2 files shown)


src/utils/thread.zig
```diff
@@ -7,41 +7,64 @@ const Mutex = std.Thread.Mutex;
...
+    maybe_coverage: ?*usize,
+    comptime taskFn: anytype,
+    params: anytype,
+) std.Thread.SpawnError!void {
+    const chunk_size, const n_threads = chunkSizeAndThreadCount(data_len, max_n_threads);
+
+    if (maybe_coverage) |coverage| coverage.* = 0;
+
+    const S = struct {
+        fn taskFnWg(wg: *std.Thread.WaitGroup, fn_params: @TypeOf(params), task_params: TaskParams) @typeInfo(@TypeOf(taskFn)).Fn.return_type.? {
```
Rexicon226, line 43: Already exists in the stdlib in the form of `wg.spawnManager`. I hate that it exists but I mean, might as well use it.

Rexicon226 (changes requested):
Good changes overall.

src/utils/thread.zig
Rexicon226: To add, I don't see a point to catching `spawn` failing. It would only fail under the most extreme conditions, it wouldn't be something that we could recover from.

src/accountsdb/db.zig
```diff
@@ -438,24 +438,14 @@ pub const AccountsDB = struct {
...
-                n_parse_threads,
-            );
+            var wg: std.Thread.WaitGroup = .{};
+            defer wg.wait();
+            try spawnThreadTasks(&wg, n_account_files, n_parse_threads, null, loadAndVerifyAccountsFilesMultiThread, .{
+                loading_threads.items,
+                accounts_dir,
+                snapshot_manifest.file_map,
+                accounts_per_file_estimate,
+            });
```
0xNineteen, line 448: can we add another comma to the end of all these calls to make the parameters organized better - its a bit long rn

```suggestion
            try spawnThreadTasks(&wg, n_account_files, n_parse_threads, null, loadAndVerifyAccountsFilesMultiThread, .{
                loading_threads.items,
                accounts_dir,
                snapshot_manifest.file_map,
                accounts_per_file_estimate,
            },);
```

src/utils/thread.zig
InKryption: The caller could pretty reasonably recover by doing the remaining work on their own thread, if it were something critical. I get that ideally we move away from spawnThreadTasks to something better, but insolong as it is here, I don't see the harm in handling edge cases.

0xNineteen (approved):
lgtm just a small nit

src/utils/thread.zig
Rexicon226: The good thing about `spawnManager` is that it can't fail, and just gives ownership of the task to the caller thread to ensure it finishes. That would be a slowdown of course, but it would ensure that what needs to be done gets done.

InKryption: I don't see the need to pay that cost in cases where the work isn't critical - this implementation lets the caller ignore it if that is the case.

Rexicon226: `spawn` failing is not something we should really be trying to optimize. The "cost" we pay as you say is that either the caller needs to explicitly handle setting up logic for re-running threads, which will almost certainly just fail again (maybe resulting in a loop, so then we need to add extra retry attempts logic), or the function call "just works" and it may have taken some extra time but the work gets done anyways. 

`spawn` return an error is not an edge case, it's a large problem and recovering from it is highly unlikely.

Rexicon226: I understand the idea now, sorry it took me a couple times to get it :). This is fine for me and better than the stdlib one.

0xNineteen (approved):
great improvement :shipit:

Rexicon226 (approved):
lgtm

PR #303, opened 2024-10-04
feat(accounts-db): fast loading using disk index

### LINKED ISSUE #299: feat(accounts-db): fast snapshot loading
one of the most expensive operations is loading from a snapshot - specifically, generating the account index - ideally if you do it once, you should be able to re-use the index created and load much faster

this will be split up into PRS:
- improved naming/organization (#320)
- changes to the account index to support fast loading 
  - improved allocation approach (#324)
  - index instead of ptr (https://github.com/Syndica/sig/pull/358)
- code which saves and loads the index data

_note:_ old branch with more context can be found in #303

0xNineteen:
attempt to close #299

attempt 1:
- store references on disk to a file and re-build the swissmap (re-building all `.next_ptr` values to point to the new memory addresses) [👎 - still slow - main bottleneck is the swissmap's `getOrPut`] 
-- add multithreading to building the swissmaps [👎 - still slow]

attempt 2: 
- read references with MAP_FIXED mmap flag to ensure the pointers in the linked-lists all remain valid so we dont have to do any work [👎 - all logic was implemented but `MAP_FIXED` fails to maintain the given memory address every time its been run]

attempt 3: 
- modify the account-index to use index's instead of pointers (ie, `next_ptr` changes from `*AccountRef` to `(Slot, Index)` with a lookup in Map(Slot, []AccountRef) so we can keep the addresses the same 
-- note: slower reads (iterating over array + lookup vs direct lookup) but supports fast loading
-- need to benchmark the change in performance if its work
-- -- need to improve benchmarking process -- poor rn (cli usage + exporting data) [🧙‍♂️ -- david/I are here]

...

initial implementation (with lots of todos):
- [x] map out thread-safety of new index design
- [x] modify code to allocate a single `AccountRef` slice which can be used with RecycleFBA to manage state
- [x] save underlying_ref_memory from the RecycleFBA to disk after done snapshot loading 
- [x] modify `reference_memory_map` to use an index approach (not pointer) 
- [x] change `AccountRef` code to use `next_index` instead of pointers

\>> initial implementation is roughly done (enough to benchmark) - want to finish benchmarking story before going further to confirm performance hasnt regressed significantly - so switching to that

full implementation (throughout full codebase):
- [ ] modify code to allocate a single `AccountRef` slice which can be used with RecycleFBA to manage state
- [x] save underlying_ref_memory from the RecycleFBA to disk after done snapshot loading 
- [ ] modify `reference_memory_map` to use an index approach (not pointer) 
- [ ] change `AccountRef` code to use `next_index` instead of pointers

...
Screenshot 2024-10-09 at 4 12 16 PM

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -9,13 +9,14 @@ pub fn build(b: *Build) void {
     const optimize = b.standardOptimizeOption(.{});
     const filters = b.option([]const []const u8, "filter", "List of filters, used for example to filter unit tests by name"); // specified as a series like `-Dfilter="filter1" -Dfilter="filter2"`
     const enable_tsan = b.option(bool, "enable-tsan", "Enable TSan for the test suite");
+    const no_run = b.option(bool, "no-run", "Do not run the selected step and install it") orelse false;
 
     // CLI build steps
-    const run_step = b.step("run", "Run the sig executable");
+    const sig_step = b.step("run", "Run the sig executable");
     const test_step = b.step("test", "Run library tests");
     const fuzz_step = b.step("fuzz", "Gossip fuzz testing");
     const benchmark_step = b.step("benchmark", "Benchmark client");
-    const geyser_reader_step = b.step("geyser_reader", "read data from geyser");
+    const geyser_reader_step = b.step("geyser_reader", "Read data from geyser");
 
     // Dependencies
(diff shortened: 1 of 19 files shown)


0xNineteen:
this code is pretty nasty rn - gonna restart fresh

PR #304, opened 2024-10-05
naming(random): rename rng -> rand

sonicfromnewyoke:
closes https://github.com/orgs/Syndica/projects/2/views/1?pane=issue&itemId=78473765

- [x] ensure usage of `rng` name for `std.Random.DefaultPrng`
- [x] ensure usage of `rand` name for `std.Random`

### DIFF
--- a/src/accountsdb/genesis_config.zig
+++ b/src/accountsdb/genesis_config.zig
@@ -86,7 +86,7 @@ pub const Rent = extern struct {
     pub fn initRandom(random: std.Random) Rent {
         return .{
             .lamports_per_byte_year = random.int(u64),
-            .exemption_threshold = @bitCast(random.int(u64)),
+            .exemption_threshold = random.float(f64),
             .burn_percent = random.uintAtMost(u8, 100),
         };
     }
@@ -114,12 +114,12 @@ pub const Inflation = struct {
 
     pub fn initRandom(random: std.Random) Inflation {
         return .{
-            .initial = @bitCast(random.int(u64)),
-            .terminal = @bitCast(random.int(u64)),
-            .taper = @bitCast(random.int(u64)),
-            .foundation = @bitCast(random.int(u64)),
-            .foundation_term = @bitCast(random.int(u64)),
-            .__unused = @bitCast(random.int(u64)),
+            .initial = random.float(f64),
+            .terminal = random.float(f64),
+            .taper = random.float(f64),
+            .foundation = random.float(f64),
+            .foundation_term = random.float(f64),
+            .__unused = random.float(f64),
         };
     }
 };
--- a/src/accountsdb/genesis_config.zig
+++ b/src/accountsdb/genesis_config.zig
@@ -56,14 +56,14 @@ pub const FeeRateGovernor = struct {
 
     pub const @"!bincode-config:lamports_per_signature" = bincode.FieldConfig(u64){ .skip = true };
 
-    pub fn random(rand: std.Random) FeeRateGovernor {
+    pub fn initRandom(random: std.Random) FeeRateGovernor {
         return .{
-            .lamports_per_signature = rand.int(u64),
-            .target_lamports_per_signature = rand.int(u64),
-            .target_signatures_per_slot = rand.int(u64),
-            .min_lamports_per_signature = rand.int(u64),
-            .max_lamports_per_signature = rand.int(u64),
-            .burn_percent = rand.uintAtMost(u8, 100),
+            .lamports_per_signature = random.int(u64),
+            .target_lamports_per_signature = random.int(u64),
+            .target_signatures_per_slot = random.int(u64),
+            .min_lamports_per_signature = random.int(u64),
+            .max_lamports_per_signature = random.int(u64),
+            .burn_percent = random.uintAtMost(u8, 100),
         };
     }
 };
@@ -83,11 +83,11 @@ pub const Rent = extern struct {
     /// distributed to validators.
     burn_percent: u8,
 
-    pub fn random(rand: std.Random) Rent {
+    pub fn initRandom(random: std.Random) Rent {
         return .{
-            .lamports_per_byte_year = rand.int(u64),
-            .exemption_threshold = @bitCast(rand.int(u64)),
-            .burn_percent = rand.uintAtMost(u8, 100),
+            .lamports_per_byte_year = random.int(u64),
+            .exemption_threshold = @bitCast(random.int(u64)),
+            .burn_percent = random.uintAtMost(u8, 100),
         };
     }
 };
@@ -112,14 +112,14 @@ pub const Inflation = struct {
     /// DEPRECATED, this field is currently unused
     __unused: f64,
 
-    pub fn random(rand: std.Random) Inflation {
+    pub fn initRandom(random: std.Random) Inflation {
         return .{
-            .initial = @bitCast(rand.int(u64)),
-            .terminal = @bitCast(rand.int(u64)),
-            .taper = @bitCast(rand.int(u64)),
-            .foundation = @bitCast(rand.int(u64)),
-            .foundation_term = @bitCast(rand.int(u64)),
-            .__unused = @bitCast(rand.int(u64)),
+            .initial = @bitCast(random.int(u64)),
+            .terminal = @bitCast(random.int(u64)),
+            .taper = @bitCast(random.int(u64)),
+            .foundation = @bitCast(random.int(u64)),
+            .foundation_term = @bitCast(random.int(u64)),
+            .__unused = @bitCast(random.int(u64)),
         };
     }
 };
--- a/src/accountsdb/snapshots.zig
+++ b/src/accountsdb/snapshots.zig
@@ -85,11 +85,11 @@ pub fn stakeDelegationsRandom(
     errdefer stake_delegations.deinit();
 
     try sig.rand.fillHashmapWithRng(&stake_delegations, random, random.uintAtMost(usize, max_list_entries), struct {
-        pub fn randomKey(_random: std.Random) !Pubkey {
-            return Pubkey.initRandom(_random);
+        pub fn randomKey(rand: std.Random) !Pubkey {
+            return Pubkey.initRandom(rand);
         }
-        pub fn randomValue(_random: std.Random) !Delegation {
-            return Delegation.initRandom(_random);
+        pub fn randomValue(rand: std.Random) !Delegation {
+            return Delegation.initRandom(rand);
         }
     });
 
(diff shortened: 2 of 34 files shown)


InKryption:
I think this is a good structural change, however, if we are renaming these with a consistent convention, I would prefer all `rng`s be `prng`s, as it communicates the fact they are, and the intent for them to be, pseudo random number generators, which can be important for things like the potential runtime of a function when using certain algorithms (ie the biased vs non-biased ranged int functions).

dnut:
Nice change, thanks for the contribution.

Why "rand" instead of "random"? The type name is "Random" so I feel like using the same word is slightly clearer.

src/gossip/active_set.zig
```diff
@@ -177,12 +177,13 @@ test "init/denit" {
 test "gracefully rotates with duplicate contact ids" {
     const alloc = std.testing.allocator;
 
-    var rng = std.rand.DefaultPrng.init(100);
+    // NOTE: DefaultPrng == Xoshiro256, so mb it would be better to use only DefaultPrng
```
Rexicon226, line 180: What are these comments getting at? "it would be better to use only DefaultPrng" seems like a TODO comment to me?

sonicfromnewyoke: "it would be better to use only DefaultPrng" sounds like "Remove part which uses Xoshiro256 directly" and may lead to removing important tests

Rexicon226: I don't understand what you mean. Why would we be removing tests? 
I might be missing something really obvious I'm sorry.

0xNineteen:
imo always better to be more explicit with naming -- would support both `random` and `prng` requests

Rexicon226 (changes requested):
Overall good changes :)
There are just a couple of places where us incorrectly seeding with a timestamp led to needing to deviate from the requested naming scheme.

src/net/net.zig
```diff
@@ -540,8 +540,8 @@ test "valid ipv4 socket parsing" {
 }
 
 test "SocketAddr.random" {
-    var rng = std.rand.DefaultPrng.init(@intCast(std.time.milliTimestamp()));
-    const addr = SocketAddr.random(rng.random());
+    var prng = std.rand.DefaultPrng.init(@intCast(std.time.milliTimestamp()));
```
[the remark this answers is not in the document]
sonicfromnewyoke: done  [c78469b](https://github.com/Syndica/sig/pull/304/commits/c78469ba8027ccfbb5c3ec4b4c96f382a27f82b4#diff-f371061337ac8309638b3df49171a7870a01cb555abe905931dd2d0fe8d8d100R544)

src/gossip/service.zig
```diff
@@ -3265,23 +3265,23 @@ pub const BenchmarkGossipServiceGeneral = struct {
         const outgoing_channel = gossip_service.packet_incoming_channel;
 
         // generate messages
-        var rand = std.rand.DefaultPrng.init(19);
-        const rng = rand.random();
+        var prand = std.rand.DefaultPrng.init(19);
```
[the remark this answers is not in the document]
sonicfromnewyoke: done [c78469b](https://github.com/Syndica/sig/pull/304/commits/c78469ba8027ccfbb5c3ec4b4c96f382a27f82b4#diff-fbd2efdcd434a0d1a58c3990d825690ff474f00df7a7204e00b3d87cec3bbfa4R3264)

src/gossip/service.zig
```diff
@@ -2925,8 +2925,8 @@ test "test large push messages" {
         var as_lock = gossip_service.active_set_rw.write();
         var as: *ActiveSet = as_lock.mut();
         const prng_seed: u64 = @intCast(std.time.milliTimestamp());
```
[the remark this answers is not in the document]
sonicfromnewyoke: done [c78469b](https://github.com/Syndica/sig/pull/304/commits/c78469ba8027ccfbb5c3ec4b4c96f382a27f82b4#diff-fbd2efdcd434a0d1a58c3990d825690ff474f00df7a7204e00b3d87cec3bbfa4R2927)

src/gossip/service.zig
```diff
@@ -2743,7 +2743,7 @@ test "test build prune messages and handle push messages" {
 
 test "build pull requests" {
     const allocator = std.testing.allocator;
-    var prng = std.rand.DefaultPrng.init(91);
+    var default_rng = std.rand.DefaultPrng.init(91);
```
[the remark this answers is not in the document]
sonicfromnewyoke: done [c78469b](https://github.com/Syndica/sig/pull/304/commits/c78469ba8027ccfbb5c3ec4b4c96f382a27f82b4#diff-fbd2efdcd434a0d1a58c3990d825690ff474f00df7a7204e00b3d87cec3bbfa4R2749)

src/gossip/service.zig
```diff
@@ -2371,8 +2371,8 @@ test "handling prune messages" {
         var as_lock = gossip_service.active_set_rw.write();
         var as: *ActiveSet = as_lock.mut();
         const prng_seed: u64 = @intCast(std.time.milliTimestamp());
```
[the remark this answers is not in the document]
sonicfromnewyoke: done [c78469b](https://github.com/Syndica/sig/pull/304/commits/c78469ba8027ccfbb5c3ec4b4c96f382a27f82b4#diff-fbd2efdcd434a0d1a58c3990d825690ff474f00df7a7204e00b3d87cec3bbfa4R2379)

src/gossip/service.zig
```diff
@@ -2831,23 +2831,23 @@ test "test build push messages" {
...
     var keypair = try KeyPair.create([_]u8{1} ** 32);
     // var id = Pubkey.fromPublicKey(&keypair.public_key);
-    const value = try SignedGossipData.random(rng.random(), &keypair);
+    const value = try SignedGossipData.random(prng.random(), &keypair);
 
     // set the active set
     {
         var as_lock = gossip_service.active_set_rw.write();
         var as: *ActiveSet = as_lock.mut();
         const prng_seed: u64 = @intCast(std.time.milliTimestamp());
```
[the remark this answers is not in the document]
sonicfromnewyoke: done [c78469b](https://github.com/Syndica/sig/pull/304/commits/c78469ba8027ccfbb5c3ec4b4c96f382a27f82b4#diff-fbd2efdcd434a0d1a58c3990d825690ff474f00df7a7204e00b3d87cec3bbfa4R2927)

Rexicon226 (approved):
LGTM

Rexicon226:
Just needs a rebase and should be good to go, unless someone else has any comments.

InKryption:
> ensured the usage of rand parameter for std.Random. rand was used instead of random in this case, to not shadowing function declaration of fn random(..)

@dnut I think this would be a good rationale for using `rand` as the conventional name for an instance of the `std.Random` interface instead of `random`; gives increased consistency, and less rules to think about when naming stuff.

If you agree, I'd like to see that change, otherwise I'm happy to sign off on this as is.

dnut:
> I think this would be a good rationale for using rand as the conventional name for an instance of the std.Random interface instead of random; gives increased consistency, and less rules to think about when naming stuff.

That's a good point. If most people prefer "rand", I won't get in the way.

Personally I still have a mild preference for using the name "random" because I feel it is slightly more clear, and it is more consistent with the broader pattern of naming things like allocator, logger, reader, writer, etc. after the type.

Avoiding shadowing is a pain point with zig in general. I'm starting to grow accustomed to using special names where needed.

InKryption (dismissed):

sonicfromnewyoke:
> Personally I still have a mild preference for using the name "random" because I feel it is slightly more clear, and it is more consistent with the broader pattern of naming things like allocator, logger, reader, writer, etc. after the type.
> 
> Avoiding shadowing is a pain point with zig in general. I'm starting to grow accustomed to using special names where needed.

from my observation, all `random(...)` functions are used for generating random fixtures and for testing purposes only. What about changing the name of such functions to something like `gen_fixture(...)`, `init_fixture(..)`, etc?   It should simplify tackling the obstacle of shadowing

Rexicon226:
`initRandom` would be a suitable name I think. Thoughts?

InKryption:
I think that'd make sense, makes sense to annotate a function used to initialize a new instance of a data type as an init function.

sonicfromnewyoke:
- [x] rename `fn random(...)` -> `fn initRandom(...)`
- [x] make sure all `rand` fn parameters and struct fields have a name `random`

src/accountsdb/genesis_config.zig
```diff
@@ -83,11 +83,11 @@ pub const Rent = extern struct {
...
     burn_percent: u8,
 
-    pub fn random(rand: std.Random) Rent {
+    pub fn initRandom(random: std.Random) Rent {
         return .{
-            .lamports_per_byte_year = rand.int(u64),
-            .exemption_threshold = @bitCast(rand.int(u64)),
-            .burn_percent = rand.uintAtMost(u8, 100),
+            .lamports_per_byte_year = random.int(u64),
+            .exemption_threshold = @bitCast(random.int(u64)),
```
Rexicon226, line 89: You've already done a bit of cleanup in this PR which I really appreciate, so if you don't mind could you change these bit casts to call `random.float(f64)`? ditto the ones in the `Inflation` struct.

src/accountsdb/snapshots.zig
```diff
@@ -370,31 +370,31 @@ pub const HashAge = struct {
...
     max_list_entries: usize,
 ) std.mem.Allocator.Error!BlockhashQueueAges {
     var ages = BlockhashQueueAges.Managed.init(allocator);
     errdefer ages.deinit();
 
-    try sig.rand.fillHashmapWithRng(&ages, rand, rand.uintAtMost(usize, max_list_entries), struct {
-        pub fn randomKey(_rand: std.Random) !Hash {
-            return Hash.random(_rand);
+    try sig.rand.fillHashmapWithRng(&ages, random, random.uintAtMost(usize, max_list_entries), struct {
+        pub fn randomKey(_random: std.Random) !Hash {
```
Rexicon226, line 393: I would be happy calling these `rand`, but I don't think `_random` is a good name. Ditto all other ones.

src/accountsdb/snapshots.zig
```diff
@@ -829,64 +829,64 @@ pub const BankFields = struct {
...
+            .transaction_count = random.int(u64),
+            .tick_height = random.int(u64),
+            .signature_count = random.int(u64),
+            .capitalization = random.int(u64),
+            .max_tick_height = random.int(u64),
+            .hashes_per_tick = if (random.boolean()) random.int(u64) else null,
+            .ticks_per_slot = random.int(u64),
+            .ns_per_slot = random.int(u128),
+            .genesis_creation_time = random.int(sig.accounts_db.genesis_config.UnixTimestamp),
+            .slots_per_year = @bitCast(random.int(u64)),
```
Rexicon226, line 873: If you could call `random.float(f64)` here please.

Rexicon226 (changes requested):
Just a couple of changes, this is really great work :)

src/accountsdb/snapshots.zig
sonicfromnewyoke: done b4ffcdef27060b8d7bca2cf85e36c714f9db647c

src/accountsdb/snapshots.zig
sonicfromnewyoke: done b4ffcdef27060b8d7bca2cf85e36c714f9db647c

src/accountsdb/genesis_config.zig
sonicfromnewyoke: done b4ffcdef27060b8d7bca2cf85e36c714f9db647c

Rexicon226 (dismissed):
LGTM!

InKryption (approved):

PR #306, opened 2024-10-07
feat(prometheus,shred-collector): metrics improvements + add metrics to shred collector

dnut:
The aim of this PR is to improve the observability of the shred collector with prometheus metrics. While doing so, I also made various improvements to the prometheus library. I also made changes throughout the rest of the codebase to take advantage of these improvements.
## Shred Collector
Added metrics to:
- shred receiver
- shred verifier
- shred processor
- repair service
- repair peer provider

### VariantCounter

I added a new metric type: `VariantCounter`: This separately count the occurrence of each variant within an enum or error set. This is a composite metric, similar to a histogram, except the "buckets" are discrete, unordered, and identified by name, instead of representing numeric ranges. Currently it's only used for counting errors but it can also be used to count enum variants, since enums and errors are conceptually almost the same thing. 

I was struggling to decide if each error should be reported as a completely separate metric, or if the VariantCounter as a whole should be a single metric with many labels. I ended up decided to take the label approach because I think it's more flexible, but I'd appreciate feedback on this. See this commit for switching to labels: 0afc7500e7f5ef7caa60a931c0522e2c1e167a59

### Initiailize structs containing many metrics

We have some structs like this:
```zig
const MyMetricsStruct = struct {
    event_count: *Counter,
    current_state: *Gauge(u64),
    my_hist: *Histogram,
};
```

Usually we define a custom init function for each of these structs. These init functions use comptime code to initialize all the fields. I moved all this logic into a single place. It is now in these methods in the Registry struct:

- `initStruct`: Init an entire metrics struct that contains many metrics
- `initFields`: Init metrics fields within a struct that has other data

Configuration:
  - `Metric.metric_type` This is provided within the definitions for Counter, Gauge, etc. to identify the type of metric so `initStruct` and `initFields` can work properly.
  - `MyMetricsStruct.prefix` optional: For namespacing of metrics. Within the struct containing metrics, this will string will be prefixed to any metric names
  - `MyMetricsStruct.buckets` required for histograms. If the struct contains any histograms, you need to specify the buckets either with a const or a function that takes the field name and returns its buckets.

### Existing metrics tweaks
- `Counter`: remove `set` method
- `Gauge`: add min and max methods


src/shred_collector/shred_receiver.zig
```diff
@@ -106,12 +116,16 @@ pub const ShredReceiver = struct {
         comptime is_repair: bool,
     ) !void {
         if (packet.size == REPAIR_RESPONSE_SERIALIZED_PING_BYTES) {
-            if (try self.handlePing(&packet)) |p| try response_sender.send(p);
+            if (try self.handlePing(&packet)) |p| {
```
0xNineteen, line 119: rename `p` to `pong_packet` ?

src/shred_collector/shred_receiver.zig
```diff
@@ -192,10 +211,34 @@ const REPAIR_RESPONSE_SERIALIZED_PING_BYTES = 132;
 const RepairPing = union(enum) { Ping: Ping };
 
 pub const ShredReceiverMetrics = struct {
-    invalid_repair_pings: *sig.prometheus.Counter,
+    received_count: *Counter,
```
0xNineteen, line 214: we use `num_` in `BlockstoreRpcApiMetrics` and here we use `_count` and in geyser we use `_total` and `_n` lol -- im cool with any of them but should be consistent

src/shred_collector/shred_verifier.zig
```diff
@@ -7,38 +7,71 @@ const shred_layout = sig.ledger.shred.layout;
...
     unverified_shred_receiver: *Channel(Packet),
     /// me --> shred processor
     verified_shred_sender: *Channel(Packet),
     leader_schedule: SlotLeaderProvider,
 ) !void {
+    const metrics = try registry.initStruct(Metrics);
     while (!exit.load(.acquire) or
         unverified_shred_receiver.len() != 0)
     {
+        var count: usize = 0;
```
0xNineteen, line 31: `packet_count` ?

src/prometheus/variant_counter.zig
```diff
@@ -0,0 +1,143 @@
...
+const Atomic = std.atomic.Value;
+
+const Metric = prometheus.metric.Metric;
+const MetricType = prometheus.metric.MetricType;
+
+/// Separately count the occurrence of each variant within an enum or error set.
+///
+/// This is a composite metric, similar to a histogram, except the "buckets" are
+/// discrete, unordered, and identified by name, instead of representing numeric ranges.
+pub fn VariantCounter(comptime EnumOrError: type) type {
```
0xNineteen, line 13: :fire:

src/prometheus/variant_counter.zig
```diff
@@ -0,0 +1,143 @@
...
+            .int_to_index = &int_to_index,
+            .len = index_to_int.len,
+        };
+    }
+
+    pub fn index(self: Self, err: self.EnumOrError) usize {
+        return self.int_to_index[toInt(err)];
+    }
+
+    pub fn get(self: Self, index_: usize) self.EnumOrError {
```
0xNineteen, line 108: is this used anywhere?

src/prometheus/variant_counter.zig
```diff
@@ -0,0 +1,143 @@
...
+/// not guaranteed. This is particularly useful for variants, whose
+/// integers are likely not continuous.
+///
+/// You can provide an variant to determine its index, or provide an
+/// index to get the respective variant.
+///
+/// This struct must be initialized at comptime, but it can be used
+/// at any time.
+const VariantIndexer = struct {
+    EnumOrError: type,
```
0xNineteen, line 59: this originally confused me because i thought it was a struct which includes either enums or errors -- not one or the other -- maybe we rename to `Set` and then comptime ensure its either an enum or error set (similar to how we have logger) -- something like:
```
pub const Set = union(enum) {
    error_set: ErrorSet,
    enum: Enum,
```

src/prometheus/variant_counter.zig
```diff
@@ -0,0 +1,143 @@
...
+    offset: u16,
+    index_to_int: []const u16,
+    int_to_index: []const u16,
```
0xNineteen, line 62: not sure why we need all this -- `toInt` `fromInt` and `toName` can just use `@intFromEnum/fromError` at runtime (rn its like a cache for a really cheap operation?)

0xNineteen (changes requested):
looks like a great improvement - a few nits - and a few comments that are more global questions of 'how we name/organize metrics' -- which is more than what this PR asked for but i think it has the chance to make a large improvement to consistency if we decide these things here

src/shred_collector/shred_receiver.zig
dnut: I think `_count` and `_total` conform to the standard naming conventions, whereas `num_` does not. Grafana will complain if you don't use one of these.

I prefer "count" over "total" because I feel it is less ambiguous. "Count" clearly describes a number of discrete items or events, which is what we're typically using Counters for. I would expect anything called a "count" to be a positive integer. "Total" on the other hand is much more general, it could refer to a sum of any arbitrary numbers.

src/shred_collector/repair_service.zig
```diff
@@ -63,16 +65,28 @@ pub const RepairService = struct {
         }
     });
 
+    const Metrics = struct {
```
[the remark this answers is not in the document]
dnut: In this file there are multiple structs each with their own metrics that are specific to the struct. Each of those structs are small so it doesn't make them unwieldy to include a small metrics struct, and those metrics structs are not used anywhere outside of them. So it felt right to define it within the struct. It feels like slightly more clutter for the file as a whole if the metrics are mixed in with all the other structs.

Initially I had the fields from the metrics struct just as fields directly within the service structs. The `Metrics` struct was just a way to group the metrics more cleanly within the struct. I think it would also be fine to put metrics directly in service structs, but at this point I would personally only do that if it was only a single metric.

I don't think it makes a big difference though, so I'm fine with it either way. 

Personally I don't think I would advocate for a strict rule to standardize how this is done. I feel that the context dictates what approach looks the cleanest. But I can also see an advantage of consistency. So I don't have very strong feelings about it.

src/prometheus/registry.zig
```diff
@@ -71,6 +72,75 @@ pub fn Registry(comptime options: RegistryOptions) type {
...
+                .counter => try self.getOrCreateCounter(name),
+                .variant_counter => try self.getOrCreateVariantCounter(name, MetricType.Type),
+                .gauge => try self.getOrCreateGauge(name, MetricType.Data),
+                .gauge_fn => @compileError("GaugeFn does not support auto-init."),
+                .histogram => try self.getOrCreateHistogram(
+                    name,
+                    if (@typeInfo(@TypeOf(Config.buckets)) == .Fn)
+                        Config.buckets(name)
+                    else
+                        &Config.buckets,
```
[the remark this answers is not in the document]
dnut: > maybe change this field name to default_buckets so its more clear that it will be used for all histograms

_First some context:_ Initially I had `Config.buckets(name)` as the only option, but I found that in many cases I was just implementing that function by ignoring the name and returning the same buckets no matter what. The function has to be written a bit strangely to ensure the comptime slice works correctly. I decided to allow it to be declared as a const for that common case, instead of the function, to reduce confusing boilerplate.

Normally I'd think of the word "default" to indicate that it's sort of like a fallback in the absence of explicit configuration. But in this case, you need to manually configure the buckets regardless. You need a `buckets` decl that can configure every single histogram in the struct. You just have the option to specify that configuration as being either dependent or independent of the field name. But you can't do both, this would make no sense. Since they are mutually exclusive, there's a benefit of using the same name for both, the compiler won't allow you to specify both by accident.

> is there a comptime way to ensure Config has either a buckets field or method 

Yes it currently checks this at comptime.

> and print a more helpful panic? (rn if its missing its not that great of an error message)

here's the compile error that we have now:
```zig
src/prometheus/registry.zig:136:49: error: struct 'ledger.writer.ScanAndFixRootsMetrics' has no member named 'buckets'
                    if (@typeInfo(@TypeOf(Config.buckets)) == .Fn)
                                          ~~~~~~^~~~~~~~
src/ledger/writer.zig:306:36: note: struct declared here
pub const ScanAndFixRootsMetrics = struct {
                                   ^~~~~~
referenced by:
    initStruct__anon_41729: src/prometheus/registry.zig:80:36
    test.setAndChainConnectedOnRootAndNextSlots: disconnected: src/ledger/writer.zig:475:62
    remaining reference traces hidden; use '-freference-trace' to see all reference traces
```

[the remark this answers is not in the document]
dnut: > -- rn the way we handle buckets is kind of hidden logic but not sure if theres a better way 🤔

I agree this could be better but I couldn't immediately think of a way to improve the situation. There is something nice about having explicit traits/interfaces built into the language.

src/prometheus/registry.zig
```diff
@@ -71,6 +72,75 @@ pub fn Registry(comptime options: RegistryOptions) type {
...
+        /// Returns the number of fields that were *not* initialized.
+        pub fn initFields(
+            self: *Self,
+            /// Mutable pointer to a struct containing metrics.
+            metrics_struct: anytype,
+        ) GetMetricError!usize {
+            const Struct = @typeInfo(@TypeOf(metrics_struct)).Pointer.child;
+            const fields = @typeInfo(Struct).Struct.fields;
+            var num_fields_skipped: usize = fields.len;
+            inline for (@typeInfo(Struct).Struct.fields) |field| {
```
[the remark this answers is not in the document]
dnut: Debug logging mean you need to pass in a logger whenever you call this, which is doable but adds some dependency burden.

I made this function return a usize with the total number of skipped fields, so you can assert it in the calling context to ensure you're not missing anything. That ends up looking something like this:

```zig
        var self: Self = undefined;
        const registry = globalRegistry();
        std.debug.assert(try registry.initFields(&self) == 1);
        self._logging_fields = .{ .logger = logger };
```

I feel like the assertion approach is pretty good at detecting whether you're missing some fields. It's not perfect but if you add new fields and forget to init them in the struct, you'll get an error pretty quickly when you run the code. I could also make it a comptime check if desired, the tradeoff would be mean you *always* have to run the check, and don't have the option to ignore it. I went with the assertion approach because I liked the flexibility and it seemed good enough.

src/prometheus/variant_counter.zig
dnut: It could be a union but the whole idea of this field is to store the concrete type of the data. So the two variants of Set would need to have type `type`, and you'd need some way to extract the type out of it. Something like this:

```zig
pub const Set = union(enum) {
    error_set: type,
    enum: type,

    fn Type(self: Set) type {
        return switch(self) {
            .error_set => |e| e,
            .enum => |e| e,
        };
    }
}
```
and then most places where `self.EnumOrError` is used, you'll need to instead call it like `self.EnumOrError.Type()`

src/prometheus/variant_counter.zig
dnut: There's an array of counters for each error. You need to be able to associate error to one of those counters. Let's say you have three errors, then you'd like to assign each of those errors one of the numbers 0,1,2 so you can identify which counter they are associated with in the array of counters. That's the job of this struct.

The three errors that you're tracking may have integer representations like 121, 249, and 310. They need to be assumed that they are basically random. The goal here is to map those numbers to 0, 1, and 2.

src/prometheus/variant_counter.zig
dnut: I guess not. It's the inverse of `index` but not used so I can remove it c42bd45a1621a68db97e32d507b1529e3c1f9007

src/prometheus/registry.zig
```diff
@@ -10,6 +10,7 @@ const OnceCell = @import("../sync/once_cell.zig").OnceCell;
 
 const Metric = @import("metric.zig").Metric;
 const Counter = @import("counter.zig").Counter;
+const VariantCounter = @import("variant_counter.zig").VariantCounter;
 const Gauge = @import("gauge.zig").Gauge;
```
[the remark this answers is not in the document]
dnut: d3009b67be3c2d67b12928921d2744ae06fdfbc4

src/geyser/core.zig
```diff
@@ -77,13 +77,7 @@ pub const GeyserWriterStats = struct {
     geyser_writer_total_bytes: *Counter,
```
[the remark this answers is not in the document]
dnut: e7a1ec10808703990147bd6c80551dab99d8b926

src/geyser/core.zig
```diff
@@ -260,17 +254,7 @@ pub const GeyserReaderStats = struct {
     const GaugeU64 = Gauge(u64);
```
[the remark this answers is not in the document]
dnut: e7a1ec10808703990147bd6c80551dab99d8b926 (same as above)

src/shred_collector/shred_processor.zig
```diff
@@ -94,14 +108,32 @@ fn processShred(
...
-            error.SlotUnderflow, error.SlotOverflow => return,
+            error.SlotUnderflow, error.SlotOverflow => {
+                metrics.set_last_shred_error.observe(err);
+                return;
+            },
         };
     }
 }
+
+const Metrics = struct {
```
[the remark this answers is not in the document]
dnut: My mentality behind the concise naming is that the context is already specified - in this case by the file name. This is the Metrics struct in shred_processor.zig, i.e. `shred_processor.Metrics`, i.e. the metrics for the shred processor. I'm fine calling this ShredProcessorMetrics if you think it improves clarity, it just seems unnecessary to me.

Regarding stats vs metrics I don't have a strong opinion on that either. Metrics seems like the natural choice since that's the terminology used in prometheus. But after I started that pattern I noticed that it conflicts with the Stats name in other places. So I'd definitely like to consolidate on a single name, so I'm happy to switch everything to Metrics

Maybe I'm being pedantic, but I think it's worth noting that the pattern as defined `{struct_name}Metrics` won't always apply, because metrics structs are not necessarily 1-to-1 with service structs. In the shred collector case, case there is no associated struct (except I guess the file that contains it, which is technically a struct). Sometimes there will be multiple associated structs, and sometimes there will be multiple metrics structs for one service struct.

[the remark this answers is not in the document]
0xNineteen: ok makes sense - im flexible, maybe we keep it `Metrics` if the context is already defined and theres no conflict

src/shred_collector/shred_receiver.zig
0xNineteen: im good with using `_count` everywhere

src/prometheus/variant_counter.zig
0xNineteen: huh ok that makes sense i was unaware that error sets would map to random integer representations (assumed it would be incremental from 0)

src/prometheus/variant_counter.zig
0xNineteen: mmm i see yeah thats not great - the way it is rn is ok

0xNineteen:
can we add the comments on metrics to contributing.md?
- call the struct `Metrics` if the context is already defined and theres no conflict in the file
- if the Metrics struct is small enough, it can be defined within the struct, otherwise it can be defined directly underneath the struct

src/prometheus/registry.zig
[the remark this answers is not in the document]
0xNineteen: > here's the compile error that we have now:

yeah i more so meant can we have a better comptime error message -- one which describes how Metric structs which have histograms need to either provide a bucket field which will be used for all histograms or a bucket(field_name) method which returns the buckets for a given field name

usually i would be fine with keeping the compilers default compiler error but the bucket's usage is so hidden that a more descriptive comptime message would help devs debug a significant amount imo

src/prometheus/variant_counter.zig
dnut: I put something like that in the struct's doc, but I accidentally deleted the word "error". It was supposed to say this:
```
/// Assigns a continuous sequence of integers starting at 0 to the
/// variants. For example, if you have an variant set with 10 variants,
/// it will assign the numbers 0-9 to each of the variants.
///
/// Enums usually already have numbers with this property, but it's
/// not guaranteed. This is particularly useful for **errors**, whose
/// integers are likely not continuous.
```
Note where it says **errors** (currently in the code it says "variants" which is a mistake)

Is this good enough or do you think it needs further explanation?

dnut: Also FYI, error integers are globally unique. They're not only unique within an error set. This is required because error sets need to intersect with each other. The same error needs to be usable across multiple error sets. The most extreme example is `anyerror`, which is the set of all errors. Enums on the other hand, are strictly non-intersecting sets. But then you have the option to specify arbitrary integers for each variant, so it could also be non-continuous.

It's unspecified how error integers are assigned by the compiler. I've seen large gaps between the integers for errors within the same set. They're also not guaranteed to be consistent across different compilations.

0xNineteen: yeah thats good -- huh very interesting thanks for explaining

0xNineteen:
also - should this PR include a corresponding grafana dashboard? doesnt seem to be commited

src/prometheus/registry.zig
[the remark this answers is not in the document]
dnut: I like the idea of just having the simplest word "buckets" just be the canonical way to configure the histogram buckets

I agree a more descriptive error message is nice to have. I was just trying to keep the implementation simple, and I figured it was enough information to figure it out. It's significantly more complex to have custom error messages.

Also regarding the naming, I like something about just having the simplest and most obvious word "buckets" just be the canonical way to configure the histogram buckets, regardless of whether it's function or an array. I think "buckets" the easiest name to remember when writing the struct. If I kinda forgot exactly what name to use, "buckets" would probably be my first guess. You're less likely to see any comptime errors at all, compared to seeing an error and needing to have a nice error message, because your first guess would be correct. Whereas having different names with multiple words, it's a little harder to get it correct unless you see it written in front of you already. Also I felt that "buckets" is descriptive enough of a name to be pretty clear what it means. But I do also see the advantage of an explicit function name helping clarify the purpose of the string parameter.

src/accountsdb/db.zig
```diff
@@ -113,25 +83,38 @@ pub const AccountsDBStats = struct {
...
-                    break :blk try registry.getOrCreateHistogram(field.name, histogram_kind.buckets());
-                },
-                else => @compileError("Unsupported field type"),
-            };
-        }
-        return self;
+        return globalRegistry().initStruct(Self);
+    }
+
+    pub fn buckets(comptime field_name: []const u8) []const f64 {
```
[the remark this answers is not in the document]
dnut: 38c9d9c6c17beea7b56a12925f7552845c5a41d1

what about histogramBucketsForField?

dnut:
> also - should this PR include a corresponding grafana dashboard? doesnt seem to be commited

I figured I would do this later when I actually need to gain insights from the metrics. For now I just wanted to expose them in the code. It would be great to also have a dashboard already, but I just didn't prioritize it at this point.

src/shred_collector/shred_receiver.zig
dnut: e6312ca0a63ec918fb67d3a891fcaab44e9464be

src/shred_collector/shred_receiver.zig
```diff
@@ -91,9 +94,16 @@ pub const ShredReceiver = struct {
     ) !void {
         while (!self.exit.load(.acquire)) {
             for (receivers) |receiver| {
+                var count: usize = 0;
```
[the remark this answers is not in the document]
dnut: e6312ca0a63ec918fb67d3a891fcaab44e9464be

src/shred_collector/shred_processor.zig
```diff
@@ -10,10 +10,15 @@ const Atomic = std.atomic.Value;
...
+const Histogram = sig.prometheus.Histogram;
 const Logger = sig.trace.Logger;
 const Packet = sig.net.Packet;
+const Registry = sig.prometheus.Registry;
 const Shred = sig.ledger.shred.Shred;
 const ShredInserter = sig.ledger.ShredInserter;
+const SlotOutOfBounds = shred_collector.shred_tracker.SlotOutOfBounds;
+const VariantCounter = sig.prometheus.VariantCounter;
 
 // TODO: add metrics (e.g. total count of shreds processed)
```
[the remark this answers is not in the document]
dnut: ca6e0de009a42df00787291b21ee1654b46514d0

src/shred_collector/shred_processor.zig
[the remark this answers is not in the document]
dnut: Stats->Metrics: 90689ef28bc892cbea8d903212b7d3633e008e77

src/shred_collector/shred_receiver.zig
dnut: Are you suggesting changing the metrics names now to all be consistent? My only concern here is breaking existing grafana dashboards.

dnut:
> can we add the comments on metrics to contributing.md?
> 
>     * call the struct `Metrics` if the context is already defined and theres no conflict in the file
> 
>     * if the Metrics struct is small enough, it can be defined within the struct, otherwise it can be defined directly underneath the struct

9b17bfb177eb1cd3e24ae8e42269091e8eb0f601

src/shred_collector/shred_receiver.zig
0xNineteen: true - pushed it to a backlog issue - https://github.com/orgs/Syndica/projects/2/views/10?pane=issue&itemId=82929996

0xNineteen:
nice - just commited a quick fit to the naming -- some cases updated the struct name but not the field (ie, `stats: GossipMetrics` -- also for AccountsDB reverted the changes to `BankHashStatsMap` (this struct isnt used for prometheus metrics)

PR #307, opened 2024-10-08
refactor(trace): reimplement Entry to reduce allocations and simplify code

dnut:
This changes the `Entry` type so that it directly contains any logger "fields" within the struct as normal struct fields instead of allocating strings repeatedly along the way. This unlocked some simplifications to the code and allows the code to be more organized. It also reduces allocations which theoretically could have a performance benefit.

There are no meaningful changes to the logger api. You still use it in exactly the same way as before. Note that I haven't changed anything outside the `trace` module. While it's true the Entry type is different, the entry type is not intended for use outside the logger, so it's not really an api change in my opinion.

## Explanation of Change
The basic unchanged idea of the logger is that you call some method like `logger.info()` and then you're supposed to directly chain that with other method calls. For normal logging, you would just chain that with a call to `log` or `logf` as in `logger.info().log("hello world")`. If you want to include fields in the log, you would put some calls to `field` in between `info` and `logf`, as in `logger.info().field("src_addr", src_ip).log("message receive")`.

The way this already worked is with the builder pattern. The `info` method returns an item of type `Entry` which is responsible for tracking all the accumulating state. Then when you finish up by calling `Entry.log`, this function is responsible for calling back to the logger to actually log the message.

That describes how it was, and how it still is. So, what have I changed?

Previously `Entry` was a tagged union that kept a bunch of state relevant to the specific logger implementation, with different variants for different logger implementations. The `field` functions perform allocations to construct the field string in the way that this particular logger wants to have a field string look. Then the `log` function implements the approach to logging for this particular logger. So basically, each logger implementation had its implementation fractured across two structs, its logger (e.g. ChannelPrintLogger) and its Entry (e.g. ChannelPrintEntry).

With the new change, Entry has only one responsibility. All it does is store the fields. And it stores them exactly as they were passed, instead of formatting them for the specific logger. Likewise, it's just a struct with a single implementation, instead of being a tagged union representing different implementations. When you call `log`, it just passes the fields directly to the logger. So the loggers now include their own `log` function instead of that logic being defined in various Entry implementations.

To enable the Entry to store the data as is, without needing to format it or produce any allocations, the Entry struct needs to be generic. The entry contains a generic Fields struct that contains whatever fields were added so far. Each time you add a field to the Entry, it returns a new version of Entry that has a fields struct with one more field.

## Reduce Allocations
This reduces allocations to either 0 or 1 allocation per log message.

For the direct print logger, there are zero allocations needed. The fields and the log message are directly formatted by the stderr writer.

For the channel logger, you only need a single allocation. The fields and the log message are written to a single allocated string. That string is sent over the channel, and then written to stderr.

## Code Simplification
Several layers of the code were able to be removed. For example I was able to remove some code from each logger implementation that was redundant with the ScopedLogger interface. I was also able to simplify log formatting because there's no longer a need to bundle up all the parameters into a struct and pass that around before formatting it. The data can be directly formatted to a single string.

The main simplification is that the Entry is just a basic struct containing fields. It doesn't need to do anything more than add fields to a struct. The tradeoff here is that you need to understand comptime code to fully grasp how the entry works under the hood. But the comptime code is fully encapsulated in a single place that doesn't concern itself with any other responsibilities, so that knowledge is not needed unless you're interested making an improvement to that very mechanism.

## Code Clarity & Organization
The code is clearer and more organized because the logger implementations fully encapsulate what it means to log with that logger. You no longer need to fracture a logger implementation halfway between the logger and halfway in some struct that represents an undefined "entry" concept. Previously the entry was just a spillover of the logger and it was unclear why it needs to exist, and just caused confusion until you fully understood the whole logger paradigm. Now it's very clear. The entry holds the fields, and nothing else. The logger implements logging in its entirety.

The code is also clearer now in what it means to "implement a logger". All you need to do is implement a struct with a `log` method conforming to the function signature expected by `ScopedLogger.private_log`. Implement that one method, and you have a fully functional logger.

### DIFF
--- a/src/trace/entry.zig
+++ b/src/trace/entry.zig
@@ -1,245 +1,90 @@
 const std = @import("std");
-const logfmt = @import("logfmt.zig");
-const Level = @import("level.zig").Level;
-const Channel = @import("../sync/channel.zig").Channel;
+const sig = @import("../sig.zig");
+const trace = @import("lib.zig");
+
+const logfmt = trace.logfmt;
+
+const Level = trace.level.Level;
+const ScopedLogger = trace.log.ScopedLogger;
+const Channel = sig.sync.Channel;
 const AtomicBool = std.atomic.Value(bool);
 
-pub const Entry = union(enum) {
-    channel_print: ChannelPrintEntry,
-    direct_print: DirectPrintEntry,
-    noop,
+pub fn NewEntry(comptime scope: ?[]const u8) type {
+    return Entry(struct {}, scope);
+}
 
-    const Self = @This();
+pub fn Entry(comptime Fields: type, comptime scope: ?[]const u8) type {
+    return struct {
+        logger: ScopedLogger(scope),
+        level: Level,
+        fields: Fields,
 
-    pub fn init(allocator: std.mem.Allocator, channel: *Channel(logfmt.LogMsg), log_level: Level) Self {
-        return .{ .channel_print = ChannelPrintEntry.init(allocator, channel, log_level) };
-    }
+        const Self = @This();
 
-    pub fn deinit(self: Self) void {
-        switch (self) {
-            .noop => {},
-            .channel_print => |impl| impl.deinit(),
-            .direct_print => |impl| impl.deinit(),
+        pub inline fn init(logger: ScopedLogger(scope), level: Level) Self {
+            return .{ .logger = logger, .level = level, .fields = .{} };
         }
-    }
 
-    pub fn field(self: Self, name: []const u8, value: anytype) Self {
-        switch (self) {
-            .noop => {
-                return self;
-            },
-            .channel_print => |entry| {
-                var log_entry = entry;
-                _ = log_entry.field(name, value);
-                return Entry{ .channel_print = log_entry };
-            },
-            .direct_print => |entry| {
-                var log_entry = entry;
-                _ = log_entry.field(name, value);
-                return Entry{ .direct_print = log_entry };
-            },
-        }
-    }
-
-    pub fn log(self: Self, comptime msg: []const u8) void {
-        switch (self) {
-            .noop => {},
-            .channel_print => |impl| {
-                var logger = impl;
-                logger.log(msg);
-            },
-            .direct_print => |impl| {
-                var logger = impl;
-                logger.log(msg);
-            },
-        }
-    }
-
-    pub fn logf(self: Self, comptime fmt: []const u8, args: anytype) void {
-        switch (self) {
-            .noop => {},
-            .channel_print => |impl| {
-                var logger = impl;
-                logger.logf(fmt, args);
-            },
-            .direct_print => |impl| {
-                var logger = impl;
-                logger.logf(fmt, args);
-            },
-        }
-    }
-};
-
-pub const ChannelPrintEntry = struct {
-    allocator: std.mem.Allocator,
-    scope: ?[]const u8,
-    log_level: Level,
-    fields: std.ArrayList(u8),
-    exit_sig: std.atomic.Value(bool),
-    channel: *Channel(logfmt.LogMsg),
-    const Self = @This();
-
-    pub fn init(
-        allocator: std.mem.Allocator,
-        scope: ?[]const u8,
-        channel: *Channel(logfmt.LogMsg),
-        log_level: Level,
-    ) Self {
-        return .{
-            .allocator = allocator,
-            .scope = scope,
-            .exit_sig = AtomicBool.init(false),
-            .log_level = log_level,
-            .fields = std.ArrayList(u8).init(allocator),
-            .channel = channel,
-        };
-    }
-
-    pub fn deinit(self: *Self) void {
-        self.fields.deinit();
-    }
-
-    pub fn field(self: *Self, name: []const u8, value: anytype) *Self {
-        const min_capacity = self.fields.items.len + logfmt.countField(name, value);
-        self.fields.ensureTotalCapacity(min_capacity) catch return self;
-        logfmt.fmtField(self.fields.writer(), name, value);
-        return self;
-    }
-
-    pub fn log(self: *Self, comptime msg: []const u8) void {
-        const log_msg = logfmt.LogMsg{
-            .level = self.log_level,
-            .maybe_scope = self.scope,
-            .maybe_msg = msg,
-            .maybe_fields = self.fields.toOwnedSlice() catch |err| {
-                std.debug.print("Processing fields failed with err: {any}", .{err});
-                self.deinit();
-                return;
-            },
-            .maybe_fmt = null,
-        };
-
-        self.channel.send(log_msg) catch |err| {
-            std.debug.print("Send msg through channel failed with err: {any}", .{err});
-            self.deinit();
-            return;
-        };
-    }
-
-    pub fn logf(self: *Self, comptime fmt: []const u8, args: anytype) void {
-        // Get memory for formatting message.
-        const msg_buf = self.allocBuf(std.fmt.count(fmt, args)) catch |err| {
-            std.debug.print("allocBuff failed with err: {any}", .{err});
-            self.deinit();
-            return;
-        };
-        var fmt_message = std.io.fixedBufferStream(msg_buf);
-        // Format message.
-        logfmt.fmtMsg(fmt_message.writer(), fmt, args);
-
-        const log_msg = logfmt.LogMsg{
-            .level = self.log_level,
-            .maybe_scope = self.scope,
-            .maybe_msg = null,
-            .maybe_fields = self.fields.toOwnedSlice() catch |err| {
-                std.debug.print("Processing fields failed with err: {any}", .{err});
-                self.deinit();
-                return;
-            },
-            .maybe_fmt = fmt_message.getWritten(),
-        };
-
-        self.channel.send(log_msg) catch |err| {
-            std.debug.print("Send msg through channel failed with err: {any}", .{err});
-            self.deinit();
-            return;
-        };
-    }
-
-    // Utility function for allocating memory from RecycleFBA for part of the log message.
-    fn allocBuf(self: *const Self, size: u64) ![]u8 {
-        const buf = blk: while (true) {
-            const buf = self.allocator.alloc(u8, size) catch {
-                std.time.sleep(std.time.ns_per_ms);
-                if (self.exit_sig.load(.unordered)) {
-                    return error.MemoryBlockedWithExitSignaled;
-                }
-                continue;
+        /// Add a field to the log message.
+        pub inline fn field(
+            self: Self,
+            comptime name: [:0]const u8,
+            value: anytype,
+        ) Entry(FieldsPlus(name, @TypeOf(value)), scope) {
+            if (self.logger == .noop) return .{
+                .logger = .noop,
+                .level = undefined,
+                .fields = undefined,
             };
-            break :blk buf;
-        };
-        errdefer {
-            self.allocator.free(buf);
+            var new_fields: FieldsPlus(name, @TypeOf(value)) = undefined;
+            inline for (@typeInfo(Fields).Struct.fields) |existing_field| {
+                @field(new_fields, existing_field.name) = @field(self.fields, existing_field.name);
+            }
+            @field(new_fields, name) = value;
+            return .{
+                .logger = self.logger,
+                .level = self.level,
+                .fields = new_fields,
+            };
         }
-        return buf;
-    }
-};
 
-/// An instance of `Entry` that logs directly to std.debug.print, instead of sending to channel.
-pub const DirectPrintEntry = struct {
-    const builtin = @import("builtin");
+        /// Log the message using the logger, including all fields that are saved in the entry.
+        pub inline fn log(self: Self, comptime message: []const u8) void {
+            self.logger.private_log(self.level, self.fields, message, .{});
+        }
 
-    log_level: Level,
-    scope: ?[]const u8,
-    allocator: std.mem.Allocator,
-    log_msg: std.ArrayList(u8),
+        /// Log the message using the logger, including all fields that are saved in the entry.
+        pub inline fn logf(self: Self, comptime fmt: []const u8, args: anytype) void {
+            self.logger.private_log(self.level, self.fields, fmt, args);
+        }
 
-    const Self = @This();
-
-    pub fn init(
-        allocator: std.mem.Allocator,
-        scope: ?[]const u8,
-        log_level: Level,
-    ) Self {
-        return .{
-            .allocator = allocator,
-            .scope = scope,
-            .log_level = log_level,
-            .log_msg = std.ArrayList(u8).init(allocator),
-        };
-    }
-
-    pub fn deinit(self: *Self) void {
-        self.log_msg.deinit();
-    }
-
-    pub fn field(self: *Self, name: []const u8, value: anytype) *Self {
-        logfmt.fmtField(self.log_msg.writer(), name, value);
-        return self;
-    }
-
-    pub fn log(self: *Self, comptime msg: []const u8) void {
-        defer self.deinit();
-        const log_msg = logfmt.LogMsg{
-            .level = self.log_level,
-            .maybe_scope = self.scope,
-            .maybe_msg = msg,
-            .maybe_fields = null,
-            .maybe_fmt = null,
-        };
-
-        const writer = self.log_msg.writer();
-        logfmt.writeLog(writer, log_msg) catch @panic("Failed to write log");
-        std.debug.print("{s}", .{self.log_msg.items});
-    }
-
-    pub fn logf(self: *Self, comptime fmt: []const u8, args: anytype) void {
-        defer self.deinit();
-        // Format message.
-        var fmt_msg = std.ArrayList(u8).initCapacity(self.allocator, 256) catch @panic("could not initCapacity for message");
-        defer fmt_msg.deinit();
-        logfmt.fmtMsg(fmt_msg.writer(), fmt, args);
-
-        const log_msg = logfmt.LogMsg{
-            .level = self.log_level,
-            .maybe_scope = self.scope,
-            .maybe_msg = null,
-            .maybe_fields = null,
-            .maybe_fmt = fmt_msg.items,
-        };
-
-        const writer = self.log_msg.writer();
-        logfmt.writeLog(writer, log_msg) catch @panic("Failed to write log");
-        std.debug.print("{s}", .{self.log_msg.items});
-    }
-};
+        /// Returns a new struct type based on Fields, just with one more field added.
+        fn FieldsPlus(comptime field_name: [:0]const u8, comptime FieldType: type) type {
+            const info = @typeInfo(Fields);
+            var new_fields: [1 + info.Struct.fields.len]std.builtin.Type.StructField = undefined;
+            for (info.Struct.fields, 0..) |existing_field, i| {
+                new_fields[i] = existing_field;
+            }
+            const ActualFieldType = switch (@typeInfo(FieldType)) {
+                .ComptimeFloat => f64,
+                .ComptimeInt => u64,
+                else => FieldType,
+            };
+            new_fields[info.Struct.fields.len] = .{
+                .name = field_name,
+                .type = ActualFieldType,
+                .default_value = null,
+                .is_comptime = false,
+                .alignment = @alignOf(FieldType),
+            };
+            const new_struct = std.builtin.Type.Struct{
+                .layout = .auto,
+                .backing_integer = null,
+                .fields = &new_fields,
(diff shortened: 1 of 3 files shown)


src/trace/entry.zig
```diff
@@ -1,245 +1,87 @@
...
+                .ComptimeInt => u64,
+                else => FieldType,
+            };
+            new_fields[info.Struct.fields.len] = .{
+                .name = field_name,
+                .type = ActualFieldType,
+                .default_value = null,
+                .is_comptime = false,
+                .alignment = @alignOf(FieldType),
             };
```
InKryption, line 76: I know this is in draft, but just want to quickly provide this suggestion:
```suggestion
            const ActualFieldType = switch (@typeInfo(FieldType)) {
                .ComptimeFloat => f64,
                .ComptimeInt => u64,
                else => FieldType,
            };
            const new_fields = info.Struct.fields ++ &[_]std.builtin.Type.StructField{.{
                .name = field_name,
                .type = ActualFieldType,
                .default_value = null,
                .is_comptime = false,
                .alignment = 0,
            }};
```
And then also change the line below from `.fields = &new_fields` to `.fields = new_fields` - github is stupid and won't let me include the lines below this one in the suggestion.
```

**Comment by InKryption** [comment]
By the way, interesting approach with the entry names used as the field names directly. It seems like it's the right way to go, and gives us the nice property that entry names are statically verified at compile time to each be unique (I assume we don't need multiple same-named entries anyway).

File: src/trace/entry.zig [hunk 2]
```diff
@@ -1,245 +1,87 @@
...
-    direct_print: DirectPrintEntry,
-    noop,
+pub fn NewEntry(comptime scope: ?[]const u8) type {
+    return Entry(struct {}, scope);
+}
 
-    const Self = @This();
+pub fn Entry(comptime Fields: type, comptime scope: ?[]const u8) type {
+    return struct {
+        logger: ScopedLogger(scope),
```
dadepo, line 15: This is a nice improvement 👍. I considered having the Entry contain the actual logger instead of it being the one responsible for sending the message for logging but ran into some issue I could not get to compile.

src/trace/entry.zig
```diff
@@ -1,245 +1,87 @@
...
+        level: Level,
+        fields: Fields,
```
dadepo, line 17: The win here is that `fields` is no longer an `ArrayList` (which requires allocation) but it is now essentially a struct.

src/trace/entry.zig
```diff
@@ -1,245 +1,87 @@
...
-                var log_entry = entry;
-                _ = log_entry.field(name, value);
-                return Entry{ .direct_print = log_entry };
-            },
+        /// Add a field to the log message.
+        pub inline fn field(
+            self: Self,
+            comptime name: [:0]const u8,
+            value: anytype,
+        ) Entry(FieldsPlus(name, @TypeOf(value)), scope) {
```
dadepo, line 30: Interesting function types can be used this way.

src/trace/entry.zig
dnut: 80e603c7cc88d00e324a4ed85602df2afcad2cfe

src/trace/entry.zig
dnut: Is it clear what's going on here?

src/trace/entry.zig
```diff
@@ -1,245 +1,83 @@
...
-        return .{ .channel_print = ChannelPrintEntry.init(allocator, channel, log_level) };
-    }
+        const Self = @This();
 
-    pub fn deinit(self: Self) void {
-        switch (self) {
-            .noop => {},
-            .channel_print => |impl| impl.deinit(),
-            .direct_print => |impl| impl.deinit(),
+        pub inline fn init(logger: ScopedLogger(scope), level: Level) Self {
```
Rexicon226, line 21: Unjustified usage of `inline`, no reason for it to be inline.

src/trace/entry.zig
```diff
@@ -1,245 +1,83 @@
...
-                _ = log_entry.field(name, value);
-                return Entry{ .channel_print = log_entry };
-            },
-            .direct_print => |entry| {
-                var log_entry = entry;
-                _ = log_entry.field(name, value);
-                return Entry{ .direct_print = log_entry };
-            },
+        /// Add a field to the log message.
+        pub inline fn field(
```
Rexicon226, line 26: Shouldn't be `inline`.

src/trace/entry.zig
```diff
@@ -1,245 +1,83 @@
...
-            .channel_print => |impl| {
-                var logger = impl;
-                logger.log(msg);
-            },
-            .direct_print => |impl| {
-                var logger = impl;
-                logger.log(msg);
-            },
+        /// Log the message using the logger, including all fields that are saved in the entry.
+        pub inline fn log(self: Self, comptime message: []const u8) void {
```
Rexicon226, line 49: Shouldn't be `inline`.

src/trace/entry.zig
```diff
@@ -1,245 +1,83 @@
...
-            .channel_print => |impl| {
-                var logger = impl;
-                logger.logf(fmt, args);
-            },
-            .direct_print => |impl| {
-                var logger = impl;
-                logger.logf(fmt, args);
-            },
+        /// Log the message using the logger, including all fields that are saved in the entry.
+        pub inline fn logf(self: Self, comptime fmt: []const u8, args: anytype) void {
```
Rexicon226, line 54: Shouldn't be `inline`.

src/trace/log.zig
```diff
@@ -171,79 +169,55 @@ pub const ChannelPrintLogger = struct {
...
-        if (@intFromEnum(self.max_level) >= @intFromEnum(level)) {
-            switch (level) {
-                .err => self.err(scope).logf(fmt, args),
-                .warn => self.warn(scope).logf(fmt, args),
-                .info => self.info(scope).logf(fmt, args),
-                .debug => self.debug(scope).logf(fmt, args),
-                .trace => self.trace(scope).logf(fmt, args),
-            }
+    // Utility function for allocating memory from RecycleFBA for part of the log message.
+    fn allocBuf(self: *const Self, size: u64) ![]u8 {
```
Rexicon226, line 210: Could you please explain the purpose of this function? Can `log_allocator.alloc` spuriously fail and so we need to call it multiple times? I'm pretty confused why this exists.

src/trace/log.zig
```diff
@@ -171,79 +169,55 @@ pub const ChannelPrintLogger = struct {
...
+        for (0..100) |_| {
+            return self.log_allocator.alloc(u8, size) catch {
+                std.time.sleep(std.time.ns_per_ms);
+                if (self.exit.load(.unordered)) {
```
Rexicon226, line 214: Needs to be `monotonic`.

Rexicon226 (changes requested):
A couple points to fix up. 

To add to my "removing inline" comments, `inline` is a very specific keyword that should only be used to 1. propagate comptime-ness in specific functions or 2. there is an obvious and clear benefit from it, either by i.e removing call frames, LLVM failing to inline the call itself, or something else.

Just putting `inline` everywhere is terrible and will cause slowdowns in runtime and compile time. By using `inline`, you're making the optimizer optimize every single callsite individually, instead of optimizing the function itself and then inlining the callsites as needed. None of the functions here that used `inline` would benefit or have a clear usecase for `inline`.
A small rant to explain for future usages as well :)

src/trace/entry.zig
dadepo: It is. But that is a bit of a biased answer given the familiarity with this part of the code the past couple of weeks.

src/trace/log.zig
dnut: I don't like it either, just copied the existing code to keep the behavior the same.

The basic idea is that you may need to wait for the RecycleFBA allocator to have some memory available. Personally I don't think we should rely on a mechanism like this, but I wasn't trying to deal with it in this pr.

src/trace/entry.zig
dnut: I'm fine with removing inline but just want to go through the reasoning a bit more to flesh out any grey area.

> To add to my "removing inline" comments, inline is a very specific keyword that should only be used to 1. propagate comptime-ness in specific functions or 2. there is an obvious and clear benefit from it, either by i.e removing call frames, LLVM failing to inline the call itself, or something else.

In the past, I've actually made the same argument about avoiding the use of the inline keyword. The basic idea is that you should either trust that the compiler is smarter than you are about making this decision, or prove otherwise. This idea makes sense in principle. Generally I assume the compiler knows better than I do. But occasionally I feel pretty confident about some judgment, and there's a cost in proving myself right: it takes time.

In this case, I assume the compiler agrees with me about inlining these. It seems absurd to push an extra stack frame for each of these method calls, especially for the `log` and `logf` methods which do nothing except pass the exact same parameters to another function with the same signature. In the event that the compiler doesn't agree, I'm betting that it's wrong. But I was too lazy to actually investigate it. Maybe I'm being too bold, assuming the compiler is wrong if it disagrees with me. Maybe the compiler always knows best. I just haven't collected enough evidence about this to be 100% convinced of it.

Let's look at this from another angle. Sometimes a programmer is faced with a related decision. Should I write some code directly in function A, or should I extract that code out into function B and call function B from within function A. Assume it makes no difference from a code readability or organizational perspective. Following your line of reasoning, it would always be better to extract the code into a separate function, because it allows the compiler to decide whether to inline it, rather than forcibly inlining it by writing the code directly in function A. Do you agree with that?

src/trace/log.zig
dnut: d9dc8e71c73207e9d2d5ea03b3c8878a32be553a

src/trace/entry.zig
dnut: tagging @InKryption on this since we've had a similar discussion previously

src/trace/log.zig
dnut: @0xNineteen what do you think? Do we need this? How about we just provide enough memory so we don't need to worry about this?

src/trace/entry.zig
Rexicon226: >  It seems absurd to push an extra stack frame for each of these method calls, especially for the log and logf methods which do nothing except pass the parameters to another function with the same signature.

Irrelevant, this costs virtually nothing. It's not even a factor that should be considered. No stack frame would be created either way.

> In the event that the compiler doesn't agree, I'm betting that it's wrong. But I was too lazy to actually investigate it. Maybe I'm being too bold, assuming the compiler is wrong if it disagrees with me. Maybe the compiler always knows best. I just haven't collected enough evidence about this to be 100% convinced of it.

There is no "wrong" here, the compiler not inlining will cost us nothing. I am confident that the optimizer will make the correct choice whatever that may be. If you were to go and take extensive benchmarks across all common use cases and really find that adding `inline` were to make it faster, then it could make sense. 

> Following your line of reasoning, it would always be better to extract the code into a separate function, because it allows the compiler to decide whether to inline it, rather than forcibly inlining it by writing the code directly in function A. Do you agree with that?

When code is used multiple times throughout the codebase and this block of code has a clear intention by itself, such as in this case, `field`, which is a component with a clear function, I agree. 
I draw the line at things like getter and setter functions which *are* ridiculous and provide no benefit to just accessing the object directly.

src/trace/entry.zig
InKryption: There is actually a benefit to this being `inline`: if `value` is a comptime-known or comptime-only value (ie consider `.field(metric_name, INT_CONSTANT).log("Beginning work")`), the `inline` here will allow us to retain that comptime-knowness through the call, allowing the value to be directly embedded as type information in a comptime field, rather than taking up space on the stack, achieving the same properties as we would get if we did the log in a single call.

src/trace/entry.zig
InKryption: I think it's be fair to say that how this code works may not be obvious to someone who doesn't understand that, in zig, types can be computed through arbitrary logic just as any other values.
But I think understanding that facet of comptime is a fair prerequisite to reading sig, as long as we don't get much more complicated than this.

src/trace/entry.zig
Rexicon226: I don't see howargument makes sense. `new_fields` already cannot be a "comptime-only" struct given that it copies values from `self.fields`. The copy of `value` into `new_fields` will not benefit at all from the comptime-ness of `value`.

Secondly, we need to stop thinking in terms of "space on the stack." This is an irrelevant idea that should have no say in reasoning about inline functions. It's not "bad" to take up stack space. It's not "bad" to have extra stack frames. These things take orders of magnitude less time than half the logic surrounding them not to mention the fact that the optimizer will easily see over-allocated stack partitions and remove them.

src/trace/entry.zig
InKryption: My thoughts:

I think there is an argument to be made that, while there's no cost to not inlining the function, there's also no cost to inlining it either in this case. My reasoning: the reason it's not recommended to inline things willy nilly is because it limits what transformations the compiler is able to make to the code; but for this specific function, I am 99.99999% confident there is no transformation the optimizer could feasibly apply beyond simply inlining it (happy to be proven to otherwise). With that in mind, I think it's perfectly reasonable to specify the intent for this function to be inlined (again, specifically in this case, as soon as there's any sort of potential safety check, inner function calls, etc, I think it best to leave it to the optimizer).

I think a bigger issue here is that this `init` function is mostly superfluous, we could remove it by just initializing `Entry` directly. With that done we don't even really have to consider the implications or tradeoffs of inlining this function, because its purpose is already perfectly accomplished by a simpler construct.

Though if a designated initialization really is desired, I'd vote for just having a constant called `init`, which will benefit from the decl literal feature in 0.14.

src/trace/entry.zig
InKryption: > I don't see how this argument makes sense. new_fields already cannot be a "comptime-only" struct given that it copies values from self.fields. The copy of value into new_fields will not benefit at all from the comptime-ness of value.

I'm not saying it becomes comptime-only, I am saying it will gain comptime fields. This runtime struct can perfectly well contain a `comptime` field.

To the rest: I guess talking about taking up space on the stack was the wrong framing; my principle concern is about the information the compiler has when reasoning about the code. If we did `logger.infof(fields, fmt_str, args)`, the compiler would have available a greater amount of information available to it by virtue of `.{ .field = 120 }` being encoded into the type as `struct { comptime field: comptime_int = 120 }`. With this inlined, we retain that property despite our current choice of API (although I just noticed for this to work, we'd actually need to take this into account in the `FieldsPlus` logic as well @dnut).

Though we could also say we just don't care about this information being conveyed to the compiler, uninline it and call it a day, I'm just as well in favor of that.

src/trace/entry.zig
Rexicon226: > With that in mind, I think it's perfectly reasonable to specify the intent for this function to be inlined

Our intent is not to inline it. Inlining it provides no change as you've pointed out, and would be distracting and confusing to a reader. "Why was this inlined?" they may ask themselves. 

re init function, I totally agree with all points there, this shouldn't be an inline functions + it'll be nice to use decl literals when we have them.

src/trace/entry.zig
Rexicon226: > I'm not saying it becomes comptime-only, I am saying it will gain comptime fields. This runtime struct can perfectly well contain a comptime field.

As you pointed out, we don't make it one, nor should we. https://github.com/Syndica/sig/pull/307/files#diff-d4c236e79e7de5edd2999e5ecbe1cc72433b23964fa8c6e40df19926cad8bc0eR70.
I am strongly against using any type of tuple is_comptime hack here. It's unnecessary and is extremely ugly.

> the compiler would have available a greater amount of information available to it by virtue of .{ .field = 120 } being encoded into the type as struct { comptime field: comptime_int = 120 }

Given the line I linked above, this is an optimization that LLVM itself would need to make, and I'm certain it will. It also, again, does not matter if it does.

src/trace/entry.zig
dnut: f0c6b320cbe51686c31cf5e6ffffadd2325d52d7

src/trace/entry.zig
dnut: f0c6b320cbe51686c31cf5e6ffffadd2325d52d7

src/trace/entry.zig
dnut: Thanks for all the feedback. I could still see arguments on both sides, but I'm satisfied at this point with removing inline.

Regarding the init function, typically I'll provide an init function if it abstracts away any fields. In this case it abstracts the initialization of the `fields` fields. But it's trivial and only used in one spot, so I'm happy to remove it.

f0c6b320cbe51686c31cf5e6ffffadd2325d52d7

src/trace/log.zig
0xNineteen: this way provides a way to recover from no memory available which you could eventually recover from -- not sure how we could do it better -- if you want to provide enough memory so we don't need to worry about this, you can increasing the buffer size and panic on fail instead of retrying?

0xNineteen:
removing my review on this - seems like theres good discussion on this and plenty of reviews - trust your guys choice

src/trace/entry.zig
dnut: > we'd actually need to take this into account in the FieldsPlus logic as well @dnut).

Yeah, due to this, I don't think it's worth the complexity.

FieldsPlus would need a parameter to embed the default value into the type, and the `value: anytype` parameter of `field` would need to be conditionally included in the function's return type depending on whether its value is available comptime. Here's what I came up with:

```diff
diff --git a/src/trace/entry.zig b/src/trace/entry.zig
index c4b00ea..70c042c 100644
--- a/src/trace/entry.zig
+++ b/src/trace/entry.zig
@@ -23,13 +23,21 @@ pub fn Entry(comptime Fields: type, comptime scope: ?[]const u8) type {
             self: Self,
             comptime name: [:0]const u8,
             value: anytype,
-        ) Entry(FieldsPlus(name, @TypeOf(value)), scope) {
+        ) Entry(FieldsPlus(
+            name,
+            @TypeOf(value),
+            if (isComptime(@TypeOf(value))) value else null,
+        ), scope) {
             if (self.logger == .noop) return .{
                 .logger = .noop,
                 .level = undefined,
                 .fields = undefined,
             };
-            var new_fields: FieldsPlus(name, @TypeOf(value)) = undefined;
+            var new_fields: FieldsPlus(
+                name,
+                @TypeOf(value),
+                if (isComptime(@TypeOf(value))) value else null,
+            ) = undefined;
             inline for (@typeInfo(Fields).Struct.fields) |existing_field| {
                 @field(new_fields, existing_field.name) = @field(self.fields, existing_field.name);
             }
@@ -51,19 +59,25 @@ pub fn Entry(comptime Fields: type, comptime scope: ?[]const u8) type {
             self.logger.private_log(self.level, self.fields, fmt, args);
         }
 
+        fn isComptime(comptime T: type) bool {
+            return switch (@typeInfo(T)) {
+                .ComptimeInt, .ComptimeFloat => true,
+                else => false,
+            };
+        }
+
         /// Returns a new struct type based on Fields, just with one more field added.
-        fn FieldsPlus(comptime field_name: [:0]const u8, comptime FieldType: type) type {
+        fn FieldsPlus(
+            comptime field_name: [:0]const u8,
+            comptime FieldType: type,
+            comptime default_value: ?FieldType,
+        ) type {
             const info = @typeInfo(Fields);
-            const ActualFieldType = switch (@typeInfo(FieldType)) {
-                .ComptimeFloat => f64,
-                .ComptimeInt => u64,
-                else => FieldType,
-            };
             const new_fields = info.Struct.fields ++ &[_]std.builtin.Type.StructField{.{
                 .name = field_name,
-                .type = ActualFieldType,
-                .default_value = null,
-                .is_comptime = false,
+                .type = FieldType,
+                .default_value = if (default_value) |dv| @ptrCast(&dv) else null,
+                .is_comptime = isComptime(FieldType),
                 .alignment = @alignOf(FieldType),
             }};
             const new_struct = std.builtin.Type.Struct{
```

src/trace/log.zig
InKryption: > this way provides a way to recover from no memory available which you could eventually recover from

I believe this idea in and of itself is a reasonable goal, I think the main issue is that it's being done through the `std.mem.Allocator` interface, which does not provide any guarantee that this is the behavior it will display, and it's a fairly unconventional pattern as opposed to something more common such as "`arena`-like behavior".

A simple direct improvement imo would be to directly use `log_allocator_state` instead of going through the `log_allocator` field, encoding the intention for this behavior right in the code itself.

dnut: > feels like changes are outside the scope of this PR though

https://github.com/orgs/Syndica/projects/2/views/10?pane=issue&itemId=83096299

Rexicon226 (approved):
overall looks good to me

InKryption (approved):
lgtm

PR #308, opened 2024-10-09
fix(trace): incorrectly logging string fields

dnut:
This fixes a logic bug that I introduced in the logger pr. The bug is that string fields are formatted incorrectly. `[]const u8` ends up being formatted as `{any}` instead of `{s}`.

This branch of the if-else will actually never execute because it has a redundant condition with the first branch. It's just not supposed to have that condition. I think I copy and pasted it by mistake.

### DIFF
--- a/src/trace/logfmt.zig
+++ b/src/trace/logfmt.zig
@@ -32,7 +32,7 @@ fn fieldFmtString(comptime Value: type) []const u8 {
         // Assume arrays of u8 are strings.
         .Pointer => |ptr| if (ptr.size == .One)
             fieldFmtString(ptr.child)
-        else if (ptr.size == .One and ptr.child == u8)
+        else if (ptr.child == u8)
             "{s}={s} "
         else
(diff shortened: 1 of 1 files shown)


InKryption (approved):

src/trace/logfmt.zig
```diff
@@ -32,7 +32,7 @@ fn fieldFmtString(comptime Value: type) []const u8 {
         // Assume arrays of u8 are strings.
         .Pointer => |ptr| if (ptr.size == .One)
             fieldFmtString(ptr.child)
-        else if (ptr.size == .One and ptr.child == u8)
+        else if (ptr.child == u8)
```
dadepo, line 35: This is what I was asking about here https://github.com/Syndica/sig/commit/2869887ca94a3ae61e75ed4b2189f12bd8324277#r147741038 having the `ptr.size == .One` in both if and the else looked off.

Nice it got fixed.

PR #323, opened 2024-10-21
refactor(docs): Update, rephrase & improve CONTRIBUTING.md

InKryption:
This codifies the conclusions reached in #304, as well as those reached in the discussions in and originating from #315.

The main additions are `Identifiers`/`Interfaces & Interface State`, and `Identifiers`/`Constants`.

While adding these, I also took the liberty of making some details more consistent, like removing the newline between a header and the first paragraph, rewording some other sections to be more accurate, and removing some unnecessary detail.

I also split up some short paragraphs into separate lines which were a single continuous line before, which made it easier to visualize all the information in a single screen.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -1,31 +1,44 @@
 # Contributing to Sig
-
-Thank you for considering contributing to Syndica's Sig project! We appreciate your interest and support in helping us make this project better. By participating in this project, you are joining a community of developers and contributors working together to create value for the Solana ecosystem.
-
+Thank you for considering contributing to Syndica's Sig project! We appreciate your interest and support in helping us make this project better.
+By participating in this project, you are joining a community of developers and contributors working together to create value for the Solana ecosystem.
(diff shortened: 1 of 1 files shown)


0xNineteen (approved):

Sobeston (approved):

sonicfromnewyoke:
WDYT, does [getting rid of `const Self = @This()` for simple structs](https://zig.news/kristoff/dont-self-simple-structs-fj8) (e.g. small structs in the `src>ledger>shred.zig`)  make the code more readable?

InKryption:
> WDYT, does [getting rid of `const Self = @This()` for simple structs](https://zig.news/kristoff/dont-self-simple-structs-fj8) (e.g. small structs in the `src>ledger>shred.zig`) make the code more readable?

In my own personal opinion, I prefer avoiding `const Self = @This()`, however the compromise we reached previously was to continue to use it broadly in the codebase.

Though I have noticed a recent trend where we've naturally tended to forget to follow that convention, so maybe we should revisit the topic at some point, and either refactor to more consistently use `const Self = This()`, or instead eschew it where unnecessary.

PR #325, opened 2024-10-22
perf(accountsdb): improve benchmarking

Sobeston:
- makes AccountsCache configurable
- introduces pseudo-random  weighted sampling of accounts using the [Alias Method](https://en.wikipedia.org/wiki/Alias_method) (before this code accessed every account once in order).
- adds a small optimisation to AccountsCache.CachedAccount

### DIFF
--- a/src/accountsdb/cache.zig
+++ b/src/accountsdb/cache.zig
@@ -13,6 +13,8 @@ const Slot = sig.core.Slot;
 pub const AccountsCache = struct {
     lru: LRU,
     allocator: std.mem.Allocator,
+    cache_hits: usize = 0,
+    cache_misses: usize = 0,
 
     /// Atomically refcounted account
     pub const CachedAccount = struct {
@@ -83,10 +85,15 @@ pub const AccountsCache = struct {
 
     /// User is expected to release the returned CachedAccount.
     pub fn get(self: *Self, pubkey: Pubkey, slot: Slot) ?*CachedAccount {
-        const account: *CachedAccount = self.lru.get(pubkey) orelse return null;
+        const account: *CachedAccount = self.lru.get(pubkey) orelse {
+            self.cache_misses += 1;
+            return null;
+        };
         if (account.slot == slot) {
+            self.cache_hits += 1;
             return account.copyRef();
         } else {
+            self.cache_misses += 1;
             return null;
         }
     }
--- a/src/accountsdb/cache.zig
+++ b/src/accountsdb/cache.zig
@@ -114,6 +114,7 @@ pub const AccountsCache = struct {
             if (existing_entry.slot < slot) {
                 existing_entry.releaseOrDestroy(self.allocator);
 
+                // TODO: this could be turned into one allocation
                 const new_entry = try self.allocator.create(CachedAccount);
                 new_entry.* = try CachedAccount.init(self.allocator, account, slot);
                 _ = self.lru.put(pubkey, new_entry);
--- a/src/accountsdb/cache.zig
+++ b/src/accountsdb/cache.zig
@@ -8,13 +8,13 @@ const Pubkey = sig.core.pubkey.Pubkey;
 const LruCacheCustom = sig.common.lru.LruCacheCustom;
 const ReferenceCounter = sig.sync.reference_counter.ReferenceCounter;
 const Slot = sig.core.Slot;
+const Counter = sig.prometheus.counter.Counter;
 
 /// Stores read-only in-memory copies of commonly used *rooted* accounts
 pub const AccountsCache = struct {
     lru: LRU,
     allocator: std.mem.Allocator,
-    cache_hits: usize = 0,
-    cache_misses: usize = 0,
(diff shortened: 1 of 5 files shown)


src/accountsdb/db.zig
```diff
@@ -72,7 +72,7 @@ pub const AccountsDB = struct {
     unrooted_accounts: RwMux(SlotPubkeyAccounts),
 
     // pubkey->account LRU maps
-    accounts_cache: RwMux(AccountsCache),
+    maybe_accounts_cache_mux: ?RwMux(AccountsCache),
```
0xNineteen, line 75: the convention weve been following elsewhere is `_rw` for `RwMux`

src/accountsdb/db.zig
```diff
@@ -4251,15 +4307,129 @@ pub const BenchmarkAccountsDB = struct {
...
+        }
+
         return elapsed;
     }
 };
+
+/// Implementation of Alastair J. Walker's "Alias method".
+/// Once constructed, this allows for efficiently getting pseudorandom indeces
+/// that fit the given probabilities.
+const WeightedIndexer = struct {
```
0xNineteen, line 4373: should probs be in src/rand/rand.zig

src/accountsdb/cache.zig
```diff
@@ -13,6 +13,8 @@ const Slot = sig.core.Slot;
 pub const AccountsCache = struct {
     lru: LRU,
     allocator: std.mem.Allocator,
+    cache_hits: usize = 0,
+    cache_misses: usize = 0,
```
0xNineteen, line 17: would probs be best to make these use prometheus counters

0xNineteen (changes requested):
few small nits - benchmark results on CI show cache with lru < 100_000 are slower than no LRU -- any reason why you might think this is?

src/accountsdb/cache.zig
Sobeston: https://github.com/Syndica/sig/pull/325/commits/85ec8c3441692383fed14ff3f66f0c71e1cd35df#diff-7f62328c8d0fd51e206b91ff2b72aef34de0fae176ba46106a0c768d29f306f8R14-R18

src/accountsdb/db.zig
Sobeston: https://github.com/Syndica/sig/pull/325/commits/85ec8c3441692383fed14ff3f66f0c71e1cd35df#diff-ad5c9043014490f7d5452d02439dcbb3bf5eb93306be2a8b12ba05f7ddad5e3aR429-R438

src/accountsdb/db.zig
0xNineteen: few places where _mux is still used, eg:  `const accounts_cache_mux = &(self.maybe_accounts_cache_rw orelse break :blk null);`

src/rand/rand.zig
```diff
@@ -426,3 +426,75 @@ const rust_expected_weights: [1000]u64 = .{
...
     81032160488073578,  173298561206951917, 154421360720272045, 15981567533895039,  88393403508710277,
 };
+
+/// Implementation of Alastair J. Walker's "Alias method".
+/// Once constructed, this allows for efficiently getting pseudorandom indeces
+/// that fit the given probabilities.
+pub const WeightedAliasSampler = struct {
+    probability: []const f32,
+    alias: []const usize,
+    pub fn init(allocator: std.mem.Allocator, probabilities: []const f32) !WeightedAliasSampler {
```
dnut, line 436: The sampler will silently misbehave if the probabilities do not add up to 1.

Either of these changes will make the sampler safer:
1. return error if they do not sum to 1
2. treat them as weights instead of probabilities

Personally I prefer 2 because it's more flexible, won't result in runtime errors, and it makes the api more consistent with the other sampler. What do you think?

src/accountsdb/db.zig
Sobeston: https://github.com/Syndica/sig/pull/325/commits/cf31bb0ad2fed66f340c14b84582ab47fcb50f3f

src/rand/rand.zig
```diff
@@ -426,3 +426,75 @@ const rust_expected_weights: [1000]u64 = .{
...
+            .alias = alias,
+        };
+    }
+
+    pub fn deinit(self: WeightedAliasSampler, allocator: std.mem.Allocator) void {
+        allocator.free(self.probability);
+        allocator.free(self.alias);
+    }
+
+    pub fn sample(self: WeightedAliasSampler, random: std.Random) usize {
```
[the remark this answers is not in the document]
Sobeston: https://github.com/Syndica/sig/pull/325/commits/b93bc85de9a922ccfde8cc5f14049b513bd9cea4

src/rand/rand.zig
Sobeston: https://github.com/Syndica/sig/pull/325/commits/8adc271af982bddc22bacde3020a617314895ae5

0xNineteen (dismissed):

dnut (approved):

PR #329, opened 2024-10-23
refactor(ledger): organize into folders

dnut:
This PR only changes the organization of files in the ledger, and likewise the imports. I moved some files from the ledger folder into two new subfolders: `database` and `shred_inserter`.

I've been finding it hard to make sense of the number of files in the ledger. I would look at the list of files and struggle to answer questions like, "Which files are for inserting shreds?" This is not a good sign, since I'm the one who wrote it. So I think it's appropriate to add a bit of structure to show how some of these apparently disparate files are actually intimately linked.

For example, there is one function `insertShreds` which is the sole entrypoint for all the code within 5 files in the ledger. Those files were scattered throughout the ledger folder and it wasn't totally clear what their purpose was. Their sole purpose is to be a part of inserting shreds. So in my mind it makes sense to group them all together under a `shred_inserter` umbrella. One way is by putting them all in a single file, but I think that makes the situation worse because the file becomes very large, which makes it hard to discern any organization within it. The other alternative is grouping it together in a folder, which preserves the distinction between the files while also showing that they are related.

The other files I combined into a folder are the database implementations.

### DIFF
--- a/src/ledger/schema.zig
+++ b/src/ledger/schema.zig
@@ -1,10 +1,11 @@
 const std = @import("std");
 const sig = @import("../sig.zig");
+const ledger = @import("lib.zig");
 
-const meta = sig.ledger.meta;
+const meta = ledger.meta;
 
-const ColumnFamily = sig.ledger.database.ColumnFamily;
-const ErasureSetId = sig.ledger.shred.ErasureSetId;
+const ColumnFamily = ledger.database.ColumnFamily;
+const ErasureSetId = ledger.shred.ErasureSetId;
 const Pubkey = sig.core.Pubkey;
 const Signature = sig.core.Signature;
 const Slot = sig.core.Slot;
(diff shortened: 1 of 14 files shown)


src/ledger/schema.zig
```diff
@@ -63,7 +64,6 @@ pub const schema = struct {
     pub const address_signatures: ColumnFamily = .{
         .name = "address_signatures",
         .Key = struct {
-            // NOTE: rn we sort by pubkey first, maybe we want to sort by slot first?
```
dnut, line 66: This column family is used when querying for all signatures from a particular address over a particular timespan. It's a lot more efficient to filter first by address. If it's sorted by slot first, it means the vast majority of items you iterate over will be discarded.

0xNineteen (approved):

PR #330, opened 2024-10-23
improve(cli): return `MissingPort` error if no port is specified in the IPv4 address passed to `gossip-host` 

dadepo:
The compile error now shows as

```
time=2024-10-23T13:40:20.873Z level=info entrypoints: { 35.203.170.30:8001, 139.178.94.143:8001, 139.178.94.143:8001 }
time=2024-10-23T13:40:21.727Z level=info shred version: 27799 - from entrypoint ip echo: 35.203.170.30:8001
error: MissingPort
```

Which is a bit more informative than just `InvalidIp` it currently returns

### DIFF
--- a/src/net/net.zig
+++ b/src/net/net.zig
@@ -91,9 +91,7 @@ pub const SocketAddr = union(enum(u8)) {
             }
         }
 
-        if (!colon_found) {
-            return error.MissingPort;
-        }
+        if (!colon_found) return error.MissingPort;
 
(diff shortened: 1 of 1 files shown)


dadepo:
Given the comment here https://github.com/Syndica/sig/pull/331#issuecomment-2435124429 the better approach is to remove the requirement of a port when using the `--gossip-host` flag.

I'll close this for now and create another PR that does just that.

PR #331, opened 2024-10-24
improve(cli, gossip): get gossip port from host if given, fallback to gossip port

dadepo:
The gossip cli has these two options:

```
 --gossip-host <Gossip Host>           
  -p, --gossip-port <Gossip Port>      
```

`--gossip-host` requires that the port is set. 

This begs the question if `--gossip-host` is passed and also `--gossip-port`, which value should be used as the port? 

Current behaviour is that the port passed with the host is always ignored, and `--gossip-port` is used if set, if not it defaults to `8001`.

Given the port passed with the host, IMO should be more authoritative, this PR tries to use that first, if not then it defaults to 
 `--gossip-port`, and if that is not set, defaults to `8001`.

### DIFF
--- a/src/cmd/cmd.zig
+++ b/src/cmd/cmd.zig
@@ -1107,6 +1107,7 @@ const AppBase = struct {
     entrypoints: std.ArrayList(SocketAddr),
     shred_version: u16,
     my_ip: IpAddr,
+    my_port: u16,
 
     fn init(allocator: Allocator) !AppBase {
         const logger = try spawnLogger();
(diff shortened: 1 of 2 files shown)


src/cmd/cmd.zig
```diff
@@ -1123,6 +1124,8 @@ const AppBase = struct {
         errdefer entrypoints.deinit();
 
         const ip_echo_data = try getMyDataFromIpEcho(logger, entrypoints.items);
+        // Try to extract the port from the given host. If no port is found, fall back to the configured default port.
+        const my_port = (config.current.gossip.getPortFromHost() orelse config.current.gossip.port) catch config.current.gossip.port;
```
dadepo, line 1128: Another option I considered was to error out if `--gossip-host` is set (which must have a port portion) and if `--gossip-port`.

Yet another option is to drop the requirement that `--gossip-host` require the port portion.

Out of all these, I felt this fallback was the least intrusive approach.

0xNineteen (approved):

dnut:
`--gossip-host` should strictly be for *only* the IP address. It was a bug to expect a port in there. We should remove that functionality because it's confusing, overly complicated, and contradicts the name of the cli option

dadepo:
> `--gossip-host` should strictly be for _only_ the IP address. It was a bug to expect a port in there. We should remove that functionality because it's confusing, overly complicated, and contradicts the name of the cli option

I agree. I will work on a follow up PR to do this.

PR #332, opened 2024-10-24
refactor(ledger): flatten shred field nesting

### LINKED ISSUE #212: Shred struct redesign
Consider a new way to organize data for the Shred/*Shred/GenericShred paradigm, with this discussion in mind: https://github.com/Syndica/sig/pull/194#discussion_r1696015339

@0xNineteen said:
> out of scope comment:
> 
>     fields: Fields,
>     const Fields = GenericShred(DataShredHeader, data_shred);
> 
> looking back at some prev code - the GenericShred approach seems a bit confusing - it would probs be a lot easier to understand the way agave did it [here](https://github.com/anza-xyz/agave/blob/b0df3945e8c9a2f85093fb720d349b536040a8d4/ledger/src/shred/legacy.rs#L36) by putting the common_fields onto the struct itself
> 
> the fromPayload method can probably be done with a custom bincode deserialization method
> 
> data_shred -> DATA_SHRED_CONSTANTS would be more descriptive, also since its a global constant

@dnut said:

> 
> > looking back at some prev code - the GenericShred approach seems a bit confusing - it would probs be a lot easier to understand the way agave did it [here](https://github.com/anza-xyz/agave/blob/b0df3945e8c9a2f85093fb720d349b536040a8d4/ledger/src/shred/legacy.rs#L36) by putting the common_fields onto the struct itself
> 
> I don't like that there are so many layers either. I've been trying to come up with a better alternative but haven't come up with anything yet. In my opinion, the agave approach is even more confusing.
> 
> The issue with putting the fields directly in CodingShred and DataShred is that you need to copy identical duplicates for all the generic methods into both of those structs.
> 
> In agave, they accomplished this with macros, like impl_shred_common and impl_merkle_shred. Personally I find it easier to make sense of a nested struct rather than wading through macro definitions.
> 
> The other approach is just having duplicate code. One disadvantage to this is that you lose the idea that these methods apply to all shreds. Another thing to keep in mind is that we are going to need a lot more of them. The insert shreds pr doubles the number of generic methods on shreds. Maintaining duplicates of each is extra work and will almost inevitably lead to accidental divergence at some point.
> 
> I can spend some time looking for an approach that flattens some nesting without resorting to confusing metaprogramming or excessive code duplication. Some repetition may be inevitable, but I'd like to avoid duplicating every single detail. I think it will make more sense to do this on top of the changes in the insert shreds pr, first of all because GenericShred isn't introduced by the current pr, and also the insert shreds pr has more concrete methods that are very relevant to this. I'd like to ensure any change to the structure here is compatible with all of those methods, and avoid too many complicated merges.

dnut:
closes #212

- Moved shred fields out of GenericShred struct into each respective shred struct: DataShred and CodeShred.
- Anywhere that fields from `fields` were being accessed, switched it to get the fields directly from the shred struct. so `shred.fields.common` is now `shreds.common`
- For any usages of methods in GenericShred via `fields` outside shred.zig, analogous methods were added to DataShred and CodeShred to serve the same purpose. this is in service to the point above, so things like `shred.fields.merkleRoot()` are now `shred.merkleRoot()`
- All the methods from GenericShred are now just functions in a private namespace called `generic_shred`. other shred structs that used to call `self.fields.function()` now call `generic.function(self)`
- Renamed `default` initializers in the shred structs to `zeroedForTest` since the output data is actually an invalid shred and should probably not be used for anything. it's barely distinguishable from initializing the data as undefined. In the future i think these methods should be replaced with a way to initialize well-formed shreds. For now, we need it for tests, so I'd like it to be named in a way that indicates clearly what it does and what it's for, rather than implying it is a well-formed "default."

### DIFF
--- a/src/ledger/meta.zig
+++ b/src/ledger/meta.zig
@@ -187,13 +187,13 @@ pub const ErasureMeta = struct {
 
     pub fn fromCodeShred(shred: CodeShred) ?Self {
         return .{
-            .fec_set_index = shred.fields.common.fec_set_index,
+            .fec_set_index = shred.common.fec_set_index,
             .config = ErasureConfig{
-                .num_data = shred.fields.custom.num_data_shreds,
-                .num_code = shred.fields.custom.num_code_shreds,
+                .num_data = shred.custom.num_data_shreds,
+                .num_code = shred.custom.num_code_shreds,
             },
             .first_code_index = shred.firstCodeIndex() catch return null,
-            .first_received_code_index = shred.fields.common.index,
+            .first_received_code_index = shred.common.index,
         };
     }
 
@@ -358,9 +358,9 @@ pub const MerkleRootMeta = struct {
             // `None` for those cases in blockstore, as a later
             // shred that contains a proper merkle root would constitute
             // a valid duplicate shred proof.
-            .merkle_root = shred.fields.merkleRoot() catch null,
-            .first_received_shred_index = shred.fields.common.index,
-            .first_received_shred_type = shred.fields.common.variant.shred_type,
+            .merkle_root = shred.merkleRoot() catch null,
+            .first_received_shred_index = shred.common.index,
+            .first_received_shred_type = shred.common.variant.shred_type,
         };
     }
 };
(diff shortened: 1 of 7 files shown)


0xNineteen (approved):

0xNineteen:
lgtm - is a large improvement, much cleaner and easier to follow

PR #337, opened 2024-10-29
feat(RecycleFBA): Add metrics

dadepo:
The goal of this PR is to add the ability to collect metrics from the RecycleFBA.

TODO
- [ ] Clarify on the metrics we want to collect.
- [ ] How to label the metircs? Given there would be many instance of recyle fba
- [ ] Ensure metrics collection does not affect the performance of allocating.
- [ ] Add metrics and confirm it works on ec2.

### DIFF
--- a/src/utils/allocators.zig
+++ b/src/utils/allocators.zig
@@ -1,6 +1,8 @@
 const std = @import("std");
 const builtin = @import("builtin");
 const sig = @import("../sig.zig");
+const Counter = sig.prometheus.counter.Counter;
+const Gauge = sig.prometheus.gauge.Gauge;
 
 pub fn RecycleFBA(config: struct {
     /// If enabled, all operations will require an exclusive lock.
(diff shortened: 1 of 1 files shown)


src/utils/allocators.zig
```diff
@@ -204,6 +210,17 @@ pub fn RecycleFBA(config: struct {
     };
 }
 
+pub const RecycleFBAMetrics = struct {
```
dadepo, line 213: @0xNineteen Your ideas on metrics that should be collected here will be appreciated.

src/utils/allocators.zig
```diff
@@ -16,13 +18,16 @@ pub fn RecycleFBA(config: struct {
...
+        // for tracking metrics.
+        maybe_metrics: ?RecycleFBAMetrics = null,
 
         const Record = struct { is_free: bool, buf: [*]u8, len: u64 };
         const AllocatorConfig = struct {
             // used for the records array
             records_allocator: std.mem.Allocator,
             // used for the underlying memory for the allocations
             bytes_allocator: std.mem.Allocator,
+            maybe_metrics: ?RecycleFBAMetrics = null,
```
0xNineteen, line 30: this should take an optional registry which is used to init the metrics (see how the account-cache in accountsdb works)

src/utils/allocators.zig
0xNineteen: i think thats a great starting point -- we could also add a historgram of the sizes that are free

0xNineteen:
gonna close this for now since it seems dead - can reopen if we want to continue tackling it

PR #341, opened 2024-10-30
fix(rpc): return types

0xNineteen:
- log the RPC client version on `init`
- improve getClusterNodes return type to have all fields (renamed to match anza `RpcContactInfo`)
- add `getVersion` rpc call 
- fix `getAccountInfo` value to be optional
- add error type for `getSignatureStatuses` to fix parsing error

### DIFF
--- a/src/rpc/client.zig
+++ b/src/rpc/client.zig
@@ -32,16 +32,30 @@ pub const Client = struct {
             .LocalHost => "http://localhost:8899",
             .Custom => |cluster| cluster.url,
         };
-        return .{
+
+        var self: Client = .{
             .http_endpoint = http_endpoint,
             .http_client = std.http.Client{ .allocator = allocator },
             .max_retries = options.max_retries,
             .logger = options.logger,
         };
+
+        self.logVersion(allocator) catch |err| {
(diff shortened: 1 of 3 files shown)


src/rpc/client.zig
```diff
@@ -282,6 +301,9 @@ pub const Client = struct {
             break;
         }
 
+        // sometimes the response is "x-y" where our fields follow "x_y" format, so this
+        // replaces "-" with "_" so we can properly parse the response
+        _ = std.mem.replace(u8, response.bytes.items, "-", "_", response.bytes.items);
```
yewman, line 306: This seems potentially risky in a general request method, is there an issue with a field name in the RPC data structure which uses hyphens? I thought they were camel case. Anyhow I think it would make more sense to use the same field name as the Rpc response using the @"<name>" in this case. Let me know what you think

yewman (changes requested):
Just one issue with field naming otherwise lgtm

0xNineteen:
merging into [19/tx-forwarding-cicd](https://github.com/Syndica/sig/tree/19/tx-forwarding-cicd) bc it has specific RPC methods implemented we want to support

yewman (approved):
lgtm!

PR #345, opened 2024-10-31
fix(trace): use scope in logger 

dadepo:
This PR adds scope to logger in appropriate places. 

The kinds of scoping that was added are:

- Logger is scoped to struct if the logger is part of the structs states.
   -  Exception to this is anonymous struct in GossipMetrics which is scoped to GossipService.
- Free standing function are scoped to the function name using `@src().fn_name`
- runService is defined to receive the logger as `anytype` this is because, without that, scoping to `@src().fn_name` will produce this output:

```
[runService__anon_51517] time=2024-11-05T06:16:33.852Z level=info Starting gossip readSocket
[runService__anon_51575] time=2024-11-05T06:16:33.852Z level=info Starting gossip verifyPackets
[runService__anon_52325] time=2024-11-05T06:16:33.853Z level=info Starting gossip processMessages
[runService__anon_53622] time=2024-11-05T06:16:33.853Z level=info Starting gossip buildMessages
[runService__anon_53953] time=2024-11-05T06:16:33.853Z level=info Starting gossip sendSocket
``` 
using anytype, the output becomes

```
[gossip readSocket] time=2024-11-05T06:50:36.060Z level=info Starting gossip readSocket
[gossip verifyPackets] time=2024-11-05T06:50:36.061Z level=info Starting gossip verifyPackets
[gossip processMessages] time=2024-11-05T06:50:36.061Z level=info Starting gossip processMessages
[gossip sendSocket] time=2024-11-05T06:50:36.061Z level=info Starting gossip sendSocket
[gossip buildMessages] time=2024-11-05T06:50:36.061Z level=info Starting gossip buildMessages
```
- printTimeEstimate is defined to receive the logger as `anytype`. This is to allow the scope to be set by the caller (instead of it being set within the function)

TODO
- [x] Cover the remaining places
- [x] Confirm all the cli services works fine.


src/accountsdb/db.zig
```diff
@@ -122,7 +123,7 @@ pub const AccountsDB = struct {
     bank_hash_stats: RwMux(BankHashStatsMap) = RwMux(BankHashStatsMap).init(.{}),
 
     metrics: AccountsDBMetrics,
-    logger: Logger,
+    logger: ScopedLogger(@typeName(@This())),
```
0xNineteen, line 126: can you change this to self

0xNineteen (commented):
a lot of this stuff isnt really scoped well - eg, gossip has output like:
```
[gossip.service.GossipService] time=2024-11-01T19:20:08.072Z level=debug from_endpoint=139.178.94.143:8001 from_pubkey=5D1fNXzvv5NjV1ysLjirC4WY92RNsVH18vjmcszZd8on gossip: recv ping
time=2024-11-01T19:20:18.085Z level=info gossip: recv 4494: 0 ping, 1 pong, 0 push, 0 pull request, 4493 pull response, 0 prune
```

accountsdb has things like 
```
time=2024-11-01T19:44:57.135Z level=info n_threads_snapshot_load: 12
time=2024-11-01T19:44:57.135Z level=info using ram memory for account index
[accountsdb.db.AccountsDB] time=2024-11-01T19:44:57.297Z level=info loading from snapshot...
[accountsdb.db.AccountsDB] time=2024-11-01T19:44:57.297Z level=info found 405721 account files
```

im thinking the type approach might be safer but restricts the usage and implementation and isnt very easy to use

dadepo:
> a lot of this stuff isnt really scoped well - eg, gossip has output like:
> 
> ```
> [gossip.service.GossipService] time=2024-11-01T19:20:08.072Z level=debug from_endpoint=139.178.94.143:8001 from_pubkey=5D1fNXzvv5NjV1ysLjirC4WY92RNsVH18vjmcszZd8on gossip: recv ping
> time=2024-11-01T19:20:18.085Z level=info gossip: recv 4494: 0 ping, 1 pong, 0 push, 0 pull request, 4493 pull response, 0 prune
> ```
> 
> accountsdb has things like
> 
> ```
> time=2024-11-01T19:44:57.135Z level=info n_threads_snapshot_load: 12
> time=2024-11-01T19:44:57.135Z level=info using ram memory for account index
> [accountsdb.db.AccountsDB] time=2024-11-01T19:44:57.297Z level=info loading from snapshot...
> [accountsdb.db.AccountsDB] time=2024-11-01T19:44:57.297Z level=info found 405721 account files
> ```

Yeah, this is due to the approach I took to only scope the loggers that are on the struct state. As for those it is clear what the scope is. Some of the places you mentioned are either from functions without state, or init functions. I'll go over them again to add the scope to places it makes sense.

> as an aside, im thinking the type approach might be safer but restricts the usage and isnt very easy to use/change -- wdyt about changing it to be just `ScopedLogger` (with no type saftey) -- would also save us from having `.unscoped()` everywhere for general functions

I did notice having to call `.unscoped()` was a bit of a chore, but one can argue that is a reasonable price to pay for the type safety which prevents adding a wrong scope to the log - which might be misleading. Having said that, I have no strong objection to dropping it as you suggested, but I think this is a decision that would probably need to be discussed.

src/accountsdb/db.zig
dadepo: 4d4ed66ea7e9138e47053e83a1ab6cbd2f515fcc

dadepo:
> a lot of this stuff isnt really scoped well - eg, gossip has output like:
> 
> ```
> [gossip.service.GossipService] time=2024-11-01T19:20:08.072Z level=debug from_endpoint=139.178.94.143:8001 from_pubkey=5D1fNXzvv5NjV1ysLjirC4WY92RNsVH18vjmcszZd8on gossip: recv ping
> time=2024-11-01T19:20:18.085Z level=info gossip: recv 4494: 0 ping, 1 pong, 0 push, 0 pull request, 4493 pull response, 0 prune
> ```

This now 

```
[gossip.service.GossipService] time=2024-11-04T12:58:35.227Z level=debug from_endpoint=109.236.92.86:8000 from_pubkey=ADiBUmi7zhg7fS5iy595iaNJvCXiZKjDrYUqrNHud2n7 gossip: recv ping
[gossip.service.GossipService] time=2024-11-04T12:58:37.171Z level=info gossip: recv 22135: 4 ping, 1 pong, 0 push, 0 pull request, 21602 pull response, 528 prune
[gossip.service.GossipService] time=2024-11-04T12:58:37.171Z level=info gossip: sent: 0 ping, 4 pong, 2317 push, 12 pull request, 0 pull response, 0 prune
[gossip.service.GossipService] time=2024-11-04T12:58:48.779Z level=info gossip: recv 12878: 0 ping, 20 pong, 0 push, 0 pull request, 12860 pull response, 0 prune
[gossip.service.GossipService] time=2024-11-04T12:58:48.779Z level=info gossip: sent: 21 ping, 0 pong, 4196 push, 12 pull request, 0 pull response, 0 prune
```

> accountsdb has things like
> 
> ```
> time=2024-11-01T19:44:57.135Z level=info n_threads_snapshot_load: 12
> time=2024-11-01T19:44:57.135Z level=info using ram memory for account index
> [accountsdb.db.AccountsDB] time=2024-11-01T19:44:57.297Z level=info loading from snapshot...
> [accountsdb.db.AccountsDB] time=2024-11-01T19:44:57.297Z level=info found 405721 account files
> ```

This is now

```
[cmd.cmd] time=2024-11-04T13:00:19.477Z level=info n_threads_snapshot_load: 8
[accountsdb.index.AccountIndex] time=2024-11-04T13:00:19.477Z level=info using ram memory for account index
[accountsdb.db.AccountsDB] time=2024-11-04T13:00:19.525Z level=info loading from snapshot...
[accountsdb.db.AccountsDB] time=2024-11-04T13:00:19.525Z level=info found 403736 account files
[accountsdb.db.AccountsDB] time=2024-11-04T13:00:19.533Z level=info [8 threads]: reading and indexing accounts...
```

dnut:
> the type approach might be safer but restricts the usage and isnt very easy to use/change

My view is that the only imposed restriction is that we're forced to do things that we already want to do. It only annoys you to let you know that you made a mistake.

> would also save us from having .unscoped() everywhere for general functions

The only time you need to call `unscoped` is when you do actually want to ensure that the next scope does not inherit the current scope. Regardless of whether scope is integrated into the type, I would still advocate for explicitly calling an `unscoped` method that wipes out the scope before passing the logger to other scopes.

0xNineteen:
> My view is that the only imposed restriction is that we're forced to do things that we already want to do.

not really, one thing we do often is print progress results (using `printTimeEstimate`) which should have an associated scope - but thats not possible because you need to use `.unscoped()` since its a general method which accepts a Logger and is used across the codebase

dadepo:
> @dnut - not really, one thing we do often is print progress results (using `printTimeEstimate`) which should have an associated scope - but thats not possible because you need to use `.unscoped()` since its a general method which accepts a Logger and is used across the codebase

@0xNineteen @dnut for this (and also for `runService`) I tried the approach of having the logger be received as `anytype`. This allows for the scope to be assigned by the caller, instead of from within the function. Do let me know your thoughts with this approach.

dnut:
The case where you want to inherit scope from another context is something I had not considered. But it seems unusual. Why wouldn't you want the scope to indicate where the actual log message comes from?

The only case I can think of is if the code calling the logger is actually intended to be an extension to the logger itself. It's like you're extending the logger interface through composition.

I still think it would be preferable for the default behavior to be each scope having its own distinct logging scope, and inheriting scope should be explicit.

src/ledger/database/rocksdb.zig
```diff
@@ -331,7 +332,7 @@ pub fn RocksDB(comptime column_families: []const ColumnFamily) type {
     };
 }
 
-fn callRocks(logger: Logger, comptime func: anytype, args: anytype) ReturnType(@TypeOf(func)) {
+fn callRocks(logger: anytype, comptime func: anytype, args: anytype) ReturnType(@TypeOf(func)) {
```
dnut, line 335: This should use the same scope as RocksDB.

src/ledger/cleanup_service.zig
```diff
@@ -31,19 +31,20 @@ const DEFAULT_CLEANUP_SLOT_INTERVAL: u64 = 512;
...
     exit: *AtomicBool,
 ) !void {
+    const logger = logger_.withScope(@src().fn_name);
     var last_purge_slot: Slot = 0;
 
     logger.info().log("Starting blockstore cleanup service");
     while (!exit.load(.acquire)) {
         last_purge_slot = try cleanBlockstore(
-            logger,
+            logger.unscoped(),
```
dnut, line 47: This file can use a single scope.

src/shred_collector/service.zig
```diff
@@ -38,7 +38,7 @@ pub const ShredCollectorConfig = struct {
 /// Resources that are required for the Shred Collector to operate.
 pub const ShredCollectorDependencies = struct {
     allocator: Allocator,
-    logger: Logger,
+    logger: ScopedLogger(@typeName(@This())),
```
dnut, line 41: This should not be a scoped logger

src/utils/service.zig
```diff
@@ -186,7 +187,7 @@ pub const ReturnHandler = struct {
 /// It's guaranteed to run at least once in order to not race initialization with
 /// the `exit` flag.
 pub fn runService(
-    logger: Logger,
+    logger: anytype,
```
dnut, line 190: Every log message already specifies the name of the service, so this function could just use ServiceManager as its scope.

If you think the service name should really be part of the scope, that's fine, but then I would make it something like "ServiceManager.\<service name\>" and also remove the service name from each of the log messages. Also you could accept Logger as the input here, and add the scope within this function.

dnut (commented):
This is great to see. The logs are so much better with scopes.

I might just make a few tweaks here and there. I think some of the scopes can be broader, which would simplify things and reduce the amount of `unscoped` `withScope` and `anytype` usages.

src/accountsdb/db.zig
```diff
@@ -703,11 +704,12 @@ pub const AccountsDB = struct {
     /// combines multiple thread indexes into the given index.
     /// each bin is also sorted by pubkey.
     pub fn combineThreadIndexesMultiThread(
-        logger: Logger,
+        logger_: Logger,
         index: *AccountIndex,
         thread_dbs: []const AccountsDB,
         task: sig.utils.thread.TaskParams,
     ) !void {
+        const logger = logger_.withScope(@src().fn_name);
```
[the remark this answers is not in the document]
0xNineteen: imo i could see the pros of knowing what method is currently being executed which you can ctrl+f find in the codebase and know exactly where its executing -- if its not too much work the ideal would be something like `AccountsDB.combineThreadIndexesMultiThread`

0xNineteen (approved):
overall lgtm

src/accountsdb/db.zig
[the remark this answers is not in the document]
dnut: Where do you draw the line? Would you rescope the logger for every function?

The information comes with a tradeoff of visual noise and code complexity. The costs are not too high, but neither is the benefit since you can also search for the log message itself.

In terms of understanding the logs on their own, I feel the file/struct/component name is much more useful than the function name, but I could see the function name providing valuable depth in some situations.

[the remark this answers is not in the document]
dnut: There could be a concept of nesting scope. for example:
```zig
const global_logger = logger.withScope("global");
const local_logger = global_logger.addScope("local");
```
Using local_logger would result in a scope like "global.local".

If you want to rescope the logger at a very granular level, you could add some functions like `infoScoped` that take a scope parameter.

0xNineteen:
something i noticed yew do is use this format:
```
time=2024-11-04T12:58:48.779Z level=info (gossip.service.GossipService) sent: 21 ping, 0 pong, 4196 push, 12 pull request, 0 pull response, 0 prune
```
-- showing the scope after the info and with circle brackets -- i kinda pref it like that but curious what others think too

src/ledger/database/rocksdb.zig
dadepo: I think it does. Since the `callRocks` function is called from within the `RocksDB ` type function, and the logger scoped within `RocksDB` is passed to it. 

Just that, since this is being called from the `RocksDB` type function, I could not find a way to properly typed the logger in `callRocks` but to use anytype.

src/accountsdb/db.zig
```diff
@@ -31,6 +31,7 @@ const AccountIndex = sig.accounts_db.index.AccountIndex;
 const AccountRef = sig.accounts_db.index.AccountRef;
 const RwMux = sig.sync.RwMux;
 const Logger = sig.trace.log.Logger;
+const ScopedLogger = sig.trace.log.ScopedLogger;
```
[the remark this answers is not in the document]
dadepo: > What do you think? Any downsides?

Applying the suggestion to `GossipDumpService`, it becomes

```zig
pub const GossipDumpService = struct {
    allocator: Allocator,
    logger: Logger,
    gossip_table_rw: *RwMux(GossipTable),
    counter: *Atomic(usize),

    const Self = @This();
    const Logger = sig.trace.ScopedLogger(@typeName(Self));
    ...
```

previously it was

```zig
pub const GossipDumpService = struct {
    allocator: Allocator,
    gossip_table_rw: *RwMux(GossipTable),
    counter: *Atomic(usize),

    const Self = @This();
```

One minor downside is that it does away with the unwritten convention that `Logger` refers to the upscoped logger, and the scoped logger will be obvious in the type, ie ` logger: ScopedLogger(@typeName(Self)),` in the above.

With the suggestion, the type definition does look cleaner? ie just `Logger`. But there is something to be said about being able to look at the logger and know this is using a scoped logger or not.

src/utils/service.zig
dadepo: [81aab5c](https://github.com/Syndica/sig/pull/345/commits/81aab5c850faeafb73b0727b3e575e97917f4a2b)

src/shred_collector/service.zig
dadepo: [ffec062](https://github.com/Syndica/sig/pull/345/commits/ffec062b6033b04605c35af9e0aae033fa1752ff)

src/accountsdb/db.zig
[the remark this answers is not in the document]
dadepo: > There could be a concept of nesting scope. 

Indeed. This can be a follow up task, if it is agreed as a new feature to be added to the logger.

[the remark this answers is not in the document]
dnut: My opinion is that we should never do this:
```zig
const logger = logger_.withScope(@src().fn_name);
```
...because it *only* includes the name of the function, which is probably not enough information for most people to understand where it actually comes from. The logs will be confusing with just a bunch of individual function names. I think it's more valuable to have higher level scopes like "gossip" and "accountsdb" so you can take a quick glance at the logs to figure out what's going on. If you want to know exactly where a log is coming from you can always search the code for the log message.

I'm fine with fully-qualified scoping, where the function name comes after the file name or struct name, but I think it's too obscure to have *only* the function name.

If you're going to do fully qualified scope, personally I think nesting scope would be the easiest way to do it. I don't think this needs to be a follow up task. It only adds a few lines of code.
```zig
pub fn addScope(self: Self, comptime new_scope: []const u8) ScopedLogger(scope ++ new_scope) {
    return self.withScope(scope ++ new_scope);
}
```

src/ledger/database/rocksdb.zig
dnut: Oh I see. I didn't read this carefully enough.

It might be best to move callRocks into the RocksDB struct, so it can utilize the generic type without resorting to anytype.

dnut: Actually, I think it would probably be better to set the scope manually to "rocksdb" for the entire file. As is, the scope is going to be pretty weird, because it will include every column family. I think the string will be huge and messy.

src/ledger/cleanup_service.zig
dnut: That's not necessary. You can just specify the same scope in both places. 
```zig
logger.withScope("ledger.cleanup_service");
...
logger_: sig.trace.ScopedLogger("ledger.cleanup_service"),
```
Or you can declare a const for the scoped logger
```zig
const Logger = sig.trace.ScopedLogger("ledger.cleanup_service");
```

src/accountsdb/db.zig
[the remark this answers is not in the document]
dnut: I feel like you could make the same argument that the `logger` field in a struct should also indicate that it is a scoped logger. For example, why wouldn't you do it like this?
```zig
pub const GossipDumpService = struct {
    allocator: Allocator,
    gossip_dump_service_scoped_logger: ScopedLogger(@typeName(Self)),
    gossip_table_rw: *RwMux(GossipTable),
    counter: *Atomic(usize),

    const Self = @This();
```
It just seems kind of redundant to specify the scope in every variable's name, when the scope itself already is a namespace.

Upon further thought, I think I see your point. I'll answer my own question...

We often need to have the same scope deal with both scoped and unscoped loggers. For example, an `init` function may need to take an unscoped logger, and then convert it to a scoped logger. 

To deal with that you have several options:
1. "Logger" always means unscoped, and use `ScopedLogger("thescope")` anywhere you need a scoped logger.
2. "Logger" always means unscoped, and you assign a new name to the individually scoped logger, like `const ScopedLogger = ScopedLogger("gossip")` (or "GossipLogger" or whatever)
3. "Logger" always means "the correct logger for this scope" and you can specify `sig.trace.Logger` if you need to deal with an unscoped logger in a context where the .
4. "Logger" always refers the most generic logger to use in the current scope. If `sig.trace.Logger` is needed in this scope, you assign it the name "Logger" and use a different name for scoped loggers. If this scope only needs scoped loggers, you can assign the scoped logger to the name "Logger".

I still like option 3 but I do see the potential for confusion. If you prefer options 1 and 2 then that's fine by me. Option 4 feels a bit too chaotic.

src/ledger/cleanup_service.zig
dadepo: [3b7bc77](https://github.com/Syndica/sig/pull/345/commits/3b7bc77f96f10e2a94dd07636156730e3e8e6640)

src/ledger/database/rocksdb.zig
dadepo: [cae2755](https://github.com/Syndica/sig/pull/345/commits/cae27554ca6894d0d8a05e56f7baa8b499b5bf05)

src/accountsdb/db.zig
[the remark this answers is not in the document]
dadepo: > so I don't know what's best. If you prefer options 1 and 2 then that's fine by me. 

I do have a slight preference for option 1. Mainly because variable names can be free to be `logger` or whatever makes sense in the context, and a look at the corresponding type gives the extra info (if it is scoped with a non-null value or not)...mirroring how any type inherently offers context for any variable.

> Option 4 feels a bit too chaotic.

Indeed.

src/accountsdb/db.zig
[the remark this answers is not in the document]
dadepo: So I don't think the nesting scope feature solves the problem that led to the use of `@src().fn_name` in the first place. 

The `@src().fn_name` was introduced as the scope to to use for free standing functions that are not part of a struct. 

For logging that occurs within a struct, the scope for that is quite clear, it is scoped to `@typeName(Self)`, but for free standing function, it is not that clear, hence the idea to scope it to `@src().fn_name`.

But given the cons of that approach, I have removed the usage of `@src().fn_name` by using a scope constant defined in the file. Ref: [45affc0](https://github.com/Syndica/sig/pull/345/commits/45affc0daea265c58a86c773f8320b7402819004)

Another approach might be to have the logger typed as `anytype` and leave it to the caller to provide either a scoped logger or un-scoped logger. But went with the using a scope constant instead. 

The nested scope might still be a feature to consider (although might have to be evaluated against the current setup where upscoped logger is always used in the boundary where a logger is passed in, into a struct - where it then get scoped appropriately) .

dadepo:
> something i noticed yew do is use this format:
> 
> ```
> time=2024-11-04T12:58:48.779Z level=info (gossip.service.GossipService) sent: 21 ping, 0 pong, 4196 push, 12 pull request, 0 pull response, 0 prune
> ```
> 
> -- showing the scope after the info and with circle brackets -- i kinda pref it like that but curious what others think too

To make it more concrete, in our case, instead of 

```
[accountsdb.index.AccountIndex] time=2024-11-04T13:00:19.477Z level=info using ram memory for account index
```

we would then have

```
time=2024-11-04T13:00:19.477Z level=info [accountsdb.index.AccountIndex] using ram memory for account index
```

Both actually works for me, so no strong preference for one.

src/accountsdb/db.zig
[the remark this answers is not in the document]
dnut: Sounds good to me. Feel free to resolve this conversation. I'm just leaving it open for visibility.

> But given the cons of that approach, I have removed the usage of @src().fn_name by using a scope constant defined in the file. Ref: [45affc0](https://github.com/Syndica/sig/pull/345/commits/45affc0daea265c58a86c773f8320b7402819004)

I like the idea of scoping by file. To me that feels like the most "normal" way of doing scoping. Initially, I assumed all loggers would be scoped by file. But scoping by struct also seems good.

I'm not saying we need to add the nested scope now. I just misunderstand why you used function names as the scope. I assumed that it was because you wanted extra granularity. I was just suggesting the nested scope as a way to include the granularity without sacrificing high level context.

> The nested scope might still be a feature to consider (although might have to be evaluated against the current setup where upscoped logger is always used in the boundary where a logger is passed in, into a struct - where it then get scoped appropriately) .

Let's say you have a file with one public function and five private functions. Since the five are private, they are all in some way called by the public function. Let's assume all of those private functions need a logger. You can define a single logger scope for the entire file. I would expect the public function to accept an unscoped logger as a parameter, and then it would convert it to a scoped logger. Then all the private functions would accept the *scoped* logger as a parameter. That way you don't need to pointlessly convert back and forth between the scoped and unscoped variants of the logger.

Now, let's say you wanted to increase the granularity of the scope in the logs for each individual logger by including the function name in the scope. I think that would be best served by appending the scopes together, so you don't lose the filename from the scope. This is where the idea of "nesting scope" could come into play.

dnut (approved):
Looks good. Nice work.

Regarding the log format, I personally prefer the time to be the first thing in the log line. So I would be happy if you want to move the scope to the right. I'm indifferent about parentheses vs brackets. These changes can come in a separate pr.

dnut:
I think you need to sign some of your commits before github will let this merge. It'll require a force push and an re-approval.

src/accountsdb/db.zig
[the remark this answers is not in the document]
dadepo: > Let's say you have a file with one public function and five private functions. Since the five are private, they are all in some way called by the public function. Let's assume all of those private functions need a logger. You can define a single logger scope for the entire file. I would expect the public function to accept an unscoped logger as a parameter, and then it would convert it to a scoped logger. Then all the private functions would accept the scoped logger as a parameter. That way you don't need to pointlessly convert back and forth between the scoped and unscoped variants of the logger

That is indeed a valid scenario having the nested scoping helps. I'll add a task for this so it does not fall through the cracks.

dnut (approved):

PR #347, opened 2024-10-31
fix(ledger): backward merkle root chaining bugs

dnut:
## bug fixes
Previously, backwards merkle root chaining checks were:
- ignoring the actual result of the check
  - always marked it as a duplicate (as if the check failed)
  - always returned true (as if the check succeeded)
- comparing chained root to chained root, which is wrong. it should be chained to actual root

## refactoring
consolidate common chaining logic into single function "checkAndReportChaining"

Improve dependency management by:
- organizing merkle root check functions into a struct since they all share some common dependencies. this simplifies the function signatures and keeps things standardized and consistent
- utilizing working stores.
  - this is better than passing down the granular transitive dependencies (database and hashmap) separately because it clarifies what this code actually needs. all it really needs is some kind of store for erasure metas. it's not actually important in this context that it's managed through a negotiation between database and hashmap calls. this is the whole point of the pending state, to abstract away that negotiation. passing down all the transitive dependencies makes the function parameters a mess, and it becomes unclear why the code actually needs any of the dependencies.
  - this is better than passing down the whole pending state struct because it reduces the number of dependencies and it clarifies what is actually needed, rather than pretending the entire state is needed.

## change in behavior

merkle root checks now return an error if the merkle root is missing, instead of ignoring it, because that means there is a severe bug in the validator and it needs to be shutdown and patched immediately.

### DIFF
--- a/src/ledger/shred_inserter/merkle_root_checks.zig
+++ b/src/ledger/shred_inserter/merkle_root_checks.zig
@@ -22,8 +22,13 @@ const MerkleRootMeta = ledger.meta.MerkleRootMeta;
 const PossibleDuplicateShred = shred_inserter.working_state.PossibleDuplicateShred;
 const Shred = ledger.shred.Shred;
 const ShredId = ledger.shred.ShredId;
+
+const DuplicateShredsWorkingStore = shred_inserter.working_state.DuplicateShredsWorkingStore;
+const ErasureMetaWorkingStore = shred_inserter.working_state.ErasureMetaWorkingStore;
+const MerkleRootMetaWorkingStore = shred_inserter.working_state.MerkleRootMetaWorkingStore;
+const PendingInsertShredsState = shred_inserter.working_state.PendingInsertShredsState;
+const ShredWorkingStore = shred_inserter.working_state.ShredWorkingStore;
 const WorkingEntry = shred_inserter.working_state.WorkingEntry;
-const WorkingShredStore = shred_inserter.working_state.WorkingShredStore;
 
 const key_serializer = ledger.database.key_serializer;
 const value_serializer = ledger.database.value_serializer;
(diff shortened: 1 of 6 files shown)


src/ledger/shred_inserter/merkle_root_checks.zig
```diff
@@ -228,43 +197,88 @@ pub fn checkBackwardsChainedMerkleRootConsistency(
...
-    const merkle_root = shred_mod.layout.getChainedMerkleRoot(prev_shred);
-    const chained_merkle_root = shred.chainedMerkleRoot() catch null;
 
-    if (!checkChaining(merkle_root, chained_merkle_root)) {
-        logger.warn().logf(&newlinesToSpaces(
-            \\Received conflicting chained merkle roots for slot: {}, shred {any} type {any} \
-            \\chains to merkle root {any}, however previous fec set code \
-            \\shred {any} has merkle root {any}. Reporting as duplicate
-        ), .{
+    const early_shred, const late_shred = switch (direction) {
```
dadepo, line 235: nit: kinda find `prev_shred/next_shred` easier to parse than `early_shred/late_shred`.

dnut: I intentionally made this change to move away from "previous" and "next" because I felt that those names were misleading. When they're mixed up together in the same context, I feel that it implies they come before and after some central shred, implying there are three. But there are always only two shreds.

I get that early and late aren't the best names either. What about old_shred and new_shred?

dadepo: Gotcha. I think `old_shred` and `new_shred` is better. Or maybe `older_shred`/`newer_shred` to convey it is some sort of comparison.

src/ledger/shred_inserter/merkle_root_checks.zig
```diff
@@ -228,43 +197,88 @@ pub fn checkBackwardsChainedMerkleRootConsistency(
...
+        .forward => .{ shred.payload(), other_shred },
+        .backward => .{ other_shred, shred.payload() },
+    };
+
+    const chained_merkle_root = shred_mod.layout.getChainedMerkleRoot(late_shred);
```
dadepo, line 240: Agave has this as

```
let merkle_root = shred::layout::get_merkle_root(&prev_shred);
 let chained_merkle_root = shred.chained_merkle_root().ok();
```

where 
 - `prev_shred` -> `merkle_root` and
 - `current_shred` -> `chained_merkle_root`

You seem to have flipped that into

- `late_shred` aka `prev_shred` -> `chained_merkle_root`
- `early_shred` aka current shred - > `merkle_root`

and I guess that is because it is only the later shred that can only be considered to have chained merkle root? and not the one currently being processed?

dnut: > late_shred aka prev_shred -> chained_merkle_root
> early_shred aka current shred - > merkle_root

You have it backwards. `prev_shred` in the backward chaining function becomes `early_shred` in this function.

Late shred comes after early shred. A shred's chained merkle root is always supposed to point to the erasure set that came before it. So the late shred's chained merkle root is supposed to equal the early shred's merkle root.

src/ledger/shred_inserter/merkle_root_checks.zig
dnut: e827130473ba05e0cd0b88c3e4b6c6080f660f02

dadepo (approved):

PR #348, opened 2024-11-01
perf(accountsdb): experiment with lmdb for rooted account storage

Sobeston:
notes:
- this PR does not use serialisation, but is still doing copies (zero-copy is possible by persisting the transaction)
   - we could architect account fetching around this, where the code that fetched the account is expected to end the transaction once it's done with the account so that it doesn't need to copy or allocate for the account at all (!).
      - could perhaps use refcounted accounts around the codebase that end their transaction upon destroying.
- read performance is pretty good, but still a little slower than previous benchmarks. 
   - It seems suspect that lmdb is supposed to be extremely fast for reads, yet our previous naiive implementation is ~40% faster at reading somehow. 
   - There's some things that can be done to improve read speed a little, like reusing transaction and cursor objects, but they're small wins.
      - If we use `MDB_NOTLS`, we can do this without complicating our usage.
   - CPU time is dominated by mdb_get and mdb_cursor_get, which are out of my control, unless I've used them incorrectly (which I don't think I have).
- we would also want an additional index for Slot->Pubkey for snapshot generation. Worth noting that this would make writes a bit slower.
- writes do not block reads at all; as soon as a write transaction is committed, new read transactions will be able to see it.
- potentially unsafe options (MDB_NOSYNC | MDB_NOLOCK) do not improve read performance.
- a benefit of lmdb, is that read speed scales linearly with the number of threads - there's no limit to how parallel our reads can be. If we expect our reads to be highly parallel this would be an advantage. 
- Writes are slow, but still usable. Thought on write performance:
   - we can actually use MDB_NOLOCK, iff we have a single writer thread. This is probably a good idea; we can send to it via a channel. 
   - we could also trivially shard the database by pubkey to effectively have multiple writer threads
   - using MDB_NOSYNC massively increases write performance - 100K random account writes goes from 100 seconds to only 2.8 seconds. This flag is potentially unsafe.
      - MDB_NOSYNC would be safe for us to use on a filesystem that preserves write order (I believe this is by default on ext4 systems, which will be most linux machines). We could lose the most recent writes before crashing, but in the context of a blockchain, if we crash we'd need to catch up anyway, so this is fine.
   -  I think we should use MDB_NOSYNC | MDB_NOLOCK.
      - this means we can't use `MDB_WRITEMAP`, which allows us to write to mmapped memory. I can't think of an advantage of doing that so this is fine.
 

A flamegraph of doing 100K random reads (100 iterations):
image

data layout is as follows:
- a map from Pubkey to Slot (allowing duplicates, kept sorted so we can quickly fetch latest rooted slot)
- a map from Pubkey+Slot to (serialised) Account

when viewed using lmdbnav
image
image
image

## Final thoughts

- For LMDB to make sense, it would have to perform well at the scale of accountsdb - the network has somewhere around 650 million accounts. If it can't handle billions of entries it's not a good idea.
   - Unfortunately its performance totally blew up when testing with large amounts of accounts; the performance was already not perfect at 100K accounts, but when grown to 10M, it was slow enough to be unusable (>7 hours to populate the database, >3 hours to read 10M entries from it).
   
Further optimisations not fully tested:
- using one table instead of two, which would halve our lookups (albeit at the expense of a larger table, which might slow down lookups & writes)
- reusing transaction and cursor objects
- using custom functions for comparisons (e.g. integer comparison is faster than memeql)
- sharding the database
   - we could super trivially shard the database by pubkey, allowing:
      - one writer thread per shard, for much higher write throughput
      - much smaller tables for faster reads and writes
      - we would lose some consistency 
         - i.e. if we want to look up every account modified between slot X and slot Y, we would need to query multiple dbs, which means we wouldn't have the safety of using one transaction - dbs could be in slightly different states. 
         - I think for our use case this would probably be fine
      - this might be taking things too far - we'd be squaring the circle. At this point I think it's obvious that it isn't the right solution.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -41,6 +41,9 @@ pub fn build(b: *Build) void {
     const rocksdb_dep = b.dependency("rocksdb", dep_opts);
     const rocksdb_mod = rocksdb_dep.module("rocksdb-bindings");
 
+    const lmdb_dep = b.dependency("lmdb", dep_opts);
+    const lmdb_mod = lmdb_dep.module("lmdb");
+
     // expose Sig as a module
     const sig_mod = b.addModule("sig", .{
         .root_source_file = b.path("src/sig.zig"),
@@ -52,6 +55,7 @@ pub fn build(b: *Build) void {
     sig_mod.addImport("zstd", zstd_mod);
     sig_mod.addImport("curl", curl_mod);
     sig_mod.addImport("rocksdb", rocksdb_mod);
+    sig_mod.addImport("lmdb", lmdb_mod);
 
     // main executable
     const sig_exe = b.addExecutable(.{
@@ -69,6 +73,7 @@ pub fn build(b: *Build) void {
     sig_exe.root_module.addImport("zig-network", zig_network_module);
     sig_exe.root_module.addImport("zstd", zstd_mod);
     sig_exe.root_module.addImport("rocksdb", rocksdb_mod);
+    sig_exe.root_module.addImport("lmdb", lmdb_mod);
     sig_exe.linkLibC();
 
     const main_exe_run = b.addRunArtifact(sig_exe);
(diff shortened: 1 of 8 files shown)


0xNineteen:
can we open a PR to add the PR description/details into a `docs/research/` folder ?

PR #352, opened 2024-11-01
feat(ledger): support LMDB as a blockstore database backend

dnut:
https://github.com/Syndica/lmdb-zig

The bulk of this PR is an implementation for the blockstore's `Database` interface using LMDB, and adding more tests for the interface.

I also added a build option to customize the database backend, for example:
```
zig build -Dblockstore-db=lmdb
```
RocksDB remains the default for now.

For any build, only the necessary dependencies will be compiled. For example, lmdb will never be compiled, unless you specify `-Dblockstore-db=lmdb`.

A few bugs were uncovered in the hashmap db when testing the databases more strictly, so I fixed those in here as well.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -63,8 +63,11 @@ pub fn build(b: *Build) void {
     sig_mod.addImport("httpz", httpz_mod);
     sig_mod.addImport("zstd", zstd_mod);
     sig_mod.addImport("curl", curl_mod);
-    if (blockstore_db == .rocksdb) sig_mod.addImport("rocksdb", rocksdb_mod);
-    if (blockstore_db == .lmdb) sig_mod.addImport("lmdb", lmdb_mod);
+    switch (blockstore_db) {
+        .rocksdb => sig_mod.addImport("rocksdb", rocksdb_mod),
+        .lmdb => sig_mod.addImport("lmdb", lmdb_mod),
+        .hashmap => {},
+    }
     sig_mod.addOptions("build-options", build_options);
 
     // main executable
@@ -82,8 +85,11 @@ pub fn build(b: *Build) void {
     sig_exe.root_module.addImport("zig-cli", zig_cli_module);
     sig_exe.root_module.addImport("zig-network", zig_network_module);
     sig_exe.root_module.addImport("zstd", zstd_mod);
-    if (blockstore_db == .rocksdb) sig_exe.root_module.addImport("rocksdb", rocksdb_mod);
-    if (blockstore_db == .lmdb) sig_exe.root_module.addImport("lmdb", lmdb_mod);
+    switch (blockstore_db) {
+        .rocksdb => sig_exe.root_module.addImport("rocksdb", rocksdb_mod),
+        .lmdb => sig_exe.root_module.addImport("lmdb", lmdb_mod),
+        .hashmap => {},
+    }
     sig_exe.root_module.addOptions("build-options", build_options);
     sig_exe.linkLibC();
 
@@ -164,11 +170,12 @@ pub fn build(b: *Build) void {
     benchmark_exe.root_module.addImport("zig-network", zig_network_module);
     benchmark_exe.root_module.addImport("httpz", httpz_mod);
     benchmark_exe.root_module.addImport("zstd", zstd_mod);
-    benchmark_exe.root_module.addImport("rocksdb", rocksdb_mod);
-    benchmark_exe.root_module.addImport("lmdb", lmdb_mod);
     benchmark_exe.root_module.addImport("prettytable", pretty_table_mod);
-    if (blockstore_db == .rocksdb) sig_exe.root_module.addImport("rocksdb", rocksdb_mod);
-    if (blockstore_db == .lmdb) sig_exe.root_module.addImport("lmdb", lmdb_mod);
+    switch (blockstore_db) {
+        .rocksdb => benchmark_exe.root_module.addImport("rocksdb", rocksdb_mod),
+        .lmdb => benchmark_exe.root_module.addImport("lmdb", lmdb_mod),
+        .hashmap => {},
+    }
     benchmark_exe.root_module.addOptions("build-options", build_options);
     benchmark_exe.linkLibC();
 
--- a/build.zig
+++ b/build.zig
@@ -10,6 +10,11 @@ pub fn build(b: *Build) void {
     const filters = b.option([]const []const u8, "filter", "List of filters, used for example to filter unit tests by name"); // specified as a series like `-Dfilter="filter1" -Dfilter="filter2"`
     const enable_tsan = b.option(bool, "enable-tsan", "Enable TSan for the test suite");
     const no_run = b.option(bool, "no-run", "Do not run the selected step and install it") orelse false;
+    const blockstore_db = b.option(BlockstoreDB, "blockstore-db", "Blockstore database backend") orelse .rocksdb;
+
+    // Build options
+    const build_options = b.addOptions();
+    build_options.addOption(BlockstoreDB, "blockstore_db", blockstore_db);
 
     // CLI build steps
     const sig_step = b.step("run", "Run the sig executable");
@@ -58,8 +63,9 @@ pub fn build(b: *Build) void {
     sig_mod.addImport("httpz", httpz_mod);
     sig_mod.addImport("zstd", zstd_mod);
     sig_mod.addImport("curl", curl_mod);
-    sig_mod.addImport("rocksdb", rocksdb_mod);
-    sig_mod.addImport("lmdb", lmdb_mod);
+    if (blockstore_db == .rocksdb) sig_mod.addImport("rocksdb", rocksdb_mod);
+    if (blockstore_db == .lmdb) sig_mod.addImport("lmdb", lmdb_mod);
+    sig_mod.addOptions("build-options", build_options);
 
     // main executable
     const sig_exe = b.addExecutable(.{
@@ -76,8 +82,9 @@ pub fn build(b: *Build) void {
     sig_exe.root_module.addImport("zig-cli", zig_cli_module);
     sig_exe.root_module.addImport("zig-network", zig_network_module);
     sig_exe.root_module.addImport("zstd", zstd_mod);
-    sig_exe.root_module.addImport("rocksdb", rocksdb_mod);
-    sig_exe.root_module.addImport("lmdb", lmdb_mod);
+    if (blockstore_db == .rocksdb) sig_exe.root_module.addImport("rocksdb", rocksdb_mod);
+    if (blockstore_db == .lmdb) sig_exe.root_module.addImport("lmdb", lmdb_mod);
+    sig_exe.root_module.addOptions("build-options", build_options);
     sig_exe.linkLibC();
 
     const main_exe_run = b.addRunArtifact(sig_exe);
@@ -117,6 +124,7 @@ pub fn build(b: *Build) void {
     unit_tests_exe.root_module.addImport("zstd", zstd_mod);
     unit_tests_exe.root_module.addImport("rocksdb", rocksdb_mod);
     unit_tests_exe.root_module.addImport("lmdb", lmdb_mod);
+    unit_tests_exe.root_module.addOptions("build-options", build_options);
     unit_tests_exe.linkLibC();
 
     const unit_tests_exe_run = b.addRunArtifact(unit_tests_exe);
@@ -159,6 +167,9 @@ pub fn build(b: *Build) void {
     benchmark_exe.root_module.addImport("rocksdb", rocksdb_mod);
     benchmark_exe.root_module.addImport("lmdb", lmdb_mod);
     benchmark_exe.root_module.addImport("prettytable", pretty_table_mod);
+    if (blockstore_db == .rocksdb) sig_exe.root_module.addImport("rocksdb", rocksdb_mod);
+    if (blockstore_db == .lmdb) sig_exe.root_module.addImport("lmdb", lmdb_mod);
+    benchmark_exe.root_module.addOptions("build-options", build_options);
     benchmark_exe.linkLibC();
 
     const benchmark_exe_run = b.addRunArtifact(benchmark_exe);
@@ -202,3 +213,9 @@ fn makeZlsNotInstallAnythingDuringBuildOnSave(b: *Build) void {
         artifact.generated_bin = null;
     }
 }
+
+const BlockstoreDB = enum {
+    rocksdb,
+    lmdb,
+    hashmap,
+};
--- a/build.zig
+++ b/build.zig
@@ -42,6 +42,9 @@ pub fn build(b: *Build) void {
     const rocksdb_dep = b.dependency("rocksdb", dep_opts);
     const rocksdb_mod = rocksdb_dep.module("rocksdb-bindings");
 
+    const lmdb_dep = b.dependency("lmdb", dep_opts);
+    const lmdb_mod = lmdb_dep.module("lmdb");
+
     const pretty_table_dep = b.dependency("prettytable", dep_opts);
     const pretty_table_mod = pretty_table_dep.module("prettytable");
 
@@ -56,6 +59,7 @@ pub fn build(b: *Build) void {
     sig_mod.addImport("zstd", zstd_mod);
     sig_mod.addImport("curl", curl_mod);
     sig_mod.addImport("rocksdb", rocksdb_mod);
+    sig_mod.addImport("lmdb", lmdb_mod);
 
     // main executable
     const sig_exe = b.addExecutable(.{
@@ -73,6 +77,7 @@ pub fn build(b: *Build) void {
     sig_exe.root_module.addImport("zig-network", zig_network_module);
     sig_exe.root_module.addImport("zstd", zstd_mod);
     sig_exe.root_module.addImport("rocksdb", rocksdb_mod);
+    sig_exe.root_module.addImport("lmdb", lmdb_mod);
     sig_exe.linkLibC();
 
     const main_exe_run = b.addRunArtifact(sig_exe);
@@ -111,6 +116,7 @@ pub fn build(b: *Build) void {
     unit_tests_exe.root_module.addImport("zig-network", zig_network_module);
     unit_tests_exe.root_module.addImport("zstd", zstd_mod);
     unit_tests_exe.root_module.addImport("rocksdb", rocksdb_mod);
+    unit_tests_exe.root_module.addImport("lmdb", lmdb_mod);
     unit_tests_exe.linkLibC();
 
     const unit_tests_exe_run = b.addRunArtifact(unit_tests_exe);
@@ -151,6 +157,7 @@ pub fn build(b: *Build) void {
     benchmark_exe.root_module.addImport("httpz", httpz_mod);
     benchmark_exe.root_module.addImport("zstd", zstd_mod);
     benchmark_exe.root_module.addImport("rocksdb", rocksdb_mod);
+    benchmark_exe.root_module.addImport("lmdb", lmdb_mod);
     benchmark_exe.root_module.addImport("prettytable", pretty_table_mod);
     benchmark_exe.linkLibC();
 
--- a/build.zig
+++ b/build.zig
@@ -128,8 +128,11 @@ pub fn build(b: *Build) void {
     unit_tests_exe.root_module.addImport("httpz", httpz_mod);
     unit_tests_exe.root_module.addImport("zig-network", zig_network_module);
     unit_tests_exe.root_module.addImport("zstd", zstd_mod);
-    unit_tests_exe.root_module.addImport("rocksdb", rocksdb_mod);
-    unit_tests_exe.root_module.addImport("lmdb", lmdb_mod);
+    switch (blockstore_db) {
+        .rocksdb => unit_tests_exe.root_module.addImport("rocksdb", rocksdb_mod),
+        .lmdb => unit_tests_exe.root_module.addImport("lmdb", lmdb_mod),
+        .hashmap => {},
+    }
     unit_tests_exe.root_module.addOptions("build-options", build_options);
     unit_tests_exe.linkLibC();
 
--- a/src/ledger/database/hashmap.zig
+++ b/src/ledger/database/hashmap.zig
@@ -178,9 +178,17 @@ pub fn SharedHashMapDB(comptime column_families: []const ColumnFamily) type {
                         const cf_index, const key = delete_ix;
                         self.maps[cf_index].delete(self.allocator, key);
                     },
-                    .delete_range => {
-                        // TODO: also add to database tests
-                        @panic("not implemented");
+                    .delete_range => |delete_range_ix| {
+                        const cf_index, const start, const end = delete_range_ix;
+                        const keys, _ = self.maps[cf_index].map.range(start, end);
+                        const to_delete = try batch.allocator.alloc([]const u8, keys.len);
+                        defer batch.allocator.free(to_delete);
+                        for (keys, 0..) |key, i| {
+                            to_delete[i] = key;
+                        }
+                        for (to_delete) |delete_key| {
+                            self.maps[cf_index].delete(self.allocator, delete_key);
+                        }
                     },
                 }
             }
@@ -266,7 +274,7 @@ pub fn SharedHashMapDB(comptime column_families: []const ColumnFamily) type {
                 const end_bytes = try key_serializer.serializeAlloc(self.allocator, end);
                 errdefer self.allocator.free(end_bytes);
                 const cf_index = cf.find(column_families);
-                self.instructions.append(
+                try self.instructions.append(
                     self.allocator,
                     .{ .delete_range = .{ cf_index, start_bytes, end_bytes } },
                 );
--- a/src/ledger/database/hashmap.zig
+++ b/src/ledger/database/hashmap.zig
@@ -26,9 +26,10 @@ pub fn SharedHashMapDB(comptime column_families: []const ColumnFamily) type {
 
         pub fn open(
             allocator: Allocator,
-            _: Logger,
+            logger: Logger,
             _: []const u8,
         ) Allocator.Error!Self {
+            logger.info().log("Initializing SharedHashMapDB");
             var maps = try allocator.alloc(SharedHashMap, column_families.len);
             errdefer {
                 for (maps) |*m| m.deinit();
--- a/src/ledger/database/hashmap.zig
+++ b/src/ledger/database/hashmap.zig
@@ -455,7 +455,5 @@ const SharedHashMap = struct {
 };
 
 comptime {
-    if (build_options.blockstore_db == .hashmap) {
-        _ = &database.interface.testDatabase(SharedHashMapDB);
-    }
+    _ = &database.interface.testDatabase(SharedHashMapDB);
 }
--- a/src/ledger/database/hashmap.zig
+++ b/src/ledger/database/hashmap.zig
@@ -241,6 +241,7 @@ pub fn SharedHashMapDB(comptime column_families: []const ColumnFamily) type {
                 key: cf.Key,
                 value: cf.Value,
             ) anyerror!void {
+                std.debug.assert(!self.executed.*);
                 const k_bytes = try key_serializer.serializeAlloc(self.allocator, key);
                 errdefer self.allocator.free(k_bytes);
                 const v_bytes = try value_serializer.serializeAlloc(self.allocator, value);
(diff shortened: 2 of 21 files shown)


src/ledger/database/interface.zig
```diff
@@ -546,5 +547,169 @@ fn tests(comptime Impl: fn ([]const ColumnFamily) type) type {
 
             try std.testing.expectEqual(null, try iter.next());
         }
+
+        pub fn @"iterator forward start before all"() !void {
```
Rexicon226, line 551: We should merge #353 and then rebase + migrate these tests.

dnut: e2910ce044b991a4efd475b5c95bae141ffa5d67

src/ledger/database/lmdb.zig
```diff
@@ -0,0 +1,716 @@
+const std = @import("std");
+const c = @import("lmdb");
```
Rexicon226, line 2: Do you prefer to use `c`? I like to use the actual name, since there could be multiple translate-c imports, and it wouldn't make sense. Just a nit.

src/ledger/database/lmdb.zig
```diff
@@ -0,0 +1,716 @@
...
+        .vtable = &vtable,
+    };
+}
+
+fn resetTxnFree(ctx: *anyopaque, _: []u8, _: u8, _: usize) void {
+    const txn: *c.MDB_txn = @ptrCast(@alignCast(ctx));
+    c.mdb_txn_abort(txn);
+}
+
+fn ret(constructor: anytype, args: anytype) LmdbError!TypeToCreate(constructor) {
```
Rexicon226, line 396: This, and the functions under it are clearly relying on some behavior that makes no sense to me, a casual reader. 
Could you document them better?

src/ledger/database/lmdb.zig
```diff
@@ -0,0 +1,716 @@
...
+            // allocate cf handles
+            const dbis = try allocator.alloc(c.MDB_dbi, column_families.len);
+            errdefer allocator.free(dbis);
+
+            // save cf handles
+            inline for (column_families, 0..) |cf, i| {
+                // open cf/database, creating if necessary
+                dbis[i] = try ret(c.mdb_dbi_open, .{
+                    txn,
+                    @as([*c]const u8, @ptrCast(cf.name)),
```
Rexicon226, line 47: This is not a translated file, it should never use C pointers.

src/ledger/database/lmdb.zig
```diff
@@ -0,0 +1,716 @@
...
+                }
+            };
+        }
+    };
+}
+
+fn toVal(bytes: []const u8) c.MDB_val {
+    return .{
+        .mv_size = bytes.len,
+        .mv_data = @constCast(@ptrCast(bytes.ptr)),
```
Rexicon226, line 370: I dunno about this. While I see what you're doing with this constCast, it can lead to very illegal behavior. One of the *only* use cases of constCast is to mitigate incorrect APIs, which I actually don't think this is. 
I bet the implementation stores a mutable pointer because it frees it at some point. I have not looked into lmdb, so don't quote me on that.

Please take a `[]u8` here and duplicate and constant strings or string literals instead. I refuse to use constCast anywhere in this codebase.

Rexicon226, line 370: I dunno about this. While I see what you're doing with this constCast, it can lead to very illegal behavior. One of the *only* use cases of constCast is to mitigate incorrect APIs, which I actually don't think this is. 
I bet the implementation stores a mutable pointer because it frees it at some point. I have not looked into lmdb, so don't quote me on that.

Please take a `[]u8` here and duplicate and constant strings or string literals instead. I refuse to use constCast anywhere in this codebase.

src/ledger/database/lmdb.zig
```diff
@@ -0,0 +1,716 @@
...
+    };
+}
+
+fn fromVal(value: c.MDB_val) []const u8 {
+    const ptr: [*c]u8 = @ptrCast(value.mv_data);
```
Rexicon226, line 375: This isn't a translated file, it shouldn't be using C pointers.

src/ledger/database/lmdb.zig
```diff
@@ -0,0 +1,716 @@
...
+    prev_dup = 13,
+    /// Position at last data item of previous key
+    prev_nodup = 14,
+
+    /// Position at previous page and return up to a page of duplicate data items.
+    /// Only for #MDB_DUPFIXED
+    prev_multiple = 18,
+};
+
+fn result(int: isize) LmdbError!void {
```
Rexicon226, line 515: I don't think `result` is the best name for this. 

These seems more like a "convert errno to error set" function, so I'd recommend it returns an error set and then be wrapped in a "check for error return code" function if that makes sense.

src/ledger/database/lmdb.zig
Rexicon226: Interesting arguments, I agree with them.

src/ledger/database/lmdb.zig
dnut: I'd love to find a better name for this, but I don't see the point of splitting it into two functions.

Assuming the name is descriptive, I feel that it helps readability for the name to be concise, since it's used for every call to lmdb. I'm struggling to find a name that feels satisfying in that regard. I think "lmdbReturnCodeToErrorUnion" is sufficiently descriptive, but it feels like clutter.

src/ledger/database/lmdb.zig
dnut: > Please take a []u8 here and duplicate and constant strings or string literals instead.

This seems doable. I probably won't need to dupe anything. It will require some changes to preexisting code, but shouldn't be too bad.

--------

> I bet the implementation stores a mutable pointer because it frees it at some point.

toVal is used for inputs to LMDB. These are pointers that are allocated by the caller (in zig) and passed into an LMDB function as an input. They are not freed by LMDB.

> One of the only use cases of constCast is to mitigate incorrect APIs, which I actually don't think this is.

My understanding is that when MDB_val is passed as an input into LMDB, the data behind the pointer is never mutated. It only reads the data.

Would you call that an incorrect API? If it says it takes mutable pointers but never mutates them?

> it can lead to very illegal behavior

What kind of illegal behavior could happen in this situation?

dnut: I'm resolving this comment since it is a duplicate

Sobeston: Yeah lmdb's api is a bit odd w.r.t parameters there. As far as I understand it, yes, it will not modify that value at most call sites. The reason for this is some functions & options will use the key as an out param, e.g. `mdb_cursor_get` with `MDB_GET_CURRENT` will write to both the key and value params.

If you want to assert that it's const, I think it's best to be super explicit at the callsite of toVal. We wouldn't want to e.g. assert that something's const when it isn't (even if the memory is safely writeable).

src/ledger/database/lmdb.zig
```diff
@@ -0,0 +1,716 @@
...
+    };
+}
+
+fn IntermediateType(function: anytype) type {
+    const params = @typeInfo(@TypeOf(function)).Fn.params;
+    return @typeInfo(params[params.len - 1].type.?).Pointer.child;
+}
+
+const MDB_CREATE = 0x40000;
+const MDB_RDONLY = 0x20000;
```
Sobeston, line 424: Could take these from your c import

src/ledger/database/lmdb.zig
```diff
@@ -0,0 +1,716 @@
...
+        25 => error.ENOTTY,
+        26 => error.ETXTBSY,
+        27 => error.EFBIG,
+        28 => error.ENOSPC,
+        29 => error.ESPIPE,
+        30 => error.EROFS,
+        31 => error.EMLINK,
+        32 => error.EPIPE,
+        33 => error.EDOM,
+        34 => error.ERANGE,
```
Sobeston, line 572: These numbers are technically platform dependent (yeah... :facepalm:)

image

src/ledger/database/lmdb.zig
InKryption: Weighing in:
> Would you call that an incorrect API? If it says it takes mutable pointers but never mutates them?

Yeah, this is explicitly an incorrect API; it doesn't adequately communicate its intent.

src/ledger/database/lmdb.zig
```diff
@@ -0,0 +1,716 @@
...
+    const vtable = .{
+        .alloc = &sig.utils.allocators.noAlloc,
+        .resize = &Allocator.noResize,
+        .free = &resetTxnFree,
+    };
+    return .{
+        .ptr = @ptrCast(@alignCast(txn)),
+        .vtable = &vtable,
+    };
+}
```
Sobeston, line 389: A bit confused by this, could we get a comment?

src/ledger/database/lmdb.zig
```diff
@@ -0,0 +1,716 @@
...
+pub fn LMDB(comptime column_families: []const ColumnFamily) type {
+    return struct {
+        allocator: Allocator,
+        env: *c.MDB_env,
+        dbis: []const c.MDB_dbi,
+        path: []const u8,
+
+        const Self = @This();
+
+        pub fn open(allocator: Allocator, logger: Logger, path: []const u8) anyerror!Self {
```
Sobeston, line 25: What's the reasoning for anyerror everywhere?

src/ledger/database/lmdb.zig
InKryption: How about `maybeError`, which reads nicely in `try maybeError(int)`, ie "try unwrap this which is maybe an error".

Sobeston (changes requested):
Broad strokes seems good, got a few comments for now

src/ledger/database/lmdb.zig
Rexicon226: Thanks for explaining everyone, I agree lmdb is in the wrong here. Nevertheless, we should not use `@constCast` if we can avoid it.

> What kind of illegal behavior could happen in this situation?

I originally assumed that either 
1. lmdb is freeing the data at some point, since free takes a mutable pointer. freeing a piece of data in a rodata section is invalid and would result in a panic.
2. it's writing to the data somehow. writing to data in rodata is also invalid and would cause a fault.

src/ledger/database/lmdb.zig
Rexicon226: Ah, I missed this, this is very incorrect. You need to integrate with `std.posix.errno`.

src/ledger/database/lmdb.zig
dnut: I wrote a comptime function to enforce interface implementation by these databases. That code does not work if a functions has an inferred error return type that may differ depending on comptime parameters, because the return type is null when you do `@typeInfo` on the function.

I can improve some of these (like open) that do not use bincode. But for functions that call bincode, we'll need anyerror, unless I loosen up the interface enforcement a bit more.

build.zig
```diff
@@ -51,7 +59,9 @@ pub fn build(b: *Build) void {
     sig_mod.addImport("httpz", httpz_mod);
     sig_mod.addImport("zstd", zstd_mod);
     sig_mod.addImport("curl", curl_mod);
-    sig_mod.addImport("rocksdb", rocksdb_mod);
+    if (blockstore_db == .rocksdb) sig_mod.addImport("rocksdb", rocksdb_mod);
```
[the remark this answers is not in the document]
Rexicon226: Looking back at this comment, I still think this should be a switch, but for a different reason. Saying "so it won't be missed" doesn't make sense, since obviously it won't be missed. Nothing would work otherwise.
My new reason for it is that I just think a switch looks better. It also shows that you don't need to have two db imports at the same time.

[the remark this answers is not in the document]
dnut: eaac69fd0355df004e33c20b21c9d0f2ae13a6f2

src/ledger/database/lmdb.zig
dnut: 890d797380ee60f0a6513a5708ec0d79754fa586

src/ledger/database/lmdb.zig
dnut: 890d797380ee60f0a6513a5708ec0d79754fa586

src/ledger/database/lmdb.zig
dnut: 551ecf6c7ecb49ba5da09d5997beec6625947f41

src/ledger/database/lmdb.zig
```diff
@@ -0,0 +1,713 @@
...
+            errdefer allocator.free(dbis);
+
+            // save cf handles
+            inline for (column_families, 0..) |cf, i| {
+                // open cf/database, creating if necessary
+                dbis[i] = try ret(c.mdb_dbi_open, .{
+                    txn,
+                    @as([*]const u8, @ptrCast(cf.name)),
+                    c.MDB_CREATE,
+                });
```
Sobeston, line 49: a nit:
- no need to ptrCast, `.ptr` on the slice will give you a `[*]const u8`

the issue:
I'm pretty sure lmdb will be expecting a c-style 0-terminated string. So giving it this pointer will not be safe - if this worked so far, it's likely only down to coincidence that the slice's data ended with a 0.
Two options:
- use a `[:0]const u8` for cf.name, which is a valid slice *and* a valid sentinel terminated pointer
- use allocator.dupeZ to copy it into a buffer that ends with a 0 

(this compiling at all is the issue with using C pointer types in general)

src/ledger/database/lmdb.zig
Rexicon226: Two things here.
1. I took a quick glance at lmbd, and it calls `strlen` on the name passed in. This should be a `[*:0]const u8` type. Either call `allocator.dupeZ` on the name, or just store `cf.name` as a `[:0]const u8`.
2. There is no need for a `@ptrCast` here, doing `slice.ptr` yields `[*]T`.

build.zig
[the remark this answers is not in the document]
Rexicon226: Perfect. I'd still like to use lmdb as a lazy dependency, however, that can easily be a follow-up enchantment.

src/ledger/database/lmdb.zig
```diff
@@ -0,0 +1,713 @@
...
+        const Self = @This();
+
+        pub fn open(allocator: Allocator, logger: Logger, path: []const u8) anyerror!Self {
+            logger.info().log("Initializing LMDB");
+            const owned_path = try allocator.dupe(u8, path);
+
+            // create and open the database
+            const env = try ret(c.mdb_env_create, .{});
+            try result(c.mdb_env_set_maxdbs(env, column_families.len));
+            try result(c.mdb_env_open(env, @ptrCast(path), 0, 0o700));
```
dadepo, line 32: should the @ptrCast not be using `owned_path` instead?

src/ledger/database/lmdb.zig
dnut: 229156237aa44950615578a02e4285f7d153b473

src/ledger/database/lmdb.zig
dnut: > I'm pretty sure lmdb will be expecting a c-style 0-terminated string. 

I don't think so. LMDB always requires you to specify the length. It needs to use this because there is no safe sentinel value for a slice of arbitrary bytes.

dnut: ac1e07da776c82a5ca62f1bae7590c8342d4aa1a

src/ledger/database/lmdb.zig
```diff
@@ -0,0 +1,594 @@
...
+            defer key_bytes.deinit();
+            const val_bytes = try value_serializer.serializeToRef(self.allocator, value);
+            defer val_bytes.deinit();
+
+            const txn = try ret(c.mdb_txn_begin, .{ self.env, null, 0 });
+            errdefer c.mdb_txn_abort(txn);
+
+            var key_val = toVal(key_bytes.data);
+            var val_val = toVal(val_bytes.data);
+            try result(c.mdb_put(txn, self.dbi(cf), &key_val, &val_val, 0));
```
dadepo, line 93: whats the expected behaviour regarding putting key that already exist? Will it override, error or append and does the `0` flag in `c.mdb_put` aligns with the expected behaviour?

src/ledger/database/lmdb.zig
dnut: I forgot the context here is actually a string, not the storage of bytes in the database. I'll fix it

src/ledger/database/lmdb.zig
dnut: 6f3c9ccf6846de4a4638886851f56daf092a9b1f

src/ledger/database/lmdb.zig
dnut: It doesn't need to be owned. LMDB dupes it:
```C
env->me_path = strdup(path);
```
But it does need to be zero terminated string. So I decided to just make it use the owned path for that.

b8adf694e731c1828150f8c802193b60abcd45f7

src/ledger/database/lmdb.zig
dnut: b8adf694e731c1828150f8c802193b60abcd45f7

src/ledger/database/lmdb.zig
dnut: ac1e07da776c82a5ca62f1bae7590c8342d4aa1a
b8adf694e731c1828150f8c802193b60abcd45f7

src/ledger/database/lmdb.zig
dnut: 98f4d11be351c8ea5e95916f726e92aaeaff41df

src/ledger/database/lmdb.zig
dnut: f086a562142cb7b76e1bdfcf93ebed7f3e742c17

src/ledger/database/lmdb.zig
dnut: The default is to replace the existing data. 0 just means no flags, so use the default.

```c
	/** @brief Store items into a database.
	 *
	 * This function stores key/data pairs in the database. The default behavior
	 * is to enter the new key/data pair, replacing any previously existing key
	 * if duplicates are disallowed, or adding a duplicate data item if
	 * duplicates are allowed (#MDB_DUPSORT).
	 * @param[in] txn A transaction handle returned by #mdb_txn_begin()
	 * @param[in] dbi A database handle returned by #mdb_dbi_open()
	 * @param[in] key The key to store in the database
	 * @param[in,out] data The data to store
	 * @param[in] flags Special options for this operation. This parameter
	 * must be set to 0 or by bitwise OR'ing together one or more of the
	 * values described here.
	 * <ul>
	 *	<li>#MDB_NODUPDATA - enter the new key/data pair only if it does not
	 *		already appear in the database. This flag may only be specified
	 *		if the database was opened with #MDB_DUPSORT. The function will
	 *		return #MDB_KEYEXIST if the key/data pair already appears in the
	 *		database.
	 *	<li>#MDB_NOOVERWRITE - enter the new key/data pair only if the key
	 *		does not already appear in the database. The function will return
	 *		#MDB_KEYEXIST if the key already appears in the database, even if
	 *		the database supports duplicates (#MDB_DUPSORT). The \b data
	 *		parameter will be set to point to the existing item.
	 *	<li>#MDB_RESERVE - reserve space for data of the given size, but
	 *		don't copy the given data. Instead, return a pointer to the
	 *		reserved space, which the caller can fill in later - before
	 *		the next update operation or the transaction ends. This saves
	 *		an extra memcpy if the data is being generated later.
	 *		LMDB does nothing else with this memory, the caller is expected
	 *		to modify all of the space requested. This flag must not be
	 *		specified if the database was opened with #MDB_DUPSORT.
	 *	<li>#MDB_APPEND - append the given key/data pair to the end of the
	 *		database. This option allows fast bulk loading when keys are
	 *		already known to be in the correct order. Loading unsorted keys
	 *		with this flag will cause a #MDB_KEYEXIST error.
	 *	<li>#MDB_APPENDDUP - as above, but for sorted dup data.
	 * </ul>
	 * @return A non-zero error value on failure and 0 on success. Some possible
	 * errors are:
	 * <ul>
	 *	<li>#MDB_MAP_FULL - the database is full, see #mdb_env_set_mapsize().
	 *	<li>#MDB_TXN_FULL - the transaction has too many dirty pages.
	 *	<li>EACCES - an attempt was made to write in a read-only transaction.
	 *	<li>EINVAL - an invalid parameter was specified.
	 * </ul>
	 */
int  mdb_put(MDB_txn *txn, MDB_dbi dbi, MDB_val *key, MDB_val *data,
			    unsigned int flags);
```

https://github.com/LMDB/lmdb/blob/da9aeda08c3ff710a0d47d61a079f5a905b0a10a/libraries/liblmdb/lmdb.h#L1335-L1383

src/ledger/database/lmdb.zig
dnut: Changing this to require a mutable pointer without adding any extra allocations is doable but awkward. It requires some extra complexity to distinguish between input types, which would need to be mutable, and output types, which should probably be immutable.

It's strange to require callers to always pass mutable pointers into the database, even though it will never be mutated. It might encourage callers to allocate dupes before passing it in, which would be pointless.

src/ledger/database/hashmap.zig
```diff
@@ -178,9 +180,17 @@ pub fn SharedHashMapDB(comptime column_families: []const ColumnFamily) type {
...
                     },
-                    .delete_range => {
-                        // TODO: also add to database tests
-                        @panic("not implemented");
+                    .delete_range => |delete_range_ix| {
+                        const cf_index, const start, const end = delete_range_ix;
+                        const keys, _ = self.maps[cf_index].map.range(start, end);
+                        const to_delete = try batch.allocator.alloc([]const u8, keys.len);
+                        defer batch.allocator.free(to_delete);
+                        for (keys, 0..) |key, i| {
```
dadepo, line 188: Not sure I understand the need of first filling the `to_delete` array first, to then iterate over it, retrieving the keys and using that as the `delete_key` in

```
self.maps[cf_index].delete(self.allocator, delete_key);
```

Any reason for not directly using the key from `keys` as the `delete_key`?

src/ledger/database/hashmap.zig
```diff
@@ -351,12 +364,12 @@ pub fn SharedHashMapDB(comptime column_families: []const ColumnFamily) type {
 
                 pub fn nextKey(self: *@This()) anyerror!?cf.Key {
                     const index = self.nextIndex() orelse return null;
-                    return key_serializer.deserialize(cf.Key, self.allocator, self.keys[index]);
+                    return try key_serializer.deserialize(cf.Key, self.allocator, self.keys[index]);
```
dadepo, line 367: Given that the value gotten from `try` is not used within the function but returned immediately, is there a functional difference/advantage over how it was before without the `try`

src/ledger/database/interface.zig
```diff
@@ -536,5 +540,178 @@ pub fn testDatabase(comptime Impl: fn ([]const ColumnFamily) type) type {
...
+            defer db.deinit();
+
+            try db.put(cf1, 40, .{ .hello = 44 });
+            try db.put(cf1, 10, .{ .hello = 111 });
+            try db.put(cf1, 30, .{ .hello = 33 });
+            try db.put(cf1, 20, .{ .hello = 222 });
+
+            var batch = try db.initWriteBatch();
+            defer batch.deinit();
+            try batch.deleteRange(cf1, 0, 100);
```
dadepo, line 708: Nit: if there is a bug in `deleteRange` where it deletes everything instead of what's in between start and end index, this test won't catch this. Having `try batch.deleteRange(cf1, 20, 35);` might be better to confirm 20 and 30 were deleted but 10 and 40 were not

src/ledger/database/lmdb.zig
```diff
@@ -0,0 +1,641 @@
...
+
+const BytesRef = database.interface.BytesRef;
+const ColumnFamily = database.interface.ColumnFamily;
+const IteratorDirection = database.interface.IteratorDirection;
+const Logger = sig.trace.Logger;
+
+const key_serializer = database.interface.key_serializer;
+const value_serializer = database.interface.value_serializer;
+
+pub fn LMDB(comptime column_families: []const ColumnFamily) type {
```
dadepo, line 17: The RocksDB implementation has the logger as part of the struct's states. Any reason not to have same for the `LMDB`? 

Also, the logger can be scoped.

src/ledger/database/lmdb.zig
```diff
@@ -0,0 +1,641 @@
...
+            return .{
+                .allocator = allocator,
+                .env = env,
+                .dbis = dbis,
+                .path = owned_path,
+            };
+        }
+
+        pub fn deinit(self: *Self) void {
+            self.allocator.free(self.dbis);
```
dadepo, line 61: More of a question. There is also `self.env`, which is a pointer. What takes care of freeing that?

src/ledger/database/hashmap.zig
dnut: Good catch. I was wondering if someone would comment on this.

The problem is that it's not safe to mutate the hash map while iterating over a pointer to its data. This loop is iterating over the data returned by `ArrayHashMap.keys`.

```zig
/// Returns the backing array of keys in this map. Modifying the map may
/// invalidate this array. Modifying this array in a way that changes
/// key hashes or key equality puts the map into an unusable state until
/// `reIndex` is called.
pub fn keys(self: Self) []K {
    return self.unmanaged.keys();
}
```

src/ledger/database/hashmap.zig
dnut: Without try, this won't compile:
```zig
src/ledger/database/hashmap.zig:367:54: error: expected type 'anyerror!?ledger.schema.schema.address_signatures__struct_16305', found '@typeInfo(@typeInfo(@TypeOf(ledger.database.interface.serializer(.big).deserialize__anon_50347)).Fn.return_type.?).ErrorUnion.error_set!ledger.schema.schema.address_signatures__struct_16305'
                    return key_serializer.deserialize(cf.Key, self.allocator, self.keys[index]);
                           ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/ledger/database/hashmap.zig:367:54: note: error union payload 'ledger.schema.schema.address_signatures__struct_16305' cannot cast into error union payload '?ledger.schema.schema.address_signatures__struct_16305'
src/ledger/schema.zig:87:16: note: struct declared here
        .Key = struct {
               ^~~~~~
src/ledger/database/hashmap.zig:365:56: note: function return type declared here
                pub fn nextKey(self: *@This()) anyerror!?cf.Key {
                                               ~~~~~~~~^~~~~~~~
```

Before this PR, this code was unused.

src/ledger/database/lmdb.zig
dnut: The only reason rocksdb gets a logger is to communicate its errors through strings. Any time it returns an error, it also logs an error message, so the cause of the error is clear. LMDB only communicates errors through various error codes, which can be represented with zig errors.

src/ledger/database/hashmap.zig
```diff
@@ -442,5 +455,7 @@ const SharedHashMap = struct {
 };
 
 comptime {
-    _ = &database.interface.testDatabase(SharedHashMapDB);
+    if (build_options.blockstore_db == .hashmap) {
```
Rexicon226, line 458: When testing, we want to test all implementations. Ditto for the other 2 places where you do this.

src/ledger/database/lmdb.zig
dnut: Yeah it should be closed here.

Adding this also revealed another bug, which I fixed in the same commit. Oddly, aborting write transactions seems to lead to memory corruption, which is detected as a double free when closing the env.

2b7c2da5a67994ff946d6ba1822f03bf3659a20c

src/ledger/database/hashmap.zig
dnut: I made the hashmap tests unconditional, since they don't have any external dependencies: a283dcce07d1617aabf46810a1329ed753a9f247. But I haven't done it for rocksdb or lmdb.

Running all the tests was my initial intent, but I changed my mind because it's not actually possible to run all the tests in a single build.

Sure, you could run the tests in testDatabase for each database. But you can't run all the ledger tests for every database, unless you rebuild it separately for each database and execute separate test runs.

If you want to truly properly test all the databases, there is no avoiding it. You need to run three separate commands. I added these to CI to ensure they all pass before a merge.
```
zig build test
zig build test -Dblockstore-db=lmdb -Dfilter=ledger
zig build test -Dblockstore-db=hashmapdb -Dfilter=ledger
```

At that point, why force people to build every database when running tests? Sure it gets a little more test coverage, but it's still not sufficient, and it wastes time building extra dependencies.

Rexicon226: Hmm, I didn't consider that the ledger tests. 

I'm convinced, the CI checks should be sufficient, and it's not very likely that there would be a bug coming from a specific backend.

Rexicon226:
Note that the "check_style" step that isn't "passing", is because #360 needs to be merged first. 
I changed the CI step name, and Github's branch protection rules work on specific names.

dnut:
I split this up into two stacked changesets that are now represented in two prs: #372 and #373

PR #353, opened 2024-11-02
test(ledger): cleanup database testing

Rexicon226:
The previous strategy was... interesting, to put it mildly. It should be noted that putting `test` blocks inside of generic structs, *is* allowed. 

Also removed some usages of `catch unreachable` that actually were reachable.

### DIFF
--- a/src/core/signature.zig
+++ b/src/core/signature.zig
@@ -28,19 +28,25 @@ pub const Signature = struct {
         return .{ .data = try base58.decode(str) };
     }
 
-    pub fn verify(self: Self, pubkey: Pubkey, msg: []const u8) bool {
+    pub fn verify(
+        self: Self,
+        pubkey: Pubkey,
+        msg: []const u8,
+    ) e.NonCanonicalError!bool {
         const signature = Ed25519.Signature.fromBytes(self.data);
-        signature.verify(msg, Ed25519.PublicKey.fromBytes(pubkey.data) catch unreachable) catch
-            return false;
+        const byte_pubkey = try Ed25519.PublicKey.fromBytes(pubkey.data);
+        signature.verify(msg, byte_pubkey) catch return false;
         return true;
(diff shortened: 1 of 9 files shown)


src/core/signature.zig
```diff
@@ -28,19 +28,25 @@ pub const Signature = struct {
         return .{ .data = try base58.decode(str) };
     }
 
-    pub fn verify(self: Self, pubkey: Pubkey, msg: []const u8) bool {
+    pub fn verify(
+        self: Self,
+        pubkey: Pubkey,
+        msg: []const u8,
+    ) e.NonCanonicalError!bool {
```
dadepo, line 35: Should there be a consensus on when to use horizontal/vertical formatting? I have seen instances where horizontal formatting has been updated to vertical formatting (and vice versa) in some PRs.

Rexicon226: I format it vertically when the line takes up more than ~100 characters since it goes off my screen.
We could maybe come to a maximum number of characters per row as part of our style guide?

dnut: I also aim for 100 characters and would like it in the style guide. I have a slight hesitation because there are some exceptions to this rule, but they are very rare.

dnut (approved):

src/core/signature.zig
Rexicon226: I agree. Could we check this in CI along with unused imports?

dnut: I would not be opposed to that.

ultd: IMO we should always vertical align function args for readability unless it's < 2/3 args

Rexicon226: I think the official guide should be “format it vertically or as needed in order for each line to take less than 100 characters”, and after that we can just go depending on what looks better.

PR #369, opened 2024-11-08
feat(rpc,accountsdb): RPC Server for snapshot propagation

### LINKED ISSUE #327: Snapshot Propagation
This issue can be closed once we are able to successfully and efficiently respond to requests for our snapshots through an RPC socket.

Right now, this is the basic outline for the steps to take:
- [x] Report an RPC socket in our contact info in gossip.
- [x] Set up logic to report the latest generated snapshots in AccountsDB to gossip.
- [x] Make gossip propagate our RPC socket & up-to-date snapshot infos.
- [ ] Set up http server to listen on our reported RPC socket.
  - Note: the server will simply be listening on and responding to simple HTTP requests, the actual RPC request handling can be factored in at a later date.

InKryption:
Closes #327.

### DIFF
--- a/src/accountsdb/db.zig
+++ b/src/accountsdb/db.zig
@@ -1012,7 +1012,8 @@ pub const AccountsDB = struct {
     };
 
     /// Validates accountsdb against some snapshot info - if used, it must
-    /// be after loading the snapshot(s) whose information is supplied.
+    /// be after loading the snapshot(s) whose information is supplied,
+    /// and before mutating accountsdb.
     pub fn validateLoadFromSnapshot(
         self: *Self,
         params: ValidateLoadFromSnapshotParams,
@@ -1052,6 +1053,11 @@ pub const AccountsDB = struct {
         }
 
         if (maybe_latest_snapshot_info.*) |latest_snapshot_info| {
+            // ASSERTION: nothing has changed if we previously successfully
+            // verified a load from the snapshot; ie, calling this function
+            // after calling it once should not produce any mutations.
+            // The assertion may also trip if any mutations to accountsdb
+            // have occurred since the first call to this function.
             std.debug.assert(latest_snapshot_info.full.slot == params.full_slot);
             std.debug.assert(latest_snapshot_info.full.hash.eql(accounts_hash));
             std.debug.assert(latest_snapshot_info.full.capitalization == total_lamports);
@@ -1095,6 +1101,8 @@ pub const AccountsDB = struct {
                 return error.IncorrectAccountsDeltaHash;
             }
 
+            // ASSERTION: same idea as the previous assertion, but applied to
+            // the incremental snapshot info.
             if (p_maybe_first_inc.*) |first_inc| {
                 std.debug.assert(first_inc.slot == inc_slot);
                 std.debug.assert(first_inc.hash.eql(accounts_delta_hash));
--- a/src/accountsdb/db.zig
+++ b/src/accountsdb/db.zig
@@ -4275,7 +4275,7 @@ test "generate snapshot & update gossip snapshot hashes" {
 
     // pretend `all_snapshot_fields`/`snap_files` refers to `tmp_snap_dir`, even though the archive file isn't actually in there, just the unpacked contents.
     // TODO: this is not nice, make sure the API for loading from archives outside of the snapshot dir is improved.
-    _ = try accounts_db.loadWithDefaults(allocator, &all_snapshot_fields, 1, true, 1500, false, false);
+    _ = try accounts_db.loadWithDefaults(allocator, &all_snapshot_fields, 1, true, 300, false, false);
 
     var bank_fields = try BankFields.initRandom(allocator, random, 128);
     defer bank_fields.deinit(allocator);
--- a/src/accountsdb/db.zig
+++ b/src/accountsdb/db.zig
@@ -130,6 +130,7 @@ pub const AccountsDB = struct {
     /// Used to potentially skip the first `computeAccountHashesAndLamports`.
     first_snapshot_load_info: RwMux(?SnapshotGenerationInfo),
     /// Represents the largest slot info used to generate a full snapshot, and optionally an incremental snapshot relative to it, which currently exists.
+    /// It also protects access to the snapshot archive files it refers to - as in, the caller who has a lock on this has a lock on the snapshot archives.
     latest_snapshot_gen_info: RwMux(?SnapshotGenerationInfo),
 
     // TODO: populate this during snapshot load
@@ -1016,6 +1017,9 @@ pub const AccountsDB = struct {
         self: *Self,
         params: ValidateLoadFromSnapshotParams,
     ) !void {
+        const maybe_latest_snapshot_info: *?SnapshotGenerationInfo, var latest_snapshot_info_lg = self.latest_snapshot_gen_info.writeWithLock();
+        defer latest_snapshot_info_lg.unlock();
+
         const maybe_first_snapshot_info: *?SnapshotGenerationInfo, var first_snapshot_info_lg = self.first_snapshot_load_info.writeWithLock();
         defer first_snapshot_info_lg.unlock();
 
@@ -1047,6 +1051,11 @@ pub const AccountsDB = struct {
             return error.IncorrectTotalLamports;
         }
 
+        if (maybe_latest_snapshot_info.*) |latest_snapshot_info| {
+            std.debug.assert(latest_snapshot_info.full.slot == params.full_slot);
+            std.debug.assert(latest_snapshot_info.full.hash.eql(accounts_hash));
+            std.debug.assert(latest_snapshot_info.full.capitalization == total_lamports);
+        }
         maybe_first_snapshot_info.* = .{
             .full = .{
                 .slot = params.full_slot,
@@ -1086,12 +1095,19 @@ pub const AccountsDB = struct {
                 return error.IncorrectAccountsDeltaHash;
             }
 
+            if (p_maybe_first_inc.*) |first_inc| {
+                std.debug.assert(first_inc.slot == inc_slot);
+                std.debug.assert(first_inc.hash.eql(accounts_delta_hash));
+                std.debug.assert(first_inc.capitalization == incremental_lamports);
+            }
             p_maybe_first_inc.* = .{
                 .slot = inc_slot,
                 .hash = accounts_delta_hash,
                 .capitalization = incremental_lamports,
             };
         }
+
+        maybe_latest_snapshot_info.* = maybe_first_snapshot_info.*;
     }
 
     /// multithread entrypoint for getHashesFromIndex
--- a/src/cmd/cmd.zig
+++ b/src/cmd/cmd.zig
@@ -687,7 +687,7 @@ fn identity() !void {
     defer logger.deinit();
     const keypair = try getOrInitIdentity(gpa_allocator, logger);
     const pubkey = Pubkey.fromPublicKey(&keypair.public_key);
-    try std.io.getStdOut().writer().print("Identity: {s}\n", .{pubkey.string().constSlice()});
+    try std.io.getStdOut().writer().print("Identity: {s}\n", .{pubkey});
 }
 
 /// entrypoint to run only gossip
--- a/src/cmd/cmd.zig
+++ b/src/cmd/cmd.zig
@@ -58,8 +58,6 @@ const gossip_value_gpa_allocator = if (builtin.mode == .Debug)
 else
     std.heap.c_allocator;
 
-const base58Encoder = base58.Encoder.init(.{});
-
 // The identifier for the scoped logger used in this file.
 const LOG_SCOPE = "cmd";
 
@@ -685,19 +683,11 @@ pub fn run() !void {
 
 /// entrypoint to print (and create if NONE) pubkey in ~/.sig/identity.key
 fn identity() !void {
-    var std_logger = try ChannelPrintLogger.init(.{
-        .allocator = gpa_allocator,
-        .max_level = config.current.log_level,
-        .max_buffer = 1 << 20,
-    });
-    defer std_logger.deinit();
-
-    const logger = std_logger.logger();
-
+    const logger = try spawnLogger();
+    defer logger.deinit();
     const keypair = try getOrInitIdentity(gpa_allocator, logger);
-    var pubkey: [50]u8 = undefined;
-    const size = try base58Encoder.encode(&keypair.public_key.toBytes(), &pubkey);
-    try std.io.getStdErr().writer().print("Identity: {s}\n", .{pubkey[0..size]});
+    const pubkey = Pubkey.fromPublicKey(&keypair.public_key);
+    try std.io.getStdOut().writer().print("Identity: {s}\n", .{pubkey.string().constSlice()});
 }
 
 /// entrypoint to run only gossip
--- a/src/rpc/server.zig
+++ b/src/rpc/server.zig
@@ -59,7 +59,7 @@ pub const Server = struct {
     }) InitError!Server {
         var tcp_server = try params.socket_addr.listen(.{
             // NOTE: ideally we would be doing this nonblockingly, however this doesn't work properly on mac,
-            // so for testing purposes we can't teste the `serve` functionality directly.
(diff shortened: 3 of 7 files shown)


src/rpc/server.zig
```diff
@@ -0,0 +1,547 @@
...
+            HandleConnectionTask.initBufferSize(server.read_buffer_size),
+        );
+        errdefer server.allocator.free(hct_buffer);
+
+        const new_hct = try HandleConnectionTask.initBufferAndReceiveHead(
+            hct_buffer,
+            server,
+            conn,
+        );
+        errdefer new_hct.http_server.connection.stream.close();
```
0xNineteen, line 130: would probs be simpler to have this logic in one `init` method -- its a bit confusing since we dont typically setup thread pool tasks like this

src/rpc/server.zig
```diff
@@ -0,0 +1,547 @@
...
+        const ptr: [*]align(@alignOf(HandleConnectionTask)) u8 = @ptrCast(hct);
+        return ptr[0..initBufferSize(read_buffer_size)];
+    }
+
+    fn getReadBuffer(
+        read_buffer_size: usize,
+        hct: *HandleConnectionTask,
+    ) []u8 {
+        return getFullBuffer(read_buffer_size, hct)[@sizeOf(HandleConnectionTask)..];
+    }
```
0xNineteen, line 159: would be good to have init methods first, then these helpers can come after - bit confusing navigating the code with this difference

src/rpc/server.zig
```diff
@@ -0,0 +1,547 @@
...
+        return hct;
+    }
+
+    fn callback(task: *ThreadPool.Task) void {
+        const hct: *HandleConnectionTask = @fieldParentPtr("task", task);
+        const server = hct.server;
+        const logger = server.logger;
+
+        const full_buffer = getFullBuffer(hct.server.read_buffer_size, hct);
+        defer hct.server.allocator.free(full_buffer);
```
0xNineteen, line 189: would be good to add a comment explaining how this free's the underlying task's memory

src/rpc/server.zig
```diff
@@ -0,0 +1,547 @@
...
+
+const AcceptConnectionError = error{
+    ProcessFdQuotaExceeded,
+    SystemFdQuotaExceeded,
+    SystemResources,
+    ProtocolFailure,
+    BlockedByFirewall,
+    NetworkSubsystemFailed,
+} || std.posix.UnexpectedError;
+fn acceptConnection(
```
0xNineteen, line 335:
```suggestion
} || std.posix.UnexpectedError;

fn acceptConnection(
```

src/rpc/server.zig
```diff
@@ -0,0 +1,547 @@
...
+        .logger = logger,
+        .snapshot_dir = snap_dir,
+        .geyser_writer = null,
+        .gossip_view = null,
+        .index_allocation = .ram,
+        .number_of_index_shards = 4,
+        .lru_size = null,
+    });
+    defer accountsdb.deinit();
+    _ = try accountsdb.loadWithDefaults(allocator, &all_snap_fields, 1, true, 1500, false, false);
```
0xNineteen, line 458: 1500 is overkill here, should change to 300

src/cmd/cmd.zig
```diff
@@ -651,19 +649,11 @@ pub fn run() !void {
 
 /// entrypoint to print (and create if NONE) pubkey in ~/.sig/identity.key
 fn identity() !void {
-    var std_logger = try ChannelPrintLogger.init(.{
```
0xNineteen, line 654: do we want to add the RPC server to any command?

0xNineteen (changes requested):
few nits otherwise lgtm

0xNineteen:
also - note the gossip echo server is broken rn (because we have been missing a server implementation) so would want to fix that once this is merged to main: https://github.com/orgs/Syndica/projects/2/views/10?pane=issue&itemId=88750700

src/rpc/server.zig
InKryption: To be clear I was just using the value used by most other sites I saw; will change it to 300.

InKryption: The other instance of this is in [accountsdb itself](https://github.com/Syndica/sig/blob/82a241843c3d1ba27c898d32f2ac27741f9dabe4/src/accountsdb/db.zig#L4198); should I change that too?

src/rpc/server.zig
InKryption: ac09a281826828456304fa84f495fbbbee6d08e7

src/accountsdb/db.zig
```diff
@@ -1047,6 +1051,11 @@ pub const AccountsDB = struct {
             return error.IncorrectTotalLamports;
         }
 
+        if (maybe_latest_snapshot_info.*) |latest_snapshot_info| {
```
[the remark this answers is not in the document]
InKryption: e7e3eb8e55fdcf35222e950176e0824b73ef0419

src/cmd/cmd.zig
```diff
@@ -651,19 +649,11 @@ pub fn run() !void {
...
-    const logger = std_logger.logger();
-
+    const logger = try spawnLogger();
+    defer logger.deinit();
     const keypair = try getOrInitIdentity(gpa_allocator, logger);
-    var pubkey: [50]u8 = undefined;
-    const size = try base58Encoder.encode(&keypair.public_key.toBytes(), &pubkey);
-    try std.io.getStdErr().writer().print("Identity: {s}\n", .{pubkey[0..size]});
+    const pubkey = Pubkey.fromPublicKey(&keypair.public_key);
+    try std.io.getStdOut().writer().print("Identity: {s}\n", .{pubkey.string().constSlice()});
```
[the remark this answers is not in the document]
InKryption: 4acd8db393d9add22877f2b1399bc82d1ed4bbe0

src/cmd/cmd.zig
InKryption: I think I'll rather do this in a follow-up PR, to keep the surface area of this PR small, and so that it can be added in after fixing the snapshot manifest loading/generation (meaning it will actually be reachable and testable).

src/rpc/server.zig
InKryption: ce0c47543da8b6befabc76e4892e50154b0904ec

src/rpc/server.zig
InKryption: In light of ce0c47543da8b6befabc76e4892e50154b0904ec, I think this is now clear just through the code itself, lmk wyt.

src/rpc/server.zig
0xNineteen: yeah should change that instance too - missed that - thanks

src/cmd/cmd.zig
0xNineteen: makes sense - can you add a ticket + link in the comments? then can resolve

src/rpc/server.zig
InKryption: 78fc6dec173897d993bed8c9d82b6808aeaf8b3d
Also changed it in the fuzz testing code

src/cmd/cmd.zig
InKryption: https://github.com/orgs/Syndica/projects/2/views/10?pane=issue&itemId=89375974

0xNineteen (approved):

src/rpc/server.zig
```diff
@@ -0,0 +1,554 @@
...
+        /// The size for the read buffer allocated to every request.
+        /// Clamped to be greater than or equal to `MIN_READ_BUFFER_SIZE`.
+        read_buffer_size: u32,
+        /// The socket address to listen on for incoming HTTP and/or RPC requests.
+        socket_addr: std.net.Address,
+
+        pub const MIN_READ_BUFFER_SIZE = 256;
+    };
+
+    pub const InitError = std.net.Address.ListenError;
```
yewman, line 58: I don't think we need this

src/rpc/server.zig
```diff
@@ -0,0 +1,554 @@
...
+
+    /// The returned result must be pinned to a memory location before calling any methods.
+    pub fn init(params: InitParams) std.net.Address.ListenError!Server {
+        var tcp_server = try params.socket_addr.listen(.{
+            // NOTE: ideally we would be doing this nonblockingly, however this doesn't work properly on mac,
+            // so for testing purposes we can't teste the `serve` functionality directly.
```
yewman, line 64:
```suggestion
            // so for testing purposes we can't test the `serve` functionality directly.
```

src/rpc/server.zig
```diff
@@ -0,0 +1,554 @@
+const std = @import("std");
+const sig = @import("../sig.zig");
+
+const SnapshotGenerationInfo = sig.accounts_db.AccountsDB.SnapshotGenerationInfo;
+const FullSnapshotFileInfo = sig.accounts_db.snapshots.FullSnapshotFileInfo;
+const IncrementalSnapshotFileInfo = sig.accounts_db.snapshots.IncrementalSnapshotFileInfo;
+const ThreadPool = sig.sync.ThreadPool;
+
+const logger_scope = "rpc.Server";
```
yewman, line 9: NIT: I'd prefer constant values defined after type aliases

yewman (commented):
Couple of minor comments otherwise lgtm!

src/rpc/server.zig
0xNineteen: would also be good to keep global constants with SCREAMING_CASE

src/rpc/server.zig
```diff
@@ -0,0 +1,547 @@
...
+    /// Wait group for all currently running tasks, used to wait for
+    /// all of them to finish before deinitializing.
+    wait_group: std.Thread.WaitGroup,
+    thread_pool: *ThreadPool,
+
+    /// Must not be mutated.
+    read_buffer_size: usize,
+    tcp: std.net.Server,
+
+    pub const InitParams = struct {
```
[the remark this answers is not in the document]
InKryption: I'm not hugely attached either way; I default to defining parameter structs separately for the sake of it making it potentially easier for future code to re-use it, but in this case I don't foresee that being an all-too-common use case.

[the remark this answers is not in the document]
InKryption: 8ca27c531d7dc3ad01782ea641fa6bc6160c6e00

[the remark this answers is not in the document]
InKryption: 8ca27c531d7dc3ad01782ea641fa6bc6160c6e00

src/rpc/server.zig
```diff
@@ -0,0 +1,547 @@
...
+        error.SystemResources,
+        error.ProtocolFailure,
+        error.BlockedByFirewall,
+        error.NetworkSubsystemFailed,
+        => |e| return e,
+
+        error.FileDescriptorNotASocket,
+        error.SocketNotListening,
+        error.OperationNotSupported,
+        => unreachable, // Improperly initialized server.
```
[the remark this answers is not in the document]
InKryption: The only circumstance under which these errors ever occur is if you initialize a `Server` directly with bogus values, ie, intentionally misusing the API, which is something that would be pretty easily flagged in review; I don't really think it's worth the cost, but I'll change this if you're set in that position.

src/rpc/server.zig
InKryption: Forgot to use this as the error set for `init`.

InKryption: 4b7beebaff498199724eeb771ca4cd4215f015d8

src/rpc/server.zig
InKryption: e8e8dbb36634fffa7f3cc1c1a2a3d3f856459ab7

InKryption: @0xNineteen are you referring to something other than this `logger_scope` constant?

0xNineteen: nope - lgtm

0xNineteen (dismissed):

src/rpc/server.zig
yewman: I sort of think that given it is only a single error, it is clearer to not alias it. lmk your thoughts

src/rpc/server.zig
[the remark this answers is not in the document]
yewman: Whats the cost? I think we should we should stick to the approach of only using unreachable when a given path is **provably** unreachable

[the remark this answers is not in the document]
yewman: @panic("Improperly initialised server") is also self commenting code

src/rpc/server.zig
InKryption: Well, it isn't a single error, it's an error set, specifically a union of 5 error sets.

yewman: pub const InitError = std.net.Address.ListenError;

yewman: is that not a single error?

src/rpc/server.zig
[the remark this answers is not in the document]
InKryption: The cost is the branch; an unreachable will be compiled out, a panic forces another code path, even if that branch never actually occurs.
"Provably unreachable" is a bit nebulous, since the most literal interpretation just means it can be encoded more simply directly in the type system, eschewing the need for `unreachable`; for all sane intents and purposes, this code is provably unreachable, because it will always be a socket, it will always be listening, and the operation will be supported.

src/rpc/server.zig
yewman: I think this is great
```zig
const AcceptConnectionError = error{
    ProcessFdQuotaExceeded,
    SystemFdQuotaExceeded,
    SystemResources,
    ProtocolFailure,
    BlockedByFirewall,
    NetworkSubsystemFailed,
} || std.posix.UnexpectedError;
```

InKryption: Further, this is a public API, and I would posit that all public APIs should ideally describe their error set.

src/rpc/server.zig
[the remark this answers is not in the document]
yewman: "Provably unreachable" came directly from our chat recently in sig channel, I get that it compiles out the branch but is this actually a considerable cost in this instance? Genuinely curious here

src/rpc/server.zig
yewman: I think this is going a little far and is overly verbose, would you also alias 
```zig
pub fn base58EncodeAlloc(self: Hash, allocator: Allocator) Allocator.Error![]const u8 {
    return base58.encodeAlloc(self.data, allocator);
}
```
to 
```zig
const Base58EncodeError = Allocator.Error;
pub fn base58EncodeAlloc(self: Hash, allocator: Allocator) Base58EncodeError![]const u8 {
    return base58.encodeAlloc(self.data, allocator);
}
```

src/rpc/server.zig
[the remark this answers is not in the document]
InKryption: It just seems like a pointless cost to pay at all, it's comparable to writing `ptr orelse @panic("null pointer")` instead of `ptr.?` after branching on `ptr != null`; if the handle came from some generic source that we ought to guard against carefully, I'd agree with your assessment, but in this case the code is in full control of the handle.

src/rpc/server.zig
InKryption: I think allocation errors are a particular case of convention being sufficient; everyone intuitively understands that if you pass in an allocator, and an error is returned, among those errors will be `std.mem.Allocator.Error`, and can easily reach for it; in this case, and most other similar cases, it is not obvious that `init` would return `std.net.Address.ListenError`.

0xNineteen: I agree with harold on this one tbh - lets not over alias things which dont need to be

src/rpc/server.zig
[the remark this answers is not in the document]
0xNineteen: agreed with harold in this case - if its improperly initialized then we should panic (as the 'unreachable' is actually reachable) given the server is a parameter - this isnt performance sensitive code, one branch is okay imo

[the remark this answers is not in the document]
InKryption: 3761d42d41ee276fdc40dfb74e0499d34ac74249

src/rpc/server.zig
InKryption: 3761d42d41ee276fdc40dfb74e0499d34ac74249

yewman (approved):
lgtm

PR #372, opened 2024-11-08
feat(ledger): user-selectable db, hashmapdb fixes, add tests, safer memory management

dnut:
I broke up the lmdb pr #352 into two prs: this one and #373. This PR contains all the changes from #352 *except* for adding support for lmdb. I stripped out lmdb because it has a bug that I'm not sure how to fix yet. I'd like to get these improvements merged before figuring out lmdb.

-----------------------------------

This adds a build option to customize the database backend, for example:

```bash
zig build -Dblockstore-db=lmdb
```

RocksDB remains the default for now.

For any build, only the necessary dependencies will be compiled. For example, if you specify `-Dblockstore-db=hashmap`, then rocksdb will not be built.

A few bugs were uncovered in the hashmap db when testing the databases more strictly, so I fixed those in here as well.

I also upgrade the rocksdb library because it had an unsound usage of the allocator interface. I made changes in both rocksdb-zig and sig to make this safer.

### DIFF
--- a/src/ledger/database/hashmap.zig
+++ b/src/ledger/database/hashmap.zig
@@ -1,6 +1,7 @@
 const std = @import("std");
 const sig = @import("../../sig.zig");
 const database = @import("lib.zig");
+const build_options = @import("build-options");
 
 const Allocator = std.mem.Allocator;
 const DefaultRwLock = std.Thread.RwLock.DefaultRwLock;
@@ -26,9 +27,10 @@ pub fn SharedHashMapDB(comptime column_families: []const ColumnFamily) type {
 
         pub fn open(
             allocator: Allocator,
-            _: Logger,
+            logger: Logger,
             _: []const u8,
         ) Allocator.Error!Self {
+            logger.info().log("Initializing SharedHashMapDB");
             var maps = try allocator.alloc(SharedHashMap, column_families.len);
             errdefer {
                 for (maps) |*m| m.deinit();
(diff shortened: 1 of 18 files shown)


src/ledger/database/hashmap.zig
```diff
@@ -20,31 +21,36 @@ pub fn SharedHashMapDB(comptime column_families: []const ColumnFamily) type {
         maps: []SharedHashMap,
         /// shared lock is required to call locking map methods.
         /// exclusive lock is required to call non-locking map methods.
-        transaction_lock: DefaultRwLock = .{},
+        /// to avoid deadlocks, always acquire the shared lock *before* acquiring the map lock.
+        transaction_lock: *RwLock,
```
InKryption, line 25: Would be good to understand why we're allocating this field now, any comment?

src/utils/allocators.zig
```diff
@@ -445,6 +445,10 @@ pub const failing = struct {
     }
 };
 
+pub fn noAlloc(_: *anyopaque, _: usize, _: u8, _: usize) ?[*]u8 {
+    return null;
+}
+
```
InKryption, line 451: What's this for? Appears to be dead code, and serves a purpose redundant with `failing.allocator(.{}).vtable.alloc`.

InKryption (commented):
Seems like a solid changeset, just have a couple of questions I'd like answered and then I'd be eager to merge.

src/ledger/database/hashmap.zig
dnut: The database is copied sometimes based on the assumption that it only contains pointers. If this lock is copied then it doesn't work. It needs to always use the same lock.

src/utils/allocators.zig
dnut: I forgot to remove this when I moved away from the approach that relied on it.

dnut: 2b5c2496e7685cb9a8b5eba15047cb6c1cb2aa00

InKryption (approved):

PR #378, opened 2024-11-12
refactor(merkle_tree): general purpose nested list

0xNineteen:
- uses a new general purposed nested list struct which can be used with any `[][]T` type


src/utils/merkle_tree.zig
```diff
@@ -0,0 +1,125 @@
...
+                    const index_in_nested = index - search_index;
+                    return &slice[index_in_nested];
+                } else {
+                    search_index += slice.len;
+                }
+            }
+            unreachable;
+        }
+
+        pub fn getSlice(self: *const Self, index: u64, length: u64) []T {
```
InKryption, line 25: What exactly is the point of this function? I don't see it used anywhere so it's hard to judge what exactly the point is; there's no API for getting the length of the slice that the element at `index` would be in, which makes me question how exactly the `length` should be computed by the caller, or what the usage is in general.

InKryption (commented):
Just one question, otherwise looks solid.

src/utils/merkle_tree.zig
0xNineteen: ah great catch - it was some old code i used previously but dont need anymore

d89bf4a

InKryption (approved):
lgtm

PR #379, opened 2024-11-12
refactor(bincode): file_id config

0xNineteen:


InKryption:
Not a fan of the specific change; I don't think we should be moving support for specific types like FileId into bincode.
I do like the direction though: I think it would make more sense to directly declare the `pub const @"!bincode-config": bincode.FieldConfig(FileId) = .{...};` inside `FileId`.

0xNineteen:
excellent point - completely agree - 56b7def

InKryption (dismissed):
lgtm

src/accountsdb/accounts_file.zig
```diff
@@ -49,6 +54,24 @@ pub const FileId = enum(Int) {
...
+    ) anyerror!void {
+        try sig.bincode.write(writer, @as(usize, data.toInt()), params);
+    }
+
+    fn deserialize(
+        _: std.mem.Allocator,
+        reader: anytype,
+        params: sig.bincode.Params,
+    ) anyerror!FileId {
+        const int = try sig.bincode.readInt(usize, reader, params);
```
Rexicon226, line 71: I'm wondering about this `readInt(usize, ...)`, it doesn't seem correct to me.
Does everything here just assume that you're on a 64-bit system? That seems like a very weak guarantee.

0xNineteen: usize is used in agave which we kind of need to follow to make sure we get the de/serialization right - could probs safely change to a u64 but also probs not worth

Rexicon226: > usize is used in agave which we kind of need to follow to make sure we get the de/serialization right

this doesn't make much sense to me though. is it guaranteed that all nodes are going to be 64-bit machines? if we take an example where a 32-bit client sends some bincode information to a 64-bit client, it will decode it incorrectly.
same question applies with account files, are they always encoded on a 64-bit machine?

if the answer is yes, why not make that immediately obvious when reading the code by using `u64`?
if the answer is no, then it's a bug.

0xNineteen: yeah thats a great point - ea8a517c

Rexicon226 (approved):

PR #380, opened 2024-11-12
feat(bincode): config default values

0xNineteen:
- see doc comment on how this new field is useful
- we also remove dead test code


InKryption (approved):
lgtm

PR #381, opened 2024-11-12
feat(allocator): add recycle buffer

0xNineteen:
- see doc comment on struct to understand its usage
- this new allocator is specifically for the account index and is built with that in mind (vs a more general purpose allocator) 
  - it doesnt even support the .allocator interface because its very niche
- usage and more will be shown in the stack above (#382)

### DIFF
--- a/src/utils/allocators.zig
+++ b/src/utils/allocators.zig
@@ -175,12 +175,12 @@ pub fn RecycleBuffer(comptime T: type, config: struct {
 
         /// frees the unused space of a buf.
         /// this is useful when a buf is initially overallocated and then resized.
-        pub fn tryRecycleUnusedSpace(self: *Self, buf: []T, used_len: u64) !bool {
+        pub fn tryRecycleUnusedSpace(self: *Self, buf_ptr: [*]T, used_len: u64) !bool {
             if (config.thread_safe) self.mux.lock();
             defer if (config.thread_safe) self.mux.unlock();
 
             for (self.records.items) |*record| {
-                if (record.buf.ptr == buf.ptr) {
+                if (record.buf.ptr == buf_ptr) {
                     return self.tryRecycleUnusedSpaceWithRecordUnsafe(record, used_len);
                 }
             }
@@ -717,7 +717,7 @@ test "recycle buffer: freeUnused" {
     defer allocator.free(bytes.ptr);
 
     // free the unused space
-    const did_recycle = try allocator.tryRecycleUnusedSpace(bytes, 50);
+    const did_recycle = try allocator.tryRecycleUnusedSpace(bytes.ptr, 50);
     try std.testing.expectEqual(true, did_recycle);
 
     // this should be ok now
--- a/src/utils/allocators.zig
+++ b/src/utils/allocators.zig
@@ -705,7 +705,7 @@ pub const failing = struct {
     }
 };
 
-test "recycle buffer : save and load" {
+test "recycle buffer: save and load" {
     const backing_allocator = std.testing.allocator;
     const X = struct {
         a: u8,
--- a/src/utils/allocators.zig
+++ b/src/utils/allocators.zig
@@ -77,7 +77,7 @@ pub fn RecycleBuffer(comptime T: type, config: struct {
         }
 
         /// append a block of N elements to the manager
-        pub fn append(self: *Self, n: u64) !void {
+        pub fn append(self: *Self, n: u64) std.mem.Allocator.Error!void {
             if (n == 0) return;
 
             if (config.thread_safe) self.mux.lock();
@@ -99,7 +99,7 @@ pub fn RecycleBuffer(comptime T: type, config: struct {
             self.capacity += buf.len;
         }
 
-        pub fn alloc(self: *Self, n: u64) !struct { []T, u64 } {
+        pub fn alloc(self: *Self, n: u64) error{ AllocTooBig, AllocFailed }!struct { []T, u64 } {
             if (config.thread_safe) self.mux.lock();
             defer if (config.thread_safe) self.mux.unlock();
 
@@ -122,7 +122,7 @@ pub fn RecycleBuffer(comptime T: type, config: struct {
             @panic("not enough memory, and collapse failed max times");
         }
 
-        pub fn allocUnsafe(self: *Self, n: u64) !struct { []T, u64 } {
+        pub fn allocUnsafe(self: *Self, n: u64) error{ AllocTooBig, AllocFailed, CollapseFailed }!struct { []T, u64 } {
             // this would never succeed
             if (n > self.capacity) return error.AllocTooBig;
 
--- a/src/utils/allocators.zig
+++ b/src/utils/allocators.zig
@@ -8,7 +8,7 @@ const sig = @import("../sig.zig");
 /// - record serialization is supported for fast loading this memory back
 /// - the global index is also returned on allocation to help with additional book-keeping
 /// NOTE: T needs to have a .DEFAULT field that is used to zero out the memory allocated in `append()`
-pub fn RecycleBuffer(comptime T: type, config: struct {
+pub fn RecycleBuffer(comptime T: type, default_init: T, config: struct {
     /// If enabled, all operations will require an exclusive lock.
     thread_safe: bool = !builtin.single_threaded,
     max_collapse_tries: u32 = 5,
@@ -89,7 +89,7 @@ pub fn RecycleBuffer(comptime T: type, config: struct {
             const buf = try self.memory_allocator.alloc(T, n);
             // NOTE: we do this here so bincode serialization can work correctly
             // otherwise, we run into undefined memory which breaks bincode
-            @memset(buf, T.DEFAULT);
+            @memset(buf, default_init);
 
             self.records.appendAssumeCapacity(.{
                 .is_free = true,
@@ -102,7 +102,14 @@ pub fn RecycleBuffer(comptime T: type, config: struct {
             self.capacity += buf.len;
         }
 
-        pub fn alloc(self: *Self, n: u64) error{ AllocTooBig, AllocFailed }!struct { []T, u64 } {
+        const AllocError = error{
+            AllocTooBig,
+            AllocFailed,
+            // NOTE: even though this doesnt get hit on `alloc`, zig isnt smart enough to know that
+            CollapseFailed,
+        } || std.mem.Allocator.Error;
+
+        pub fn alloc(self: *Self, n: u64) AllocError!struct { []T, u64 } {
             if (config.thread_safe) self.mux.lock();
             defer if (config.thread_safe) self.mux.unlock();
 
@@ -122,10 +129,10 @@ pub fn RecycleBuffer(comptime T: type, config: struct {
             }
             // not enough memory to allocate and no possible recycles will be perma stuck.
             // though its possible to recover from this, its very unlikely, so we panic
-            @panic("not enough memory, and collapse failed max times");
+            @panic("not enough memory and collapse failed max times");
         }
 
-        pub fn allocUnsafe(self: *Self, n: u64) error{ AllocTooBig, AllocFailed, CollapseFailed }!struct { []T, u64 } {
+        pub fn allocUnsafe(self: *Self, n: u64) AllocError!struct { []T, u64 } {
             // this would never succeed
             if (n > self.capacity) return error.AllocTooBig;
 
@@ -190,7 +197,7 @@ pub fn RecycleBuffer(comptime T: type, config: struct {
             @panic("attempt to recycle invalid buf");
         }
 
-        fn tryRecycleUnusedSpaceWithRecordUnsafe(self: *Self, record: *Record, used_len: u64) !bool {
+        fn tryRecycleUnusedSpaceWithRecordUnsafe(self: *Self, record: *Record, used_len: u64) std.mem.Allocator.Error!bool {
             const unused_len = record.buf.len -| used_len;
             if (unused_len > config.min_split_size) {
                 // update the state of the record
@@ -704,7 +711,7 @@ test "recycle buffer: freeUnused" {
         pub const DEFAULT = .{ .a = 0 };
     };
 
-    var allocator = RecycleBuffer(X, .{
+    var allocator = RecycleBuffer(X, X.DEFAULT, .{
         .min_split_size = 10,
     }).init(.{
         .records_allocator = backing_allocator,
--- a/src/utils/allocators.zig
+++ b/src/utils/allocators.zig
@@ -83,19 +83,22 @@ pub fn RecycleBuffer(comptime T: type, config: struct {
             if (config.thread_safe) self.mux.lock();
             defer if (config.thread_safe) self.mux.unlock();
 
+            try self.records.ensureUnusedCapacity(1);
+            try self.memory.ensureUnusedCapacity(1);
+
             const buf = try self.memory_allocator.alloc(T, n);
             // NOTE: we do this here so bincode serialization can work correctly
             // otherwise, we run into undefined memory which breaks bincode
             @memset(buf, T.DEFAULT);
 
-            try self.records.append(.{
+            self.records.appendAssumeCapacity(.{
                 .is_free = true,
                 .buf = buf,
                 .global_index = self.capacity,
                 .len = buf.len,
                 .memory_index = self.memory.items.len,
             });
-            try self.memory.append(buf);
+            self.memory.appendAssumeCapacity(buf);
             self.capacity += buf.len;
         }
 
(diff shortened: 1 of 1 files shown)


0xNineteen:
* **#381** <a href="https://app.graphite.dev/github/pr/Syndica/sig/381?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @0xNineteen and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

src/utils/allocators.zig
```diff
@@ -2,6 +2,251 @@ const std = @import("std");
...
+                        try new_records.append(record);
+                    }
+                } else {
+                    try new_records.append(record);
+                    last_was_free = false;
+                }
+            }
+            self.records.deinit();
+            self.records = new_records;
+        }
```
InKryption, line 237: If I've understood the logic correctly, I think this would be an improvement:

```suggestion
        pub fn collapse(self: *Self) void {
            const records = &self.records;

            var i: usize = 1;
            while (i < records.items.len) {
                const prev = records.items[i - 1];
                const curr = records.items[i];

                const both_free = prev.is_free and curr.is_free;
                const shared_memory_index = prev.memory_index == curr.memory_index;
                if (both_free and shared_memory_index) {
                    records.items[i - 1].buf.len += curr.buf.len;
                    records.orderedRemove(i);
                } else {
                    i += 1;
                }
            }
        }
```

Related: why is this `pub`? When is it useful for a user to call this manually?

src/utils/allocators.zig
```diff
@@ -2,6 +2,251 @@ const std = @import("std");
...
+            defer if (config.thread_safe) self.mux.unlock();
+
+            const buf = try self.memory_allocator.alloc(T, n);
+            // NOTE: we do this here so bincode serialization can work correctly
+            // otherwise, we run into undefined memory which breaks bincode
+            @memset(buf, T.DEFAULT);
+
+            try self.records.append(.{
+                .is_free = true,
+                .buf = buf,
```
dadepo, line 93: more of a question, i believe the `buf` created on [line 86](https://github.com/Syndica/sig/pull/381/files#diff-28acc4df26271c03a683bc2a56dc0ae9aa1199fb8b2732b93aa247ae89e3eaa6R86) is being copied to the struct here, and also copied in the `try self.memory.append(buf)` call on line 98

So one alloc of `buf` at line 86 will lead to consuming 2x of the space occupied by `buf`. Do I get that right and is that by design?

src/utils/allocators.zig
```diff
@@ -2,6 +2,251 @@ const std = @import("std");
...
+                .global_index = self.capacity,
+                .len = buf.len,
+                .memory_index = self.memory.items.len,
+            });
+            try self.memory.append(buf);
+            self.capacity += buf.len;
+        }
+
+        pub fn alloc(self: *Self, n: u64) !struct { []T, u64 } {
```
dadepo, line 102: Similar to `append`, should this not also have a `if (n == 0) return;` check? if not it looks like it would attempt to allocate 0 sized buf, which is not that useful.

src/utils/allocators.zig
```diff
@@ -2,6 +2,251 @@ const std = @import("std");
...
+            if (n > self.capacity) return error.AllocTooBig;
+
+            // this is used to index directly into the buffer's underlying data
+            var global_index: u64 = 0;
+            // check for a buf to recycle
+            var is_possible_to_recycle = false;
+            for (self.records.items) |*record| {
+                if (record.buf.len >= n) { // allocation is possible
+                    if (record.is_free) {
+                        record.is_free = false;
```
dadepo, line 136: Given `is_free` is set to false here, what happens if `_ = try self.tryRecycleUnusedSpaceUnsafe(record, n)` call on line 138 fails, would the record be left with `is_false` set to false?

src/utils/allocators.zig
```diff
@@ -2,6 +2,251 @@ const std = @import("std");
...
+                        return .{ buf, global_index };
+                    } else {
+                        is_possible_to_recycle = true;
+                    }
+                }
+                global_index += record.buf.len;
+            }
+
+            if (is_possible_to_recycle) {
+                // they can try again later since recycle is possible
```
dadepo, line 148: From the caller's perspective, not sure the error value of `error.AllocFailed` convey the information that allocation can be tried again later.

dadepo: I actually see this is called in a loop within `alloc` method.

Is it possible to have `allocUnsafe` private? Then it won't be possible to call by a client, and the retry is handled by the `alloc` method.

src/utils/allocators.zig
```diff
@@ -2,6 +2,251 @@ const std = @import("std");
...
+
+            for (self.records.items) |*record| {
+                if (record.buf.ptr == buf.ptr) {
+                    return self.tryRecycleUnusedSpaceUnsafe(record, used_len);
+                }
+            }
+            @panic("attempt to recycle invalid buf");
+        }
+
+        pub fn tryRecycleUnusedSpaceUnsafe(self: *Self, record: *Record, used_len: u64) !bool {
```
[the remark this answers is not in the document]
0xNineteen: good point - renamed and made private - ab9c5fa6

src/utils/allocators.zig
```diff
@@ -2,6 +2,251 @@ const std = @import("std");
...
+                    // exit here
+                    return self.allocUnsafe(n);
+                } else {
+                    // up to the caller but should sleep and try collapse/alloc again
+                    return error.CollapseFailed;
+                }
+            }
+        }
+
+        pub fn free(self: *Self, buf_ptr: [*]T) void {
```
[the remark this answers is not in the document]
0xNineteen: the ptr is the minimum required information - changed recycleUnused too f0af00fd

src/utils/allocators.zig
```diff
@@ -2,6 +2,251 @@ const std = @import("std");
...
+                        record.is_free = false;
+                        const buf = record.buf[0..n]; // local copy because next line will likely change the record pointer
+                        _ = try self.tryRecycleUnusedSpaceUnsafe(record, n);
+                        return .{ buf, global_index };
+                    } else {
+                        is_possible_to_recycle = true;
+                    }
+                }
+                global_index += record.buf.len;
+            }
```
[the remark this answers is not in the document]
0xNineteen: yup - `is_possible_to_recycle` is more of a safety check than anything -- when its not possible we try to collapse to make it possible, but if that fails then we panic because the allocation will never be able to succeed

src/utils/allocators.zig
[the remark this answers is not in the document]
0xNineteen: 658fc6d2

src/utils/allocators.zig
```diff
@@ -2,6 +2,251 @@ const std = @import("std");
...
+
+            for (self.memory.items) |block| {
+                self.memory_allocator.free(block);
+            }
+            self.memory.deinit();
+            self.records.deinit();
+        }
+
+        /// append a block of N elements to the manager
+        pub fn append(self: *Self, n: u64) !void {
```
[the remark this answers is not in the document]
0xNineteen: 658fc6d2

src/utils/allocators.zig
```diff
@@ -2,6 +2,251 @@ const std = @import("std");
...
+            @memset(buf, T.DEFAULT);
+
+            try self.records.append(.{
+                .is_free = true,
+                .buf = buf,
+                .global_index = self.capacity,
+                .len = buf.len,
+                .memory_index = self.memory.items.len,
+            });
+            try self.memory.append(buf);
```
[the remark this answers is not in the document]
0xNineteen: e1ad2781

src/utils/allocators.zig
```diff
@@ -2,6 +2,251 @@ const std = @import("std");
...
+        pub fn append(self: *Self, n: u64) !void {
+            if (n == 0) return;
+
+            if (config.thread_safe) self.mux.lock();
+            defer if (config.thread_safe) self.mux.unlock();
+
+            const buf = try self.memory_allocator.alloc(T, n);
+            // NOTE: we do this here so bincode serialization can work correctly
+            // otherwise, we run into undefined memory which breaks bincode
+            @memset(buf, T.DEFAULT);
```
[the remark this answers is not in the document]
0xNineteen: 0441e213

src/utils/allocators.zig
```diff
@@ -2,6 +2,251 @@ const std = @import("std");
...
+        capacity: u64,
+        // for thread safety
+        mux: std.Thread.Mutex = .{},
+
+        // NOTE: we use the global_index to support fast loading the state
+        pub const Record = struct {
+            is_free: bool,
+            buf: []T,
+            global_index: u64,
+            len: u64,
```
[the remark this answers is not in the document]
0xNineteen: len is actually used for save/load which is used in another substack - fixed the len bug and added a test to show this in this branch - e7e11868

src/utils/allocators.zig
0xNineteen: v nice - e3268b72

src/utils/allocators.zig
```diff
@@ -2,6 +2,251 @@ const std = @import("std");
...
+    std.debug.assert(config.min_split_size > 0);
+
+    return struct {
+        // records are used to keep track of the memory blocks
+        records: std.ArrayList(Record),
+        // memory holds blocks of memory ([]T) that can be allocated/deallocated
+        memory: std.ArrayList([]T),
+        // allocator used to alloc the memory blocks
+        memory_allocator: std.mem.Allocator,
+        capacity: u64,
```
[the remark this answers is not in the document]
0xNineteen: d6c9bd91

src/utils/allocators.zig
0xNineteen: yeah we store the buf in multiple places - but is just a pointer and a len which is very cheap to store (two u64 values) so its okay

src/utils/allocators.zig
0xNineteen: 32f5d8a0

src/utils/allocators.zig
0xNineteen: 32f5d8a0

src/utils/allocators.zig
0xNineteen: added a doc comment to explain it better - 246a8db3 - would pref to keep that method public because so that the CollapseFailed case can be handled specifically by the user (in alloc we just panic but more advanced things could be done to save it)

src/utils/allocators.zig
InKryption: Still would like to know, when or why is it useful for a user to call `collapse`?

0xNineteen: if they see theres many small records and they want to call collapse to increase the performance, they can call it themselves - in general, im pro public methods to let people do what they want to do

InKryption: I'm moreso pro public data, public methods as-needed. Publicizing methods in an API force you to freeze the behavior of that method, or otherwise risk silently breaking code by changing the contract in unexpected ways. If this is going to be a simple public method, I think this should at the very least be divided into `collapseUnsafe` and `collapse` like other public methods.

InKryption (dismissed):
lgtm

src/utils/allocators.zig
```diff
@@ -2,6 +2,269 @@ const std = @import("std");
 const builtin = @import("builtin");
 const sig = @import("../sig.zig");
 
+/// very similar to RecycleFBA but with a few differences:
+/// - this uses an explicit T type and only returns slices of that type (instead of a generic u8)
+/// - additional memory blocks are supported (instead of using a fixed-buffer-allocator approach)
+/// - record serialization is supported for fast loading this memory back
+/// - the global index is also returned on allocation to help with additional book-keeping
+/// NOTE: T needs to have a .DEFAULT field that is used to zero out the memory allocated in `append()`
```
dadepo, line 10: I think with the api updated to take `default_init`, this comment about T needing to have a .DEFAULT is now outdated.

src/utils/allocators.zig
```diff
@@ -2,6 +2,269 @@ const std = @import("std");
...
+
+            for (self.memory.items) |block| {
+                self.memory_allocator.free(block);
+            }
+            self.memory.deinit();
+            self.records.deinit();
+        }
+
+        /// append a block of N elements to the manager
+        pub fn append(self: *Self, n: u64) std.mem.Allocator.Error!void {
```
dadepo, line 86: nit: `append` is mostly used to add actual elements. This seems to be expanding the capacity of the memory that can be later allocated. Maybe similar to zig's std have name that include `ensure*capacity` or `expand*capacity`

src/utils/allocators.zig
```diff
@@ -2,6 +2,269 @@ const std = @import("std");
...
+                if (both_free and shared_memory_index) {
+                    records.items[i - 1].buf.len += curr.buf.len;
+                    _ = records.orderedRemove(i);
+                } else {
+                    i += 1;
+                }
+            }
+        }
+
+        pub fn isPossibleToAllocateUnsafe(self: *Self, n: u64) bool {
```
dadepo, line 257: Similar to the other comment about the `Unsafe` suffix, not clear why this method is considered unsafe.

src/utils/allocators.zig
```diff
@@ -2,6 +2,269 @@ const std = @import("std");
...
+            if (n == 0) return error.AttemptedZeroAlloc;
+
+            for (0..config.max_collapse_tries) |_| {
+                return self.allocUnsafe(n) catch |err| {
+                    switch (err) {
+                        error.CollapseFailed => {
+                            if (config.thread_safe) self.mux.unlock();
+                            defer if (config.thread_safe) self.mux.lock();
+                            std.time.sleep(std.time.ns_per_ms * config.collapse_sleep_ms);
+                            // try again -- will hit collapse again
```
dadepo, line 132: for my understanding, looking at the implementation of `collapseUnsafe` I do not see what will make it succeed on retry so not sure why the logic here is to retry after some time

src/utils/allocators.zig
```diff
@@ -2,6 +2,269 @@ const std = @import("std");
...
+/// very similar to RecycleFBA but with a few differences:
+/// - this uses an explicit T type and only returns slices of that type (instead of a generic u8)
+/// - additional memory blocks are supported (instead of using a fixed-buffer-allocator approach)
+/// - record serialization is supported for fast loading this memory back
+/// - the global index is also returned on allocation to help with additional book-keeping
+/// NOTE: T needs to have a .DEFAULT field that is used to zero out the memory allocated in `append()`
+///
+/// when `alloc` returns error.AllocFailed the allocation failed and the caller should try again
+/// (after some records/buffers have been free'd/recycle'd).
+/// If the allocation size being too to be recycled after trying to collapse smaller records
```
dadepo, line 14: Looks like a typo here
```suggestion
/// If the allocation size being too big to be recycled after trying to collapse smaller records
```

dadepo (dismissed):

src/utils/allocators.zig
```diff
@@ -2,6 +2,269 @@ const std = @import("std");
...
+                        else => return err,
+                    }
+                };
+            }
+            // not enough memory to allocate and no possible recycles will be perma stuck.
+            // though its possible to recover from this, its very unlikely, so we panic
+            @panic("not enough memory and collapse failed max times");
+        }
+
+        pub fn allocUnsafe(self: *Self, n: u64) AllocError!struct { []T, u64 } {
```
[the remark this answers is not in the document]
dadepo: That makes sense.

src/utils/allocators.zig
dadepo: Seems to be about thread safety. https://github.com/Syndica/sig/pull/381#discussion_r1846755166

src/utils/allocators.zig
0xNineteen: if collapse fails then there wasnt enough free records which could collapse to support the size requested - if we wait some time and try to collapse again there may be _new_ free records (which were free'd by some other consumer thread) which can be collapsed and have the alloc call succeed

src/utils/allocators.zig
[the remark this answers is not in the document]
0xNineteen: completely agree - 0173e4d8

src/utils/allocators.zig
0xNineteen: added a comment - 0173e4d8

src/utils/allocators.zig
0xNineteen: 0173e4d8

dadepo (approved):

PR #385, opened 2024-11-13
Harnew/for-review

yewman:


yewman (commented):
Checked it out looks okay I think.

src/gossip/fuzz_service.zig
```diff
@@ -63,11 +63,22 @@ pub fn randomPongPacket(random: std.rand.Random, keypair: *const KeyPair, to_add
...
+    value.data.LegacyContactInfo = LegacyContactInfo.default(try Pubkey.fromPublicKey(&keypair.public_key));
+    try value.sign(&keypair);
+
+    const should_pass_sig_verification = maybe_should_pass_sig_verification orelse random.boolean();
+    if (should_pass_sig_verification) {
+        value.data.setId(pubkey);
+        try value.sign(&keypair);
+    }
+
+    return value;
```
InKryption, line 81: Wrt to the argument against the `sign` method, this, for example, seems like a downgrade compared to the original code: it doesn't even generated random data, it overwrites the randomly generated data with the default value, and redundantly re-sets the id.

InKryption (changes requested):
A few comments from an initial peruse of the changes.

0xNineteen:
fyi this is just an example branch which has everything in one branch - not for merge - will be split up into multiple branches which will be easier to review

PR #387, opened 2024-11-14
feat(channel): add blocking receive methods

dnut:
This adds two new functions `receive` and `receiveTimeout` to Channel to enable threads to patiently wait on an empty channel using a std.Thread.Condition instead of busy spinning. The existing `receive` function is renamed to `tryReceive`. It is still available for use cases that would benefit from it. I have not changed any existing usages to switch away from tryReceive.

To prevent these functions from blocking indefinitely during shutdown, a `close` method has also been added to Channel, which sets `closed` to true, and broadcasts to all waiting threads to wake up.


src/sync/channel.zig
```diff
@@ -153,12 +163,61 @@ pub fn Channel(T: type) type {
                     // Release the exclusive lock on the slot's value, which allows a consumer
                     // to read the data we've just assigned.
                     _ = slot.state.fetchOr(WRITTEN_TO, .release);
+                    channel.condition.signal();
```
Rexicon226, line 166: There is no predictable scheduling ordering between `signal` and `wait`.

> however, if predictable scheduling
       behavior is required, then that mutex shall be locked by the
       thread calling pthread_cond_broadcast() or pthread_cond_signal().

Locking the mutex ensures the other side sees the operation.

dnut: 6966c533dc25412a443b1f9de21ff9e196b57a82

Rexicon226:
Let's make the blocking methods and non-blocking methods two separate benchmarks. 
The blocking ones will be slower, so they're measuring two different things.

dnut:
> Let's make the blocking methods and non-blocking methods two separate benchmarks. The blocking ones will be slower, so they're measuring two different things.

What exactly are you proposing here? The benchmarks only test the non-blocking methods. They work the same as they were before. I haven't added benchmarks for the blocking methods.

Rexicon226:
> > Let's make the blocking methods and non-blocking methods two separate benchmarks. The blocking ones will be slower, so they're measuring two different things.
> 
> What exactly are you proposing here? The benchmarks only test the non-blocking methods. They work the same as they were before. I haven't added benchmarks for the blocking methods.

Ah, I missed the change in `testUsizeSender`. I think the Github Actions can just be run on machines of different power then. Maybe that's something that should be considered.

0xNineteen (approved):
lgtm :shipit:

Rexicon226:
I'm hoping to hold off on this PR for now. I'd like King to evaluate the situation and give some comments beforehand.

0xNineteen:
would rather merge now since i think its good and fixes a current problem - can have king check it out after

PR #395, opened 2024-11-19
feat(ci): save benchmark results

0xNineteen:
three main scripts are used
- `scripts/collect_benchmarks.sh` is periodically called using a cron job to run the benchmarks on new git commits
- `scripts/benchmark_server.py` is run as a server to visualize the results over time on port 8050
- `.github/workflows/check.yml` checks for performance regressions + comments on PRs which regress

the results are a server which re-loads the results on each page-refresh - http://18.226.150.130:8050/


0xNineteen:
^ https://github.com/Syndica/sig/actions/runs/11959247019/job/33340615173?pr=395

src/accountsdb/swiss_map.zig
```diff
@@ -624,22 +624,22 @@ fn generateData(allocator: std.mem.Allocator, n_accounts: usize) !struct {
...
     pub const args = [_]BenchArgs{
-        BenchArgs{
-            .n_accounts = 100_000,
-            .name = "100k accounts",
-        },
-        BenchArgs{
-            .n_accounts = 500_000,
-            .name = "500k accounts",
-        },
+        // BenchArgs{
```
dadepo, line 635: Can it be removed instead? if not, can there be a comment explaining why it is commented out?

dadepo (approved):

src/benchmarks.zig
```diff
@@ -565,3 +649,50 @@ pub fn benchmarkCSV(
...
     }
 }
+
+const Metric = struct {
+    name: []const u8,
+    unit: []const u8,
+    value: u64,
+    // used to deallocate the metric name which
+    // may be on the heap due to runtime formatting
+    maybe_allocator: ?std.mem.Allocator = null,
```
dadepo, line 659: More of a question: All the instances of `Metric` had an allocator set, so is there a need to have this field as optional?

src/benchmarks.zig
```diff
@@ -424,6 +473,21 @@ pub fn benchmarkCSV(
...
+                    // collect the metric
+                    if (maybe_metrics.*) |*metrics| {
+                        const fmt_size = std.fmt.count("{s}({s})", .{ def.name, arg_name });
+                        const name_buf = try allocator.alloc(u8, fmt_size);
+                        const name = try std.fmt.bufPrint(name_buf, "{s}({s})", .{ def.name, arg_name });
+
+                        const metric = Metric{
+                            .name = name,
+                            .unit = time_unit.toString(),
+                            .value = mean,
```
dadepo, line 486: Question: Any reason why the mean is set directly here, without the processing that is applied in the float case as done in here https://github.com/Syndica/sig/pull/395/files#diff-0fb4beeff922a420b2b5b7caff8f618005553add01a26cd17fe65bf29bb5b213R544

src/accountsdb/swiss_map.zig
0xNineteen: 09842663

src/benchmarks.zig
0xNineteen: in this case the return type is a Duration, which cant be a float so we know that mean will be a u64 and require no conversion -- returning multiple values supports floats which requires the additional work

src/benchmarks.zig
0xNineteen: true - 85c037de

dadepo (approved):

Rexicon226:
A couple of questions on things that are unclear to me.

1. Why are we running a corn job for benchmarks if we could just have an action that runs on commit to master? Are these benchmarks somehow different from the ones in the actions?
2. Uploading the file to the cache from open PRs is incorrect. We should only be storing the results of the master branch, since who knows what a PR could be doing. This ties back to the first question, why not just have an action that runs only on new commits to master and does the uploading, where the PR action reads that file and checks against it?

0xNineteen:
> Why are we running a cron job for benchmarks if we could just have an action that runs on commit to master? Are these benchmarks somehow different from the ones in the actions?

yes these benchmarks are different, some benchmarks (like downloading + validating a snapshot) require more hardware than whats available on CI which we will need to support

> Uploading the file to the cache from open PRs is incorrect. ...

i agree thats the best approach - i assumed the CI action was doing for us, but maybe its not - let me look into how we can do this

Rexicon226 (dismissed):
Looks fine to merge as is. Curious to start collecting metrics.

Rexicon226 (approved):
LGTM

PR #396, opened 2024-11-20
 fix(cleanup service) Improve findSlotsToClean test

dadepo:
- [x] The memtable is only flushed to file on second run of the tests. Need to figure out how to trigger the flush programmatically.


src/ledger/database/rocksdb.zig
```diff
@@ -87,6 +87,12 @@ pub fn RocksDB(comptime column_families: []const ColumnFamily) type {
         pub fn count(self: *Self, comptime cf: ColumnFamily) Allocator.Error!u64 {
             const live_files = try self.db.liveFiles(self.allocator);
             defer live_files.deinit();
+            defer for (live_files.items) |file| {
+                self.allocator.free(file.column_family_name);
+                self.allocator.free(file.name);
+                self.allocator.free(file.start_key.?);
+                self.allocator.free(file.end_key.?);
```
dnut, line 94:
```suggestion
                file.deinit();
```

src/ledger/tests.zig
```diff
@@ -330,12 +330,18 @@ pub const TestState = struct {
 };
 
 pub const TestDB = struct {
-    const dir = sig.TEST_STATE_DIR ++ "/blockstore";
+    const dir = sig.TEST_STATE_DIR ++ "blockstore";
```
dnut, line 333: why remove the slash? I put this here because I try to make a habit of including the delimiter when constructing paths to depend on fewer assumptions.

0xNineteen (dismissed):
lgtm 🚀

src/ledger/tests.zig
```diff
@@ -330,12 +330,18 @@ pub const TestState = struct {
...
 
     pub fn init(comptime test_src: std.builtin.SourceLocation) !BlockstoreDB {
         return try initCustom(std.testing.allocator, test_src);
     }
 
+    pub fn reuseBlockstore(comptime test_src: std.builtin.SourceLocation) !BlockstoreDB {
+        const path = comptimePrint("{s}/{s}/{s}", .{ dir, test_src.file, test_src.fn_name });
+        try std.fs.cwd().makePath(path);
```
dnut, line 341: If this reuses an existing blockstore, I'd expect this to return `error.PathAlreadyExists`. Are you expecting this folder to exist?

src/ledger/cleanup_service.zig
```diff
@@ -401,53 +402,99 @@ const TestDB = ledger.tests.TestDB;
...
+        {
+            var write_batch = try db.initWriteBatch();
+            defer write_batch.deinit();
+            for (0..1000) |i| {
+                try write_batch.put(ledger.schema.schema.data_shred, .{ 19, i }, data_shred.payload);
+            }
+            try db.commit(&write_batch);
+        }
+        // When implementation is not rocksdb, we do not need to flush memtable to disk to be able to assert.
+        if (build_options.blockstore_db != .rocksdb) {
```
dnut, line 467: Instead of separate scopes and repeating the same assertions multiple times, what do you think about just adding this before the assertions?

```zig
    if (build_options.blockstore_db == .rocksdb) {
        db.deinit();
        db = try TestDB.reuseBlockstore(@src());
        reader.db = db;
    }
```

dnut (commented):
Nice work, just a few minor nitpicks, and one thing I'm confused about regarding the database reuse.

src/ledger/database/rocksdb.zig
dadepo: Thanks! I'll apply this suggestion. 

I'm also curious why this leak did not show up until now. Got any ideas?

dadepo: [74a13e2](https://github.com/Syndica/sig/pull/396/commits/74a13e281a85d357ee4d9fc41ab2f16cd04f406c)

src/ledger/tests.zig
dadepo: > I'd expect this to return error.PathAlreadyExists

That is actually not the documented behaviour. As can be seen here:

https://github.com/ziglang/zig/blob/87863a834b074311d70ec4dd3b10c74705fff54d/lib/std/fs/Dir.zig#L1142

It would return success if the path already exists and is a director.

> Are you expecting this folder to exist?

Yes. Maybe I rename it to `reusePath`?

src/ledger/tests.zig
dadepo: This was actually part of changes made while troubleshooting the issue to figure out what could be going wrong.

Returned in [11489a7](https://github.com/Syndica/sig/pull/396/commits/11489a75baa3471c069f493b15f6f630c2b5e65a)

src/ledger/database/rocksdb.zig
```diff
@@ -87,6 +87,9 @@ pub fn RocksDB(comptime column_families: []const ColumnFamily) type {
         pub fn count(self: *Self, comptime cf: ColumnFamily) Allocator.Error!u64 {
             const live_files = try self.db.liveFiles(self.allocator);
             defer live_files.deinit();
+            defer for (live_files.items) |file| {
```
yewman, line 90: Thoughts on single line? 
```suggestion
            defer for (live_files.items) |file| file.deinit();
```

src/ledger/tests.zig
dnut: my mistake. seems fine as is

src/ledger/database/rocksdb.zig
dnut: 9b16c32fc54d6075160bfd48ed1d682d2f481bd7

0xNineteen (approved):

PR #399, opened 2024-11-22
fix(ledger): database reliability

dnut:
the current approach i'm going with is to switch from rocksdb to the hashmapdb and get it working reliably by using the disk backed allocator. it seems to be the fastest way to get a ledger that isn't broken. in the future we should also fix rocksdb or lmdb


dnut:
closing to change base branch

PR #400, opened 2024-11-22
fix(ledger): improve database reliability with hashmap db

dnut:
This switches the default database from rocksdb to the hashmap db.

To get the hashmap db working reliably over extended periods of time, it uses the disk backed allocator. Without this, the entire database would be stored in memory, eventually resulting in OOM.

To avoid creating absurd quantities of files that would cause errors due to too many open files, the disk allocator is only used occasionally to allocate 1 GB all at once. This 1 GB is consumed gradually over time. I implemented a new allocator called BatchAllocator that is responsible for executing this large batching. It acts as an intermediary between the disk allocator and the database. It provides an ordinary allocator interface which is used by the hashmap db.

To get parity with rocksdb in the benchmarks, I added reference counting in the database. When an entire shred needs to be read out of the hashmap db, previously it allocated 1 KB and copied the shred. Now it just copies the pointer out of the database, and it uses reference counting to ensure that the pointer lives sufficiently long.


0xNineteen (approved):
lgtm

src/utils/allocators.zig
```diff
@@ -514,6 +520,184 @@ pub fn RecycleFBA(config: struct {
...
+        return ptr;
+    }
+
+    fn writeBatchPtr(batch: *Batch, slice_with_batch: []u8) void {
+        const batch_int = @intFromPtr(batch);
+        const batch_ptr_bytes: [@sizeOf(usize)]u8 = @bitCast(batch_int);
+        @memcpy(
+            slice_with_batch[slice_with_batch.len - @sizeOf(usize) .. slice_with_batch.len],
+            &batch_ptr_bytes,
+        );
```
InKryption, line 675:
```suggestion
        const batch_ptr_slice = slice_with_batch[slice_with_batch.len - @sizeOf(usize)..][0..@sizeOf(*Batch)];
        std.mem.bytesAsValue(*Batch, batch_ptr_slice) = batch;
```

src/utils/allocators.zig
```diff
@@ -514,6 +520,184 @@ pub fn RecycleFBA(config: struct {
...
+    new_batch_lock: std.Thread.RwLock = .{},
+    new_batch_waiters: Atomic(usize) = Atomic(usize).init(0),
+    new_batch_wait_lock: std.Thread.RwLock = .{},
+
+    const Self = @This();
+
+    const Batch = struct {
+        ptr: [*]u8,
+        len: usize,
+        fba: FixedBufferAllocator,
```
InKryption, line 537: why `ptr` and `len` instead of `slice: []u8`, and why separately outside of `fba`?

src/utils/allocators.zig
```diff
@@ -514,6 +520,184 @@ pub fn RecycleFBA(config: struct {
...
+        batch.fba.threadSafeAllocator().rawFree(full_buf, ptr_align, return_address);
+        if (Batch.DONE + 1 == batch.num_allocs.fetchSub(1, .monotonic)) {
+            batch.deinit(self.backing_allocator);
+        }
+    }
+
+    fn getBatch(allocated_slice: []u8) *Batch {
+        const batch_ptr_bytes = allocated_slice.ptr[allocated_slice.len..][0..@sizeOf(usize)].*;
+        const batch_int: usize = @bitCast(batch_ptr_bytes);
+        return @ptrFromInt(batch_int);
```
InKryption, line 697:
```suggestion
        const batch_ptr_bytes = allocated_slice.ptr[allocated_slice.len..][0..@sizeOf(Batch)];
        return std.mem.bytesToValue(*Batch, batch_ptr_bytes);
```

build.zig
```diff
@@ -12,7 +12,7 @@ pub fn build(b: *Build) void {
     const filters = b.option([]const []const u8, "filter", "List of filters, used for example to filter unit tests by name"); // specified as a series like `-Dfilter="filter1" -Dfilter="filter2"`
     const enable_tsan = b.option(bool, "enable-tsan", "Enable TSan for the test suite");
     const no_run = b.option(bool, "no-run", "Do not run the selected step and install it") orelse false;
-    const blockstore_db = b.option(BlockstoreDB, "blockstore-db", "Blockstore database backend") orelse .rocksdb;
+    const blockstore_db = b.option(BlockstoreDB, "blockstore", "Blockstore database backend") orelse .hashmap;
```
InKryption, line 15: Non-actionable comment: would probably be good to make our rocksdb dependency lazy and dependent on this (ditto for lmdb).

InKryption (changes requested):
Just a few comments, suggestions, and a comment.

build.zig
dnut: is it not?

src/utils/allocators.zig
```diff
@@ -1066,29 +1250,95 @@ test "fuzzDiskMemoryAllocator - past failures" {
...
+        // .{ 0, 10, 32457527 },
+        // .{ 13, 10, 1315344 },
+        // .{ 44, 10, 122063 },
+        // .{ 200, 10, 1439666 },
+        // .{ 1139, 3, 240 },
+        // .{ 144, 10, 293 },
+        // .{ 12, 10000, 1067751 },
+        // 1, iterations: 5, batch_size: 1433431
+    };
+    const debug = true;
```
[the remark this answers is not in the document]
dnut: good catch. i left this test in a bad state. daa73d137f7d1172ceb650baaa7a30daa5466f11

src/utils/allocators.zig
```diff
@@ -1066,29 +1250,95 @@ test "fuzzDiskMemoryAllocator - past failures" {
...
+    var batch_allocator = BatchAllocator{
+        .backing_allocator = std.testing.allocator,
+        .batch_size = 1433431,
+    };
+    defer batch_allocator.deinit();
+    const allocator = batch_allocator.allocator();
+    const item0 = try allocator.alloc(u8, 747105);
+    const item1 = try allocator.alloc(u8, 746217);
+    allocator.free(item0);
+    allocator.free(item1);
```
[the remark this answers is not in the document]
dnut: this test was actually redundant with the "past failures" test. this code was auto-generated by `debug` mode of the fuzzer. 18bdc73bd014dcc32b00d15252a9f53da7d961ab

build.zig
InKryption: Just referring to `b.lazyDependency`

src/utils/allocators.zig
dnut: It's not clear to me why `bytesAsValue` exists in std when `@ptrCast` already exists. `bytesAsValue` is less widely used, which makes it esoteric. In my opinion, this makes the code harder to understand compared to using `@ptrCast` directly. What do you think?

There is also a change in semantics with your suggestion. Previously, this was just copying the bytes of the pointer. In that case, it doesn't matter how the bytes are aligned. With your suggestion, the bytes in the allocation are directly interpreted as the pointer. This means the bytes need to be properly aligned.

To align the bytes properly, we need some new logic to adjust the size and alignment of allocation. Also, the `writeBatchPtr` and `getBatch` functions need to change in tandem. I also ended up refactoring this a bit to consolidate/encapsulate this logic and clarify the meaning of the inputs. Hopefully the code is easy to understand.

03c83a4cf3e4baf994783a93a6e3eca6bea0d36f

src/utils/allocators.zig
dnut: see other thread: https://github.com/Syndica/sig/pull/400#discussion_r1880200087

src/utils/allocators.zig
dnut: I guess I did this for convenience, without putting much thought into it.

f833bf9f07dd43a0f53961eb553de57609efcb64

src/utils/allocators.zig
InKryption: `bytesAsValue` directly copies the pointer attributes of the source byte ptr - so for example, in this case, the pointer it returns will be of type `*align(1) *Batch`. It results in the same fundamental operation as was being done here, ie it copies the bytes representing the pointer to the slice, which ends up being an unaligned copy; the benefit here is the type safety (ie avoiding `@intFromPtr`, and automatically inferring the appropriate pointer alias properties).

The change you've applied does improve things, but it does incur potentially more memory usage than my suggestion, since it increases the alignment requirement for storing metadata.

I'm fine to accept the changes you've made, but perhaps consider them in light of my explanation.

dnut: > it does incur potentially more memory usage than my suggestion, since it increases the alignment requirement for storing metadata.

My point was that your suggestion does not work without increasing memory usage. Your suggestion requires additional changes to fix the alignment, which is what I implemented.

src/utils/allocators.zig
InKryption: Hm, I will say I didn't realize the full relationship, it's now clearer through the updated code; not pretty, but I think that's better than hiding that representation.

src/utils/allocators.zig
InKryption: Can maintain conversation in other thread.

src/utils/allocators.zig
InKryption: It doesn't increase memory usage though; it's functionally identical to the original code. The change you applied is not the same as what I suggested, where you are casting to `**Batch`; my suggestion is effectively your change, but casting to `*align(1) *Batch`.

InKryption (dismissed):

src/sync/reference_counter.zig
```diff
@@ -84,3 +162,13 @@ test "sync.ref_counter: ReferenceCounter works" {
     try std.testing.expect(!x.release());
     try std.testing.expect(x.release());
 }
+
+test "RcSlice payload has the correct data" {
```
dadepo, line 166: It would be nice to also have tests covering the reference counting implementation.

Tried writing one as follows but was getting segfault

```
test "RcSlice reference counting" {
    const slice1 = try RcSlice(u8).alloc(std.testing.allocator, 5);
    defer slice1.deinit(std.testing.allocator);

    // Copy data into the slice1's payload
    @memcpy(slice1.payload(), "hello");

    // Acquire a second reference to the RcSlice
    const slice2 = slice1.acquire();

    // Assert that both slices see the same payload
    const hello: *const [5]u8 = "hello";
    try std.testing.expect(sig.utils.types.eql(hello, slice1.payload()[0..5]));
    try std.testing.expect(sig.utils.types.eql(hello, slice2.payload()[0..5]));

    // Modify the payload via the second reference
    slice2.payload()[0] = 'H';

    // Assert the change is visible through the first reference
    const hello2: *const [5]u8 = "Hello";
    try std.testing.expect(sig.utils.types.eql(hello2, slice1.payload()[0..5]));
    try std.testing.expect(sig.utils.types.eql(hello2, slice2.payload()[0..5]));

    // Release the first reference
    slice1.deinit(std.testing.allocator);

    // The second reference should still be valid
    try std.testing.expect(sig.utils.types.eql(hello2, slice2.payload()[0..5]));
    // Release the final reference
    slice2.deinit(std.testing.allocator);
}
```

Not sure if the issue is in the way I am attempting the test

.github/workflows/check.yml
```diff
@@ -55,8 +55,8 @@ jobs:
 
       - name: test
         run: |
-          zig build test -Denable-tsan=true
-          zig build test -Denable-tsan=true -Dblockstore-db=hashmap -Dfilter=ledger
+          zig build test -Denable-tsan=true --verbose --summary all 
+          zig build test -Denable-tsan=true --verbose --summary all -Dblockstore=rocksdb -Dfilter=ledger
```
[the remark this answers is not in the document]
dnut: yes this change is intended. this is necessary to ensure all the same tests are run as previously.

the purpose of these commands as they were already written was to test both the default database and the non-default database (which was hashmapdb). now that rocksdb is the non-default database, it needs to be tested explicitly, instead of hashmapdb.

i don't see how this could be causing problems in ci since it's running the same tests, and when i run it locally it finishes very quickly. but i can't explain what's going on in ci, so i guess anything is possible.

src/sync/reference_counter.zig
```diff
@@ -74,7 +74,85 @@ pub const ReferenceCounter = struct {
...
+            const self = Self{
+                .ptr = @alignCast(value_ptr - reserved_space),
+                .len = value.len,
+            };
+
+            self.deinit(allocator);
+        }
+
+        pub fn acquire(self: Self) Self {
+            std.debug.assert(self.refCount().acquire());
```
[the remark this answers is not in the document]
dnut: this is like any other usage of assert. it asserts an invariant that is assumed to be guaranteed to be true. if this check returns false, it means the entire data structure is implemented incorrectly, and this function has undefined behavior. in that case, this code could trigger a segfault before it even runs the assertion.

the main idea of using ReleaseFast or ReleaseSmall is to skip assertions like this. so i don't see it as a drawback that assertions are not run in those modes. i don't think we should never run sig in ReleaseFast or ReleaseSmall unless we are confident that all of our assertions are 100% correct. i would use ReleaseSafe to build code for production deployments.

you could make an argument that there are two classes of assertions:
1. assertions that you eventually are confident are definitely correct
2. assertions that you can never know are actually correct

then you could argue that `assert` or `unreachable` can be used for `1` and `panic` can be used for `2`. personally i lean towards saying that `1` doesn't even exist, and we should always use ReleaseSafe unless we're willing to risk undefined behavior for maximum performance.

src/sync/reference_counter.zig
dnut: there's a segfault because you deinit it three times. the test passes if you remove the defer near the beginning.

I made another tweak to your test to validate exactly when the pointer is freed.

e8653ff09b526f2a11132e9e9041a58f9cd2f53c

InKryption (dismissed):

InKryption (approved):

PR #404, opened 2024-11-25
fix(cmd): fix downloadSnapshot gossip loading

0xNineteen:
- removes old and unused `initGossip` method


InKryption (approved):
code removal W

PR #407, opened 2024-11-26
refactor(gossip): improved init flow

0xNineteen:
- Moved GossipService allocation to heap with new `create()` function
  - there were a random number of structs which were heap allocated, whereas others were not - this change puts everything on the stack, and uses the `create` method to make everything on the heap (which it should be due to the multi-threaded nature of gossip)
- Removed thread pool dependency from GossipTable initialization (was dead code)
- Removed unused echo server implementation (was dead code)
  - ticket to add it back with working server [here](https://github.com/orgs/Syndica/projects/2/views/10?pane=issue&itemId=88750700)
- Added proper cleanup of muxes in GossipService deinit

### DIFF
--- a/src/gossip/service.zig
+++ b/src/gossip/service.zig
@@ -57,6 +57,7 @@ const Duration = sig.time.Duration;
 const endpointToString = sig.net.endpointToString;
 const globalRegistry = sig.prometheus.globalRegistry;
 const getWallclockMs = sig.time.getWallclockMs;
+const deinitMux = sig.sync.mux.deinitMux;
 
 const PACKET_DATA_SIZE = sig.net.packet.PACKET_DATA_SIZE;
 const UNIQUE_PUBKEY_CAPACITY = sig.gossip.table.UNIQUE_PUBKEY_CAPACITY;
@@ -127,10 +128,11 @@ const MAX_PROCESS_BATCH_SIZE = 64;
 ///           after the `sendSocket` thread exits.
 ///
 pub const GossipService = struct {
+    /// used for general allocation purposes
     allocator: std.mem.Allocator,
+    /// used specifically to allocate the gossip values
     gossip_value_allocator: std.mem.Allocator,
 
-    // note: this contact info should not change
     gossip_socket: UdpSocket,
     /// This contact info is mutated by the buildMessages thread (specifically, .shred_version and .wallclock),
     /// so it must only be read by that thread, or it needs a synchronization mechanism.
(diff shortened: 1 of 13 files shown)


0xNineteen:
* **#407** <a href="https://app.graphite.dev/github/pr/Syndica/sig/407?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Syndica/sig/407?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

src/gossip/service.zig
```diff
@@ -145,36 +148,59 @@ pub const GossipService = struct {
...
-    thread_pool: *ThreadPool,
-    echo_server: EchoServer,
-
     metrics: GossipMetrics,
 
     const Self = @This();
 
     const Entrypoint = struct { addr: SocketAddr, info: ?ContactInfo = null };
 
+    pub fn create(
```
yewman, line 179: NIT: Given that init is not used anywhere else from what I can tell, perhaps we don't need the new method. Also it would almost be nice to indicate in the method name whether we are initialising a struct on the heap or the stack. There is something I don't like about Struct.init(...) returning !*Self or !Self arbitrarily throughout sig. Interested to here your thoughts

yewman, line 179: NIT: Given that init is not used anywhere else from what I can tell, perhaps we don't need the new method. Also it would almost be nice to indicate in the method name whether we are initialising a struct on the heap or the stack. There is something I don't like about Struct.init(...) returning !*Self or !Self arbitrarily throughout sig. Interested to here your thoughts

src/gossip/service.zig
```diff
@@ -2920,6 +2938,7 @@ test "test large push messages" {
     defer {
         gossip_service.shutdown();
         gossip_service.deinit();
+        allocator.destroy(gossip_service);
```
yewman, line 2941: Is there an issue with calling self.allocator.destroy from within the deinit method?

yewman (commented):
Love these changes, just a couple of comments that I am interested to hear your view on

src/gossip/service.zig
0xNineteen: the idea was 'create' would create objects on the heap, and 'init' on the stack, similar to the allocator interface which uses `allocator.create(T)` -- the same pattern is followed by the channel -- we could delete `init` but i think the separation makes things clear and its nice to have both

src/gossip/service.zig
0xNineteen: if we did we would make the assumption that the service is always on the heap which i dont think we want to make

src/gossip/service.zig
yewman: Yeah great, I like that. I think there are some other init methods which return pointers so I was confused. This is great then!

PR #412, opened 2024-11-29
fix(snapshots): Update & fix snapshot manifest epoch rewards type & refactor extra bank fields

InKryption:
The snapshot format was updated, replacing `epoch_rewards_status` with `versioned_epoch_stakes`, and adding `accounts_lt_hash` (lt = lattice). This is in line with the updates in the agave implementation.

This new `VersionedEpochStakesMap` also required adding the other `Stakes<>` types, so I simply made that generic like it is in the agave codebase.

This also changes the `ExtraFields`' (previously known as `BankFields.Extra`) bincode [de]serialization description to simply be specified directly, instead of through the usual declarative `@"!bincode-config:<field_name>"` glue style; this actually fixes bad eof behavior it potentially had before. This also has been accounted for in the `initRandom` function.

### Update (Edit)
This PR now also updates the test snapshots to ones generated more recently, courtesy of @0xNineteen - documentation for how this was done is underway.

In tandem with that, it also updates a large number of tests to be more sound and account for this update, as well as the deletion of a few redundant test-data files. This entails unpacking snapshots into temporary directories instead of mutating test-data, and in general avoiding mutating our test-data folder.

This change also brought to light some leaks, that apparently simply weren't occurring with the previous snapshot pair, so included are changes to a couple of allocating APIs.


src/accountsdb/snapshots.zig
```diff
@@ -923,6 +1078,171 @@ pub const BankFields = struct {
     }
 };
 
+/// Analogous to https://github.com/anza-xyz/agave/blob/8d1ef48c785a5d9ee5c0df71dc520ee1a49d8168/runtime/src/serde_snapshot.rs#L396
```
0xNineteen, line 1081: would be good to include the rust struct name which its analogous too (ie, ExtraFieldsToDeserialize) -- this mainly helps with searching across clients (if you know the name in agave but want to see the equivalent in sig, you just search the rust struct)

src/accountsdb/snapshots.zig
```diff
@@ -923,6 +1078,171 @@ pub const BankFields = struct {
...
+
+    pub fn deinit(extra: *const ExtraFields, allocator: std.mem.Allocator) void {
+        var versioned_epoch_stakes = extra.versioned_epoch_stakes;
+        for (versioned_epoch_stakes.values()) |ves| ves.deinit(allocator);
+        versioned_epoch_stakes.deinit(allocator);
+    }
+
+    pub const VersionedEpochStakesMap = std.AutoArrayHashMapUnmanaged(u64, VersionedEpochStake);
+
+    /// TODO: Investigate this `LtHash` business:
```
0xNineteen, line 1105: just a note we have context here: https://github.com/orgs/Syndica/projects/2/views/10?pane=issue&itemId=85238686

src/accountsdb/snapshots.zig
```diff
@@ -923,6 +1078,171 @@ pub const BankFields = struct {
...
+                .versioned_epoch_stakes => value.count() == 0,
+                .accounts_lt_hash => value == null,
+            };
+
+            if (is_eof_value) {
+                inline for (fields[0 .. i + 1]) |field| {
+                    try bincode.write(writer, @field(data, field.name), params);
+                }
+                return;
+            }
```
0xNineteen, line 1236: if an eof_value is never found, then nothing is written -- i think it still should be valid to serialize the full optional/zero values -- the default-on-eof should just be a saftey thing across client versions which have more/less data to read

src/accountsdb/snapshots.zig
```diff
@@ -1052,22 +1372,129 @@ pub const AccountsDbFields = struct {
...
+                }
+                const value = try bincode.read(assert_allocator, AccountFileInfo, reader, params);
+
+                const gop = file_map.getOrPutAssumeCapacity(key);
+                if (gop.found_existing) return error.DuplicateFileMapEntry;
+                gop.value_ptr.* = value;
+            }
+
+            break :blk file_map;
+        };
```
0xNineteen, line 1429: why couldnt we just call bincode.read(Hashmap(Slot, []AccountFileInfo), ...) ? - feels like it would be better to have this as a per-field config as it was before?

0xNineteen (changes requested):

src/accountsdb/snapshots.zig
InKryption: Oversight, good catch

src/accountsdb/snapshots.zig
InKryption: Well, reading it as a `HashMap(Slot, []AccountFileInfo)` directly would require an extra allocating conversion, so that would be less than ideal, this mimics the behavior of the original field config.

The per-field config approach was what originally hid the incorrect deserialization via the default-on-eof config, which I'm trying to avoid.

If the concern is the amount of logic here, I would suggest instead adding a simple function that does the equivalent, to be called here.
For example:
```zig
var file_map = try bincode.hashmap.readCtx(allocator, FileMap, params, struct {
    pub const readKey = {};
    pub fn readValue(_allocator: std.mem.Allocator, _reader: anytype, _params: bincode.Params) !AccountFileInfo {
        if (try bincode.readIntAsLength(usize, reader, params) != 1) {
            return error.TooManyAccountFileInfos;
        }
        return try bincode.read(_allocator, AccountFileInfo, _reader, _params);
    }
});
```
What do you think of that?

src/accountsdb/snapshots.zig
InKryption: Good catch

src/accountsdb/snapshots.zig
0xNineteen: nice - yeah im good with that - we would also want to remove `hashMapFieldConfig` too right? since it would be deadcode

src/accountsdb/snapshots.zig
InKryption: a60c1c1da8bfca250ac09a60dcc8a9752e8936bc

src/accountsdb/snapshots.zig
InKryption: If we're just going to serialize it without eof logic, we can just use the default logic: 64562124471e41821c1895fb585e69a84521b450

src/accountsdb/snapshots.zig
InKryption: For the time being it isn't feasible to delete that just yet, it's still the basis of the [de]serialization of a lot of code we'll have to chip away at.

InKryption: ca28a947b0176442e8c44995f87a9f0156e239b2

src/bincode/hashmap.zig
```diff
@@ -7,11 +7,113 @@ const hashMapInfo = sig.utils.types.hashMapInfo;
...
+
+    const ctx_maybe_namespace = if (@TypeOf(ctx) != void) ctx else struct {
+        pub const readKey = {};
+        pub const freeKey = {};
+        pub const readValue = {};
+        pub const freeValue = {};
+    };
+    const ctx_impl = if (@TypeOf(ctx_maybe_namespace) != type) ctx_maybe_namespace else struct {
+        pub fn readKey(
+            _allocator: std.mem.Allocator,
```
yewman, line 42: why are we underscoring params?

yewman (commented):
Couple if minor queries, otherwise lgtm

src/accountsdb/snapshots.zig
InKryption: da113b56002e2d6bba09c235fc44a19ea9f2b848

src/bincode/hashmap.zig
InKryption: Because they can't shadow the parameters of the surrounding function; not much else to name them.

src/accountsdb/db.zig
```diff
@@ -274,14 +273,15 @@ pub const AccountsDB = struct {
...
+        /// Must have been allocated with `self.allocator`.
         snapshot_fields_and_paths: *AllSnapshotFields,
         n_threads: u32,
         validate: bool,
         accounts_per_file_estimate: u64,
         should_fastload: bool,
         save_index: bool,
     ) !SnapshotFields {
-        const snapshot_fields = try snapshot_fields_and_paths.collapse();
+        const snapshot_fields = try snapshot_fields_and_paths.collapse(self.allocator);
```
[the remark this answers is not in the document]
InKryption: `all_snapshots` sounds good to me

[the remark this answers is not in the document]
InKryption: 596a0644d70fad17747a51cbc8b91a591d3e99c3

InKryption:
Non-canonical strikes again: https://github.com/Syndica/sig/actions/runs/12248154039/job/34167259228?pr=412

src/accountsdb/db.zig
```diff
@@ -3259,6 +3266,14 @@ fn testWriteSnapshotIncremental(
 }
 
 test "testWriteSnapshot" {
+    // TODO: loading once from the full snapshot, and then a second time from the incremental snapshot,
```
0xNineteen, line 3269: can we create a ticket for solving this in a future pr

0xNineteen (approved):

src/accountsdb/db.zig
0xNineteen: https://github.com/Syndica/sig/pull/412#discussion_r1877984200

PR #415, opened 2024-12-02
fix(build): temporary fix to detect the M3 cpu by default

InKryption:
We should remove this code when we update to 0.14, where this feature detection is implemented properly in the underlying code used by the build system.

This will allow us to just `zig build` without needing to specify the CPU target for the time being.


Rexicon226 (approved):
LGTM!

PR #423, opened 2024-12-05
feat(shred-network): minimize dependencies

dnut:
The shred-collector command no longer needs the following since it uses RPC to fill in any missing data.
- accountsdb
- `--leader-schedule` cli argument
- `--test-repair-for-slot` cli argument

You can run the shred-collector command more simply now, and it will start up pretty quickly without needing a snapshot:

```bash
sig shred-collector -n testnet
```

## Implementation

This change makes it so the retransmit service does not have a hardcoded dependency on BankFields and LeaderScheduleCache to provide the epoch schedule, staked nodes, or the leader schedule. Instead, it expresses that it has a dependency on these three individual data types. By more specifically expressing its actual requirements, it reduces unnecessary coupling between components. This makes the code more flexible, easier to change, and clearer about its inputs.

This was implemented for static data (the EpochSchedule) by passing the EpochSchedule itself directly into the retransmit service. This was implemented for dynamic data (staked nodes and leader schedule) by passing an interface into the retransmit service, which provides the required data on demand.

The dependencies can be satisfied by either AccountsDB or RPC. AccountsDB is used for the validator command, and RPC is used for the shred-collector command. The RPC client is also used to provide the start slot, if it was not provided in the cli.

### DIFF
--- a/src/adapter.zig
+++ b/src/adapter.zig
@@ -1,4 +1,4 @@
-//! Glue code for hooking up application components from distant areas of the code.
+//! Links dependencies with dependents. Connects components from distant regions of the code.
 
 const std = @import("std");
 const sig = @import("sig.zig");
@@ -30,17 +30,17 @@ pub const RpcSlotLeaders = struct {
     }
 
     pub fn slotLeaders(self: *Self) leader_schedule.SlotLeaders {
-        return leader_schedule.SlotLeaders.init(self, slotLeader);
+        return leader_schedule.SlotLeaders.init(self, get);
     }
 
-    fn slotLeader(self: *Self, slot: sig.core.Slot) ?sig.core.Pubkey {
-        return self.slotLeaderFallible(slot) catch |e| {
+    fn get(self: *Self, slot: sig.core.Slot) ?sig.core.Pubkey {
+        return self.getFallible(slot) catch |e| {
             self.logger.err().logf("error getting leader for slot {} - {}", .{ slot, e });
             return null;
         };
     }
 
-    fn slotLeaderFallible(self: *Self, slot: sig.core.Slot) !?sig.core.Pubkey {
+    fn getFallible(self: *Self, slot: sig.core.Slot) !?sig.core.Pubkey {
         if (self.cache.slotLeader(slot)) |leader| {
             return leader;
         }
@@ -66,29 +66,23 @@ pub const RpcStakedNodes = struct {
     const Self = @This();
     const StakedNodes = sig.shred_network.shred_retransmitter.StakedNodes;
     const NodeToStakeMap = StakedNodes.NodeToStakeMap;
-    const Cache = sig.utils.lru.LruCacheCustom(
-        .locking,
+    const Cache = sig.utils.lru.SharedPointerLru(
         sig.core.Epoch,
-        *NodeToStakeMap,
+        NodeToStakeMap,
         Allocator,
-        destroyCacheItem,
+        NodeToStakeMap.deinit,
     );
 
     pub fn init(allocator: Allocator, rpc_client: sig.rpc.Client) !Self {
         return .{
             .allocator = allocator,
             .rpc_client = rpc_client,
-            .cache = try Cache.initWithContext(allocator, 8, allocator),
+            .cache = try Cache.init(allocator, 8, allocator),
         };
     }
 
     pub fn stakedNodes(self: *Self) StakedNodes {
-        return StakedNodes.init(self, get);
-    }
-
-    fn destroyCacheItem(element: *NodeToStakeMap, allocator: std.mem.Allocator) void {
-        element.deinit(allocator);
-        allocator.destroy(element);
+        return StakedNodes.init(self, get, release);
     }
 
     fn get(self: *Self, epoch: sig.core.Epoch) anyerror!*const NodeToStakeMap {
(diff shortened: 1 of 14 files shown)


yewman:
I agree that as it stands, the way dependencies are managed in the retransmit service needs improvement. I am not sure I agree that interfaces are the right approach though. Whilst they do reduce coupling, they also introduce an additional layer of indirection which increases complexity. 

If when running the full validator we want to be able to switch between implementations of the proposed interfaces via config then I would be more in favour of the changes, however, since the full validator will only ever run a single implementation I think it is clearer if we use concrete types. We can easily populate these types via rpc in order to run the shred collector without a snapshot.

In terms of coupling, the data structures in question are very simple so I think the benefit gained with respect to flexibility and ease of change does not justify the additional indirection. 

Anyway these are just my thoughts for now and I would be interested to hear others takes on this issue. @0xNineteen

dnut:
In my eyes, the interfaces are justified for the general benefits of decoupling on their own.

To be clear, I don't think interfaces are the only way to accomplish this decoupling. It can also be accomplished with other mechanisms like message passing or sharing state. In this case, I felt that interfaces were simplest, but I'm open to alternatives. A nice thing about interfaces is they are the easiest form of decoupling to adapt into another form of decoupling.

I feel that high-level application components should be loosely coupled. I've dealt with too many codebases that evolved into a tangled mess of spaghetti that was too difficult to maintain due to the sheer number of components that would need to changes as soon as one component changes. The complexity sneaks up on you. It's especially painful when there are cyclic dependencies, or when breakages cascade transitively.

I also find it easier to understand code when it has minimal dependencies. It's easier to reason about code whose inputs are only the actual data it needs, and not a massive superset of the actual required data. I think one of the reasons why agave's code is difficult to make sense of is the number of unnecessary dependencies. Sometimes it feels that I can't understand any part of the codebase until I understand every detail of the entire thing. If the code is implemented as modular/pluggable components with clear and minimal APIs between them, it becomes easier to understand each component in isolation.

RPC, accountsdb, and shred-network are pretty distant areas of the code, and I think we would benefit from minimizing the amount that such distinct components directly depend on each other. 

----

But actually, this is not the primary motivation for my change. My primary motivation was just to implement the desired feature as correctly as possible.

> since the full validator will only ever run a single implementation I think it is clearer if we use concrete types. We can easily populate these types via rpc in order to run the shred collector without a snapshot.

The interfaces are only being used for dynamic data. I used concrete types for static data. "Dynamic" data in this case is data that is different for every epoch, and can only be calculated based on data that is generated when the epoch arrives.

If you want to use concrete types, you have two options: 
1. Settle on a single implementation that is used everywhere. This means you pick either RPC or AccountsDB, and you use that exclusively, for both `validator` and `shred-collector`. This is not compatible with the idea of using accountsdb for `validator` and rpc for `shred-collector`.
2. Calculate the data in the main function, and pass down the result of that calculation. This data will become invalid after epoch transitions.

I understand epochs are long, so `2` is not a big deal for short-term testing of the validator. But sig also should be able to work for multiple days without failing. I think we should actually set up a server pretty soon to start running sig continuously. With approach `2`, *neither* `validator` nor `shred-collector` will work properly across epoch boundaries.

yewman:
> What about BankFields?

I agree that it should not depend on BankFields, my comment referring to the simplicity of the data structures is more addressing the fact that the leader schedule and staked nodes are simple, and only change once per epoch. We need some data structures which abstract away the transition logic and allows a user to get the leader schedule or staked nodes for the correct epoch during a transition but that is it. 

> In my eyes, the interfaces are justified for the general benefits of decoupling on their own.

If there is only one implementation of an interface it provides no decoupling benefits, just code duplication and indirection. I would argue that we don't need an ppc implementation for slot leaders and staked nodes, for running test services across epochs we an just schedule rpc calls to fetch epoch staked nodes when required and save that into a data structure which provides the epoch boundary transition logic, similar to LeaderScheduleCache (however should be better named as it is not a cache). 

> I feel that high-level application components should be loosely coupled.

I agree, but I don't think these interfaces actually improve the coupling. The public api of a struct is essentially its own interface. If we only require one implementation, which I argue we do, then the interface does not reduce coupling.

> I also find it easier to understand code when it has minimal dependencies

I agree, but I don't think interfaces necessarily reduce dependencies. 

> I think one of the reasons why agave's code is difficult to make sense of is the number of unnecessary dependencies.

I agree, I would also argue that one of the reasons agave is so difficult to understand is the way they use traits, it can be difficult to track down what code is actually being run in many instances.

> RPC, accountsdb, and shred-network are pretty distant areas of the code, and I think we would benefit from minimizing the amount that such distinct components directly depend on each other.

Agreed, I definitely think the dependency management here needs improvement, I just don't think interfaces are the correct solution. 

> If you want to use concrete types, you have two options:

I strongly disagree with your conclusions here. The main epoch dependant data is the epoch staked nodes which are required in a number of places throughout the validator for generating pseudo-random behaviour among other things. We could create a read write locked EpochStakesProvider and when running the `validator` update it when the data becomes availble, or for running the `shred-collector` just use an rpc call to get the required data at epoch transitions and update the provider. This would mean we have **one** implementation but use rpc to fetch data when running without the full validator. 

Note: These are my initial thoughts and I acknowledge that they may be flawed, I will create a PR to better express / flesh out my thoughts as to an alternative.

dnut:
closing in favor of #462

PR #425, opened 2024-12-06
perf(bechmarks): use direct print logger instead of channel logger

dnut:
When I look at the flamegraphs from running benchmarks, the majority of the time is spent in the logger, busy spinning while waiting for log messages. This makes it harder to understand benchmark results.

I don't see a reason to use something as complex as a channel logger just to print information about the benchmarks.

Seeing these flamegraphs makes me concerned about the channel still being overly greedy of system resources. But I'm not trying to address this in the current pr. I have another PR which adds blocking recv methods to the channel to prevent busy spinning.

### DIFF
--- a/src/benchmarks.zig
+++ b/src/benchmarks.zig
@@ -76,12 +76,10 @@ fn exitWithUsage() noreturn {
 /// zig build benchmark -- gossip
 pub fn main() !void {
     const allocator = std.heap.c_allocator;
-    var std_logger = try sig.trace.ChannelPrintLogger.init(.{
-        .allocator = allocator,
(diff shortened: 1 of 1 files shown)


dadepo (approved):

Rexicon226:
What are the logs that are actually happening? Why not just use a no-op logger by default and manually swap it when you need to debug something?

0xNineteen (approved):

dnut:
> What are the logs that are actually happening? Why not just use a no-op logger by default and manually swap it when you need to debug something?

The benchmarking framework itself outputs log messages

PR #427, opened 2024-12-06
fix(cmd): move resolveSocketAddr into net

0xNineteen:
Moved socket address resolution logic from cmd.zig to net.zig and made it more reusable

PR #428, opened 2024-12-06
fix(sig): make cmd/ folder accessible from sig.zig

0xNineteen:
various improvements and separations of methods in cmd.zig

- Moved cmd.zig identity management logic into dedicated helpers module
- Added new cluster configuration enum (to core/lib.zig) and entrypoint management (into gossip/service.zig)
- Consolidated socket address string handling
- Extracted geyser writer creation into separate function
- Added status cache initialization from directory
- Improved IP and shred version discovery from echo server


0xNineteen:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Syndica/sig/428?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#428** <a href="https://app.graphite.dev/github/pr/Syndica/sig/428?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Syndica/sig/428?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#427** <a href="https://app.graphite.dev/github/pr/Syndica/sig/427?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

dnut:
Why is this needed? It seems like a red flag if other parts of the code depend on cmd.

0xNineteen:
cmd has a bunch of useful setup/helper code which we should have accessible to the rest of the repo

dnut:
The link you posted is code in gossip.

There's probably too much code in cmd and it should be moved into various other places.

0xNineteen:
sorry should have been more specific - the main point of the PR is to show the usage of the helper functions in cmd/ which needed to be imported to be used -- ie, 
```
const getMyDataFromIpEcho = sig.cmd.cmd.getMyDataFromIpEcho;
const getOrInitIdentity = sig.cmd.helpers.getOrInitIdentity;
```

dnut:
I think reusing the logic from cmd could be done more cleanly, if that's the end goal. But I'm assuming this is just the fastest way to get something specific working, which then raises the question of what you're trying to get working. I think it would be easier to review this stack of prs as a single PR, since each change requires the broader context to justify it.

0xNineteen:
@dnut this PR addresses your comments from #429  - should be good for review

src/net/echo.zig
```diff
@@ -1,18 +1,56 @@
...
+    var my_ip: ?IpAddr = null;
+    var my_shred_version: ?u16 = null;
+
+    for (socket_addresses) |socket_addr| {
+        if (requestIpEcho(allocator, socket_addr.toAddress(), .{})) |response| {
+            // ip is optional, but we should have it if given in response
+            if (my_ip == null)
+                my_ip = response.address;
+
+            // shred version is always mandatory (so we break)
```
dnut, line 36: I find this comment a little confusing. I don't think I would break because I've acquired something that is *mandatory*. I would break once I've received something that is *sufficient*, which is as much a statement about the ip address as it as about the shred version.

If shred version is strictly mandatory, then it's confusing to make it optional in the return type.

I feel like it would be better if this function made its best effort to provide both values. What's the benefit of breaking when there is no IP? Why not just keep iterating over the provided addresses until you either exhaust all options, or acquire both pieces of data?

src/cmd/config.zig
```diff
@@ -67,53 +71,43 @@ const GossipConfig = struct {
...
+        self: GossipConfig,
+        allocator: std.mem.Allocator,
+    ) ![]SocketAddr {
+        var entrypoint_set = std.AutoArrayHashMap(SocketAddr, void).init(allocator);
+        defer entrypoint_set.deinit();
+
+        // add cluster entrypoints
+        if (try self.getCluster()) |cluster| {
+            for (sig.gossip.getClusterEntrypoints(cluster)) |entrypoint| {
+                const socket_addr = try resolveSocketAddr(allocator, entrypoint);
```
dnut, line 92: It's a little unexpected to find network io in this file, but I guess it's fine.

src/core/lib.zig
```diff
@@ -30,3 +30,5 @@ pub const MessageHeader = transaction.MessageHeader;
 pub const Transaction = transaction.Transaction;
 pub const VersionedTransaction = transaction.VersionedTransaction;
 pub const V0Message = transaction.V0Message;
+
+pub const Cluster = enum { mainnet, devnet, testnet, localnet };
```
dnut, line 34: You don't need to change this now, but just some food for thought....

I think we may want to consolidate this somewhat with the ClusterType in accountsdb. We could keep the enum and union separate, but this enum could be the tag for that union. What do you think?

Maybe one issue is that there are some weird variants in the one in accountsdb. Development and Localhost seem redundant for example. I would say it's more important to support "custom" than either one of those, but maybe it's a good idea to explicitly have a localhost.

src/cmd/cmd.zig
```diff
@@ -1216,47 +1200,63 @@ pub fn testTransactionSenderService() !void {
...
+        const echo_data = try getShredAndIPFromEchoServer(
+            logger.unscoped(),
+            allocator,
+            entrypoints,
+        );
+        const my_shred_version = echo_data.shred_version orelse 0;
+        logger.info().logf("my shred version: {d}", .{my_shred_version});
+
+        const config_ip = config.current.gossip.getHost() catch null;
+        const my_ip = config_ip orelse echo_data.ip orelse IpAddr.newIpv4(127, 0, 0, 1);
```
dnut, line 1244:
```suggestion
        const my_ip = config.current.gossip.getHost() catch
            echo_data.ip orelse IpAddr.newIpv4(127, 0, 0, 1);
```

dnut (dismissed):
Overall this looks good, I just had some various thoughts I wanted to highlight. After you take a look at my comments, I think this is good to merge.

src/accountsdb/snapshots.zig
```diff
@@ -1834,6 +1834,13 @@ pub const StatusCache = struct {
         return readFromFile(allocator, status_cache_file);
     }
 
+    /// opens the status cache using path {dir}/snapshots/status_cache
+    pub fn initFromDir(allocator: std.mem.Allocator, dir: std.fs.Dir) !StatusCache {
+        const status_cache_file = try dir.openFile("snapshots/status_cache", .{});
```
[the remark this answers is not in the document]
0xNineteen: thats a great point - i think a general re-thinking of the directory names / variable names needs to be re-thought - `snapshots_dir` is what contains the snapshot files, ie, the .zstd.tar files (and the unpacked values since we unpack to the same directory) -- which is usually 'validator/accounts_db/'-- because of the way agave packs snapshots, directly unpacking leads to another 'snapshots/' folder being created which can cause confusion -- on top of that, we store other folders in the 'accounts_db' directory which is typically where snapshots are stored leading to more confusion

will leave for now but will create a ticket to reference this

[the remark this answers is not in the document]
0xNineteen: https://github.com/orgs/Syndica/projects/2/views/10?pane=issue&itemId=90745328

src/net/echo.zig
0xNineteen: true, i was also thinking this but the previous code followed this logic so i thought i would minimize the changes and follow the same approach - but agreed we should keep iterating and provide the best effort

src/core/lib.zig
0xNineteen: yeah was thinking this as well but wasnt sure how - ill try the union tag approach

0xNineteen:
kinda messed up the commits by accident but addressed all your comments

src/cmd/cmd.zig
```diff
@@ -735,7 +735,13 @@ fn validator() !void {
         allocator.destroy(gossip_service);
     }
 
-    const geyser_writer = try buildGeyserWriter(allocator, app_base.logger.unscoped());
+    const geyser_writer: ?*GeyserWriter = createGeyserWriterFromConfig(
+        allocator,
+        config.current.geyser,
+    ) catch |err| switch (err) {
+        error.GeyserWriterIsDisabled => null,
```
dadepo, line 742: Nit: Previously implementation logs `GeyserWriter is disabled.` when geyser is disabled.

src/cmd/config.zig
```diff
@@ -67,53 +71,43 @@ const GossipConfig = struct {
...
+        // add config entrypoints
+        for (self.entrypoints) |entrypoint| {
+            const socket_addr = SocketAddr.parse(entrypoint) catch brk: {
+                break :brk try resolveSocketAddr(allocator, entrypoint);
+            };
+            try entrypoint_set.put(socket_addr, {});
+        }
+
+        const entrypoints = try allocator.dupe(SocketAddr, entrypoint_set.keys());
+        return entrypoints;
```
dadepo, line 106: Nit:

```suggestion
        return try allocator.dupe(SocketAddr, entrypoint_set.keys());
```

src/net/net.zig
```diff
@@ -259,16 +259,7 @@ pub const SocketAddr = union(enum(u8)) {
...
-    /// - array: the string, plus some extra bytes at the end
-    /// - integer: length of the string within the array
-    pub fn toString(self: Self) struct { [53]u8, usize } {
-        var buf: [53]u8 = undefined;
-        const len = self.toStringBuf(&buf);
-        return .{ buf, len };
-    }
-
-    pub fn toStringBounded(self: Self) std.BoundedArray(u8, 53) {
+    pub fn toString(self: Self) std.BoundedArray(u8, 53) {
```
dadepo, line 262: For my own understanding. What advantage does returning `std.BoundedArray(u8, 53)` has over `struct { [53]u8, usize }`?

Also could it be possible to document why capacity of 53?

dadepo (approved):

src/net/net.zig
0xNineteen: if you look at the implementation of BoundedArray youll see its actually the same as the struct  { ... } approach we had previously

src/cmd/cmd.zig
```diff
@@ -1160,15 +1144,15 @@ pub fn testTransactionSenderService() !void {
     }
 
     // define cluster of where to land transactions
-    const cluster: ClusterType = if (try config.current.gossip.getNetwork()) |n| switch (n) {
+    const rpc_cluster: ClusterType = if (try config.current.gossip.getCluster()) |n| switch (n) {
```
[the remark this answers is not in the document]
0xNineteen: yeah its a bit strange - see my comment in drew's review

src/cmd/cmd.zig
```diff
@@ -1504,8 +1428,8 @@ fn loadSnapshot(
 
     // this should exist before we start to unpack
     logger.info().log("reading genesis...");
-    result.genesis_config = readGenesisConfig(allocator, genesis_file_path) catch |err| {
-        if (err == error.GenesisNotFound) {
+    result.genesis_config = GenesisConfig.init(allocator, genesis_file_path) catch |err| {
```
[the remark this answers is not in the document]
0xNineteen: GenesisNotFound and FileNotFound are the same thing - this makes things simpler keeping it to just file not found

PR #433, opened 2024-12-11
fix(rpc): skip rpc client tests

dnut:
these tests are blocking all of our other prs. a follow-up pr can actually convert these tests into something that works locally without internet access

### DIFF
--- a/src/rpc/client.zig
+++ b/src/rpc/client.zig
@@ -415,166 +415,166 @@ pub const Client = struct {
     }
 };
 
-test "getAccountInfo: null value" {
-    const allocator = std.testing.allocator;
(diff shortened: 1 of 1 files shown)


InKryption:
I would instead suggest adding `if (true) return error.SkipZigTest;` instead of commenting them out, in order to have it noted in the test run that tests were skipped, as a reminder to eventually re-enable them.

dnut:
@InKryption 

[84c5283](https://github.com/Syndica/sig/pull/433/commits/84c52835aca6e2bf4d84e9e6507ac490695100d6)

InKryption (approved):

PR #434, opened 2024-12-12
Delete Compression, improve parsing, reorder stuff and rename a few things and rename a few things

InKryption:

PR #436, opened 2024-12-12
refactor(snapshots): Improve `SnapshotFiles.find`

InKryption:
Improve `SnapshotFiles.find`

Delete unused `readDirectory`


InKryption:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Syndica/sig/436?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#436** <a href="https://app.graphite.dev/github/pr/Syndica/sig/436?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Syndica/sig/436?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#435** <a href="https://app.graphite.dev/github/pr/Syndica/sig/435?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#434** <a href="https://app.graphite.dev/github/pr/Syndica/sig/434?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

0xNineteen:
i feel like we can name SnapshotFiles better (there are a lot of "files" pertaining to snapshots) - maybe something like SnapshotZstdTarFiles ? wdyt

src/accountsdb/snapshots.zig
```diff
@@ -2215,44 +2214,72 @@ pub const SnapshotFiles = struct {
...
+            if (current.base_slot != latest_full.slot) continue;
+            const prev = maybe_latest_incremental orelse {
+                maybe_latest_incremental = current;
+                continue;
+            };
+            if (prev.slot < current.slot) {
+                maybe_latest_incremental = current;
+                continue;
+            }
+            if (prev.slot == current.slot) {
```
0xNineteen, line 2271: why would this be an error?

src/accountsdb/snapshots.zig
```diff
@@ -2215,44 +2214,72 @@ pub const SnapshotFiles = struct {
...
-                // this unwrap is safe because count > 0
-                snapshot.slot > maybe_latest_incremental_snapshot.?.slot))
-            {
-                maybe_latest_incremental_snapshot = snapshot;
+        const latest_full = maybe_latest_full orelse return error.NoFullSnapshotFileInfoFound;
+
+        var maybe_latest_incremental: ?IncrementalSnapshotFileInfo = null;
+        for (incremental_snaps.items) |current| {
+            if (current.base_slot != latest_full.slot) continue;
+            const prev = maybe_latest_incremental orelse {
```
0xNineteen, line 2263: would prefer to avoid 'prev' and 'current' naming when all we need is the incremental snapshot and the latest incremental snapshot

src/accountsdb/snapshots.zig
```diff
@@ -2215,44 +2214,72 @@ pub const SnapshotFiles = struct {
...
-            // need to match the base slot
-            if (snapshot.base_slot == latest_full_snapshot.slot and (count == 0 or
-                // this unwrap is safe because count > 0
-                snapshot.slot > maybe_latest_incremental_snapshot.?.slot))
-            {
-                maybe_latest_incremental_snapshot = snapshot;
+        const latest_full = maybe_latest_full orelse return error.NoFullSnapshotFileInfoFound;
+
+        var maybe_latest_incremental: ?IncrementalSnapshotFileInfo = null;
+        for (incremental_snaps.items) |current| {
```
0xNineteen, line 2261:
```suggestion
        for (incremental_snapshots.items) |incremental| {
```

src/accountsdb/snapshots.zig
```diff
@@ -2215,44 +2214,72 @@ pub const SnapshotFiles = struct {
...
+            if (IncrementalSnapshotFileInfo.parseFileNameTarZst(filename)) |inc_snap| {
+                if (maybe_latest_full) |latest_full| {
+                    if (inc_snap.slot < latest_full.slot) continue;
+                    if (inc_snap.base_slot < latest_full.slot) continue;
+                }
+                try incremental_snaps.append(allocator, inc_snap);
+                continue;
+            } else |_| {}
+
+            const current = FullSnapshotFileInfo.parseFileNameTarZst(filename) catch continue;
```
0xNineteen, line 2241: current/prev makes this more confusing imo - see other comment

0xNineteen (changes requested):

src/accountsdb/snapshots.zig
InKryption: If they're the same slot, it means they're different hashes, since they're different file names. If there are two snapshots with the same snapshot range, but different hashes, it's not possible to unambiguously select one; rather call it out than just defer to some arbitrary behavior like selecting the first one that the filesystem reports.

src/accountsdb/snapshots.zig
InKryption: 58aa0bf0103d4bc4cefde2059d862f1fc9c0467c

src/accountsdb/snapshots.zig
InKryption: The prev and current naming is useful for quickly understanding the relationship between the two; how about using it as a suffix?

src/accountsdb/snapshots.zig
0xNineteen: good point - can we add a comment explaining this above the if statement

src/accountsdb/snapshots.zig
0xNineteen: still says current, but assuming it relates to the other comment so will resolve this

src/accountsdb/snapshots.zig
InKryption: We care because one of these variables is the one that is currently selected, the other is the one we're comparing against, naming them too similarly is an easy way to misunderstand code; they can't both be called just "incremental". What would you suggest instead?

0xNineteen: one would be called `latest_incremental` and the other is called `incremental`

InKryption: "incremental" shadows the method `incremental`; how about `latest_info` and `info`, since we're clearly looping over the "incremental infos"?

0xNineteen: i feel like this is an okay time to use `_incremental` -- its done elsewhere in the codebase too when naming conflicts happen like this - infos not great because you have to remember what method you are in

InKryption: 4ebc727e2269d93e845611ff5083e141176c0dca

0xNineteen (approved):

PR #422, opened 2024-12-04
refactor(shred-networking): move turbine into shred_collector 

yewman:
Consolidates logic associated with the transmission of shreds in Solana by  moving the implementation of turbine tree and retransmit stage into the shred collector module, which is renamed to **shred_networking** 

Changes in this PR are kept **minimal**, refactoring of data-structures and dependencies can be addressed in follow up PR's.

PR #437, opened 2024-12-13
fix(shred-networking): Minor fixes after move turbine into shred_collector 

dadepo:
Address the comments post merged on https://github.com/Syndica/sig/pull/422


0xNineteen (approved):
nice

PR #438, opened 2024-12-13
fix(ledger): avoid rocksdb segfault

dnut:
A segfault occurs in RocksDB when built in ReleaseSafe. This can be reproduced with a minimal example of code that only depends on the C library. It does not segfault when RocksDB is built using cmake. It does not segfault in any other zig optimize mode.

This PR is a quick fix to avoid the segfault by preventing RocksDB from being built by zig with ReleaseSafe. Whenever ReleaseSafe is specified as the build mode, RocksDB will be built with ReleaseFast. Otherwise it will use the same optimize mode as the build as a whole.


build.zig
```diff
@@ -46,7 +46,11 @@ pub fn build(b: *Build) void {
     const curl_dep = b.dependency("curl", dep_opts);
     const curl_mod = curl_dep.module("curl");
 
-    const rocksdb_dep = b.dependency("rocksdb", dep_opts);
+    const rocksdb_dep = b.dependency("rocksdb", .{
```
dadepo, line 49: I think this is fine, if building a dependency in a different release mode like this won't also lead to some other issues?

dadepo (approved):
LGTM as a quick fix.

PR #444, opened 2024-12-16
fix(ledger): use rocksdb version containing fix for llvm bug

dnut:
this also removes the hack to avoid ReleaseSafe


Rexicon226 (approved):
LGTM

PR #446, opened 2024-12-16
fix(build): support m4 cpus

kprotty:
Reuse the `defaultTargetDetectM3()` workaround for m4 cpus as well.


0xNineteen (approved):

0xNineteen (approved):

0xNineteen:
@kprotty seems like your commits need to be signed

PR #448, opened 2024-12-17
perf: replace busy-wait channel receives with receiveTimeout

dnut:
Every thread in sig that receives on a channel utilizes a CPU core at 100% from busy-waiting in the receive loop. This PR eliminates the busy-waiting by switching from `tryReceive` to `receiveTimeout` in those places. After this change, I see a dramatic decrease in CPU usage while running sig.

I did not remove `tryReceive` from defer statements or unit tests. It does not seem appropriate to use blocking calls in those contexts if non-blocking calls are already known to be sufficient.

### Handling `error.Closed`

Unlike `tryReceive`, `receiveTimeout` can return an error to indicate that the channel is closed. I needed to make a decision about how to handle this error. To replicate the same behavior as `tryReceive`, the error should simply be ignored with `catch null` in every scenario. But I didn't take this approach in most cases.

`error.Closed` provides new information that we didn't have before, and we can use that information to our advantage on a case-by-case basis. If that information can lead us to a definitive conclusion that it would be pointless to remain in a particular context, then I've changed the code to exit from that context. I've done this conservatively: the control flow will remain in a context if there is any possibility of doing more useful work in that context.

I used the following where appropriate:
- `catch return`: There is no more meaningful work that can be done in this function if the channel is closed.
- `catch break`: The current scope has nothing else to do, but there is some additional work that can be done in the function after exiting the current scope. 
- `catch null`: There is meaningful work that can be done within the current scope even if the channel is closed. Often this is because we collect a batch of items from the channel before processing any of them.

### DIFF
--- a/scripts/style.py
+++ b/scripts/style.py
@@ -206,8 +206,9 @@ def line_length(args, files_to_check):
             # are "//"
             if line.strip().startswith("//"):
                 continue
-            if len(line) > MAX_LINE_LENGTH:
-                print(f"{path}:{i + 1} is too long: {len(line)}")
+            line_length = len(line) - 1  # -1 for \n
+            if line_length > MAX_LINE_LENGTH:
+                print(f"{path}:{i + 1} is too long: {line_length}")
                 lines_found += 1
                 if path not in unique_files:
                     unique_files[path] = 1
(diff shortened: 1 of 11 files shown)


dnut:
closing since this has some performance issues, and some other improvements should occur before blocking receive is used. for example #465

PR #452, opened 2024-12-18
fix(ledger): fuzz testing

dadepo:


dadepo:
```
$ zig build fuzz -- ledger_rocksdb 10 1_000_000 > output.txt 2>&1
$ grep -c 'Put actions' output.txt
890
$ grep -c 'Delete actions' output.txt
889
$ grep -c 'Delete Files in Range actions' output.txt
925
$ grep -c 'getBytes actions' output.txt
10
$ grep -c 'get actions' output.txt
11
$ grep -c 'count actions' output.txt
997
$ grep -c 'contains actions' output.txt
11
$ grep -c 'Batch actions' output.txt
904
```

`get`, `getBytes` and `contains` are order of magnitude slower. Need to investigate

src/ledger/fuzz.zig
```diff
@@ -0,0 +1,330 @@
...
+    .name = "data",
+    .Key = u64,
+    .Value = Data,
+};
+pub const BlockstoreDB = switch (build_options.blockstore_db) {
+    .rocksdb => ledger.database.RocksDB(&.{cf1}),
+    .hashmap => ledger.database.SharedHashMapDB(&.{cf1}),
+};
+
+const build_options = @import("build-options");
```
0xNineteen, line 25: can we keep all imports at the top

src/ledger/fuzz.zig
```diff
@@ -0,0 +1,330 @@
...
+    const rocksdb_path =
+        try std.fmt.allocPrint(allocator, "{s}/ledger/rocksdb", .{sig.FUZZ_DATA_DIR});
+
+    // ensure we start with a clean slate.
+    if (std.fs.cwd().access(rocksdb_path, .{})) |_| {
+        try std.fs.cwd().deleteTree(rocksdb_path);
+    } else |_| {}
+    try std.fs.cwd().makePath(rocksdb_path);
+
+    var db: BlockstoreDB = try BlockstoreDB.open(
```
0xNineteen, line 60:
```suggestion
    var db = try BlockstoreDB.open(
```

src/ledger/fuzz.zig
```diff
@@ -0,0 +1,330 @@
...
+        allocator,
+        logger,
+        rocksdb_path,
+    );
+
+    defer db.deinit();
```
0xNineteen, line 66:
```suggestion
    defer db.deinit();
```

src/ledger/fuzz.zig
```diff
@@ -0,0 +1,330 @@
...
+    max_actions: ?usize,
+) !void {
+    const start = random.int(u32);
+    const end = blk: {
+        const end_ = random.int(u32);
+        if (end_ < start)
+            break :blk (end_ +| start)
+        else
+            break :blk end_;
+    };
```
0xNineteen, line 207: since the values are random -- i think this is the same but simpler
```suggestion
    const start = random.int(u32);
    const end = random.int(u32) +| start;
```

src/ledger/fuzz.zig
```diff
@@ -0,0 +1,330 @@
...
+            }
+        }
+        const start = random.int(u32);
+        const end = blk: {
+            const end_ = random.int(u32);
+            if (end_ < start)
+                break :blk (end_ +| start)
+            else
+                break :blk end_;
+        };
```
0xNineteen, line 302: same here

src/ledger/fuzz.zig
```diff
@@ -0,0 +1,330 @@
...
+        for (0..buffer.len) |i| {
+            buffer[i] = @intCast(random.int(u8));
+        }
+
+        const value: []const u8 = buffer[0..];
+
+        var batch = try db.initWriteBatch();
+        defer batch.deinit();
+
+        try batch.put(cf1, key, Data{ .value = value });
```
0xNineteen, line 317: why have a put in a delete range?

0xNineteen (changes requested):
one thing thats great about this is that it performs many actions in parallel and if something errors it will break and catch it - however, theres no consistency check that the values which are read, are what they should be - which i think we care about - since we have multiple database backends, i think this should be simple to do: perform the same operations and after each 'step' ensure they are returning equivalent values -- the accountsdb fuzzer does this which could be a reference

dnut (changes requested):
Ideally a fuzzer would be deterministic and reproducible. When a fuzz test fails you should be able to use the same seed to trigger the same error and debug what went wrong. But in this case, the multithreaded behavior is random and not reproducible by using the same seed. I do see the value of identifying issues with concurrency though, and I'm not sure how to test that in a reproducible way. Any thoughts?

Also I noticed stdout is flooded very fast with a ton of text when running these fuzz tests. It might be good to slow that down a bit. Normally, for fuzz tests, I would use stdout to print one line for each discrete test case that can be individually reproduced using a particular seed that is included on that line. Something on the order of 1 log message per second seems reasonable.

This test only deals with the database implementation which is a pretty small part of the ledger. In the future we should write some fuzz tests for the ShredInserter and BlockstoreReader, which is where I would expect to see more problems than RocksDB for example.

src/ledger/fuzz.zig
dadepo: https://github.com/Syndica/sig/pull/452/commits/c7efe51d70e31b6f8575e2298fa7793946060cb4

src/ledger/fuzz.zig
dadepo: https://github.com/Syndica/sig/pull/452/commits/52bbbb212263ef81abab187ef584661a398f4f94

src/ledger/fuzz.zig
dadepo: Can't find the exact commit but this has also been processed. The extra newline was removed.

src/ledger/fuzz.zig
dadepo: Applied as part of the commit that checks the methods return expected output

https://github.com/Syndica/sig/pull/452/commits/9e06b51485f7c019d6d2404e4ade42e270ac008a

src/ledger/fuzz.zig
dadepo: Applied as part of the commit that checks the methods return expected output

https://github.com/Syndica/sig/pull/452/commits/9e06b51485f7c019d6d2404e4ade42e270ac008a

src/ledger/fuzz.zig
dadepo: This is because the batch API is being performed, which involves batching the put and the delete actions.

src/ledger/fuzz.zig
```diff
@@ -0,0 +1,330 @@
...
+        defer db_contains_thread.join();
+
+        // Batch API
+        var batch_delete_range_thread = try std.Thread.spawn(
+            .{},
+            batchDeleteRange,
+            .{ &db, &random, &total_action_count, maybe_max_actions },
+        );
+        defer batch_delete_range_thread.join();
+    }
```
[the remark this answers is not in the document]
dadepo: Applied while removing the concurrency. https://github.com/Syndica/sig/pull/452/commits/b419b4becebf3ad18e872e4005915161f9f2f91a

src/ledger/fuzz.zig
```diff
@@ -0,0 +1,330 @@
...
+    const maybe_max_actions = blk: {
+        if (maybe_max_actions_string) |max_actions_str| {
+            break :blk try std.fmt.parseInt(usize, max_actions_str, 10);
+        } else {
+            break :blk null;
+        }
+    };
+    defer {
+        _ = total_action_count.fetchAdd(1, .monotonic);
+    }
```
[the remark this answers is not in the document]
dadepo: https://github.com/Syndica/sig/pull/452/commits/9cfe206d3892210badb757d709a503a1f37e6f32

> Also, what is the purpose of this defer fetchAdd?

That was a mistake. Removed.

src/ledger/fuzz.zig
```diff
@@ -0,0 +1,330 @@
...
+        allocator,
+        .debug,
+    );
+    const logger = std_logger.logger();
+
+    var prng = std.rand.DefaultPrng.init(seed);
+    const random = prng.random();
+
+    const rocksdb_path =
+        try std.fmt.allocPrint(allocator, "{s}/ledger/rocksdb", .{sig.FUZZ_DATA_DIR});
```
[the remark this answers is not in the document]
dadepo: https://github.com/Syndica/sig/pull/452/commits/076d4b323de44d8407ebaa3520d824a53db535de

src/ledger/fuzz.zig
```diff
@@ -0,0 +1,330 @@
...
+) !void {
+    const key = random.int(u32);
+    var buffer: [61]u8 = undefined;
+    // Fill the buffer with random bytes
+    for (0..buffer.len) |i| {
+        buffer[i] = @intCast(random.int(u8));
+    }
+    const value: []const u8 = buffer[0..];
+    try performDbAction(
+        "RocksDb.put",
```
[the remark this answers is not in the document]
dadepo: https://github.com/Syndica/sig/pull/452/commits/076d4b323de44d8407ebaa3520d824a53db535de

src/ledger/fuzz.zig
```diff
@@ -0,0 +1,330 @@
...
+            last_print_msg_count = current_count;
+        }
+
+        _ = count.fetchAdd(1, .monotonic);
+    }
+}
+
+fn dbPut(
+    db: *BlockstoreDB,
+    random: *const std.rand.Random,
```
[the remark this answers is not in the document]
dadepo: https://github.com/Syndica/sig/pull/452/commits/3e07071e964b0ed5a4f31f7aa7e0345285187829

src/ledger/fuzz.zig
```diff
@@ -0,0 +1,330 @@
...
+    while (true) {
+        if (max_actions) |max| {
+            if (count.load(.monotonic) >= max) {
+                std.debug.print("{s} reached max actions: {}\n", .{ action_name, max });
+                break;
+            }
+        }
+
+        _ = try @call(.auto, func, args);
+        const current_count = count.load(.monotonic);
```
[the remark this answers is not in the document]
dadepo: All of the atomics has been removed as part of dropping the concurrency https://github.com/Syndica/sig/pull/452/commits/b419b4becebf3ad18e872e4005915161f9f2f91a

src/ledger/fuzz.zig
```diff
@@ -0,0 +1,330 @@
...
+    db: *BlockstoreDB,
+    random: *const std.rand.Random,
+    count: *std.atomic.Value(u64),
+    max_actions: ?usize,
+) !void {
+    const key = random.int(u32);
+    try performDbAction(
+        "RocksDb.contains",
+        BlockstoreDB.contains,
+        .{ db, cf1, key },
```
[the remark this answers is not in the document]
dadepo: > I think it would be a more thorough test if it actually returned true sometimes.

I think this should now be the case.

src/ledger/fuzz.zig
0xNineteen: its still there 
```
    var db = try createBlockstoreDB();

    defer db.deinit();
```

src/ledger/fuzz.zig
```diff
@@ -0,0 +1,466 @@
...
+        // - batch.put
+        // - batch.delete
+        // - batch.deleteRange
+        batchAPI,
+    };
+
+    inline for (functions) |function| {
+        const fn_args = .{ &db, random, maybe_max_actions };
+        _ = try @call(.auto, function, fn_args);
+    }
```
0xNineteen, line 87: i feel like it would be better to have a similar approach to accountsdb where you have a enum over all functions and then randomly select which function to run and have a single max-actions to run would be better/more random

src/ledger/fuzz.zig
```diff
@@ -0,0 +1,466 @@
...
+};
+
+const cf1 = ColumnFamily{
+    .name = "data",
+    .Key = u64,
+    .Value = Data,
+};
+
+var dataMap = std.AutoHashMap(u32, Data).init(allocator);
+var dataKeys = std.ArrayList(u32).init(allocator);
```
0xNineteen, line 21: i dont think we actually need this - we can just use dataMap.keys() when we need to

src/ledger/fuzz.zig
```diff
@@ -0,0 +1,466 @@
...
+    value: []const u8,
+};
+
+const cf1 = ColumnFamily{
+    .name = "data",
+    .Key = u64,
+    .Value = Data,
+};
+
+var dataMap = std.AutoHashMap(u32, Data).init(allocator);
```
0xNineteen, line 20: might want to add a doc-comment above this explaning that this is a simpler blockstore which we will use to make sure our data is up to date -- also since its initialized it should be `data_map`

src/ledger/fuzz.zig
```diff
@@ -0,0 +1,466 @@
...
+                    buffer[i] = @intCast(random.int(u8));
+                }
+
+                const value: []const u8 = try allocator.dupe(u8, buffer[0..]);
+                const data = Data{ .value = value };
+
+                try batch.put(cf1, key, data);
+            }
+            try batch.deleteRange(cf1, startKey, endKey);
+            // Commit batch put and delete range.
```
0xNineteen, line 451: if these methods dont modify the dataMap - then we should probably add a note saying there is no confirmation in these methods

0xNineteen (changes requested):

src/ledger/fuzz.zig
dadepo: > i feel like it would be better to have a similar approach to accountsdb where you have a enum over all functions and then randomly select which function to run 

I considered this, but the reasoning why I did not go with that approach was because I felt the randomness should be in how the methods are called (ie the inputs passed to them) and not if the methods we are called - which would be the case if the fuzz tests randomly selects the methods to execute while fuzzing.

The advantage of going with the enum approach is for dependencies between the method calls and if randomizing that would expose any hidden bugs.

I think it should be possible to adapt what is there now to that. I'll give it a shot

src/ledger/fuzz.zig
dadepo: I actually had that but had some issues. So just quickly went with keeping a separate list. I'll revisit this and see if I can drop it

src/ledger/fuzz.zig
dnut: >  I felt the randomness should be in how the methods are called (ie the inputs passed to them) and not if the methods we are called

why not both? more variability here seems like a bonus

src/ledger/fuzz.zig
dadepo: > I actually had that but had some issues.

It was the fact that `AutoHashMap` don't have `.keys()` method, but it only returns a keyIterator. So added a separate list to quickly get going

Dropped it in [c695ce7](https://github.com/Syndica/sig/pull/452/commits/c695ce776cf0e94aa2a502233de86c3696196a62)

src/ledger/fuzz.zig
dadepo: [e29fef7](https://github.com/Syndica/sig/pull/452/commits/e29fef7149e4106793db48735e2411d147b9fb94)

src/ledger/fuzz.zig
dadepo: [0688ef9](https://github.com/Syndica/sig/pull/452/commits/0688ef9b24d019d4b10ab23d0ccdebcdf0d9a7dd)

src/ledger/fuzz.zig
dadepo: [c695ce7](https://github.com/Syndica/sig/pull/452/commits/c695ce776cf0e94aa2a502233de86c3696196a62)

src/ledger/fuzz.zig
0xNineteen: should still be `data_map`

0xNineteen (dismissed):
lgtm :shipit:

src/ledger/fuzz.zig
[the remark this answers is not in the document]
dnut: FYI `std.rand` is an alias for `std.Random`. Random contains an alias to itself only for compatibility with code written for old versions where `std.rand` was a separate namespace.

src/ledger/fuzz.zig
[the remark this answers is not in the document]
dnut: Now it's going to return true 100% of the time instead of 0% of the time. I was suggesting that the behavior should be variable. 50% of the time would be nice.

dnut:
> > Also I noticed stdout is flooded very fast with a ton of text when running these fuzz tests. It might be good to slow that down a bit. Normally, for fuzz tests, I would use stdout to print one line for each discrete test case that can be individually reproduced using a particular seed that is included on that line. Something on the order of 1 log message per second seems reasonable.
> 
> This should be addressed now. I got the idea for logging every 1000 count from another fuzzer in the codebase. But I've dropped it now.

You've removed all printing. I was suggesting that there should be discrete test cases that are actually printed.

Here's the problem I'm trying to address. I just ran the fuzzer for a long time. Maybe 20 minutes. The fuzzer eventually crashed due to an error in a test. I'd like to reproduce this, but the only way to do that is by running the fuzzer again for 20+ minutes. Identifying the cause of this failure will not be easy.

If instead the fuzzer ran many distinct and isolated test cases, and it printed the seed for each case, it would be easy to rerun a single one of those cases when they fail. You can look at the allocator fuzzer to see what I mean.

dadepo:
> If instead the fuzzer ran many distinct and isolated test cases, and it printed the seed for each case, it would be easy to rerun a single one of those cases when they fail. You can look at the allocator fuzzer to see what I mean.

The allocator fuzzer seems to have a different structure. It allows fuzzing two different implementations (disk or batch) or both and provides ability to select which to run. 

This is different from the structure for the ledger fuzzing as it fuzzes various methods on the same ledger implementation. And the pattern used for it is similar to what is used for the accounts db, and gossip fuzzer.

> Here's the problem I'm trying to address. I just ran the fuzzer for a long time. Maybe 20 minutes. The fuzzer eventually crashed due to an error in a test. I'd like to reproduce this, but the only way to do that is by running the fuzzer again for 20+ minutes. Identifying the cause of this failure will not be easy.

I am not sure if it is possible to shortcut this, as it may be that it requires that much time for the random interaction of the methods being fuzzed to trigger the issue.

dadepo:
> I got the idea for logging every 1000 count from another fuzzer in the codebase. But I've dropped it now.

Checked again. This is was from `gossip_fuzz_service`.

dnut:
> The allocator fuzzer seems to have a different structure. It allows fuzzing two different implementations (disk or batch) or both and provides ability to select which to run.

This is not the relevant part of the fuzzer that I'm talking about. You can say the same thing about the ledger fuzzer allowing multiple databases. It just happens to deal with that abstraction at comptime whereas the allocator fuzzer does it at runtime. But this is a separate concern.

> This is different from the structure for the ledger fuzzing as it fuzzes various methods on the same ledger implementation. And the pattern used for it is similar to what is used for the accounts db, and gossip fuzzer.

This is exactly the same pattern followed by the allocator fuzzer.

> > Here's the problem I'm trying to address. I just ran the fuzzer for a long time. Maybe 20 minutes. The fuzzer eventually crashed due to an error in a test. I'd like to reproduce this, but the only way to do that is by running the fuzzer again for 20+ minutes. Identifying the cause of this failure will not be easy.
> 
> I am not sure if it is possible to shortcut this, as it may be that it requires that much time for the random interaction of the methods being fuzzed to trigger the issue.

I agree it's not going to be possible to shortcut that specific scenario. The idea is to create different scenarios that can be reproduced

dadepo:
> instead it could look like something like this:
> 
> ```zig
> while (true) {
>      seed += 1;
>      print("using seed: {}", .{seed});
>      const random = rng(seed).random();
>      for (0..test_case_size) |_| {
>           runIteration(random);
>      }
> }
> ```

I am not sure I follow the intended benefit here. Given that the number of runs depends on `test_case_size` what this modification does is to add 1 to the provided seed and used that to run the iteration.  

If the intention is to get to the seed that caused the issue faster,  I am also not sure if the proposal solves the original problem statement. Given that if an error is triggered at `seed:10`, that does not make it easier to be reproduced using that `seed:10` directly. This is because the test is stateful, and the error case that appeared at `seed:10` might be because of the 9 previous seed runs.

Also to be sure I am on the same page with the proposed improvement, is the current ledger fuzzer implementation similarly in structure to exiting fuzzer like accounts_db and gossip? and is the proposed modification something that would also need to be applied to existing fuzzer? If that is the case, what about discussing the proposed improvement separately (as I am still not sure I follow how it solves the intended problem)  - and then apply to all the other fuzzers if it improves things?

dnut:
It solves the problem if you reset the state for each test case. I was just giving a basic pseudocode example to express the general idea. I wasn't trying to write the code that solves the problem in its entirety. At that point I think it's more efficient to just make the change in the code, so I committed it already.

Regarding gossip and accountsdb, I do think those fuzzers are flawed and need to be reworked, but that's not in scope right now. I just think we should not be adding more fuzzers that have the same flaw.

dnut (approved):

dnut (approved):

dadepo:
@dnut I think the recent changes might not be working as intended.

Running

`zig build fuzz -Dblockstore=rocksdb -- ledger 10 10 -freference-traced`

continually prints:

```
action_name reached max actions: 10
action_name reached max actions: 10
action_name reached max actions: 10
```

Until I kill the process. 

Also the `action_name` bit is not providing the intended information on which action is being executed.

dnut:
My mistake. I thought you added this to CI. I'll open a pr to add it to CI and fix the loop issue.

src/ledger/fuzz.zig
```diff
@@ -0,0 +1,344 @@
...
+    var count: u64 = 0;
+
+    while (true) outer: {
+        var prng = std.Random.DefaultPrng.init(seed);
+        const random = prng.random();
+        // This is a simpler blockstore which is used to make sure
+        // the method calls being fuzzed return expected data.
+        var data_map = std.AutoHashMap(u32, Data).init(allocator);
+        defer data_map.deinit();
+        for (0..1_000) |_| {
```
dadepo, line 97: @dnut The benefit of this change is that the seed used is not just the one passed in.  The observation though is that, this fixes the iteration for each seed at 1000. - which may seem arbitrary. 

It would probably be an improvement to have this configurable (which will differ from the other fuzzer) or

Do not change the seed value like this put have a random value generated by CI.

PR #455, opened 2024-12-19
feat(rpc-client): support getVoteAccounts

dnut:


yewman (commented):
Interested to get your thoughts in keeping to primitive data types where possible in the Rpc client. I know as it stands there are already a couple of instance in contradiction to this but we can tidy this up if you agree it makes sense.

src/rpc/client.zig
```diff
@@ -344,6 +344,25 @@ pub const Client = struct {
         return self.sendFetchRequest(allocator, types.RpcVersionInfo, request, .{});
     }
 
+    const GetVoteAccountsConfig = struct {
```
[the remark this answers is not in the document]
dnut: The whole point of the rpc client is to convert between strings and types. Otherwise the user might as well just use json directly with a plain old http client. Not using `Pubkey` here just seems to make client less useful, without any benefit that I can see.

I would be careful about where the types come from though. The RPC client should probably not depend on anything outside of `core`. Depending on plain old data types seems that define the general domain of Solana seems completely fine to me. But depending on other components like the ledger which depends on rocksdb, that's where you start running into actual costs.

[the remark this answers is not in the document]
yewman: In my mind the whole purpose of the rpc client is to provide a simple api for making rpc requests. As far as requests go I don't think the above example makes to client any less useful since it is just a couple of extra characters. For responses I also prefer to selectively parse into sig types when needed rather than parsing everything by default. At the end of the day though I don't really mind so if you want you can just go ahead and resolve this comment.

yewman (approved):

src/core/pubkey.zig
```diff
@@ -54,6 +57,13 @@ pub const Pubkey = extern struct {
     ) !void {
         return base58.format(self.data, writer);
     }
+
+    pub fn jsonParse(_: Allocator, source: anytype, _: ParseOptions) !Pubkey {
```
dadepo, line 61: Any reason to have the `jsonParse` function take an `Allocator` and `ParseOptions` even though these two are not used within the function?

src/rpc/types.zig
```diff
@@ -151,3 +151,19 @@ pub const RpcVersionInfo = struct {
 };
 
 pub const Signature = []const u8;
+
```
dadepo, line 154: More of a question: What goes into `src/rpc/types.zig` and what goes into `src/rpc/client.zig`. On a cursory look it looks as if the RPC responses are defined in `src/rpc/types.zig` while inpurs are defined inline in `src/rpc/client.zig`

Would that be a correct read of the pattern? If so, what the benefit to that vs having all types (both input and output) defined in `src/rpc/types.zig`

src/core/pubkey.zig
dnut: Yes. This function exists so that std.json can call it in a generic context. It always passes the same parameters to any instance of jsonParse.

src/rpc/types.zig
dnut: I don't know why it's organized this way. I agree with you about reorganizing this. I've been messing with this in my spare time and I have `methods` file that defines all the request and response types, and associates the types with each other. I'll probably open a PR with this pretty soon.

src/rpc/client.zig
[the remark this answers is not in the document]
dnut: > I don't really mind so if you want you can just go ahead and resolve this comment.

I appreciate this, but I'd like to get on the same page about how the rpc client should be designed overall. I still don't follow your reasoning. If you prefer, we can continue this discussion in another medium.

> the whole purpose of the rpc client is to provide a simple api for making rpc requests

But what does this mean? In this context, "providing an API" means you're allowing the user to work with pre-defined structs instead of strings. There's not much else going on here other than sending strings with an http client. Imagine if we removed all type conversions from this client. Then the rpc client is barely distinguishable from a generic http client. In that case, there would not be much of a point in having an rpc client library.

> I don't think the above example makes to client any less useful since it is just a couple of extra characters.

I could extend the same argument to `GetAccountInfoConfig`. Why don't we eliminate that struct and have the user pass a json string instead? It's just a couple extra characters. In my mind, using `Pubkey` follows the same logic as using the other structs used by the rpc client. If you're going to provide an rpc client that does the type conversations, why stop halfway? Why not convert everything into safer and more useful types?

PR #458, opened 2024-12-19
refactor(sync): adapt RcSlice for single-item pointers

dnut:
Most of the logic from RcSlice is moved into RcBase, which is a minimal and less opinionated version of RcSlice that does not contain the length of the pointer. RcBase is intended to be more composable than RcSlice. The parent struct containing RcBase becomes responsible for determining the length of the pointer.

RcBase serves as the implementation for both RcSlice, which contains any number of items, and a new struct Rc, which contains exactly one item.

This change is a dependency of #461


Sobeston (approved):
Would be nice to see some basic tests to show usage, but otherwise seems like a great api improvement.

PR #460, opened 2024-12-19
perf(shred-network): reduce gossip lock hold time

dnut:
Previously the gossip table lock was held for the entire function. It's only needed for this initial operation though.


yewman (approved):
lgtm

PR #461, opened 2024-12-19
feat: add Window and SharedPointerWindow data structures

dnut:
These changes are needed by #459, where they are used to track epoch states.

Previously I used an Lru to track epoch states, but the Lru requires exclusive locks on read operations, which introduces too much contention in a data structure that will be read often by many threads. The data structures in this PR were implemented to handle this use case a bit better. They can be read without an exclusive lock, and are intended to handle the predictable temporal nature of epoch transitions.

`Window` is the basic data structure that supports the idea of tracking a moving window of values.

`SharedPointerWindow` is a wrapper for `Window` that adds two features:
- thread safety via RwLock.
- manages the lifetimes of the contained data with reference counting.

`SharedPointerWindow` _could_ be implemented totally generically, enabling the user to specify any arbitrary internal container type, such as an Lru or a HashMap. But this makes the type a little bit more complex/opaque so I haven't implemented it generically. This could be done in the future if the behavior is actually needed to wrap multiple different underlying data containers.

### DIFF
--- a/src/sync/shared_memory.zig
+++ b/src/sync/shared_memory.zig
@@ -69,7 +69,7 @@ pub fn SharedPointerWindow(
         /// call `release` when you're done with the pointer
         pub fn get(self: *Self, index: usize) ?*const T {
             self.lock.lockShared();
-            defer self.lock.lockShared();
+            defer self.lock.unlockShared();
 
             if (self.window.get(index)) |element| {
                 return element.acquire().payload();
@@ -81,7 +81,7 @@ pub fn SharedPointerWindow(
         /// call `release` when you're done with the pointer
         pub fn contains(self: *Self, index: usize) bool {
             self.lock.lockShared();
-            defer self.lock.lockShared();
+            defer self.lock.unlockShared();
 
             return self.window.contains(index);
         }
@@ -91,7 +91,7 @@ pub fn SharedPointerWindow(
 
             const items_to_release = blk: {
                 self.lock.lock();
-                defer self.lock.lock();
+                defer self.lock.unlock();
 
                 self.center.store(new_center, .monotonic);
                 break :blk self.window.realignGet(new_center, self.discard_buf);
--- a/src/sync/shared_memory.zig
+++ b/src/sync/shared_memory.zig
@@ -1,0 +1,148 @@
+const std = @import("std");
+const sig = @import("../sig.zig");
+
+const Allocator = std.mem.Allocator;
+
+/// Thread safe Window that stores a single copy of data that is shared with
+/// readers as a pointer to the underlying data inside the Window.
+///
+/// - this struct owns the data and is responsible for freeing it
+/// - the lifetime of returned pointer exceeds every read operation of that pointer,
+///   even if another thread evicts it from the Window, as long as `release` is used properly.
+pub fn SharedPointerWindow(
+    T: type,
+    deinitItem_: anytype,
+    DeinitContext: type,
+) type {
+    const Window = sig.utils.collections.Window;
+    const Rc = sig.sync.Rc;
+    const deinitItem = normalizeDeinitFunction(T, DeinitContext, deinitItem_);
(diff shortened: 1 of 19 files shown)


src/sync/shared_memory.zig
```diff
@@ -0,0 +1,148 @@
...
+) type {
+    const Window = sig.utils.collections.Window;
+    const Rc = sig.sync.Rc;
+    const deinitItem = normalizeDeinitFunction(T, DeinitContext, deinitItem_);
+
+    return struct {
+        allocator: Allocator,
+        window: Window(Rc(T)),
+        center: std.atomic.Value(usize),
+        lock: std.Thread.RwLock = .{},
```
yewman, line 25: Unless explicitly required I prefer the use of a RwLock wrapper around the necessary data which in this instance is `window` as far as I understand. I understand this is a matter of taste to some extent though so am ok with leaving it as is if you feel more strongly about it.

src/utils/collections.zig
```diff
@@ -486,6 +486,121 @@ pub fn orderSlices(
...
+        }
+
+        pub fn deinit(self: Self, allocator: Allocator) void {
+            allocator.free(self.state);
+        }
+
+        /// Inserts the item into the Window, as long as its index
+        /// is within the current allowed bounds of the Window.
+        pub fn put(self: *Self, index: usize, item: T) error{OutOfBounds}!?T {
+            if (!self.isInRange(index)) {
```
yewman, line 529: NIT: tend to prefer single line if it fits within out col limits but not that fussed

yewman (approved):
lgtm, a couple of minor comments. Feel free to resolve them if you disagree as I do not feel strongly about either comment, just thought I'd share my view

src/utils/collections.zig
dnut: Typically I'll agree with collapsing if-statements when they can fit on a single line, but if it's a return or a break, I give it a second thought. This was a conscious decision to break it into multiple lines.

I feel that the exit points of a function are critical to understand how the function works, and they should be given extra emphasis compared to other code. Putting it at the end of the line makes it less obvious. I find it easier to comprehend a function's overall flow when all the return statements are isolated on their own lines.

I think the benefit of reducing lines of code is not worth the clarity tradeoff, unless there are many instances of the same pattern, and the repetition makes the pattern obvious. For example, if the function starts with five lines of code that all end in `orelse return null`, then I think it's overkill to allocate extra lines of code for that. But if it's just a single return, I try to highlight it.

src/sync/shared_memory.zig
dnut: I agree and I would argue that I am already using the pattern that you are advocating. SharedWindowPointer is the locking wrapper struct which wraps the Window type. The only reason this struct exists is to serve as an alternative to RwMux that provides stronger guarantees about synchronization correctness.

src/utils/collections.zig
yewman: I tend to find syntax highlighting sufficient for this purpose but also don't feel that strongly to happy to leave it as is

src/sync/shared_memory.zig
yewman: The `lock` held by the struct is primarily used to control synchronisation on the `window` field. Changing the `window` field type to `RwMux(Window(Rc(T)))` reflects that fact in the type system. I have noticed a preference for this approach in previous PR reviews as well so thought it worth a mention.  At the end of the day I don't feel strongly enough to hold things up so feel free to resolve this comment.

dnut: You could make the same argument about the inner fields within RwMux. RwMux contains a RwLock to protect a specific field that it contains. Obviously you wouldn't wrap the field inside a RwMux with another struct that's just like RwMux, since that would be redundant.

This data structure is analogous to RwMux. I feel that the same reasoning applies.

PR #465, opened 2024-12-19
perf(networking): add xev socket backend option

kprotty:
Add an implementation of `socket_utils.SocketThread` using libxev. Disabled for now, but will be evaluated for use in the future.

### DIFF
--- a/src/net/socket_utils.zig
+++ b/src/net/socket_utils.zig
@@ -395,7 +395,7 @@ const PerThread = struct {
     }
 };
 
-// TODO: Evaluate when XevThread socket backend is beneficial. 
+// TODO: Evaluate when XevThread socket backend is beneficial.
 const SocketBackend = PerThread;
 
 pub const SocketThread = struct {
--- a/src/net/socket_utils.zig
+++ b/src/net/socket_utils.zig
@@ -2,6 +2,7 @@ const std = @import("std");
 const builtin = @import("builtin");
 const sig = @import("../sig.zig");
 const network = @import("zig-network");
+const xev = @import("xev");
 
 const Allocator = std.mem.Allocator;
 const UdpSocket = network.Socket;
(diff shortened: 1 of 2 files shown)


src/net/socket_utils.zig
```diff
@@ -18,74 +19,344 @@ pub const PACKETS_PER_BATCH: usize = 64;
...
 
-pub const SocketThread = struct {
-    allocator: Allocator,
-    handle: std.Thread,
+const XevThread = struct {
+    pub const Handle = std.Thread.ResetEvent;
 
-    pub fn spawnSender(
+    const List = std.DoublyLinkedList(struct {
         allocator: Allocator,
```
Sobeston, line 26: Does it make sense to have an allocator per-node?

src/net/socket_utils.zig
```diff
@@ -18,74 +19,344 @@ pub const PACKETS_PER_BATCH: usize = 64;
...
+        st: ?*SocketThread,
+        packet: Packet = undefined,
+        udp: xev.UDP,
+        udp_state: xev.UDP.State = undefined,
+        udp_completion: xev.Completion = undefined,
+        io: enum { idle, pending, cancelled } = .idle,
+    });
+
+    var ref_count = std.atomic.Value(usize).init(0);
+    var register = std.atomic.Value(?*List.Node).init(null);
```
Sobeston, line 36: Looks like you could use std.once here? Though might not be an improvement

src/net/socket_utils.zig
```diff
@@ -18,74 +19,344 @@ pub const PACKETS_PER_BATCH: usize = 64;
...
+    var active = List{};
+
+    pub fn spawn(st: *SocketThread) !void {
+        // Make the socket non-blocking (required by xev).
+        var flags = try std.posix.fcntl(st.socket.internal, std.posix.F.GETFL, 0);
+        flags |= @as(c_int, @bitCast(std.posix.O{ .NONBLOCK = true }));
+        _ = try std.posix.fcntl(st.socket.internal, std.posix.F.SETFL, flags);
+
+        // Start xev thread if not running.
+        if (ref_count.fetchAdd(2, .acquire) == 0) {
```
Sobeston, line 49: Is this meant to be 2?

src/net/socket_utils.zig
```diff
@@ -18,74 +19,344 @@ pub const PACKETS_PER_BATCH: usize = 64;
...
+        timer.run(&loop, &timer_completion, SOCKET_TIMEOUT_US, xev.Timer, &timer, onTick);
+
+        try loop.run(.until_done);
     }
 
-    pub fn join(self: *SocketThread) void {
-        self.handle.join();
-        self.allocator.destroy(self);
+    fn onNotify(
+        _: ?*void,
```
Sobeston, line 127: Is this param necessary?

src/net/socket_utils.zig
```diff
@@ -18,74 +19,344 @@ pub const PACKETS_PER_BATCH: usize = 64;
...
+        var pushed = register.swap(null, .acquire);
+        while (pushed) |node| {
+            pushed = node.next;
+            active.append(node);
+        }
+
+        // Iterate active nodes and poll them. If shutting down, close them instead.
+        var iter = active.first;
+        while (iter) |node| {
+            iter = node.next;
```
Sobeston, line 167: Could put the next in the while's continue expression?

src/net/socket_utils.zig
Sobeston: Seems intentional, would be good to comment why

src/net/socket_utils.zig
```diff
@@ -18,74 +19,344 @@ pub const PACKETS_PER_BATCH: usize = 64;
...
+    fn onTick(
+        maybe_timer: ?*xev.Timer,
+        loop: *xev.Loop,
+        completion: *xev.Completion,
+        result: xev.Timer.RunError!void,
+    ) xev.CallbackAction {
+        result catch |e| std.debug.panic("xev notify event failed: {}", .{e});
+
+        pollIo(loop);
+        maybe_timer.?.run(loop, completion, SOCKET_TIMEOUT_US, xev.Timer, maybe_timer, onTick);
```
Sobeston, line 149: We take the timer as an optional, and then almost immediately assert it isn't

Sobeston: Okay, seems the optional is required to satisfy libxev's types? In this case, maybe just not use the "maybe_" param names on functions where we actually require it.

src/net/socket_utils.zig
Sobeston: Ignore me, required for libxev's types.

Sobeston (changes requested):
Some small nits, overall seems good

src/net/socket_utils.zig
kprotty: Needs to track deinit + support being initialized multiple times i.e. during test/benchmark

src/net/socket_utils.zig
kprotty: SocketThreads which make the Nodes could be created with diffferent allocators

src/net/socket_utils.zig
kprotty: `closeNode()` could dealloc the node so the continue expr would ref invalid memory

src/net/socket_utils.zig
kprotty: https://github.com/Syndica/sig/pull/465/commits/693219d352ad864800c5ad9da20211113ba6afab

src/net/socket_utils.zig
kprotty: https://github.com/Syndica/sig/pull/465/commits/693219d352ad864800c5ad9da20211113ba6afab

Sobeston (approved):
All looks good

PR #476, opened 2025-01-03
perf(accountsdb): bufferpool-integration

Sobeston:
Rough TL;DR:
- Renamed CachedRead to ReadHandle, support multiple kinds of reads (i.e. cached reads, plus a couple others)
- AccountInFile stores values rather than pointers into mmapped data, its data is now a Readhandle (therefore, must be deinitialised)
- AccountFile holds a file rather than mmapped memory
- Several things in buffer_pool were changed to be threadsafe, plus a few more utils
- lots of changes around accountsdb to accommodate for these lifetime changes

---

- [X] benchmark results
- [x] lock over fifo eviction
- ~~[ ] fuzzer tests~~
   - we do already have accountsdb fuzzing, which does have coverage for this
   - going to expand this fuzzing in a followup PR
      - edit: https://github.com/Syndica/sig/pull/532

---

benchmarks

### This branch

| Scenario                                           | read_time (min) | read_time (max) | read_time (mean) | read_time (std_dev) | write_time (min) | write_time (max) | write_time (mean) | write_time (std_dev) |
|----------------------------------------------------|-----------------|-----------------|------------------|---------------------|------------------|------------------|-------------------|----------------------|
| 10k accounts (10_slots - ram index - disk accounts) | 3               | 3               | 3                | 0                   | 134              | 138              | 136               | 1                    |
| 500k accounts (1_slot - ram index - disk accounts)  | 526             | 628             | 590              | 45                  | 911              | 1012             | 971               | 43                   |

(computeAccountHashesAndLamports:n_threads=4)
| Scenario                      | load_time   | validate_time | fastload_save_time | fastload_time |
|-------------------------------|-------------|---------------|--------------------|---------------|
| testnet - ram index - 4 threads | 79745       | 1351613       | 122914             | 33262         |

(computeAccountHashesAndLamports:n_threads=32)
| Scenario                      | load_time | validate_time | fastload_save_time | fastload_time |
|-------------------------------|-----------|---------------|--------------------|---------------|
| testnet - ram index - 4 threads | 119468    | 1484989       | 88949              | 32348         |

### Main

| Scenario                                           | read_time (min) | read_time (max) | read_time (mean) | read_time (std_dev) | write_time (min) | write_time (max) | write_time (mean) | write_time (std_dev) |
|----------------------------------------------------|-----------------|-----------------|------------------|---------------------|------------------|------------------|-------------------|----------------------|
| 10k accounts (10_slots - ram index - disk accounts) | 3               | 6               | 3                | 1                   | 14               | 17               | 15                | 0                    |
| 500k accounts (1_slot - ram index - disk accounts)  | 600             | 619             | 608              | 8                   | 308              | 437              | 390               | 58                   |

(computeAccountHashesAndLamports:n_threads=4)
| Scenario                      | load_time | validate_time | fastload_save_time | fastload_time |
|-------------------------------|-----------|---------------|--------------------|---------------|
| testnet - ram index - 4 threads | 110886    | 131546        | 89096              | 33880         |

(computeAccountHashesAndLamports:n_threads=32)
| Scenario                      | load_time      | validate_time  | fastload_save_time | fastload_time  |
|-------------------------------|----------------|----------------|--------------------|----------------|
| testnet - ram index - 4 threads | 132466         | 52177          | 116869             | 68824          |

Performance notes
- "collecting hashes from accounts using 32 threads..." is a significant slowdown, due to the extra contention. "validate_time" really suffers at higher thread counts. 32 (=total hardware threads) is likely excessive here? 
- testing in a somewhat memory-constrained environment would be a more realistic test - the kernel was doing a great job at caching things behind the scenes here.
- could experiment with `O_DIRECT` -  we don't need the kernel cache on top of this. 
- FRAME_SIZE is tweakable.
- we could very easily introduce a lazy reading mode, where file io is moved to the iterator. However, we normally either a) read the account "header", or b) read all the account data. So unsure if this is useful.
- io_uring isn't actually faster than looping with pread - we use fewer syscalls, but it doesn't seem to parallelise our reads at all. This could be because I benchmarked with a SATA SSD?
- would be easy to add a 2nd frame manager with large frames, like how linux has its huge pages. Could improve performance of large reads greatly. A single bit of frame indices could be spared for `enum(u1) { normal, large }`.
- the accountsdb read test on main cheats a bit as it doesn't actually read the .data (if you don't access the bytes, mmap won't necessarily do the work). But this is factored into my main benchmarks using a tiny patch.
- there's still hot-path allocations going on which could be removed fairly easily


src/test_runner.zig
```diff
@@ -0,0 +1,142 @@
...
+        const inverse_filters: []const []const u8 = &.{
+            "shrink account file works",
+            // zig/std/posix.zig:978:22: 0x3c42ad5 in pread (test)
+            //             .BADF => return error.NotOpenForReading, // Can be a race condition.
+
+            // leaky tests
+            "streaming accounts",
+            "stream on load",
+            "buf resizing",
+        };
```
Sobeston, line 52: Need to fix up these tests - mostly geyser memory leaks.

src/test_runner.zig
```diff
@@ -0,0 +1,142 @@
+//! Slightly modified test-runner from std, remove before merge
+const std = @import("std");
```
Sobeston, line 2: To be removed before merge

Sobeston (commented):
Commented a couple pre-merge things.

Not in the diff, but I also need to use atomics for `FramesMetadata`, but unsure how to do a saturating add on a `u2`.

Sobeston:
Notes for reviewers:
- If you can, double-check my usage of atomics, that's where I'm the least confident. It's enough to make tsan happy, but I'm not sure that it's enough.
- There are some hot-path allocations going on now, which I'll be removing (either in this PR, or a follow-up).

src/accountsdb/buffer_pool.zig
```diff
@@ -604,32 +709,32 @@ pub const HierarchicalFIFO = struct {
...
             .ghost => {
-                std.debug.assert(metadata.freq[key] == 0);
-                metadata.freq[key] = 1;
+                std.debug.assert(metadata.freqIsZero(key));
+                metadata.freqSetToOne(key);
                 // Add key to main too - important to note that the key *still*
                 // exists within ghost, but from now on we'll ignore that entry.
                 self.main.writeItemAssumeCapacity(key);
-                metadata.in_queue[key] = .main;
+                metadata.in_queue[key].store(.main, .release);
```
kprotty, line 722: If this is callable by multiple threads, there could be spurious `writeItemAssumeCapacity` calls. Or the `freqIsZero` could work on one thread but fail on another.

Might be better to wrap the whole function inside a mutex since theres multiple changes that need to be atomic w.r.t. each other.

src/accountsdb/buffer_pool.zig
```diff
@@ -196,12 +244,11 @@ pub const BufferPool = struct {
         size: FrameOffset,
     ) error{CannotOverwriteAliveInfo}!void {
```
0xNineteen, line 245: theres a `    /// TODO: atomics` above this method which i dont think we need?

src/accountsdb/cache.zig
```diff
@@ -9,6 +9,7 @@ const LruCacheCustom = sig.common.lru.LruCacheCustom;
 const ReferenceCounter = sig.sync.reference_counter.ReferenceCounter;
 const Slot = sig.core.Slot;
 const Counter = sig.prometheus.counter.Counter;
+const ReadHandle = sig.accounts_db.buffer_pool.ReadHandle;
 
 /// Stores read-only in-memory copies of commonly used *rooted* accounts
 pub const AccountsCache = struct {
```
0xNineteen, line 15: do we even need the account_cache anymore? feels like all its logic is now in the bufferpool

src/accountsdb/db.zig
```diff
@@ -652,11 +665,16 @@ pub const AccountsDB = struct {
             };
             errdefer accounts_file.deinit();
 
+            const file_id = file_info.id;
```
0xNineteen, line 668: this is suppose to be after the `if (n_accounts_this_slot == 0) {` check so we dont add empty files to the map

src/accountsdb/db.zig
```diff
@@ -3164,11 +3246,20 @@ pub fn indexAndValidateAccountFile(
...
+            error.EOF => break,
+            else => |e| return e,
+        };
+        defer account.deinit(allocator);
+
         try account.validate();
 
         if (geyser_storage) |storage| {
-            try storage.cloneAndTrack(account);
+            try storage.cloneAndTrack(allocator, account);
```
0xNineteen, line 3262: just a note we probably gotta come up with a better allocation strategy here -- ideally prealloc a batch of mem for account data because the alloc calls will slow it down a lot

src/accountsdb/buffer_pool.zig
```diff
@@ -774,91 +968,260 @@ pub const CachedRead = struct {
...
+                    .buffer_pool = buffer_pool,
+                    .frame_indices = frame_indices,
+                    .first_frame_start_offset = first_frame_start_offset,
+                    .last_frame_end_offset = last_frame_end_offset,
+                },
+            },
+        };
+    }
+
+    pub fn dupeExternalOwned(self: ReadHandle, allocator: std.mem.Allocator) !ReadHandle {
```
0xNineteen, line 1024: do we really need to handle this as a readHandle -- feels like once they want to get an owned Account, it can just be an Account (unrelated to the ReadHandle) -- we could remove the External enum from all this logic

src/core/account.zig
```diff
@@ -134,7 +140,7 @@ pub fn hashAccount(
     std.mem.writeInt(u64, &int_buf, rent_epoch, .little);
     hasher.update(&int_buf);
 
-    hasher.update(data);
+    while (data.next()) |byte| hasher.update((&byte)[0..1]);
```
0xNineteen, line 143: this will be very slow relative to how it was before -- ref to other iter comment

src/core/account.zig
```diff
@@ -3,21 +3,25 @@ const sig = @import("../sig.zig");
 const Pubkey = sig.core.Pubkey;
 const Epoch = sig.core.Epoch;
 const AccountInFile = sig.accounts_db.accounts_file.AccountInFile;
+const ReadHandle = sig.accounts_db.buffer_pool.ReadHandle;
 
 pub const Account = struct {
     lamports: u64,
-    data: []u8,
+    data: ReadHandle,
```
0xNineteen, line 10: feels like this could be an enum { slice: []u8, buffer_pool: BufferPoolRead }

0xNineteen (changes requested):

src/core/account.zig
Sobeston: Replaced with faster iterator that iterates over frames fc69c06bd38d6f2dd73fcc3cf311e2cb2881d9cb

src/accountsdb/buffer_pool.zig
```diff
@@ -449,35 +499,40 @@ pub const BufferPool = struct {
             try self.eviction_lfu.insert(self.frames_metadata, f_idx.*);
         }
 
-        return CachedRead{
-            .buffer_pool = self,
-            .frame_indices = frame_indices,
-            .first_frame_start_offset = @intCast(file_offset_start % FRAME_SIZE),
-            .last_frame_end_offset = @intCast(((file_offset_end - 1) % FRAME_SIZE) + 1),
-        };
+        for (frame_indices) |f_idx| lifetime_probe(f_idx, .init);
```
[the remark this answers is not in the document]
Sobeston: Opted to remove - f995ed4e7efd2a42c62f44efecef19ff49581028.

src/accountsdb/buffer_pool.zig
```diff
@@ -113,19 +150,30 @@ pub const BufferPool = struct {
...
         self.eviction_lfu.deinit(init_allocator);
         const frame_map, var frame_map_lg = self.frame_map_rw.writeWithLock();
         frame_map.deinit(init_allocator);
         frame_map_lg.unlock();
+
+        lifetime_probe(0, .reset_all);
+    }
+
+    const ProbeKind = enum(u8) { init, deinit, reset_all };
+    noinline fn lifetime_probe(frame: FrameIndex, kind: ProbeKind) callconv(.C) void {
```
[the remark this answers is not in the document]
Sobeston: Opted to remove f995ed4e7efd2a42c62f44efecef19ff49581028

src/accountsdb/buffer_pool.zig
```diff
@@ -41,6 +58,36 @@ const FileIdFileOffset = packed struct(u64) {
     file_offset: FileOffset,
 };
 
+fn readError() type {
```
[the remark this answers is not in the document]
Sobeston: I mean, I have a strong preference against `anyerror` - I don't think we should allow it in the codebase outside where it's super necessary, it just makes proper error handling kinda impossible. We could use an implicit error set, which would be simpler? But I think having an error set that callers can reference is useful

src/accountsdb/buffer_pool.zig
```diff
@@ -735,20 +840,109 @@ pub const HierarchicalFIFO = struct {
...
+                },
+                .sub_read => |_| {},
+                .external => |external| {
+                    if (external.owned) allocator.free(external.slice);
+                },
+            }
+        }
+    };
+
+    pub const Reader = std.io.GenericReader(*Iterator, error{}, Iterator.readBytes);
```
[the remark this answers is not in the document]
Sobeston: So I made the Reader more efficient. I think we should keep a Reader around because you won't always want to read an entire account? Though I suppose there's probably only two real cases
a) reading the whole account
b) reading the account's metadata

src/accountsdb/buffer_pool.zig
```diff
@@ -735,20 +840,109 @@ pub const HierarchicalFIFO = struct {
...
 /// view over one or more buffers owned by the BufferPool
-pub const CachedRead = struct {
-    buffer_pool: *BufferPool,
-    frame_indices: []const FrameIndex,
-    /// inclusive, the offset into the first frame
-    first_frame_start_offset: FrameOffset,
-    /// exclusive, the offset into the last frame
-    last_frame_end_offset: FrameOffset,
+pub const ReadHandle = struct {
+    inner: Inner,
```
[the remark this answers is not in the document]
Sobeston: Removed 990a2da77bcccce15225f4b9434094115bf7bda8.

src/accountsdb/buffer_pool.zig
```diff
@@ -488,21 +543,41 @@ pub const FramesMetadata = struct {
         errdefer allocator.free(rc);
         @memset(rc, .{ .state = .{ .raw = 0 } });
 
-        const key = try allocator.alignedAlloc(FileIdFileOffset, std.mem.page_size, num_frames);
+        const key = try allocator.alignedAlloc(
+            std.atomic.Value(u64),
+            std.mem.page_size,
```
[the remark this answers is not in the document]
Sobeston: Removed in bad14b6eec5138408b1d5cdf60c1df50fc62e469. I was thinking it might be good later to use mmap directly, e.g.
a) using hugepages
b) using `MAP_POPULATE` to reduce the cost of page faults

src/accountsdb/buffer_pool.zig
```diff
@@ -296,7 +347,7 @@ pub const BufferPool = struct {
         file_offset_start: FileOffset,
         /// exclusive
         file_offset_end: FileOffset,
-    ) !CachedRead {
+    ) ReadError!ReadHandle {
         if (!use_io_uring) @compileError("io_uring disabled");
 
         const frame_indices = try self.computeFrameIndices(
```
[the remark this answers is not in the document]
Sobeston: Maybe. 

a) We could have a generous upper bound of frame indices pre-allocated by the buffer pool, and have an allocator internal to the buffer pool manage that buffer. There is no fixed upper-bound, so we'd have to make a very generous guess (total frame indices count can be larger than total frame count). I'm not sure if we have an allocator that would work well for it?

b) We could just put an instance of the gpa or something inside the buffer pool and have that dish out the index slices. But could perform much worse than if the caller used e.g. an FBA for the indices.

src/accountsdb/db.zig
```diff
@@ -3221,8 +3312,11 @@ pub fn writeSnapshotTarWithFields(
         std.debug.assert(account_file.length == file_info.length);
 
         try snapgen.writeAccountFileHeader(archive_writer_counted, file_slot, file_info);
-        try archive_writer_counted.writeAll(account_file.memory);
-        try snapgen.writeAccountFilePadding(archive_writer_counted, account_file.memory.len);
+
+        var fifo = std.fifo.LinearFifo(u8, .{ .Static = std.mem.page_size }).init();
+        try fifo.pump(account_file.file.reader(), archive_writer_counted);
```
InKryption, line 3317: Do we understand our guarantees around the file's cursor position? If not, maybe want to `seekTo(0)` just before this, to ensure the file cursor isn't in a bad state.

src/accountsdb/db.zig
```diff
@@ -3221,8 +3312,11 @@ pub fn writeSnapshotTarWithFields(
...
+
+        try snapgen.writeAccountFilePadding(archive_writer_counted, (try account_file.file.stat()).size); // feels silly, should probably store this
```
InKryption, line 3319:
```suggestion
        try snapgen.writeAccountFilePadding(archive_writer_counted, try account_file.file.getEndPos()); // feels silly, should probably store this
```
Also, yeah; the file's "actual length field" was removed in a PR at one point since it was redundant with the mmap'd buffer's length - may make sense to add back now in order to avoid needing to do this, especially considering account files are immutable.

src/accountsdb/buffer_pool.zig
```diff
@@ -735,130 +803,322 @@ pub const HierarchicalFIFO = struct {
...
 
-    pub fn readByte(self: CachedRead, index: usize) u8 {
-        std.debug.assert(self.frame_indices.len != 0);
-        std.debug.assert(index < self.len());
-        const offset = index + self.first_frame_start_offset;
-        std.debug.assert(offset >= self.first_frame_start_offset);
+    /// Copies entire read into specified buffer. Must be correct length.
+    pub fn readAll(self: ReadHandle, buf: []u8) error{InvalidArgument}!void {
+        try self.read(0, self.len(), buf);
+    }
```
InKryption, line 894: In tandem with my other suggestion concerning `read`, would then need to adjust this to explicitly check `self.len() != buf.len`.

src/accountsdb/accounts_file.zig
```diff
@@ -259,7 +267,8 @@ pub const AccountInFile = struct {
 /// Analogous to [AccountStorageEntry](https://github.com/anza-xyz/agave/blob/4c921ca276bbd5997f809dec1dd3937fb06463cc/accounts-db/src/accounts_db.rs#L1069)
 pub const AccountFile = struct {
     // file contents
-    memory: []align(std.mem.page_size) u8,
+    file: std.fs.File,
```
InKryption, line 270: One concern I have here is that this moves us back from lots of mmap'd data to potentially lots of open file handles, which may wind up with us running into the "fd quota exceeded" errors again; are there any specific mitigations you implemented? If no, that's fine, we can work around them by increasing the fd quota like before, however in future we ought to get around to replacing stored file handles with some userland reference that we can use to derive the file handle on demand (ie the file id + slot pair used to derive the file name).

Non-actionable comment, feel free to mark as resolved if you don't have anything to comment and didn't implement any specific mitigations that I missed.

InKryption (changes requested):
A few observations, suggestions, and concerns. Overall I'm happy with how the integration ended up.

src/accountsdb/accounts_file.zig
Sobeston: > which may wind up with us running into the "fd quota exceeded" errors again

Hm, I guess we'll have to wait and see. Hopefully not.

> we ought to get around to replacing stored file handles with some userland reference that we can use to derive the file handle on demand (ie the file id + slot pair used to derive the file name).

Sounds easily doable, makes sense to me

src/accountsdb/buffer_pool.zig
Sobeston: 26cd92e81d02dd6073b05ab15975611a7d696393

src/accountsdb/db.zig
Sobeston: Sure, why not 31231a58cdb2b48c82ec8b5c55f5178e185d60c2

src/accountsdb/buffer_pool.zig
```diff
@@ -735,130 +803,322 @@ pub const HierarchicalFIFO = struct {
...
+    pub fn dupeAllocatedOwned(self: ReadHandle, allocator: std.mem.Allocator) !ReadHandle {
+        const data_copy = try self.readAllAllocate(allocator);
+        return initAllocatedOwned(data_copy);
     }
 
-    pub fn deinit(self: CachedRead, allocator: std.mem.Allocator) void {
-        for (self.frame_indices) |frame_index| {
-            std.debug.assert(frame_index != INVALID_FRAME);
+    pub fn slice(self: *ReadHandle, start: usize, end: usize) ReadHandle {
+        return ReadHandle{ .sub_read = .{
```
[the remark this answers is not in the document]
Sobeston: https://github.com/Syndica/sig/pull/476/commits/141d58b0749965bbb4bf7cec58689ef4bd0fcc70#diff-9334997573b7124ee92d8ed1efaa5a70b34f8531988349743900d1cf39014569R1025-R1027

src/accountsdb/buffer_pool.zig
```diff
@@ -735,130 +803,322 @@ pub const HierarchicalFIFO = struct {
...
+        return self.readAllocate(allocator, 0, self.len());
     }
 
     /// Copies entire read into specified buffer. Must be correct length.
-    pub fn readAll(
-        self: CachedRead,
+    pub fn read(
+        self: *const ReadHandle,
+        start: FileOffset,
+        end: FileOffset,
```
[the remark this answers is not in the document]
Sobeston: https://github.com/Syndica/sig/commit/26cd92e81d02dd6073b05ab15975611a7d696393

src/accountsdb/buffer_pool.zig
```diff
@@ -735,130 +803,322 @@ pub const HierarchicalFIFO = struct {
...
-        bytes_read: u32 = 0,
+    const AllocatedRead = struct {
+        slice: []u8,
+        /// owned => ReadHandle.deinit() will free the slice
+        owned: bool,
+    };
 
-        const Reader = std.io.GenericReader(*Iterator, error{}, readBytes);
+    const SubRead = packed struct(u128) {
+        parent: *ReadHandle,
```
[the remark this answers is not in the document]
Sobeston: eb024022cef3c36bb86df49cf1244de6e20229ff

src/accountsdb/buffer_pool.zig
```diff
@@ -735,130 +803,322 @@ pub const HierarchicalFIFO = struct {
...
+            self.bytes_read += @intCast(frame_buf.len);
+            return frame_buf;
+        }
+    };
+
+    pub const @"!bincode-config" = bincode.FieldConfig(ReadHandle){
+        .deserializer = bincodeDeserialize,
+        .serializer = bincodeSerialize,
+        .free = bincodeFree,
+    };
```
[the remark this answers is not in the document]
Sobeston: https://github.com/Syndica/sig/pull/476/commits/141d58b0749965bbb4bf7cec58689ef4bd0fcc70#diff-9334997573b7124ee92d8ed1efaa5a70b34f8531988349743900d1cf39014569R891-R1158

src/accountsdb/buffer_pool.zig
```diff
@@ -735,130 +803,322 @@ pub const HierarchicalFIFO = struct {
...
-    frame_indices: []const FrameIndex,
-    /// inclusive, the offset into the first frame
-    first_frame_start_offset: FrameOffset,
-    /// exclusive, the offset into the last frame
-    last_frame_end_offset: FrameOffset,
+pub const ReadHandle = union(enum) {
+    /// Data owned by BufferPool, returned by .read() - do not construct this yourself (!)
+    cached: CachedRead,
+    /// Data allocated elsewhere, not owned or created by BufferPool.
+    allocated_read: AllocatedRead,
```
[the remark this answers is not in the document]
Sobeston: 32cde18ad924f26146557a073bc4e18ccdc38c23

src/accountsdb/db.zig
Sobeston: e424ea6479c52f62572e0ab5eff02f509c63307a

src/accountsdb/buffer_pool.zig
```diff
@@ -735,20 +840,109 @@ pub const HierarchicalFIFO = struct {
...
+    ) anyerror!Inner {
+        const data = try bincode.read(alloc, []u8, reader, params);
+        return ReadHandle.initExternalOwned(data).inner;
+    }
+
+    fn bincodeFree(allocator: std.mem.Allocator, inner: anytype) void {
+        (ReadHandle{ .inner = inner }).deinit(allocator);
+    }
+
+    const Inner = union(enum) {
```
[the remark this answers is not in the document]
Sobeston: Inner removed https://github.com/Syndica/sig/commit/990a2da77bcccce15225f4b9434094115bf7bda8

src/accountsdb/buffer_pool.zig
```diff
@@ -735,20 +840,109 @@ pub const HierarchicalFIFO = struct {
...
 /// slice-like datatype
 /// view over one or more buffers owned by the BufferPool
-pub const CachedRead = struct {
-    buffer_pool: *BufferPool,
-    frame_indices: []const FrameIndex,
-    /// inclusive, the offset into the first frame
-    first_frame_start_offset: FrameOffset,
-    /// exclusive, the offset into the last frame
-    last_frame_end_offset: FrameOffset,
+pub const ReadHandle = struct {
```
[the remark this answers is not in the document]
Sobeston: Not sure. The idea is to have an abstraction over readable data, BufferPoolRead doesn't capture the rest

src/accountsdb/accounts_file.zig
```diff
@@ -270,40 +279,45 @@ pub const AccountFile = struct {
 
     const Self = @This();
 
+    // TODO: document the difference between .length and the file size, this tripped me up
```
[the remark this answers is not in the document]
Sobeston: e424ea6479c52f62572e0ab5eff02f509c63307a

src/accountsdb/buffer_pool.zig
```diff
@@ -735,20 +840,109 @@ pub const HierarchicalFIFO = struct {
...
+        .serializer = bincodeSerialize,
+        .free = bincodeFree,
+    };
+
+    fn bincodeSerialize(writer: anytype, inner: anytype, params: bincode.Params) anyerror!void {
+        var iter = (ReadHandle{ .inner = inner }).iterator();
+
+        try bincode.write(writer, @as(u64, iter.cached_read.len()), params);
+        while (iter.next()) |byte| {
+            try writer.writeByte(byte);
```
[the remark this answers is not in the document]
Sobeston: On further thought, I think the iterator is probably the way to go? It's been replaced with one that's as fast as possible for our needs fc69c06bd38d6f2dd73fcc3cf311e2cb2881d9cb

src/accountsdb/cache.zig
Sobeston: Probably not, it's likely superseded, will know for sure after benchmarks.

src/accountsdb/db.zig
Sobeston: 7ff1dd4a3535a6bec16e9bba5e949335f7df88f1

src/accountsdb/buffer_pool.zig
```diff
@@ -70,9 +122,6 @@ pub const BufferPool = struct {
     /// used for eviction to free less popular (rc=0) frames first
     eviction_lfu: HierarchicalFIFO,
 
-    /// NOTE: we might want this to be a threadlocal for best performance? I don't think this field is threadsafe
-    io_uring: if (use_io_uring) std.os.linux.IoUring else void,
-
     pub fn init(
         init_allocator: std.mem.Allocator,
```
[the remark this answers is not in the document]
Sobeston: 9699d7ea1e12aa978ff8da04c32690a0fe3a6d2c

src/accountsdb/buffer_pool.zig
```diff
@@ -113,19 +150,30 @@ pub const BufferPool = struct {
             .free_list = free_list,
             .frame_map_rw = frame_map_rw,
             .eviction_lfu = try HierarchicalFIFO.init(init_allocator, num_frames / 10, num_frames),
-            .io_uring = io_uring,
         };
     }
 
     pub fn deinit(self: *BufferPool, init_allocator: std.mem.Allocator) void {
```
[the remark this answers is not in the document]
Sobeston: 9699d7ea1e12aa978ff8da04c32690a0fe3a6d2c

src/accountsdb/accounts_file.zig
```diff
@@ -270,40 +279,45 @@ pub const AccountFile = struct {
...
-            std.posix.MAP{ .TYPE = .PRIVATE },
-            file.handle,
-            0,
-        );
-
         return .{
-            .memory = memory,
-            .length = accounts_file_info.length,
+            .file = file,
+            .length = accounts_file_info.length, // is this wrong?
```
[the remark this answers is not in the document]
Sobeston: https://github.com/Syndica/sig/commit/e424ea6479c52f62572e0ab5eff02f509c63307a

src/accountsdb/db.zig
Sobeston: 045fd3a274689c7eb2eca57426831317847a20ea

src/accountsdb/accounts_file.zig
```diff
@@ -365,27 +381,95 @@ pub const AccountFile = struct {
...
+        buffer_pool: *BufferPool,
+        start_offset: usize,
+    ) !AccountInFile {
         var offset = start_offset;
 
-        const store_info = try self.getType(&offset, AccountInFile.StorageInfo);
-        const account_info = try self.getType(&offset, AccountInFile.AccountInfo);
-        const hash = try self.getType(&offset, Hash);
-        const data = try self.getSlice(&offset, store_info.data_len);
+        // TODO efficiency: could reduce this to one buffer_pool call, and slice that
```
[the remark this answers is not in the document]
Sobeston: Going to defer any efficiency-related issues until I've got more satisfactory benchmarking data (-e still failing)

> level=info scope=accounts_db.snapshot_manifest message="reading *full* snapshot fields from: ./data/test-data/bench_snapshot/snapshots/313289869/313289869" time=2025-01-22T01:09:34.594Z
level=info scope=accounts_db.snapshot_manifest message="reading *incremental* snapshot fields from: (error.FileNotFound)/snapshots/313291496/313291496" time=2025-01-22T01:09:38.338Z
error: FileNotFound

src/accountsdb/buffer_pool.zig
```diff
@@ -27,6 +28,22 @@ const linux_io_mode: LinuxIoMode = .IoUring;
 
 const use_io_uring = builtin.os.tag == .linux and linux_io_mode == .IoUring;
```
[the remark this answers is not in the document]
Sobeston: https://github.com/Syndica/sig/issues/509 6056fb78374d79de62f8e186ed3532f4b6325587

src/accountsdb/buffer_pool.zig
```diff
@@ -533,10 +608,40 @@ pub const FramesMetadata = struct {
...
+    fn freqIncrement(self: FramesMetadata, index: FrameIndex) void {
+        const old_freq = @atomicRmw(u2, &self.freq[index], .Add, 1, .acquire);
+        if (old_freq == 0) {
+            // we overflowed (3->0), set back to max
+            @atomicStore(u2, &self.freq[index], 3, .release);
+        }
+    }
+
+    fn freqDecrement(self: FramesMetadata, index: FrameIndex) void {
+        const old_freq = @atomicRmw(u2, &self.freq[index], .Add, 1, .acquire);
```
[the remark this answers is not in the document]
Sobeston: Using .sub, now behind a mutex e91f7ca39c43bf182de25213b9136a8a9d854ef3

src/accountsdb/buffer_pool.zig
```diff
@@ -920,6 +1283,47 @@ test AtomicStack {
...
+            return .{
+                .buf = buf,
+                .head = std.atomic.Value(usize).init(0),
+                .tail = std.atomic.Value(usize).init(0),
+            };
+        }
+
+        pub fn writeItemAssumeCapacity(self: *Self, item: T) void {
+            const tail = self.tail.fetchAdd(1, .monotonic) % self.buf.len;
+            self.buf[tail] = item;
```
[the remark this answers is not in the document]
Sobeston: now behind a mutex https://github.com/Syndica/sig/commit/e91f7ca39c43bf182de25213b9136a8a9d854ef3

src/accountsdb/buffer_pool.zig
```diff
@@ -920,6 +1283,47 @@ test AtomicStack {
...
+        }
+
+        pub fn readItem(self: *Self) ?T {
```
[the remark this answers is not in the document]
Sobeston: now behind a mutex https://github.com/Syndica/sig/commit/e91f7ca39c43bf182de25213b9136a8a9d854ef3

src/accountsdb/buffer_pool.zig
Sobeston: now behind a mutex https://github.com/Syndica/sig/commit/e91f7ca39c43bf182de25213b9136a8a9d854ef3

src/accountsdb/buffer_pool.zig
Sobeston: This is gone / not sure what commit I removed it in

src/accountsdb/buffer_pool.zig
```diff
@@ -27,6 +28,22 @@ const linux_io_mode: LinuxIoMode = .IoUring;
...
+            io_uring_entries,
+            0,
+        );
+        // TODO: to deinit reliably we could hook thread exit? Not sure if we even need to.
+        break :io_ur &(maybe_io_uring.?);
+    };
+}
+
 const FileIdFileOffset = packed struct(u64) {
     const INVALID: FileIdFileOffset = .{
```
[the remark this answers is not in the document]
Sobeston: Reordered a lot of stuff - need to add this to contributing.md

src/accountsdb/buffer_pool.zig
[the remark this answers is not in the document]
0xNineteen: do we really need that abstraction? imo there should be a simpler way

src/accountsdb/cache.zig
0xNineteen: nice, created a ticket here: https://github.com/orgs/Syndica/projects/2/views/10?pane=issue&itemId=94926926

src/accountsdb/db.zig
0xNineteen: not sure what this commit solves?

src/core/account.zig
0xNineteen: this is what almost feels wrong imo - ReadHandle has a lot of methods which associate with the buffer pool whereas in this case we _only_ want to represent the data as either a slice or a buffer_pool_read and let each of those cases handle the logic themselves - separating the structs more make things easier to follow what struct contains what logic

src/accountsdb/buffer_pool.zig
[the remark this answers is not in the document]
0xNineteen: okay if theres a strong preference for it then im okay with it - my thought is just how much added work/code there is for an explicit error set which is not readable (or is a doom-wall list) and we dont use/switch on it anywhere

src/accountsdb/db.zig
Sobeston: In this codepath, the data is now not cloned, the allocator is just for copying the frames slice. Though this doesn't reduce the number of allocations, which is unfortunate. Maybe a better way of sharing ReadHandles is in order

src/accountsdb/accounts_file.zig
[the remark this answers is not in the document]
Sobeston: This ended up fixed in 6208d729261204a9187e2e4ea3d61625a3d4f3ee

src/accountsdb/buffer_pool.zig
```diff
@@ -920,6 +1283,47 @@ test AtomicStack {
...
+
+        pub fn writableLengthIsZero(self: *Self) bool {
+            const current_head = self.head.load(.acquire);
+            const current_tail = self.tail.load(.acquire);
+            return (current_tail - current_head) >= self.buf.len;
+        }
+    };
+}
+
 test "BufferPool indicesRequired" {
```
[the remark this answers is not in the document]
Sobeston: Removed

src/core/account.zig
Sobeston: Hm. The point of the methods on Read Handle is to be generic, i.e. usable in cases where the data could be stored in different ways, e.g.
- deinit
- len
- read/readAll/readAllAllocate
- iterator, reader

These methods all work as expected on slices/cached reads/etc, so it's a shared API. I guess there are a few ways we could cut this?

src/accountsdb/buffer_pool.zig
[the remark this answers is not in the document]
Sobeston: https://github.com/Syndica/sig/pull/529

Sobeston:
Got some tsan warnings, however these are all surrounding prometheus metrics / its arena allocator? This is unrelated to the PR.

src/accountsdb/buffer_pool.zig
```diff
@@ -735,190 +757,356 @@ pub const HierarchicalFIFO = struct {
...
 /// slice-like datatype
 /// view over one or more buffers owned by the BufferPool
-pub const CachedRead = struct {
-    buffer_pool: *BufferPool,
-    frame_indices: []const FrameIndex,
-    /// inclusive, the offset into the first frame
-    first_frame_start_offset: FrameOffset,
-    /// exclusive, the offset into the last frame
-    last_frame_end_offset: FrameOffset,
+pub const ReadHandle = union(enum) {
```
0xNineteen, line 760: AccountDataHandle ?

src/accountsdb/buffer_pool.zig
```diff
@@ -735,190 +757,356 @@ pub const HierarchicalFIFO = struct {
...
+    cached: CachedRead,
+
+    /// Data allocated elsewhere, not owned or created by BufferPool. BufferPool will deallocate.
+    owned_allocation: []const u8,
+    /// Data allocated elsewhere, not owned or created by BufferPool.
+    unowned_allocation: []const u8,
+    /// Data owned by parent ReadHandle
+    sub_read: SubRead,
+    /// Used in place of a read, in callsites where it is not actually needed. Provides .len().
+    unread: Unread,
```
0xNineteen, line 771: empty?

src/accountsdb/buffer_pool.zig
```diff
@@ -19,26 +21,75 @@ comptime {
     _ = offset;
 }
 
+const FrameRef = packed struct(u32) {
+    f_idx: FrameIndex,
```
0xNineteen, line 25: `index`? feels like the frame part is already infered from the struct name?

src/accountsdb/buffer_pool.zig
```diff
@@ -19,26 +21,75 @@ comptime {
...
+    hit: bool,
+    const INIT = FrameRef{
```
0xNineteen, line 27: spacing nice

```suggestion
    hit: bool,
    
    pub const INIT = FrameRef{
```

src/accountsdb/buffer_pool.zig
```diff
@@ -19,26 +21,75 @@ comptime {
...
     Blocking,
     IoUring,
 };
-const linux_io_mode: LinuxIoMode = .IoUring;
+const linux_io_mode: LinuxIoMode = .Blocking;
 
+// TODO: ideally we should be able to select this with a cli flag. (#509)
 const use_io_uring = builtin.os.tag == .linux and linux_io_mode == .IoUring;
 
+const io_uring_entries = 128;
```
0xNineteen, line 44: screaming case form constants?

src/accountsdb/buffer_pool.zig
```diff
@@ -19,26 +21,75 @@ comptime {
...
+
 const LinuxIoMode = enum {
     Blocking,
     IoUring,
 };
-const linux_io_mode: LinuxIoMode = .IoUring;
+const linux_io_mode: LinuxIoMode = .Blocking;
 
+// TODO: ideally we should be able to select this with a cli flag. (#509)
 const use_io_uring = builtin.os.tag == .linux and linux_io_mode == .IoUring;
```
0xNineteen, line 42: same here

src/accountsdb/buffer_pool.zig
```diff
@@ -296,247 +241,299 @@ pub const BufferPool = struct {
...
+
+        var i: u32 = 0;
+        var n_read: u32 = 0;
+        while (i < frame_refs.len or n_read < frame_refs.len) {
+            var queue_full = false;
+
+            if (i < frame_refs.len) {
+                const frame_ref = &frame_refs[i];
+
+                const populated = self.manager.contains_valid_data[frame_ref.f_idx].load(.acquire);
```
0xNineteen, line 265: probs need to change/double check the var names match the field name changes

src/accountsdb/buffer_pool.zig
```diff
@@ -296,247 +241,299 @@ pub const BufferPool = struct {
...
-            f_idx.* = blk: while (true) {
-                if (self.free_list.popOrNull()) |free_idx| {
-                    break :blk free_idx;
-                } else {
-                    try self.evictUnusedFrame();
-                }
+    /// Evicts unused frames for reuse.
+    eviction_lfu: sig.sync.RwMux(HierarchicalFIFO),
+
+    /// Per-frame refcounts. Used to track what frames still have handles associated with them.
```
0xNineteen, line 337: frame_ref_counts?

src/accountsdb/buffer_pool.zig
```diff
@@ -19,26 +21,75 @@ comptime {
...
+const FrameRef = packed struct(u32) {
+    f_idx: FrameIndex,
+    hit: bool,
+    const INIT = FrameRef{
+        .f_idx = INVALID_FRAME,
+        .hit = false,
+    };
+};
+
+const Rc = std.atomic.Value(u32);
```
0xNineteen, line 33: do we need this alias? -- throughout the codebase we usually rename `const Atomic = Value` and then use `Atomic(u32)` throughout -- would make `frame_ref_counts: []Atomic(u32)` much easier to read -- wdyt

src/accountsdb/buffer_pool.zig
```diff
@@ -53,208 +104,85 @@ const FileIdFileOffset = packed struct(u64) {
...
-    free_list: AtomicStack(FrameIndex),
-
-    /// uniquely identifies a frame
-    /// for finding your wanted index
-    /// TODO: a concurrent hashmap would be more appropriate
-    frame_map_rw: sig.sync.RwMux(FrameMap),
-
     frames: []Frame,
-    frames_metadata: FramesMetadata,
+    manager: Manager,
```
0xNineteen, line 108: `frame_manager: FrameManager`

src/accountsdb/buffer_pool.zig
```diff
@@ -53,208 +104,85 @@ const FileIdFileOffset = packed struct(u64) {
...
 
-    /// used for eviction to free less popular (rc=0) frames first
-    eviction_lfu: HierarchicalFIFO,
-
-    /// NOTE: we might want this to be a threadlocal for best performance? I don't think this field is threadsafe
-    io_uring: if (use_io_uring) std.os.linux.IoUring else void,
+    pub const ReadBlockingError = FrameManager.GetError || std.posix.PReadError;
+    pub const ReadIoUringError = FrameManager.GetError || IoUringError;
+    pub const ReadError = if (use_io_uring) ReadIoUringError else ReadBlockingError;
+    pub const Manager = FrameManager;
```
0xNineteen, line 113: should probs reduce rexports if we dont neeeeed them

src/accountsdb/db.zig
```diff
@@ -108,6 +110,8 @@ pub const AccountsDB = struct {
     /// NOTE: see accountsdb/readme.md for more details on how these are used
     file_map_fd_rw: std.Thread.RwLock,
```
0xNineteen, line 112: would be good to create a ticket to remove `maybe_accounts_cache_rw` -- so we have it tracked

src/accountsdb/db.zig
```diff
@@ -1235,17 +1253,26 @@ pub const AccountsDB = struct {
                     // hashes aren't always stored correctly in snapshots
                     if (account_hash.eql(Hash.ZEROES)) {
                         const account, var lock_guard = try self.getAccountFromRefWithReadLock(max_slot_ref);
```
0xNineteen, line 1255: ticket for fixing/optimizing `getAccountHashAndLamportsFromRef` (not in this diff but near this location) would be good too since its the main bottleneck (including removing the bufferpool from this usage since the data is thrown away immediately after and we just need to read the file directly)

src/accountsdb/buffer_pool.zig
```diff
@@ -19,26 +21,75 @@ comptime {
     _ = offset;
 }
 
+const FrameRef = packed struct(u32) {
+    f_idx: FrameIndex,
+    hit: bool,
```
0xNineteen, line 26: `is_in_cache`?

0xNineteen (changes requested):
few naming nits - otherwise lgtm

src/accountsdb/buffer_pool.zig
Sobeston: Renamed to AccountDataHandle f40689214b7f8a0945d24c5a2847c29af6db0203

src/accountsdb/buffer_pool.zig
Sobeston: renamed to empty f40689214b7f8a0945d24c5a2847c29af6db0203

src/accountsdb/buffer_pool.zig
Sobeston: renamed to index f40689214b7f8a0945d24c5a2847c29af6db0203

src/accountsdb/buffer_pool.zig
Sobeston: added spacing f40689214b7f8a0945d24c5a2847c29af6db0203

src/accountsdb/buffer_pool.zig
Sobeston: made them screaming case f40689214b7f8a0945d24c5a2847c29af6db0203

src/accountsdb/buffer_pool.zig
Sobeston: made them screaming case f40689214b7f8a0945d24c5a2847c29af6db0203

src/accountsdb/buffer_pool.zig
Sobeston: Removed / gave it a double check https://github.com/Syndica/sig/commit/f40689214b7f8a0945d24c5a2847c29af6db0203

src/accountsdb/buffer_pool.zig
Sobeston: Renamed to frame_ref_counts https://github.com/Syndica/sig/commit/f40689214b7f8a0945d24c5a2847c29af6db0203

src/accountsdb/buffer_pool.zig
Sobeston: Removed alias, replaced with Atomic(u32) https://github.com/Syndica/sig/commit/f40689214b7f8a0945d24c5a2847c29af6db0203

src/accountsdb/buffer_pool.zig
Sobeston: renamed https://github.com/Syndica/sig/commit/f40689214b7f8a0945d24c5a2847c29af6db0203

src/accountsdb/buffer_pool.zig
Sobeston: removed Manager re-export https://github.com/Syndica/sig/commit/f40689214b7f8a0945d24c5a2847c29af6db0203. If we want we could only export ReadError?

src/accountsdb/buffer_pool.zig
Sobeston: Not exactly *is* in cache, called it `found_in_cache` https://github.com/Syndica/sig/commit/f40689214b7f8a0945d24c5a2847c29af6db0203

src/accountsdb/buffer_pool.zig
0xNineteen: still some places like `const ReadHandle = sig.accounts_db.buffer_pool.AccountDataHandle;`

src/accountsdb/buffer_pool.zig
0xNineteen: missed this last review, but can we keep all constants at the top of the file

src/accountsdb/buffer_pool.zig
```diff
@@ -735,190 +757,356 @@ pub const HierarchicalFIFO = struct {
...
-pub const CachedRead = struct {
-    buffer_pool: *BufferPool,
-    frame_indices: []const FrameIndex,
-    /// inclusive, the offset into the first frame
-    first_frame_start_offset: FrameOffset,
-    /// exclusive, the offset into the last frame
-    last_frame_end_offset: FrameOffset,
+pub const AccountDataHandle = union(enum) {
+    /// Data owned by BufferPool, returned by .read() - do not construct this yourself (!)
+    cached: CachedRead,
```
0xNineteen, line 762: buffer_pool_read ?

src/accountsdb/buffer_pool.zig
0xNineteen: im okay either way

src/accountsdb/buffer_pool.zig
Sobeston: 55f52cf76c01ca23a88562a668737b36f28ca235

src/accountsdb/buffer_pool.zig
Sobeston: Hm I think I got all of them, `grep -r "ReadHandle" src/` shows nothing.
```
src/accountsdb/accounts_file.zig:
  11: const AccountDataHandle = sig.accounts_db.buffer_pool.AccountDataHandle;
src/accountsdb/cache.zig:
  12: const AccountDataHandle = sig.accounts_db.buffer_pool.AccountDataHandle;
src/accountsdb/db.zig:
  37: const AccountDataHandle = sig.accounts_db.buffer_pool.AccountDataHandle;
src/accountsdb/fuzz.zig:
  14: const AccountDataHandle = sig.accounts_db.buffer_pool.AccountDataHandle;
src/core/account.zig:
  6: const AccountDataHandle = sig.accounts_db.buffer_pool.AccountDataHandle;
  ```

File: src/accountsdb/buffer_pool.zig (hunk 55 above)
Comment (Sobeston) [reply src/accountsdb/buffer_pool.zig]: renamed in 9d0bf9d3b8a4304181e7f7ed9252f4c4aad847eb

File: src/accountsdb/db.zig (hunk 52 above)
Comment (Sobeston) [reply src/accountsdb/db.zig:112]: #537

File: src/accountsdb/db.zig (hunk 53 above)
Comment (Sobeston) [reply src/accountsdb/db.zig:1213]: #538

**Review by 0xNineteen — DISMISSED** [review]
👌  :shipit:

File: src/accountsdb/buffer_pool.zig [hunk 56]
```diff
@@ -19,26 +28,68 @@ comptime {
...
+};
+
 const LinuxIoMode = enum {
     Blocking,
     IoUring,
 };
-const linux_io_mode: LinuxIoMode = .IoUring;
 
-const use_io_uring = builtin.os.tag == .linux and linux_io_mode == .IoUring;
+fn io_uring() !*IoUring {
```
Rexicon226, line 46: should be camel case

src/accountsdb/buffer_pool.zig
```diff
@@ -284,9 +211,26 @@ pub const BufferPool = struct {
                 file_offset_start,
                 file_offset_end,
             );
+
+        return handle;
+    }
+
+    pub fn computeNumberofFrameIndices(
```
Rexicon226, line 218:
```suggestion
    pub fn computeNumberOfFrameIndices(
```

src/accountsdb/buffer_pool.zig
```diff
@@ -638,24 +663,24 @@ pub const HierarchicalFIFO = struct {
     /// This does not return an optional, as the caller *requires* a key to be
     /// evicted. Not being able to return a key means illegal internal state in
     /// the BufferPool.
-    pub fn evict(self: *HierarchicalFIFO, metadata: Metadata) Key {
+    pub fn evict(self: *HierarchicalFIFO, frame_ref_counts: []Atomic(u32)) Key {
```
Rexicon226, line 666:
```suggestion
    pub fn evict(self: *HierarchicalFIFO, frame_ref_counts: []const Atomic(u32)) Key {
```

Rexicon226 (changes requested):
Just gave it a quick once over, I trust the other reviews to have caught the more major stuff

src/accountsdb/buffer_pool.zig
```diff
@@ -53,208 +104,84 @@ const FileIdFileOffset = packed struct(u64) {
...
+            file_id,
+            file_offset_start,
+            file_offset_end,
+        );
+        errdefer allocator.free(frame_refs);
+        errdefer self.frame_manager.getAllIndicesRollback(frame_refs);
 
-        if (n_indices > self.frames.len) return error.OffsetsOutOfBounds;
+        // read into frames without valid data
+        for (0.., frame_refs) |i, *frame_ref| {
```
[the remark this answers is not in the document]
Sobeston: Do you mean `for (frame_refs, 0..) |*frame_ref, i|`?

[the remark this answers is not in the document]
Sobeston: Ah, yeah, didn't realise we used `0..` last basically always.

src/accountsdb/buffer_pool.zig
Sobeston: fixed in 549a6f3986e7ee9d607a5517f45ee1ad6b40ad08

Sobeston:
Managed to pull out some stats. We've used a little more RssAnon, but cut down massively on RssFile, which is to be expected.

image
image

0xNineteen (dismissed):
nice

ultd (approved):

Rexicon226 (approved):

PR #477, opened 2025-01-03
perf(rpc): io_uring integration & redesign

InKryption:
- [x] get receive working
- [x] get send working
- error handling:
  - [x] eliminate explicit TODO panics
  - [x] improve `first_error` error handling if practical

follow up work (not part of this pr): https://github.com/Syndica/sig/issues/517


src/rpc/server.zig
```diff
@@ -47,485 +37,265 @@ pub const Server = struct {
...
+pub const WorkPool = union(enum) {
+    basic,
+    linux_io_uring: switch (LinuxIoUring.can_use) {
+        .yes, .check => *LinuxIoUring,
+        .no => noreturn,
+    },
 
-        const server = hct.server;
-        const logger = server.logger;
+    pub const LinuxIoUring = @import("server/linux_io_uring.zig").LinuxIoUring;
```
0xNineteen, line 103: can we stick to normal import rules (all at top)

0xNineteen (changes requested):
few nits on imports otherwise looks great - would be good to get another look over the io_uring logic from either sobe or king before we merge

src/rpc/server.zig
```diff
@@ -1,39 +1,29 @@
+const builtin = @import("builtin");
 const std = @import("std");
 const sig = @import("../sig.zig");
 
+const connection = @import("server/connection.zig");
```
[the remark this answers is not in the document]
InKryption: Well, I don't think that format particularly makes sense for this group of files here, because they're not being exposed as a public API, they're internal to the RPC server, and I don't think it makes sense to expose them.
server.zig & server/ feel appropriate to me with that in mind.

src/rpc/server.zig
InKryption: My contention there is that `LinuxIoUring` should be directly associated with `WorkPool`, as it should only ever be instantiated in relation to it; `WorkPool.LinuxIoUring` makes this clear, and makes it so one only needs to import `WorkPool` in order to gain access to it, rather than having to import both separately from the server namespace.

src/rpc/server.zig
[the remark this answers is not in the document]
0xNineteen: my preference is we keep all files in a public api and accessible as we have with the rest of the repo - the public file and private folder scheme makes sense but for consistency would pref not do that (it gets confusing to find things with two possible locations)

src/rpc/server.zig
0xNineteen: makes sense i was meaning that we have `pub const LinuxIoUring = LinuxIoUring` and removing the @import

src/rpc/server.zig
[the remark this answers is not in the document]
dnut: Even if you want files to be private, I think it's more organized to explicitly define the private namespace in a single place, rather than scattering filename imports throughout the codebase.

0xNineteen (dismissed):
nice

Sobeston (commented):
Looking good, just a couple Qs

src/utils/io.zig
```diff
@@ -87,3 +87,87 @@ fn NarrowAnyStream(comptime Error: type) type {
...
 }
+
+/// Writer which captures only an offset window of data into a buffer.
+/// This can be useful for incrementally capturing formatted data.
+pub const WindowedWriter = struct {
+    remaining_to_ignore: u64,
+    end_index: usize,
+    buffer: []u8,
+
+    pub fn init(
```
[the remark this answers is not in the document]
InKryption: Could you elaborate?

src/rpc/server/linux_io_uring.zig
```diff
@@ -537,9 +523,10 @@ fn handleRecvBody(
                 } };
                 const sfh = &entry_data.state.send_file_head;
                 switch (try sfh.computeAndMaybePrepSend(entry, &liou.io_uring)) {
-                    .sending_more => return,
                     .all_sent => unreachable, // we know this for certain
+                    .sending_more => {},
                 }
+                return;
```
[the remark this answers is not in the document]
InKryption: The back-and-forth in calling `computeAndMaybePrepSend` is described in the name: first compute, and then maybe prep send; the condition for the send being prepped is that we still have more data to send. We know for a fact that `sfh.sent_bytes == 0`, since that's what we just set it to, meaning this function will have to see there's still data remaining.
If we've prepped the send, that means we're "sending more"; if it detects that, based on `sent_bytes`, we've already sent everything, it's "all sent", meaning there's no need to prep another send.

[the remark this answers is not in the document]
Sobeston: Sounds good, could use a slightly better comment? But fine as-is.

src/rpc/server/linux_io_uring.zig
```diff
@@ -0,0 +1,847 @@
...
+                null,
+                std.os.linux.SOCK.CLOEXEC,
+            ) catch |err| return switch (err) {
+                error.SubmissionQueueFull => |e| {
+                    server_ctx.logger.err().log(
+                        "Under normal circumstances the accept_multishot would be" ++
+                            " the first SQE to be queued, but somehow the queue was full.",
+                    );
+                    return e;
+                },
```
kprotty, line 99: is this worth being a panic?

src/rpc/server/server.zig
```diff
@@ -0,0 +1,291 @@
...
+            .socket_addr = sock_addr,
+            .read_buffer_size = 4096,
+            .reuse_address = true,
+        });
+        defer rpc_server_ctx.joinDeinit();
+
+        var exit = std.atomic.Value(bool).init(false);
+        const serve_thread = try rpc_server_ctx.serveSpawn(work_pool, &exit);
+        defer serve_thread.join();
+        defer exit.store(true, .release);
```
kprotty, line 205: does the server (basic or io_uring) have a way to check for the exit flag once set (to then stop and complete the `join()`)?

src/rpc/server/connection.zig
```diff
@@ -0,0 +1,265 @@
...
+
+pub const HandleAcceptResult = enum {
+    success,
+    intr,
+    again,
+    conn_aborted,
+};
+
+/// Resembles the error handling of `std.posix.accept`.
+pub fn handleAcceptResult(
```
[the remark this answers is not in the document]
InKryption: Both; the main purpose is to share error handling code between backends that use `accept` - and since we're defining it anyway, it makes sense to use panics as we've prescribed in the rest of the codebase.

src/rpc/server/linux_io_uring.zig
```diff
@@ -0,0 +1,847 @@
...
+    multishot_accept_submitted: bool,
+    pending_cqes_count: u8,
+    pending_cqes_buf: [255]std.os.linux.io_uring_cqe,
+
+    pub const can_use: enum { no, yes, check } = switch (builtin.os.getVersionRange()) {
+        .linux => |version| can_use: {
+            const min_version: std.SemanticVersion = .{ .major = 6, .minor = 0, .patch = 0 };
+            const is_at_least = version.isAtLeast(min_version) orelse break :can_use .check;
+            break :can_use if (is_at_least) .yes else .no;
+        },
```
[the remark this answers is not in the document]
InKryption: This is an interesting point; do you think it should just always be `.check` for the linux target then?

src/rpc/server/linux_io_uring.zig
InKryption: Hmm, I'm a bit indifferent, is there a specific reason you'd want it to be a panic vs a propagated error? Mind you, this isn't necessarily a one-time thing, `self.multishot_accept_submitted` can be reset if the accept_multishot is terminated (see line 184).

src/rpc/server/server.zig
InKryption: I'm not 100% sure on how to answer this question. The server itself doesn't need to worry about the exit flag, just the main serve loop; the waitgroup.wait() in deinit ensures all jobs are complete before deinitializing the tcp server and all the rest.

src/rpc/server/connection.zig
```diff
@@ -0,0 +1,265 @@
...
+    /// NOTE: this is *only* a hint, and may not apply on all platforms.
+    /// See `have_accept4`.
+    sync_hint: enum { blocking, nonblocking },
+) AcceptHandledError!std.net.Server.Connection {
+    const accept_flags: u32 = blk: {
+        var accept_flags: u32 = std.posix.SOCK.CLOEXEC;
+        accept_flags |= switch (sync_hint) {
+            .blocking => 0,
+            .nonblocking => std.posix.SOCK.NONBLOCK,
+        };
```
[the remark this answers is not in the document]
InKryption: 87e691747e2605ba109d2c13e59e18c371a1674d

src/rpc/server/linux_io_uring.zig
kprotty: The submissions queue being full would occur if theres many IO operations started. Could the RPC server reach that high, given enough accepted connections doing IO? If so, to continue serving it should probably submit() what's currently in the submission queue and try enqueueing the operation again (could maybe make this a wrapper fn)

src/rpc/server/server.zig
kprotty: Presumably, the exit flag is a way to tell the server to stop running. If there were no IO and the exit flag was set, would the server be able to eventually wake and see it to exit? Or would that be way too unlikely?

src/rpc/server/linux_io_uring.zig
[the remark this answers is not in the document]
kprotty: yea, that makes sense.

src/rpc/server/linux_io_uring.zig
InKryption: Under normal circumstances, it shouldn't occur, however it's not impossible that there is some subtle control flow oversight which could lead to that situation; thinking about it more, we probably would want it to panic in such a circumstance.

src/rpc/server/server.zig
InKryption: Yeah, when the exit flag is set, both backends have been designed to eventually wake up and observe that; for the basic backend, this is accomplished by making the accepting socket nonblocking, so that instead of blocking, it just continues and checks the exit flag; for io_uring, it just sort of behaves asynchronously by design. When it finishes reacting to CQEs and finishes dispatching whatever other work it needs to, it will come back up to the loop, and check the exit flag.

src/rpc/server/linux_io_uring.zig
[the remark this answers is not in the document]
InKryption: 36379e38fa2a5474c82f5d06a6b9d1c48594fe2f

src/rpc/server/linux_io_uring.zig
```diff
@@ -0,0 +1,835 @@
...
+            data_ptr.* = .{
+                .buffer = buffer,
+                .stream = stream,
+                .state = EntryState.INIT,
+            };
+            break :entry .{ .ptr = data_ptr };
+        };
+        errdefer if (new_recv_entry.ptr) |data_ptr| server_ctx.allocator.destroy(data_ptr);
+
+        _ = liou.io_uring.recv(
```
dnut, line 205: Just thinking out loud here... Imagine if io_uring had an sqe called "accept_multishot_recv" that didn't bother with a cqe for the accept, and the kernel would just assume automatically that we want to receive the data, and userspace wouldn't be bothered until after a recv cqe was ready. That way we could avoid these steps here. Do you think something like that would be useful?

src/rpc/server/linux_io_uring.zig
```diff
@@ -0,0 +1,835 @@
...
+    comptime unreachable;
+}
+
+const HandleRecvBodyError =
+    IouGetSqeError ||
+    std.fs.Dir.StatFileError ||
+    std.fs.File.OpenError ||
+    std.fs.File.GetSeekPosError ||
+    std.posix.PipeError;
+fn handleRecvBody(
```
dnut, line 430: I think it's easier to read with a blank line. I see you've done this in other places too. Curious what others think.
```suggestion
    std.posix.PipeError;

fn handleRecvBody(
```

0xNineteen: i pref the blank lines too

src/rpc/server/server.zig
```diff
@@ -0,0 +1,288 @@
...
+
+const SnapshotGenerationInfo = sig.accounts_db.AccountsDB.SnapshotGenerationInfo;
+
+pub const connection = @import("connection.zig");
+pub const requests = @import("requests.zig");
+
+pub const basic = @import("basic.zig");
+pub const LinuxIoUring = @import("linux_io_uring.zig").LinuxIoUring;
+
+pub const Context = struct {
```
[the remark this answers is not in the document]
0xNineteen: i think this relates to the namespace its in, ie, `rpc.server.Context` which gives it more meaning -- imo i would prefer the `RpcServer` approach since the name contains everything (even if it is redundant to the namespace) but can see the pros/cons of both - curious what others think

src/rpc/server/linux_io_uring.zig
InKryption: Discussed on slack, resolution: 1242b40725d54ee352f8f47cb860851b790bfe82

src/rpc/server/server.zig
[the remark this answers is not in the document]
InKryption: Well, it doesn't actually contain everything, the backends are defined separately. Calling it the `RpcServer` doesn't entirely make sense, because it isn't the server unto itself, it is merely the "context" required for the WorkPool to operate, and telling it how to dispatch its work. The "server" is `Context` and `WorkPool` working together.

src/rpc/server/server.zig
```diff
@@ -0,0 +1,288 @@
...
+    read_buffer_size: u32,
+
+    pub const LOGGER_SCOPE = "rpc.server";
+    pub const ScopedLogger = sig.trace.log.ScopedLogger(LOGGER_SCOPE);
+
+    pub const MIN_READ_BUFFER_SIZE = 4096;
+
+    pub const WorkPool = union(enum) {
+        basic,
+        linux_io_uring: if (LinuxIoUring.can_use) *LinuxIoUring else noreturn,
```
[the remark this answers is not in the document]
InKryption: It's not an advantage, it's required, the `IoUring` code fails to compile on targets which are not linux.

src/rpc/server/linux_io_uring.zig
```diff
@@ -0,0 +1,835 @@
...
+        PermissionDenied,
+        SystemOutdated,
+    };
+
+    // NOTE(ink): constructing the return type as `E!?T`, where `E` and `T` are resolved
+    // separately seems to help ZLS with understanding the types involved better, which is
+    // why I've done it like that here. If ZLS gets smarter in the future, you could probably
+    // inline this into a single branch in the return type expression.
+    const InitErrOrEmpty = if (!can_use) error{} else InitError;
+    const InitResultOrNoreturn = if (!can_use) noreturn else LinuxIoUring;
```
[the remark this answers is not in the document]
InKryption: Same response as the above, the `IoUring` code will not compile on other targets, so it needs to be comptime-unreachable if we aren't on linux; the reason this is done specifically for the WorkPool definition and this `init` signature is because these are the initial entrypoints to the whole backend; everything else is only reachable if those are reachable.

src/rpc/server/linux_io_uring.zig
```diff
@@ -0,0 +1,835 @@
...
+                self.pending_cqes_buf[i + 1 ..][0..self.pending_cqes_count],
+            );
+            const our_cqe = OurCqe.fromCqe(raw_cqe);
+            switch (try consumeOurCqe(self, server_ctx, our_cqe)) {
+                .ok => {},
+                .signal_interrupted_operation,
+                .misc_connection_failure,
+                .http_head_parse_error,
+                .splice_error,
+                => {},
```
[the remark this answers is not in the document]
InKryption: I've removed these as of 1242b40725d54ee352f8f47cb860851b790bfe82, however we could still add metrics for their sources in the code itself, yeah, would be interesting data to observe.

src/rpc/server/linux_io_uring.zig
```diff
@@ -0,0 +1,835 @@
...
+
+        server_ctx.wait_group.start();
+        errdefer server_ctx.wait_group.finish();
+
+        const buffer = try server_ctx.allocator.alloc(u8, server_ctx.read_buffer_size);
+        errdefer server_ctx.allocator.free(buffer);
+
+        const new_recv_entry: Entry = entry: {
+            const data_ptr = try server_ctx.allocator.create(EntryData);
+            errdefer comptime unreachable;
```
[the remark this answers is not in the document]
InKryption: It's just to avoid needing to handle resource cleanup of `data_ptr` beyond this point; we do this in other parts of the codebase for various reasons. It's just a code assertion to communicate "we just did some expensive operation, and undoing that operation in case of failure beyond this point is complicated or expensive, or would need to be duplicated elsewhere, so just don't do that".

src/rpc/server/linux_io_uring.zig
```diff
@@ -0,0 +1,835 @@
...
+
+            data_ptr.* = .{
+                .buffer = buffer,
+                .stream = stream,
+                .state = EntryState.INIT,
+            };
+            break :entry .{ .ptr = data_ptr };
+        };
+        errdefer if (new_recv_entry.ptr) |data_ptr| server_ctx.allocator.destroy(data_ptr);
+
```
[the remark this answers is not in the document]
InKryption: Not a bad point; was originally like this for other reasons, when the code was a fair bit different than it is now.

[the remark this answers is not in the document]
InKryption: 1f6c7bc6ac2c8149d0a722581bb503cd241dc848

src/rpc/server/linux_io_uring.zig
InKryption: That seems like it would be useful, yeah, but I don't really think there's a way to represent that; can't link the SQEs because the `recv` SQE needs to know immediately what socket it's receiving from, and there's no mechanism to represent using a handle that's only later created or anything like that.

src/rpc/server/linux_io_uring.zig
InKryption: Small oversight, changed in 1242b40725d54ee352f8f47cb860851b790bfe82

src/rpc/server/linux_io_uring.zig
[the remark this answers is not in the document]
dnut: I can see that it's a replacement for `errdefer allocator.destroy(data_ptr)` but I don't see how it's actually better. This code is more esoteric, no more concise, and you have the cleanup code just outside the block anyway. It doesn't seem a big deal either way, and you've removed this in a later commit, so I'll just resolve this comment.

src/rpc/server/linux_io_uring.zig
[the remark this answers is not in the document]
dnut: https://github.com/orgs/Syndica/projects/2/views/10?pane=issue&itemId=96509434

src/rpc/server/server.zig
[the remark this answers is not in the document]
InKryption: Because that was the cleanest way to implement the code; making WorkPool a direct part of the Context was fairly awkward.

The WorkPool functions still require passing in the Context, which is both redundant, and adds a possible invalid invariant where `self != &context.work_pool.linux_io_uring` - which, while this should never happen in practice, is mental overhead for the reader of the IoUring backend and any other future backends.

It adds many responsibilities to `Context.init`, due to the fact that initializing the work pool means not only forwarding init parameters, but implementing a strategy for deciding which workpool backend to select; right now this is simple enough, since there's only `.basic` and `.linux_io_uring`, however the point of the way this is implemented is for adding future alternative backends.

Even if you figure out the perfect strategy for how to make this a workable API, it just adds a lot of complexity and mental overhead to the init function, expanding the error set to need to include all the possible errors from any possible backend - this detracts from the focus of `Context.init`, which is to initialize the basic state required for the server functioning.

As well, earlier in the design process, having them be part of the same thing required an awkward dance to support pinned memory; currently this is no longer an issue, and may not be an issue if they were re-united because of specific design changes made after the fact, but I could absolutely forsee the possibility adding a backend which (or even changing the `.basic` backend such that it) needs a stable pointer to `Context`, e.g. for referencing the `wait_group` across threads; this separation of Context and WorkPool facilitates this and other designs.

TL;DR: the separation of WorkPool and Context is motivated by the cumulative effect of many small, but together significant factors, and I am strongly opposed to unifying them, unless a stronger rationale and solution is discovered.

### Rename

With that said, I have taken your feedback into account, and I would like to propose an alternative API structure:
```zig
var rpc_server: sig.rpc.Server = try .init(...);
defer rpc_server.joinDeinit();

var maybe_liou: ?sig.rpc.Server.LinuxIoUring = try .init(...);
defer if (maybe_liou) |*liou| liou.deinit();

var exit: std.atomic.Value(bool) = .init(false);
errdefer exit.store(true, .release);
(try rpc_server.serveSpawn(&exit, .{ .linux_io_uring = &maybe_liou.? })).detach();
```
Ie, rename `Context` just to `Server`, and expose all relevant declarations in the `Server` namespace, rather than `sig.rpc.server.RpcServer` & co.
I don't think it's perfect, since I stand by the fact that the Context isn't the Server unto itself, but if you think this would be a more comprehensible API, so be it.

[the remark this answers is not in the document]
dnut: Everything you're saying makes sense.

I don't have a clear vision on how this should be structured. I'm asking questions because something about it feels unusual or unfamiliar to me. I'm probing to understand the intent while brainstorming other ideas to understand the problem space and convince myself that this is a good design.

Answering the question, "how do I run an rpc server?" took a little more investigation than I expected. You have to dig around a bit to see what constitutes a server.

Regarding "Server" vs "Context", I understand now why you chose context. Like I said it feels like a grey area, so I'm ambivalent. I like that the word "server" directs your attention to the struct you need to create, which orchestrates the server at a high level. But I also see how it's unusual that the WorkPool would accept something called "Server" as an input, when the WorkPool itself implements most of the server logic.

I think a part of my confusion comes from the `serve` and `acceptAndServeConnection` functions being defined as methods on Context, which make it seem like Context is the core server struct. What do you think about moving those functions outside of Context?

Either as plain functions:
```zig
/// Calls `acceptAndServeConnection` in a loop until `exit.load(.acquire)`.
pub fn serve(
    context: *Context,
    work_pool: *WorkPool,
    /// The pool to dispatch work to.
    exit: *std.atomic.Value(bool),
) AcceptAndServeConnectionError!void {
    context.wait_group.start();
    defer context.wait_group.finish();
    while (!exit.load(.acquire)) {
        try acceptAndServeConnection(context, work_pool);
    }
}

pub fn acceptAndServeConnection(
    context: *Context,
    work_pool: *WorkPool,
) AcceptAndServeConnectionError!void {
    switch (work_pool) {
        .basic => try sig.rpc.server.basic.acceptAndServeConnection(&context),
        .linux_io_uring => |linux| try linux.acceptAndServeConnections(&context),
    }
}
```

Or as methods on a Server struct that combines the two components of a server:
```zig
pub const Server = struct {
    context: Context,
    work_pool: WorkPool,

    /// Calls `acceptAndServeConnection` in a loop until `exit.load(.acquire)`.
    pub fn serve(
        self: *Server,
        /// The pool to dispatch work to.
        exit: *std.atomic.Value(bool),
    ) AcceptAndServeConnectionError!void {
        self.context.wait_group.start();
        defer self.context.wait_group.finish();
        while (!exit.load(.acquire)) {
            try self.acceptAndServeConnection();
        }
    }

    pub fn acceptAndServeConnection(self: *Server) AcceptAndServeConnectionError!void {
        switch (self.work_pool) {
            .basic => try sig.rpc.server.basic.acceptAndServeConnection(&self.context),
            .linux_io_uring => |linux| try linux.acceptAndServeConnections(&self.context),
        }
    }
};
```

Regarding "Server" vs "RpcServer", I think clarity is improved when type names speak for themselves. I also get that this is redundant, and you can imply context from the namespace. It's a fine line deciding where to reduce redundancy vs increase clarity. In this case I have a slight preference for RpcServer, but both ways seem reasonable, so I won't press the issue.

src/rpc/server/lib.zig
```diff
@@ -0,0 +1,11 @@
...
+
+pub const Context = server.Context;
+pub const basic = server.basic;
+pub const LinuxIoUring = server.LinuxIoUring;
+
+comptime {
+    _ = Context;
+    _ = basic;
+    _ = LinuxIoUring;
+}
```
0xNineteen, line 11: whats going on here

src/rpc/server/server.zig
```diff
@@ -0,0 +1,293 @@
...
+const std = @import("std");
+const sig = @import("../../sig.zig");
+
+const SnapshotGenerationInfo = sig.accounts_db.AccountsDB.SnapshotGenerationInfo;
+
+pub const connection = @import("connection.zig");
+pub const requests = @import("requests.zig");
+
+pub const basic = @import("basic.zig");
+pub const LinuxIoUring = @import("linux_io_uring.zig").LinuxIoUring;
```
0xNineteen, line 11: can we keep the imports in order following contributing.md

src/rpc/server/connection.zig
```diff
@@ -0,0 +1,264 @@
...
+    }
+}
+
+pub const SetSocketSync = std.posix.FcntlError;
+
+/// Ensure the socket is set to be blocking or nonblocking.
+pub fn setSocketSync(
+    socket: std.posix.socket_t,
+    sync: enum { blocking, nonblocking },
+) !void {
```
0xNineteen, line 73: is this import meant to be like this? could we also keep all imports at the top of the file

src/rpc/server/connection.zig
```diff
@@ -0,0 +1,264 @@
...
+    var addr: std.net.Address = .{ .any = undefined };
+    var addr_len: std.posix.socklen_t = @sizeOf(@TypeOf(addr.any));
+    try std.posix.getsockname(socket_handle, &addr.any, &addr_len);
+    return addr;
+}
+
+/// When this is false, it means `accept[Handled]` can't apply
+/// flags to the accepted socket, and the caller will have to
+/// to ensure relevant flags are enabled/disabled after acceptance.
+pub const have_accept4 = !builtin.target.isDarwin();
```
0xNineteen, line 17: screaming case ? and at top of file

src/rpc/server/connection.zig
```diff
@@ -0,0 +1,264 @@
+const builtin = @import("builtin");
+const std = @import("std");
+
+pub const GetSockNameError = std.posix.GetSockNameError;
+pub fn getSockName(
```
0xNineteen, line 5: can we get some space up in here

src/rpc/server/linux_io_uring.zig
```diff
@@ -0,0 +1,837 @@
...
+    std.mem.Allocator.Error ||
+    connection.HandleAcceptError ||
+    connection.HandleRecvError ||
+    connection.HandleSendError ||
+    connection.HandleSpliceError;
+
+/// Panic message for handling `EAGAIN`; we're not using nonblocking sockets at all,
+/// so it should be impossible to receive that error, or for such an error to be
+/// triggered just from malicious connections.
+const EAGAIN_PANIC_MSG =
```
0xNineteen, line 99: global constants at top of file ?

dnut: Why do you think they belong on top? What if it's only used by two functions in the file? Couldn't you put it next to those two functions?

0xNineteen: nbd but mainly for consistency - when i see a screaming case constant i auto just go to the top of the file to find it - searching between methods or wherever is just more work imo

InKryption: I could probably just inline this into the relevant portion of `consumeOurCqe` - originally was being used by a couple other functions, but ended up only being used in that.

dnut: The advantage of putting it right next to where it's used is that you don't need to go anywhere. It's visible at the same time as its usage.

If it's not immediately visible, I use go-to-definition in the IDE (ctrl-click or F12 in vscode). I assume you use that for structs and functions?

src/rpc/server/connection.zig
InKryption: It isn't an import so much as it is a re-export; `SetSocketSync` is to be interpreted as the error set of of `setSocketSync` - though should be `SetSocketSyncError`.

src/rpc/server/lib.zig
InKryption: These are references to force analysis of the namespaces; otherwise tests don't run.

src/rpc/server/server.zig
InKryption: Hmm, we want to group exports and internal imports?

src/rpc/server/connection.zig
InKryption: 706f136b75d03abed6a66023acce4c00a76e85c2

src/rpc/server/connection.zig
InKryption: 706f136b75d03abed6a66023acce4c00a76e85c2

src/rpc/server/linux_io_uring.zig
InKryption: 706f136b75d03abed6a66023acce4c00a76e85c2

src/rpc/server/server.zig
0xNineteen: yeah sure - i think its more so ordering of namespaces first then structs

src/rpc/server/linux_io_uring.zig
0xNineteen: if its out of the editor's scope (ie, you will need to scroll for it) then you will still need to go somewhere - theres cases when your reading code not in the editor without go-to-def shortcuts

src/rpc/server/lib.zig
0xNineteen: can we add a note about that - i dont think thats the case for any other struct/files in the project so not sure why this is the case here 🤔

InKryption: Actually I just realized; the reason the tests aren't being picked up is because we're relying on the `refAllDeclsRecursive` function, which can only access public declarations, which `server` is not, which is where the main test is actually defined - it also works just by referencing `_ = server;` directly.

InKryption: Amended to this: 1e042dd9711bf93019c24cf95a90ba5046090762

src/rpc/server/server.zig
InKryption: I think that's kind of weird - in my mind, exports are a separate group unto themselves, intermixing them with internal imports feels wrong. But I'm not going to die on this hill.

InKryption: 76292804a4d84c6ed947e9fd3187329be0c24ea3

src/rpc/server/lib.zig
dnut: This is an annoying thing about zig. I can't figure out a good way to ensure every test is run. I'd like to just say once "please run every single test in the src folder" and zig figures out how to do that. But it doesn't work like that. Instead you need a bunch of weird hacks to make that happen.

InKryption: For what it's worth, they are working on a solution for this. I'll see if I can find the relevant issue or discussion.

0xNineteen (approved):
nice :shipit:

src/rpc/server/server.zig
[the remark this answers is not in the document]
InKryption: Resolution: 3e6c4fb0977eabd23703bf4e5ebd377db6e8e01b

dnut (approved):

0xNineteen (dismissed):

0xNineteen (approved):
amazing work :shipit:

PR #479, opened 2025-01-04
fix(metrics) misc metrics fixes & cleanup

Sobeston:
Confirmed working on Linux (docker engine) and macOS (via docker desktop). Didn't quite work with podman-compose on macOS, but unsure if we'll ever use that.

Summary of changes
- removed linux and mac specific compose files
- removed network mode host
- gave only prometheus host access through `host.docker.internal:host-gateway`
- got prometheus to stop complaining
- got loki to work properly
- updated readme a bit / misc cleanup


0xNineteen:
awesome 🔥

PR #481, opened 2025-01-06
fix(shred-network): rpc epoch ctx service bugs

dnut:
- several issues with memory management
- thread safety
- nonsensical/redundant epoch calculations
- add tests


0xNineteen:
whoops didnt see auto approve set

dnut:
> whoops didnt see auto approve set

I tried to keep these changes uncontroversial. They just basic bug fixes. There may be some improved design choices inspired by these bugs, but I decided not to do that here.

PR #482, opened 2025-01-07
feat(perf) Add continuous profiling support

Sobeston:
Pretty easy change, should be handy.

Has a good amount of detail, even in ReleaseSafe (thanks @Rexicon226 for suggesting to disable omit_frame_pointer).

If you're using docker desktop + macos, this won't work - the profiler won't be able to find your host's process.

image

(focused in on the gossip verification)

image

(focused further, into bincode)

image

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -85,6 +85,12 @@ pub fn build(b: *Build) void {
         .optimize = optimize,
         .sanitize_thread = enable_tsan,
     });
+
+    // make sure pyroscope's got enough info to profile
+    sig_exe.build_id = .fast;
+    sig_exe.root_module.omit_frame_pointer = false;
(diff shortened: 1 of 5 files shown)


Rexicon226:
Wow, nice! How can I test this out?

Sobeston:
> Wow, nice! How can I test this out?

Start Sig, e.g.
`./zig-out/bin/sig gossip -n testnet 2>&1 | tee -a logs/sig.log`
then should be as simple as
`[metrics]$ SIG_PID=$(pgrep sig) docker compose up`

(I'm not sure how common pgrep is, but there's 1000 ways to do that)

0xNineteen (approved):
wow awesome :shipit:

PR #485, opened 2025-01-08
fix(ci/cd): provision benchmark collection

0xNineteen:
- provisions cron jobs to periodically run benchmarks 
- adds support for expensive benchmarks which should only be run for ci/cd (not locally) -- ie, snapshot downloading + unpacking 
- removes extra gossip logs which dont add value (info can be found on grafana)


0xNineteen:
* **#485** <a href="https://app.graphite.dev/github/pr/Syndica/sig/485?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Syndica/sig/485?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/accountsdb/download.zig
```diff
@@ -451,31 +463,22 @@ pub fn getOrDownloadAndUnpackSnapshot(
...
-            std.posix.close(dir.fd);
-            break :blk true;
-        } else |_| {
-            break :blk false;
+    if (std.fs.cwd().openDir(snapshot_path, .{ .iterate = true })) |dir| {
+        defer std.posix.close(dir.fd);
+        // clear old snapshots, if we will download a new one
+        if (force_new_snapshot_download) {
+            logger.info().log("deleting accounts_db dir...");
+            try dir.deleteTreeMinStackSize("accounts_db");
```
InKryption, line 471: Doesn't this change make it so this function is always forced to download a new snapshot, instead of being able to detect and return an existing one?

src/accountsdb/download.zig
```diff
@@ -18,7 +18,9 @@ const FullAndIncrementalManifest = sig.accounts_db.FullAndIncrementalManifest;
 
 const parallelUnpackZstdTarBall = sig.accounts_db.parallelUnpackZstdTarBall;
 
-const DOWNLOAD_PROGRESS_UPDATES_NS = 6 * std.time.ns_per_s;
+// NOTE: this also represents the amount of time a connection is given
+// to warm up before we start checking the download speed
+const DOWNLOAD_PROGRESS_UPDATES_NS = 20 * std.time.ns_per_s;
```
InKryption, line 23: Maybe we should rename this to something like "DOWNLOAD_WARMUP_TIME_NS", and then change the comment to "This also represents the interval at which progress updates are issued.".

InKryption (changes requested):
One rename suggestion that I realized because of a good comment, and a concern about the changes to `getOrDownloadAndUnpackSnapshot`.

src/accountsdb/download.zig
0xNineteen: feels like the change would be equivalent (having the var name describe something and the comment describe its other usage) so wouldnt matter much -- unless you think the warmup time is more important to define vs progress updates?

src/accountsdb/download.zig
0xNineteen: only if`force_new_snapshot_download` is set, we delete the directory - otherwise we re-use it

InKryption: Yeah, that fixes it.

src/accountsdb/download.zig
InKryption: Yeah, I think the warmup time is a more functionally important aspect of this constant; changing it doesn't only impact the aesthetic of the output, it impacts how much warmup time we're permitting for a connection.

0xNineteen: agreed - 3bcc70e8

InKryption (approved):
LGTM

PR #486, opened 2025-01-09
perf(channel): hooks to avoid spinning wait

kprotty:
Adds a Channel.SendHook concept to observe & override the behavior when a value is sent on a channel. It's used to notify a ResetEvent so that receivers can block on it without spinning. Exit flag is periodically checked by using `ResetEvent.timedWait(1s)` to block.

### DIFF
--- a/src/gossip/service.zig
+++ b/src/gossip/service.zig
@@ -149,7 +149,8 @@ pub const GossipService = struct {
     /// Indicates if the gossip service is closed.
     closed: bool,
 
-    /// Piping between the gossip_socket
+    /// Piping between the gossip_socket. 
+    /// Set to null until start() is called as they represent threads.
     incoming_pipe: ?*SocketPipe = null,
     outgoing_pipe: ?*SocketPipe = null,
 
@@ -403,9 +404,8 @@ pub const GossipService = struct {
             },
         };
 
-        self.incoming_pipe = try SocketPipe.init(
+        self.incoming_pipe = try SocketPipe.initReceiver(
             self.allocator,
-            .receiver,
             self.logger.unscoped(),
             self.gossip_socket,
             self.packet_incoming_channel,
@@ -435,10 +435,8 @@ pub const GossipService = struct {
             exit_condition.ordered.exit_index += 1;
         }
 
-        // SocketPipe overrides `packet_outgoing_channel.send_hook`
-        self.outgoing_pipe = try SocketPipe.init(
+        self.outgoing_pipe = try SocketPipe.initSender(
             self.allocator,
-            .sender,
             self.logger.unscoped(),
             self.gossip_socket,
             self.packet_outgoing_channel,
--- a/src/gossip/service.zig
+++ b/src/gossip/service.zig
@@ -53,7 +53,7 @@ const PingAndSocketAddr = sig.gossip.ping_pong.PingAndSocketAddr;
 const ServiceManager = sig.utils.service_manager.ServiceManager;
 const Duration = sig.time.Duration;
 const ExitCondition = sig.sync.ExitCondition;
-const SocketPipe = sig.net.SocketPipe;
+const SocketThread = sig.net.SocketThread;
 
 const endpointToString = sig.net.endpointToString;
 const globalRegistry = sig.prometheus.globalRegistry;
@@ -105,7 +105,7 @@ const GOSSIP_PRNG_SEED = 19;
 
 /// The flow of data goes as follows:
 ///
-/// `SocketPipe` ->
+/// `SocketThread.initReceiver` ->
 ///         - reads from the gossip socket
 ///         - puts the new packet onto `packet_incoming_channel`
 ///         - repeat until exit
@@ -121,14 +121,14 @@ const GOSSIP_PRNG_SEED = 19;
 ///         - processes the verified message it has received
 ///         - depending on the type of message received, it may put something onto `packet_outgoing_channel`
 ///
-///  `SocketPipe` ->
+///  `SocketThread.initSender` ->
 ///         - receives from `packet_outgoing_channel`
 ///         - sends the outgoing packet onto the gossip socket
 ///         - repeats while `exit` is false and `packet_outgoing_channel`
-///         - when `SocketPipe` sees that `exit` has become `true`, it will begin waiting on
+///         - when `SocketThread` sees that `exit` has become `true`, it will begin waiting on
 ///           the previous thing in the chain to close, that usually being `processMessages`.
 ///           this ensures that `processMessages` doesn't add new items to `packet_outgoing_channel`
-///           after the `SocketPipe` thread exits.
+///           after the `SocketThread` exits.
 ///
 pub const GossipService = struct {
     /// used for general allocation purposes
@@ -149,10 +149,10 @@ pub const GossipService = struct {
     /// Indicates if the gossip service is closed.
     closed: bool,
 
-    /// Piping between the gossip_socket.
+    /// Piping data between the gossip_socket and the channels.
     /// Set to null until start() is called as they represent threads.
-    incoming_pipe: ?*SocketPipe = null,
-    outgoing_pipe: ?*SocketPipe = null,
+    incoming_socket_thread: ?*SocketThread = null,
+    outgoing_socket_thread: ?*SocketThread = null,
 
     /// communication between threads
     packet_incoming_channel: *Channel(Packet),
@@ -335,8 +335,8 @@ pub const GossipService = struct {
         self.service_manager.deinit();
 
         // Wait for pipes to shutdown if any
-        if (self.incoming_pipe) |pipe| pipe.deinit(self.allocator);
-        if (self.outgoing_pipe) |pipe| pipe.deinit(self.allocator);
+        if (self.incoming_socket_thread) |thread| thread.join();
+        if (self.outgoing_socket_thread) |thread| thread.join();
 
         // assert the channels are empty in order to make sure no data was lost.
         // everything should be cleaned up when the thread-pool joins.
@@ -404,7 +404,7 @@ pub const GossipService = struct {
             },
         };
 
-        self.incoming_pipe = try SocketPipe.initReceiver(
+        self.incoming_socket_thread = try SocketThread.spawnReceiver(
             self.allocator,
             self.logger.unscoped(),
             self.gossip_socket,
@@ -435,7 +435,7 @@ pub const GossipService = struct {
             exit_condition.ordered.exit_index += 1;
         }
 
-        self.outgoing_pipe = try SocketPipe.initSender(
+        self.outgoing_socket_thread = try SocketThread.spawnSender(
             self.allocator,
             self.logger.unscoped(),
             self.gossip_socket,
--- a/src/gossip/service.zig
+++ b/src/gossip/service.zig
@@ -149,7 +149,7 @@ pub const GossipService = struct {
     /// Indicates if the gossip service is closed.
     closed: bool,
 
-    /// Piping between the gossip_socket. 
+    /// Piping between the gossip_socket.
     /// Set to null until start() is called as they represent threads.
     incoming_pipe: ?*SocketPipe = null,
     outgoing_pipe: ?*SocketPipe = null,
@@ -524,7 +524,7 @@ pub const GossipService = struct {
 
         // loop until the previous service closes and triggers us to close
         while (true) {
-            self.packet_incoming_channel.wait(exit_condition) catch break;
+            self.packet_incoming_channel.waitToReceive(exit_condition) catch break;
 
             // verify in parallel using the threadpool
             // PERF: investigate CPU pinning
@@ -621,7 +621,7 @@ pub const GossipService = struct {
         // - `exit` isn't set,
         // - there isn't any data to process in the input channel, in order to block the join until we've finished
         while (true) {
-            self.verified_incoming_channel.wait(exit_condition) catch break;
+            self.verified_incoming_channel.waitToReceive(exit_condition) catch break;
 
             var msg_count: usize = 0;
             while (self.verified_incoming_channel.tryReceive()) |message| {
--- a/src/gossip/service.zig
+++ b/src/gossip/service.zig
@@ -159,7 +159,7 @@ pub const GossipService = struct {
     packet_outgoing_channel: *Channel(Packet),
     verified_incoming_signal: Channel(GossipMessageWithEndpoint).SendSignal = .{},
     verified_incoming_channel: *Channel(GossipMessageWithEndpoint),
-    
+
     /// table to store gossip values
     gossip_table_rw: RwMux(GossipTable),
     /// manages push message peers
--- a/src/gossip/service.zig
+++ b/src/gossip/service.zig
@@ -53,6 +53,7 @@ const PingAndSocketAddr = sig.gossip.ping_pong.PingAndSocketAddr;
 const ServiceManager = sig.utils.service_manager.ServiceManager;
 const Duration = sig.time.Duration;
 const ExitCondition = sig.sync.ExitCondition;
+const SocketPipe = sig.net.SocketPipe;
 
 const endpointToString = sig.net.endpointToString;
 const globalRegistry = sig.prometheus.globalRegistry;
@@ -104,7 +105,7 @@ const GOSSIP_PRNG_SEED = 19;
 
 /// The flow of data goes as follows:
 ///
-/// `readSocket` ->
+/// `SocketPipe` ->
 ///         - reads from the gossip socket
 ///         - puts the new packet onto `packet_incoming_channel`
 ///         - repeat until exit
@@ -120,14 +121,14 @@ const GOSSIP_PRNG_SEED = 19;
 ///         - processes the verified message it has received
 ///         - depending on the type of message received, it may put something onto `packet_outgoing_channel`
 ///
-///  `sendSocket` ->
+///  `SocketPipe` ->
 ///         - receives from `packet_outgoing_channel`
 ///         - sends the outgoing packet onto the gossip socket
 ///         - repeats while `exit` is false and `packet_outgoing_channel`
-///         - when `sendSocket` sees that `exit` has become `true`, it will begin waiting on
+///         - when `SocketPipe` sees that `exit` has become `true`, it will begin waiting on
 ///           the previous thing in the chain to close, that usually being `processMessages`.
 ///           this ensures that `processMessages` doesn't add new items to `packet_outgoing_channel`
-///           after the `sendSocket` thread exits.
+///           after the `SocketPipe` thread exits.
 ///
 pub const GossipService = struct {
     /// used for general allocation purposes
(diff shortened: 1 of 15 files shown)


src/net/socket_utils.zig
```diff
@@ -10,125 +9,109 @@ const Channel = sig.sync.Channel;
 const Logger = sig.trace.Logger;
 const ExitCondition = sig.sync.ExitCondition;
 
-const UdpSocket = @import("zig-network").Socket;
+const network = @import("zig-network");
+const UdpSocket = network.Socket;
```
0xNineteen, line 13: can we fix the imports in this file -- see contributing.md

src/net/socket_utils.zig
```diff
@@ -10,125 +9,109 @@ const Channel = sig.sync.Channel;
...
-/// A thread that is dedicated to either sending or receiving data over a socket.
-/// The included channel can be used communicate with that thread.
-///
-/// The channel only supports one: either sending or receiving, depending how it
-/// was initialized. While you *could* send data to the channel for a "receiver"
-/// socket, the underlying thread won't actually read the data from the channel.
-pub const SocketThread = struct {
-    channel: *Channel(Packet),
-    exit: *Atomic(bool),
+pub const SocketPipe = struct {
```
0xNineteen, line 21: would SocketManager be a better name or SocketThread -- pipe doesnt really describe what its doing

src/net/socket_utils.zig
```diff
@@ -10,125 +9,109 @@ const Channel = sig.sync.Channel;
...
 
     const Self = @This();
 
-    pub fn initSender(
+    pub const Direction = enum {
+        sender,
+        receiver,
+    };
+
+    pub fn init(
```
0xNineteen, line 31: do you think it would be better to have two seperate methods initSender and initReciever ? -- direction seems almost hidden with how key it is to how this method operates

src/net/socket_utils.zig
```diff
@@ -10,125 +9,109 @@ const Channel = sig.sync.Channel;
...
-        const channel = try Channel(Packet).create(allocator);
-        return .{
-            .channel = channel,
-            .exit = exit,
-            .handle = try std.Thread.spawn(
-                .{},
-                sendSocket,
-                .{ socket, channel, logger, .{ .unordered = exit } },
-            ),
+        channel: *Channel(Packet),
```
0xNineteen, line 36:
```suggestion
        incoming_channel: *Channel(Packet),
```

src/gossip/service.zig
```diff
@@ -147,6 +148,10 @@ pub const GossipService = struct {
     /// Indicates if the gossip service is closed.
     closed: bool,
 
+    /// Piping between the gossip_socket
```
0xNineteen, line 151:
```suggestion
    /// Piping between the gossip_socket (are set to null until the service 
    /// starts because they represent threads)
```

src/shred_network/repair_service.zig
```diff
@@ -251,7 +252,8 @@ pub const RepairRequester = struct {
     logger: ScopedLogger(@typeName(Self)),
     random: Random,
     keypair: *const KeyPair,
-    sender: SocketThread,
+    sender_pipe: *SocketPipe,
+    sender_channel: *Channel(Packet),
```
0xNineteen, line 256: it almost feels like the channel should belong (or be initialized) inside the pipe -- wdyt

src/sync/channel.zig
```diff
@@ -5,14 +5,29 @@ const Backoff = @import("backoff.zig").Backoff;
...
         closed: Atomic(bool) = Atomic(bool).init(false),
         allocator: Allocator,
-        mutex: std.Thread.Mutex = .{},
-        condition: std.Thread.Condition = .{},
+        event: std.Thread.ResetEvent = .{},
+        send_hook: ?*SendHook = null,
+
+        pub const SendHook = struct {
+            /// Called before a value is sent to a channel.
+            /// If it returns false, the value ia not pushed to the channel.
```
0xNineteen, line 21:
```suggestion
            /// If it returns false, the value is not pushed to the channel.
```

src/sync/channel.zig
```diff
@@ -128,19 +143,27 @@ pub fn Channel(T: type) type {
...
             if (channel.closed.load(.monotonic)) {
                 return error.ChannelClosed;
             }
+
+            const send_hook = channel.send_hook;
+            if (send_hook) |hook| {
+                if (!hook.before_send(hook, channel, value)) {
+                    return;
+                }
+            }
```
0xNineteen, line 165: i think adding an intermediate makes this much clearer

```suggestion
            if (send_hook) |hook| {
                const should_send_channel = hook.before_send(hook, channel, value)
                if (!should_send_channel) {
                    return;
                }
            }
```

src/sync/channel.zig
```diff
@@ -128,19 +143,27 @@ pub fn Channel(T: type) type {
...
         }
 
         pub fn send(channel: *Self, value: T) !void {
             if (channel.closed.load(.monotonic)) {
                 return error.ChannelClosed;
             }
+
+            const send_hook = channel.send_hook;
+            if (send_hook) |hook| {
+                if (!hook.before_send(hook, channel, value)) {
```
0xNineteen, line 162: i might be confused but shouldnt it just be `hook.before_send(channel, value)`

src/trace/log.zig
```diff
@@ -141,8 +142,9 @@ pub const ChannelPrintLogger = struct {
             .exit = AtomicBool.init(false),
             .max_level = config.max_level,
             .handle = null,
-            .channel = try Channel([]const u8).create(config.allocator),
+            .channel = try Channel([]const u8).init(config.allocator),
```
0xNineteen, line 145: why is this not `create`?

0xNineteen (changes requested):

src/sync/channel.zig
dnut: The function is able to return a bool indicating whether the send is actually necessary. The boolean indicates whether to send it. When it returns false, that means "we already handled it, no need to send it." That way you don't get redundant processing of the message.

src/sync/channel.zig
```diff
@@ -128,19 +143,27 @@ pub fn Channel(T: type) type {
 
         /// to deinit channels created with `create`
         pub fn destroy(channel: *Self) void {
+            const allocator = channel.allocator; // read allocator from channel before freeing it
```
dnut, line 146: Why is this necessary? `deinit` does not impact the allocator field.

src/trace/log.zig
dnut: The logger itself is already a self-managed allocation. It's not necessary to do nested allocations for its fields.

src/net/socket_utils.zig
dnut: The directionality depends on whether this is a send or receive socket. Also it's not clear what "incoming" means. Does it mean that it's channel used packets that are incoming from the network? Does it mean that it's for packets that are incoming from the caller?

src/net/socket_utils.zig
```diff
@@ -10,125 +9,109 @@ const Channel = sig.sync.Channel;
...
+            // empty the channel
+            while (outgoing_channel.tryReceive()) |_| {}
+            exit.afterExit();
+            logger.debug().log("sendSocket loop closed");
+        }
+
+        while (true) {
+            outgoing_channel.wait(exit) catch break;
+            while (outgoing_channel.tryReceive()) |p| {
+                if (exit.shouldExit()) break; // drop the rest (like above) if exit prematurely.
```
dnut, line 102:
```suggestion
        outer: while (true) {
            outgoing_channel.wait(exit) catch break;
            while (outgoing_channel.tryReceive()) |p| {
                if (exit.shouldExit()) break :outer; // drop the rest (like above) if exit prematurely.
```

src/gossip/service.zig
```diff
@@ -426,12 +434,15 @@ pub const GossipService = struct {
             exit_condition.ordered.exit_index += 1;
         }
 
-        try self.service_manager.spawn("[gossip] sendSocket", socket_utils.sendSocket, .{
+        // SocketPipe overrides `packet_outgoing_channel.send_hook`
```
dnut, line 437: Is this accurate? I don't see this happening anywhere inside SocketPipe.

src/shred_network/repair_service.zig
dnut: The advantage of decoupling them is that it's less opinionated about the relationship between channels and pipes. For example you may want multiple sockets using the same channel.

src/shred_network/shred_receiver.zig
```diff
@@ -41,68 +42,102 @@ pub const ShredReceiver = struct {
...
+        defer response_receiver_pipe.deinit(self.allocator);
+
+        // Create N pipes from turbine_socket -> turbine_receiver.
+        const turbine_receiver = try Channel(Packet).create(self.allocator);
+        turbine_receiver.send_hook = &receive_signal.hook;
+        defer turbine_receiver.destroy();
+
+        var turbine_receiver_pipes: std.BoundedArray(*SocketPipe, NUM_TVU_RECEIVERS) = .{};
+        defer for (turbine_receiver_pipes.slice()) |pipe| pipe.deinit(self.allocator);
+        for (0..NUM_TVU_RECEIVERS) |_| {
```
dnut, line 109: Given that we're moving towards reducing the number of threads dedicated to network IO, I'm curious on your opinion about using multiple threads here.

Agave actually uses eight threads for this. I decided to support the same logic of being able to run on multiple threads, basically just for feature parity, but I kept it a bit more conservative at 2 threads since 8 felt like overkill. Now I'm thinking we could just simplify this and just move it to one thread.

I don't know why they dedicated multiple threads to this. Turbine does receive a lot of data. But occasionally repair is expected to receive a lot more data than turbine ever receives. So I don't get it. Is it conceivable that a single CPU would be insufficient for simply reading packets and writing them to a channel?

src/shred_network/repair_service.zig
kprotty: This was the use-case of `SocketThread`. But of the 4 places using SocketPipe, there's one where it cant couple them:

`GossipService` can be initialized separately & optionally creates the pipe (among other things) in `.start()`. Some tests dont call start() and instead mimic a receiver pipe by pushing data directly to the channels themselves (see `test packet verification` and `benchmarkGossipService`).

src/shred_network/shred_receiver.zig
```diff
@@ -41,68 +42,102 @@ pub const ShredReceiver = struct {
...
+            .receiver,
+            self.logger.unscoped(),
+            self.repair_socket,
+            response_receiver,
+            exit,
+        );
+        defer response_receiver_pipe.deinit(self.allocator);
+
+        // Create N pipes from turbine_socket -> turbine_receiver.
+        const turbine_receiver = try Channel(Packet).create(self.allocator);
```
dnut, line 103: What do you think about the contention from sending to the same channel from multiple threads? Typical network conditions are about 5000 packets per second. Firedancer aims to support 1 million.

src/gossip/service.zig
kprotty: Not atm, leftover comment from when SocketPipe setup the ResetEvent SendHook itself (now its been moved into the channel).

src/shred_network/shred_receiver.zig
```diff
@@ -41,68 +42,102 @@ pub const ShredReceiver = struct {
 
     const Self = @This();
 
+    /// A shared instance of an event to support waiting on multiple Channels.
+    const ReceiverSignal = struct {
```
dnut, line 46: I can see now why you envisioned all signals being hooks.

What do you think about having the channels share the same ResetEvent? For example you could change this Channel field to a pointer:
```
event: *std.Thread.ResetEvent = .{},
```

src/shred_network/shred_receiver.zig
```diff
@@ -41,68 +42,102 @@ pub const ShredReceiver = struct {
...
+        event: std.Thread.ResetEvent = .{},
+        hook: Channel(Packet).SendHook = .{ .after_send = afterSend },
+
+        fn afterSend(hook: *Channel(Packet).SendHook, _: *Channel(Packet)) void {
+            const signal: *ReceiverSignal = @alignCast(@fieldParentPtr("hook", hook));
+            signal.event.set();
+        }
+
+        fn wait(signal: *ReceiverSignal, exit: ExitCondition) error{Exit}!void {
```
dnut, line 55:
```suggestion
        fn wait(self: *ReceiverSignal, exit: ExitCondition) error{Exit}!void {
```

src/sync/channel.zig
kprotty: Im actually fine with removing it for now, then adding it back in #465 if needed.

src/net/socket_utils.zig
0xNineteen: moving data 'from the network to a channel' or 'from a channel to the network' feels like its managing data related to sockets -- how would pipe describe it?

src/net/socket_utils.zig
0xNineteen: ah true i guess it could get confusing without more context

src/shred_network/shred_verifier.zig
```diff
@@ -29,9 +29,11 @@ pub fn runShredVerifier(
 ) !void {
     const metrics = try registry.initStruct(Metrics);
     var verified_merkle_roots = try VerifiedMerkleRoots.init(std.heap.c_allocator, 1024);
-    while (!exit.load(.acquire) or
-        unverified_shred_receiver.len() != 0)
-    {
+    while (true) {
+        unverified_shred_receiver.wait(.{ .unordered = exit }) catch |e| switch (e) {
+            error.Exit => if (unverified_shred_receiver.isEmpty()) break,
```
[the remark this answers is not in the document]
kprotty: Could call it `.waitToReceive()` and have it return early if `!isEmpty()`. It should keep the same semantics.

src/shred_network/shred_receiver.zig
kprotty: Wrote a quick [channel benchmark](<https://gist.github.com/kprotty/8e4f57e21e80992655e26d5823ee4266>) which tests throughput under max contention (constantly sending). M4 with 12 senders is ~70mil/sec (~20mil/sec with receiver). 5900x with 24 senders is ~30mil/sec (~15mi/sec with receiver).

The channel parts at least should hold up. I think the bigger bottlenecks would be either IO or thread parking/unparking on ResetEvent.

src/shred_network/shred_receiver.zig
kprotty: Maybe to offset inter-thread/channel communication? Not sure. I also think it could move to one thread. The saturation would matter more if the recvfrom CPU was doing many thread unparks on the channel's ResetEvent. Or maybe if its doing other IO work.

src/net/socket_utils.zig
0xNineteen: ah i see -- still not very clear imo - okay its a pipe and has something to do with a socket but cant really infer much more other than that from the name - would rather call it `SocketReadWriteThreads` or something much more explicit

src/shred_network/shred_receiver.zig
0xNineteen: we should keep this in sig as a load tester

src/net/socket_utils.zig
kprotty: ah ok. `SocketThread` sounds good then.

src/net/socket_utils.zig
kprotty: https://github.com/Syndica/sig/pull/486/commits/767607c3c46beb88cf4533d9bbabba71b9b3e634

src/net/socket_utils.zig
kprotty: https://github.com/Syndica/sig/pull/486/commits/3a56b911d94df285a9e3942b92a160dbefa91db1

src/net/socket_utils.zig
kprotty: https://github.com/Syndica/sig/pull/486/commits/767607c3c46beb88cf4533d9bbabba71b9b3e634

src/gossip/service.zig
kprotty: https://github.com/Syndica/sig/pull/486/commits/767607c3c46beb88cf4533d9bbabba71b9b3e634

src/sync/channel.zig
kprotty: https://github.com/Syndica/sig/pull/486/commits/767607c3c46beb88cf4533d9bbabba71b9b3e634 (removed `beforeSend` for now, reasoning: https://github.com/Syndica/sig/pull/486#discussion_r1917161745)

src/sync/channel.zig
kprotty: https://github.com/Syndica/sig/pull/486/commits/767607c3c46beb88cf4533d9bbabba71b9b3e634 (removed `beforeSend` for now, reasoning: https://github.com/Syndica/sig/pull/486#discussion_r1917161745)

src/sync/channel.zig
kprotty: https://github.com/Syndica/sig/pull/486/commits/767607c3c46beb88cf4533d9bbabba71b9b3e634

src/net/socket_utils.zig
kprotty: https://github.com/Syndica/sig/pull/486/commits/767607c3c46beb88cf4533d9bbabba71b9b3e634

src/gossip/service.zig
kprotty: https://github.com/Syndica/sig/pull/486/commits/767607c3c46beb88cf4533d9bbabba71b9b3e634

src/shred_network/shred_verifier.zig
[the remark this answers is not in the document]
kprotty: https://github.com/Syndica/sig/pull/486/commits/3281d2654aa2bfa4964d1b019866fdcd4cc9e825

src/shred_network/shred_receiver.zig
kprotty: https://github.com/Syndica/sig/pull/486/commits/767607c3c46beb88cf4533d9bbabba71b9b3e634

src/shred_network/shred_receiver.zig
kprotty: https://github.com/Syndica/sig/pull/486/commits/00592b9394273a16c0384109fb7295acd8158df4

src/shred_network/shred_receiver.zig
kprotty: Since it's only this case, decided to make the `turbine_socket` & `repair_socket` run their own `runPacketHandler` threads for now (instead of having them share the same thread + ReceiveEvent hook).

https://github.com/Syndica/sig/pull/486/commits/84836dda2f90eac80aa474ae4ded2c704ef08f2b

0xNineteen (approved):
lgtm :shipit:

dnut (approved):

dnut (approved):

PR #488, opened 2025-01-10
improve(quic): remove x509 cert workaround in favor of openssl calls

Rexicon226:
Agave does a hack where it precomputes the bytes of the certificate and public key because rustls doesn't support, something. I didn't exactly understand what it doesn't support.

I originally just copied straight from what agave was doing under the time pressure from the demo, however, now we can take a step back and fix up some of the uglier parts of the quic client. We're not using rustls, so we can confidently use the cleaner approach of constructing the certificate and keys using the boringssl API.

After a bit of fiddling around with the proper settings and realizing that `X509_gmtime_adj` returns an error if the value passed in is too large, this branch passes the transaction sender test and general usage by me.


yewman (approved):
lgtm

PR #489, opened 2025-01-10
feat(ci): ensure building passes on macos

Rexicon226:
I'm hoping to one day setup CI runners to run the test suite on macos as well, however, Github's CI runners seem to be trolling me a bit and keep swapping the runners between aarch64 and x86-64. This wouldn't be a problem if the aarch64 runner's hwinfo weren't seemingly screwed up and proving incorrect CPU feature detection.

For now, the best solution we can manage is just ensuring that the build succeeds when targeting aarch64-macos from x86_64-linux. Maybe we can setup self-hosted runners at one point?


build.zig.zon
```diff
@@ -36,8 +36,8 @@
             .hash = "12207766d25ba350d6e2f2153fc74a2b3ff204224e1c08adf211cd9e400075033898",
         },
         .lsquic = .{
-            .url = "https://github.com/Syndica/lsquic/archive/7c34f2472f390482dffada7a65b8231f6b5b3a3b.tar.gz",
-            .hash = "1220ff572d14b18d9774a298db0b8aeb046800039ccc8e10646992b074e26ea4c38a",
+            .url = "https://github.com/Syndica/lsquic/archive/ed6ced0cbc6447f7135a32db491e398debdf8af7.tar.gz",
```
Rexicon226, line 39: The update here is to vendor `sys/queue.h` so we can build lsquic on musl (which is the default when you set `x86_64-linux`).

.github/workflows/check.yml
```diff
@@ -36,6 +36,23 @@ jobs:
...
+    steps:
+        - name: checkout
+          uses: actions/checkout@v3
+        
+        - name: setup-zig
+          uses: mlugg/setup-zig@v1
+          with:
+            version: 0.13.0
+        
+        # ensure builds pass for common targets we have.
```
0xNineteen, line 50: might be good to add a comment on context why we dont run the tests for mac

0xNineteen (changes requested):
small nit - otherwise lgtm

.github/workflows/check.yml
Rexicon226: Done in 75c3b9b8

0xNineteen (approved):
:shipit:

0xNineteen (approved):

PR #490, opened 2025-01-12
refactor(build,download): Remove curl dependency, replace with stdlib's http client

InKryption:
Written to retain equivalent or better download speed, and some enhancement to the behavior of output.
Now in the absence of C wrappers and callback indirection, I find the code to be much simpler to understand; I also simplified some of the logic in general.

I also took the liberty of now returning the resulting file, and propagating the resulting full and incremental snapshot file handles; for now this doesn't do anything, but in the future I would like to explore propagating them further in order to short-circuit the snapshot detecting logic - as in, making it unnecessary to scan the filesystem if we've just downloaded a snapshot pair.

Lastly, I tidied up a couple of other things, and publicized some decls that are depended on by a public function.

### DIFF
--- a/src/accountsdb/download.zig
+++ b/src/accountsdb/download.zig
@@ -1,7 +1,6 @@
 //! logic for downloading a snapshot
 
 const std = @import("std");
-const curl = @import("curl");
 const sig = @import("../sig.zig");
 
 const SlotAndHash = sig.accounts_db.snapshots.SlotAndHash;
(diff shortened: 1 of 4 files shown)


src/accountsdb/download.zig
```diff
@@ -227,321 +229,188 @@ pub fn downloadSnapshotsFromGossip(
...
-            return;
+            return .{ full_archive_file, inc_archive_file };
         }
     }
 }
 
-const DownloadProgress = struct {
-    file: std.fs.File,
-    min_mb_per_second: ?usize,
+const DOWNLOAD_FILE_BUFFER_SIZE = 1 * BYTE_PER_MIB + 4096;
```
0xNineteen, line 328: can we keep all global consts at the top of the file

0xNineteen: yeah nbd - i would say its more related to the actual buffer initialization in `downloadSnapshotsFromGossip` (not downloadFile) -- which caused my confusion and then i search at the top of the file where i expect all constants to be and its not there -- would rather keep it consistent and all at the top

InKryption: Well, the buffer initialized in that function is initialized with the constraints defined by `downloadFile` - the latter defines the kind of buffer it wants.

InKryption: Amended as discussed on slack:

26cb7b1d8a8852edd5b6431f80bb8a2c618203a5

0xNineteen (approved):

PR #492, opened 2025-01-13
refactor(build): Add 'no-bin' & rework install/run behavior, rename 'run' to 'sig'

InKryption:
Basic changes:
* Adds no-bin.
* Renames 'run' to 'sig'.
* Re-organizes 'doc' step code.
* Re-organizes and makes some code more consistent.
* Step dependency graph rework:
  An artifact's top level step now by default depends on the compile step, the install step, and the run step; the install and run steps can be individually disabled using `no-bin` and `no-run`. If both are disabled, there now will be no binary even compiled, allowing us to run semantic analysis without wasting time emitting a binary.


dnut:
What do you think about `norun` and `nobin` to make it easier to type? Are those names still clear? I'm mildly annoyed by typing dashes haha. Not a strong preference

InKryption:
I'm not against it; if nobody else has an opinion against, I'm fine to change it to that, otherwise I'll push the change in a few.

0xNineteen (approved):
nice :shipit:

PR #493, opened 2025-01-14
fix(shred-network): keep up with turbine

dnut:
> ⚠️ *Note: most of the lines added in this PR are for the grafana dashboard*

This change was primarily motivated as a bugfix to deal with the issue that the shred collector would consistently get stuck at a certain slot and stop progressing.

### Problem

What this means is that the shred tracker would think that it hasn't received all the shreds for a certain slot, and it would keep trying to repair that slot. Eventually it would lag so far behind that it would stop tracking any future slots ahead. This is because the shred tracker only worries about 1024 slots at a time. So the repair service would get into a degraded state with poor performance because it's looking at a ton of slot unnecessarily, and it would lose the ability to repair slots beyond a certain point. And our metrics in grafana would appear like the shred collector has halted.

The general problem was that slots were skipped but the shred tracker would not understand that it was skipped, so it would keep trying to repair that slot. There are a lot of complicated scenarios were slots can be skipped, and the repair service was not doing a good job of acquiring overall context about these scenarios.

### Fix

The fix was to to make the repair service more persistent about requesting high-level information about each slot. It uses `HighestShred` repairs to get general context about slots and there were some situations where it wouldn't send those requests for slots. For example if it already send out a lot of `Shred` requests to ask for specific shreds, it might stop sending out HighestShred repairs slots that came after it (which may be skipping it). Also it did not send out HighestShred repairs if it already thought a slot was complete. I changed it to always send out HighestShred repairs for every slot it is aware of, rather than making assumptions about those slots, or stopping due to reaching a limit.

### Test

With these changes, the behavior is fixed. I no longer see the shred tracker getting stuck. It is now about able to catch up from behind and remain caught up. I've tested this dozens of times starting from 1000 slots behind, and it catches up reliably within a minute or so, and stays caught up. After about an hour it runs out of memory due to a leak, which is a separate issue that I'll fix in another pr.

## Other changes

I made some other changes that are mainly for performance and simplification of the code.

In a way I actually *reduced* the eagerness of sending HighestShred repairs for future slots. There's a difference between sending out requests for slots we know about and are caught behind (what I described increasing above) and sending out requests for slots that don't even appear to exist (this is what I decreased). It's not necessary to constantly send out hundreds of requests per second for future slots that have not happened yet. It's good enough to just send out a few probing requests to see if we're behind. Otherwise turbine will let us know where we are, and we can fill in the gaps with the logic described above.

I made the repair service a bit more dynamic in its resource usage. When catching up from far behind, it needs to scale up the number of threads it uses to sign and serialize tens of thousands of repair requests. Once it's caught up, it can work with just a single thread.

I made some changes to the way that shreds are registered with the tracker by moving this into the ShredInserter. This unlocked the ability to also register recovered shreds, so we don't attempt to repair shreds that have already been recovered with Reed Solomon. Eventually we can probably remove BasicShredTracker since its role is partly handled by the Index in the blockstore. But for now the shred tracker is working fine.

### DIFF
--- a/src/cmd/config.zig
+++ b/src/cmd/config.zig
@@ -15,7 +15,7 @@ const getAccountPerFileEstimateFromCluster =
 pub const Config = struct {
     identity: IdentityConfig = .{},
     gossip: GossipConfig = .{},
-    shred_network: ShredCollectorConfig = shred_network_defaults,
+    shred_network: ShredNetworkCliArgs = .{},
     accounts_db: AccountsDBConfig = .{},
     geyser: GeyserConfig = .{},
     turbine: TurbineConfig = .{},
@@ -110,10 +110,22 @@ pub const GossipConfig = struct {
     }
 };
 
-pub const shred_network_defaults = ShredCollectorConfig{
-    .turbine_recv_port = 8002,
-    .repair_port = 8003,
-    .start_slot = null,
+const ShredNetworkCliArgs = struct {
+    start_slot: ?sig.core.Slot = null,
+    repair_port: u16 = 8003,
+    turbine_recv_port: u16 = 8002,
+    no_retransmit: bool = true,
+    dump_shred_tracker: bool = false,
+
+    pub fn toConfig(self: ShredNetworkCliArgs, fallback_slot: sig.core.Slot) ShredCollectorConfig {
+        return .{
+            .start_slot = self.start_slot orelse fallback_slot,
+            .repair_port = self.repair_port,
+            .turbine_recv_port = self.turbine_recv_port,
+            .retransmit = !self.no_retransmit,
+            .dump_shred_tracker = self.dump_shred_tracker,
+        };
+    }
 };
 
 /// Analogous to [AccountsDbConfig](https://github.com/anza-xyz/agave/blob/4c921ca276bbd5997f809dec1dd3937fb06463cc/accounts-db/src/accounts_db.rs#L597)
--- a/src/cmd/config.zig
+++ b/src/cmd/config.zig
@@ -110,6 +110,10 @@ pub const GossipConfig = struct {
     }
 };
 
+/// The command-line arguments that are used to configure the shred network. The
+/// CLI args are slightly different from the `shred_network.start` inputs, so it
+/// gets its own struct. `ShredNetworkConfig` represents the inputs to the start
+/// function.
 const ShredNetworkCliArgs = struct {
     start_slot: ?sig.core.Slot = null,
     repair_port: u16 = 8003,
@@ -117,6 +121,7 @@ const ShredNetworkCliArgs = struct {
     no_retransmit: bool = true,
     dump_shred_tracker: bool = false,
 
+    /// Converts from the CLI args into the `shred_network.start` parameters
     pub fn toConfig(self: ShredNetworkCliArgs, fallback_slot: sig.core.Slot) ShredNetworkConfig {
         return .{
             .start_slot = self.start_slot orelse fallback_slot,
--- a/src/cmd/config.zig
+++ b/src/cmd/config.zig
@@ -2,7 +2,7 @@ const std = @import("std");
 const sig = @import("../sig.zig");
 
 const ACCOUNT_INDEX_SHARDS = sig.accounts_db.db.ACCOUNT_INDEX_SHARDS;
-const ShredCollectorConfig = sig.shred_network.ShredCollectorConfig;
+const ShredNetworkConfig = sig.shred_network.ShredNetworkConfig;
 const IpAddr = sig.net.IpAddr;
 const LogLevel = sig.trace.Level;
 const Cluster = sig.core.Cluster;
@@ -117,7 +117,7 @@ const ShredNetworkCliArgs = struct {
     no_retransmit: bool = true,
     dump_shred_tracker: bool = false,
 
-    pub fn toConfig(self: ShredNetworkCliArgs, fallback_slot: sig.core.Slot) ShredCollectorConfig {
+    pub fn toConfig(self: ShredNetworkCliArgs, fallback_slot: sig.core.Slot) ShredNetworkConfig {
         return .{
             .start_slot = self.start_slot orelse fallback_slot,
             .repair_port = self.repair_port,
--- a/src/ledger/shred_inserter/shred_inserter.zig
+++ b/src/ledger/shred_inserter/shred_inserter.zig
@@ -150,6 +150,7 @@ pub const ShredInserter = struct {
         retransmit_sender: ?PointerClosure([]const []const u8, void),
         shred_tracker: ?*sig.shred_network.shred_tracker.BasicShredTracker,
     ) !InsertShredsResult {
+        const milli_timestamp = std.time.milliTimestamp();
         ///////////////////////////
         // check inputs for validity and edge cases
         //
@@ -191,7 +192,8 @@ pub const ShredInserter = struct {
             switch (shred) {
                 .data => |data_shred| {
                     if (shred_tracker) |tracker| {
-                        tracker.registerDataShred(&shred.data) catch |err| switch (err) {
+                        tracker.registerDataShred(&shred.data, milli_timestamp) catch |err|
+                            switch (err) {
                             error.SlotUnderflow, error.SlotOverflow => {
                                 self.metrics.register_shred_error.observe(@errorCast(err));
                             },
@@ -284,7 +286,8 @@ pub const ShredInserter = struct {
                     continue;
                 }
                 if (shred_tracker) |tracker| {
-                    tracker.registerDataShred(&shred.data) catch |err| switch (err) {
+                    tracker.registerDataShred(&shred.data, milli_timestamp) catch |err|
+                        switch (err) {
                         error.SlotUnderflow, error.SlotOverflow => {
                             self.metrics.register_shred_error.observe(@errorCast(err));
                         },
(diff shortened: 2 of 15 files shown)


src/shred_network/shred_tracker.zig
```diff
@@ -211,27 +255,61 @@ pub const SlotReport = struct {
...
+        std.debug.assert(index < ShredSet.bit_length);
+        const mask_bit = maskBit(index);
+        const mask_index = maskIndex(index);
+        defer self.masks[mask_index] |= mask_bit;
+        return self.masks[mask_index] & mask_bit != 0;
+    }
+
+    fn maskBit(index: usize) ShredSet.MaskInt {
+        return @as(ShredSet.MaskInt, 1) << @as(ShredSet.ShiftInt, @truncate(index));
+    }
```
yewman, line 269:
```suggestion
    }
    
```

src/shred_network/shred_tracker.zig
```diff
@@ -47,89 +50,137 @@ pub const BasicShredTracker = struct {
...
+            }
         }
+        return end - start;
     }
 
     /// returns whether it makes sense to send any repair requests
     pub fn identifyMissing(
         self: *Self,
         slot_reports: *MultiSlotReport,
+        milli_timestamp: i64,
```
yewman, line 171: Thoughts on using sig time types?

yewman (commented):
Just a couple questions otherwise lgtm

src/ledger/shred_inserter/shred_inserter.zig
```diff
@@ -189,6 +191,15 @@ pub const ShredInserter = struct {
             const shred_source: ShredSource = if (is_repair) .repaired else .turbine;
             switch (shred) {
                 .data => |data_shred| {
+                    if (shred_tracker) |tracker| {
+                        tracker.registerDataShred(&shred.data, milli_timestamp) catch |err|
+                            switch (err) {
```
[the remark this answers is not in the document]
dnut: 2fc5b5c0c22b15335acf846dd7e0553120165418

src/shred_network/shred_tracker.zig
dnut: 14a9b6767e74b3042f9d4b163688eed47f6893c5

src/shred_network/shred_tracker.zig
dnut: c4178eaa2d7acd453991b2564e909573c36ce618

src/shred_network/repair_service.zig
```diff
@@ -52,11 +53,11 @@ pub const RepairService = struct {
     shred_tracker: *BasicShredTracker,
     logger: ScopedLogger(@typeName(Self)),
     exit: *Atomic(bool),
-    last_big_request_timestamp_ms: i64 = 0,
     /// memory to re-use across iterations. initialized to empty
     report: MultiSlotReport,
     thread_pool: RequestBatchThreadPool,
     metrics: Metrics,
+    rng: std.Random.DefaultPrng,
```
0xNineteen, line 60: i think the convention is to name this `prng` -- @InKryption is this correct? do we have a link to where we defined the random rules?

src/shred_network/repair_service.zig
```diff
@@ -110,14 +123,17 @@ pub const RepairService = struct {
         self.report.deinit();
     }
 
-    const min_loop_duration_ns = 100 * std.time.ns_per_ms;
+    const min_loop_duration_ns = 500 * std.time.ns_per_ms;
```
0xNineteen, line 126: sig Durations instead of `_ns` ?

src/shred_network/repair_service.zig
```diff
@@ -110,14 +123,17 @@ pub const RepairService = struct {
...
 
     pub fn run(self: *Self) !void {
         var waiting_for_peers = false;
         var timer = try std.time.Timer.start();
         var last_iteration: u64 = 0;
         while (!self.exit.load(.acquire)) {
-            if (self.sendNecessaryRepairs()) |_| {
+            _ = timer.read();
```
0xNineteen, line 133: this does nothing?

src/shred_network/repair_service.zig
```diff
@@ -158,71 +182,56 @@ pub const RepairService = struct {
             }
         }
 
-        // TODO less often
-        if (addressed_requests.items.len > 0) {
-            self.logger.debug().logf(
-                "sent {} repair requests",
```
0xNineteen, line 164: nice 👌 - logs be gone

src/shred_network/repair_service.zig
```diff
@@ -495,6 +545,10 @@ pub const RepairPeerProvider = struct {
     }
 };
 
+/// The maximum time to wait between repair loop iterations.
+/// Sort of like a timeout for repair responses.
+const MAX_REPAIR_DELAY_MILLIS = 1_000;
```
0xNineteen, line 550: Duration? -- feel like we should push for always using sig durations in constants, wdyt? -- maybe add to contributing too?

src/shred_network/service.zig
```diff
@@ -203,6 +207,20 @@ pub fn start(
         .{repair_svc},
     );
 
+    if (conf.dump_shred_tracker) {
+        try service_manager.spawn("dump shred tracker", struct {
```
0xNineteen, line 211: this is a good tmp solution but i feel like maybe we can push these to geyser features later? (same as the gossip dump) maybe as a future ticket?

0xNineteen (changes requested):
overall looks great - few things

src/shred_network/repair_service.zig
dnut: Harold had a similar comment on some other code. It was a bit complicated since our time types did not have sufficient support for what I needed. Sometimes it's more straightforward to work with integers than opaque time structs. But I added the necessary methods for what he requested. I'll see about these other ones too.

src/shred_network/service.zig
dnut: I don't understand. How does this relate to geyser?

src/shred_network/repair_service.zig
dnut: ba1baafae1c09cefc295c08a238dce1828f697b5

src/shred_network/repair_service.zig
dnut: 403d5f7a71d7597123eae9999bae8913912bfc40

src/shred_network/service.zig
0xNineteen: im guessing the main reason you would want to dump shred information is to debug the information externally / while the process is still running - you can dump this information in a text file while it runs in a separate thread (ie, what we do in gossip) - or just stream the data out of the process using geyser and then having a reading process handle the rest (which is what im thinking - at least for a future ticket)

src/shred_network/repair_service.zig
dnut: I know you resolved this but I made some changes already. Also tried to document the code better since it's a bit arbitrary. c323f03d45f332b533ef2a96d7ba81144f6325b4

src/shred_network/service.zig
dnut: oh that makes sense. I wasn't thinking of geyser so generally. That's a good idea but it would require a more complex setup in order to debug the code

src/shred_network/repair_service.zig
dnut: This wasn't even used, I deleted it here c323f03d45f332b533ef2a96d7ba81144f6325b4

src/cmd/config.zig
```diff
@@ -110,10 +110,22 @@ pub const GossipConfig = struct {
     }
 };
 
-pub const shred_network_defaults = ShredCollectorConfig{
-    .turbine_recv_port = 8002,
-    .repair_port = 8003,
-    .start_slot = null,
+const ShredNetworkCliArgs = struct {
```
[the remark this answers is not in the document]
dnut: If we want to allow the cli and the service to have different apis from each other, then I don't see a good way to do it. This could be done by adding some global variables, but then it's more of a burden to ensure you're setting the values correctly for every command.

```zig
var cli_no_retransmit = false;
var cli_start_slot: ?Slot = null;

fn run() {
    var retransmit_option = cli.Option{
        .value_ref = cli.mkRef(&cli_no_retransmit),
    };
    var test_repair_option = cli.Option{
        .value_ref = cli.mkRef(&cli_start_slot),
    };
}

fn validator() {
    config.current.shred_network.retransmit = !no_retransmit;
    config.current.shred_network.start_slot = cli_start_slot orelse fallback;
}

fn shredCollector() {
    config.current.shred_network.retransmit = !no_retransmit;
    config.current.shred_network.start_slot = cli_start_slot orelse fallback;
}
```

Alternatively we can just force the service to conform to the CLI. It would accept `no_retransmit` instead of `retransmit` and `start_slot` would be optional. I think this solution would be better than the above. I just thought it made more sense for the service to have slightly different inputs than the CLI. but it's not important.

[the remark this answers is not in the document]
0xNineteen: ah i see, makes sense - what if we just rename ShredNetworkCliArgs to ShredNetworkConfig ? the naming inconsistency is the only thing that seems off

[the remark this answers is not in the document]
dnut: Actually neither of them should say "collector" in the name. They're both for the shred network. The difference is that one of them represents the config parameters for the shred network's start function (currently called ShredCollectorConfig) and the other one is for cli arguments (currently called ShredNetworkCliArgs)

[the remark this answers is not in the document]
dnut: My instinct would be to just rename the config struct to ShredNetworkConfig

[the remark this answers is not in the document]
0xNineteen: ah okay - yeah that name change would make the most sense then - maybe just add a comment about the Cli struct about the difference between the two / why we have a different naming scheme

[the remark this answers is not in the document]
dnut: comment: 8f66f721d5eb62aba2473cf43489da78b9eb18f4

0xNineteen (approved):
:shipit:

PR #495, opened 2025-01-15
fix(ledger): fuzzer loop bug and add ledger fuzzing to ci

dnut:


.github/workflows/check.yml
```diff
@@ -176,6 +176,26 @@ jobs:
...
+        with:
+          submodules: recursive
+      - name: setup zig
+        uses: mlugg/setup-zig@v1
+        with:
+          version: 0.13.0
+      - name: build
+        run: zig build -Doptimize=ReleaseSafe -Dno-run fuzz
+      - name: run
+        run: ./zig-out/bin/fuzz ledger 19 10000
```
dadepo, line 197: Might be an idea to have the seed randomly generated on each run? Because right now we are fuzzing the same path with every CI build given we have a fixed seed.

dadepo (approved):

.github/workflows/check.yml
dnut: I actually had the same thought when I was doing this.

https://github.com/orgs/Syndica/projects/2/views/10?pane=issue&itemId=93890772

PR #498, opened 2025-01-15
feat(CI): Randomize fuzzer seed on CI

dadepo:


src/gossip/fuzz_service.zig
```diff
@@ -35,6 +35,7 @@ const SLEEP_TIME = Duration.zero();
 // const SLEEP_TIME = Duration.fromSecs(10);
 
 pub fn run(seed: u64, args: *std.process.ArgIterator) !void {
+    std.debug.print("Running gossip_service fuzzer with seed: {}\n", .{seed});
```
dnut, line 38: `src/fuzz.zig` already has `std.debug.print("using seed: {d}\n", .{seed});`

dadepo: [54c8504](https://github.com/Syndica/sig/pull/498/commits/54c8504757b350bd2edcb4c3ab3a6fbdaad0ebc8)

dnut:
I like that this change increases test coverage. Each time CI runs, it's going to run additional test cases. We're more likely to detect bugs like this.

I also like that this can be accomplished without sacrificing reproducibility. The seed can be used locally to run the same test.

My concern is that this could make CI flaky. We require certain checks to pass before merging a PR to ensure that *this specific change* did not introduce a bug. If the tests are flaky, then you never really know which change was responsible for the failure. However much this randomness improves our test coverage, it would equally harm the fidelity of our PR checks. This could result in changes getting blocked by something unrelated. It could also result in people ignoring the fuzzer results and just retrying them until they succeed.

So I'm torn about introducing this change and would like other opinions. cc: @0xNineteen 

I'll propose an alternative as well. What if we keep the PR-based CI process more consistent, but we add a separate fuzzing job that runs on main, with the goal of increasing test coverage beyond what would be practical to run for PRs. This job could run for much more time, and it could use random seeds.

dadepo:
> My concern is that this could make CI flaky. We require certain checks to pass before merging a PR to ensure that _this specific change_ did not introduce a bug. If the tests are flaky, it's hard to tell which change was responsible for the failure.

That is indeed a valid concern

> Another idea would be to add a separate fuzzing job that runs on main, with the goal of increasing test coverage beyond what would be practical to run for PRs. This job could run for much more time, and it could use random seeds. We would lose the visibility and coverage of running a new test on every single PR commit, but it could provide similar benefits without the fidelity tradeoff.
>> imo this would likley be the best approach - a continuously running fuzzing program which tracks main 😈 -

If this is set up in such a way that the seeds can also be randomized (I think this also adds to the randomness apart from being long running) then this should work

> We would lose the visibility and coverage of running a new test on every single PR commit, but it could provide similar benefits without the fidelity tradeoff.

Maybe we could set up some alerting system? That way we do get notified if the continuous fuzzer hits a bug

dadepo:
Closing this in favour of the issues that would come out of this https://github.com/orgs/Syndica/projects/2/views/10?pane=issue&itemId=94007773

dadepo:
@dnut This just crossed my mind [b8f8172](https://github.com/Syndica/sig/pull/498/commits/b8f81722f7921822f6d8330002f952719dc3ba42)

Will that be a relatively inexpensive way to get more random fuzzing with the CI without the flakiness issue you raised?

0xNineteen:
> Will that be a relatively inexpensive way to get more random fuzzing with the CI without the flakiness issue you raised?

the flakiness comes from having different seeds - deriving the seed from the commit hash will still be different seed for each PR which could result in random failures

dadepo:
> the commit hash will still be different seed for each PR which could result in random failures unrelated to the PR who is trying to get merged

True, the part it improves on is this part:

> It could also result in people ignoring the fuzzer results and just retrying them until they succeed

Because in this case, retrying won't make the test succeed.

imo the benefit of deriving the seed from the hash is that we increase the probability of unearthing issues via each CI run as new seeds get used by the fuzzer, with the drawback that there is a chance the bug found was not introduced in that particular PR, which I think is an improvement over the current state where using a fixed seed means there is no possibility for the bug to be found anyways as a fix seed is always used. 

But having said that, the continuous fuzzing is also another good approach.

PR #512, opened 2025-01-23
perf(channel): fix contention in benchmarks

kprotty:
Have a continous stream of senders (& maybe a receiver) then observe and report the channel throughput from the outside as nanoseconds per `send()`. This avoids senders completing early from only having to push a fixed amount of items, thus reducing contention (as they're out of the race) and randomizing throughput results.

### DIFF
--- a/src/sync/channel.zig
+++ b/src/sync/channel.zig
@@ -594,127 +594,90 @@ pub const BenchmarkChannel = struct {
 
     pub const BenchmarkArgs = struct {
         name: []const u8 = "",
-        n_items: usize,
         n_senders: usize,
-        n_receivers: usize,
+        receives: bool,
     };
 
     pub const args = [_]BenchmarkArgs{
         .{
-            .name = "  10k_items-   1_senders-   1_receivers ",
-            .n_items = 10_000,
+            .name = "1_senders-   1_receivers ",
             .n_senders = 1,
-            .n_receivers = 1,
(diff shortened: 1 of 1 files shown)


src/sync/channel.zig
```diff
@@ -594,127 +594,90 @@ pub const BenchmarkChannel = struct {
...
-            .n_senders = 8,
-            .n_receivers = 8,
-        },
-        .{
-            .name = "   1m_items-  16_senders-  16_receivers ",
-            .n_items = 1_000_000,
-            .n_senders = 16,
-            .n_receivers = 16,
+            .name = "N_senders-   1_receivers ",
+            .n_senders = std.math.maxInt(usize),
```
0xNineteen, line 614: would be good to add a comment somewhere saying maxInt == n_cpus

0xNineteen (changes requested):
small nit - otherwise lgtm

src/sync/channel.zig
Rexicon226: What about having it be an optional instead, `null` meaning `n_cpus`? That seems like a better API to me.

kprotty: Took null approach https://github.com/Syndica/sig/pull/512/commits/eb8092ea573799c9e8f576d36998b361f78357bf

0xNineteen (approved):
nice :shipit:

PR #513, opened 2025-01-23
refactor(svm): refactor transaction implementation 

yewman:
# Issue
Our initial implementation of transactions was heavily based off Agave to support the development of other components. The way agave implements transactions is overly complex, involving the following traits and structs, of which we implement the first seven:
- VersionedTransaction
- VersionedMessage
- Message
- V0Message
- MessageHeader
- CompiledInstruction
- MessageAddressTableLookup
- SanitizedTransaction
- SanitizedVersionedTransaction
- SanitizedMessage
- SanitizedVersionedMessage
- SVMTransaction
- SVMMessage
- SVMInstruction

# Resolution
This PR simplifies our current transaction implementation whilst satisfying our current requirements, it implements the following structs:
- sig.core.Transaction 
- sig.core.Transaction.Instruction 
- sig.core.Transaction.AddressLookup
- sig.utils.io.CheckedReader

The CheckedReader is used to read from a slice when deserialising transactions. It returns a NoBytesLeft error when attempting to access outside the underlying bounds. It differs from an AnyReader with a std.io.FixedBufferStream context in that provides some methods which reference the underlying slice directly when returning from function calls, rather than reading into a provided buffer. This may not be required anymore and could potentially be replaced by using a fixed buffer stream reader. From chatting with David, however, he mentioned some other potential use cases for such a reader so perhaps it has a place even if it is not used for transaction deserialisation right now.

Whilst functionality provided by the Agave implementations in Agave will undoubtedly become necessary down the line, this refactor is based on the philosophy that it is better to omit complexity until it is required. 

# Impacts
The changes to other modules are limited, there are two noteworthy areas of change:

1. Entry Serialize/Deserialize: The Entry struct contains a list of transactions which now use a custom deserialise method with a CheckedReader, for this reason the default bincode deserialisation of Entry's no longer worked. Custom serialise and deserialise methods where implemented to account for this. If the CheckedReader is reverted to anytype, we could revert to using bincode config to set the deserialize method for the Transaction struct.

2. Transfer Instruction Compilation: This involved the removal of the transfer instruction compilation logic previously contained in `src/core/transaction.zig`. This implementation was auxiliary to the gulf stream milestone and has been replaced by a manual transaction builder in `src/transaction_sender/mock_transfer_generator.zig`. This better reflects its purpose and removes otherwise redundant code. Transaction compilation can be re-implemented down as part of an sdk or purely for testing purposes.


src/core/transaction.zig
```diff
@@ -110,521 +49,372 @@ pub const V0Message = struct {
...
-    message: Message,
+    pub const Version = enum(u8) {
+        /// Legacy transaction without address lookups.
+        Legacy = 0xFF,
+        /// Transaction with address lookups.
+        V0 = 0x00,
+    };
 
-    pub const @"!bincode-config:signatures" = shortVecConfig([]const Signature);
+    pub const Instruction = struct {
```
0xNineteen, line 89: can we keep this out of the struct to reduce nesting?

src/core/transaction.zig
```diff
@@ -110,521 +49,372 @@ pub const V0Message = struct {
...
+            };
+        }
+    };
 
-    pub fn sanitize(self: *const Transaction) !void {
-        const num_required_sigs = self.message.header.num_required_signatures;
-        const num_signatures = self.signatures.len;
-        if (num_required_sigs > num_signatures) {
-            return error.InsufficientSignatures;
+    pub const AddressLookup = struct {
```
0xNineteen, line 148: same here

src/core/transaction.zig
```diff
@@ -110,521 +49,372 @@ pub const V0Message = struct {
...
+                .writable_indexes = writable_indexes,
+                .readonly_indexes = readonly_indexes,
+            };
+        }
+    };
 
-    pub fn clone(self: *const Message, allocator: std.mem.Allocator) error{OutOfMemory}!Message {
-        const account_keys = try allocator.dupe(Pubkey, self.account_keys);
-        errdefer allocator.free(account_keys);
+    pub fn clone(self: Transaction, allocator: std.mem.Allocator) !Transaction {
```
0xNineteen, line 207: can we keep clone below deinit? noticed its like that some places in the other codebase which it shouldnt be imo

src/core/transaction.zig
```diff
@@ -110,521 +49,372 @@ pub const V0Message = struct {
...
-            if (ci.program_id_index == 0) {
-                return error.ProgramIdCannotBePayer;
-            }
-            for (ci.accounts) |ai| {
-                if (ai >= self.account_keys.len) {
-                    return error.AccountIndexOutOfBounds;
-                }
-            }
-        }
+    pub fn sanitize(self: Transaction) !void {
```
0xNineteen, line 251: in other places (gossipdata) we use `verify` instead of sanitize -- wdyt

0xNineteen (changes requested):
overall lgtm - just a nit on the struct nesting

src/core/transaction.zig
yewman: We can, although I do think that this is prime example of where nesting makes sense since the Instruction type makes no sense without an associated Transaction. It sort of leaves me wondering if we do not want nesting in an instance such as this, what would justify nesting? I am pretty impartial, just trying to understand for the future

dnut (commented):
just some initial thoughts (need to submit review before i can reply in other threads)

src/core/transaction.zig
dnut: I can see how this is a reasonable candidate for nesting, but I don't think it's the strongest example. This is highly subjective so I'll just explain my personal thoughts.

I wouldn't go so far to say that Instructions make no sense without an associated Transaction. I wrote solana program client libraries that dealt extensively with `Instruction` but they never did anything with `Transaction`. From the perspective of a client, I think instructions are in the domain of the program's client library, and transactions are in the domain of the wallet.

I think the strongest case of nested structs is where the inner struct would never be used in any situation except when directly interacting with the outer struct. For example if you have a `Params` struct that defines the parameters to one of the struct methods.

Also I think it becomes hard to make sense of code when there is a lot of nesting. When you have a method inside a struct inside another struct that also has methods, and both structs are larger than what you can fit on the screen, the context can get a little confusing. Typically I would prefer nested structs to have no methods, or at least be very small. I think of nested structs as a way to combine a few pieces of data together for convenient usage in the outer struct. When it has a bunch of logic of its own, it starts to feel like a first-class citizen and not just a helper for another struct.

src/core/transaction.zig
dnut, line 251: This is already implemented in main. Shouldn't be too hard to port over.

src/core/transaction.zig
```diff
@@ -110,521 +49,372 @@ pub const V0Message = struct {
...
+        }
     }
 };
 
-pub const AccountMeta = struct {
-    pubkey: Pubkey,
-    is_signer: bool,
-    is_writable: bool,
+/// Serialize a **valid** transaction to a slice of bytes.
+pub fn serialize(writer: anytype, data: anytype, params: sig.bincode.Params) !void {
```
dnut, line 288: Do you think we could lean on bincode more for transaction serialization? Something like this?

```zig
    pub const @"!bincode-config:signatures" = sig.bincode.shortvec.sliceConfig([]Signature);
    pub const @"!bincode-config:account_keys" = sig.bincode.shortvec.sliceConfig([]const Pubkey);
    pub const @"!bincode-config:instructions" = sig.bincode.shortvec.sliceConfig([]const Instruction);
    pub const @"!bincode-config:address_table_lookups" = sig.bincode.shortvec.sliceConfig([]const AddressLookup);

    pub const @"!bincode-config:version": sig.bincode.FieldConfig(Version) = .{
        .deserializer = deserializeVersion,
        .serializer = serializeVersion,
    };

    /// Bit mask that indicates whether a serialized message is versioned.
    const MESSAGE_VERSION_PREFIX: u8 = 0x80;

    fn serializeVersion(writer: anytype, data: anytype, _: sig.bincode.Params) !void {
        switch (data) {
            .Legacy => {},
            .V0 => try writer.writeByte(0x80),
        }
    }

    fn deserializeVersion(
        _: std.mem.Allocator,
        original_reader: anytype,
        _: sig.bincode.Params,
    ) !Version {
        var peekable = sig.utils.io.peekableReader(original_reader);
        const reader = peekable.reader();

        return if (try peekable.peekByte() & MESSAGE_VERSION_PREFIX != 0)
            switch (try reader.readByte() & ~MESSAGE_VERSION_PREFIX) {
                0 => .V0,
                127 => error.OffChainMessage,
                else => error.InvalidMessageTag,
            }
        else
            return .Legacy;
    }

```

src/core/transaction.zig
```diff
@@ -110,521 +49,372 @@ pub const V0Message = struct {
...
-    allocator: std.mem.Allocator,
-    from_pubkey: Pubkey,
-    to_pubkey: Pubkey,
-    lamports: u64,
-) !Instruction {
-    var account_metas = try allocator.alloc(AccountMeta, 2);
-    account_metas[0] = AccountMeta.newMutable(from_pubkey, true);
-    account_metas[1] = AccountMeta.newMutable(to_pubkey, false);
-    return try Instruction.initSystemInstruction(
+test "legacy_transaction_parse" {
```
dnut, line 365: Can you bring back the old tests? I'm pretty sure they are all still applicable, they just might need some minor adjustments for the new structure.

And you don't like `bincode.testRoundTrip` which was written for exactly this use case? This same test was previously written as two lines of code.

dnut (changes requested):

src/core/transaction.zig
yewman: I agree with almost everything you say here, however, I'd like to clarify that the `Instruction` defined here does not make sense without an associated transaction since they contian indexes into the accounts array of their associated transaction. I imagine the `Instruction` you mention dealing would have been [this one](https://github.com/anza-xyz/agave/blob/faea52f338df8521864ab7ce97b120b2abb5ce13/sdk/program/src/instruction.rs#L332) yet the Instruction we are discussing here is akin to the Agave [CompiledInstruction](https://github.com/anza-xyz/agave/blob/faea52f338df8521864ab7ce97b120b2abb5ce13/sdk/program/src/instruction.rs#L641). Given that this has evidently been misleading I will make the name more descriptive when removing the nesting. I think going with `TransactionInstruction` makes sense since the name reflects its close link to transactions, however, we could mirror agave and use `CompiledInstruction`. Let me know if you have a preference otherwise I'll proceed with `TransactionInstruction`

src/core/transaction.zig
yewman: The issue here is that the peekable reader does not acutally peek when called in the above manner, the peekable reader provides true peek functionality only if the peeked byte is 'read' by the peekable reader, otherwise the peeked byte is lost. Hence the above approach results in a deserialisation error for legacy transactions because the signature count byte is consumed. It would be nice if the std library reader had a peek method even if it returned an unimplemented error when called with an underlying reader that does not support that functionality

src/core/transaction.zig
```diff
@@ -110,521 +49,372 @@ pub const V0Message = struct {
...
+    };
 
-pub const Transaction = struct {
-    signatures: []const Signature,
-    message: Message,
+    pub const Version = enum(u8) {
+        /// Legacy transaction without address lookups.
+        Legacy = 0xFF,
+        /// Transaction with address lookups.
+        V0 = 0x00,
```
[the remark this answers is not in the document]
yewman: f32f7eb3

src/core/transaction.zig
yewman: f32f7eb3 for bincode.testRoundTrip
Tests coming

src/core/transaction.zig
```diff
@@ -110,521 +49,372 @@ pub const V0Message = struct {
...
+            );
+            errdefer allocator.free(readonly_indexes);
+            try reader.readNoEof(readonly_indexes);
+
+            return .{
+                .table_address = table_address,
+                .writable_indexes = writable_indexes,
+                .readonly_indexes = readonly_indexes,
+            };
+        }
```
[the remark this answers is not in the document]
yewman: f32f7eb3

src/core/transaction.zig
```diff
@@ -110,521 +49,372 @@ pub const V0Message = struct {
...
+            );
+            errdefer allocator.free(data);
+            try reader.readNoEof(data);
+
+            return .{
+                .program_index = program_index,
+                .account_indexes = account_indexes,
+                .data = data,
+            };
+        }
```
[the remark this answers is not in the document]
yewman: f32f7eb3

src/core/transaction.zig
```diff
@@ -1,105 +1,44 @@
...
 const Hash = sig.core.Hash;
 const Pubkey = sig.core.Pubkey;
 const Signature = sig.core.Signature;
-const indexOf = sig.utils.slice.indexOf;
 
-const peekableReader = sig.utils.io.peekableReader;
-const shortVecConfig = sig.bincode.shortvec.sliceConfig;
+pub const Transaction = struct {
+    /// MAX_BYTES is the maximum size of a transaction.
+    pub const MAX_BYTES: u32 = 1232;
```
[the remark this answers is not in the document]
yewman: f32f7eb3

src/core/transaction.zig
yewman: f32f7eb3

src/core/transaction.zig
yewman: f32f7eb3

src/core/transaction.zig
yewman: f32f7eb3

src/core/transaction.zig
yewman: I'm fine with changing it to verify, do you have an opion here @dnut?

yewman: cc2b65e

src/core/transaction.zig
dnut: Oh ok I see the issue. It would only work if peekable reader is also being used for the next field after version, and it's not being used like that in my example. I think it's possible to achieve this, but it would require a custom deserialize method for Transaction, so the benefit is unclear.

src/core/transaction.zig
yewman: dd14bcb7 testing

src/core/transaction.zig
dnut: Fine with me. I think I like `validate` better than `verify` though, since verify has other meanings that are relevant, like verifying a signature.

src/core/transaction.zig
dnut: what about "Message sanitize fails if program id has index 0"

src/core/transaction.zig
dnut: something like this might be workable, but it fails the tests for some reason. up to you if you want to do anything with this

```zig
pub fn deserialize(allocator: std.mem.Allocator, reader: anytype, params: sig.bincode.Params) !Transaction {
    var peekable = sig.utils.io.peekableReader(reader);
    return try sig.bincode.readWithConfig(allocator, Transaction, peekable.reader(), params, .{});
}

fn deserializeVersion(
    _: std.mem.Allocator,
    /// This reader must be a peekable reader. Transaction's bincode config ensures this.
    reader: anytype,
    _: sig.bincode.Params,
) !TransactionVersion {
    return if (try reader.context.peekByte() & MESSAGE_VERSION_PREFIX != 0)
        switch (try reader.readByte() & ~MESSAGE_VERSION_PREFIX) {
            0 => .v0,
            127 => error.OffChainMessage,
            else => error.InvalidMessageTag,
        }
    else
        return .legacy;
}

```

src/core/transaction.zig
yewman: I'm probably with you on the use of `validate` instead. Thoughts @0xNineteen ?

0xNineteen: validates good

src/core/transaction.zig
yewman: 063439ea

src/core/transaction.zig
yewman: Yeah I think this is reasonable, I am curious where the error.OffChainMessage has come from in your suggested implementation?

dnut: > where the error.OffChainMessage has come from

It's from the code on main:
```zig
fn deserialize(
    allocator: std.mem.Allocator,
    original_reader: anytype,
    params: sig.bincode.Params,
) !VersionedMessage {
    var peekable = peekableReader(original_reader);
    const reader = peekable.reader();

    if (try peekable.peekByte() & MESSAGE_VERSION_PREFIX != 0) {
        return switch (try reader.readByte() & ~MESSAGE_VERSION_PREFIX) {
            0 => .{ .v0 = try sig.bincode.read(allocator, V0Message, reader, params) },
            127 => error.OffChainMessage,
            else => error.InvalidMessageTag,
        };
    } else {
        return .{ .legacy = try sig.bincode.read(allocator, Message, reader, params) };
    }
}
```

When I wrote that, I was copying this from agave:

```rust
MessagePrefix::Versioned(version) => {
    match version {
        0 => {
            Ok(VersionedMessage::V0(seq.next_element()?.ok_or_else(
                || {
                    // will never happen since tuple length is always 2
                    de::Error::invalid_length(1, &self)
                },
            )?))
        }
        127 => {
            // 0xff is used as the first byte of the off-chain messages
            // which corresponds to version 127 of the versioned messages.
            // This explicit check is added to prevent the usage of version 127
            // in the runtime as a valid transaction.
            Err(de::Error::custom("off-chain messages are not accepted"))
        }
        _ => Err(de::Error::invalid_value(
            de::Unexpected::Unsigned(version as u64),
            &"a valid transaction message version",
        )),
    }
}
```

src/transaction_sender/mock_transfer_generator.zig
```diff
@@ -375,4 +375,62 @@ pub const MockTransferService = struct {
...
+        const from_pubkey = Pubkey.fromPublicKey(&from_keypair.public_key);
+
+        const signatures = try allocator.alloc(Signature, 1);
+        errdefer allocator.free(signatures);
+
+        const addresses = try allocator.alloc(Pubkey, 3);
+        errdefer allocator.free(addresses);
+        addresses[0] = from_pubkey;
+        addresses[1] = to_pubkey;
+        addresses[2] = try Pubkey.fromString("11111111111111111111111111111111");
```
dadepo, line 396: Nit: Is it possible to comment what the `addresses[2]` is for? the other two is clear based on the variables `from_pubkey` and `to_pubkey` used to set their values.

src/transaction_sender/mock_transfer_generator.zig
```diff
@@ -375,4 +375,62 @@ pub const MockTransferService = struct {
...
+        allocator: std.mem.Allocator,
+        random: std.Random,
+        from_keypair: KeyPair,
+        to_pubkey: Pubkey,
+        lamports: u64,
+        recent_blockhash: Hash,
+    ) !sig.core.Transaction {
+        const from_pubkey = Pubkey.fromPublicKey(&from_keypair.public_key);
+
+        const signatures = try allocator.alloc(Signature, 1);
```
dadepo, line 389: More of a question: Does this needs to be dynamically allocated? given it is passed on to `sig.core.Transaction` which expects it as an array? Also won't there be an issue given it is being deinited when this function exists.

I guess same question applies to `instructions`, `addresses` etc

src/transaction_sender/mock_transfer_generator.zig
```diff
@@ -375,4 +375,62 @@ pub const MockTransferService = struct {
...
+        instructions[0] = .{
+            .program_index = 2,
+            .account_indexes = account_indexes,
+            .data = try allocator.dupe(u8, &data),
+        };
+
+        const transaction = sig.core.Transaction{
+            .signatures = signatures,
+            .version = .legacy,
+            .signature_count = 1,
```
dadepo, line 420: Nit:

```suggestion
            .signature_count = signatures.len,
```

src/crypto/base58.zig
```diff
@@ -16,7 +16,7 @@ pub fn Base58Sized(decoded_size: usize) type {
 
         pub fn decode(str: []const u8) ![decoded_size]u8 {
             var result_data: [decoded_size]u8 = undefined;
-            @setEvalBranchQuota(6100);
+            @setEvalBranchQuota(decoded_size * 145);
```
dadepo, line 19: More of a question: Was this diff intended to be part of this PR (it does not like it related to transactions refactor) and what is the reason/benefit from going to a fix value to one that depends on`decoded_size`. Also any reason for the `145`?

src/core/transaction.zig
```diff
@@ -110,749 +36,517 @@ pub const V0Message = struct {
...
+    ///   1) `account_keys`
+    ///   2) ordered list of account_keys loaded from `writable` lookup table indexes
+    ///   3) ordered list of account_keys loaded from `readable` lookup table indexes
+    instructions: []const TransactionInstruction,
 
-    pub fn sanitize(_: V0Message) !void {
-        // TODO
-        std.debug.print("V0Message.sanitize not implemented", .{});
-    }
+    /// `AddressLookup`'s are used to load account account addresses from lookup tables.
```
dadepo, line 49: Looks like a typo?

```suggestion
    /// `AddressLookup`'s are used to load account addresses from lookup tables.
```

src/core/transaction.zig
```diff
@@ -110,749 +36,517 @@ pub const V0Message = struct {
...
         allocator.free(self.data);
     }
 };
 
-pub const AccountMeta = struct {
-    pubkey: Pubkey,
-    is_signer: bool,
-    is_writable: bool,
+pub const TransactionAddressLookup = struct {
+    /// Adadress of the lookup table
```
dadepo, line 291:
```suggestion
    /// Address of the lookup table
```

src/core/transaction.zig
```diff
@@ -110,749 +36,517 @@ pub const V0Message = struct {
...
+    pub fn clone(self: *const TransactionAddressLookup, allocator: std.mem.Allocator) !TransactionAddressLookup {
+        const writable_indexes = try allocator.dupe(u8, self.writable_indexes);
+        errdefer allocator.free(writable_indexes);
         return .{
-            .pubkey = pubkey,
-            .is_signer = is_signer,
-            .is_writable = false,
+            .table_address = self.table_address,
+            .writable_indexes = writable_indexes,
+            .readonly_indexes = try allocator.dupe(u8, self.readonly_indexes),
```
dadepo, line 307: Any reason why this does not protected by `errdefer` but `writable_indexes` is?

yewman: since it is the last allocation, if it fails there is no need for the errdefer

0xNineteen (dismissed):
nice :shipit:

src/transaction_sender/mock_transfer_generator.zig
yewman: Transactions `own` their member slices and free them when they are deinitialized, if we didn't dynamically allocate the transtion deinit would fail

src/crypto/base58.zig
yewman: @Rexicon226 can you please the above, I think you will do a better job than me 😂

src/core/transaction.zig
yewman: dd73ede4

src/core/transaction.zig
yewman: dd73ede4

src/transaction_sender/mock_transfer_generator.zig
yewman: dd73ede4

src/transaction_sender/mock_transfer_generator.zig
yewman: dd73ede4

src/crypto/base58.zig
Rexicon226: This code doesn't exist anymore on master, you really should rebase!

src/core/transaction.zig
yewman: I have extracted the signable component into a TransactionMessage struct however relying on bincode for the TransactionMessage doesn't work (at least from my understanding) since we need conditional serialisation/deserialisation based on the version field. Otherwise we end up with a trailing byte when serialising legacy transactions due to the empty address lookups, and will error when deserialising legacy transactions since we will be expecting a zero byte to encode the empty address lookups. Not sure if there is a clean way around this in bincode so I went with the custom ser/der methods for Transaction, TransactionVersion and TransactionMessage. At least it is all still compatible with bincode ser/der interface so is probably fine

0xNineteen (dismissed):
:shipit:

0xNineteen (approved):

PR #518, opened 2025-01-24
feat(svm): system-program and related context

yewman:
# Overview
The primary objective of this PR is to complete the implementation of the System Program, achieving this goal will require the the full and partial implementation of a number of other data structures and system variables. These additional implementations will be completed only to the extent required to facilitate the correct execution of the System Program.

Since this is a large PR, it is being implemented in relative 'isolation' from other Sig components which should make it easier to review. For example, rather than moving 'Rent' sysvar from its definition in accounts db it is simply redefined in the sysvar module of the runtime. Refactoring will be completed in a follow-up PR to simplify the review process. 

README.md files are used to document and  track the progress of program, sysvar, and runtime context implementations with the long term goal of becoming useful references for understanding the SVM specification. 

**Note**: For detailed documentation of changes, please see the relevant README.md (hopefully it will be up to date 😝)

**Note on Naming**: I have tried to choose sensible names although I think there is probably room for improvement so am keen to hear any suggestions if something comes across as confusing etc.


src/core/instruction.zig
```diff
@@ -0,0 +1,207 @@
...
+pub const AccountMeta = struct {
+    /// An account's public key
+    pubkey: Pubkey,
+    /// True if account must sign the transaction
+    is_signer: bool,
+    /// True if the account is mutable
+    is_writable: bool,
+};
+
+pub const InstructionError = error{
```
0xNineteen, line 28: i think this is already defined in accountsdb/snapshots.zig ?

src/core/instruction.zig
```diff
@@ -0,0 +1,207 @@
...
+    /// Program address
+    program_id: Pubkey,
+    /// Accounts that the command references
+    accounts: []const AccountMeta,
+    /// Data is the binary encoding of the program instruction and its
+    /// arguments. The lifetime of the data must outlive the instruction.
+    data: []const u8,
+};
+
+pub const AccountMeta = struct {
```
0xNineteen, line 19:
```suggestion
pub const InstructionAccount = struct {
```

src/runtime/program/lib.zig
```diff
@@ -0,0 +1,4 @@
+pub const system_program_execute = @import("system_program_execute.zig");
+pub const system_program = @import("system_program.zig");
```
dadepo, line 2: What about having a directory for each program? 

I feel having everything in the top level `/program` may be too much making it harder to locate the files for a particular program, 

but `program/system`, `program/stake`, `program/vote` etc would make for better organization.

src/runtime/program/README.md
```diff
@@ -0,0 +1,123 @@
...
+## Loader V4 Program
+- [ ] loader_v4_program
+- [ ] loader_v4_program_execute
+
+## Precompile Programs
+- [ ] ed25519Verify
+- [ ] secp256k1Verify
+- [ ] secp256r1Verify
+
+## Stake Program
```
dadepo, line 60: Stake Program already listed above in line 43.

src/runtime/program/system_program.zig
dadepo, line 1: This file contains the instructions and not the actual system program implementation? Maybe if the suggestion about having a folder for each program is accepted (see https://github.com/Syndica/sig/pull/518/files#r1946992532) then this file can be renamed to `program/system/instruction.zig`?

src/runtime/program/lib.zig
yewman: I am happy with that, I just know in past we have favoured keeping the directory structure as flat as possible. It might be worth @0xNineteen and @dnut weighing in on this one..

src/runtime/program/system_program.zig
yewman: Agreed, if we go with the above approach I would also want to pull InstructionError into its own file as well. Almost mirroring the agave program interface implementations for example https://github.com/solana-program/system/tree/main/interface/src

src/runtime/program/README.md
yewman: 11063ffe
I just removed the extras while I was at it because I think they are futile rn

src/runtime/program/lib.zig
Rexicon226: How common is it that a program will require multiple files?

dadepo: The ones already in this PR are two files. Going through the `programs` in Agave there are a couple containing >=2 files.

At the end of the day it comes down to the file organization that fits well with Sig's structure. Because instead of the current two files in this PR, the whole implementation of a program can actually go into one file - (and I think that might even makes for better org, instead of having programs with different files with all of them sprawled within the `/program` directory)

dadepo: Firedancer uses a single file per program structure. That can also serve as another inspiration.

Rexicon226: > The ones already in this PR are two files. 

Looking through it, I see no reason for it to be in two files.

I would like us to try to keep each program as a single file, and if necessary, create a folder with the same name as the program which contains the auxiliary files.

src/runtime/sysvar/clock.zig
```diff
@@ -0,0 +1,31 @@
+const sig = @import("../../sig.zig");
+
+const Pubkey = sig.core.Pubkey;
+const Slot = sig.core.Slot;
+const Epoch = sig.core.Epoch;
+
+/// [agave] https://github.com/anza-xyz/agave/blob/8db563d3bba4d03edf0eb2737fba87f394c32b64/sdk/clock/src/lib.rs#L184
+pub const Clock = extern struct {
```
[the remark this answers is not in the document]
Rexicon226: It shouldn't be an `extern`, we don't rely on the C ABI anywhere here.

src/runtime/sysvar_cache.zig
```diff
@@ -0,0 +1,37 @@
...
+    maybe_last_restart_slot: ?sysvar.LastRestartSlot = null,
+    maybe_rent: ?sysvar.Rent = null,
+    maybe_slot_hashes: ?sysvar.SlotHashes = null,
+    maybe_stake_history: ?sysvar.StakeHistory = null,
+
+    // deprecated sysvars, these should be removed once practical
+    maybe_fees: ?sysvar.Fees = null,
+    maybe_recent_blockhashes: ?sysvar.RecentBlockhashes = null,
+
+    pub fn get(self: SysvarCache, comptime T: type) ?T {
```
Rexicon226, line 23: A couple of points on this struct. What do you think about not prepending each field with `maybe`, since it somewhat loses its value if they're all optional? And if you were to do that, you could rework `get` to look something like:
```zig
pub fn get(self: SysvarCache, comptime field: @TypeOf(.EnumLiteral) ?T {
    return @field(self, @tagName(field));
}
```
and it could be accessed like:
```zig
.get(.clock);
```

src/runtime/program/system_program_execute.zig
```diff
@@ -0,0 +1,1629 @@
...
+fn executeAdvanceNonceAccount(
+    allocator: std.mem.Allocator,
+    ic: *InstructionContext,
+) !void {
+    try ic.checkNumberOfAccounts(1);
+    var account = try ic.borrowInstructionAccount(0);
+    defer account.release();
+
+    const recent_blockhashes = try ic.getSysvarWithAccountCheck(RecentBlockhashes, 1);
+    if (recent_blockhashes.isEmpty()) {
```
dadepo, line 202: I have a question about the `execute<TheMethod>` and `theMethod` pattern. Which is how is it determined what logic goes in `execute<TheMethod>` and which ones to have in `theMethod`?

For example, getting all the sysvar looks like something to go in the `execute<TheMethod>`. But what about checking for constraint that could lead to error? For example here, the `if (recent_blockhashes.isEmpty())` leads to `InstructionError.Custom` but within `initializeNonceAccount` there are more error conditions.

Would it be an idea to restricting the code that goes into  `execute<TheMethod>` as the ones needed to retrieve the dependencies from the context, while the rest of the logic including checking constraints are only in `theMethod`?

Another idea is to maybe drop `execute<TheMethod>` and have everything in one method?

src/runtime/program/system_program_execute.zig
```diff
@@ -0,0 +1,1629 @@
...
+            .{account.getPubkey()},
+        );
+        return InstructionError.InvalidArgument;
+    }
+
+    const versioned_nonce = try account.getState(allocator, nonce.Versions);
+    switch (versioned_nonce.getState()) {
+        .unintialized => {
+            const min_balance = rent.minimumBalance(account.getData().len);
+            if (min_balance > account.getLamports()) {
```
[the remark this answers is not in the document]
yewman: We could, although in this instance it would just mean that we need to calculate min balance twice for no real benefit imo

src/runtime/sysvar_cache.zig
yewman: 0baf390
I've removed the maybe_'s. If I was to modify the get method as you suggest how would you approach the following method and it's call sites? https://github.com/Syndica/sig/blob/0baf390825b75e48733a5410c48f9690ee5e87df/src/runtime/instruction_context.zig#L112

src/runtime/borrowed_account.zig
```diff
@@ -0,0 +1,194 @@
...
+        allocator: std.mem.Allocator,
+        comptime T: type,
+    ) error{InvalidAccountData}!T {
+        return bincode.readFromSlice(allocator, T, self.account.data, .{}) catch {
+            return InstructionError.InvalidAccountData;
+        };
+    }
+
+    /// [agave] https://github.com/anza-xyz/agave/blob/faea52f338df8521864ab7ce97b120b2abb5ce13/sdk/src/transaction_context.rs#L976
+    pub fn setState(
```
0xNineteen, line 123: `serializeIntoAccountData` ?

src/runtime/borrowed_account.zig
```diff
@@ -0,0 +1,194 @@
...
+        self: *BorrowedAccount,
+        state: anytype,
```
0xNineteen, line 125: would be good to document any state needs to support `serializedSize()` and `bincode`

src/runtime/borrowed_account.zig
```diff
@@ -0,0 +1,194 @@
...
+    pub fn setDataLength(
+        self: *BorrowedAccount,
+        allocator: std.mem.Allocator,
+        tc: *TransactionContext,
+        length: usize,
+    ) InstructionError!void {
+        try self.checkCanSetDataLength(tc, length);
+        try self.checkDataIsMutable();
+        if (self.getData().len == length) return;
+        tc.accounts_resize_delta +|= @intCast(length -| self.getData().len);
```
0xNineteen, line 150: dont you need to cast length to i64 before doing this operation? (same with data.len)

src/runtime/borrowed_account.zig
```diff
@@ -0,0 +1,194 @@
...
+        if (written.len != serialized_size)
+            return InstructionError.GenericError;
+    }
+
+    /// [agave] https://github.com/anza-xyz/agave/blob/faea52f338df8521864ab7ce97b120b2abb5ce13/sdk/src/transaction_context.rs#L885
+    pub fn setDataLength(
+        self: *BorrowedAccount,
+        allocator: std.mem.Allocator,
+        tc: *TransactionContext,
+        length: usize,
```
0xNineteen, line 145: new_length ?

src/runtime/feature_set.zig
```diff
@@ -0,0 +1,15 @@
...
+const Pubkey = sig.core.Pubkey;
+
+/// [agave] https://github.com/anza-xyz/agave/blob/8db563d3bba4d03edf0eb2737fba87f394c32b64/sdk/feature-set/src/lib.rs#L1188
+pub const FeatureSet = struct {
+    active: std.AutoArrayHashMapUnmanaged(Pubkey, u64),
+    inactive: std.AutoArrayHashMapUnmanaged(Pubkey, void),
+
+    pub const EMPTY = FeatureSet{
+        .active = std.AutoArrayHashMapUnmanaged(Pubkey, u64){},
+        .inactive = std.AutoArrayHashMapUnmanaged(Pubkey, void){},
```
0xNineteen, line 13:
```suggestion
        .active = .{},
        .inactive = .{},
```

src/runtime/nonce.zig
```diff
@@ -0,0 +1,74 @@
...
+        lamports_per_signature: u64,
+    ) Data {
+        return .{
+            .authority = authority,
+            .durable_nonce = durable_nonce,
+            .fee_calculator = .{ .lamports_per_signature = lamports_per_signature },
+        };
+    }
+
+    pub fn getDurableNonce(self: Data) Hash {
```
0xNineteen, line 61: do we need these? would pref not to have getters

src/runtime/transaction_context.zig
```diff
@@ -0,0 +1,117 @@
...
+
+    /// If an error other than an InstructionError occurs during execution its value will
+    /// be set here and InstructionError.custom will be returned
+    custom_error: ?u32,
+
+    /// Optional log collector
+    log_collector: ?LogCollector,
+
+    // TODO: the following feilds should live above the transaction level, however, they are
+    // defined here temporarily for convenience.
```
0xNineteen, line 43: would be good to create a future ticket on this -- probs need a new category 'Runtime' too

0xNineteen (changes requested):
few things

src/runtime/program/lib.zig
yewman: 94b2097b

src/runtime/program/system_program.zig
yewman: 94b2097b

src/runtime/tmp_utils.zig
```diff
@@ -0,0 +1,13 @@
+// Utils required for the runtime.
+// Some of these utils are present elswhere in random places, for now to keep the pr review simple
+// they are redefined here. This will be addressed in a follow up pr.
```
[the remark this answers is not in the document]
yewman: 94b2097b

src/runtime/account_shared_data.zig
```diff
@@ -0,0 +1,31 @@
+const std = @import("std");
+const sig = @import("../sig.zig");
+
+const Pubkey = sig.core.Pubkey;
+const Epoch = sig.core.Epoch;
+
+/// [agave] https://github.com/anza-xyz/agave/blob/faea52f338df8521864ab7ce97b120b2abb5ce13/sdk/src/account.rs#L118
+pub const AccountSharedData = struct {
```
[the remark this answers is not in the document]
yewman: 94b2097b

src/runtime/instruction_context.zig
```diff
@@ -0,0 +1,125 @@
...
+        self: *const InstructionContext,
+        index: anytype,
+    ) error{NotEnoughAccountKeys}!bool {
+        if (index >= self.accounts.len) return InstructionError.NotEnoughAccountKeys;
+        return self.accounts[index].is_signer;
+    }
+
+    /// Replaces Agave's approach to checking if a pubkey is a signer which is to precompute a
+    /// hashmap of signers to parse during instruction execution
+    pub fn isPubkeySigner(self: *const InstructionContext, pubkey: anytype) bool {
```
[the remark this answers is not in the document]
yewman: 94b2097b

src/runtime/borrowed_account.zig
```diff
@@ -0,0 +1,194 @@
...
+    borrow_context: struct {
+        program_id: Pubkey,
+        is_writable: bool,
+    },
+
+    pub fn release(self: *BorrowedAccount) void {
+        self.account_write_guard.release();
+    }
+
+    pub fn getPubkey(self: BorrowedAccount) Pubkey {
```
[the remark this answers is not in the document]
yewman: 94b2097b

src/runtime/borrowed_account.zig
```diff
@@ -0,0 +1,194 @@
...
+        const resize_delta: i64 = @intCast(length -| old_length);
+        const new_accounts_resize_delta = etc.accounts_resize_delta +| resize_delta;
+
+        if (new_accounts_resize_delta > MAX_PERMITTED_ACCOUNTS_DATA_ALLOCATIONS_PER_TRANSACTION)
+            return InstructionError.MaxAccountsDataAllocationsExceeded;
+    }
+
+    /// Deserialize the account data into a type `T`.
+    /// [agave] https://github.com/anza-xyz/agave/blob/faea52f338df8521864ab7ce97b120b2abb5ce13/sdk/src/transaction_context.rs#L968
+    pub fn getState(
```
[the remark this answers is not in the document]
yewman: 94b2097b

src/runtime/borrowed_account.zig
yewman: 94b2097b

src/runtime/borrowed_account.zig
yewman: 94b2097b

src/runtime/borrowed_account.zig
yewman: 94b2097b

src/runtime/feature_set.zig
yewman: 94b2097b

src/runtime/transaction_context.zig
yewman: 94b2097b

src/runtime/borrowed_account.zig
yewman: be1acfda

src/runtime/borrowed_account.zig
```diff
@@ -0,0 +1,194 @@
...
+    pub fn getPubkey(self: BorrowedAccount) Pubkey {
+        return self.pubkey;
+    }
+
+    pub fn getLamports(self: BorrowedAccount) u64 {
+        return self.account.lamports;
+    }
+
+    pub fn getData(self: BorrowedAccount) []const u8 {
+        return self.account.data;
```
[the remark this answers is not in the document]
yewman: As in remove any getters which do not perform checks at present? For the case of the data I think it probably makes sense to keep a getter which returns a `[]const u8`, there will be `getDataMutable` which returns a `[]u8` and performs appropriate checks

src/core/instruction.zig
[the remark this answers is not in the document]
yewman: 721ba09d

src/core/instruction.zig
```diff
@@ -0,0 +1,207 @@
...
+    /// An account's public key
+    pubkey: Pubkey,
+    /// True if account must sign the transaction
+    is_signer: bool,
+    /// True if the account is mutable
+    is_writable: bool,
+};
+
+pub const InstructionError = error{
+    /// Deprecated! Use CustomError instead!
```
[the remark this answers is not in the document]
yewman: It is actually defined in both `src/ledger/transaction_status.zig` and `src/accountsdb/snapshots.zig`. In both the above it is defined as an union(enum) which I presume is relevant for ser/der. In the runtime it is beneficial to keep the InstructionError as an zig error set to allow `try` semantics. I'll have look into the usage in the above scenarios and see if I can come up with a good solution

src/runtime/program/system_program_execute.zig
yewman: I would just be wary of altering behaviour on error cases, for example if there are two points of failure A and B which return errorA and errorB respectively, then if we change the order of execution relative to agave we will emit different transaction errors when receiving input which would fail both A and B. Otherwise should be fine :)

src/runtime/account_shared_data.zig
```diff
@@ -0,0 +1,40 @@
+const std = @import("std");
+const sig = @import("../sig.zig");
+
+const Pubkey = sig.core.Pubkey;
+const Epoch = sig.core.Epoch;
+
+/// `AccountSharedData` holds account information with a shared reference to the account data field
+/// `AccountSharedData`'s are loaded from `accounts_db` during the transaction loading phase
+///
+/// TODO: move to `accounts_db`?
```
dadepo, line 10: ping: Move? or keep where it is? If keep, then maybe this TODO can be removed?

src/runtime/account_shared_data.zig
```diff
@@ -0,0 +1,40 @@
...
+    rent_epoch: Epoch,
+
+    pub fn isZeroed(self: AccountSharedData) bool {
+        // TODO: naive implementation
+        for (self.data) |byte| if (byte != 0) return false;
+        return true;
+    }
+
+    pub fn resize(self: *AccountSharedData, allocator: std.mem.Allocator, new_size: usize) !void {
+        // TODO: naive implementation
```
dadepo, line 32: Ping: Want to improve the implementation in this PR? if not, maybe have an issue for them? As I think these are things that should still be done.

src/runtime/account_shared_data.zig
```diff
@@ -0,0 +1,40 @@
...
+        const new_memory = try allocator.alloc(u8, new_size);
+        @memset(new_memory, 0);
```
dadepo, line 34: Any reason to not use `allocator.resize`?

src/runtime/borrowed_account.zig
```diff
@@ -0,0 +1,201 @@
...
+const Epoch = sig.core.Epoch;
+const InstructionError = sig.core.instruction.InstructionError;
+const Pubkey = sig.core.Pubkey;
+
+const AccountSharedData = sig.runtime.AccountSharedData;
+const TransactionContext = sig.runtime.TransactionContext;
+const WLockGuard = sig.runtime.TransactionAccount.WLockGuard;
+
+const MAX_PERMITTED_ACCOUNTS_DATA_ALLOCATIONS_PER_TRANSACTION =
+    sig.runtime.program.system_program.MAX_PERMITTED_ACCOUNTS_DATA_ALLOCATIONS_PER_TRANSACTION;
```
dadepo, line 15: Nit: Maybe have the directory named as only `system`? ie drop the `_program` suffix, given everything is already within `system` directory.

src/runtime/sysvar/epoch_schedule.zig
```diff
@@ -0,0 +1,29 @@
...
+
+    /// The first epoch after the warmup period.
+    ///
+    /// Basically: `log2(slots_per_epoch) - log2(MINIMUM_SLOTS_PER_EPOCH)`.
+    first_normal_epoch: u64,
+
+    /// The first slot after the warmup period.
+    ///
+    /// Basically: `MINIMUM_SLOTS_PER_EPOCH * (2.pow(first_normal_epoch) - 1)`.
+    first_normal_slot: u64,
```
Sobeston, line 25: Could use slot type in this strict

src/runtime/sysvar/last_restart_slot.zig
```diff
@@ -0,0 +1,12 @@
+const sig = @import("../../sig.zig");
+
+const Pubkey = sig.core.Pubkey;
+
+/// [agave] https://github.com/anza-xyz/agave/blob/8db563d3bba4d03edf0eb2737fba87f394c32b64/sdk/last-restart-slot/src/lib.rs#L15
+pub const LastRestartSlot = struct {
+    /// The last restart `Slot`.
+    last_restart_slot: u64,
```
Sobeston, line 8: This struct too

Sobeston (commented):
Looks great, just a couple nits

src/runtime/sysvar/rent.zig
```diff
@@ -0,0 +1,50 @@
...
+    /// The percentage of collected rent that is burned.
+    ///
+    /// Valid values are in the range [0, 100]. The remaining percentage is
+    /// distributed to validators.
+    burn_percent: u8,
+
+    pub const ID =
+        Pubkey.parseBase58String("SysvarRent111111111111111111111111111111111") catch unreachable;
+
+    pub fn default() Rent {
```
dadepo, line 35: Is the convention `default()` or `DEFAULT`?

src/core/instruction.zig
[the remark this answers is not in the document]
0xNineteen: ah good point - didnt realize - maybe just add a note in the comments of both structs mentioning how there are duplicates but one is an enum and one is an error type - maybe remove the duplicate in ledger/accountsdb too

src/runtime/instruction_context.zig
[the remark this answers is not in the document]
0xNineteen: theres another one right above : 

```
    pub fn isIndexSigner(
        self: *const InstructionContext,
        index: anytype, // should be u64 ?
    ) error{NotEnoughAccountKeys}!bool {
```

src/runtime/borrowed_account.zig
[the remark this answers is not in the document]
0xNineteen: ok - we'll still need checks on the `getData` to make sure we dont race on the data, right? - im guessing that will come in a future PR? any reason we dont just add it in this PR?

src/runtime/borrowed_account.zig
0xNineteen: documentation would need to be in setState doc comment -- since its parameter state can be anytype right? - or could we just change the `anytype` to `nonce.Versions` ? not sure why we have so many anytypes lol

src/runtime/sysvar/rent.zig
0xNineteen: should be `const DEFAULT`

src/runtime/instruction_context.zig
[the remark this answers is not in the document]
yewman: 021a5e8f
ahh I had an isSigner method but since they do slightly different things I split it into separate methods and forgot to change the arg type!

src/runtime/borrowed_account.zig
yewman: anytype here because this method is not unique to storing `nonce.Versions`. Is there a better way to acheive this without the use of `anytype`?

src/runtime/sysvar/rent.zig
yewman: ce9fd270

src/runtime/sysvar/last_restart_slot.zig
yewman: ce9fd270

src/runtime/sysvar/epoch_schedule.zig
yewman: ce9fd270

src/runtime/account_shared_data.zig
yewman: I think I'd rather leave it here until we implement account loading, I have added a more descriptive comment dedbf586

src/runtime/program/system_program/execute.zig
```diff
@@ -0,0 +1,1509 @@
...
+    try checkSeedAddress(
+        ic,
+        account.pubkey,
+        base,
+        owner,
+        seed,
+        "Create: address {} does not match derived address {}",
+    );
+    try allocate(allocator, ic, &account, space, base);
+    try assign(ic, &account, owner, base);
```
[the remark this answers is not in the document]
yewman: But in this instance the `base` of the derived address is the `authority` for the respective calls

src/runtime/borrowed_account.zig
0xNineteen: ah makes sense, i guess were just not using it anywhere yet rn

src/runtime/account_shared_data.zig
yewman: 1996fabf

src/runtime/account_shared_data.zig
yewman: 1996fabf

Rexicon226: You should still use a resize solution:
```zig
if (allocator.resize(data, new_size)) {
    const old_len = data.len;
    data.len = new_size;
    @memset(data[@min(old_len, data.len)..], 0);
} else {
    const new_memory = try allocator.alloc(u8, new_size);
    @memset(new_memory, 0);
    @memcpy(new_memory.ptr, data[0..@min(data.len, new_size)]);
    allocator.free(data);
    data = new_memory;
}
```

src/runtime/borrowed_account.zig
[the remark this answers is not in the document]
yewman: 78b0c12e
I don't think data races are an issue since transaction execution is synchronous

0xNineteen (dismissed):
nice :shipit:

src/runtime/account_shared_data.zig
0xNineteen: i think this may be the wrong commit

yewman: ahh yes it is 😂 , 1996fabf

src/runtime/account_shared_data.zig
yewman: 10b67796

Sobeston (approved):
:shipit: nice

src/runtime/nonce.zig
```diff
@@ -0,0 +1,67 @@
...
+
+/// [agave] https://github.com/anza-xyz/agave/blob/faea52f338df8521864ab7ce97b120b2abb5ce13/sdk/program/src/nonce/state/current.rs#L10-L11
+const DURABLE_NONCE_HASH_PREFIX = "DURABLE_NONCE";
+
+/// Current variants have durable nonce and blockhash domains separated.\
+///
+/// Must support `bincode` and `serializedSize` methods for writing to the account data.\
+///
+/// [agave] https://github.com/anza-xyz/agave/blob/faea52f338df8521864ab7ce97b120b2abb5ce13/sdk/program/src/nonce/state/mod.rs#L12
+pub const Versions = union(enum) {
```
dadepo, line 16: Agave's implementation contains a couple more methods not implemented here? Is the idea to implement those later? if so maybe a TODO would be nice

0xNineteen (approved):
had to fix the docs for CI so pushed some mods

docs/check.py
```diff
@@ -19,13 +19,23 @@
     for src_path, docs_path in g.get_markdown_files(args.src_dir, exclude_dirs, doc_dir_path):
         # check to see if the files are the same !
         with open(src_path, "r") as src_f:
-            with open(docs_path, "r") as docs_f:
-                src_lines = src_f.readlines()
-                # fix image paths for docusaurus
-                for i in range(0, len(src_lines)):
-                    if "/docs/docusaurus/static/img" in src_lines[i]:
-                        src_lines[i] = src_lines[i].replace("/docs/docusaurus/static/img", "/img")
+            src_lines = src_f.readlines()
```
dadepo, line 22: Nit: Is this diff meant to be in this PR? looks like result of a bad merge/rebase?

docs/generate.py
```diff
@@ -61,9 +61,20 @@ def get_markdown_files(
     for src_path, docs_path in get_markdown_files(src_path, exclude_dirs, code_docs_path):
         # copy the file to the docs/code directory
         with open(src_path, "r") as f:
+            src_lines = f.readlines()
```
dadepo, line 64: Same here. diff intended for this pr?

docs/readme.md
```diff
@@ -38,3 +38,7 @@ to ensure that the markdown is formatted correctly, we use the following
 conventions:
 - '#' for headers which will correspond to the title and sidebar name in docusaurus
 - all images need to point to /docs/docusaurus/static/img/
+
```
dadepo, line 41: Same here. diff intended for this pr?

docs/docusaurus/docs/code/runtime.md
dadepo, line 1: This looks like a duplicate of src/runtime/README.md

src/runtime/program/test_program_execute.zig
```diff
@@ -0,0 +1,211 @@
...
+            allocator.free(account.account.data);
+        allocator.free(expected_transaction_context.accounts);
+    }
+
+    try expectTransactionContextEqual(expected_transaction_context, transaction_context);
+}
+
+pub fn expectTransactionAccountEqual(
+    expected: TransactionAccount,
+    actual: TransactionAccount,
```
dadepo, line 152: Minor feature request. Is it possible to pass the index through to this function so the index of account that failed the equal check can be logged? Ran into mismatch while setting up test for the vote program and it was not quickly obvious which of the accounts are not equal.

dadepo (commented):
I have worked through half of the changed files. Will prioritize going through the rest tomorrow

dadepo (approved):
LGTM! Further improvements can be applied in follow up PRs

PR #520, opened 2025-01-27
refactor(echo,prometheus,build): Remove our dependency on the httpz library

InKryption:
There are still some definite improvements/fixes to make to prometheus, like making it possible to tell it to exit gracefully.

Note: unlike in the echo case, I decided it probably wasn't worth the complexity of dispatching to a threadpool in the prometheus case - not actually even sure if it's worth it in the echo server case either, but I'm unaware of the particular requirements there, so I retained that behavior from httpz.

Also took the liberty of refactoring adjacent code.

There's a couple TODOs I'll resolve once the RPC server PR is also merged.

### DIFF
--- a/src/net/echo.zig
+++ b/src/net/echo.zig
@@ -23,8 +23,6 @@ pub fn getShredAndIPFromEchoServer(
     for (socket_addresses) |socket_addr| {
         const response = requestIpEcho(socket_addr.toAddress(), .{}) catch continue;
         my_ip = my_ip orelse response.address;
-        if (my_ip == null)
-            my_ip = response.address;
 
         if (response.shred_version) |shred_version| {
             logger.info()
--- a/src/net/echo.zig
+++ b/src/net/echo.zig
@@ -237,7 +237,7 @@ const ConnectionTask = struct {
         const socket_addr = SocketAddr.fromIpV4Address(request.server.connection.address);
         const ip_echo_response: IpEchoResponse = .{
             .address = .{ .ipv4 = socket_addr.V4.ip },
-            .shred_version = .{ .value = 0 },
+            .shred_version = .{ .value = 0 }, // TODO: correct shred version needs to be propagated here
         };
 
         const content_len = blk: {
--- a/src/net/echo.zig
+++ b/src/net/echo.zig
@@ -147,19 +147,19 @@ pub const Server = struct {
                 _ = try std.posix.fcntl(conn.stream.handle, std.posix.F.SETFL, flags_int);
             }
 
-            const hct = try HandleConnectionTask.create(
+            const hct = try ConnectionTask.create(
                 self.allocator,
                 conn,
                 self.logger,
             );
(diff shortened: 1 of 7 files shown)


src/net/echo.zig
```diff
@@ -21,28 +15,27 @@ const HEADER_LENGTH: usize = 4;
...
-                logger.info()
-                    .field("shred_version", shred_version.value)
-                    .field("ip", socket_addr.toString().constSlice())
-                    .log("ip echo response");
-                my_shred_version = shred_version.value;
-            }
+        const response = requestIpEcho(socket_addr.toAddress(), .{}) catch continue;
+        my_ip = my_ip orelse response.address;
+        if (my_ip == null)
+            my_ip = response.address;
```
0xNineteen, line 27: isnt this if redundant?

src/net/echo.zig
```diff
@@ -51,204 +44,306 @@ pub fn getShredAndIPFromEchoServer(
...
+        self.wg.finish();
+        const allocator = self.allocator;
+        allocator.destroy(self);
+    }
+};
 
-    std.json.stringify(IpEchoServerResponse.init(IpAddr{ .ipv4 = socket_addr.V4.ip }), .{}, res.writer()) catch |err| {
-        // logger.err().logf("could not json stringify IpEchoServerResponse: {any}", .{err});
-        std.debug.print("could not json stringify ip echo server response message: {}\n", .{err});
+fn handleConnectionNoErr(hct: *HandleConnectionTask) void {
```
0xNineteen, line 206: would be good to have these in the struct

src/net/echo.zig
```diff
@@ -51,204 +44,306 @@ pub fn getShredAndIPFromEchoServer(
...
+        else => try httpRespondError(&request, .bad_request, ""),
+    };
+    const ip_echo_request = ip_echo_request_result.value;
+    _ = ip_echo_request;
+    ip_echo_request_result.deinit();
+
+    const socket_addr = SocketAddr.fromIpV4Address(request.server.connection.address);
+    const ip_echo_response: IpEchoResponse = .{
+        .address = .{ .ipv4 = socket_addr.V4.ip },
+        .shred_version = .{ .value = 0 },
```
0xNineteen, line 254: we should have a TODO or ticket that the correct shred version should be propagated here

0xNineteen (changes requested):
overall looks great

src/net/echo.zig
InKryption: Missed that whoops

src/net/echo.zig
InKryption: d5b2dcb854dacc5ce211d6251c05953c8ed79705

src/net/echo.zig
InKryption: e74ef3fb1ccbd0b50469af7798d3ebc02110e31f

0xNineteen (approved):
lgtm 🚢 :shipit:

PR #522, opened 2025-01-30
fix(adb/threadpool): remove waitgroups and fix segfault

0xNineteen:
the main fix is to remove the waitgroups in accounts-db loading/validating so we can have errors propagate correctly. while doing this, i also noticed a bug in the homo-threadpool impl where `addOne` to existing tasks will cause re-allocs on existing thread state leading to a segfault. there was also some code in repair which would `join` the threadpool after a single schedule in a loop, which i assumed was a bug and fixed as well.

there was also a bug which tsan found where an allocator was used in the registry without a lock which i fixed as well.

this lead to another bug in tsan about registry init race conditions which was fixed in 553c8039.

### What changed?
- Refactored `spawnThreadTasks` to use a homogeneous thread pool instead of raw thread spawning
- Removed single-threaded code paths in favor of unified multi-threaded implementation
- Modified `loadFromSnapshot` to return void instead of duration
- Adjusted testnet accounts per file estimate from 1000 to 500
- Added thread pool capacity pre-allocation to avoid reallocation during scheduling
- Added new test case for `spawnThreadTasks`

### DIFF
--- a/src/prometheus/registry.zig
+++ b/src/prometheus/registry.zig
@@ -241,7 +241,7 @@ pub fn Registry(comptime options: RegistryOptions) type {
             self.mutex.lock();
             defer self.mutex.unlock();
 
-            var allocator = self.arena_state.allocator();
+            const allocator = self.arena_state.allocator();
             const duped_name = try allocator.dupe(u8, name);
 
             const gop = try self.metrics.getOrPut(allocator, duped_name);
--- a/src/prometheus/registry.zig
+++ b/src/prometheus/registry.zig
@@ -238,12 +238,11 @@ pub fn Registry(comptime options: RegistryOptions) type {
             if (self.nbMetrics() >= options.max_metrics) return error.TooManyMetrics;
             if (name.len > options.max_name_len) return error.NameTooLong;
 
-            var allocator = self.arena_state.allocator();
-
-            const duped_name = try allocator.dupe(u8, name);
-
             self.mutex.lock();
             defer self.mutex.unlock();
+
+            var allocator = self.arena_state.allocator();
+            const duped_name = try allocator.dupe(u8, name);
 
             const gop = try self.metrics.getOrPut(allocator, duped_name);
             if (!gop.found_existing) {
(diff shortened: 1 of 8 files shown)


src/prometheus/registry.zig
```diff
@@ -238,13 +238,12 @@ pub fn Registry(comptime options: RegistryOptions) type {
...
 
-            var allocator = self.arena_state.allocator();
-
-            const duped_name = try allocator.dupe(u8, name);
-
             self.mutex.lock();
             defer self.mutex.unlock();
 
+            var allocator = self.arena_state.allocator();
+            const duped_name = try allocator.dupe(u8, name);
```
InKryption, line 245: Should either just be `self.arena_state.allocator().dupe(u8, name)`, or some variation of `const allocator = self.arena_state.allocator()` or `const arena = self.arena_state.allocator()`.

src/utils/thread.zig
```diff
@@ -217,32 +220,74 @@ pub fn HomogeneousThreadPool(comptime TaskType: type) type {
...
+            self.results = try std.ArrayList(TaskResult).initCapacity(self.allocator, self.tasks.capacity);
             self.tasks.clearRetainingCapacity();
             return results;
         }
 
         /// Like join, but it returns an error if any tasks failed, and otherwise discards task output.
         pub fn joinFallible(self: *Self) !void {
-            const results = self.join();
+            const results = try self.join();
             for (results.items) |result| try result;
```
InKryption, line 245: May be worth documenting that this short-circuits up to the first error.

InKryption (changes requested):
Two very minor comments, overall a solid improvement and fix.

src/utils/thread.zig
0xNineteen: [ea137f5](https://github.com/Syndica/sig/pull/522/commits/ea137f5d3ee3252bbe8fc2e625d1129e5c310da9)

src/prometheus/registry.zig
0xNineteen: [ea137f5](https://github.com/Syndica/sig/pull/522/commits/ea137f5d3ee3252bbe8fc2e625d1129e5c310da9)

InKryption (dismissed):

PR #532, opened 2025-02-04
feat(fuzz): fuzz with random multithreaded reads for accountsdb

Sobeston:
Note: I've rediscovered that accountsdb fuzzing is currently broken, and we don't run it on CI... However, this PR does not break it further.

I'll retarget this PR onto main once #476 is merged.


0xNineteen:
is it possible to target this to main now? do we need the buffer pool changes to be merged first?

Sobeston:
> is it possible to target this to main now? do we need the buffer pool changes to be merged first?

closed in favour of #552

PR #533, opened 2025-02-04
perf(channel): skip intermediary channels

kprotty:


ultd:
@kprotty please hold merge until we get some stats on packets/second performance increase/decrease by removing any work the callback/task does and running benchmark to effectively know if non-channel vs channel is beneficial. Additionally, any stats around memory utilization would be great.

kprotty:
Closing as it ends up having similar perf for `GossipService.verifyPackets`:

The `benchmark -- gossip` results are [biased](https://gist.github.com/kprotty/8e4f57e21e80992655e26d5823ee4266/revisions#diff-bc457943775cc87cf2a1ca507ae8465e222b7116f08e74705f2e1875808eab21) towards the no-channel approach as it starts processing packets while they're being built and submitted (instead of after with channel approach). Changing the latter to process them during shows the no-channel approach having [less volatile](https://gist.github.com/kprotty/8e4f57e21e80992655e26d5823ee4266/225264b58e85480e21014bb3732b2268d2f186a8) results but ultimately similar perf for a mixed stream of packet types.

When simulating max packet load, and only recording `verifyPacket` calls instead of actual work, the no-channel approach has worse throughput by default. It's bottlenecked on the allocator, where using [block-based allocation](https://gist.github.com/kprotty/8e4f57e21e80992655e26d5823ee4266#file-verify_block_allock_service-zig-L483-L539) (`rep-chan-alloc`) brings it to slightly better throughput than the channel approach. `no-pool` is provided as the baseline/max with no channel/ThreadPool scheduling at all.

```
(no verification/work):
main:            ~280,300 pps
replace-channel: ~104,000 pps
rep-chan-alloc:  ~281,500 pps
no-pool:         ~396,000 pps
```

Instead, a smaller PR with misc. changes discovered during the process that are unrelated to removing channels will be opened instead.

PR #535, opened 2025-02-04
feat(build.zig): cross-compile, install, and run on remote hosts over ssh

dnut:
This makes it easy to use `zig build` to target a remote host while building, installing, and running executables.

## Examples

Cross-compile all binaries for a remote host called `myec2`'s build target and install them on that host:
```bash
zig build -Dssh-host=myec2
```

Compile, install, and run sig on `myec2`:
```bash
zig build -Dssh-host=myec2 sig -- gossip -n testnet
```

Run the tests on `myec2`:
```bash
zig build -Dssh-host=myec2 test
```

## How it works
**Determine the build target:** `ssh.getHostTarget` - SSH into the server, run `zig targets`, then parse the json.

**Install to the remote host:** `ssh.addSendArtifact` - Analogous to `Build.addInstallArtifact`, except that it installs on a remote server. Uses `Build.addSystemCommand` to call rsync if installed, falling back to scp. rsync is preferred since it uses checksums to avoid redundant uploads. This depends on an InstallArtifact step.

**Run on the remote host:** `ssh.addRemoteCommand` - Analogous to `Build.addSystemCommand`, except that it runs on a remote host. Uses `Build.addSystemCommand` to run ssh.

## CLI arguments
```
  -Dssh-host=[string]          Builds will target this remote host, binaries will be installed there, and executables will run there.
  -Dssh-installdir=[string]    When using ssh-host, this configures the directory to install binaries (relative to ssh-workdir) (default: zig-out/bin).
  -Dssh-workdir=[string]       When using ssh-host, this configures the working directory where executables will run (default: sig).
```

This is also compatible with `no-run` and `no-bin`. Those options will be respected when deciding whether to install or run the binary on the remote host, with one exception. If you specify *only* `no-bin` for an artifact that needs to run on the remote host, then it will still install the artifact, because that is necessary in order to run it.


build.zig
```diff
@@ -351,3 +372,103 @@ fn defaultTargetDetectM3() ?std.Target.Query {
         .cpu_model = .{ .explicit = model },
     };
 }
+
+const ssh = struct {
+    /// SSH into the host and call `zig targets` to determine the compilation target
+    /// for that machine.
+    fn getHostTarget(b: *Build, remote_host: []const u8) !Build.ResolvedTarget {
+        const run_result = try std.process.Child.run(.{
```
InKryption, line 380: Should be noted that eventually this will cease to work, alongside some of our other build hacks, when build.zig becomes sandboxed under wasm - while I'm not against small hacks in the interim to improve developer workflow until the tooling catches up, I'm not sure I'm 100% comfortable with all the integrations done here to do the SSH commands.

Ideally there would be some concept of a `LazyResolvedTarget` to act as the result of a run step, which could then be used as the actual target we use; however such a thing doesn't exist, so I'd argue it makes more sense to separate out this specific ssh tooling out into its own script, whose result can be passed as the argument to `zig build`.

Ie `zig build -Dssh=$(zig run scripts/ssh.zig -- host work_dir install_dir)`, where the output is then a string that can be parsed into an override for the target, and the command to be executed on the remote host with a normal run step.

build.zig
```diff
@@ -351,3 +372,103 @@ fn defaultTargetDetectM3() ?std.Target.Query {
...
+    ) !*Build.Step.Run {
+        const local_path = b.getInstallPath(install.dest_dir.?, install.dest_sub_path);
+        defer b.allocator.free(remote_dir);
+
+        const remote_uri = try allocPrint(
+            b.allocator,
+            "{s}:{s}/{s}",
+            .{ host, remote_dir, install.dest_sub_path },
+        );
+        defer b.allocator.free(remote_uri);
```
InKryption, line 433:
```suggestion
        const remote_uri = try b.fmt(
            "{s}:{s}/{s}",
            .{ host, remote_dir, install.dest_sub_path },
        );
```

InKryption (changes requested):
I have some considerable concerns about how some of the additions here have been made, namely the direct usage of `std.process.Child` in the build script.
Some of them can be factored out, and one big one can't be, for which I have a solution in mind.

Beyond that, I actually quite like the rest of the changes factoring out of the build options into a consolidated struct, just a small comment on that.

build.zig
dnut: Originally I had this implemented as a separate script, but I didn't see an advantage to that so I moved everything to build.zig.

Introducing this code doesn't mean that our build process is now more dependent on fragile system dependencies. Nothing pre-existing from build.zig is broken if this stops working. This is just another layer that you can optionally use for extra features, like it would be if it were a separate script. So I don't see how it's helpful to move it into another script. It would just be more cumbersome to use.

build.zig
dnut: 2bb47f0743fe48a012f61ff7731c1a26eea7269e

build.zig
```diff
@@ -297,6 +222,102 @@ pub fn build(b: *Build) void {
     docs_step.dependOn(&install_sig_docs.step);
 }
 
+pub const Config = struct {
```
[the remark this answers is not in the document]
dnut: 016eacd1e4410b2d2de117a9749f373f27f903fe

build.zig
```diff
@@ -351,3 +372,103 @@ fn defaultTargetDetectM3() ?std.Target.Query {
...
+
+        const which_rsync = try std.process.Child.run(.{
+            .allocator = b.allocator,
+            .argv = &.{ "which", "rsync" },
+        });
+        const send = if (which_rsync.term == .Exited and which_rsync.term.Exited == 0)
+            b.addSystemCommand(&.{ "rsync", "--progress", local_path, remote_uri })
+        else
+            b.addSystemCommand(&.{ "scp", local_path, remote_uri });
+        send.step.dependOn(&mkdir.step);
```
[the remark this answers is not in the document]
dnut: When is build.zig going to be sandboxed? Is it a sure thing or still being discussed?

a166500810e721d3f6f6d65c19dfe85912e65777

build.zig
```diff
@@ -351,3 +372,103 @@ fn defaultTargetDetectM3() ?std.Target.Query {
...
     };
 }
+
+const ssh = struct {
+    /// SSH into the host and call `zig targets` to determine the compilation target
+    /// for that machine.
+    fn getHostTarget(b: *Build, remote_host: []const u8) !Build.ResolvedTarget {
+        const run_result = try std.process.Child.run(.{
+            .allocator = b.allocator,
+            .argv = &.{ "ssh", remote_host, "bash", "--login", "-c", "'zig targets'" },
```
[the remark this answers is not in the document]
dnut: I guess you had zig installed somewhere that's not normally in PATH and manually added it to PATH in .zprofile?

build.zig
```diff
@@ -297,6 +222,102 @@ pub fn build(b: *Build) void {
...
+/// this build script.
+fn addInstallAndRun(
+    b: *Build,
+    step: *Build.Step,
+    exe: *Build.Step.Compile,
+    config: Config,
+) !void {
+    var send_step: ?*Build.Step = null;
+
+    if (config.install or config.ssh_host != null and config.run) {
```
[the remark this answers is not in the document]
dnut: Normally when I write boolean expressions, I assume this precedence is common knowledge, like the analogous operations * and +. I'm not opposed to making it explicit here, but I'm guessing there are a lot of other places in the code where it's implicit.

8bacdea9c4f3a1cff695927d63e492204c722094

build.zig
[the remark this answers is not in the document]
Rexicon226: It is being discussed here: https://github.com/ziglang/zig/issues/14286, and I can attest that it's likely to get in.

build.zig
InKryption: I can see that argument, originally when I read this code the other adjacent additions/changes made me interpret it as being more deeply rooted than it actually is.
Add a TODO/NOTE comment linking to https://github.com/ziglang/zig/issues/14286 to clarify that we will only be relying on it as a temporary tool, until it no longer works, and as such should not be built upon.

dnut: ec9a91425ed0d989f3cbacf5f767ec7066ef5102

PR #552, opened 2025-02-07
feat(fuzz): fuzz with random multithreaded reads for accountsdb

Sobeston:
Rebased #532 to target main

### DIFF
--- a/src/accountsdb/fuzz.zig
+++ b/src/accountsdb/fuzz.zig
@@ -12,6 +12,8 @@ const AccountsDB = sig.accounts_db.AccountsDB;
 const FullSnapshotFileInfo = sig.accounts_db.snapshots.FullSnapshotFileInfo;
 const IncrementalSnapshotFileInfo = sig.accounts_db.snapshots.IncrementalSnapshotFileInfo;
 
+const N_RANDOM_THREADS = 8;
+
 pub const TrackedAccount = struct {
     pubkey: Pubkey,
     slot: u64,
@@ -134,11 +136,17 @@ pub fn run(seed: u64, args: *std.process.ArgIterator) !void {
         break :blk tracked_accounts;
     });
     var map_alive = std.atomic.Value(bool).init(true);
+
+    var threads: [N_RANDOM_THREADS]std.Thread = undefined;
+    var spawned_threads: u8 = 0;
+
     defer {
         map_alive.store(false, .seq_cst);
         const tracked_accounts, var tracked_accounts_lg = tracked_accounts_rw.writeWithLock();
         defer tracked_accounts_lg.unlock();
         tracked_accounts.deinit();
+
+        for (threads[0..spawned_threads]) |thread| thread.join();
     }
 
     var random_bank_fields = try BankFields.initRandom(allocator, random, 1 << 8);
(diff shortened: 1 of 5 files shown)


src/accountsdb/fuzz.zig
```diff
@@ -145,6 +155,17 @@ pub fn run(seed: u64, args: *std.process.ArgIterator) !void {
     var pubkeys_this_slot = std.AutoHashMap(Pubkey, void).init(allocator);
     defer pubkeys_this_slot.deinit();
 
+    for (0..8) |i| {
+        _ = try std.Thread.spawn(.{}, readRandomAccountThread, .{
```
kprotty, line 159: should `.detach()` the thread handles to avoid leaking their resources.

src/accountsdb/fuzz.zig
```diff
@@ -129,9 +129,19 @@ pub fn run(seed: u64, args: *std.process.ArgIterator) !void {
...
+    var tracked_accounts_rw = sig.sync.RwMux(Map).init(blk: {
+        var tracked_accounts = Map.init(allocator);
+        try tracked_accounts.ensureTotalCapacity(10_000);
+        break :blk tracked_accounts;
+    });
+    var map_alive = std.atomic.Value(bool).init(true);
+    defer {
+        map_alive.store(false, .seq_cst);
+        const tracked_accounts, var tracked_accounts_lg = tracked_accounts_rw.writeWithLock();
+        defer tracked_accounts_lg.unlock();
```
kprotty, line 142: Should join the spawned `readRandomAccountThread`s below. Seems like a reader could not see `map_alive=false`, get preempted before read locking, have this defer code run & deinit the map, then reader gets resumed and tries to read-lock it.

Sobeston: addressed in 972fb916f818893e5c4f7f64db5ee9dbe660a322

src/accountsdb/fuzz.zig
Sobeston: addressed in https://github.com/Syndica/sig/commit/972fb916f818893e5c4f7f64db5ee9dbe660a322

src/accountsdb/fuzz.zig
```diff
@@ -129,9 +131,25 @@ pub fn run(seed: u64, args: *std.process.ArgIterator) !void {
...
+    var threads: [N_RANDOM_THREADS]std.Thread = undefined;
+    var spawned_threads: u8 = 0;
+
+    defer {
+        map_alive.store(false, .seq_cst);
+        const tracked_accounts, var tracked_accounts_lg = tracked_accounts_rw.writeWithLock();
+        defer tracked_accounts_lg.unlock();
+        tracked_accounts.deinit();
+
+        for (threads[0..spawned_threads]) |thread| thread.join();
```
kprotty, line 151: Should prob join threads outside the writeLock to avoid a deadlock: `reader misses alive, goes to readLock; writer alive=false, writeLocks, waits on reader to join; reader readLock is waiting on writer writeLock`

Maybe move the `defer { alive=false; join threads }` after the `defer deinit`. Could also get rid of the `errdefer detach(); spawned = 0` down below.

0xNineteen: c8250b5b

src/accountsdb/fuzz.zig
```diff
@@ -140,6 +158,23 @@ pub fn run(seed: u64, args: *std.process.ArgIterator) !void {
     var pubkeys_this_slot = std.AutoHashMap(Pubkey, void).init(allocator);
     defer pubkeys_this_slot.deinit();
 
+    {
+        for (&threads) |*thread| {
+            defer spawned_threads += 1;
```
kprotty, line 163: Should happen after the spawn; Erroring out would still increment it & join an uninitialized thread in the above defer

kprotty (approved):

0xNineteen:
@Sobeston can you sign your commits

Sobeston:
Hm not sure why those weren't signed already. Had to force push to sign the old ones, sorry about that

kprotty (approved):

PR #554, opened 2025-02-07
feat(docs): add docusaurus

dutchers:
Adds docusaurus to the Sig repo.


Rexicon226:
btw, we require signed commits in the sig repo

Rexicon226 (changes requested):
You seem to have accidentally deleted the `docs/imgs/ci.png` file, we kinda need that. Maybe move it somewhere else now? I'm not sure what the intended structure of the `docs` is supposed to be now.

dutchers:
> You seem to have accidentally deleted the `docs/imgs/ci.png` file, we kinda need that. Maybe move it somewhere else now? I'm not sure what the intended structure of the `docs` is supposed to be now.

Re-added. I did a search for ci.png, but doesn't seem to be used anywhere. Let me know if I missed something.

Rexicon226:
> Re-added. I did a search for ci.png, but doesn't seem to be used anywhere. Let me know if I missed something.

Moreso used as documentation on how the current CI is laid out :) It's not used anywhere directly.

Rexicon226 (approved):
lgtm
think the commits need to be verified as well though

dutchers:
@Rexicon226 Finally got them verified. There was a discrepancy in my git config'd user.email and the email I'm using in github, so I modified the gpg key, re-added the public key to GitHub, and then resigned. Let me know if there is anything else.

PR #555, opened 2025-02-08
fix(prometheus): remove httpz again and fix prometheus metrics

Rexicon226:
After delving into the packets via Wireshark, I saw that the packets were using HTTP 1.0. This seemed to cause some issues in Prometheus, still not sure exactly what. Seems to me like it's a Prometheus bug regarding how explicit keep-alive is handled, but have no idea. HTTP sucks!

@InKryption why did you explicitly use HTTP 1.0? Kinda confused there.
@dnut I think you're the one who was having issues, can you test it out? I could reproduce the same context timeout issue and it's gone on my machine.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -47,9 +47,6 @@ pub fn build(b: *Build) void {
     const zig_cli_dep = b.dependency("zig-cli", dep_opts);
     const zig_cli_mod = zig_cli_dep.module("zig-cli");
 
-    const httpz_dep = b.dependency("httpz", dep_opts);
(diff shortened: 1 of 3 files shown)


Sobeston (approved):
I'm 100% sure that the issues we were running into with Prometheus are at the http level. I'm also not sure what, exactly, the cause is. But this works, and removing httpz is what we want to do. 

I'm happy, just double check with @InKryption before merging

InKryption (approved):

PR #562, opened 2025-02-11
refactor(cli): Some CLI options improvements and re-organization

InKryption:
The plan is to eventually move away from the zig-cli library we're using, but in the meantime it can't hurt to organize what we do have a bit better, to make the eventual switch easier to manage.

I manually checked that despite the slight re-ordering of some of the CLI options in this, no options were added to or removed from any commands.


Rexicon226 (approved):
LGTM
can't wait to move away from `zig-cli`, the API is so bad. at least in the 0.14 version, they no longer use pointers to the `Option` structs.

PR #564, opened 2025-02-11
feat(cli): Mock RPC Server

InKryption:
This will just allow me and potentially others to test the RPC Server manually (ie quickly testing and iterating on its behavior with `curl`). Also includes a refactor of `findAndUnpackTestSnapshots` that makes it easier to use its behavior elsewhere, via `findAndUnpackSnapshotFilePair` and the slightly lower level `unpackSnapshotFilePair`, the former of which is used in the mock RPC server.


src/cmd.zig
```diff
@@ -680,6 +680,24 @@ pub fn main() !void {
                             },
                         },
                     },
+                    &cli.Command{
+                        .name = "mock-rpc-server",
+                        .description = .{
+                            .one_line = "Run a mock RPC server.",
+                        },
+                        .options = &[_]*cli.Option{} ++
```
Rexicon226, line 688: Do you need this empty CLI option?

InKryption: It's just to force the following concatenations to have the same indentation.

InKryption: 87d17c0e261ee7137c12b362e3966a38d049890b

Rexicon226: Could use `\\`

Rexicon226: Bruh. Fair enough

0xNineteen (approved):

PR #565, opened 2025-02-12
refactor(rpc server): RPC Server refinements

InKryption:
* Fixes a lock guard leak in certain codepaths.
* Refactors the linux io_uring backend `Entry` definition to properly allow for more than just the `ACCEPT` alt allocationless state.
* Flushes headers in the basic server preemptively.
* Factor out http version into a constant in linux io_uring backend.
* Handle `/health` route, returning skeleton "unknown".
* Add skeleton tag for handling genesis file requests.

### DIFF
--- a/src/rpc/server/basic.zig
+++ b/src/rpc/server/basic.zig
@@ -68,6 +68,8 @@ pub fn acceptAndServeConnection(server_ctx: *server.Context) !void {
                     .content_length = archive_len,
                     .respond_options = .{},
                 });
+                // flush the headers, so that if this is a head request, we can mock the response without doing unnecessary work
+                try response.flush();
 
                 if (!response.elide_body) {
                     // use a length which is still a multiple of 2, greater than the send_buffer length,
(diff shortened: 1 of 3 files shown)


src/rpc/server/basic.zig
```diff
@@ -97,18 +99,31 @@ pub fn acceptAndServeConnection(server_ctx: *server.Context) !void {
...
+            .genesis_file => {},
+
+            .not_found => {},
         },
         .POST => {
             logger.err().logf("{} tried to invoke our RPC", .{conn_address});
-            return try request.respond("RPCs are not yet implemented", .{
+            try request.respond("RPCs are not yet implemented", .{
                 .status = .service_unavailable,
                 .keep_alive = false,
```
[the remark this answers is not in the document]
InKryption: Mostly to keep it in line with the io_uring backend which also doesn't emit any extra headers; I'm not keen to emit headers I don't understand the exact implications of, here in particular because we aren't actually doing anything to keep the connections alive.

src/rpc/server/linux_io_uring.zig
```diff
@@ -10,6 +10,8 @@ const IoUring = std.os.linux.IoUring;
 
 const LOGGER_SCOPE = "rpc.server.linux_io_uring";
 
+const HTTP_VERSION: std.http.Version = .@"HTTP/1.0";
+
```
Sobeston, line 14: Do we definitely want 1.0? If yes, a comment would be good. I think 1.1 is more commonly used

0xNineteen (approved):

src/rpc/server/linux_io_uring.zig
InKryption: We don't really support or use any HTTP/1.1 features as of now, and I don't really think the server should advertise itself as being 1.0 if it doesn't make any effort to be.

PR #566, opened 2025-02-12
feat(svm): more SBPFv3 support and add Poseidon Hash syscall

Rexicon226:
- fixes a ton of tiny little bugs with different components of the SVM
- implements the rest of the memory operations from SBPFv2/3
- adds the `sol_poseidon` syscall and testing for it


src/svm/elf.zig
```diff
@@ -475,7 +594,7 @@ pub const Elf = struct {
         if (text_section_slice.len % @sizeOf(sbpf.Instruction) != 0)
             return error.InvalidTextSectionLength;
 
-        if (self.config.minimum_version.enableElfVaddr()) {
+        if (self.version.enableElfVaddr()) {
             if (self.config.optimize_rodata != true) return error.UnsupportedSBPFVersion;
         }
```
yewman, line 599:
```suggestion
        if (self.version.enableElfVaddr() and !self.config.optimize_rodata)
            return error.UnsupportedSBPFVersion;
```

yewman (approved):
One nit suggestion otherwise looks good!

src/svm/elf.zig
Rexicon226: Agreed! 7fa6a5be

yewman (approved):
lgtm

PR #576, opened 2025-02-24
feat(vote-program): Authorize instructions

dadepo:
Implements the authorize related instructions. 

- authorize
- authorize_with_seed
- authorize_checked
- authorize_checked_with_seed

This also includes support for versioned state and ability to convert from previous states to the current vote state.

A bulk of the methods on the `VoteState` where also implemented as part of implementing the authorize instructions.


src/runtime/program/vote/state.zig
```diff
@@ -132,7 +476,100 @@ pub const VoteState = struct {
...
+        // The offset in slots `n` on which the target_epoch
+        // (default value `DEFAULT_LEADER_SCHEDULE_SLOT_OFFSET`) is
+        // calculated is the number of slots available from the
+        // first slot `S` of an epoch in which to set a new voter for
+        // the epoch at `S` + `n`
+        if (self.authorized_voters.contains(target_epoch)) {
+            return VoteError.TooSoonToReauthorize;
+        }
+
+        const epoch, const pubkey = self.authorized_voters.last() orelse
```
0xNineteen, line 495:
```suggestion
        const latest_epoch, const latest_authorized_pubkey = self.authorized_voters.last() orelse
```

src/runtime/program/vote/state.zig
```diff
@@ -132,7 +476,100 @@ pub const VoteState = struct {
...
+
+            self.prior_voters.append(PriorVote{
+                .key = pubkey,
+                .start = epoch_of_last_authorized_switch,
+                .end = target_epoch,
+            });
+        }
+
+        self.authorized_voters.insert(target_epoch, new_authorized_voter) catch
+        // TODO: Is it okay to convert out of memory to InvalidAccountData?
```
0xNineteen, line 516: doesnt make sense to me, do we need only instruction errors as return types?

src/runtime/program/vote/execute.zig
```diff
@@ -138,10 +174,287 @@ fn intializeAccount(
...
+    signers: ?std.AutoHashMap(Pubkey, void),
+) InstructionError!void {
+    const versioned_state = try vote_account.deserializeFromAccountData(
+        allocator,
+        VoteStateVersions,
+    );
+    var vote_state = versioned_state.convertToCurrent(allocator) catch {
+        // TODO okay to convert out of memory to InvalidAccountData?
+        return InstructionError.InvalidAccountData;
+    };
```
0xNineteen, line 225: i think if we OOM in the program exec we should panic - and we shouldnt do any conversions in the state methods - its equivalent to what would happen in rust

src/runtime/program/vote/execute.zig
```diff
@@ -138,10 +174,287 @@ fn intializeAccount(
...
+/// but will implicitly withdraw authorization from the previously authorized
+/// key.
+fn authorize(
+    allocator: std.mem.Allocator,
+    ic: *InstructionContext,
+    vote_account: *BorrowedAccount,
+    authorized: Pubkey,
+    vote_authorize: VoteAuthorize,
+    clock: Clock,
+    signers: ?std.AutoHashMap(Pubkey, void),
```
0xNineteen, line 216: why is signers optional here?

src/runtime/program/vote/execute.zig
```diff
@@ -138,10 +174,287 @@ fn intializeAccount(
...
+    try vote_account.serializeIntoAccountData(VoteStateVersions{ .current = vote_state });
+}
+
+/// Analogous to
+/// https://github.com/anza-xyz/agave/blob/49fb51295c1062b6b09e585b2fe0a4676c33d3d4/programs/vote/src/vote_state/mod.rs#L691-L692
+/// https://github.com/anza-xyz/agave/blob/49fb51295c1062b6b09e585b2fe0a4676c33d3d4/programs/vote/src/vote_state/mod.rs#L701-L707
+/// https://github.com/anza-xyz/solana-sdk/blob/4e30766b8d327f0191df6490e48d9ef521956495/vote-interface/src/state/mod.rs#L873
+///
+/// Check if the current authorized withdrawer or epoch authorized voter is a signer.
+fn checkAuthority(
```
0xNineteen, line 280:
```suggestion
fn validateAuthorityIsSigner(
```

i think `validate` is the word we want to use going forward - we talked about this in the past in a random PR i think - @yewman is this correct?

src/runtime/program/vote/execute.zig
```diff
@@ -138,10 +174,287 @@ fn intializeAccount(
...
+
+    // current authorized withdrawer or epoch authorized voter must say "yay"
+    if (!authorized_withdrawer_signer) {
+        const epoch_authorized_voter = try vote_state.getAndUpdateAuthorizedVoter(
+            allocator,
+            current_epoch,
+        );
+        if (!ic.info.isPubkeySigner(epoch_authorized_voter)) {
+            return InstructionError.MissingRequiredSignature;
+        }
```
0xNineteen, line 300: validate methods shouldnt have state modification code in them imo - would be good to pull this out and refactor

src/runtime/program/vote/execute.zig
```diff
@@ -138,10 +174,287 @@ fn intializeAccount(
...
+fn verifyAuthorizedSigner(
+    authorized: Pubkey,
+    signers: std.AutoHashMap(Pubkey, void),
+) InstructionError!bool {
+    if (signers.contains(authorized)) {
+        return true;
+    } else {
+        return InstructionError.MissingRequiredSignature;
+    }
 }
```
0xNineteen, line 455: should we move this now?

src/runtime/program/vote/execute.zig
```diff
@@ -138,10 +174,287 @@ fn intializeAccount(
...
+    defer vote_state.deinit();
+
+    switch (vote_authorize) {
+        .voter => {
+            const current_epoch = clock.epoch;
+            const target_epoch = std.math.add(u64, current_epoch, 1) catch {
+                return InstructionError.InvalidAccountData;
+            };
+
+            try checkAuthority(
```
0xNineteen, line 235: shouldnt this be `verifyAuthorizedSigner` ?

src/runtime/program/vote/execute.zig
```diff
@@ -138,10 +174,287 @@ fn intializeAccount(
...
+            owner,
+        ) catch |err| {
+            ic.tc.custom_error = @intFromError(err);
+            return InstructionError.Custom;
+        };
+        expected_authority_keys.put(created, {}) catch |err| {
+            // TODO okay to convert out of memory to custom error?
+            ic.tc.custom_error = @intFromError(err);
+            return InstructionError.Custom;
+        };
```
0xNineteen, line 360:
```suggestion
        var signer_account = try ic.borrowInstructionAccount(signer_index);
        defer signer_account.release();
        
        const created = pubkey_utils.createWithSeed(
            signer_account.pubkey,
            seed,
            owner,
        ) catch |err| {
            ic.tc.custom_error = @intFromError(err);
            return InstructionError.Custom;
        };
        
        expected_authority_keys.put(created, {}) catch |err| {
            // TODO okay to convert out of memory to custom error?
            ic.tc.custom_error = @intFromError(err);
            return InstructionError.Custom;
        };
```

small nit on the spacing

src/runtime/program/vote/execute.zig
```diff
@@ -138,10 +174,287 @@ fn intializeAccount(
...
+    if (!ic.info.isPubkeySigner(new_authority.pubkey)) {
+        return InstructionError.MissingRequiredSignature;
+    }
+
+    const clock = try ic.getSysvarWithAccountCheck(
+        Clock,
+        @intFromEnum(vote_instruction.VoteAuthorize.AccountIndex.clock_sysvar),
+    );
+
+    const autorize = switch (vote_authorize) {
```
0xNineteen, line 429:
```suggestion
    const authorize_pubkey = switch (vote_authorize) {
```

src/runtime/program/vote/execute.zig
0xNineteen, line 455:
```suggestion
fn validateIsSigner(
    pubkey: Pubkey,
    signers: std.AutoHashMap(Pubkey, void),
) InstructionError!void {
    if (!signers.contains(authorized)) {
        return InstructionError.MissingRequiredSignature;
    }
}
```

0xNineteen (changes requested):

src/runtime/program/vote/instruction.zig
```diff
@@ -26,6 +26,73 @@ pub const IntializeAccount = struct {
...
+        ///  `[SIGNER]` Base key of current Voter or Withdrawer authority's derived key
+        base_key = 2,
+        /// `[SIGNER]` New vote or withdraw authority
+        // TODO make this uniform.
+        new_authority = 3,
+    };
+};
+
+pub const VoteAuthorize = enum {
+    // TODO lowercase
```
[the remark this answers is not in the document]
dadepo: Note to rename the enum variant to lowercase. Will push the change in a bit.

src/runtime/program/vote/execute.zig
dadepo: [c711769](https://github.com/Syndica/sig/pull/576/commits/c71176924ce327a39bfa331a1129ca68fe7473c8)

src/runtime/program/vote/execute.zig
dadepo: [c711769](https://github.com/Syndica/sig/pull/576/commits/c71176924ce327a39bfa331a1129ca68fe7473c8)

src/runtime/program/vote/state.zig
dadepo: [c711769](https://github.com/Syndica/sig/pull/576/commits/c71176924ce327a39bfa331a1129ca68fe7473c8)

src/runtime/program/vote/state.zig
```diff
@@ -51,11 +53,351 @@ pub const EpochCredit = struct {
...
+    v0_23_5: VoteState0_23_5,
+    v1_14_11: VoteState1_14_11,
+    current: VoteState,
+
+    /// Agave https://github.com/anza-xyz/solana-sdk/blob/4e30766b8d327f0191df6490e48d9ef521956495/vote-interface/src/state/vote_state_versions.rs#L80
+    fn landedVotesFromLockouts(
+        allocator: std.mem.Allocator,
+        lockouts: std.ArrayList(Lockout),
+    ) !std.ArrayList(LandedVote) {
+        var landedVotes = std.ArrayList(LandedVote).init(allocator);
```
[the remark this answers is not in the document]
dadepo: [c711769](https://github.com/Syndica/sig/pull/576/commits/c71176924ce327a39bfa331a1129ca68fe7473c8)

src/runtime/program/vote/state.zig
```diff
@@ -51,11 +53,351 @@ pub const EpochCredit = struct {
...
+
+    pub fn insert(self: *AuthorizedVoters, epoch: Epoch, authorizedVoter: Pubkey) !void {
+        try self.authorized_voters.put(epoch, authorizedVoter);
+    }
+
+    /// Agave https://github.com/anza-xyz/solana-sdk/blob/4e30766b8d327f0191df6490e48d9ef521956495/vote-interface/src/authorized_voters.rs#L42
+    pub fn purgeAuthorizedVoters(
+        self: *AuthorizedVoters,
+        allocator: std.mem.Allocator,
+        currentEpoch: Epoch,
```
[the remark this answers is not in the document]
dadepo: [c711769](https://github.com/Syndica/sig/pull/576/commits/c71176924ce327a39bfa331a1129ca68fe7473c8)

src/runtime/program/vote/instruction.zig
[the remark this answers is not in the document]
dadepo: [c711769](https://github.com/Syndica/sig/pull/576/commits/c71176924ce327a39bfa331a1129ca68fe7473c8)

src/runtime/program/vote/execute.zig
dadepo: Actually had it with `validate` prefix but changed it when I noticed this https://github.com/Syndica/sig/blob/3516baf98b64bd8c2a6f63e14703927ddf8244c5/src/runtime/program/system_program/execute.zig#L782 in the system program

src/runtime/program/vote/execute.zig
```diff
@@ -138,10 +174,289 @@ fn intializeAccount(
...
+    authorization_type: VoteAuthorize,
+    current_authority_derived_key_owner: Pubkey,
+    current_authority_derived_key_seed: []const u8,
+) InstructionError!void {
+    try ic.info.checkNumberOfAccounts(4);
+
+    var new_authority = try ic.borrowInstructionAccount(
+        @intFromEnum(vote_instruction.VoteAuthorizeCheckedWithSeedArgs.AccountIndex.new_authority),
+    );
+    defer new_authority.release();
```
dadepo, line 393: Need to check `new_authority` is also a signer

src/runtime/program/vote/state.zig
dadepo: [5b20988](https://github.com/Syndica/sig/pull/576/commits/5b20988e3e7b71b6b3267bf464732fdfcf9c203b)

src/runtime/program/vote/execute.zig
dadepo: [5b20988](https://github.com/Syndica/sig/pull/576/commits/5b20988e3e7b71b6b3267bf464732fdfcf9c203b)

src/runtime/program/vote/state.zig
```diff
@@ -51,11 +53,351 @@ pub const EpochCredit = struct {
...
+    ) bool {
+        var expired_keys = std.ArrayList(Epoch).init(allocator);
+        defer expired_keys.deinit();
+
+        var voter_iter = self.authorized_voters.iterator();
+        while (voter_iter.next()) |entry| {
+            if (entry.key_ptr.* < currentEpoch) {
+                expired_keys.append(entry.key_ptr.*) catch unreachable;
```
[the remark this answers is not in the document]
dadepo: [eb22f80](https://github.com/Syndica/sig/pull/576/commits/eb22f803f6e7b988c196201c5b43109b533c656a)

src/runtime/program/vote/execute.zig
dadepo: > validate methods shouldnt have state modification code in them imo

I agree.

While trying to pull it out, I ended up dropping the method altogether. This is because `vote_state.getAndUpdateAuthorizedVoter` which contains the state update is core to the check that is being performed but it should only be called if initial check is `authorized_withdrawer_signer`. Trying to pull this out would mean the check function would be redundant or such that it calls a passed in function (which contains the vote_state.getAndUpdateAuthorizedVoter call)


[694a8c6](https://github.com/Syndica/sig/pull/576/commits/694a8c64b1f56d8b9f4531069ea163b9d78be19d)

src/runtime/program/vote/execute.zig
dadepo: Dropped in [694a8c6](https://github.com/Syndica/sig/pull/576/commits/694a8c64b1f56d8b9f4531069ea163b9d78be19d)

src/runtime/program/vote/state.zig
```diff
@@ -51,11 +53,351 @@ pub const EpochCredit = struct {
...
+    pub fn deinit(self: VoteStateVersions) void {
+        switch (self) {
+            .v0_23_5 => |vote_state| vote_state.deinit(),
+            .v1_14_11 => |vote_state| vote_state.deinit(),
+            .current => |vote_state| vote_state.deinit(),
+        }
+    }
+
+    // For bincode
+    pub fn serializedSize(self: VoteStateVersions) !usize {
```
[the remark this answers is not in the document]
dadepo: [8059203](https://github.com/Syndica/sig/pull/576/commits/80592039d49a2daab66dc603f566aa17d8646137)

src/runtime/program/vote/execute.zig
0xNineteen: huh yeah that ref should be validate too imo

src/runtime/program/vote/state.zig
```diff
@@ -51,11 +53,351 @@ pub const EpochCredit = struct {
...
+
+        var voter_iter = self.authorized_voters.iterator();
+        while (voter_iter.next()) |entry| {
+            if (entry.key_ptr.* < currentEpoch) {
+                expired_keys.append(entry.key_ptr.*) catch unreachable;
+            }
+        }
+
+        for (expired_keys.items) |key| {
+            _ = self.authorized_voters.orderedRemove(key);
```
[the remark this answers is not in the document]
dadepo: The `SortedMap` has a `fetchSwapRemove` method that can be used for the swap remove, but `SortedMap` does not expose a `sort` method which can be used to re-order after the loop? the `sort` method is called internally on some of the method calls. for example `keys()`. 

An approach could be to call `fetchSwapRemove` and then `keys()` but not sure if that won't end up being inefficient also (and maybe cumbersome code wise?)

src/runtime/program/vote/execute.zig
dadepo: [4fb72ac](https://github.com/Syndica/sig/pull/576/commits/4fb72acd549e37ba2bbcf7b3002908cabf671bd8)

src/runtime/program/vote/execute.zig
dadepo: The agave code has the same call. It is here https://github.com/anza-xyz/solana-sdk/blob/c07f692e41d757057c8700211a9300cdcd6d33b1/vote-interface/src/state/mod.rs#L872 

Only that it is not obvious, because it is called and passed to a closure that is passed as an argument to set_new_authorized_voter. See https://github.com/anza-xyz/agave/blob/131d3b3e007b1505231c726b98489aae40234a39/programs/vote/src/vote_state/mod.rs#L701

0xNineteen: ah i see okay thank you

src/runtime/program/vote/state.zig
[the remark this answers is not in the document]
0xNineteen: would you want to add a `swapRemoveNoSort` and make the `sort` method public?

src/runtime/program/vote/execute.zig
dadepo: Started doing that, but found out that it does not take `InstructionInfo` as self (like the other related methods: `isPubkeySigner` etc), and both of it's inputs are not tied to the instruction. So feels more like a utility function.

I have left it here, and removed the todo f804999 if this is needed somewhere else perhaps then it can be moved.

0xNineteen (dismissed):
lgtm

yewman (changes requested):
A couple of comments from an initial pass :)

src/runtime/program/vote/error.zig
```diff
@@ -0,0 +1,28 @@
+const sig = @import("../../../sig.zig");
+
+/// Agave https://github.com/anza-xyz/solana-sdk/blob/4e30766b8d327f0191df6490e48d9ef521956495/vote-interface/src/error.rs#L11
+///
+/// Reasons the vote might have had an error
+pub const VoteError = error{
```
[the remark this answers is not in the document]
dadepo: [cbf8ea4](https://github.com/Syndica/sig/pull/576/commits/cbf8ea4b4bed99831dd29a407560ba2d179b7c02) Although the return type of `setNewAuthorizedVoter` now looks a bit awkward. Not sure how that can be improved.

src/runtime/program/vote/execute.zig
```diff
@@ -129,10 +163,268 @@ fn intializeAccount(
         clock,
     );
     defer vote_state.deinit();
-    try vote_account.serializeIntoAccountData(vote_state);
+    try vote_account.serializeIntoAccountData(VoteStateVersions{ .current = vote_state });
```
Rexicon226, line 166:
```suggestion
    try vote_account.serializeIntoAccountData(.{ .current = vote_state });
```

src/runtime/program/vote/execute.zig
```diff
@@ -138,10 +173,273 @@ fn intializeAccount(
...
+/// but will implicitly withdraw authorization from the previously authorized
+/// key.
+fn authorize(
+    allocator: std.mem.Allocator,
+    ic: *InstructionContext,
+    vote_account: *BorrowedAccount,
+    authorized: Pubkey,
+    vote_authorize: VoteAuthorize,
+    clock: Clock,
+    signers: ?std.AutoHashMap(Pubkey, void),
```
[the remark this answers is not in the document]
Rexicon226: Depends if searching through the `signers` map is time-critical. Moving to a slice would make it O(N). If that doesn't matter, I agree with Harold and would prefer a slice. Either way, it shouldn't be a managed type, the allocator is passed in just above.

src/runtime/program/vote/execute.zig
```diff
@@ -129,10 +163,268 @@ fn intializeAccount(
...
+            const target_epoch = std.math.add(u64, current_epoch, 1) catch {
+                return InstructionError.InvalidAccountData;
+            };
+
+            // Analogous to
+            // https://github.com/anza-xyz/agave/blob/49fb51295c1062b6b09e585b2fe0a4676c33d3d4/programs/vote/src/vote_state/mod.rs#L691-L692
+            // https://github.com/anza-xyz/agave/blob/49fb51295c1062b6b09e585b2fe0a4676c33d3d4/programs/vote/src/vote_state/mod.rs#L701-L707
+            // https://github.com/anza-xyz/solana-sdk/blob/4e30766b8d327f0191df6490e48d9ef521956495/vote-interface/src/state/mod.rs#L873
+            {
+                const authorized_withdrawer_signer = if (signers) |signers_|
```
Rexicon226, line 226: Don't prepend or append identifiers with `_`. Call it `maybe_signers` and `signers` if it does need to be optional.

src/runtime/program/vote/execute.zig
```diff
@@ -129,10 +163,268 @@ fn intializeAccount(
...
+                    }
+                }
+            }
+
+            _, const err = try vote_state.setNewAuthorizedVoter(
+                authorized,
+                target_epoch,
+            );
+
+            if (err) |err_| {
```
Rexicon226, line 248: Same here, I'd like us to not use `_` like this.

src/runtime/program/vote/execute.zig
```diff
@@ -129,10 +163,268 @@ fn intializeAccount(
...
+        new_authority.pubkey,
+        authorize_pubkey,
+        clock,
+        null,
+    );
+}
+
+fn validateIsSigner(
+    authorized: Pubkey,
+    signers: std.AutoHashMap(Pubkey, void),
```
Rexicon226, line 418: Shouldn't be managed.

src/runtime/program/vote/execute.zig
```diff
@@ -226,3 +518,546 @@ test "executeIntializeAccount" {
...
+            },
+        },
+    );
+}
+
+test "vote_program: executeAuthorize voter signed by current withdrawer" {
+    const ids = sig.runtime.ids;
+    const testing = sig.runtime.program.testing;
+    const PriorVote = sig.runtime.program.vote_program.state.PriorVote;
+    _ = &PriorVote;
```
Rexicon226, line 629: Useless discard.

src/runtime/program/vote/state.zig
```diff
@@ -51,11 +53,338 @@ pub const EpochCredit = struct {
...
+
+    pub fn deinit(self: VoteState1_14_11) void {
+        self.votes.deinit();
+        self.authorized_voters.deinit();
+        self.epoch_credits.deinit();
+    }
+
+    /// Upper limit on the size of the Vote State
+    /// when votes.len() is MAX_LOCKOUT_HISTORY.
+    pub fn sizeOf() usize {
```
Rexicon226, line 382: This should be a constant.

src/runtime/program/vote/state.zig
```diff
@@ -132,7 +463,474 @@ pub const VoteState = struct {
...
     }
 
-    pub fn serializedSize(self: VoteState) !usize {
-        return sig.bincode.sizeOf(self, .{});
+    /// https://github.com/anza-xyz/solana-sdk/blob/4e30766b8d327f0191df6490e48d9ef521956495/vote-interface/src/state/mod.rs#L862
+    pub fn setNewAuthorizedVoter(
+        self: *VoteState,
+        new_authorized_voter: Pubkey,
+        target_epoch: Epoch,
+    ) (error{OutOfMemory} || InstructionError)!struct { void, ?VoteError } {
```
Rexicon226, line 471: I do not understand why this is returning a void tuple member.

src/runtime/program/vote/state.zig
```diff
@@ -132,7 +463,474 @@ pub const VoteState = struct {
...
+    commission: u8,
+) VoteState {
+    if (!builtin.is_test) {
+        @panic("createTestVoteState should only be called in test mode");
+    }
+
+    return .{
+        .node_pubkey = node_pubkey,
+        .authorized_voters = if (authorized_voter) |authorized_voter_|
+            AuthorizedVoters.init(allocator, 0, authorized_voter_) catch unreachable
```
Rexicon226, line 544: I see no reason not to use `try` here instead of `catch unreachable`.

Rexicon226 (changes requested):
Just gave it a quick read through, haven't compared to Agave yet.

src/runtime/program/vote/execute.zig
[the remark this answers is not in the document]
dadepo: 5eb2846

src/runtime/program/vote/state.zig
```diff
@@ -51,11 +53,338 @@ pub const EpochCredit = struct {
...
+
+    pub fn len(self: *const AuthorizedVoters) usize {
+        return self.authorized_voters.count();
+    }
+
+    pub fn contains(self: *const AuthorizedVoters, epoch: Epoch) bool {
+        return self.authorized_voters.contains(epoch);
+    }
+
+    // TODO Add method that returns iterator over authorized_voters
```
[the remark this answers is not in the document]
dadepo: It is the iterator. Put the todo there as Agave has a method that exposes the iterator...but we necessary don't need it as it can easily be accessed directly via`authorized_voters.iterator()`. I'll remove the TODO

src/runtime/program/vote/state.zig
```diff
@@ -51,11 +53,338 @@ pub const EpochCredit = struct {
...
+                .epoch_credits = state.epoch_credits,
+                .last_timestamp = state.last_timestamp,
+            },
+            .current => |state| return state,
+        }
+    }
+};
+
+/// Agave https://github.com/anza-xyz/solana-sdk/blob/4e30766b8d327f0191df6490e48d9ef521956495/vote-interface/src/state/vote_state_0_23_5.rs#L11
+pub const VoteState0_23_5 = struct {
```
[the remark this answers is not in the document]
dadepo: I think there are 3 variants of this already 😬 in the code base. I started using `/// [agave] ...` but then switched over to `/// Agave ` mainly because /// [agave] felt like an incomplete markdown link, and slightly prefer having Agave capitalized...

src/runtime/program/vote/execute.zig
dadepo: Note sure why, but this broke the test with the DataMismatch error. Perhaps this affected the bincode serialisation?

src/runtime/program/vote/execute.zig
dadepo: [0a22f64](https://github.com/Syndica/sig/pull/576/commits/0a22f6490bce6ee73f6dc066e8d50b883a4f156a)

src/runtime/program/vote/state.zig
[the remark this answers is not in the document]
dadepo: [0a22f64](https://github.com/Syndica/sig/pull/576/commits/0a22f6490bce6ee73f6dc066e8d50b883a4f156a)

src/runtime/program/vote/execute.zig
```diff
@@ -138,10 +173,273 @@ fn intializeAccount(
...
+        allocator,
+        VoteStateVersions,
+    );
+    var vote_state = try versioned_state.convertToCurrent(allocator);
+    defer vote_state.deinit();
+
+    switch (vote_authorize) {
+        .voter => {
+            const current_epoch = clock.epoch;
+            const target_epoch = std.math.add(u64, current_epoch, 1) catch {
```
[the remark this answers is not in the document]
dadepo: Curious to know to which extent changes/improvement can be made. Esp in cases where the code would semantically still be the same.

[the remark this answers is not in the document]
yewman: As long as it doesn't modify the behaviour of the program its fine

src/runtime/program/vote/execute.zig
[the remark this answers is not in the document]
yewman: Sorry, it wasn't clear in the original comment but I think the slice should not be optional and we should only search for signers within this slice. The control flow of searching maybe signers and conditionally searching the instruction context is kind of unclear and error prone imo. Lmk if I am missing something

src/runtime/program/vote/error.zig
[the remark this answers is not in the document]
yewman: We could use a mapError approach as well, see pubkey_utils. However, that approach has its own drawbacks as well

src/runtime/program/vote/execute.zig
Rexicon226: Ah, I didn't see that it's a union. My bad.

src/runtime/program/vote/execute.zig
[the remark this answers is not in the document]
dadepo: Yeah I was about getting to that, but wanted ti check the mapping to the Agave implementation again to make sure things are similar.

src/runtime/program/vote/state.zig
dadepo: See https://github.com/Syndica/sig/pull/576#discussion_r1999386049 and the diff cbf8ea4 

This became the case when VoteError was turned into an enum instead of an error, previously this method was returning `(error{OutOfMemory} || InstructionError || VoteError)!void`

src/runtime/program/vote/state.zig
dadepo: [68f58af](https://github.com/Syndica/sig/pull/576/commits/68f58af64a339c8a1f6b0d2b62d761ec0bf8e937)

src/runtime/program/vote/execute.zig
```diff
@@ -138,10 +173,273 @@ fn intializeAccount(
...
+            // https://github.com/anza-xyz/solana-sdk/blob/4e30766b8d327f0191df6490e48d9ef521956495/vote-interface/src/state/mod.rs#L873
+            {
+                const authorized_withdrawer_signer = if (signers) |signers_|
+                    try validateIsSigner(vote_state.authorized_withdrawer, signers_)
+                else
+                    ic.info.isPubkeySigner(vote_state.authorized_withdrawer);
+
+                // current authorized withdrawer or epoch authorized voter must say "yay"
+                if (!authorized_withdrawer_signer) {
+                    const epoch_authorized_voter = try vote_state.getAndUpdateAuthorizedVoter(
```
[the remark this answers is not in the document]
dadepo: You mean in Agave `get_and_update_authorized_voter()` gets called regardless of the value of `authorized_withdrawer_signer`? This was supposed to optimize that, But given that `get_and_update_authorized_voter()` takes a `&mut self`, this might introduce subtle difference in behaviour.

I'll update to also be called regardless of  `authorized_withdrawer_signer` to keep in sync.

src/runtime/program/vote/state.zig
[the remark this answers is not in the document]
0xNineteen: would prefer to use `/// [agave]`

src/runtime/program/vote/state.zig
Rexicon226: Yeah but why is it returning void?

Rexicon226: This does not seem like the correct return type to me. I think it should be a union, something like
```zig
const Result = union(enum) {
	ok: void,
	err: VoteError,
};
```

src/runtime/program/vote/state.zig
Rexicon226: Looking at agave, they have a `sizeOf` function on the `VoteStateVersions`
```rs
pub fn vote_state_size_of(is_current: bool) -> usize {
  if is_current {
      VoteState::size_of()
  } else {
      VoteState1_14_11::size_of()
  }
}
```

`VoteState::size_of()` is `3762` and `VoteState1_14_11::size_of()` is `3731`. Where does `64` come from?

Those "sub" `size_of()` calls should be constants in Sig, do we need this `vote_state_size_of` function?

src/runtime/program/vote/execute.zig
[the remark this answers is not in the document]
dadepo: [c47f772](https://github.com/Syndica/sig/pull/576/commits/c47f7723b66bb8b67f1a68d19f9c7e89deadbe26)

src/runtime/program/vote/state.zig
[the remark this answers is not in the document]
dadepo: [d8e2253](https://github.com/Syndica/sig/pull/576/commits/d8e225374584a8afae6edc8f167f2b79e2f7f342)

src/runtime/program/vote/execute.zig
```diff
@@ -138,10 +173,273 @@ fn intializeAccount(
...
+                        allocator,
+                        current_epoch,
+                    );
+                    if (!ic.info.isPubkeySigner(epoch_authorized_voter)) {
```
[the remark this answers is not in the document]
dadepo: [c47f772](https://github.com/Syndica/sig/pull/576/commits/c47f7723b66bb8b67f1a68d19f9c7e89deadbe26)

src/runtime/program/vote/execute.zig
```diff
@@ -129,10 +164,273 @@ fn intializeAccount(
...
+                target_epoch,
+            ) catch |err| {
+                switch (err) {
+                    VoteError.TooSoonToReauthorize => {
+                        ic.tc.custom_error = @intFromError(err);
+                        return InstructionError.Custom;
+                    },
+                    else => {
+                        return InstructionError.InvalidAccountData;
+                    },
```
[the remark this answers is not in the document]
dadepo: Updated as part of cbf8ea4

src/runtime/program/vote/execute.zig
```diff
@@ -129,10 +164,273 @@ fn intializeAccount(
...
+    }
+
+    try authorize(
+        allocator,
+        ic,
+        vote_account,
+        new_authority,
+        authorization_type,
+        clock,
+        if (expected_authority_keys.count() > 0)
```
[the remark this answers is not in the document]
dadepo: Updated as part of [c47f772](https://github.com/Syndica/sig/pull/576/commits/c47f7723b66bb8b67f1a68d19f9c7e89deadbe26) and I think it should now be clearer and map more easily to Agave...the signers that is checked would either be the one from the transaction or the ones derived from the seed.

src/runtime/program/vote/state.zig
dadepo: the 64 was not meant to be checked in. Fixed that here [bdd89bf](https://github.com/Syndica/sig/pull/576/commits/bdd89bfcbaa3c4db3b3c17872c9c32d8fe18dffe)

> do we need this vote_state_size_of function?

I don't think so. Everywhere it is used, true is passed to it, meaning it currently only returns the size of the current vote state. It looks like a temp function they added.

src/runtime/program/vote/execute.zig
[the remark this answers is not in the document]
dadepo: [b05f606](https://github.com/Syndica/sig/pull/576/commits/b05f606a9b4041b3beaae9e92b30bf89a61216c9)

src/runtime/program/vote/state.zig
dadepo: That can work. I'll try this and Harold's suggestion of having a `mapError` similar to `pubkey_utils`

src/runtime/program/vote/state.zig
dadepo: Happy to change it, but does the fact that it is meant only to be used in tests provide a reason? It has a `!builtin.is_test` check that enforces it. I guess I had the `catch unreachable` for the minor convenience of not having to use `try` when calling it

src/runtime/program/vote/execute.zig
dadepo: 5eb28462  [c47f772](https://github.com/Syndica/sig/pull/576/commits/c47f7723b66bb8b67f1a68d19f9c7e89deadbe26)

src/runtime/program/vote/state.zig
Rexicon226: Yeah, I was going to mention something similar. I'm happy with this version :).

src/runtime/program/vote/state.zig
Rexicon226: Do you find `catch unreachable` more convenient than using `try`? To me, it just seems long, scarier, and harder to understand to someone less experienced in Zig. Also if we're being pedantic, it's not unreachable, so it's not correct.

src/runtime/program/vote/execute.zig
```diff
@@ -129,10 +163,275 @@ fn intializeAccount(
...
+    try ic.info.checkNumberOfAccounts(4);
+
+    var new_authority = try ic.borrowInstructionAccount(
+        @intFromEnum(vote_instruction.VoteAuthorize.AccountIndex.new_authority),
+    );
+    defer new_authority.release();
+
+    if (!ic.info.isPubkeySigner(new_authority.pubkey)) {
+        return InstructionError.MissingRequiredSignature;
+    }
```
yewman, line 394:
```suggestion
    // Safe since there are at least 4 accounts, and the new_authority index is 3.
    const new_authority_meta = &ic.info.account_metas.buffer[
        @intFromEnum(vote_instruction.VoteAuthorize.AccountIndex.new_authority)
    ];
    if (!new_authority_meta.is_signer) {
        return InstructionError.MissingRequiredSignature;
    }
```

yewman (changes requested):
A couple of extra comments from second pass :)

src/runtime/program/vote/state.zig
yewman: I think just returning ?VoteError does the trick for the use cases in this PR

src/runtime/program/vote/state.zig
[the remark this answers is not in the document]
yewman: I think there are still a few left

src/runtime/program/vote/execute.zig
```diff
@@ -129,10 +164,273 @@ fn intializeAccount(
...
+    try ic.info.checkNumberOfAccounts(4);
+
+    var new_authority = try ic.borrowInstructionAccount(
+        @intFromEnum(vote_instruction.VoteAuthorizeCheckedWithSeedArgs.AccountIndex.new_authority),
+    );
+    defer new_authority.release();
+
+    if (!ic.info.isPubkeySigner(new_authority.pubkey)) {
+        return InstructionError.MissingRequiredSignature;
+    }
```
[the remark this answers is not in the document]
yewman: One caveat is iff Agave borrows an account, we should also borrow since this will throw an error if the account is not present in the transaction context. Basically we should borrow iff agave borrows in the same place. We may be able to remove some unnecessary borrows down the line but debugging will be easier if we simply borrow where agave does for now

src/runtime/program/vote/execute.zig
```diff
@@ -129,10 +163,275 @@ fn intializeAccount(
...
+        ic,
+        vote_account,
+        new_authority.pubkey,
+        authorize_pubkey,
+        clock,
+        signers.items,
+    );
+}
+
+fn validateIsSigner(
```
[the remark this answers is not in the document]
dadepo: [a383d70](https://github.com/Syndica/sig/pull/576/commits/a383d70165053e627c8cd8283bdba6794f1a251b)

src/runtime/program/vote/execute.zig
```diff
@@ -129,10 +163,275 @@ fn intializeAccount(
...
+
+    try authorize(
+        allocator,
+        ic,
+        vote_account,
+        new_authority,
+        authorization_type,
+        clock,
+        expected_authority_keys.items,
+    );
```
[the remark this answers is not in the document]
dadepo: [a383d70](https://github.com/Syndica/sig/pull/576/commits/a383d70165053e627c8cd8283bdba6794f1a251b)

src/runtime/instruction_info.zig
```diff
@@ -85,11 +85,23 @@ pub const InstructionInfo = struct {
...
+    /// Agave https://github.com/anza-xyz/agave/blob/9eee2f66775291a1ec4c4b1be32efc1d314002f7/transaction-context/src/lib.rs#L736
+    ///
+    /// Fill in the passed buffer with all keys of signer instruction accounts in this Instruction.
+    pub fn signers(self: *const InstructionInfo, buffer: *std.ArrayList(Pubkey)) !void {
+        for (self.account_metas.constSlice()) |account_meta| {
+            if (account_meta.is_signer) {
+                try buffer.append(account_meta.pubkey);
+            }
+        }
+    }
```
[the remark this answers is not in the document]
dadepo: Given that the signers would most of the time be less than `MAX_ACCOUNT_METAS`, is there not the risk of access uninitialised portion of the bounded array. I initially considered this approach but this was the main reason I considered passing in the buffer to be filled instead.

src/runtime/instruction_info.zig
```diff
@@ -85,9 +85,8 @@ pub const InstructionInfo = struct {
         self: *const InstructionInfo,
         pubkey: Pubkey,
     ) bool {
-        for (self.account_metas.constSlice()) |account_meta| {
+        for (self.account_metas.slice()) |account_meta|
```
dadepo, line 88: Updated while resolving the merge conflict when rebasing. Will revert.

src/runtime/program/vote/state.zig
dadepo: [a383d70](https://github.com/Syndica/sig/pull/576/commits/a383d70165053e627c8cd8283bdba6794f1a251b)

src/runtime/program/vote/state.zig
dadepo: The last line was `try self.authorized_voters.insert(target_epoch, new_authorized_voter)` which was returning void. Added a line after that that returns null, which made it possible to have `?VoteError` as the return type.

[a383d70](https://github.com/Syndica/sig/pull/576/commits/a383d70165053e627c8cd8283bdba6794f1a251b)

src/runtime/instruction_info.zig
dadepo: [c4ecc1c](https://github.com/Syndica/sig/pull/576/commits/c4ecc1cbbbdd2ddae69f194deed3078f31836dd2)

src/runtime/program/vote/execute.zig
dadepo: Directly accessing the buffer or using `getAccountMetaAtIndex` which is the preference? Went with `getAccountMetaAtIndex`

src/runtime/program/vote/execute.zig
```diff
@@ -129,10 +164,273 @@ fn intializeAccount(
...
+    const clock = try ic.getSysvarWithAccountCheck(
+        Clock,
+        @intFromEnum(vote_instruction.VoteAuthorize.AccountIndex.clock_sysvar),
+    );
+
+    const authorize_pubkey = switch (vote_authorize) {
+        .voter => VoteAuthorize.voter,
+        .withdrawer => VoteAuthorize.withdrawer,
+    };
+
```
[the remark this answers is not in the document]
dadepo: This is mainly because the one used in the instruction has the account indices within it. And thought having that in `states` would be misplaced. Maybe rename the the one in instruction? something like `VoteAuthorizeInstruction` as I can see how it could be confusing...welcome to other ideas too...

src/runtime/instruction_info.zig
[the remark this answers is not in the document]
yewman: Its no more risky than using an arraylist. You should never be indexing out of bounds (current len) with either

src/runtime/program/vote/execute.zig
yewman: I prefer direct access in instances where we have just checked the number of account metas since calling getAccountMetaAtIndex does a redundant check. The orelse will never hit here since the line before we check that then are enough accounts. I would use .? if you prefer the method. Alternatively a `getAccountMetaAtIndexUnchecked` might make sense.

src/runtime/program/vote/execute.zig
```diff
@@ -129,10 +163,275 @@ fn intializeAccount(
...
+    try ic.info.signers(&signers);
+
+    try authorize(
+        allocator,
+        ic,
+        vote_account,
+        pubkey,
+        vote_authorize,
+        clock,
+        signers.items,
```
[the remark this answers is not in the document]
dadepo: [e3ef100](https://github.com/Syndica/sig/pull/576/commits/e3ef1000101284f5625ab144feb689db1cfa8409)

src/runtime/program/vote/execute.zig
```diff
@@ -129,10 +163,275 @@ fn intializeAccount(
...
+) (error{OutOfMemory} || InstructionError)!void {
+    const clock = try ic.getSysvarWithAccountCheck(
+        Clock,
+        @intFromEnum(vote_instruction.Authorize.AccountIndex.clock_sysvar),
+    );
+
+    var signers = std.ArrayList(Pubkey).init(allocator);
+    defer signers.deinit();
+
+    try ic.info.signers(&signers);
```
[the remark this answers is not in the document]
dadepo: [e3ef100](https://github.com/Syndica/sig/pull/576/commits/e3ef1000101284f5625ab144feb689db1cfa8409)

src/runtime/instruction_info.zig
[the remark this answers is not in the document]
dadepo: [e3ef100](https://github.com/Syndica/sig/pull/576/commits/e3ef1000101284f5625ab144feb689db1cfa8409)

src/runtime/program/vote/execute.zig
dadepo: > Alternatively a getAccountMetaAtIndexUnchecked might make sense.

True. Went with direct access for now [c2fb042](https://github.com/Syndica/sig/pull/576/commits/c2fb042154244f52150c5a331d3285ed83793954) such a method can be added later

src/runtime/program/vote/state.zig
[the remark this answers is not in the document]
dadepo: [c330747](https://github.com/Syndica/sig/pull/576/commits/c3307472f55ebaa18b83f17d4698372a31991c02)

src/runtime/program/vote/execute.zig
```diff
@@ -129,10 +163,283 @@ fn intializeAccount(
...
+            // current authorized withdrawer must say "yay".
+            const authorized_withdrawer_signer = blk: {
+                validateIsSigner(
+                    vote_state.authorized_withdrawer,
+                    signers,
+                ) catch {
+                    break :blk false;
+                };
+                break :blk true;
+            };
```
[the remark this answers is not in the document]
dadepo: `std.meta.isError` that's a handy one! Similar to `the is_ok()`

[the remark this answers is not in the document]
yewman: Yeah it is nice :)!

src/runtime/program/vote/execute.zig
```diff
@@ -129,10 +163,270 @@ fn intializeAccount(
...
+            if (maybe_err) |err| {
+                ic.tc.custom_error = @intFromEnum(err);
+                return InstructionError.Custom;
+            }
+        },
+        .withdrawer => {
+            // current authorized withdrawer must say "yay".
+            const authorized_withdrawer_signer = blk: {
+                validateIsSigner(
+                    vote_state.authorized_withdrawer,
```
yewman, line 259: One last thing with this section, could you permalink the different blocks (permalink above each block of equivalent agave code) so that it is easy to understand the difference in control flow.


- above 234
    - https://github.com/anza-xyz/agave/blob/01e50dc39bde9a37a9f15d64069459fe7502ec3e/programs/vote/src/vote_state/mod.rs#L697-L701
- above 238
    - https://github.com/anza-xyz/solana-sdk/blob/4e30766b8d327f0191df6490e48d9ef521956495/vote-interface/src/state/mod.rs#L872
- above 244
    - https://github.com/anza-xyz/agave/blob/01e50dc39bde9a37a9f15d64069459fe7502ec3e/programs/vote/src/vote_state/mod.rs#L701-L709
    - https://github.com/anza-xyz/agave/blob/01e50dc39bde9a37a9f15d64069459fe7502ec3e/programs/vote/src/vote_state/mod.rs#L701-L709

src/runtime/program/vote/execute.zig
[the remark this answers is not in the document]
yewman: missed the one in .withdrawer branch

src/runtime/program/vote/execute.zig
dadepo: [df457b3](https://github.com/Syndica/sig/pull/576/commits/df457b3ebe69cf574445a1bed72e4d8dc8e8359a)

src/runtime/program/vote/instruction.zig
```diff
@@ -26,6 +26,71 @@ pub const IntializeAccount = struct {
...
+        account = 0,
+        /// `[]` Clock sysvar
+        clock_sysvar = 1,
+        /// `[SIGNER]` Vote or withdraw authority
+        current_authority = 2,
+    };
+};
+
+pub const VoteAuthorizeWithSeedArgs = struct {
+    authorization_type: sig.runtime.program.vote_program.state.VoteAuthorize,
```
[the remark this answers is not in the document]
dadepo: [531aad8](https://github.com/Syndica/sig/pull/576/commits/531aad8b48e52eee87311098f5bb89096af3c50c)

src/runtime/program/vote/execute.zig
dadepo: Fixing the std.meta.isError revert

dadepo: [3e916a8](https://github.com/Syndica/sig/pull/576/commits/3e916a89b09b0d53225b8acc369cd47d76a4c207)

src/runtime/program/vote/execute.zig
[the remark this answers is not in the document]
dadepo: [3e916a8](https://github.com/Syndica/sig/pull/576/commits/3e916a89b09b0d53225b8acc369cd47d76a4c207)

yewman (approved):
lgtm :)

Rexicon226 (dismissed):
Looks good to merge.

PR #581, opened 2025-02-26
fix(socket): use XevThread for MacOS instead of PerThread

Rexicon226:
There's no blocking `sento` on BSD, so to throttle the buffer usage we're switching to the `XevThread` socket backend, otherwise, when the throughput is too high, we will drop packets with `NOBUFS`.

### DIFF
--- a/src/net/socket_utils.zig
+++ b/src/net/socket_utils.zig
@@ -207,6 +207,7 @@ const XevThread = struct {
             .cancelled => unreachable, // only observable in onSend/onRecv callback.
         }
 
+        const addr = node.data.packet.addr;
         switch (st.direction) {
             .sender => {
                 node.data.packet = st.channel.tryReceive() orelse return;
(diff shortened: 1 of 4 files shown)


0xNineteen:
would be good to rebase - PR details seem out of date

src/net/socket_utils.zig
```diff
@@ -79,6 +79,7 @@ const XevThread = struct {
 
         // Create a node for the SocketThread
         const node = try st.allocator.create(List.Node);
+        defer st.allocator.destroy(node);
```
kprotty, line 82: The node lives independently of the SocketThread state and must stay alive. When the SocketThread's exit condition is set, the node will free itself and wake the handle waited on in `join()` below.

src/net/socket_utils.zig
```diff
@@ -420,7 +420,14 @@ const PerThread = struct {
 };
 
 // TODO: Evaluate when XevThread socket backend is beneficial.
-const SocketBackend = PerThread;
+const SocketBackend = switch (builtin.os.tag) {
+    // BSD doesn't have a blocking `sendto`, which means that if our
```
kprotty, line 424: Thoughts on catching NOBUFS on `sendto` then calling `posix.poll({fd, writable}, -1)` and retrying (vs switching to XevThread) ?

Rexicon226: Sure, both work. You want to take this over? I'm kinda focused on the conformance stuff.

src/net/socket_utils.zig
Rexicon226: Yeah, I realized that immediately but had to hop off. Ditto my question below.

src/net/socket_utils.zig
kprotty: Sure

kprotty (approved):

Rexicon226:
The MacOS CI doesn't track coverage, so covering the logic added in this PR is impossible for now. 
I've tested it locally and confirmed that it fixes the bug.

PR #585, opened 2025-02-28
fix(svm): more conformance!

Rexicon226:
After a grueling day in the fuzzing mines, I've returned with fixes o' plenty.


yewman (approved):
lgtm :)

PR #593, opened 2025-02-28
refactor(rpc-client): redesign RPC client

dnut:
I redesigned the RPC client to make the code more organized and composable, without adding unnecessary abstractions such as interfaces or generic types. The RPC client struct itself is now simplified, and only serves to glue together the code from other modules: `methods`, `http`, `request`, and `response`.

## Motivation

My primary motivation for these changes was to improve the organization and readability of the code. Each of these domains (networking, serialization, solana data model) are disjoint and deal with a different set of responsibilities. I feel the code is easier to understand when they are clearly delineated in separate modules.

An extra benefit of this organization is that the RPC client is more composable and flexible. You can easily combine these components, or provide custom implementations for them, to implement whatever kind of RPC client you would like in a few lines of code. The provided RPC client glues together these pieces for ease of use.

A common use case that I've needed in the past when I was writing trading bots was to blast out many rpc requests and then collect the responses asynchronously. This required some ugly hacks using the rust rpc client. With my new approach here, use cases like that are much easier to support through composition of a custom http client.

## Approach

RPC Methods, from the perspective of an RPC client, are just data. They are request and response data types that need to be converted back and forth from json. So I've described that fully with the type system. Now there is `methods.zig` with a struct for each method. These structs can be used in the client and server.

I also extracted serialization and HTTP connectivity into their own modules. Serialization is implemented in `request.zig` and `response.zig`. The HTTP logic was moved into `HttpPostFetcher` which is defined in `http.zig`.

Thus in client.zig we have the full implementation of an rpc client in the `fetchRpc` function. It runs these steps in 5 lines of code:
1. serialize the request as json
2. send json in an http post message and await the response
3. deserialize the json as the response type

While `fetchRpc` sufficiently implements an rpc client, a more opinionated `Client` struct is also provided to improve its ergonomics. It handles the following:
- initializes and contains an HttpPostFetcher
- defines two fetch methods: one to use the contained allocator, and one to use a custom allocator
- defines a helper method for each RPC method so you can call them like `client.getSlot(.{})`

### Request
I was able to simplify the code by removing the `Request` data type. It is no longer necessary since request structs can be directly serialized.

### Response

I reorganized the Response data structure to be more intuitive and safe. It no longer has intermediary states where the struct is partially invalid. You just init it from the json to get a struct that's more neatly organized than the old version.

### Tests

I removed the dependency on testnet from the rpc tests. The existing tests were validating three things:
1. serialization and deserialization works
5. we're able to connect to a testnet rpc node with http requests
6. the testnet rpc node associates our request data type with the same response data type that we do.

I converted the tests to purely validate `1`. Validations `2` and `3` are no longer being tested because it's not appropriate to test these in a unit test. This is why the tests previously were skipped. So I've actually increased coverage by unskipping the tests. We could also add tests similar to `2` that connect HttpPostFetcher to a *local* server, but I think that's overkill, since the http client comes from `std` and involving the network in unit tests is a headache and leads to flaky tests.


0xNineteen (approved):
lgtm :shipit:

yewman (approved):
Looks good to me, my only thought is whether the method structs could be be defined as something like
```zig
pub const Method = struct {
    pub const Request = struct {
        ...
        pub const Config = struct {
            ...
        };
    }

    pub const Response = struct {
         ...
    };
};
```
Not sure if its the best approach just throwing it out there. In my mind it just makes it very clear to a reader what the `Request` vs. `Response`. It is still pretty clear with the current approach though so happy for you to resolve this comment and merge either approach :)

dnut:
@yewman I thought I might get some feedback on this topic since I myself had a hard time deciding what approach to use here. I made the current choice with awareness of your suggestion as an option. I am open to changing this though. Apologies for the long message. If at any point this feels like a waste of time, we can just merge it as is. Otherwise I think we should look at the full picture to make a decision.

------------------

I agree that the Request/Response terminology helps provide some clarity. The problem is that it doesn't actually work with the way the code is currently written. Of course, we could change it to support a different design.

Somehow, we need to associate request and response types with each other. There are many ways of doing this. Some approaches can fully describe it with the type system, and other approaches rely on other contexts to describe the relationships between request and response types.

Describe the relationship outside the type system:
1. Define a function for each rpc method. *(currently supported for ease of use)*
2. Every time you call `fetch`, pass in both the request *and* the response type.

Describe the relationship fully within the type system:

3. Put a declaration inside the request type that specifies the response type. *(current approach to `fetch`)*
4. Define something else that describes the mapping from Request to Response. *(your proposal is a variant of this)*

## Outside type system
## 1. Define a function for each rpc method

We had this already, and I've also included it here in response to feedback. I'm happy to support this behavior to improve the ergonomics of the most common usage of an RPC client. But it doesn't need to be the only way that the types are correlated. I like the idea of having the type system speak for itself if it can be done cleanly, to improve both readability and flexibility.

### 2. Every time you call `fetch`, pass in both the request *and* the response type.

This is the ultimate in flexibility but it feels like overkill. It has the worst ergonomics, and it is more prone to misuse. I see this as the worst option.

## Within type system
### 3. Put a declaration inside the request type that specifies the response type.

This is currently required for how the `fetch` functions I've written work. They determine the return type by looking for a Response decl in the request type. This more ergonomic than `2` because you don't need to juggle types. The correct response type is implied by the request type.

This could also be implemented in a way that looks a lot like your code, just with a small addition.
```zig
pub const Method = struct {
    pub const Request = struct {
        ...
        pub const Config = struct {
            ...
        };

        pub const Response = Method.Response;  // <--- see here
    }

    pub const Response = struct {
         ...
    };
};
```

I think that would be OK, but I took my current approach because the code is simpler and more direct, and because I think the method itself is actually a request.

For example, consider "GetAccountInfo." Grammatically, the name is a command, which already implies that it is a request. Beyond what the name implies, it's asking for an account info that:
- has a particular pubkey
- is from the time specified by the config
- is provided in a particular response format

I think it's accurate to think of the response as an element in the request's list of demands. In general, there is a temporal asymmetry in the request/response pattern. You start with a particular request, and from that request you can derive a response. A response makes no sense in a vacuum, it's always derived from a request.

My point here is that each approach is reasonable in a way, not that one is objectively better.

### 4. Define something else that describes the mapping from Request to Response

The idea here is that you have a mapping defined somewhere outside the request and the response, which maps them together. This is what you've written. There are other ways to do this that could decouple things further, but I think that would be a downgrade compared to your suggestion.

This could fit into the `fetch` paradigm with some minor adjustments. An additional parameter would be added, and the types would be adjusted a bit. So where I currently have this:
```zig
pub fn fetchRpc(
    fetcher: *HttpPostFetcher,
    allocator: Allocator,
    request: anytype,
) FetchRpcError!Response(@TypeOf(request)) {
```
It could instead be something like this:
```zig
pub fn fetchRpc(
    fetcher: *HttpPostFetcher,
    allocator: Allocator,
    method: type
    request: method.Request,
) FetchRpcError!Response(method) {
```
I decided against this since it is less ergonomic in some cases. I think of the requests themselves as just being messages, and I'd like to be able to pass around those messages flexibly. You may not always want to initialize the struct on the same line of code that you're passing it into the fetch function

For example, you might want to construct a message in one place, and pass it into your rpc client somewhere else, without needing that "somewhere else" to specifically include the name of the message in the source code. This is easy to support with the current approach. the `fetch` functions infer the type at comptime for you. You only need to pass in the message. But with this other approach, you need to specify both the method and the request. It's doable, but more cumbersome to handle two items at once.

------------------

So, these are all the tradeoffs I considered when I was writing this code and settled on the current approach. I'm curious to hear your thoughts on this.

0xNineteen:
gonna merge it now but we can continue discussion of future changes/improvements here or offline too

PR #595, opened 2025-03-01
test: no global registry in tests

dnut:
It's a good practice to avoid sharing global state across tests. It's easy to initialize a registry, so we don't need to lean on globalRegistry to provide a registry to tests that need one.


Sobeston (approved):
Don't see why not, seems good :+1:

PR #596, opened 2025-03-03
Add tests for PeekablebReader

Raphexion:
This commit adds some unit-tests for the PeekableReader in utils/io.zig.

PR #600, opened 2025-03-04
Add tests for lazy

Raphexion:
Add some more tests for some utils modules. Including a test that was missing from #596 (test a double peek).

While writing the unit-tests for Lazy, I corrected a small issue where the final result was not actually returned. Most likely this was dormant in the code because only `Lazy(void)` is currently used.

### DIFF
--- a/src/utils/io.zig
+++ b/src/utils/io.zig
@@ -188,6 +188,14 @@ test PeekableReader {
         try testing.expect(try reader.peekByte() == 'a');
     }
 
+    // double peek
+    {
+        var stream = std.io.fixedBufferStream("abcdef");
+        var reader = peekableReader(stream.reader());
+        try testing.expect(try reader.peekByte() == 'a');
+        try testing.expect(try reader.peekByte() == 'a');
+    }
+
     // read empty data
     {
         var stream = std.io.fixedBufferStream("");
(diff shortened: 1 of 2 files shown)


dnut (commented):
Thanks for the contribution. Nice bugfix, and additional test coverage is always welcome. I'm curious about some style choices though.

Why do you combine multiple tests into a single test? Normally I would use a separate test block for each of these so the test runner has more information. Something like this:

```zig
test "Lazy void example" {...}
test "Lazy void multiple arguments" {...}
test "Lazy non-void return type" {...}
```

Is there a specific advantage to running these all in a single test?

Also, using underscore prefixes is something we've had some internal debate about because its meaning is ambiguous which can result in confusion or misuse. We don't outright ban the pattern, but we caution against it in the style guide. I think using the word "test" could improve clarity here, for example `fn test_set` or `const tests = struct { fn set() };` What do you think?

Raphexion:
@dnut Thank you for your review and feedback. I think you are correct in both of your observations. I've split the "combined test" into three separate test cases. It is much better. Both for readability and for seeing individual test results :+1: :sun_with_face:

Moreover, I good suggestion with removing the underscore. I felt that it polluted the namespace. But I agree with your point. It is unnecessary and just adds noise.

0xNineteen (approved):
lgtm - thanks for the contribution! :D

PR #602, opened 2025-03-05
fix(accounts-db): reduce bufferpool reads

0xNineteen:
replaces two bufferpool reads in `getAccountHashAndLamports` with a single call


Sobeston (commented):
Seems very reasonable, just want to see some numbers before approving

Sobeston (approved):
:shipit:

PR #608, opened 2025-03-07
fix(prometheus): flaky metrics server

dnut:
## Problems

Prometheus does not consistently get metrics from our metrics endpoint. The metrics are intermittent.

image (2)

Also, the server often crashes due to errors. In #590 I added a wrapper to catch these errors and restart the server. But sometimes an error occurs that is handled, a server restart is attempted, but the server never responds to any http requests any more, so the metrics become completely inaccessible until sig is completely restarted. Usually this is caused by `error.BrokenPipe`

## Solution

> Revert "fix(prometheus): remove httpz again and fix prometheus metrics (#555)"
> 
> This reverts commit 3502333aa0798734dd0b3384f9d41e47078a495a.

I understand the desire to eliminate unnecessary dependencies. In the grand scheme of things, eliminating httpz is not that important. Having a working metrics endpoint is critical though. We have a solution that works with httpz. We should test more thoroughly before replacing it.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -89,6 +89,9 @@ pub fn build(b: *Build) !void {
     const zig_cli_dep = b.dependency("zig-cli", dep_opts);
     const zig_cli_mod = zig_cli_dep.module("zig-cli");
 
+    const httpz_dep = b.dependency("httpz", dep_opts);
+    const httpz_mod = httpz_dep.module("httpz");
+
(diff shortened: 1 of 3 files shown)


0xNineteen (approved):

0xNineteen:
if our zig server implementation is flaky rn we should keep this in mind when were using it for rpc (and if we should take a different approach there too)

PR #611, opened 2025-03-10
feat(vote-program): Update validator identity instruction

dadepo:
Implements the insturction to update a node's identity, ie change the `node_pubkey`


src/runtime/program/vote/execute.zig
```diff
@@ -428,6 +429,61 @@ fn executeAuthorizeChecked(
...
+        VoteStateVersions,
+    );
+
+    var vote_state = versioned_state.convertToCurrent(allocator) catch {
+        // TODO okay to convert out of memory to InvalidAccountData?
+        return InstructionError.InvalidAccountData;
+    };
+    defer vote_state.deinit();
+
+    // current authorized withdrawer must say "yay"
```
0xNineteen, line 473: can we change these comments to something more descriptive instead of "must say 'yay'" XD -- tbh maybe we dont even need the comments its pretty clear your checking the signer

src/runtime/program/vote/execute.zig
```diff
@@ -428,6 +429,61 @@ fn executeAuthorizeChecked(
...
+    new_identity: Pubkey,
+) InstructionError!void {
+    const versioned_state = try vote_account.deserializeFromAccountData(
+        allocator,
+        VoteStateVersions,
+    );
+
+    var vote_state = versioned_state.convertToCurrent(allocator) catch {
+        // TODO okay to convert out of memory to InvalidAccountData?
+        return InstructionError.InvalidAccountData;
```
0xNineteen, line 469: should fix this

0xNineteen (changes requested):
few small things otherwise lgtm

src/runtime/program/vote/execute.zig
dadepo: I agree. it was blind copy from Agave. Removed instead [bc11aa3](https://github.com/Syndica/sig/pull/611/commits/bc11aa3e37295871544527b8eefbacf7a04707da)

src/runtime/program/vote/execute.zig
dadepo: [6b84b02](https://github.com/Syndica/sig/pull/611/commits/6b84b02e3962fe3bc35439ef2247851469eebeec)

src/runtime/program/vote/execute.zig
dadepo: Brought the comment back and used one `if` [f176961](https://github.com/Syndica/sig/pull/611/commits/f1769618697d46e38a8d4a5bad427cd1072c3edd)

src/runtime/program/vote/execute.zig
```diff
@@ -428,6 +429,55 @@ fn executeAuthorizeChecked(
...
+
+    var vote_state = try versioned_state.convertToCurrent(allocator);
+    defer vote_state.deinit();
+
+    // Both the current authorized withdrawer and new identity must sign.
+    if (!(ic.info.isPubkeySigner(vote_state.authorized_withdrawer) and
+        ic.info.isPubkeySigner(new_identity)))
+    {
+        return InstructionError.MissingRequiredSignature;
+    }
```
0xNineteen, line 475: small nit : feels like we can make this much easier to read like the following:

```suggestion        
        const withdrawer_signed = ic.info.isPubkeySigner(vote_state.authorized_withdrawer)
        const new_identity_signed = ic.info.isPubkeySigner(new_identity)
        // Both the current authorized withdrawer and new identity must sign.
        const has_required_sigs = withdrawer_signed and new_identity_signed
        if (!has_required_sigs) {
            return InstructionError.MissingRequiredSignature;
        }
```

also wdyt about renaming the field on vote_state from `authorized_withdrawer` to just `withdrawer`?

0xNineteen (changes requested):

src/runtime/program/vote/execute.zig
dadepo: Applied the variable binding suggestions [4b1dedc](https://github.com/Syndica/sig/pull/611/commits/4b1dedc7e541f361cfad83e46a4f6316d6aa9dd9)

> also wdyt about renaming the field on vote_state from authorized_withdrawer to just withdrawer?

I think that makes sense. The ` authorized_` is definitely redundant. But I think it would be cleaner to rename in a separate PR. There is also an `authorized_voters` field too that can be renamed to just `voters`. Added a TODO for this.

0xNineteen (dismissed):

dadepo:
Closing this. It would be reviewed in https://github.com/Syndica/sig/pull/616

PR #618, opened 2025-03-17
refactor(runtime): change system_program error to enum

yewman:
Native program errors should be enums, or implement a mapError method to map the error to the correct int according to Agave's analogous enum definition. If try semantics are required, using a map function with a switch may be the best option. Otherwise an enum is simpler.

### DIFF
--- a/src/runtime/program/system_program/error.zig
+++ b/src/runtime/program/system_program/error.zig
@@ -1,5 +1,5 @@
 /// [agave] https://github.com/solana-program/system/blob/6185b40460c3e7bf8badf46626c60f4e246eb422/interface/src/error.rs#L12
-pub const Error = error{
+pub const Error = enum(u8) {
     /// An account with the same address already exists.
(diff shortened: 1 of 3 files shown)


Rexicon226 (dismissed):

Rexicon226 (approved):
We can merge without full test coverage, as those areas require a lot of boilerplate to hit. Will revisit later to improve.

src/runtime/program/system_program/error.zig
```diff
@@ -1,5 +1,5 @@
 /// [agave] https://github.com/solana-program/system/blob/6185b40460c3e7bf8badf46626c60f4e246eb422/interface/src/error.rs#L12
-pub const Error = error{
+pub const Error = enum(u8) {
     /// An account with the same address already exists.
     AccountAlreadyInUse,
```
dadepo, line 4: With the switch to enum, I think the variants should now be snake_case

PR #626, opened 2025-03-19
fix(consensus): remove unecassary `@constCast` usage

Rexicon226:
Using `@constCast` is a very large smell, and is unnecessary here.


yewman (approved):
Makes sense

PR #631, opened 2025-03-21
feat(runtime): Implement parameter serialisation

yewman:
Implements basic serialisation and deserialisation of parameters for BPF execution. 
Some testing is implemented, more testing will be added in follow up BPF execution PR.

### DIFF
--- a/src/runtime/program/serialization.zig
+++ b/src/runtime/program/serialization.zig
@@ -1,0 +1,535 @@
+const std = @import("std");
+const sig = @import("../../sig.zig");
+
+const ids = sig.runtime.ids;
+const program = sig.runtime.program;
+const vm = sig.vm;
+
+const Pubkey = sig.core.Pubkey;
+const InstructionError = sig.core.instruction.InstructionError;
+
+const InstructionContext = sig.runtime.InstructionContext;
+const BorrowedAccount = sig.runtime.BorrowedAccount;
+
+const Region = vm.memory.Region;
+
+/// Solana BPF version flag
+pub const EF_SBPF_V2: u32 = 0x20;
+/// Maximum number of instructions in an eBPF program.
+pub const PROG_MAX_INSNS: usize = 65_536;
+/// Size of an eBPF instructions, in bytes.
+pub const INSN_SIZE: usize = 8;
+/// Frame pointer register
+pub const FRAME_PTR_REG: usize = 10;
+/// First scratch register
+pub const FIRST_SCRATCH_REG: usize = 6;
+/// Number of scratch registers
+pub const SCRATCH_REGS: usize = 4;
+/// Alignment of the memory regions in host address space in bytes
+pub const HOST_ALIGN: usize = 16;
+/// Upper half of a pointer is the region index, lower half the virtual address inside that region.
+pub const VIRTUAL_ADDRESS_BITS: usize = 32;
+
+/// Size (and alignment) of a memory region
+pub const MM_REGION_SIZE: u64 = 1 << VIRTUAL_ADDRESS_BITS;
+/// Virtual address of the bytecode region (in SBPFv3)
+pub const MM_BYTECODE_START: u64 = 0;
+/// Virtual address of the readonly data region (also contains the bytecode until SBPFv3)
+pub const MM_RODATA_START: u64 = MM_REGION_SIZE;
+/// Virtual address of the stack region
+pub const MM_STACK_START: u64 = MM_REGION_SIZE * 2;
+/// Virtual address of the heap region
+pub const MM_HEAP_START: u64 = MM_REGION_SIZE * 3;
+/// Virtual address of the input region
+pub const MM_INPUT_START: u64 = MM_REGION_SIZE * 4;
+
+/// `assert_eq(std::mem::align_of::<u128>(), 8)` is true for BPF but not for some host machines
+pub const BPF_ALIGN_OF_U128: usize = 8;
+
+/// [agave] https://github.com/anza-xyz/solana-sdk/blob/e1554f4067329a0dcf5035120ec6a06275d3b9ec/account-info/src/lib.rs#L17-L18
+pub const MAX_PERMITTED_DATA_INCREASE: usize = 1_024 * 10;
+
+pub const SerializedAccount = union(enum) {
+    account: struct { u16, BorrowedAccount },
+    duplicate: u8,
+};
+
+pub const SerializedAccountMeta = struct {
+    original_data_len: usize,
+    vm_data_addr: u64,
+    vm_key_addr: u64,
+    vm_lamports_addr: u64,
+    vm_owner_addr: u64,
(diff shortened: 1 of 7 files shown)


0xNineteen (commented):
lgtm

src/runtime/program/serialization.zig
```diff
@@ -0,0 +1,964 @@
...
+
+    /// [agave] https://github.com/anza-xyz/agave/blob/01e50dc39bde9a37a9f15d64069459fe7502ec3e/program-runtime/src/serialization.rs#L56-L57
+    pub fn write(self: *Serializer, comptime T: type, value: T) u64 {
+        const vaddr = self.vaddr +| self.buffer.items.len -| self.region_start;
+        self.buffer.appendSliceAssumeCapacity(std.mem.asBytes(&value));
+        return vaddr;
+    }
+
+    /// [agave] https://github.com/anza-xyz/agave/blob/01e50dc39bde9a37a9f15d64069459fe7502ec3e/program-runtime/src/serialization.rs#L77-L78
+    pub fn writeAll(self: *Serializer, data: []const u8) u64 {
```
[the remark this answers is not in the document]
Rexicon226: > can this method just call self.write([]const u8, data)?

Serializing the `[]const u8` makes no sense.

It should be called `writeBytes` in my opinion. If anything `write` should be calling `writeBytes`.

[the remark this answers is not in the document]
0xNineteen: im cool with that

src/runtime/program/serialization.zig
```diff
@@ -0,0 +1,964 @@
...
+                allocator.free(account_metas);
+            }
+
+            const serialized_regions = try concatRegions(allocator, regions);
+            defer allocator.free(serialized_regions);
+            if (copy_account_data) {
+                try std.testing.expectEqualSlices(u8, memory, serialized_regions);
+            }
+
+            // TODO: deserialize and compare
```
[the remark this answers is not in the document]
yewman: Yeah I would like to leave this, I have added a more informative todo. Basically we want to make sure that this method and our entrypoint deserialize method produce the same result

src/runtime/program/serialization.zig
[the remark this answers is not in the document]
yewman: is this what you are thinking  01048582

src/runtime/program/serialization.zig
```diff
@@ -0,0 +1,967 @@
...
+                BPF_ALIGN_OF_U128,
+            ) - account.constAccountData().len;
+
+            if (self.copy_account_data) {
+                self.buffer.appendNTimes(
+                    self.allocator,
+                    0,
+                    MAX_PERMITTED_DATA_INCREASE + align_offset,
+                ) catch {
+                    return InstructionError.InvalidArgument;
```
Rexicon226, line 112: The only way these could error is if there's an out-of-memory error. I don't think `InvalidArgument` is correct here, it should bubble up `OutOfMemory`.

src/runtime/program/serialization.zig
```diff
@@ -0,0 +1,967 @@
...
+        // TODO: Cow https://github.com/anza-xyz/agave/blob/01e50dc39bde9a37a9f15d64069459fe7502ec3e/program-runtime/src/serialization.rs#L134
+        if (account.constAccountData().len > 0) {
+            const state = getAccountDataRegionMemoryState(account);
+            const region = switch (state) {
+                inline .constant, .mutable => |s| Region.init(
+                    s,
+                    if (s == .constant)
+                        account.constAccountData()
+                    else
+                        try account.mutableAccountData(),
```
Rexicon226, line 161: What do you think about unifying these functions and instead have `account.accountData(.constant)`?

src/runtime/program/serialization.zig
```diff
@@ -0,0 +1,967 @@
...
+
+            start += @sizeOf(u8) // is_signer
+            + @sizeOf(u8) // is_writable
+            + @sizeOf(u8) // executable
+            + @sizeOf(u32) // original data len
+            + @sizeOf(Pubkey); // key
+
+            // read owner
+            if (start + @sizeOf(Pubkey) > memory.len) return InstructionError.InvalidArgument;
+            const owner = memory[start .. start + @sizeOf(Pubkey)];
```
Rexicon226, line 609:
```suggestion
            const owner = memory[start ..][0..@sizeOf(Pubkey)];
```
Ditto for the other usages.

src/runtime/program/serialization.zig
```diff
@@ -0,0 +1,967 @@
...
+                    post_account.account.rent_epoch,
+                );
+            }
+        }
+    }
+}
+
+fn concatRegions(allocator: std.mem.Allocator, regions: []Region) ![]u8 {
+    if (!builtin.is_test) {
+        @panic("concatRegions should only be called in test mode");
```
Rexicon226, line 955: This needs to be a `compileError`, otherwise you'll have a runtime panic.

Rexicon226 (changes requested):
Looks good, a couple notes.

src/runtime/program/serialization.zig
yewman: bde9467a

src/runtime/program/serialization.zig
yewman: bde9467a

src/runtime/program/serialization.zig
yewman: bde9467a

src/runtime/program/serialization.zig
yewman: I am okay with this, although I would rather not import MemoryState from the vm. Instead I have just defined an enum in the function arg which I think is fine, and perhaps preferable in this context. Lmk your thoughts. 
ee2e3078

src/runtime/program/serialization.zig
Rexicon226: And the one underneath it as well!

src/runtime/program/serialization.zig
Rexicon226: Then move the MemoryState enum out of the VM. I see no benefit in having two separate enums. Using the same one would allow you to simplify the code to just 
```zig
            const region = switch (state) {
                inline .constant, .mutable => |s| Region.init(
                    s,
                    account.getData(s),
                    self.vaddr,
                ),
            };
```

yewman: Yeah thats fair. On second thought I think we should just leave the const and mutable methods since the MemoryState enum feels very vm specific and I don't think it is needed outside that context. If down the line we find there are more instances like the above then perhaps we can reconsider.

src/runtime/program/serialization.zig
yewman: ooops 😆 
ddaed0e1

Rexicon226 (approved):

PR #632, opened 2025-03-21
feat(cli): add ability to specify positional arguments

InKryption:
todo: generate help output for positional arguments


Rexicon226 (approved):
lgtm, the XOR for positionals and subcmds still seems like the best solution to me

PR #638, opened 2025-03-26
feat(svm); start adding ELF comments

Rexicon226:


src/vm/elf.zig
```diff
@@ -219,39 +244,50 @@ pub const Elf = struct {
             } else return error.InvalidDynamicSectionTable;
         }
 
+        /// [agave] https://github.com/anza-xyz/sbpf/blob/615f120f70d3ef387aab304c5cdf66ad32dae194/src/elf.rs#L827-L1002
         fn parseRoSections(
             self: Data,
             headers: Headers,
             config: Config,
             version: sbpf.Version,
             allocator: std.mem.Allocator,
```
yewman, line 253: can we modify this to stick to the Sig convention of allocators being the first argument (other than self of course)

src/vm/elf.zig
```diff
@@ -116,8 +127,8 @@ pub const Elf = struct {
 
         fn parse(headers: Headers) !Data {
             const strtab = try headers.shdrSlice(headers.header.e_shstrndx);
-
             for (headers.shdrs) |shdr| {
+                // Check that none of the section names are longer than the sting table containing them.
```
yewman, line 131:
```suggestion
                // Check that none of the section names are longer than the string table containing them.
```

yewman (changes requested):

src/vm/elf.zig
Rexicon226: e56fba01

src/vm/elf.zig
Rexicon226: e56fba01

yewman: And this one please 🙏 
 https://github.com/Syndica/sig/blob/e56fba010ea113c24db1ef63349f39f0a14cd13c/src/vm/elf.zig#L406

yewman (approved):
Nice improvement

PR #639, opened 2025-03-26
feat(vm): unaligned memory map

Sobeston:
- Implemented unaligned memory maps
- Reworked the memory map code to make errors match agave
- Testing

### DIFF
--- a/src/vm/memory.zig
+++ b/src/vm/memory.zig
@@ -1,5 +1,6 @@
 const std = @import("std");
 const sbpf = @import("sbpf.zig");
+const exe = @import("executable.zig");
 
 /// Virtual address of the bytecode region (in SBPFv3)
 pub const BYTECODE_START: u64 = 0x000000000;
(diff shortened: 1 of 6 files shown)


yewman (changes requested):
Looks good! Would you mind adding permalinks to the rest of the memory.zig module as you have for new implementations?

src/vm/memory.zig
```diff
@@ -13,32 +14,74 @@ pub const HEAP_START: u64 = 0x300000000;
...
-        return .{ .aligned = try AlignedMemoryMap.init(regions, version) };
+    pub fn deinit(self: *const MemoryMap, allocator: std.mem.Allocator) void {
+        switch (self.*) {
+            .aligned => {},
+            .unaligned => |unaligned| unaligned.deinit(allocator),
+        }
     }
 
-    pub fn region(self: MemoryMap, vm_addr: u64) !Region {
+    fn find_region(
```
0xNineteen, line 45: should be camelCase

src/vm/memory.zig
```diff
@@ -85,17 +129,11 @@ pub const Region = struct {
...
-                .mutable => |mutable| mutable,
-            },
-            .mutable => switch (self.host_memory) {
-                .constant => return error.AccessViolation,
-                .mutable => |mutable| mutable,
-            },
+    /// Returns null if you're trying to get mutable access to a constant region.
+    pub fn hostSlice(self: Region, comptime state: MemoryState) ?state.Slice() {
+        return switch (self.host_memory) {
+            .constant => |constant| if (state == .mutable) null else constant,
```
0xNineteen, line 135: feels like this null case should be an error -- wdyt

src/vm/memory.zig
```diff
@@ -104,71 +142,232 @@ pub const Region = struct {
         comptime state: MemoryState,
         vm_addr: u64,
         len: u64,
-    ) !state.Slice() {
-        if (vm_addr < self.vm_addr_start) return error.InvalidVirtualAddress;
+    ) ?state.Slice() {
+        if (vm_addr < self.vm_addr_start) return null;
```
0xNineteen, line 146: why change this from an error to null?

src/vm/memory.zig
```diff
@@ -104,71 +142,232 @@ pub const Region = struct {
...
-
         return .{
             .regions = regions,
             .version = version,
+            .config = config,
         };
     }
 
-    fn region(self: *const AlignedMemoryMap, vm_addr: u64) !Region {
+    fn find_region(self: *const AlignedMemoryMap, vm_addr: u64) ?Region {
```
0xNineteen, line 202: here too

src/vm/memory.zig
```diff
@@ -104,71 +142,232 @@ pub const Region = struct {
...
+    fn deinit(self: *const UnalignedMemoryMap, allocator: std.mem.Allocator) void {
+        allocator.free(self.regions);
+        allocator.free(self.region_addresses);
+    }
+
+    // [agave] https://github.com/anza-xyz/sbpf/blob/a8247dd30714ef286d26179771724b91b199151b/src/memory_region.rs#L218
+    fn constructEytzingerOrder(
+        self: UnalignedMemoryMap,
+        ascending_regions: []Region,
+        _in_index: usize,
```
0xNineteen, line 302: is there no better var names for the vars with the _ prefix? code can become hard to read with too many

src/vm/memory.zig
```diff
@@ -104,71 +142,232 @@ pub const Region = struct {
...
+        return self.constructEytzingerOrder(
+            ascending_regions,
+            in_index +| 1,
+            (out_index *| 2) +| 2,
+        );
+    }
+
+    // [agave] https://github.com/anza-xyz/sbpf/blob/a8247dd30714ef286d26179771724b91b199151b/src/memory_region.rs#L293
+    fn find_region(self: *const UnalignedMemoryMap, vm_addr: u64) ?Region {
+        // NOTE: agave-like cache unimplemented. Does not seem necessary.
```
0xNineteen, line 330:
```suggestion
    // [agave] https://github.com/anza-xyz/sbpf/blob/a8247dd30714ef286d26179771724b91b199151b/src/memory_region.rs#L293
    // NOTE: agave-like cache unimplemented. Does not seem necessary.
    fn find_region(self: *const UnalignedMemoryMap, vm_addr: u64) ?Region {
```

Sobeston:
note: the region methods should continue to return nulls, as they do not have the version+config required to return the correct error

src/vm/memory.zig
Sobeston: https://github.com/Syndica/sig/pull/639/commits/e2a0e3c58e0f908854d59b9212eca28bc5f1bd0e#diff-23e4e8b8c3b0ebce6d44fff7c7ff93546043dcee6a94c6c2b54d5482d0457468R334-R339 moved the comment here

src/vm/memory.zig
Sobeston: renamed using in_index and new_in_index https://github.com/Syndica/sig/pull/639/commits/e2a0e3c58e0f908854d59b9212eca28bc5f1bd0e#diff-23e4e8b8c3b0ebce6d44fff7c7ff93546043dcee6a94c6c2b54d5482d0457468R305-R321

src/vm/memory.zig
Sobeston: changed to camelCase in e2a0e3c58e0f908854d59b9212eca28bc5f1bd0e

src/vm/memory.zig
Sobeston: changed to camelCase in e2a0e3c58e0f908854d59b9212eca28bc5f1bd0e

src/vm/memory.zig
Sobeston: We need additional context (version + config, which Region code is not aware of) to return the correct error value, so it's best to resolve the error elsewhere.

src/vm/memory.zig
Sobeston: Same reason as previous

yewman (approved):
lgtm!

PR #640, opened 2025-03-26
fix(build): remove `installArtifact` for unit tests

Rexicon226:


InKryption:
Don't see why we have *any* `installArtifact` calls when we have our incremental-friendly `addInstallAndRun`.

InKryption (approved):

PR #641, opened 2025-03-26
Improve error handling, logs & http responses

InKryption:
Improve error handling, logs & http responses
* And some other error handling & response improvements.
* Use informative 503 Service Unavailable response for genesis file.
* Some logger refactors.
* Only use getSockName in io_uring in error log paths.
* Add missing logs to some places

Skeleton for rpc requests in basic backend
plus some miscellaneous fixes and renames

PR #642, opened 2025-03-26
Create `Call` union for representing RPC methods

InKryption:


InKryption:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Syndica/sig/642?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#643** <a href="https://app.graphite.dev/github/pr/Syndica/sig/643?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#642** <a href="https://app.graphite.dev/github/pr/Syndica/sig/642?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Syndica/sig/642?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#641** <a href="https://app.graphite.dev/github/pr/Syndica/sig/641?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/rpc/methods.zig
```diff
@@ -14,6 +19,320 @@ const Pubkey = sig.core.Pubkey;
...
+
+        const UntaggedPayload = @Type(.{ .Union = blk: {
+            var info = @typeInfo(Method).Union;
+            info.tag_type = null;
+            info.decls = &.{};
+            break :blk info;
+        } });
+    };
+
+    pub fn jsonStringify(
```
dnut, line 99: Why did you add this function when we already have a serialize method that works fine? Personally I find this implementation a lot harder to understand than the existing code. I think it's a huge readability advantage that the correct json format is exactly the way the json library already serializes structs. Why would you write a completely custom implementation? Is there a big advantage to this approach?

src/rpc/methods.zig
```diff
@@ -14,6 +19,320 @@ const Pubkey = sig.core.Pubkey;
...
+    id: Id,
+    method: Method,
+
+    pub const Id = union(enum) {
+        null,
+        int: i128,
+        str: []const u8,
+    };
+
+    pub const Method = union(enum) {
```
dnut, line 35: To me, this feels awkward as a nested struct. Could you imagine `Method` ever be used in a context where `Call` is not relevant?

InKryption: I suppose it currently is, I just nested it because it seemed like a logical way to structure it. I can move it outside.

src/rpc/methods.zig
InKryption: The existing serialize function relies on passing in the specific request type, and deriving the `method` field from the `@typeName` (through `methodName`), which is not acceptable, because `@typeName` does not provide a stable format upon which we can rely.
Side note on that: there's discussions about removing `@typeName` and instead simply making `@compileError` and the like take varargs, in which `type` values can be directly interpolated, to satisfy the only valid use case for which `@typeName` currently exists.

Besides that, it's an improvement in these ways:
* It doesn't force the caller to allocate a string, it allows usage of either stringifyAlloc or stringify.
* It permits serializing the `"jsonrpc": "2.0"` field without needing to actually store it.
* The `Call` structure actually provides a way to represent and therefore serialize a custom `"id"` field.
* Enforces exhaustive handling of method serialization since it uses a union rather than `request: anytype`.

If the main concern here is readability, I could factor out some of the behavior and structure it more similarly to how a struct is serialized.

InKryption: Also to be clear, the PR on top of this one in the stack removes the `serialize` function.

dnut: > The existing serialize function relies on passing in the specific request type, and deriving the method field from the @typeName (through methodName), which is not acceptable, because @typeName does not provide a stable format upon which we can rely.

I'm not convinced that this is a problem we need to worry about. `@typeName` isn't going to suddenly stop working out of nowhere. It would need to be a breaking change in zig. If that does happen, the tests would fail, and it would be trivial to fix.

>  Side note on that: there's discussions about removing @typeName and instead simply making @compileError and the like take varargs, in which type values can be directly interpolated, to satisfy the only valid use case for which @typeName currently exists.

That would be a shame. I can imagine other use valid use cases of `@typeName`. What if you have a generic struct, and you'd like to print the concrete type at runtime?

> It permits serializing the "jsonrpc": "2.0" field without needing to actually store it.

I don't understand. By store, do you mean create a struct that has this field? In this case "2.0" is still data that exists at runtime. What is the difference in storage?

> Enforces exhaustive handling of method serialization since it uses a union rather than request: anytype.

The enforcement I see is that you're only allowed to serialize types that are defined within this particular tagged union. I actually that as a downside. I'd rather be able to serialize any struct instead of requiring it to be a member of some central registry. This was actually one of the deliberate goals of the existing rpc client.

InKryption: > I'm not convinced that this is a problem we need to worry about. @typeName isn't going to suddenly stop working out of nowhere. It would need to be a breaking change in zig. If that does happen, the tests would fail, and it would be trivial to fix.

It may well stop working in the way you expect it to; tomorrow the zig team could decide that it would be beneficial to append the line & column of the declaration site to the end of the name, or to begin mangling the name if it happens to coincide with a type in another module with equivalent path, or perhaps it could even decide to truncate the name based on some heuristic.

This is part of the contract of `@typeName`; relying on its format is relying on compiler details that are subject to change, and which in fact have changed in the past many times.

> That would be a shame. I can imagine other use valid use cases of @typeName. What if you have a generic struct, and you'd like to print the concrete type at runtime?

The only use case currently considered valid is `@compileError("..." ++ @typeName(T) ++ "...")`, and in general the printing of types. Any other use case is considered out of scope. A big part of the rationale for considering its removal is in consideration of incremental compilation, given that currently its existence means the names assigned to types can influence comptime control flow.

> I don't understand. By store, do you mean create a struct that has this field? In this case "2.0" is still data that exists at runtime. What is the difference in storage?

On closer inspection this comment was due to a misunderstanding, I confused this code and the code in `rpc/response.zig` which stores the `jsonrpc` field.

> The enforcement I see is that you're only allowed to serialize types that are defined within this particular tagged union. I actually that as a downside. I'd rather be able to serialize any struct instead of requiring it to be a member of some central registry. This was actually one of the deliberate goals of the existing rpc client.

This approach seems problematic for testing, code coverage, ensuring that requests are well-formed, and poses a pretty unfortunate code sharing problem, because the server is simpler and safer to implement with an exhaustive listing of all implemented requests.
It also makes it harder to keep track of and statically verify what methods are actually implemented, both for the client as well as the server.

Putting the ability to send custom requests at this level seems somewhat convoluted compared to just adding that ability at the level where the string is actually sent, ie in the client.

---

All in all, I think the current ducktyped approach offers little benefit over this exhaustive approach, and relies on a contentious feature which may well change or be removed, and seems problematic on many levels.

src/rpc/methods.zig
InKryption: 5dc22c6de2a6cd68930118671c3fa0bec4b08555

src/rpc/methods.zig
dnut: We are digressing from my initial concern, so I'm going to refocus on that. Moving the serialization code exclusively into Call means you can't serialize requests unless they are wrapped in a Call. Due to this, the client's serialization is broken in your other branch. I think a good solution is to pull some of this code out into a standalone function. I found the old implementation to be more readable because it was more declarative, but it's not that important as long as it works.

Regarding all the other discussion, we clearly think differently on some aspects, but that's OK. I agree the union is useful in the server. My main concern was that I prefer each request type to embed enough information to be serialized. I don't like the client requiring extra inputs to communicate the method name. If you can get something like that working, that would be nice. Another approach would be with `const name: []const u8`. But it's not the top priority right now. If you see too many complications, we can revisit this later.

src/rpc/methods.zig
```diff
@@ -14,6 +19,329 @@ const Pubkey = sig.core.Pubkey;
 const Signature = sig.core.Signature;
 const Slot = sig.core.Slot;
 
+/// NOTE: for the sake of simplicity, we only support `method: ..., params: ...`,
+/// and reject `params: ..., method: ...`; this is a reasonable expectation for
+/// clients to satisfy.
+pub const Call = struct {
```
dnut, line 25: I was confused by this name at first and had to read the implementation to understand that it represents requests. I think `Request` would be a clearer name.

Also I would move this to another file. I think it's cleaner if `methods.zig` is specialized to solely define the individual methods. Note that the `Response` struct is defined in its own file. I would probably move this struct into another file like `request.zig`

src/rpc/methods.zig
InKryption: 669712e85df373d108f334f09681793ec459b267

src/rpc/methods.zig
InKryption: 669712e85df373d108f334f09681793ec459b267

src/rpc/methods.zig
```diff
@@ -14,6 +19,329 @@ const Pubkey = sig.core.Pubkey;
 const Signature = sig.core.Signature;
 const Slot = sig.core.Slot;
 
+/// NOTE: for the sake of simplicity, we only support `method: ..., params: ...`,
+/// and reject `params: ..., method: ...`; this is a reasonable expectation for
+/// clients to satisfy.
```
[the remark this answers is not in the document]
InKryption: 669712e85df373d108f334f09681793ec459b267

dnut (changes requested):

src/rpc/methods.zig
dnut: This commit doesn't address my initial concern. The serialization code does not work unless you wrap the request struct in another struct. Because of this, the client code is still broken in your other branch. Your other branch is only passing the tests because you changed the tests so that they're not actually testing the code used by the client any more.

The changes in these two branches are so coupled with each other, I think it would be easier to review if it was all in a single branch.

InKryption: 7a450aa8be09a68eb7bc5e643d115083c188e168

I went ahead and merged the top branch into this one

src/rpc/request.zig
```diff
@@ -65,6 +68,284 @@ fn AsTuple(comptime Struct: type) type {
     return @Type(.{ .Struct = info });
 }
 
+/// NOTE: for the sake of simplicity, we only support `method: ..., params: ...`,
+/// and reject `params: ..., method: ...`; this is a reasonable expectation for
```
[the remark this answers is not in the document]
InKryption: 1fd6c1db84f0403079b3bedf9efbda60cd899ee5

src/rpc/methods.zig
```diff
@@ -14,6 +19,110 @@ const Pubkey = sig.core.Pubkey;
...
+    isBlockhashValid: noreturn,
+    minimumLedgerSlot: noreturn,
+    requestAirdrop: RequestAirdrop,
+    sendTransaction: SendTransaction,
+    simulateTransaction: noreturn,
+
+    pub const Tag = @typeInfo(MethodAndParams).Union.tag_type.?;
+
+    /// Returns a wrapper over `self` which will be stringified as an array.
+    pub fn jsonStringifiedAsParamsArray(self: MethodAndParams) JsonStringifiedAsParamsArray {
```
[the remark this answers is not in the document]
InKryption: 60dcffc66d2153acb83e9174c6ab0cf3b4780154

src/rpc/request.zig
```diff
@@ -65,6 +68,284 @@ fn AsTuple(comptime Struct: type) type {
...
+        source: anytype,
+        options: std.json.ParseOptions,
+    ) std.json.ParseError(@TypeOf(source.*))!Id {
+        const TokType = enum { null, number, string };
+        const tok_type: TokType = switch (try source.peekNextTokenType()) {
+            .null => .null,
+            .number => .number,
+            .string => .string,
+            else => {
+                try source.skipValue();
```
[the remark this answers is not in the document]
InKryption: It aligns the parsing behavior with most other std.json parsers, namely the fact that the offending unexpected token is consumed before issuing the error; `peekNextTokenType` doesn't consume the offending token, so we do that here before returning.

src/rpc/request.zig
```diff
@@ -65,6 +68,284 @@ fn AsTuple(comptime Struct: type) type {
...
+                defer id_buf.deinit();
+
+                const maybe_str = try source.allocNextIntoArrayList(&id_buf, options.allocate.?);
+                if (std.fmt.parseInt(i128, maybe_str orelse id_buf.items, 10)) |int|
+                    break :id .{ .int = int }
+                else |err| switch (err) {
+                    error.Overflow, error.InvalidCharacter => {},
+                }
+                const str = maybe_str orelse try id_buf.toOwnedSlice();
+                break :id .{ .str = str };
```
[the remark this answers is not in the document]
InKryption: It's just error tolerant parsing of integers cases. If it's a number, it could also be a float, or an integer outside the range `-2¹²⁷`...`2¹²⁷-1`, neither of which are representible by `i128` (the latter is not likely, but entirely possible).
It's also simpler for them to share branches since they'll both be taking advantage of the `allocNextIntoArrayList` logic anyway.

src/rpc/response.zig
```diff
@@ -1,27 +1,26 @@
...
     return struct {
         arena: *std.heap.ArenaAllocator,
         id: u64,
         jsonrpc: []const u8,
         payload: Payload,
+        const Self = @This();
 
         pub const Payload = union(enum) {
-            result: Method.Response,
+            result: T,
```
dnut, line 16: Is this needed for the server?

src/rpc/response.zig
```diff
@@ -49,13 +48,13 @@ pub fn Response(comptime Method: type) type {
             };
         }
 
-        pub fn deinit(self: Response(Method)) void {
+        pub fn deinit(self: Self) void {
```
dnut, line 51: Did you switch to `Self` because it's a generic type? I don't know much about the internals of the compiler, but it's proven to be smart about avoiding infinite recursion here. So I'm not clear on the purpose of `Self` in this case.

src/rpc/request.zig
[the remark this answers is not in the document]
dnut: Let's say a client sets the id to a json number equal to `2^127`. This code is going to fail parsing it to i128, and then parse it as a string instead. Then when this ID is serialized back into json, it's going to be converted into a json string instead of the correct type: a number. It fails round trip idempotency. The client will not be able to correlate the request and response IDs.

So it is still failing to deserialize properly, but it's masquerading as a success. Wouldn't it be preferable to fail explicitly with an error?

Alternatively, you could just represent *all* numbers with a string instead of i128. It's not like we need to do math with these numbers.

```zig
union(enum) {
    null,
    string: []const u8,
    number: []const u8,
}
```

src/rpc/request.zig
```diff
@@ -65,6 +68,284 @@ fn AsTuple(comptime Struct: type) type {
...
+            .null => .null,
+            .number => .number,
+            .string => .string,
+            else => {
+                try source.skipValue();
+                return error.UnexpectedToken;
+            },
+        };
+
+        return switch (tok_type) {
```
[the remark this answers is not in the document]
dnut: I don't see how this implies a need for two switches. You can keep the same approach to the number and string branches with just one switch. How is this code any different from yours?

```zig
        return switch (try source.peekNextTokenType()) {
            .null => id: {
                std.debug.assert(try source.next() == .null);
                break :id .null;
            },
            .number, .string => id: {
                var id_buf = std.ArrayList(u8).init(allocator);
                defer id_buf.deinit();

                const maybe_str = try source.allocNextIntoArrayList(&id_buf, options.allocate.?);
                if (std.fmt.parseInt(i128, maybe_str orelse id_buf.items, 10)) |int|
                    break :id .{ .int = int }
                else |err| switch (err) {
                    error.Overflow, error.InvalidCharacter => {},
                }
                const str = maybe_str orelse try id_buf.toOwnedSlice();
                break :id .{ .str = str };
            },
            else => {
                try source.skipValue();
                return error.UnexpectedToken;
            },
        };
```

dnut (commented):
This is looking good, but I still have a lingering concern about numbers being deserialized incorrectly. Also posted a couple more questions

src/rpc/response.zig
InKryption: It isn't needed for the server per se, but I made this change because the original way it was structured confused me for quite a bit. Having `Response` simply be generic over the response type directly, rather than "a type that exposes a type decl called 'Response', which will be the only thing that is actually used" seems like a natural decoupling.

src/rpc/response.zig
InKryption: I switched it to `Self` since that's consistent with the rest of the code base, and is what we have codified in the style guide: https://github.com/Syndica/sig/blob/main/docs/docusaurus/docs/contributing/style-guide.mdx#method-parameters

src/rpc/request.zig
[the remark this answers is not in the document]
InKryption: Comprendido, big oversight on my part.
A nice property of being able to represent a number using `i128` is that most valid IDs are likely going to fit within it, saving needless heap-allocated copies of most. A middle ground would be to have two tags which represent numbers, one being a `i128`, and one being a string. I'll probably go ahead and do that.

src/rpc/response.zig
dnut: Actually the style guide says the opposite. Take a closer look, it says to use the struct name. Not sure if you remember the discussion we had about this a few months ago when reviewing some of David's code in the VM,  and we decided to start using this pattern. It's not a big deal, and we don't need to change the entire codebase right now, but I was a little confused by your change because it goes in the opposite direction. I thought you were making an exception because it was a generic type, and wanted to understand that a little better.

src/rpc/request.zig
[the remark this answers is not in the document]
InKryption: 26c970ed

src/rpc/request.zig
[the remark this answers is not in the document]
InKryption: 26c970ed

src/rpc/response.zig
dnut: Interesting. I find `Response(GetAccountInfo)` to be very intuitive. It's a function that returns the response type for the GetAccountInfo method. Now it's used like `Response(GetAccountInfo.Response)`, and returns the response type for a method's response type? It feels more confusing/redundant/verbose to me now.

Anyway, obviously I'm going to have this opinion, since I'm the one who deliberately made the original design choice. This change feels a bit divisive, but relatively inconsequential compared to all the improvements you've made, so I'll be happy to approve the PR regardless.

src/rpc/request.zig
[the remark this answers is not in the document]
dnut: Good idea. Sounds effective, and simple enough.

I also think it's fine if v1 of our server does a ton of unnecessary allocations. Once we support every RPC method, we can go over it with a fine toothed comb and optimize its performance.

src/rpc/response.zig
InKryption: I'm referring to:
> If the type name is not available (for example in anonymous structs), define const Self = @This() and use that as the type.

I interpret "anonymous struct" refers to structs that aren't directly assigned a name in global scope, ie anonymous structs. If that isn't the intention, we should clarify that.

Personally, generic types are the exact one place where I think `const Self = @This()`, `const Name = @This()`, or even `@This()` is preferable, given that otherwise changing the parameters in any way leads to more changes than would otherwise be needed during refactoring (certainly not as simple as renaming the struct).

Do you want to keep it as is, or go forward with the change?

src/rpc/request.zig
```diff
@@ -65,6 +68,284 @@ fn AsTuple(comptime Struct: type) type {
...
+            .method = .{ .getAccountInfo = .{
+                .pubkey = test_pubkey1,
+                .config = .{
+                    .encoding = .base58,
+                },
+            } },
+        },
+    );
+
+    try testParseCall(
```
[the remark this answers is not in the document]
InKryption: 850c0ded7673e730e605b324842a443eac6b6ee1

src/rpc/response.zig
dnut: We can leave this as is since it's undecided right now and not important enough to block the merge.

My intended meaning of "name is not available" was that there is simply no name that can be given to the type. `Response(Method)` is a name that can be given to the type, so it does not qualify, according to my intention. We can clarify this language once we settle on a standard.

dnut (approved):

PR #650, opened 2025-03-31
feat(cli,cmd,geyser): Document the library - refactor & fix some minor details

InKryption:
To discuss: are the new doc comments by themselves in this format sufficient, or would they be improved as a separate markdown document?

Don't pay too much attention to the code changes, they are minor implementation-facing improvements and small API-relevant details I noticed while writing the documentation.

### DIFF
--- a/src/cli.zig
+++ b/src/cli.zig
@@ -10,12 +10,12 @@
 //! ```zig
 //! .foo = .{
 //!     .fizz = "1",
-//!     .bar = .{
+//!     .subcmd = .{ .bar = .{
 //!         .buzz = true,
-//!         .baz = .{
+//!         .subcmd = .{ .baz = .{
 //!             .fizzbuzz = true,
-//!         },
-//!     },
+//!         } },
+//!     } },
 //! }
 //! ```
 //!
@@ -123,7 +123,9 @@ pub fn OptionInfo(comptime Opt: type) type {
     return struct {
         /// A named option will be specifiable as `--{name_override orelse field_name}(=<value>)?`,
         /// irrespective of position within the scope of the (sub)command.
-        /// A positional option
+        ///
+        /// A positional option will be specifiable as `{<value>}`, with respect to the position
+        /// it is declared relative to other positional options within the scope of the (sub)command.
         kind: enum { named, positional },
 
         /// Used to override the name displayed on the command line, or null
--- a/src/cli.zig
+++ b/src/cli.zig
@@ -874,8 +874,10 @@ fn CmdHelper(
                         );
 
                         if (arg_info.alias != .none) @compileError(
-                            "Argument " ++ parent_prefix ++ field_name ++ " cannot have an alias, " ++
-                                "since it's not named, it's " ++ @tagName(arg_info.kind),
+                            "Argument " ++
+                                parent_prefix ++ field_name ++
+                                " cannot have an alias, " ++
+                                "since it's not named, it's positional",
                         );
 
                         positional_set_ptr.* =
@@ -1429,7 +1431,10 @@ fn CmdHelper(
         } {
             const help_argument_alias_name = "-h, --help";
 
-            var max_name_alias_width: u64 = @max(help_argument_alias_name.len, name_alias_base_width);
+            var max_name_alias_width: u64 = @max(
+                help_argument_alias_name.len,
+                name_alias_base_width,
+            );
             var max_default_value_width: ?u64 = null;
 
             @setEvalBranchQuota(positional_set.len * 8 + 1);
--- a/src/cli.zig
+++ b/src/cli.zig
@@ -19,39 +19,39 @@
 //! }
 //! ```
 //!
-//! # Options
-//! A given command may accept zero or more options, named arguments that may
+//! # Arguments
+//! A given command may accept zero or more arguments, named arguments that may
 //! also be specified via an alias that is unique to it (per subcommand).
-//! An option may be of type bool, int, string, enum, and an optional xor
+//! An argument may be of type bool, int, string, enum, and an optional xor
 //! list of any of the former primitives.
 //!
-//! Note: the `-h, --help` option cannot be used by the programmer, it is
+//! Note: the `-h, --help` argument cannot be used by the programmer, it is
 //! controlled by the parser.
 //!
-//! ## Optional Options, Required Options, Default Values
-//! All options must have a default value, but may be treated as "required" by
+//! ## Optional Arguments, Required Arguments, Default Values
+//! All arguments must have a default value, but may be treated as "required" by
 //! the application by using an optional type with a default value of `null`,
-//! and issue an error if the option posseses this value.
+//! and issue an error if the argument posseses this value.
 //!
 //! `null` is not a specifiable value for an optional on the command line itself,
 //! meaning if the value is `null`, the application can know for certain that the
-//! option was not specified at all.
+//! argument was not specified at all.
 //!
-//! ## List Options
-//! List options are inherently "optional", in the sense that they are not allowed
+//! ## List Arguments
+//! List arguments are inherently "optional", in the sense that they are not allowed
 //! to have a non-empty default value, and will always default to an empty value
 //! if no values are specified. For this reason, a list cannot be optional, because
-//! whether or not the option was specified can be detected by whether or not the
+//! whether or not the argument was specified can be detected by whether or not the
 //! list is empty.
 //!
-//! Syntactically, lists are specified as a repeating series of the option.
+//! Syntactically, lists are specified as a repeating series of the argument.
 //! Example: `foo --bar 1 --bar 2 --bar 3 --bar 4`, would parse to a command
-//! where the option `bar` is equal to `&.{ 1, 2, 3, 4 }`.
+//! where the argument `bar` is equal to `&.{ 1, 2, 3, 4 }`.
 //!
-//! ## Positional Options
-//! A positional option is only allowed if there is no subcommand, and vice versa.
-//! They are specified before named options, and not after; for example:
-//! `foo file/path --opt=a`, but not `foo --opt=a file/path`.
+//! ## Positional Arguments
+//! A positional argument is only allowed if there is no subcommand, and vice versa.
+//! They are specified before named arguments, and not after; for example:
+//! `foo file/path --arg=a`, but not `foo --arg=a file/path`.
 
 const std = @import("std");
 
@@ -61,8 +61,8 @@ const std = @import("std");
 ///
 /// The input `S` must be a struct type where each field is one of the following:
 /// * A union where each tag represents a subcommand.
-/// * A struct where each field is an option.
-/// * A simple value which is assigned to an option.
+/// * A struct where each field is an argument.
+/// * A simple value which is assigned to an argument.
 ///
 /// Up to one union field is allowed, because at most one subcommand field is permitted,
 /// which is what a union represents (each tag of the union is a subcommand name).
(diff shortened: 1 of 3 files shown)


src/cli.zig
```diff
@@ -1,78 +1,137 @@
...
 
-pub const CommandHelp = struct {
-    short: []const u8,
-    long: ?[]const u8,
-};
+const std = @import("std");
 
+// -- API -- //
+
+/// This is the core tool you will use to describe the shape of the command line parser.
```
dnut, line 60: Isn't the struct definition the core tool that describes the shape? Isn't CommandInfo basically just for documentation and customization?

I just noticed that you can't even use the parser without an instance of CommandInfo, which seems odd. It would be nice if users of the library wouldn't be forced to deal with the extra complexity of creating a CommandInfo for every command, unless they want to add more details into the help message. It feels like a lot of redundant boilerplate.

src/cli.zig
```diff
@@ -1,78 +1,137 @@
...
+///
+/// The `sub: SubInfo` field is a struct with all of the same fields as the input struct `S`,
+/// except:
+/// * each union field type is replaced with `CommandInfo(FieldType)`.
+/// * each struct field type is replaced with `OptionInfoGroup(FieldType)`.
+/// * each normal field type is replaced with `OptionInfo(FieldType)`.
+///
+/// NOTE: each subcommand field must be optional, such that its absence in the command line
+/// can be represented as `null`, allowing the calling code to take appropriate action for
+/// any missing subcommand.
```
dnut, line 78: I get the idea of allowing subcommands to be optional so the client can deal with it using custom logic. But it seems like unnecessary flexibility that comes at the cost of ergonomics. It seems like an advanced feature that would rarely be taken advantage of, and should not be required.

I think it's more straightforward to allow clients to define subcommands as required by using the type system and have the parser return an error. In most cases, as a client, I would prefer that, rather than needing to write custom logic to deal with a missing optional value.

src/cli.zig
```diff
@@ -1,78 +1,137 @@
...
 
+/// This struct describes everything related to a given option within the scope of a (sub)command.
+///
+/// Similar to `CommandInfo`, it can be instantiated directly at the site of usage, or it can
+/// be pre-declared, which allows re-using the option info in multiple sub-commands.
 pub fn OptionInfo(comptime Opt: type) type {
     return struct {
+        /// A named option will be specifiable as `--{name_override orelse field_name}(=<value>)?`,
+        /// irrespective of position within the scope of the (sub)command.
+        /// A positional option
```
dnut, line 126: Missing the rest of this sentence?

src/cli.zig
```diff
@@ -1,78 +1,137 @@
...
         kind: enum { named, positional },
```
dnut, line 127: So the options and the positional arguments are just intermixed as struct fields, and the only way you can distinguish them is by looking at OptionInfo?

I think I prefer position arguments being in their own field called "positional," so the main struct is more organized and self-documenting.

src/cli.zig
```diff
@@ -1,78 +1,137 @@
...
+    /// It is recommended that this be worded in a way which is not redundant with the brief, since it
+    /// will be displayed as a paragraph that follows the brief.
+    long: ?[]const u8,
 };
 
+/// This struct describes everything related to a given option within the scope of a (sub)command.
+///
+/// Similar to `CommandInfo`, it can be instantiated directly at the site of usage, or it can
+/// be pre-declared, which allows re-using the option info in multiple sub-commands.
 pub fn OptionInfo(comptime Opt: type) type {
```
dnut, line 122: I don't understand the documentation of the `config` field in this struct. I also looked at the definition of OptionConfig and it took me a long time to figure out how to populate that field. It's also not clear to me why this field does not have a default value. I think there should be a default approach to parsing.

src/cli.zig
```diff
@@ -181,44 +210,68 @@ pub const OptionAlias = enum(u7) {
...
-    /// sub-command fields which is a tagged union (possibly optional) wherein each
+    /// Must be a struct type containing zero or more option fields, and optionally
+    /// one sub-command field which is an optional tagged union, wherein each
     /// member is a sub-command struct following the same description as this one,
     /// recursively.
-    ///
-    /// An option field must either be a parsed value such as a
-    /// boolean, integer, enum, string, or list or optional there-of,
-    /// or a struct of parsed values.
+    /// See `CommandInfo`.
```
dnut, line 274: is this supposed to be one line lower?

src/cmd.zig
```diff
@@ -181,7 +181,7 @@ const Cmd = struct {
     metrics_port: u16,
     log_file: ?[]const u8,
     tee_logs: bool,
-    subcmd: union(enum) {
+    subcmd: ?union(enum) {
```
dnut, line 184: Previously it did not need to be optional? I think it would be cleaner to leave it the old way.

dnut (commented):
Overall this is very nice. I like the added documentation and think we should merge it. I just have one concern about the high level example at the top of the file because it seems inaccurate.

Also I have some feedback here about the design of the library in general. I would have ideally put this on the initial pr but it got merged before I could review it. So it doesn't need to be addressed in this PR, I just wanted to note my impressions, and we can figure out later if anything actually needs to change.

src/cli.zig
```diff
@@ -1,78 +1,137 @@
...
+//! controlled by the parser.
+//!
+//! ## Optional Options, Required Options, Default Values
+//! All options must have a default value, but may be treated as "required" by
+//! the application by using an optional type with a default value of `null`,
+//! and issue an error if the option posseses this value.
+//!
+//! `null` is not a specifiable value for an optional on the command line itself,
+//! meaning if the value is `null`, the application can know for certain that the
+//! option was not specified at all.
```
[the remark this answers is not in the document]
InKryption: The important take-away of this paragraph is that the library does not have such a concept as a "required" option, all options must have a fallback. The reason optionals are brought up is because they can be easily used by the *application* to issue an informative error message if an option is null when it's expected to be non-null, ie "required by the application".
If the application does not wish to "require" the value, it simply doesn't issue an error when the value is null.
If the application does have a simple default value, it should simply specify it.

I don't think you would be able to misuse the library, because there is only one way to accomplish the described tasks; if you wanted a required value, you'd discover that all options must have a default value and that the library will not error for a missing value on the command line, and as such you would need to use the approach making use of optionals.

src/cli.zig
```diff
@@ -1,78 +1,137 @@
...
+//!
+//! ## List Options
+//! List options are inherently "optional", in the sense that they are not allowed
+//! to have a non-empty default value, and will always default to an empty value
+//! if no values are specified. For this reason, a list cannot be optional, because
+//! whether or not the option was specified can be detected by whether or not the
+//! list is empty.
```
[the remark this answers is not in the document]
InKryption: The intention trying to be expressed here is the fact that `?[]const T = null` is illegal, because it is redundant with `[]const T = &.{}`, in how the library works.

src/cmd.zig
InKryption: It might be cleaner, but it is more complicated and less flexible, adding multiple ways doing the same thing, such that the core features are less discoverable.

src/cli.zig
InKryption: The most you need to do as a client is `const subcmd = cmd.subcmd orelse return error.Badcommand` if you don't want to do anything fancy, but having obligate optional sub commands means there's only one way to do it, and requires much fewer assumptions about the intentions or needs of the caller, while keeping all core features discoverable and simple.

src/cli.zig
InKryption: It's needed for disambiguating the interpretation of certain types, default value specification, as well as properly specifying subcommand info by extension.
It's quite a bit more complicated and hard to document without this. A previous design did in fact do such a thing, but it was unwieldy and felt very magical, in a bad way.

src/cli.zig
InKryption: In most cases, there is exactly one way in which parsing will work. Ie, for `[]const T`, where `T != u8`, it should be relatively obvious that it's a list of `T`, therefore `.config = {}`. It is only configurable in ambiguous cases, ie `[]const u8`, which could be a string, or a list of bytes.

src/cli.zig
InKryption: No, it's in correct position, it's making reference to the doc comment of `CommandInfo`, which is relevant here.

src/cli.zig
```diff
@@ -1,78 +1,137 @@
...
+//! ```zig
+//! .foo = .{
+//!     .fizz = "1",
+//!     .bar = .{
+//!         .buzz = true,
+//!         .baz = .{
+//!             .fizzbuzz = true,
+//!         },
+//!     },
+//! }
```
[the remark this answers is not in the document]
InKryption: e4fc4a41f879358a986caa635b8ad9883d1961bb

I think the unit test serves as a more illustrative and tested example that won't potentially bit rot with time.

src/cli.zig
InKryption: e4fc4a41f879358a986caa635b8ad9883d1961bb

src/cli.zig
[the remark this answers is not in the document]
dnut: > he library does not have such a concept as a "required" option ... If the application does not wish to "require" the value, it simply doesn't issue an error when the value is null.

OK I understand how this works now, and I see why you say it can't be misused.

It is unexpected to me that a cli parsing library would have no concept of a required argument. Can you make this explicit in the docs? Currently the docs explain the workaround, but not the reason for it. This is what led to my confusion.

src/cli.zig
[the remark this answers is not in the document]
dnut: The confusion I had here was caused by your the of homonyms to mean different things in the same context. In one place you say "optional" to mean "not required." In another place, you say "optional" to mean "nullable type." Can you clarify that with distinct language?

I also see "option" used here for a *third* meaning: "command-line argument." I think it would be best to avoid the usage of the word "option" since it means so many different things and the reader will have trouble figuring out what the author means when they use the word.

src/cli.zig
dnut: In my opinion, the parser should have sane defaults that don't require you to include any `CommandInfo`s, but I think this is too complicated to figure out right now. It would probably involve some design changes, like specifying defaults on the structs themselves, which I know you saw some issues with. We can revisit this another time.

src/cli.zig
dnut: I see the benefit of your approach. Once you understand how it works, there is an elegance to how this approach is both simple and flexible. But personally, I would prefer to support required arguments.

I think the "everything is optional" idea is unusual and counterintuitive, since cli arguments are often *not* optional. The library does not fully model the domain it aims to support. I understand you can still require arguments in the application, but what I mean is that the library does not define the concept for you, which makes me as a user feel lost because I expect it to be in there *somewhere*. I think this makes the library hard to understand because users need to stumble over this omission until they get creative and find a workaround. It caused me confusion when I initially reviewed this PR, and it didn't click for me until I started to read your responses to my review. I think the library would actually be easier to understand if it included the extra code complexity that would be needed for a more complete representation of its domain.

I also think there there are many more places that the library will be used than the number of places where it is defined. So I'm happy to make the library more complex to simplify its usages.

Anyway I don't think you need to change anything. I feel that your docs here are clear, and it helps to mitigate the confusion I was describing.

src/cli.zig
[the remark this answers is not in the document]
InKryption: f04f1d0931e8d6a8a22594259f143cca7dcd3050

src/cli.zig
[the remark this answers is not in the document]
InKryption: 4b4ad49a962184959618f94bb15501583548f3e6

dnut (approved):

PR #659, opened 2025-04-07
refactor(runtime): refactor runtime and testing framework

yewman:
- Create Epoch and Slot Contexts
- Add allocator to TransactionContext and remove from LogCollector
    - TransactionContext requires allocator for syscalls
    - Since TransactionContext requires an allocator, the LogCollector does not need one 
- Implement set_return_data
- Tidy testing framework


src/vm/tests.zig
```diff
@@ -2020,7 +2020,8 @@ test "pqr divide by zero" {
     program[0] = @intFromEnum(OpCode.mov32_imm);
     program[16] = @intFromEnum(OpCode.exit_or_syscall);
 
-    inline for (.{
+    // TODO: Why does this cause a transitive error when using inline?
```
Rexicon226, line 2023: See the comment above about an LLVM miscompilation, I am sure it's the same thing here.

yewman: thnx!

src/runtime/transaction_context.zig
```diff
@@ -24,9 +24,35 @@ pub const MAX_INSTRUCTION_TRACE_LENGTH = 64;
...
+};
+
 /// [agave] https://github.com/anza-xyz/agave/blob/faea52f338df8521864ab7ce97b120b2abb5ce13/sdk/src/transaction_context.rs#L136
 /// [agave] https://github.com/anza-xyz/agave/blob/faea52f338df8521864ab7ce97b120b2abb5ce13/program-runtime/src/invoke_context.rs#L192
 pub const TransactionContext = struct {
+    /// Allocator
+    allocator: std.mem.Allocator,
+
+    /// Slot Context
+    sc: *const SlotContext,
```
dnut, line 54: These two-letter field names are esoteric. Can we use more descriptive names for clarity? I lean towards using full words, as in `instruction_context`, `transaction_context`, `slot_context`, and `epoch_context`. If you think the verbosity is a problem, maybe we can do `ixn_ctx`, `txn_ctx`, `slot_ctx`, and `epoch_ctx`? Alternatively, what do you think about adding methods that drill down into nested layers? For example `TransactionContext` could have a method called `epochContext`

src/runtime/transaction_context.zig
```diff
@@ -24,9 +24,35 @@ pub const MAX_INSTRUCTION_TRACE_LENGTH = 64;
...
+    }
+};
+
+pub const SlotContext = struct {
+    /// Epoch Context
+    ec: *const EpochContext,
+
+    /// Sysvar Cache
+    sysvar_cache: SysvarCache,
+};
```
dnut, line 45: Do you envision these structs replacing the `Bank`? After these are fleshed out, I assume we'd only need something to replace `BankRc` to manage the accounts?

src/runtime/transaction_context.zig
```diff
@@ -149,36 +158,23 @@ pub const TransactionContext = struct {
...
-        if (self.log_collector) |*lc| try lc.log(fmt, args);
+        if (self.log_collector) |*lc| try lc.log(self.allocator, fmt, args);
     }
 };
 
 /// [agave] https://github.com/anza-xyz/solana-sdk/blob/e1554f4067329a0dcf5035120ec6a06275d3b9ec/transaction-context/src/lib.rs#L493
 pub const TransactionReturnData = struct {
     program_id: Pubkey = Pubkey.ZEROES,
-    data: std.ArrayListUnmanaged(u8) = .{},
+    data: std.BoundedArray(u8, MAX_RETURN_DATA) = .{},
```
dnut, line 174: nice

src/runtime/log_collector.zig
```diff
@@ -41,28 +40,29 @@ pub const LogCollector = struct {
     /// [agave] https://github.com/anza-xyz/agave/blob/faea52f338df8521864ab7ce97b120b2abb5ce13/program-runtime/src/log_collector.rs#L25
     pub fn log(
         self: *LogCollector,
+        allocator: std.mem.Allocator,
```
dnut, line 43: This doesn't need to be addressed in the current PR, but maybe we can reuse the log collector to avoid each transaction triggering allocations for its logs.

For the message contents, we could just allocate one big 10KB buffer up front, assuming that we'll need it pretty soon anyway, and pull slices from that buffer as needed for the logs.

`self.messages` could also be allocated up front, though choosing a reasonable size is a little more complicated. Maybe dynamic+recycled is fine here.

dnut (changes requested):
Looks good, mainly just have a concern about naming.

src/runtime/log_collector.zig
yewman: I had a similar thought but decided we can come back to it later, I don't think it will bottleneck anything so the impact should be pretty minimal. Something to keep in mind for perf down the line I think

src/runtime/transaction_context.zig
yewman: Not directly, in that they will contain additional information which is not in the bank. However, if we take inspiration from fd over agave we could have a SlotBank and EpochBank which live in their respective contexts

src/runtime/transaction_context.zig
yewman: Fair enough, I actually don't find them all that esoteric since they directly correspond to the acronym of each type. I think full names will make the code less readable so would be in favour of the abbreviated names such as `txn_ctx`. Instead of accessor methods, what do you think about each context containing pointers to all its higher level contexts instead?

yewman: Did these in a couple of commits for clarity:
96a0cb8c
1cea3223
c553a72b
cb246c31
f085d785
I think we can revisit accessor methods or direct access to higher level contexts later :)

yewman: Rename func args for ic 16555d2e

yewman: Completed all renames of 'ic' and 'tc' 
41073d12

dnut: To understand it, you need to know about the different contexts in the svm, you need to be familiar with the pattern of how acronyms are used for contexts, and the same pattern should not be used for other things in the same scope. When I say esoteric, I just mean it requires special knowledge like this. In general, I do not like acronyms. I think they almost always make code harder to understand, unless some specific criteria are met.

I assume you're trying to avoid verbosity/repetition like this:
```zig
instruction_context.transaction_context.slot_context.epoch_context
```
Personally, I don't mind this. I think the explicitness is justified. But I get it. Brevity is valuable - as long as it doesn't sacrifice clarity.

I'm fine with any of these and do not have a preference between them:
1. `ixn_ctx.txn_ctx.slot_ctx.epoch_ctx`
2. accessor methods
3. directly contain all higher-level pointers

1 and 2 both involve indirection. 3 bloats the structs with extra data, but it avoids indirection which is a nice bonus.

yewman: I think understanding any of the code within the runtime requires some level of understanding of the contexts it uses. I don't often like acronyms either, however, in this instance I think using acronyms makes the code easier to read and understand. It requires a little overhead in figuring out what `ic` `tc` `sc` and `ec` refer to initially but that is kind of a one time cost for a reader, once its baked in it makes the rest of the code in the runtime much easier to understand in my opinion. The pattern issue applies to any abbreviation since in Sig we tend to avoid abbreviations. Using the full names is a possibility but it will result in relatively simple statements taking up multiple lines which makes code hard to read imo. For example, of the following two blocks I find the former much cleaner as you need to read less before reaching the actual function call.
```zig
    if (ids.NATIVE_LOADER_ID.equals(&program_owner)) {
        if (bpf_loader_program.v1.ID.equals(&ic.ixn_info.program_meta.pubkey)) {
            try ic.tc.consumeCompute(bpf_loader_program.v1.COMPUTE_UNITS);
            try ic.tc.log("Deprecated loader is no longer supported", .{});
            return InstructionError.UnsupportedProgramId;
        } else if (bpf_loader_program.v2.ID.equals(&ic.ixn_info.program_meta.pubkey)) {
            try ic.tc.consumeCompute(bpf_loader_program.v2.COMPUTE_UNITS);
            try ic.tc.log(
                "BPF loader management instructions are no longer supported",
                .{},
            );
            return InstructionError.UnsupportedProgramId;
        } else if (bpf_loader_program.v3.ID.equals(&ic.ixn_info.program_meta.pubkey)) {
            try ic.tc.consumeCompute(bpf_loader_program.v3.COMPUTE_UNITS);
            return executeBpfLoaderV3ProgramInstruction(allocator, ic);
        } else if (bpf_loader_program.v4.ID.equals(&ic.ixn_info.program_meta.pubkey)) {
            try ic.tc.consumeCompute(bpf_loader_program.v4.COMPUTE_UNITS);
            return executeBpfLoaderV4ProgramInstruction(allocator, ic);
        } else {
            return InstructionError.IncorrectProgramId;
        }
    }
``` 
```zig
    if (ids.NATIVE_LOADER_ID.equals(&program_owner)) {
        if (bpf_loader_program.v1.ID.equals(&instruction_context.ixn_info.program_meta.pubkey)) {
            try instruction_context.transaction_context.consumeCompute(bpf_loader_program.vi.COMPUTE_UNITS);
            try instruction_context.transaction_context.log("Deprecated loader is no longer supported", .{});
            return InstructionError.UnsupportedProgramId;
        } else if (bpf_loader_program.v2.ID.equals(&instruction_context.ixn_info.program_meta.pubkey)) {
            try instruction_context.transaction_context.consumeCompute(bpf_loader_program.v2.COMPUTE_UNITS);
            try instruction_context.transaction_context.log(
                "BPF loader management instructions are no longer supported",
                .{},
            );
            return InstructionError.UnsupportedProgramId;
        } else if (bpf_loader_program.v3.ID.equals(&instruction_context.ixn_info.program_meta.pubkey)) {
            try instruction_context.transaction_context.consumeCompute(bpf_loader_program.v3.COMPUTE_UNITS);
            return executeBpfLoaderV3ProgramInstruction(allocator, instruction_context);
        } else if (bpf_loader_program.v4.ID.equals(&instruction_context.ixn_info.program_meta.pubkey)) {
            try instruction_context.transaction_context.consumeCompute(bpf_loader_program.v4.COMPUTE_UNITS);
            return executeBpfLoaderV4ProgramInstruction(allocator, instruction_context);
        } else {
            return InstructionError.IncorrectProgramId;
        }
    }
```
Anyway I know `ic` `tc` `sc` and `ec` is pretty extreme abbreviation so happy to concede on this, checkout the PR now using the `ixn_ctx`, ... approach and let me know what you think

dnut: in that example, you could add one line of code at the start
```zig
const any_name = instruction_context.transaction_context;
```

dnut: feel free to resolve this. leaving open for visibility

src/runtime/log_collector.zig
dnut: https://github.com/orgs/Syndica/projects/2/views/10?pane=issue&itemId=105975325

dnut (approved):

src/runtime/transaction_context.zig
yewman: Yeah but almost everywhere you use instruction_context.transaction_context you effectively 'alias' then I would argue that renaming is probably a good idea

dnut (approved):

PR #660, opened 2025-04-07
fix(runtime): implement fixes to reach system-program conformance

yewman:


src/runtime/program/bpf_loader/execute.zig
```diff
@@ -789,8 +789,8 @@ pub fn executeV3SetAuthorityChecked(
     allocator: std.mem.Allocator,
     ic: *InstructionContext,
 ) (error{OutOfMemory} || InstructionError)!void {
-    if (!ic.ec.feature_set.active.contains(
-        features.ENABLE_BPF_LOADER_SET_AUTHORITY_CHECKED_IDX,
+    if (!ic.tc.sc.ec.feature_set.active.contains(
```
kprotty, line 792: should this go through `ic.ec`?

yewman: 28d4293e

Rexicon226 (approved):
LGTM.

I want to note that our current `custom_error `strategy makes no sense. If we use `@intFromEnum`, the number starts counting from 0 for each "error" enum type, which isn't unique. On the other hand, `@intFromError` won't give us well-defined numbers or a consistent order, but the numbers will be unique. Currently, we only use the `custom_error` to compare against an expected `custom_error`. In that case, we should be using error sets + `anyerror`. Otherwise, we need to make sure that the total order of the error enums is unique.

Issue #661, opened 2025-04-09
Crash when decompressing mainnet snapshot


maierfelix:
### Description

Memory usage increases dramatically during the decompression phase of a downloaded mainnet snapshot and leads to a crash. The reason seems to be that at the moment the whole snapshot to decompress is loaded into memory.

The crash error: `"UnTarTask encountered error: OutOfMemory`

My current workaround for this is to first let sig download the snapshot, then go to `accounts_db` and manually decompress using `zstd`:
````
zstd -cd snapshot-xx.tar.zst | tar -xf -
zstd -cd incremental-snapshot-xx.tar.zst | tar -xf -
````

### How to Reproduce the Bug

Run
````
sig snapshot-download \
    --gossip-port 8001 \
    --entrypoint entrypoint.mainnet-beta.solana.com:8001 \
    --entrypoint entrypoint2.mainnet-beta.solana.com:8001 \
    --entrypoint entrypoint3.mainnet-beta.solana.com:8001 \
    --entrypoint entrypoint4.mainnet-beta.solana.com:8001 \
    --min-snapshot-download-speed 50
````

dnut:
merging with #662

> Hi! I'm running this on a AMD Ryzen 7950X3D with 128gb RAM and 2TB SSD


> when running the validator for the first time, is it normal that this step takes so long and what is the recommended hardware for running sig on mainnet?
> ```
> time=2025-04-04T07:09:09.957Z level=info scope=accounts_db message="gathering account hashes [thread0]: 1/256 (0%) (est: 1h14m12.418s elp: 17.46s)"
> time=2025-04-04T07:09:27.085Z level=info scope=accounts_db message="gathering account hashes [thread0]: 2/256 (0%) (est: 1h13m12.736s elp: 34.588s)"
> time=2025-04-04T07:09:43.065Z level=info scope=accounts_db message="gathering account hashes [thread0]: 3/256 (1%) (est: 1h11m4.649s elp: 50.568s)"
> time=2025-04-04T07:09:59.547Z level=info scope=accounts_db message="gathering account hashes [thread0]: 4/256 (1%) (est: 1h10m24.215s elp: 1m7.051s)"
> time=2025-04-04T07:10:14.878Z level=info scope=accounts_db message="gathering account hashes [thread0]: 5/256 (1%) (est: 1h8m55.577s elp: 1m22.382s)"
[...]

PR #663, opened 2025-04-09
fix(snapshot): reduce loading memory usage

Sobeston:
Investigating https://github.com/Syndica/sig/issues/661, this should be a fix.

### DIFF
--- a/src/accountsdb/snapshots.zig
+++ b/src/accountsdb/snapshots.zig
@@ -2622,7 +2622,7 @@ pub fn parallelUnpackZstdTarBall(
         null,
         file_size,
         std.posix.PROT.READ,
-        std.posix.MAP{ .TYPE = .SHARED },
+        std.posix.MAP{ .TYPE = .PRIVATE },
         file.handle,
         0,
     );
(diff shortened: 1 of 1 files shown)


src/accountsdb/snapshots.zig
```diff
@@ -2618,12 +2618,17 @@ pub fn parallelUnpackZstdTarBall(
...
+    const memory = try std.posix.mmap(
+        null,
+        file_size,
+        std.posix.PROT.READ,
+        std.posix.MAP{ .TYPE = .PRIVATE },
+        file.handle,
+        0,
+    );
+    defer std.posix.munmap(memory);
+
```
dnut, line 2630:
```suggestion
try std.posix.madvise(memory.ptr, memory.len, std.posix.MADV.SEQUENTIAL | std.posix.MADV.WILLNEED);

```

dnut: Is this assumption correct? Will it read the entire thing sequentially?

Sobeston: zstd supports streaming decompression so seems likely? But can't comment beyond that

dnut: I'm thinking this is an easy optimization we can throw in for now. Later we can redo the whole thing to stream data from the file. At least this gives us a minor improvement over normal mmap. Do you have any concerns about adding this line of code?

Sobeston: No concerns, it's likely a small improvement

dnut (approved):

dnut (approved):

PR #666, opened 2025-04-11
feat(consensus): Implement `ProgressMap`

InKryption:
First pass which mirrors the rust implementation and has a unit test
validating that the ownership semantics in and between all of the
data structures are sound.
Provides minimal implementations of two functions that may be needed
imminently.

### DIFF
--- a/src/consensus/lib.zig
+++ b/src/consensus/lib.zig
@@ -1,5 +1,13 @@
 pub const fork_choice = @import("fork_choice.zig");
+pub const progress_map = @import("progress_map.zig");
+
+comptime {
+    _ = fork_choice;
+    _ = progress_map;
+}
 
 pub const HeaviestSubtreeForkChoice = fork_choice.ForkChoice;
 pub const ForkWeight = fork_choice.ForkWeight;
 pub const ForkInfo = fork_choice.ForkInfo;
+
+pub const ProgressMap = progress_map.ProgressMap;
--- a/src/consensus/lib.zig
+++ b/src/consensus/lib.zig
@@ -1,11 +1,6 @@
 pub const fork_choice = @import("fork_choice.zig");
 pub const progress_map = @import("progress_map.zig");
 
-comptime {
-    _ = fork_choice;
-    _ = progress_map;
-}
-
 pub const HeaviestSubtreeForkChoice = fork_choice.ForkChoice;
 pub const ForkWeight = fork_choice.ForkWeight;
 pub const ForkInfo = fork_choice.ForkInfo;
--- a/src/consensus/progress_map.zig
+++ b/src/consensus/progress_map.zig
@@ -6,6 +6,52 @@ const Hash = sig.core.Hash;
 const Pubkey = sig.core.Pubkey;
 const PubkeyArraySet = std.AutoArrayHashMapUnmanaged(Pubkey, void);
 
+/// TODO: merge into real replay stage when it exists?
+const replay_stage = struct {
+    pub const SUPERMINORITY_THRESHOLD: struct {
+        comptime numerator: comptime_int = 1,
+        comptime denominator: comptime_int = 3,
+
+        pub fn asFloat(self: @This(), comptime T: type) T {
+            const num: T = @floatFromInt(self.numerator);
+            const denom: T = @floatFromInt(self.denominator);
+            return num / denom;
+        }
+
+        pub fn orderAgainstScalar(self: @This(), value: anytype) std.math.Order {
+            const T = @TypeOf(value);
+            switch (@typeInfo(T)) {
+                .ComptimeInt,
+                => std.math.order(self.numerator, self.denominator * value),
+                .Int,
+                => std.math.order(self.numerator, std.math.mulWide(T, self.denominator, value)),
+                .ComptimeFloat,
+                .Float,
+                => std.math.order(self.asFloat(T), value),
+                else => comptime unreachable,
+            }
+        }
+
+        /// Order of threshold against `fraction_num / fraction_denom` without conversion to floating point.
+        pub fn orderAgainstFraction(
+            self: @This(),
+            fraction_num: anytype,
+            fraction_denom: anytype,
+        ) std.math.Order {
+            const Num = @TypeOf(fraction_num);
+            const Denom = @TypeOf(fraction_denom);
+
+            const NormalizedNum = std.math.IntFittingRange(0, self.numerator * std.math.maxInt(Denom));
+            const self_norm_num = @as(NormalizedNum, self.numerator) * fraction_denom;
+
+            const NormalizedDenom = std.math.IntFittingRange(0, std.math.maxInt(Num) * 3);
+            const frac_norm_num = self.numerator * @as(NormalizedDenom, fraction_num);
+
+            return std.math.order(self_norm_num, frac_norm_num);
+        }
+    } = .{};
+};
+
 /// TODO: any uses of these types are to be evaluated in their context, and
 /// the actual required synchronization semantics are to be determined later.
 const stubs = struct {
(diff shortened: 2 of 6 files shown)


InKryption:
* **#666** <a href="https://app.graphite.dev/github/pr/Syndica/sig/666?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Syndica/sig/666?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/consensus/progress_map.zig
```diff
@@ -0,0 +1,1169 @@
...
+
+        return result;
+    }
+
+    pub fn getForkStats(self: ProgressMap, slot: Slot) ?*ForkStats {
+        const fork_progress = self.map.getPtr(slot) orelse return null;
+        return &fork_progress.fork_stats;
+    }
+
+    pub fn getHash(self: *const ProgressMap, slot: Slot) ?Hash {
```
dadepo, line 94: This and getForkStats are the methods that are directly used in the tower. Although the underlying map would be populated somewhere else (during the replay?) having these two methods is enough to plug it in, into the tower.

dadepo (approved):

src/consensus/lib.zig
```diff
@@ -1,5 +1,13 @@
 pub const fork_choice = @import("fork_choice.zig");
+pub const progress_map = @import("progress_map.zig");
+
+comptime {
+    _ = fork_choice;
+    _ = progress_map;
+}
```
dnut, line 7: What is this for?

src/consensus/progress_map.zig
```diff
@@ -0,0 +1,1169 @@
...
+        const fork_progress = self.map.get(slot) orelse return null;
+        return fork_progress.fork_stats.bank_hash;
+    }
+};
+
+pub const ForkProgress = struct {
```
dnut, line 100: Would you mind adding function(s) to cover the logic in ForkProgress::new and ForkProgress::new_from_bank? For the bank you can stub out the dependency with a basic struct that has the minimal fields necessary.

src/consensus/progress_map.zig
```diff
@@ -0,0 +1,1169 @@
...
+
+            return .{
+                .voted = voted,
+                .stake = self.stake,
+            };
+        }
+    };
+};
+
+pub const cluster_info_vote_listener = struct {
```
dnut, line 365: nice you've already gotten a head start on this.

src/consensus/progress_map.zig
```diff
@@ -0,0 +1,1169 @@
...
+
+    pub const ExecuteAccessoryTimings = struct {
+        feature_set_clone_us: Saturating(u64),
+        get_executors_us: Saturating(u64),
+        process_message_us: Saturating(u64),
+        process_instructions: ExecuteProcessInstructionTimings,
+    };
+};
+
+test "ProgressMap memory ownership" {
```
dnut, line 739: It would be nice to also have the same test coverage as agave's 4 tests in progress_map.rs, plus supporting whatever functions are used in those tests.

src/consensus/progress_map.zig
```diff
@@ -0,0 +1,1169 @@
...
+            .duplicate_confirmed_hash = self.duplicate_confirmed_hash,
+            .computed = self.computed,
+            .lockout_intervals = lockout_intervals,
+            .bank_hash = self.bank_hash,
+            .my_latest_landed_vote = self.my_latest_landed_vote,
+        };
+    }
+};
+
+pub const PropagatedStats = struct {
```
dnut, line 211: Could you support `add_vote_pubkey` and `add_node_pubkey`?

src/consensus/progress_map.zig
```diff
@@ -0,0 +1,1169 @@
...
+            if (info.child != Saturating(T)) lazy.nonMatchErr();
+            info.child = T;
+
+            return @Type(.{ .Pointer = info });
+        }
+    };
+}
+
+/// AUDIT: https://github.com/anza-xyz/agave/blob/cb32984a9b0d5c2c6f7775bed39b66d3a22e3c46/core/src/consensus/progress_map.rs
+pub const ProgressMap = struct {
```
dnut, line 63: Could you add some of the logic from agave's ProgressMap? You can add them all, or use your judgement about whether some of them are redundant/pointless, but at least `get_leader_propagation_slot_must_exist` and `get_bank_prev_leader_slot` seem worthwhile.

dnut (changes requested):
This looks good. I like how you organized the code.

Replay actually needs most of the methods that agave implements on these structs. Can you add those so the structs are ready for use?

src/consensus/progress_map.zig
dadepo: `progress.my_latest_landed_vote` is another function off the `ProgressMap` that has popped up in the new sets of tower related methods I am adding.

dadepo: ...and just ran into a usage of [get_propagated_stats](https://github.com/anza-xyz/agave/blob/8e5470a9c4c01f5725e2d37c475abf6980fcb435/core/src/consensus/progress_map.rs#L280) also

dadepo: and also [get_leader_propagation_slot_must_exist](https://github.com/anza-xyz/agave/blob/8e5470a9c4c01f5725e2d37c475abf6980fcb435/core/src/consensus/progress_map.rs#L347)

src/consensus/progress_map.zig
```diff
@@ -0,0 +1,1169 @@
...
+const sig = @import("../sig.zig");
+
+const Slot = sig.core.Slot;
+const Hash = sig.core.Hash;
+const Pubkey = sig.core.Pubkey;
+const PubkeyArraySet = std.AutoArrayHashMapUnmanaged(Pubkey, void);
+
+/// TODO: any uses of these types are to be evaluated in their context, and
+/// the actual required synchronization semantics are to be determined later.
+const stubs = struct {
```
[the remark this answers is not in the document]
InKryption: I'd rather not use any actual sync primitives in these yet; just because the rust code is structured in such a way that it needs them doesn't mean it makes sense for us to use the exact same synchronization structure. These stubs serve as an annotation to indicate that their usages *may* require some form of synchronization, since they did in the rust code.

src/consensus/progress_map.zig
InKryption: I think I would like to add these in a follow up PR, the scope of this PR has reached a boiling point with other methods recently added requested by Drew.

src/consensus/lib.zig
InKryption: Removed, I forget why it was there in the first place, I think I was running into an issue in an earlier iteration and this worked around it.

src/consensus/progress_map.zig
InKryption: 5a91c246b4c3e8f1ee5dfd062f5edba8280c6ed4
52b42b82333ca9ebd4cca7bc534cefb46f43e532

src/consensus/progress_map.zig
InKryption: 52b42b82333ca9ebd4cca7bc534cefb46f43e532
de067e322dee7f764a49423e5e9f633057274109

src/consensus/progress_map.zig
InKryption: I think it would be better to do this in a follow-up PR, the size of this PR is reaching a boiling point.

src/consensus/progress_map.zig
InKryption: 00f61eba5189cd56a229187677ee292f8dbe0545
Implemented three of the four (3/4) tests, I'll implement the fourth in the follow up PR, since it tests some other unimplemented bits.

src/consensus/progress_map.zig
```diff
@@ -0,0 +1,1965 @@
...
+    retry_iteration: u32,
+};
+
+pub const LockoutIntervals = struct {
+    map: HashThatShouldBeMadeBTreeMap,
+
+    /// TODO: replace this with a BTree map. In the meantime, just have to
+    /// manually keep this sorted.
+    pub const HashThatShouldBeMadeBTreeMap =
+        std.AutoArrayHashMapUnmanaged(ExpirationSlot, EntryList);
```
dnut, line 603: What about `sig.utils.collections.SortedMap`?

src/consensus/progress_map.zig
```diff
@@ -0,0 +1,1965 @@
...
+    /// Same as `addVotePubkeyAssumeCapacity`, except it [re-]allocates if needed, instead of
+    /// assuming capacity is sufficient.
+    pub fn addVotePubkey(
+        self: *PropagatedStats,
+        allocator: std.mem.Allocator,
+        vote_pubkey: Pubkey,
+        stake: u64,
+    ) std.mem.Allocator.Error!bool {
+        const gop = try self.propagated_validators.getOrPut(allocator, vote_pubkey);
+        gop.value_ptr.* = {};
```
dnut, line 527: Does this actually do anything?

dnut (approved):
nice work. just want to address a few questions before merging

src/consensus/progress_map.zig
InKryption: It doesn't do anything, it just makes it clear to a reader of the code that the value is simply a void value; omitting it would make it seem like at a glance that there's a missing initialization for `gop.value_ptr.*`, until you look deeper and see that it's a `Map(Pubkey, void)`.

src/consensus/progress_map.zig
InKryption: I don't want to use a managed data structure here considering that this is going to be stored in memory. That, plus there's currently no actual code that reads the field in any way that would rely on it being properly ordered, so the value gained from using the sorted map would be outweighed by the point about it being managed.

PR #668, opened 2025-04-11
feat(runtime): bpf-loader conformance (instructions only)

yewman:
Bug fixes for bpf-loader-conformance. Passing all public test-vectors for instructions excluding the v4 loader

### DIFF
--- a/src/runtime/instruction_info.zig
+++ b/src/runtime/instruction_info.zig
@@ -128,7 +128,12 @@ pub const InstructionInfo = struct {
     }
 };
 
-pub fn limitedFixedBufferStream(comptime Buffer: type, buffer: Buffer, limit: usize) LimitedFixedBufferStream(Buffer) {
+/// Custom limited fixed buffer stream to support limited deserialization of instructions
+pub fn limitedFixedBufferStream(
+    comptime Buffer: type,
+    buffer: Buffer,
+    limit: usize,
+) LimitedFixedBufferStream(Buffer) {
     return .{
         .buffer = buffer,
         .pos = 0,
--- a/src/runtime/instruction_info.zig
+++ b/src/runtime/instruction_info.zig
@@ -109,9 +109,12 @@ pub const InstructionInfo = struct {
         allocator: std.mem.Allocator,
         comptime T: type,
     ) InstructionError!T {
-        const max_len = @min(self.instruction_data.len, Transaction.MAX_BYTES);
-        const buffer = self.instruction_data[0..max_len];
-        return bincode.readFromSlice(allocator, T, buffer, .{}) catch {
+        var lfbs = limitedFixedBufferStream(
+            []const u8,
+            self.instruction_data,
+            Transaction.MAX_BYTES,
+        );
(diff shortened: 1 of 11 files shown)


src/runtime/program/bpf_loader/execute.zig
```diff
@@ -2826,3 +2868,41 @@ test "executeV3Migrate" {
...
     );
 }
+
+fn createValidProgramData(
+    allocator: std.mem.Allocator,
+    state: anytype,
+    state_size: usize,
+    additional_bytes: usize,
+) ![]u8 {
+    if (!@import("builtin").is_test)
```
kprotty, line 2878: should import be up top re:style?

yewman: d3f89668

src/runtime/instruction_info.zig
```diff
@@ -124,3 +127,46 @@ pub const InstructionInfo = struct {
         if (self.account_metas.len < minimum_accounts) return InstructionError.NotEnoughAccountKeys;
     }
 };
+
+/// Custom limited fixed buffer stream to support limited deserialization of instructions
```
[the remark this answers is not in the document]
yewman: Thanks for pointing this out, initially this was my approach but due to a bincode issue that I didn't look into enough it failed some instances. I have made the bincode fix and am now using the above approach. 7ddf0f93

src/vm/syscalls.zig
```diff
@@ -42,6 +45,191 @@ pub const Entry = struct {
     builtin_fn: Syscall,
 };
 
+// [agave] https://github.com/anza-xyz/agave/blob/a2af4430d278fcf694af7a2ea5ff64e8a1f5b05b/programs/bpf_loader/src/syscalls/mod.rs#L335
+pub fn register(
```
Rexicon226, line 49: I think this should either be called `registerDefault` or moved out of the VM.

Rexicon226 (changes requested):

src/runtime/program/bpf_loader/execute.zig
```diff
@@ -167,13 +167,14 @@ pub fn executeV3Write(
     offset: u32,
     bytes: []const u8,
 ) (error{OutOfMemory} || InstructionError)!void {
+    std.debug.print("bytes.len={}\n", .{bytes.len});
```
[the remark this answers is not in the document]
yewman: 502130f0f

src/vm/syscalls.zig
yewman: `registerDefault` does not make sense since it is not a default. What is your issue with it being here? I think it makes sense since we a creating a `BuiltinProgram` (vm defined) by registering syscalls (vm defined) using the FeatureSet (runtime defined) to conditionally register some syscalls.

Rexicon226: Hmm, I've thought about it some more and realized a couple of assumptions I made were incorrect. I guess this is fine as-is. Still just seems like a weird place to put it, but I cannot articulate why, unfortunately.

Rexicon226 (approved):

PR #672, opened 2025-04-15
refactor(runtime): Change SysvarCache to use byte slices similar to Agave and Fd

yewman:
Agave stores a number of sysvars in the sysvar cache as byte slices, these slices are written into vm memory by sysvar getter syscalls. This change modifies the existing sysvar cache to mimic this approach, used by both Fd and Agave.


src/runtime/sysvar_cache.zig
```diff
@@ -1,41 +1,126 @@
...
+        if (self.slot_hashes) |slot_hashes| allocator.free(slot_hashes);
+        if (self.slot_hashes_obj) |slot_hashes_obj| slot_hashes_obj.deinit(allocator);
+        if (self.stake_history) |stake_history| allocator.free(stake_history);
+        if (self.stake_history_obj) |stake_history_obj| stake_history_obj.deinit(allocator);
+        if (self.recent_blockhashes) |recent_blockhashes| recent_blockhashes.deinit(allocator);
+    }
+
+    /// Returns the sysvar as an object if it is supported
+    /// Replaces the sysvar object getters in Agave
+    /// NOTE: No allocations are actually performed here, we require the allocator purely
```
Rexicon226, line 56: You should use a failing allocator then, `std.testing.failing_allocator`.
That will simplify this PR a ton.

Rexicon226 (changes requested):

src/runtime/sysvar_cache.zig
yewman: amazing thankyou 🙏

yewman: 59a53465

Rexicon226 (approved):

PR #676, opened 2025-04-16
fix(core,bloom,consensus): refactor some data structures

Rexicon226:
Look at each commit for the refactor.


src/core/hard_forks.zig
```diff
@@ -1,104 +1,75 @@
 const std = @import("std");
-const ArrayList = std.ArrayList;
-const allocator = std.mem.Allocator;
 const Slot = @import("time.zig").Slot;
-const expect = std.testing.expect;
-const eql = std.mem.eql;
+const expectEqual = std.testing.expectEqual;
```
yewman, line 3: Personally I like to keep testing imports within tests, rather than top level imports. I find top level imports useful for understanding dependencies and think that imports for tests creates ambiguity. It can lead to repetitive imports within tests but in my opinion the tradeoff is worth it for better clarity of dependencies

src/core/shred.zig
```diff
@@ -48,30 +46,20 @@ test "core.shred: test ShredVersion" {
...
-    const logger = test_logger.logger();
-
     const shred_version_one = ShredVersion.computeShredVersion(Hash.ZEROES, null);
     try std.testing.expect(shred_version_one == 1);
-    logger.debug().logf("shred_version_one: {}", .{shred_version_one});
 
-    var hard_forks = HardForks.default(testing_alloc);
-    defer _ = hard_forks.deinit();
+    var hard_forks: HardForks = .{};
+    defer _ = hard_forks.deinit(allocator);
```
yewman, line 52:
```suggestion
    defer hard_forks.deinit(allocator);
```

scripts/style.py
```diff
@@ -142,7 +142,6 @@ def unused_imports(args, files_to_check):
     "src/bloom/bit_set.zig",
     "src/bloom/bit_vec.zig",
     "src/bloom/bitvec.zig",
-    "src/bloom/bloom.zig",
```
yewman, line 145: Why remove this?

src/bloom/bloom.zig
```diff
@@ -77,12 +77,16 @@ pub const Bloom = struct {
...
     }
 
-    pub fn initRandom(alloc: std.mem.Allocator, random: std.Random, num_items: usize, false_rate: f64, max_bits: usize) error{OutOfMemory}!Self {
+    pub fn initRandom(
+        alloc: std.mem.Allocator,
+        random: std.Random,
+        num_items: usize,
+        false_rate: f64,
+        max_bits: usize,
+    ) error{OutOfMemory}!Self {
```
yewman, line 89: If we are refactoring this, lets remove Self as well

yewman (changes requested):
I like the changes, a couple of notes otherwise lgtm!

src/core/hard_forks.zig
Rexicon226: I agree that mixing them isn't optimal, but having imports inside of tests is something I'm very much against. How about we place them above the area of the file where we have our "test" blocks?

scripts/style.py
Rexicon226: Because `bloom.zig` now passes the style check.

src/core/hard_forks.zig
InKryption: I would suggest just not importing them in the first place, but if this was a democratic vote, I'd vote for importing inside the tests.

src/bloom/bloom.zig
Rexicon226: ee91892f

src/core/shred.zig
Rexicon226: 109498a5

src/core/hard_forks.zig
Rexicon226: I'm happy with not importing them and fully qualifying the namespace.

yewman (approved):

yewman (approved):

PR #677, opened 2025-04-17
fix(accountsdb loading): improve memory usage for mainnet

Sobeston:
Before, the index would use ram for:
1. All operations in loading threads
2. Pubkey ref map

Now the disk is used for both, resulting in huge memory savings for loading mainnet snapshots. As of now, you need only 13GB of ram from the gpa, whereas before it required well over 100 (couldn't measure it - system locked up long before finishing).

This PR also greatly reduces memory allocations done by code reading accounts from the buffer pool, which was necessary for some memory tracking tools.

### DIFF
--- a/src/accountsdb/accounts_file.zig
+++ b/src/accountsdb/accounts_file.zig
@@ -292,23 +292,29 @@ pub const AccountFile = struct {
 
     pub fn validate(
         self: *const Self,
-        metadata_allocator: std.mem.Allocator,
         buffer_pool: *BufferPool,
     ) !usize {
         var offset: usize = 0;
         var number_of_accounts: usize = 0;
         var account_bytes: usize = 0;
 
+        var buffer_pool_frame_buf: [BufferPool.MAX_READ_BYTES_ALLOCATED]u8 = undefined;
+        var fba = std.heap.FixedBufferAllocator.init(&buffer_pool_frame_buf);
+        const allocator = fba.allocator();
+
         while (true) {
             const account = self.readAccount(
-                metadata_allocator,
+                allocator,
                 buffer_pool,
                 offset,
             ) catch |err| switch (err) {
                 error.EOF => break,
                 else => return err,
             };
-            defer account.deinit(metadata_allocator);
+            defer {
+                account.deinit(allocator);
+                fba.reset();
+            }
 
             try account.validate();
             offset = offset + account.len;
--- a/src/accountsdb/accounts_file.zig
+++ b/src/accountsdb/accounts_file.zig
@@ -681,7 +681,7 @@ test "core.accounts_file: verify accounts file" {
     var accounts_file = try AccountFile.init(file, file_info, 10);
     defer accounts_file.deinit();
 
-    _ = try accounts_file.validate(std.testing.allocator, &bp);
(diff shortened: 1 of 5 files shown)


src/accountsdb/db.zig
```diff
@@ -170,14 +170,18 @@ pub const AccountsDB = struct {
         number_of_index_shards: usize,
         /// Amount of BufferPool frames, used for cached reads. Default = 1GiB.
         buffer_pool_frames: u32 = 2 * 1024 * 1024,
+        /// For supplying your own ReferenceAllocator. Required for index_allocation = .parent.
+        injected_ref_allocator: ?*sig.accounts_db.index.ReferenceAllocator = null,
```
InKryption, line 174: Would suggest something akin to:
```zig
index_allocation: union(AccountIndex.AllocatorConfig.Tag) {
    ram,
    disk,
    parent: *sig.accounts_db.index.ReferenceAllocator,
},
```
to make the invalid parameter combination unrepresentable.

src/accountsdb/index.zig
```diff
@@ -104,17 +105,24 @@ pub const AccountIndex = struct {
                     .{sig.utils.fmt.tryRealPath(index_dir, ".")},
                 );
 
+                // reviewer's note: looks like we leak this + is there a point in allocating?
```
InKryption, line 108: Probably has something to do with the managed data structures inside the AccountIndex data structure holding `std.mem.Allocator` references to it, meaning it needs to be a stable pointer.

I think I might have tried to refactor this at some point but gave up, can't quite recall though. Feel free to take a stab at it if you want to see if you have a better time of it.

src/accountsdb/accounts_file.zig
```diff
@@ -292,23 +292,29 @@ pub const AccountFile = struct {
...
         var account_bytes: usize = 0;
 
+        var buffer_pool_frame_buf: [BufferPool.MAX_READ_BYTES_ALLOCATED]u8 = undefined;
+        var fba = std.heap.FixedBufferAllocator.init(&buffer_pool_frame_buf);
+        const allocator = fba.allocator();
+
         while (true) {
             const account = self.readAccount(
-                metadata_allocator,
+                allocator,
```
InKryption, line 307: If this size invariant actually holds, this is a very nice saving, both here and the other couple places you do this.

InKryption (changes requested):
Just one suggestion, and a couple neutral comments; overall, great change set, always nice to get rid of some allocator parameters.

src/accountsdb/accounts_file.zig
Sobeston: Tested with mainnet, it holds. You just need to read a valid account (10MiB max)

src/accountsdb/index.zig
Sobeston: Alright, just going to leave this be. Removed the note as it isn't a leak.

ff4b8b296aba104cf0e2f7a8c70da1e9e69914cf

src/accountsdb/db.zig
Sobeston: Changed in ae65af4b8667ee59bfc4c81a81d8f5f835493888

InKryption (approved):

PR #678, opened 2025-04-17
perf(profiling): integrate tracy

Sobeston:
Adds tracy instrumentation. Tracks various spans in our code, and tracks our allocations. Shouldn't do anything without `-Denable-tracy`.

Best explained with some screenshots:

gpa memory usage going up with accountsdb init:
image

a time distribution
image

a suspected memory leak
image
image

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -159,6 +159,10 @@ pub fn build(b: *Build) !void {
     });
     const tracy_mod = tracy_dep.module("tracy");
 
+    if (config.enable_tracy) {
+        tracy_mod.linkLibrary(tracy_dep.artifact("tracy"));
+        tracy_mod.link_libcpp = true;
+    }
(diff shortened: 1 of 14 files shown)


build.zig
```diff
@@ -147,6 +152,17 @@ pub fn build(b: *Build) !void {
...
     const pretty_table_mod = pretty_table_dep.module("prettytable");
 
+    const tracy_dep = b.dependency("tracy", .{
+        .target = config.target,
+        .optimize = config.optimize,
+        .tracy_enable = config.enable_tracy,
+    });
+    const tracy_mod = tracy_dep.module("tracy");
+
+    if (config.enable_tracy) {
```
Rexicon226, line 162: This should be handled in the dependency.

InKryption (approved):

PR #680, opened 2025-04-17
fix(testing): disable some flaky tests

Rexicon226:
These tests keep failing in the CI, making it very difficult to work. We should just disable them for now and have a tracking issue to fix them.

Tracking issue: https://github.com/Syndica/sig/issues/681


yewman (approved):
Agreed

PR #682, opened 2025-04-21
feat(vm): syscall conformance

yewman:
**ExecutionError**
The main change in this PR is the definition of the ExecutionError which is an error set consisting of the union of InstructionError, EbpfError, SyscallError, and some Sig specific errors. InstructionError.ArithmeticOverflow has been renamed to InstructionError.ProgramArithmeticOverflow to avoid conflicting with SyscallError.ArithmeticOverflow. It is necessary to understand which error set an error belongs to for conformance and logging.

### DIFF
--- a/src/runtime/compute_budget.zig
+++ b/src/runtime/compute_budget.zig
@@ -65,20 +65,9 @@ pub const ComputeBudget = struct {
     /// [fd] https://github.com/firedancer-io/firedancer/blob/211dfccc1d84a50191a487a6abffd962f7954179/src/flamenco/vm/syscall/fd_vm_syscall_crypto.c#L238-L245
     ///
     /// Returns the cost of a Poseidon hash syscall for a given input length.
-    pub fn poseidonCost(self: ComputeBudget, len: u64) SyscallError!u64 {
-        const squared_inputs = std.math.powi(u64, len, 2) catch
-            return SyscallError.ArithmeticOverflow;
-
-        const mul_result = std.math.mul(
-            u64,
-            squared_inputs,
-            self.poseidon_cost_coefficient_a,
-        ) catch return SyscallError.ArithmeticOverflow;
-
-        return std.math.add(
-            u64,
-            mul_result,
-            self.poseidon_cost_coefficient_c,
-        ) catch return SyscallError.ArithmeticOverflow;
+    pub fn poseidonCost(self: ComputeBudget, len: std.math.IntFittingRange(0, 12)) u64 {
+        const squared_inputs = std.math.powi(u64, len, 2) catch unreachable;
+        const mul = squared_inputs * self.poseidon_cost_coefficient_a;
+        return mul + self.poseidon_cost_coefficient_c;
     }
 };
--- a/src/runtime/compute_budget.zig
+++ b/src/runtime/compute_budget.zig
@@ -1,4 +1,7 @@
 const std = @import("std");
+const sig = @import("../sig.zig");
+
+const SyscallError = sig.vm.SyscallError;
 
 /// [agave] https://github.com/anza-xyz/agave/blob/a11b42a73288ab5985009e21ffd48e79f8ad6c58/compute-budget/src/compute_budget.rs#L11-L119
 pub const ComputeBudget = struct {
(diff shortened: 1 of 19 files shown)


src/vm/syscalls/lib.zig
```diff
@@ -221,64 +215,111 @@ pub fn register(
...
 
     const arg1 = registers.get(.r1);
     const arg2 = registers.get(.r2);
     const arg3 = registers.get(.r3);
     const arg4 = registers.get(.r4);
     const arg5 = registers.get(.r5);
 
-    try ctx.log(
-        "0x{x} 0x{x} 0x{x} 0x{x} 0x{x}",
+    const message = try std.fmt.allocPrint(
```
Rexicon226, line 254: Why are you using `allocPrint` here, the log collector already duplicates the message and takes ownership.

src/vm/syscalls/lib.zig
```diff
@@ -221,64 +215,111 @@ pub fn register(
...
+/// [agave] https://github.com/firedancer-io/agave/blob/66ea0a11f2f77086d33253b4028f6ae7083d78e4/programs/bpf_loader/src/syscalls/logging.rs#L107
+pub fn logData(tc: *TransactionContext, memory_map: *MemoryMap, registers: RegisterMap) Error!void {
+    const vm_addr = registers.get(.r1);
+    const len = registers.get(.r2);
+
+    try tc.consumeCompute(tc.compute_budget.syscall_base_cost);
 
-    try ctx.log("TODO: compute budget calculations", .{});
+    const vm_messages = try memory_map.translateSlice(
+        []u8,
```
Rexicon226, line 299: This is wrong, a slice cannot have a well-defined layout.

src/vm/syscalls/lib.zig
```diff
@@ -362,31 +407,42 @@ pub fn poseidon(ctx: *TransactionContext, mmap: *MemoryMap, registers: RegisterM
...
         );
-        try hasher.append(slice[0..32]);
+        hasher.append(slice[0..32]) catch {
+            // TODO: THIS IS INCORRECT, it is set to temporarily match the error set. Agave
+            // returns sets a custom poseidon error here.
+            return error.OutOfMemory;
+        };
     }
-    const result = try hasher.finish();
+    const result = hasher.finish() catch {
```
Rexicon226, line 421: This function cannot error, feel free to make it `catch unreachable`.

Rexicon226 (changes requested):

src/vm/syscalls/lib.zig
Rexicon226: https://github.com/Rexicon226/zig-poseidon/commit/a6b20b50420bdb15bfe29cafcb27e440a776ce9b please update the dep to use this one.

src/runtime/compute_budget.zig
```diff
@@ -58,17 +61,20 @@ pub const ComputeBudget = struct {
     /// [fd] https://github.com/firedancer-io/firedancer/blob/211dfccc1d84a50191a487a6abffd962f7954179/src/flamenco/vm/syscall/fd_vm_syscall_crypto.c#L238-L245
     ///
     /// Returns the cost of a Poseidon hash syscall for a given input length.
-    pub fn poseidonCost(self: ComputeBudget, len: u64) !u64 {
-        const squared_inputs = try std.math.powi(u64, len, 2);
-        const mul_result = try std.math.mul(
+    pub fn poseidonCost(self: ComputeBudget, len: u64) SyscallError!u64 {
+        const squared_inputs = std.math.powi(u64, len, 2) catch
```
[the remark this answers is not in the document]
yewman: Thats fair, I don't think we should default to returning errors not explicitly defined within Syscall, Ebpf, or Instruction error sets where possible though. Perhaps we can use an orelse return null here and use a single catch to return SyscallError.ArithmeticOverflow. Even though we may know that there is never a case when it can overflow, a new reader / auditor may not, so it is best to handle it explicitly even if it will never be encountered imo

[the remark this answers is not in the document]
Rexicon226: I'm happy just not making these safety checked.

src/vm/executable.zig
```diff
@@ -919,9 +919,10 @@ pub const BuiltinProgram = struct {
     functions: Registry(syscalls.Syscall) = .{},
 
     pub fn deinit(
-        self: *BuiltinProgram,
+        self_: BuiltinProgram,
```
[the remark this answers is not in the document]
yewman: 228dd81b

src/vm/lib.zig
```diff
@@ -14,3 +17,99 @@ pub const Config = executable.Config;
 pub const Vm = interpreter.Vm;
 pub const Elf = elf.Elf;
 pub const Section = executable.Section;
+
+const InstructionError = sig.core.instruction.InstructionError;
+
+pub const SyscallError = error{
```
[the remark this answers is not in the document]
yewman: I think we can dispense with them entirely, 228dd81b

src/vm/syscalls/lib.zig
yewman: 228dd81b

src/vm/syscalls/lib.zig
yewman: 38b9562c

Rexicon226 (approved):

PR #686, opened 2025-04-24
refactor(core,accountsdb): Move bank to core and add basic bank structs

dnut:
The diff in this PR is a bit confusing because a ton of code was moved and a recent PR was reverted. Here is a more readable summary of changes:

The main idea of this PR is to make some bank fields available to the rest of the codebase. To do this, I added three new structs to `core/bank.zig`, and I moved most of the bank code from accountsdb into core. The moved code has simply been copy and pasted so you can rest assured it is exactly the same code. Here are the new structs: https://github.com/Syndica/sig/blob/de12d5876bb8676847a6b0f2ef8ccef88b025210/src/core/bank.zig#L45-L157

I also was forced to add some unit tests to improve test coverage. This revealed a bug in `ExtraFields.clone` where the `versioned_epoch_stakes` field was not being cloned. Once this was fixed, it revealed a missing errdefer in `VersionEpochStake.clone`, which I also fixed. These changes were in 3bb8823498cff09201dad39d266308c85fd17110 and 627ee133d8bae6f75a38b3bbf9bdd21a3534cb15.

In ab829d9fba95b253df980d9225c6f402a758d07e I also reverted #687 because the merge conflicts are severe and the changes from #687 are not needed right now. This makes the diff for the current PR look a bit strange. The only changes in this PR (relative to the commit before #687) are the ones mentioned above.

## Sig's `Bank` Approach

We're not currently planning to use a `Bank` struct like agave. The `Bank` has become a catch-all bucket containing an incredible amount of state and methods, and it is used almost everywhere in the validator, despite most of that state being unnecessary in most places where the `Bank` is used. It's the ultimate spaghetti code, where everything depends on everything. It's very hard to make sense of the dependencies. A more granular expression of dependencies will make the code throughout Sig a lot clearer.

The bank data needs to be represented somehow, and it's becoming necessary for much of what we're implementing in replay and consensus. So I implemented three structs that each contain specialized subsets of the fields in agave's Bank:
- `SlotState`: data about a slot that changes as the slot is executed
- `SlotConstants`: unchanging data about a slot
- `EpochConstants`: unchanging data about an epoch

Other parts of agave's `Bank`, like the pointer to accountsdb, can be managed as separate dependencies. Tons of methods should not be needed here either. For example, the logic to execute transactions can live in the SVM module.

This is partially inspired by firedancer's approach of having `fd_slot_bank` separate from the `fd_epoch_bank`, and keeping accountsdb managed separately from the banks.

If eventually we realize that we do actually need all of these things combined together in many places, then it will be trivial to write something like this:

```zig
pub const Bank = struct {
    accounts: *AccountsDB,
    slot_state: *SlotState,
    slot_constants: SlotConstants,
    epoch_constants: EpochConstants,
};
```

This could work just like the Bank does in agave, but at least it's organized. I don't see a need for this right now, so I haven't included it.


src/core/stake.zig
```diff
@@ -0,0 +1,799 @@
...
+    }
+};
+
+/// Analogous to [Stakes](https://github.com/anza-xyz/agave/blob/1f3ef3325fb0ce08333715aa9d92f831adc4c559/runtime/src/stakes.rs#L186).
+/// It differs in that its delegation element parameterization is narrowed to only accept the specific types we actually need to implement.
+pub fn Stakes(comptime DelegationElem: type) type {
+    std.debug.assert(
+        DelegationElem == Delegation or
+            DelegationElem == Stake,
+    );
```
InKryption, line 147: Why go from `Stakes(.delegation)` to `Stakes(Delegation)`? The former seems more easily discoverable, given that the parameter type itself describes the allowed set of variants, instead of looking at the comptime logic - that, plus needing to sync exhaustive handling between here and all the sites where the type is potentially switched on.

InKryption (commented):
I have just one comment, overall everything seems to be a net improvement.

src/core/stake.zig
dnut: Regarding the list of allowed variants, in both cases you need to look at the definition of `Stakes` to figure it out. Actually I think it's more discoverable with `Stakes(Delegation)` because it's the first line of code in the function. You don't need to go look at some other function to figure it out.

Exhaustiveness is great, but you can still have exhaustive switches with either approach. `else => unreachable` makes the switch exhaustive in this case. Either way, you get a compilation error for any unhandled type.

The problem with the enum is readability. I feel that the code's meaning is obscured by unnecessary layers of indirection. Using an enum to indirectly select a type is confusing compared to just specifying the type directly. I'd like to be able to see some struct field like `stakes: Stakes(Delegation)` and get as much information out of it as possible at first glance, like, _"here's a field that uses `Stakes` to do something with the `Delegation` type."_ At first, I thought the enum was just some configuration. I didn't realize it actually determines the inner type until I read the `Stakes` implementation. `Stakes(Delegation)` more clearly communicates that it deals with the Delegation type.

InKryption (approved):
LGTM

PR #687, opened 2025-04-24
refactor(snapshots): "random init" refactor + style fix

InKryption:
Removes `fillHashmapWithRng`, it was kind of complicated and clunky compared to just writing the same logic directly. Also adds a minimum to all the data entry lengths for certain debugging purposes I found I needed recently.

Makes snapshots.zig conform to the column length limit.


Rexicon226 (approved):
LGTM

PR #689, opened 2025-04-25
feat(runtime): batch transaction execution

Sobeston:
Treating this as an MVP PR, not necessarily going to handle every path. 

Limitations:
- missing some validation code
- no "rollback" or program cache
- no support for precompile and compute budget instructions


src/runtime/account_loader.zig
```diff
@@ -41,34 +42,41 @@ pub const LoadedAccounts = struct {
     collected_rent: u64 = 0,
     loaded_account_data_size: u32 = 0,
 
+    /// n accounts loaded and returned (excludes loading of owner accounts)
+    n_loaded: u8,
```
Rexicon226, line 46: You've just created a `std.BoundedArray` here, but less safe :)

Sobeston: True, however the length is for two buffers? What do you think, make them both bounded arrays or no

Rexicon226: Yeah

yewman (commented):
some minor comments from initial pass

src/accountsdb/snapshots.zig
```diff
@@ -583,7 +583,7 @@ pub const Manifest = struct {
 };
 
 /// Analogous to [TransactionError](https://github.com/anza-xyz/agave/blob/cadba689cb44db93e9c625770cafd2fc0ae89e33/sdk/src/transaction/error.rs#L14)
-const TransactionError = union(enum) {
+pub const TransactionError = union(enum) {
```
yewman, line 586: We will need a new version of the TransactionError which holds the InstructionError error type instead of the InstructionError union type. Checkout transaction_error.zig in https://github.com/Syndica/sig/pull/733, perhaps implementing this within this PR would make sense since we will not be dealing with the InstructionError union type

src/runtime/account_loader.zig
```diff
@@ -3,72 +3,36 @@ const sig = @import("../sig.zig");
...
 
-const MockedBank = struct {
+pub const MockedAccountsDb = struct {
     allocator: std.mem.Allocator,
-    slot: Slot,
-    accounts: std.AutoArrayHashMapUnmanaged(Pubkey, sig.core.Account),
-    rent_collector: RentCollector,
-
-    fn deinit(self: *MockedBank) void {
+    accounts: std.AutoArrayHashMapUnmanaged(Pubkey, sig.core.Account) = .{},
```
yewman, line 28: Can we alias Account?

src/runtime/account_loader.zig
```diff
@@ -78,32 +42,27 @@ fn Bank(comptime kind: BankKind) type {
...
-        fn rentCollector(self: Self) RentCollector {
-            return switch (kind) {
-                .AccountsDb => self.inner.bank_fields.rent_collector,
-                .Mocked => self.inner.rent_collector,
-            };
-        }
-        fn getAccount(self: Self, pubkey: *const sig.core.Pubkey) !sig.core.Account {
+        fn getAccount(
+            self: Self,
+            pubkey: *const sig.core.Pubkey,
```
yewman, line 47: alias Pubkey?

src/runtime/account_loader.zig
```diff
@@ -120,127 +79,385 @@ fn Bank(comptime kind: BankKind) type {
...
+        };
+
+        return .{ .ok = result };
+    }
+
+    fn loadTransactionAccountsInner(
+        self: *const BatchAccountCache,
+        allocator: std.mem.Allocator,
+        transaction: *const RuntimeTransaction,
+        rent_collector: *const sig.core.rent_collector.RentCollector,
```
yewman, line 292: can we alias RentCollector and FeatureSet?

src/runtime/account_loader.zig
```diff
@@ -381,41 +503,39 @@ fn accumulateAndCheckLoadedAccountDataSize(
 // [agave] https://github.com/anza-xyz/agave/blob/cb32984a9b0d5c2c6f7775bed39b66d3a22e3c46/svm/src/account_loader.rs#L639
 fn constructInstructionsAccount(
     allocator: std.mem.Allocator,
-    tx: *const sig.core.Transaction,
+    tx: *const runtime.transaction_execution.RuntimeTransaction,
```
yewman, line 506:
```suggestion
    transaction: *const RuntimeTransaction,
```

src/runtime/program/compute_budget/lib.zig
```diff
@@ -0,0 +1,128 @@
+const sig = @import("../../../sig.zig");
+
+const InstructionInfo = sig.runtime.InstructionInfo;
```
yewman, line 3: We could probably remove a bunch of this implementation and just keep the stub plus required constants for this PR, the compute budget program PR has the rest of the completed implementation

yewman (commented):
A few comments otherwise looking good :)

src/runtime/account_loader.zig
Sobeston: aliased in 09a4180fedec79391b1136494da64a82e755015e

src/runtime/account_loader.zig
Sobeston: I wanted to more obviously disambiguate it from AccountSharedData if that makes sense

src/accountsdb/snapshots.zig
Sobeston: Yeah we could move the change to this PR, the change is good

src/runtime/account_loader.zig
Sobeston: fixed in ab2caef2e6d9f661ef8bc7da4b04a9a03f3ff587

src/runtime/program/compute_budget/lib.zig
Sobeston: stubbed it away in 3afafbb2fe7c0df368ae5b2b22ca5f201fef9230

src/runtime/account_loader.zig
Sobeston: fixed in 09a4180fedec79391b1136494da64a82e755015e

src/accountsdb/snapshots.zig
yewman: want to add it?

src/runtime/account_loader.zig
yewman: yeah that is fair :)

src/runtime/account_loader.zig
```diff
@@ -78,32 +46,27 @@ fn Bank(comptime kind: BankKind) type {
                 .Mocked => self.inner.allocator,
             };
         }
```
yewman, line 48: NIT: spaces between funcs 
```suggestion
        }
        
```

src/runtime/account_loader.zig
```diff
@@ -2,73 +2,41 @@ const std = @import("std");
...
-const MockedBank = struct {
+pub const MockedAccountsDb = struct {
     allocator: std.mem.Allocator,
-    slot: Slot,
-    accounts: std.AutoArrayHashMapUnmanaged(Pubkey, sig.core.Account),
-    rent_collector: RentCollector,
-
-    fn deinit(self: *MockedBank) void {
+    accounts: std.AutoArrayHashMapUnmanaged(Pubkey, sig.core.Account) = .{},
+    fn deinit(self: *MockedAccountsDb) void {
```
yewman, line 33: NIT: 
```suggestion
    
    fn deinit(self: *MockedAccountsDb) void {
```

yewman (commented):
Great stuff! Just have a few minor nits, otherwise looks good to go imo

src/runtime/account_loader.zig
Sobeston: fixed in ad41b539b9c9696ce4e727bf086dce6f741fd394

src/runtime/account_loader.zig
Sobeston: fixed in ad41b539b9c9696ce4e727bf086dce6f741fd394

src/runtime/account_loader.zig
```diff
@@ -120,245 +83,430 @@ fn Bank(comptime kind: BankKind) type {
...
-        ) !?LoadedTransactionAccount {
-            const account = blk: {
-                if (self.account_cache.get(key.*)) |account| {
-                    // a previous transaction deallocated this account.
-                    break :blk if (account.lamports == 0) null else account;
-                }
+    // holds SYSVAR_INSTRUCTIONS_ID accounts, purely so we can deallocate them later. Index of each
+    // one isn't intended to be meaningful.
+    sysvar_instruction_account_datas: std.ArrayListUnmanaged(AccountSharedData) = .{},
+    // NOTE: we could take this field out, and have the caller deinit it from inside LoadedTransactionAccounts
```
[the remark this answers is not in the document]
Sobeston: fixed in ad41b539b9c9696ce4e727bf086dce6f741fd394

src/runtime/transaction_error.zig
```diff
@@ -0,0 +1,139 @@
...
+    };
+}
+
+/// NOTE: This type is implement multiple times in sig. We need to clean this up however this falls
+/// outside the scope of this PR. It is re-implemented here because the other definitions use the
+/// InstructionError union type within the InstructionError variant. This implementation requires the
+/// InstructionError error set for compatibility with the runtime. Ideally, we may be able to replace
+/// the TransactionError union type with a TransactionError error set however this will be dependent
+/// on coming up with a nice solution for propagating any required error state within the runtime.
+pub const TransactionError = union(enum) {
```
dnut, line 18: > This type is implement multiple times in sig. We need to clean this up however this falls outside the scope of this PR.

I don't think it's necessarily out of scope when you're introducing the technical debt in this pr.

> This implementation requires the InstructionError error set for compatibility with the runtime.

Isn't a conversion method already implemented? How do you handle custom error payloads with the error set?

> Ideally, we may be able to replace the TransactionError union type with a TransactionError error set

I don't think transaction failures should be represented with an error set because they aren't treated like a normal error and will be too easy to conflate with normal errors. Transaction failures aren't really exceptional cases or problems in the validator, they're just a type of transaction result. Also union payloads seem useful here.

src/runtime/transaction_execution.zig
```diff
@@ -0,0 +1,586 @@
...
+}
+
+/// [agave] https://github.com/firedancer-io/agave/blob/403d23b809fc513e2c4b433125c127cf172281a2/svm/src/transaction_processor.rs#L323-L324
+pub fn loadAndExecuteTransaction(
+    allocator: std.mem.Allocator,
+    transaction: *const RuntimeTransaction,
+    batch_account_cache: *BatchAccountCache,
+    environment: *const TransactionExecutionEnvironment,
+    config: *const TransactionExecutionConfig,
+) error{OutOfMemory}!TransactionResult(ProcessedTransaction) {
```
dnut, line 173: This return value is a bit confusing because you have an error case in the first layer, and then in the next layer, fees only transactions always have an error, and executed transactions can also have errors. What is the difference between TransactionResult.err and fees_only? Can you document the return value here? Alternatively, what do you think about adding `err` to ProcessedTransaction (maybe with a more specific name) and removing the TransactionResult layer? Then the type can be documented centrally and speak for itself.

src/runtime/transaction_execution.zig
```diff
@@ -0,0 +1,586 @@
...
+//     - Load accounts
+//
+// Once the accounts have been loaded, the transaction is commitable, even if its
+// execution fails.
+
+pub const StatusCache = struct {};
+pub const EpochStakes = struct {};
+
+pub const RuntimeTransaction = struct {
+    pub const Accounts = std.MultiArrayList(sig.core.instruction.InstructionAccount);
```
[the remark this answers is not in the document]
Sobeston: It helps to be able to pass in a []Pubkey to functions that may need it. It could just be a slice? But I like being able to make the fields separately. MultiArrayList(_).Slice would also work.

src/runtime/transaction_execution.zig
Sobeston: The first layer is just for OutOfMemory, in which case this is our problem, not the transaction's problem. 

But I see what you mean with the Result ProcessedTransaction stuff. I think this just mirrors how Agave's error handling is structured, might require another look

src/runtime/transaction_error.zig
Sobeston: The `TransactionError` enum that rust uses can either store either:
a) an account index
b) an instruction index
c) an instruction error (which can be store a custom u32)
So I think we can come up with a nicer implementation that integrates better with Zig's errors.

> will be too easy to conflate with normal errors. Transaction failures aren't really exceptional cases or problems in the validator

True... But I think using Zig's error types for them in the general case does make a lot of sense. e.g. Agave uses the same TransactionError type everywhere, which ends up being a bit horrible. All code that might return a transaction error uses the full type. Being able to make a type on the fly like 
```zig
error { ProgramAccountNotFound, InvalidProgramForExecution, MaxLoadedAccountsDataSizeExceeded }
```
, and have that coerce to the "full" error type is great

What if we had something like:
```zig
const TransactionError = struct {
   err: TransactionErrorSet, // error {}
   
   account_index: ?u8,
   instruction_index: ?u16,
   custom: ?u32, // null unless self.err == error.Custom
 }
 ```

dnut (approved):

PR #690, opened 2025-04-26
fix(runtime): repr check before vm translate

kprotty:
Assert that the types being translated from VM memory have a stable memory layout.


yewman (approved):
lgtm

Rexicon226 (approved):

PR #693, opened 2025-04-28
feat(consensus): Implement the vote listener

InKryption:


InKryption:
Usurped by #700 stack

PR #694, opened 2025-04-28
feat(grafana): customizable password and enable embedding

dnut:
The grafana password is now configurable in a .env file, and this is used in the run-and-update-service. GF_SECURITY_ALLOW_EMBEDDING is now enabled so public dashboards can be embedded in iframes.


Sobeston (approved):
:ok_hand:

PR #698, opened 2025-04-29
refactor(accountsdb): move manager loop to own file

dnut:
accountsdb.zig and the AccountsDB struct are overwhelmingly large. I'm writing code in replay that needs to read data from the db and it's hard to figure out where to even begin looking at this giant file to find the public facing api for loading accounts out of the database.

`AccountsDB.runManagerLoop` and the functions it calls are not ordinary functions for interacting with the database. It's really a behind-the-scenes service that runs constantly and keeps AccountsDB internally tidy, whereas AccountsDB is a data structure that other components need to read from and write to. That service and its test alone account for 1500 lines of code, contributing a lot to `accoutsdb.zig`.

I decided to split out the manager service into its own separate file. I think it's sensible to keep the data structure in one place, and the service in another. The scope of the manager thread is now more clearly delineated, and the AccountsDB struct is brought down to a more reasonable size that's easier to wrap your head around.

accountsdb.zig is still almost 4000 lines of code and I think it would be nice to break it up a bit further. I just did the manager for now because it was the most obvious candidate for splitting into a separate file, and it only took a few minutes.

This should have little impact on anything anyone else is working on since it basically only touches the manager loop in accountsdb, which no one else is currently working on.

### DIFF
--- a/src/accountsdb/db.zig
+++ b/src/accountsdb/db.zig
@@ -66,11 +66,9 @@ const LOG_SCOPE = "accounts_db";
 const ScopedLogger = sig.trace.log.ScopedLogger(LOG_SCOPE);
 
 pub const DB_LOG_RATE = sig.time.Duration.fromSecs(5);
-pub const DB_MANAGER_LOOP_MIN = sig.time.Duration.fromSecs(5);
 
 pub const MERKLE_FANOUT: usize = 16;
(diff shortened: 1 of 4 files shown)


InKryption (approved):
Yeah, this seems entirely reasonable to me, great change.

PR #699, opened 2025-04-30
refactor(core,accountsdb): Move bank to core and add basic bank structs (#686)

InKryption:
refactor(core,accountsdb): Move bank to core and add basic bank structs (#686)

The diff in this PR is a bit confusing because a ton of code was moved
and a recent PR was reverted. Here is a more readable summary of
changes:

The main idea of this PR is to make some bank fields available to the
rest of the codebase. To do this, I added three new structs to
`core/bank.zig`, and I moved most of the bank code from accountsdb into
core. The moved code has simply been copy and pasted so you can rest
assured it is exactly the same code. Here are the new structs:
https://github.com/Syndica/sig/blob/de12d5876bb8676847a6b0f2ef8ccef88b025210/src/core/bank.zig#L45-L157

I also was forced to add some unit tests to improve test coverage. This
revealed a bug in `ExtraFields.clone` where the `versioned_epoch_stakes`
field was not being cloned. Once this was fixed, it revealed a missing
errdefer in `VersionEpochStake.clone`, which I also fixed. These changes
were in 3bb8823498cff09201dad39d266308c85fd17110 and
627ee133d8bae6f75a38b3bbf9bdd21a3534cb15.

In ab829d9fba95b253df980d9225c6f402a758d07e I also reverted #687 because
the merge conflicts are severe and the changes from #687 are not needed
right now. This makes the diff for the current PR look a bit strange.
The only changes in this PR (relative to the commit before #687) are the
ones mentioned above.

## Sig's `Bank` Approach

We're not currently planning to use a `Bank` struct like agave. The
`Bank` has become a catch-all bucket containing an incredible amount of
state and methods, and it is used almost everywhere in the validator,
despite most of that state being unnecessary in most places where the
`Bank` is used. It's the ultimate spaghetti code, where everything
depends on everything. It's very hard to make sense of the dependencies.
A more granular expression of dependencies will make the code throughout
Sig a lot clearer.

The bank data needs to be represented somehow, and it's becoming
necessary for much of what we're implementing in replay and consensus.
So I implemented three structs that each contain specialized subsets of
the fields in agave's Bank:
- `SlotState`: data about a slot that changes as the slot is executed
- `SlotConstants`: unchanging data about a slot
- `EpochConstants`: unchanging data about an epoch

Other parts of agave's `Bank`, like the pointer to accountsdb, can be
managed as separate dependencies. Tons of methods should not be needed
here either. For example, the logic to execute transactions can live in
the SVM module.

This is partially inspired by firedancer's approach of having
`fd_slot_bank` separate from the `fd_epoch_bank`, and keeping accountsdb
managed separately from the banks.

If eventually we realize that we do actually need all of these things
combined together in many places, then it will be trivial to write
something like this:

```zig
pub const Bank = struct {
    accounts: *AccountsDB,
    slot_state: *SlotState,
    slot_constants: SlotConstants,
    epoch_constants: EpochConstants,
};
```

This could work just like the Bank does in agave, but at least it's
organized. I don't see a need for this right now, so I haven't included
it.

ReferenceCounter: don't use `Self`

VoteTracker MVP

### DIFF
--- a/src/consensus/lib.zig
+++ b/src/consensus/lib.zig
@@ -1,5 +1,8 @@
 pub const fork_choice = @import("fork_choice.zig");
+pub const vote_tracker = @import("vote_tracker.zig");
 
 pub const HeaviestSubtreeForkChoice = fork_choice.ForkChoice;
 pub const ForkWeight = fork_choice.ForkWeight;
 pub const ForkInfo = fork_choice.ForkInfo;
+
+pub const VoteTracker = vote_tracker.VoteTracker;
--- a/src/consensus/vote_tracker.zig
+++ b/src/consensus/vote_tracker.zig
@@ -1,0 +1,275 @@
+const std = @import("std");
+const sig = @import("../sig.zig");
+const builtin = @import("builtin");
+
+const Slot = sig.core.Slot;
+const Hash = sig.core.Hash;
+const Pubkey = sig.core.Pubkey;
+
+/// TODO: these decls should be moved into their own actual namespaces in the codebase eventually.
+const commitment = struct {
+    pub const VOTE_THRESHOLD_SIZE: f64 = 2.0 / 3.0;
+};
+
+pub const VoteTracker = struct {
+    /// Protects all access to `map`, and partial access to its elements.
+    ///
+    /// Any direct mutation to, or direct read from, `map`, must first acquire
+    /// an appropriate guard on this lock.
+    ///
+    /// Each entry of `map` is guarded by its own `rwlock`. In order to obtain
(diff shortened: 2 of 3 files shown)


InKryption:
* **#699** <a href="https://app.graphite.dev/github/pr/Syndica/sig/699?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Syndica/sig/699?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

InKryption:
Ignore this, graphite fail

PR #700, opened 2025-04-30
feat(consensus): Implement VoteTracker

InKryption:
ReferenceCounter: don't use `Self`

VoteTracker MVP

### DIFF
--- a/src/consensus/vote_tracker.zig
+++ b/src/consensus/vote_tracker.zig
@@ -1,0 +1,274 @@
+const std = @import("std");
+const sig = @import("../sig.zig");
+const builtin = @import("builtin");
+
+const Slot = sig.core.Slot;
+const Hash = sig.core.Hash;
+const Pubkey = sig.core.Pubkey;
+
+/// TODO: these decls should be moved into their own actual namespaces in the codebase eventually.
+const commitment = struct {
+    pub const VOTE_THRESHOLD_SIZE: f64 = 2.0 / 3.0;
+};
+
+pub const VoteTracker = struct {
+    /// Protects all access to `map`, and partial access to its elements.
+    ///
+    /// Any direct mutation to, or direct read from, `map`, must first acquire
+    /// an appropriate guard on this lock.
+    ///
+    /// Each entry of `map` is guarded by its own `rwlock`. In order to obtain
+    /// a guard on any of the aformentioned locks, a guard on this map lock
+    /// must first be acquired, and then the entry's lock may be acquired
+    /// before releasing the guard on this lock (but not after).
+    ///
+    /// TODO: document whether a read guard on this map lock permits acquiring
+    /// a write guard on the entry's lock lock, and other potential restrictions
+    /// and/or allowances around safely accessing any data.
+    map_rwlock: std.Thread.RwLock,
+
+    /// Map from a slot to a set of validators who have voted for that slot.
+    /// See the doc comment on `map_rwlock` for commentary on accessing this field and its contents.
+    map: Map,
+
+    pub const EMPTY: VoteTracker = .{
+        .map_rwlock = .{},
+        .map = .{},
+    };
+
+    pub const Map = std.AutoArrayHashMapUnmanaged(Slot, RwMuxSlotVoteTracker);
+    pub const RwMuxSlotVoteTracker = struct {
+        rwlock: std.Thread.RwLock,
+        tracker: SlotVoteTracker,
+
+        pub const EMPTY_ZEROES: RwMuxSlotVoteTracker = .{
+            .rwlock = .{},
+            .tracker = SlotVoteTracker.EMPTY_ZEROES,
+        };
+    };
+
+    pub fn deinit(self: *VoteTracker, allocator: std.mem.Allocator) void {
+        self.map_rwlock.lock();
+        defer self.map_rwlock.unlock();
+        const map = &self.map;
+        for (map.values()) |*svt| {
+            svt.rwlock.lock();
+            defer svt.rwlock.unlock();
+            svt.tracker.deinit(allocator);
+        }
+        map.deinit(allocator);
+    }
+
+    /// TODO: investigate why this alias even exists in the agave code
+    const progressWithNewRootBank = purgeStaleState;
+
+    /// Purge any outdated slot data
+    fn purgeStaleState(self: *VoteTracker, new_root_bank_slot: Slot) void {
+        self.map_rwlock.lock();
+        defer self.map_rwlock.unlock();
+        const map = &self.map;
+
+        var start_idx: usize = 0;
+        outer: while (start_idx != map.count()) {
+            for (map.keys()[start_idx..], start_idx..) |slot, i| {
+                if (slot >= new_root_bank_slot) continue;
+                std.debug.assert(map.swapRemove(slot));
+                start_idx = i;
+                continue :outer;
+            }
+        }
+    }
+
+    /// NOTE: intended for tests
+    pub fn insertVote(
+        self: *VoteTracker,
+        allocator: std.mem.Allocator,
+        slot: Slot,
+        pubkey: Pubkey,
+    ) !void {
+        if (!builtin.is_test) @compileError(@src().fn_name ++ " can only be used in tests.");
+
+        self.map_rwlock.lock();
+        defer self.map_rwlock.unlock();
+        const map = &self.map;
+
+        const gop = try map.getOrPut(allocator, slot);
+        errdefer if (!gop.found_existing) std.debug.assert(map.pop().key == slot);
+        if (!gop.found_existing) gop.value_ptr.* = RwMuxSlotVoteTracker.EMPTY_ZEROES;
+        const w_slot_vote_tracker = &gop.value_ptr.tracker;
+
+        map.lockPointers();
+        defer map.unlockPointers();
+
+        const voted = &w_slot_vote_tracker.voted;
+        errdefer if (!gop.found_existing) voted.deinit(allocator);
+
+        const voted_slot_updates = w_slot_vote_tracker.initAndOrGetUpdates();
+        errdefer if (!gop.found_existing) voted_slot_updates.deinit(allocator);
+
+        if (!voted.contains(pubkey)) try voted.ensureUnusedCapacity(allocator, 1);
+        try voted_slot_updates.ensureUnusedCapacity(allocator, 1);
+
+        voted.putAssumeCapacity(pubkey, true);
+        voted_slot_updates.appendAssumeCapacity(pubkey);
+    }
+};
+
+pub const SlotVoteTracker = struct {
+    /// Maps pubkeys that have voted for this slot
+    /// to whether or not we've seen the vote on gossip.
+    /// True if seen on gossip, false if only seen in replay.
+    voted: std.AutoArrayHashMapUnmanaged(Pubkey, bool),
+    optimistic_votes_tracker: std.AutoArrayHashMapUnmanaged(Hash, VoteStakeTracker),
+    voted_slot_updates: ?std.ArrayListUnmanaged(Pubkey),
+    gossip_only_stake: u64,
+
+    pub const EMPTY_ZEROES: SlotVoteTracker = .{
+        .voted = .{},
+        .optimistic_votes_tracker = .{},
+        .voted_slot_updates = null,
+        .gossip_only_stake = 0,
+    };
+
+    pub fn deinit(self: SlotVoteTracker, allocator: std.mem.Allocator) void {
+        var copy = self;
+        copy.voted.deinit(allocator);
+        copy.optimistic_votes_tracker.deinit(allocator);
+        copy.initAndOrGetUpdates().deinit(allocator);
+    }
+
+    /// If there already are vote slot updates, returns the pointer to them.
+    /// Otherwise, initializes them, and then returns the pointer.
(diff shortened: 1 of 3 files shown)


InKryption:
* **#700** <a href="https://app.graphite.dev/github/pr/Syndica/sig/700?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Syndica/sig/700?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/consensus/vote_tracker.zig
```diff
@@ -0,0 +1,343 @@
...
+    pub fn getOrInsertSlotTracker(
+        self: *VoteTracker,
+        allocator: std.mem.Allocator,
+        slot: Slot,
+    ) std.mem.Allocator.Error!*RcRwSlotVoteTracker {
+        blk: {
+            self.map_rwlock.lockShared();
+            defer self.map_rwlock.unlockShared();
+            const rc_rw_svt = self.map.get(slot) orelse break :blk;
+            // we acquired a lock on the map, this shouldn't be possible
```
dadepo, line 88: Looks like this comment is not correct? Given we retrieved the `rc_rw_svt` and then we call rc_rw_svt.rc.acquire().

src/consensus/vote_tracker.zig
```diff
@@ -0,0 +1,343 @@
...
+const Slot = sig.core.Slot;
+const Hash = sig.core.Hash;
+const Pubkey = sig.core.Pubkey;
+
+/// TODO: these decls should be moved into their own actual namespaces in the codebase eventually.
+const commitment = struct {
+    pub const VOTE_THRESHOLD_SIZE: f64 = 2.0 / 3.0;
+};
+
+pub const VoteTracker = struct {
```
dadepo, line 14: Noticed the implementation does not include `get_slot_vote_tracker`. Is that intentional?

src/consensus/vote_tracker.zig
```diff
@@ -0,0 +1,343 @@
...
+
+    pub const EMPTY_ZEROES: SlotVoteTracker = .{
+        .voted = .{},
+        .optimistic_votes_tracker = .{},
+        .voted_slot_updates = null,
+        .gossip_only_stake = 0,
+    };
+
+    pub fn deinit(self: SlotVoteTracker, allocator: std.mem.Allocator) void {
+        var copy = self;
```
dadepo, line 193: Is this meant to work based on the assumption that `SlotVoteTracker` would be passed in as a shallow copy? and hence the fields can still be reached and de-allocated?

src/consensus/vote_tracker.zig
```diff
@@ -0,0 +1,343 @@
...
+
+        if (!voted.contains(pubkey)) try voted.ensureUnusedCapacity(allocator, 1);
+        try voted_slot_updates.ensureUnusedCapacity(allocator, 1);
+
+        voted.putAssumeCapacity(pubkey, true);
+        voted_slot_updates.appendAssumeCapacity(pubkey);
+    }
+};
+
+pub const SlotVoteTracker = struct {
```
dadepo, line 176: Seems this is missing `optimistic_votes_tracker` is that intentional?

dadepo (commented):
Looks good in general. Left a couple of comments/questions.

src/consensus/vote_tracker.zig
InKryption: That function is just `svt.optimistic_votes_tracker.get(hash)`, didn't see a point in adding it.

src/consensus/vote_tracker.zig
InKryption: I have it implemented locally on the top PR, I've pulled it down here now:
a69ed52cb91ace5bea5c7399a0476aac785cecfa

src/consensus/vote_tracker.zig
InKryption: To be clear, the intended reading of the comment is that the assertion shouldn't be triggered. Is that what you're saying is incorrect, or is the comment badly phrased?

src/consensus/vote_tracker.zig
```diff
@@ -0,0 +1,343 @@
...
+    /// before releasing the guard on this lock (but not after).
+    ///
+    /// TODO: document whether a read guard on this map lock permits acquiring
+    /// a write guard on the entry's lock lock, and other potential restrictions
+    /// and/or allowances around safely accessing any data.
+    map_rwlock: std.Thread.RwLock,
+
+    /// Map from a slot to a set of validators who have voted for that slot.
+    /// See the doc comment on `map_rwlock` for commentary on accessing this field and its contents.
+    map: Map,
```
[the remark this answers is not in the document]
InKryption: None at all, just a relic of an old changeset: 72b960ed41ad774b4cb46291c34e519c6243f62c

src/consensus/vote_tracker.zig
```diff
@@ -0,0 +1,343 @@
...
+
+    pub const EMPTY: VoteTracker = .{
+        .map_rwlock = .{},
+        .map = .{},
+    };
+
+    pub const Map = std.AutoArrayHashMapUnmanaged(Slot, *RcRwSlotVoteTracker);
```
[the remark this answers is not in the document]
InKryption: 72b960ed41ad774b4cb46291c34e519c6243f62c
was just a relic of an old changeset

src/consensus/vote_tracker.zig
```diff
@@ -0,0 +1,343 @@
...
+                if (slot >= new_root_bank_slot) continue;
+                map.fetchSwapRemove(slot).?.value.deinit(allocator);
+                start_idx = i;
+                continue :outer;
+            }
+        }
+    }
+
+    /// NOTE: intended for tests
+    pub fn insertVote(
```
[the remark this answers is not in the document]
InKryption: 55a38223fc423ed972d6f7f157ffc61ad2c971f8

src/consensus/vote_tracker.zig
```diff
@@ -0,0 +1,343 @@
...
+
+        var start_idx: usize = 0;
+        outer: while (start_idx != map.count()) {
+            for (map.keys()[start_idx..], start_idx..) |slot, i| {
+                if (slot >= new_root_bank_slot) continue;
+                map.fetchSwapRemove(slot).?.value.deinit(allocator);
+                start_idx = i;
+                continue :outer;
+            }
+        }
```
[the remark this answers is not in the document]
InKryption: > Also, does fetchSwapRemove actually invalidate the slice? I assumed it would only be for operations that trigger a reallocation. Maybe keys can just be set once at the beginning.

Yeah. Whilst removing an entry doesn't trigger a reallocation, it does shrink the slices by one - but if we don't update our view of the keys slice, we'll continue to observe a slice that includes the elements that are invalidated at the end, and thus end up operating on undefined memory in the last iterations if any elements were removed.

> Another thing, do you know why fetchSwapRemoveAt does not exist?
I'm not sure, might just be an oversight. Could probably implement the behaviour via an adapted context with `fetchSwapRemoveAdapted`.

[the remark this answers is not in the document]
InKryption: 13ce4ccb42e49a01d251ebbd49b4a86a9037d5b6

src/consensus/vote_tracker.zig
```diff
@@ -0,0 +1,343 @@
...
+            // and in theory this is the only remaining reference anyway.
+            const tracker, _ = self.tracker.writeWithLock();
+            tracker.deinit(allocator);
+            allocator.destroy(self);
+        }
+    };
+
+    pub fn deinit(self: *VoteTracker, allocator: std.mem.Allocator) void {
+        self.map_rwlock.lock();
+        defer self.map_rwlock.unlock();
```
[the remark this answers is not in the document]
InKryption: This wouldn't really avoid UB so much as move it from the sites which still incorrectly hold a reference to the deinit function. At least in the former case we could hope to get a segfault pointing to the invalid usage of the map, rather than an assertion failure in the deinit that could have resulted from any number of other sites in the codebase continuing to hold a reference.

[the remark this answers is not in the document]
InKryption: Also important to note, if we reach a point where we are deinitializing the vote tracker, the validator is exiting anyway, and we may want to allow other threads to finish whatever they were doing with the vote tracker correctly before cleaning up and exiting, rather than crashing the cleanup of the vote tracker.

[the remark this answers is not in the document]
dnut: I didn't say this would help avoid UB. My point was that it's the responsibility of the caller not to deinit things that are still in use. If you follow that practice, there is no need to acquire the lock here. Acquiring the lock is not meaningful if you already know there are not any other threads using this. That was my main point. You could just delete these lines because they're unnecessary.

That being said, there could of course be a bug in the code, with other threads actually using this while it's being deinitialized. But that's a bug that should be fixed. Acquiring the lock here means you are *less* likely to interfere with whatever that other thread is doing, so you are less likely to see errors. That may have an advantage, but it's a workaround that hides the bug. Personally I think I'd rather raise the alarm in this case. If you want to do that by seeing segfaults where it's still being used, then you should just delete the lock acquisition and deinit immediately. If you want to raise alarms with assertion failures in the deinit function, then you would put an assertion here.

Regarding the decision between segfault and assertion, the segfault may have more information per event (the stack trace), versus the assertion which has a higher likelihood of being triggered when the code is incorrect. I'd rather have the assertion as the first line of defense, so at least I know there's an issue, then I always have the option to run it again with the assertion removed to see where segfaults take place.

src/consensus/vote_tracker.zig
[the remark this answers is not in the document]
dnut: I put `keys = map.keys();` within the specific branch so it doesn't need to compute the slice on every iteration, only the ones where the slice would actually change

src/consensus/vote_tracker.zig
dnut: I think it's safe to assume that it will always return true here. If it returns false, that means there's another function that's destroying the item without removing it from the map, or another function is not respecting the lock. Either of those sound like bugs that assertions like this are supposed to help us identify.

In general, for typical reference counting use cases (where the counter itself gets destroyed with the resource), the return value of this function should never be false. If it's false, that means you have already triggered undefined behavior. You're actually lucky if it returns false instead of segfaulting or something worse.

InKryption: Well, there's also the possibility that `rc.release()` is called one too many times, without actually freeing any memory, which isn't that far-fetched.

src/consensus/vote_tracker.zig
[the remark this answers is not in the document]
InKryption: That doesn't seem that important, and seems more brittle to refactors.

src/consensus/vote_tracker.zig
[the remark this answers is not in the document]
InKryption: 4a5b20e455151d368dca9bdb602a43d26bc34c6f

src/consensus/vote_tracker.zig
dnut: good point

dnut (dismissed):

src/consensus/vote_tracker.zig
dadepo: > or is the comment badly phrased?

I think having something like "this assert should not fail" instead of "this shouldn't be possible" may be more clearer?

src/consensus/vote_tracker.zig
dadepo: Seems a comment was unintentionally committed

> // self.slot_vote_trackers.read().unwrap().get(&slot).cloned()

src/consensus/vote_tracker.zig
dadepo: I was actually asking about how self is passed in. ie `deinit(self: SlotVoteTracker..`? 

Was wondering about having `pub fn deinit(self: *SlotVoteTracker` instead? and why would that be a bad/good approach.

dadepo (dismissed):

src/consensus/vote_tracker.zig
InKryption: Passing it in mutably would mean its mutability is viral. I prefer mutability not be required to deinitialize something is my point.

src/consensus/vote_tracker.zig
InKryption: b8f0cafe528832ab9e34c3f9a4824e80855702ee

src/consensus/vote_tracker.zig
dadepo: The options available I believe are:

- Passing it as a mutable ref
- Passing it as an immutable ref
- Passing it as an immutable copy

And in your approach, you are using the last approach. My question was about how that works (and if that is safe) since it is being deinitialize via a copy.

InKryption: Yeah, it's perfectly safe. The pointers passed to `allocator.free` under the hood are still the same pointers, the only reason those methods take in a mutable reference to the data structures is to do `self.* = undefined`, which I view as pointless, and is the subject of debate as to whether that should even be the convention.
See `std.ArrayList(T).deinit`, which in fact doesn't do this, representing the current split in the debate.

dadepo (approved):

PR #704, opened 2025-04-30
feat(core): transaction and entry verification and hashing

dnut:
This adds some core functionality for verifying entries and transactions that is needed by replay.

- verify transactions
- hash transactions
- verify poh chain of entries
- verify ticks of entries

I also consolidated the various approaches to sha256 hashing into a single function `Hash.generateSha256`. Most of the files changed were to use this function.


src/core/transaction.zig
```diff
@@ -278,6 +345,16 @@ pub const TransactionMessage = struct {
...
         }
     }
+
+    /// Return the blake3 hash of the pre-serialized message.
+    pub fn hash(serialized_message: []const u8) Hash {
+        var hasher = Blake3.init(.{});
+        hasher.update("solana-tx-message-v1");
+        hasher.update(serialized_message);
+        var the_hash: Hash = undefined;
+        hasher.final(&the_hash.data);
```
InKryption, line 355: very minor nit: would be nice tor to use `var the_hash: Hash = .{ .data = undefined };` here, to avoid the uncertainty of undefined padding. I know `Hash` is an extern struct which has zero padding outside its singular `data` field, but I did have to double take before I figured that, so this change would probably be useful for a less familiar auditor.

dnut: 9df975fd3a9a26c76e6cd09bd8b4caa8598b771e

I never actually thought about this before, but I see what you mean. Is there a specific problem that might arise from having undefined padding? Does the compiler normally set the padding to a specific value that is used at runtime? Or are you just saying the compiler reserves the right to utilize the padding as it sees fit, so we shouldn't mess with it.

InKryption: In the case of an extern struct like this, there's no real threat aside from the concern that might arise before realizing that it's extern. For auto layout structs, what you say is exactly the case, the compiler reserves the right to use the padding, so it should be in a valid state. It's similar to how it's a footgun to initialize a sentinel array directly using `undefined`, since it will also make the sentinel value undefined, versus using `@splat(undefined)`.

Also contentious for packed structs; while they don't have padding, my understanding is that any undefined bits mean the whole struct is undefined, since the compiler/optimizer doesn't/can't distinguish between a value having a subset of bits be undefined, and being entirely undefined, with the current memory model, so best to steer clear of initializing them to undefined.

Also can be problematic for exotic integers, and by extension packed structs which are backed by exotic integers, being that these do having padding also reserved by the compiler.

InKryption (approved):
LGTM

PR #713, opened 2025-05-02
refactor(threadpool): consolidate thread pool usages

dnut:
This consolidates the functionality from `ThreadPoolTask` with `HomogeneousThreadPool` since I need features from both in replay. Also, they are redundant - it's confusing how there are two ways to do almost the same thing. 

I added a couple features to `HomogeneousThreadPool` that were in use with `ThreadPoolTask`, and deleted `ThreadPoolTask`:
- Allow usage with a pre-existing `ThreadPool` instead of always creating its own. I specifically wanted this for replay. This is implemented with `initBorrowed`
- Constrain the number of concurrent tasks to something less than the number of threads in the pool. As a simplification, I implemented this a little different than `ThreadPoolTask`. I just track the limit with an integer and keep throwing tasks into the same array list instead of monitoring a bunch of tasks in a buffer and popping out results as soon as they're done. I think this makes it easier to manage the results with because you can handle them all at the end instead of needing to conditionally handle some in the loop and then the rest at the end.

### DIFF
--- a/src/gossip/service.zig
+++ b/src/gossip/service.zig
@@ -26,7 +26,7 @@ const Counter = sig.prometheus.Counter;
 const Gauge = sig.prometheus.Gauge;
 const Histogram = sig.prometheus.Histogram;
 const GetMetricError = sig.prometheus.registry.GetMetricError;
-const ThreadPoolTask = sig.utils.thread.ThreadPoolTask;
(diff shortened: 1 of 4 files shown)


InKryption (approved):
I quite like the unification in this PR, I'm happy to merge, but curious if @kprotty has any comments or concerns.

PR #714, opened 2025-05-03
fix(runtime): vote program conformance

yewman:
Fixes for vote conformance

### DIFF
--- a/src/runtime/program/vote/state.zig
+++ b/src/runtime/program/vote/state.zig
@@ -141,6 +141,65 @@ pub const VoteStateUpdate = struct {
     }
 };
 
+pub fn serializeCompactVoteStateUpdate(writer: anytype, data: anytype, _: sig.bincode.Params) anyerror!void {
+    // Calculate lockout offsets
+    var slot = data.root orelse 0;
+    var lockouts = std.BoundedArray(struct { Slot, u8 }, MAX_LOCKOUT_HISTORY){};
+    for (data.lockouts.items) |lockout| {
+        lockouts.appendAssumeCapacity(.{
+            try std.math.sub(Slot, lockout.slot, slot),
+            @intCast(lockout.confirmation_count),
+        });
+        slot = lockout.slot;
+    }
+
+    // Serialize in compact format
+    try writer.writeInt(Slot, slot, std.builtin.Endian.little);
+    try sig.bincode.varint.serialize_short_u16(writer, @intCast(lockouts.len), .{});
+    for (lockouts.constSlice()) |lockout| {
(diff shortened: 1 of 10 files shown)


src/utils/collections.zig
```diff
@@ -818,44 +818,6 @@ pub fn Window(T: type) type {
     };
 }
 
-/// Analogous to [CircBuf](https://github.com/anza-xyz/solana-sdk/blob/e1554f4067329a0dcf5035120ec6a06275d3b9ec/vote-interface/src/state/vote_state_0_23_5.rs#L44)
-/// `RingBuffer` is a fixed-size circular buffer (ring buffer) implementation.
-/// It stores a fixed number of elements of type `I` and overwrites the oldest elements
-/// when the buffer is full.
-pub fn RingBuffer(comptime I: type, comptime Size: usize) type {
```
dadepo, line 825: The ring buffer data structure might be coming in handy somewhere else. See https://github.com/orgs/Syndica/projects/2/views/10?pane=issue&itemId=108874668

So might still be worth it to have it in the src/utils/collections path to prevent duplication

(except the implementation used in vote program has requirement peculiar to it, that makes it not re-usable)

yewman: I would leave it as is until the use case becomes apparent, then can review pulling it out if it makes sense

src/runtime/program/vote/state.zig
```diff
@@ -117,6 +116,81 @@ pub const VoteStateUpdate = struct {
     timestamp: ?i64,
 };
 
+pub fn serializeCompactVoteStateUpdate(
+    writer: anytype,
+    data: anytype,
```
Rexicon226, line 121: Why not just make `data: VoteStateUpdate`?

Rexicon226 (changes requested):

src/runtime/program/vote/state.zig
yewman: because its used by bincode field config it needs to be anytype

Rexicon226: anytype isn't a type, it's a form of parametric polymorphism. theentire point is that it's a top type.

yewman: changing to VoteStateUpdate causes a transitive error, so I must be doing something else wrong then

Rexicon226: Ah, this isn't a compilation error, this is the compiler crashing. I already fixed this bug in the master branch of Zig, so we can update it later.

Rexicon226 (approved):

src/utils/collections.zig
dadepo: That works.

PR #715, opened 2025-05-03
feat(consensus): Implement subset of vote_parser code for vote listener

InKryption:
Add deinit functions & turn functions into consts

Implement subset of vote parser code

### DIFF
--- a/src/consensus/lib.zig
+++ b/src/consensus/lib.zig
@@ -7,6 +7,9 @@ pub const tower_storage = @import("tower_storage.zig");
 pub const unimplemented = @import("unimplemented.zig");
 pub const vote_tracker = @import("vote_tracker.zig");
 pub const vote_transaction = @import("vote_transaction.zig");
+comptime {
+    _ = @import("vote_parser.zig");
+}
 
 pub const HeaviestSubtreeForkChoice = fork_choice.ForkChoice;
 pub const ForkWeight = fork_choice.ForkWeight;
--- a/src/consensus/vote_parser.zig
+++ b/src/consensus/vote_parser.zig
@@ -1,0 +1,159 @@
+//! Based on https://github.com/anza-xyz/agave/blob/182823ee353ee64fde230dbad96d8e24b6cd065a/vote/src/vote_parser.rs
+
+// TODO: this is probably/definitely the wrong place for this file to be,
+// but it's the only place it's needed right now, and I don't feel like
+// committing to a solid structure yet.
+
+const std = @import("std");
+const sig = @import("../sig.zig");
+
+const vote_program = sig.runtime.program.vote;
+
+const Slot = sig.core.Slot;
+const Hash = sig.core.Hash;
+const Pubkey = sig.core.Pubkey;
+const Signature = sig.core.Signature;
+const Transaction = sig.core.Transaction;
+
+// #[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
+pub const VoteTransaction = union(enum) {
+    vote: sig.runtime.program.vote_program.state.Vote,
+    vote_state_update: sig.runtime.program.vote_program.state.VoteStateUpdate,
+    // #[serde(with = "serde_compact_vote_state_update")]
+    compact_vote_state_update: sig.runtime.program.vote_program.state.VoteStateUpdate,
+    // #[serde(with = "serde_tower_sync")]
+    tower_sync: sig.runtime.program.vote_program.state.TowerSync,
+
+    pub fn deinit(self: VoteTransaction, allocator: std.mem.Allocator) void {
+        switch (self) {
+            .vote => |vote| vote.deinit(allocator),
+            .vote_state_update => |vsu| vsu.deinit(allocator),
+            .compact_vote_state_update => |cvsu| cvsu.deinit(allocator),
+            .tower_sync => |ts| ts.deinit(allocator),
+        }
+    }
+
+    pub fn lastVotedSlot(self: VoteTransaction) ?Slot {
+        return switch (self) {
+            .vote => |vote| if (vote.slots.len != 0) vote.slots[vote.slots.len - 1] else null,
+            inline //
+            .vote_state_update,
+            .compact_vote_state_update,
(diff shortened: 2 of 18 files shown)


InKryption:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Syndica/sig/715?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#715** <a href="https://app.graphite.dev/github/pr/Syndica/sig/715?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Syndica/sig/715?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#700** <a href="https://app.graphite.dev/github/pr/Syndica/sig/700?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/consensus/vote_parser.zig
```diff
@@ -0,0 +1,126 @@
...
+    switch_proof_hash: ?Hash,
+    signature: Signature,
+
+    pub fn deinit(self: ParsedVote, allocator: std.mem.Allocator) void {
+        self.vote.deinit(allocator);
+    }
+};
+
+// Used for parsing gossip vote transactions
+pub fn parseVoteTransaction(
```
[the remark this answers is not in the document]
InKryption: Implemented as part of the PR stacked on top of this one, #743, which includes other related changes that will also assist in porting of other tests from agave similar to this one.

src/consensus/vote_parser.zig
```diff
@@ -0,0 +1,126 @@
...
+    allocator: std.mem.Allocator,
+    tx: Transaction,
+) std.mem.Allocator.Error!?ParsedVote {
+    // Check first instruction for a vote
+    const message = tx.msg;
+
+    if (message.instructions.len == 0) return null;
+    const first_instruction = message.instructions[0];
```
dnut, line 39: Is this a requirement for vote transactions? It must be in the first instruction?

dnut (dismissed):
Looks good, but I think this should merge concurrently with its test coverage.

src/consensus/vote_parser.zig
InKryption: Very much appears so: https://github.com/anza-xyz/agave/blob/640a749ea599a05a1e2e553ed70f835752cb68e8/vote/src/vote_parser.rs#L25-L39

src/consensus/vote_parser.zig
```diff
@@ -0,0 +1,286 @@
...
+            vote_tx_msg,
+            &.{},
+        );
+    };
+    defer vote_tx.deinit(allocator);
+    try std.testing.expectEqual(null, parseVoteTransaction(allocator, vote_tx));
+}
+
+/// Reimplemented locally from Vote program.
+fn newVoteTransaction(
```
dadepo, line 211: Nit: if these functions are meant only for testing, it would be nice to have `builtin.is_test` guard for them.

dadepo (dismissed):

src/consensus/vote_parser.zig
InKryption: 8fa5bfca7a900106b2c7911333f92aaa01d6227a

dadepo (dismissed):

dnut (dismissed):

src/consensus/lib.zig
```diff
@@ -7,6 +7,9 @@ pub const tower_storage = @import("tower_storage.zig");
 pub const unimplemented = @import("unimplemented.zig");
 pub const vote_tracker = @import("vote_tracker.zig");
 pub const vote_transaction = @import("vote_transaction.zig");
+comptime {
```
dadepo, line 10: Nit: I think David recently had a PR where these kind of `comptime` where removed. So if it can be avoided might be an idea to drop it now

dadepo: Not a blocker so resolving this so as not to block merging

dadepo (approved):

src/consensus/lib.zig
InKryption: Unfortunately, as it stands, this is needed in order to run the vote parser tests without actually exposing its API here. Which, I would rather not expose it, as it's not a full vote parser API, it features only what is needed for implementing tests and other bits; at some point in the future it will make sense to extract it when we get around to porting that set of features.

dnut: This is how we do imports. Every file is exposed as a `pub const` in `lib.zig`, and then you can access them through their qualified path `sig.consensus.vote_parser.whatever`.

Including or excluding it in lib.zig doesn't actually change anything about how exposed the API is. The API is exposed regardless because the file can always be directly imported. lib.zig just serves as a better way of importing the same thing.

If we want to define a constrained public API for consumers of sig as a module, then we should define some separate files for that. The current approach is not set up with that in mind. Every single file is exposed. Going against that pattern for new files is just going to introduce chaos.

InKryption: That's not entirely accurate, there is precedent for internal modules in the RPC server. If the issue is centralization, can take a similar approach as there and have an internal-only module importable by peers, but not by lib.zig.

I don't think "expose everything" is a sound design principle, it creates the potential for an implicit dependency between two things which were never intended to be interlinked.

> Including or excluding it in lib.zig doesn't actually change anything about how exposed the API is. The API is exposed regardless because the file can always be directly imported. lib.zig just serves as a better way of importing the same thing.

This is false in a couple ways:
* Like you say, the way by which we reference things in the API is by qualified namespaces, not direct imports, so creating such a delineation is enforced by convention.
* If sig is to be consumable as a library, what modules are and are not exposed has some implications for the contracts formed with its consumers, because they cannot just import the files relative to the root, they are constrained to the qualified namespaces.

dnut (approved):

InKryption:
Replaced with #776 to try and get around the codecov weirdness.

PR #719, opened 2025-05-04
feat(svm): implement `sol_alt_bn128_*`

Rexicon226:
Pure-zig impl of alt-bn128 (bn254) following close to Firedancer's code, since they have done all of the interesting and hard investigative work to determine a close-to-optimal implementation. In the future I have a few interesting idea I'd like to explore but it's probably out of scope for this PR.


Rexicon226:
Putting this on hold for a bit, due to libff's sqrt implementation being a bit buggy. Will continue after progress has been made on more critical tasks, such as the zksdk.

kprotty (approved):

PR #722, opened 2025-05-06
feat(replay): basic loop, load blocks, validations

dnut:


src/core/poh.zig
```diff
@@ -0,0 +1,205 @@
...
+    /// Number of hashes since the last entry
+    num_hashes: u64,
+    /// Static number of hashes expected between tick entries
+    hashes_per_tick: u64,
+    /// Current remaining number of hashes until the next tick
+    remaining_hashes: u64,
+    /// Counter for the number of ticks that happened so far
+    tick_count: u64,
+
+    const Self = @This();
```
Rexicon226: Use `Poh` direct with a `self` param name.

src/core/genesis_config.zig
```diff
@@ -57,6 +61,70 @@ pub const FeeRateGovernor = struct {
...
+        var self = base.*;
+
+        if (self.target_signatures_per_slot > 0) {
+            // lamports_per_signature can range from 50% to 1000% of
+            // target_lamports_per_signature
+            self.min_lamports_per_signature = @max(1, self.target_lamports_per_signature / 2);
+            self.max_lamports_per_signature = self.target_lamports_per_signature * 10;
+
+            // What the cluster should charge at `latest_signatures_per_slot`
+            const desired_lamports_per_signature =
```
Rexicon226: I wonder if we can make these chains of min-max calls a bit more readable by moving them into a few more well-named variables. Just a thought.

src/core/leader_schedule.zig
```diff
@@ -119,6 +119,17 @@ pub const LeaderSchedule = struct {
         self.allocator.free(self.slot_leaders);
     }
 
+    pub fn slotLeaders(self: *const LeaderSchedule, comptime first_slot: Slot) SlotLeaders {
+        return SlotLeaders.init(@as(*LeaderSchedule, @constCast(self)), struct {
```
Rexicon226: No `@constCast`. I really don't see a purpose in having `SlotLeaders` be a runtime interface, all usages can be easily converted into comptime interfaces and we can avoid making very strange hacks like this one. The current approach is going against our goal of readability.

src/replay/confirm_slot.zig
```diff
@@ -0,0 +1,555 @@
...
+            .max_tick_height = poh.tick_count,
+            .tick_height = 0,
+            .slot_is_full = true,
+            .tick_hash_count = &tick_hash_count,
+        },
+    );
+    defer future.destroy();
+
+    const result = try testAwait(future);
+    errdefer std.debug.print("failed with: {any}\n", .{result});
```
Rexicon226: Do we need these debug prints?

src/replay/confirm_slot.zig
```diff
@@ -0,0 +1,555 @@
...
+    try std.testing.expectEqual(
+        ConfirmSlotStatus{ .err = .{ .invalid_transaction = .SignatureFailure } },
+        result,
+    );
+}
+
+pub fn testAwait(
+    future: anytype,
+) !ret: {
```
Rexicon226:
```suggestion
) @TypeOf(future.poll()) {
```

src/replay/confirm_slot.zig
```diff
@@ -0,0 +1,555 @@
...
+
+    const result = try testAwait(future);
+    errdefer std.debug.print("failed with: {any}\n", .{result});
+    try std.testing.expectEqual(
+        ConfirmSlotStatus{ .err = .{ .invalid_block = .InvalidEntryHash } },
+        result,
+    );
+}
+
+test "confirmSlot fail: sigverify" {
```
Rexicon226: For future reference, you really don't need `confirmSlot` in the name of the test, the `confirm_slot.zig` namespace already provides it.

src/replay/resolve_lookup.zig
```diff
@@ -43,11 +43,15 @@ pub const ResolvedTransaction = struct {
     accounts: std.MultiArrayList(InstructionAccount),
     instructions: []const InstructionInfo,
 
-    pub fn deinit(self: ResolvedTransaction, allocator: Allocator) void {
+    /// The pointers in the transaction are not freed because it is assumed to
+    /// be borrowed. The transaction is returned here to clarify that to the
+    /// caller, and to give them the option to free the transaction if it's no
+    /// longer needed.
+    pub fn deinit(self: ResolvedTransaction, allocator: Allocator) Transaction {
```
Rexicon226: Could we be storing `transaction` by reference here if it's borrowed?

src/replay/scheduler.zig
```diff
@@ -0,0 +1,294 @@
...
+        try Transaction.initRandom(allocator, rng.random()),
+        try Transaction.initRandom(allocator, rng.random()),
+        try Transaction.initRandom(allocator, rng.random()),
+        try Transaction.initRandom(allocator, rng.random()),
+        try Transaction.initRandom(allocator, rng.random()),
+    };
+    defer for (transactions) |tx| tx.deinit(allocator);
+
+    {
+        const batch1 = try replay.resolve_lookup
```
Rexicon226: This happens in a lot of places throughout the code, but instead of putting the `.resolveBatch` on the next line, which I personally think is less readable, you should be doing 
```zig
const batch1 = try replay.resolve_lookup.resolveBatch(
    allocator,
    undefined,
    &unresolved_batch,
);
```
with a trailing-comma after the last argument.

src/replay/scheduler.zig
```diff
@@ -0,0 +1,294 @@
...
+    var rng = std.Random.DefaultPrng.init(0);
+
+    var thread_pool = ThreadPool.init(.{});
+    var scheduler = try TransactionScheduler.initCapacity(allocator, 10, &thread_pool);
+    defer scheduler.deinit();
+
+    const tx = try Transaction.initRandom(allocator, rng.random());
+    defer tx.deinit(allocator);
+
+    const unresolved_batch = .{ tx, tx };
```
Rexicon226: Give this an array result location,
```suggestion
    const unresolved_batch: [2]Transaction = .{ tx, tx };
```
or just
```suggestion
    const unresolved_batch: []const Transaction = &.{ tx, tx };
```

src/replay/confirm_slot.zig
```diff
@@ -0,0 +1,555 @@
...
+};
+
+/// Tracks the state of a slot confirmation execution.
+///
+/// Do not share across threads.
+///
+/// agave: confirm_slot and confirm_slot_entries
+/// fd: runtime_process_txns_in_microblock_stream
+pub const ConfirmSlotFuture = struct {
+    allocator: Allocator,
```
Rexicon226: Why does it need to be managed?

Rexicon226 (changes requested):
Overall great stuff, have a few comments though :)

src/replay/confirm_slot.zig
dnut: When the test fails, it helps to understand what went wrong. `expectEqual` doesn't show as much information.

src/replay/scheduler.zig
dnut: I'm aware of this approach and deliberately avoid it. I'm surprised you prefer to split up the arguments on separate lines when it's not necessary. With consolidated function calls, I find it easier to follow the high level flow. When most lines of code are dedicated to arguments, I find it noisy and distracting.

src/core/poh.zig
dnut: you caught me recycling some old code ecd1c695222d639c1950562157072239ab34116f

src/core/genesis_config.zig
dnut: I agree it is confusing and can be improved.

This calculation needs to be identical to agave, and I'm not aware of the reasoning behind it, so I ported it as directly as possible without putting much thought into the meaning of every operation.

src/core/leader_schedule.zig
dnut: Assuming we stay with a runtime interface, I think constCast is justified here. It's only working around an API limitation and not being used to mutate a const pointer. You can see on the next line of code that it's cast right back to a const pointer in its only usage.

By comptime interface, you mean passing down an additional parameter to indicate the type? or using anytype? 

> The current approach is going against our goal of readability.

I'm open to other approaches but I think they all come with some kind of readability tradeoff. Passing extra parameters is cumbersome, and anytype makes the interface unclear. It feels like a toss-up to me.

src/replay/confirm_slot.zig
dnut: Wow I didn't know this was legal zig. It feels semantically wrong somehow, like it shouldn't be allowed. Anyway, it's better than the mess I wrote. 0fd7758fe5feb5fa9968d79315cc001c03a1c36a

src/replay/resolve_lookup.zig
dnut: Harold asked a similar question on the pr that introduced this: https://github.com/Syndica/sig/pull/769#discussion_r2116643542

src/replay/confirm_slot.zig
dnut: I actually made a very similar comment on a previous pr: https://github.com/Syndica/sig/pull/769#discussion_r2116599199

In this case I was torn because there are other functions in this file aside from confirmSlot that we could also add tests for, and this clarifies specifically what it's testing within the file.

src/replay/scheduler.zig
dnut: What's the benefit of this?

src/accountsdb/buffer_pool.zig
```diff
@@ -883,13 +883,16 @@ pub const AccountDataHandle = union(enum) {
         self: *const AccountDataHandle,
         start: FileOffset,
         buf: []u8,
-    ) void {
+    ) u32 {
```
InKryption, line 889: What's the point of returning the length here? It appears that it will always copy `self.len()` bytes to `buf[0..self.len()]`.

src/core/bank.zig
```diff
@@ -110,8 +145,55 @@ pub const SlotState = struct {
     /// The value is only meaningful after freezing.
     accounts_lt_hash: sig.sync.Mux(LtHash),
 
-    pub fn isFrozen(self: *const SlotState) bool {
-        return self.hash.read().get() != null;
+    pub const GENESIS = SlotState{
```
InKryption, line 145:
```suggestion
    pub const GENESIS: SlotState = .{
```

src/replay/confirm_slot.zig
dnut: I guess it doesn't need to be. I did this to be defensive since it's going to cross multiple contexts, but it's not really necessary. I don't feel strongly either way.

472f374ea3494bb6be40afb7404448e7a46b0daa

src/sync/mux.zig
```diff
@@ -270,6 +284,24 @@ pub fn RwMux(comptime T: type) type {
...
+
+        /// Acquires the lock just long enough to clone the item using the
+        /// item's `clone` method, and returns the clone.
+        ///
+        /// Will not compile unless T.clone(T, Allocator) exists.
+        pub fn readClone(self: *Self, allocator: std.mem.Allocator) !T {
+            self.private.r.lockShared();
+            defer self.private.r.unlockShared();
+            return self.private.v.clone(allocator);
+        }
```
InKryption: I don't think this method belongs on the rwmux, this creates an implicit, optional contract that the user must then understand, both in how this operates on the protected data, and in the semantics of how the `clone` method of the protected data works.
I see exactly one usage of this, for which I think is no worse for simply inlining this logic:
```diff
-const foo = try rwmux.readClone(allocator);
+const foo = foo: {
+    var lg = rwmux.read();
+    defer lg.unlock();
+    break :foo try lg.get().clone();
+};
```
if there is a widespread need for this, I think it would be better suited as its own function, outside of the Lock wrappers.

src/utils/thread.zig
```diff
@@ -250,6 +252,20 @@ pub fn HomogeneousThreadPool(comptime TaskType: type) type {
             return true;
         }
 
+        /// Checks if all tasks are complete.
+        /// Returns a result indicating the outcome:
+        /// - done: all succeeded.
+        /// - pending: some are still running.
+        /// - err: all completed, and at least one failed.
+        pub fn pollFallible(self: *Self) union(enum) { done, pending, err: anyerror } {
```
InKryption: Let's not propagate anyerror. If the task wants to return some specific error, it can set a value in its context pointer or something, and inform the caller of the error that way. I would suggest either changing the err tag to also be void and allow the caller to do `.err => return error.PollFailed` or `.err => return task_ctx.err.?`, or returning `error{PollFailed}!enum { done, pending }`.

src/core/entry.zig
```diff
@@ -15,15 +15,23 @@ pub const Entry = struct {
...
-        return self.transactions.items.len == 0;
+        return self.transactions.len == 0;
+    }
+
+    pub fn init(poh_entry: sig.core.poh.PohEntry, transactions: []const Transaction) Entry {
+        return .{
+            .num_hashes = poh_entry.num_hashes,
+            .hash = poh_entry.hash,
+            .transactions = transactions,
+        };
```
InKryption: I don't see what the point of this initialization method is, it mostly seems to obfuscate the relationship between `transactions` field and the `deinit` method.

src/core/entry.zig
```diff
@@ -185,10 +193,7 @@ pub const test_entry = struct {
         .num_hashes = 149218308,
         .hash = sig.core.Hash
             .parseBase58String("G8T3smgLc4XavAtxScD3u4FTAqPtwbFCEJKwJbfoECcd") catch unreachable,
-        .transactions = .{
-            .items = txns[0..2],
-            .capacity = 2,
-        },
+        .transactions = txns[0..2],
```
InKryption: Can we now inline `txns`? ie
```suggestion
        .transactions = &.{
            sig.core.transaction.transaction_v0_example.as_struct,
            sig.core.transaction.transaction_v0_example.as_struct,
        },
```

src/core/genesis_config.zig
```diff
@@ -57,6 +61,70 @@ pub const FeeRateGovernor = struct {
 
     pub const @"!bincode-config:lamports_per_signature" = bincode.FieldConfig(u64){ .skip = true };
 
+    pub const DEFAULT = FeeRateGovernor{
```
InKryption, line 64:
```suggestion
    pub const DEFAULT: FeeRateGovernor = .{
```

src/core/hash.zig
```diff
@@ -129,5 +129,7 @@ pub const Hash = extern struct {
 pub const LtHash = struct {
     data: [NUM_ELEMENTS]u16,
 
+    pub const ZEROES = LtHash{ .data = .{0} ** NUM_ELEMENTS };
```
InKryption, line 132:
```suggestion
    pub const ZEROES: LtHash = .{ .data = .{0} ** NUM_ELEMENTS };
```

src/core/leader_schedule.zig
InKryption: Suggestion: make the interface use an untagged of const and non-const pointer, allowing the implementation to choose which it is using. Effectively allows constness to be selected at runtime.

src/core/poh.zig
```diff
@@ -0,0 +1,203 @@
...
+    /// Calling this function when hashes_per_tick < 2 has undefined behavior.
+    pub fn recordAndMaybeTick(self: *Poh, mixin: Hash) struct {
+        maybe_tick: ?PohEntry,
+        record: PohEntry,
+    } {
+        assert(self.hashes_per_tick > 1);
+        var maybe_tick: ?PohEntry = null;
+        var record: PohEntry = undefined;
+        if (self.remaining_hashes == 1) maybe_tick = self.tick() orelse unreachable;
+        record = self.tryRecord(mixin) orelse unreachable;
```
InKryption:
```suggestion
        if (self.remaining_hashes == 1) maybe_tick = self.tick().?;
        const record: PohEntry = self.tryRecord(mixin).?;
```

src/core/poh.zig
```diff
@@ -0,0 +1,203 @@
...
+        if (self.remaining_hashes == 1) {
+            return null; // needs a tick first
+        }
+
+        self.latest_hash = self.latest_hash.extendAndHash(mixin.data);
+        const num_hashes = self.num_hashes + 1;
+        self.num_hashes = 0;
+        self.remaining_hashes -= 1;
+
+        return PohEntry{
```
InKryption, line 78:
```suggestion
        return .{
```

src/core/poh.zig
```diff
@@ -0,0 +1,203 @@
...
+        if (self.remaining_hashes != 0) {
+            return null;
+        }
+
+        const num_hashes = self.num_hashes;
+        self.remaining_hashes = self.hashes_per_tick;
+        self.num_hashes = 0;
+        self.tick_count += 1;
+
+        return PohEntry{
```
InKryption, line 101:
```suggestion
        return .{
```

src/core/transaction.zig
```diff
@@ -75,7 +75,50 @@ pub const Transaction = struct {
...
     }
 
-    pub const InitOwnedMsgWithSigningKeypairsError = error{
+    /// Basic transaction with randomized addresses and data.
+    ///
+    /// The number of instructions, pubkeys, and program/writable/signer/readonly
+    /// indexes are hardcoded, not randomized.
+    pub fn initRandom(allocator: std.mem.Allocator, random: std.Random) !Transaction {
+        const KeyPair = std.crypto.sign.Ed25519.KeyPair;
+        const keypair = try KeyPair.generateDeterministic(.{random.int(u8)} ** 32);
```
InKryption, line 84: Just to clarify: the intention is to generate a 32 byte array of one single homogeneous, randomly selected value?

src/core/transaction.zig
```diff
@@ -75,7 +75,50 @@ pub const Transaction = struct {
...
+        errdefer allocator.free(account_indexes);
+
+        const instructions = try allocator.dupe(Instruction, &.{.{
+            .program_index = 2,
+            .account_indexes = account_indexes,
+            .data = data,
+        }});
+        errdefer allocator.free(instructions);
+
+        const message = Message{
```
InKryption, line 108:
```suggestion
        const message: Message = .{
```

src/ledger/reader.zig
```diff
@@ -1209,22 +1209,23 @@ pub const BlockstoreReader = struct {
...
     /// Analogous to [get_slots_since](https://github.com/anza-xyz/agave/blob/15dbe7fb0fc07e11aaad89de1576016412c7eb9e/ledger/src/blockstore.rs#L3821)
-    pub fn getSlotsSince(self: *Self, slots: []const Slot) !AutoHashMap(Slot, ArrayList(Slot)) {
+    pub fn getSlotsSince(
+        db: *BlockstoreDB,
+        allocator: Allocator,
+        slots: []const Slot,
+    ) !std.AutoArrayHashMapUnmanaged(Slot, ArrayList(Slot)) {
         // TODO perf: support multi_get in db
-        var map = AutoHashMap(Slot, ArrayList(Slot)).init(self.allocator);
+        var map = std.AutoArrayHashMapUnmanaged(Slot, ArrayList(Slot)).empty;
```
InKryption:
```suggestion
        var map: std.AutoArrayHashMapUnmanaged(Slot, ArrayList(Slot)) = .empty;
```

src/ledger/reader.zig
```diff
@@ -1209,22 +1209,23 @@ pub const BlockstoreReader = struct {
     /// element's children slots.
     ///
     /// Analogous to [get_slots_since](https://github.com/anza-xyz/agave/blob/15dbe7fb0fc07e11aaad89de1576016412c7eb9e/ledger/src/blockstore.rs#L3821)
-    pub fn getSlotsSince(self: *Self, slots: []const Slot) !AutoHashMap(Slot, ArrayList(Slot)) {
+    pub fn getSlotsSince(
+        db: *BlockstoreDB,
+        allocator: Allocator,
+        slots: []const Slot,
+    ) !std.AutoArrayHashMapUnmanaged(Slot, ArrayList(Slot)) {
```
InKryption: Can we use an unmanaged arraylist here?

src/replay/confirm_slot.zig
```diff
@@ -0,0 +1,548 @@
...
+        errdefer scheduler.deinit();
+
+        const poh_verifier = try HomogeneousThreadPool(PohTask)
+            .initBorrowed(allocator, thread_pool, thread_pool.max_threads);
+        errdefer poh_verifier.deinit(allocator);
+
+        const future = try allocator.create(ConfirmSlotFuture);
+        errdefer allocator.free(future);
+
+        future.* = ConfirmSlotFuture{
```
InKryption, line 139:
```suggestion
        future.* = .{
```

src/replay/confirm_slot.zig
```diff
@@ -0,0 +1,548 @@
...
+        logger.warn().logf("Tick with invalid number of hashes found in slot: {}", .{params.slot});
+        return .InvalidTickHashCount;
+    }
+
+    return null;
+}
+
+/// Analogous to [BlockstoreProcessorError](https://github.com/anza-xyz/agave/blob/161fc1965bdb4190aa2d7e36c7c745b4661b10ed/ledger/src/blockstore_processor.rs#L779)
+pub const ConfirmSlotError = union(enum) {
+    failed_to_load_entries: anyerror,
```
InKryption: Can we avoid anyerror here? If this is some generic error that encompasses too much to describe concretely, I don't think it's worth propagating.

src/replay/execution.zig
```diff
@@ -0,0 +1,198 @@
...
+/// 2. Store the replay results into the relevant data structures.
+///
+/// Analogous to [replay_active_banks](https://github.com/anza-xyz/agave/blob/3f68568060fd06f2d561ad79e8d8eb5c5136815a/core/src/replay_stage.rs#L3356)
+pub fn replayActiveSlots(state: *ReplayExecutionState) !bool {
+    const active_slots = try state.slot_tracker.activeSlots(state.allocator);
+    if (active_slots.len == 0) {
+        return false;
+    }
+
+    var slot_statuses = std.ArrayListUnmanaged(struct { Slot, ReplaySlotStatus }){};
```
```suggestion
    var slot_statuses: std.ArrayListUnmanaged(struct { Slot, ReplaySlotStatus }) = .empty;
```

src/replay/execution.zig
```diff
@@ -0,0 +1,198 @@
...
+    defer {
+        for (slot_statuses.items) |status| status[1].confirm.destroy(state.allocator);
+        slot_statuses.deinit(state.allocator);
+    }
+    for (active_slots) |slot| {
+        const result = try replaySlot(state, slot);
+        try slot_statuses.append(state.allocator, .{ slot, result });
```
InKryption: This seems like a leak, shouldn't there be a `errdefer result.destroy();` here?

src/replay/scheduler.zig
```diff
@@ -0,0 +1,294 @@
...
+        allocator: Allocator,
+        batch_capacity: usize,
+        thread_pool: *ThreadPool,
+    ) !TransactionScheduler {
+        var batches = try std.ArrayListUnmanaged(ResolvedBatch)
+            .initCapacity(allocator, batch_capacity);
+        errdefer batches.deinit(allocator);
+
+        const pool = try HomogeneousThreadPool(ProcessBatchTask)
+            .initBorrowed(allocator, thread_pool, null);
```
InKryption, line 79:
```suggestion
        const pool: HomogeneousThreadPool(ProcessBatchTask) =
            try .initBorrowed(allocator, thread_pool, null);
```

src/replay/scheduler.zig
```diff
@@ -0,0 +1,294 @@
...
+        errdefer pool.deinit(allocator);
+
+        var channel = try Channel(?TransactionError).init(allocator);
```
InKryption:
```suggestion
        var channel: Channel(?TransactionError) = try .init(allocator);
```

src/replay/scheduler.zig
```diff
@@ -0,0 +1,294 @@
...
+                switch (e) {
+                    error.LockFailed => if (self.batches_started == self.batches_finished) {
+                        return .AccountInUse;
+                    } else {
+                        break;
+                    },
+                    else => return e,
+                }
+            };
+            assert(try self.thread_pool.trySchedule(self.allocator, .{
```
InKryption, line 166: Can we get a comment on why this assertion should hold true? It's not obvious to me.

src/replay/scheduler.zig
```diff
@@ -0,0 +1,294 @@
...
+            .pending => return .pending,
+            .err => |err| return err,
+        }
+    }
+
+    fn tryScheduleSome(self: *TransactionScheduler) !?TransactionError {
+        while (self.batches.items.len > self.batches_started) {
+            const batch = self.batches.items[self.batches_started];
+            self.locks.lockStrict(self.allocator, batch.accounts) catch |e| {
+                switch (e) {
```
InKryption: Why the dedicated block for this? Seems like it would be cleaner and easier to read as `catch |err| switch (err) {`, which is also a pattern specifically recognized by zig, to my understanding (ie taken advantage of in error traces and optimization).

src/replay/scheduler.zig
```diff
@@ -0,0 +1,294 @@
...
+
+    const transactions = [_]Transaction{
+        try Transaction.initRandom(allocator, rng.random()),
+        try Transaction.initRandom(allocator, rng.random()),
+        try Transaction.initRandom(allocator, rng.random()),
+        try Transaction.initRandom(allocator, rng.random()),
+        try Transaction.initRandom(allocator, rng.random()),
+        try Transaction.initRandom(allocator, rng.random()),
+    };
+    defer for (transactions) |tx| tx.deinit(allocator);
```
InKryption: This would leak a bunch if any of the transactions besides the first fail. I'd suggest using an arena instead, if the objective is terseness.
```suggestion
    var tx_arena_state: std.heap.ArenaAllocator = .init(allocator);
    defer tx_arena_state.deinit();
    const tx_arena = tx_arena_state.allocator();
    const transactions = [_]Transaction{
        try .initRandom(tx_arena, rng.random()),
        try .initRandom(tx_arena, rng.random()),
        try .initRandom(tx_arena, rng.random()),
        try .initRandom(tx_arena, rng.random()),
        try .initRandom(tx_arena, rng.random()),
        try .initRandom(tx_arena, rng.random()),
    };
```

src/replay/scheduler.zig
```diff
@@ -0,0 +1,294 @@
...
+
+    var transactions = [_]Transaction{
+        try Transaction.initRandom(allocator, rng.random()),
+        try Transaction.initRandom(allocator, rng.random()),
+        try Transaction.initRandom(allocator, rng.random()),
+        try Transaction.initRandom(allocator, rng.random()),
+        try Transaction.initRandom(allocator, rng.random()),
+        try Transaction.initRandom(allocator, rng.random()),
+    };
+    defer for (transactions) |tx| tx.deinit(allocator);
```
InKryption: Same suggestion:
```suggestion
    var tx_arena_state: std.heap.ArenaAllocator = .init(allocator);
    defer tx_arena_state.deinit();
    const tx_arena = tx_arena_state.allocator();
    const transactions = [_]Transaction{
        try .initRandom(tx_arena, rng.random()),
        try .initRandom(tx_arena, rng.random()),
        try .initRandom(tx_arena, rng.random()),
        try .initRandom(tx_arena, rng.random()),
        try .initRandom(tx_arena, rng.random()),
        try .initRandom(tx_arena, rng.random()),
    };
```

src/replay/scheduler.zig
```diff
@@ -0,0 +1,294 @@
...
+
+    const replaced_sigs = try allocator.dupe(sig.core.Signature, transactions[5].signatures);
+    replaced_sigs[0].data[0] +%= 1;
+    allocator.free(transactions[5].signatures);
+    transactions[5].signatures = replaced_sigs;
```
InKryption, line 288: Isn't this leaking the original `transactions[5].signatures`?
Well, with the arena suggestion, this wouldn't matter so much.

src/replay/scheduler.zig
```diff
@@ -0,0 +1,294 @@
...
+    };
+    defer for (transactions) |tx| tx.deinit(allocator);
+
+    {
+        const batch1 = try replay.resolve_lookup
+            .resolveBatch(allocator, undefined, transactions[0..3]);
+        errdefer batch1.deinit(allocator);
+
+        const batch2 = try replay.resolve_lookup
+            .resolveBatch(allocator, undefined, transactions[3..6]);
```
InKryption: Why is passing undefined here okay?

src/replay/trackers.zig
```diff
@@ -17,27 +17,90 @@ const SlotState = sig.core.SlotState;
 /// kitchen-sink style approach of storing everything under the sun.
 ///
 /// [BankForks](https://github.com/anza-xyz/agave/blob/161fc1965bdb4190aa2d7e36c7c745b4661b10ed/runtime/src/bank_forks.rs#L75)
+///
+/// This struct is *not* thread safe, and the lifetimes of the returned pointers
+/// will end as soon as the items are removed.
 pub const SlotTracker = struct {
-    slots: std.AutoArrayHashMapUnmanaged(Slot, Element) = .{},
+    slots: std.AutoArrayHashMapUnmanaged(Slot, *Element),
+    root: std.atomic.Value(Slot),
```
InKryption: Rather not make this atomic; if SlotTracker is not already properly internally locking, and not exposing a well defined API to access it safely accross threads, I don't think it's a sound decision to create a false sense of security or lattice of "some parts are thread safe, others are not".

src/replay/trackers.zig
```diff
@@ -17,27 +17,90 @@ const SlotState = sig.core.SlotState;
...
+    pub fn put(
+        self: *SlotTracker,
+        allocator: Allocator,
+        slot: Slot,
+        constants: SlotConstants,
+        state: SlotState,
+    ) !void {
+        const elem = try allocator.create(Element);
+        elem.* = .{ .constants = constants, .state = state };
+        try self.slots.put(allocator, slot, elem);
```
InKryption:
```suggestion
        try self.slots.ensureUnusedCapacity(allocator, 1);
        const elem = try allocator.create(Element);
        elem.* = .{ .constants = constants, .state = state };
        self.slots.putAssumeCapacity(slot, elem);
```
Otherwise `elem` may be leaked.

src/replay/trackers.zig
```diff
@@ -17,27 +17,90 @@ const SlotState = sig.core.SlotState;
...
             }
         }
         return try list.toOwnedSlice(allocator);
     }
+
+    pub fn frozenSlots(
+        self: *const SlotTracker,
+        allocator: Allocator,
+    ) Allocator.Error!std.AutoArrayHashMapUnmanaged(Slot, Reference) {
+        var frozen_slots = std.AutoArrayHashMapUnmanaged(Slot, Reference).empty;
```
InKryption, line 92:
```suggestion
        var frozen_slots: std.AutoArrayHashMapUnmanaged(Slot, Reference) = .empty;
        errdefer frozen_slots.deinit(allocator);
```

src/replay/trackers.zig
```diff
@@ -17,27 +17,90 @@ const SlotState = sig.core.SlotState;
...
+        var iter = self.slots.iterator();
+        while (iter.next()) |entry| {
```
InKryption: I would suggest doing `for (self.slots.keys(), self.slots.values()) |key, value| {`, as well as `frozen_slots.ensureTotalCapacity(allocator, self.slots.count())` in order to turn the `try put` into a `putAssumeCapacity`.

InKryption (changes requested):
The ovearching structure is fine, but I have some questions, comments, and suggestions on some of the fine details.

src/core/bank.zig
```diff
@@ -422,22 +515,33 @@ pub const BlockhashQueue = struct {
     /// hashes older than `max_age` will be dropped from the queue
     max_age: usize,
 
-    pub fn deinit(bhq: BlockhashQueue, allocator: std.mem.Allocator) void {
-        var ages = bhq.ages;
+    pub const DEFAULT = BlockhashQueue.init(300);
```
dadepo: Maybe have this in a constant `MAX_RECENT_BLOCKHASHES` similar to Agave?

src/replay/confirm_slot.zig
InKryption: I would suggest using `std.log.err` instead, which will integrate with the test runner, and avoid it getting overwritten by test runner output.

src/core/leader_schedule.zig
Rexicon226: > I think constCast is justified here. 

using `@constCast` is basically never justified. The only places I could think of in this context would be *inside* of the interface implementation itself, similar to how the stdlib http server uses `@constCast`, even though that's being moved away from in the lastest release cycle with the refined writer interface.

> By comptime interface, you mean passing down an additional parameter to indicate the type? or using anytype?

None of these. This can be easily described as a union enum which will contain entries for the `EpochContextManager`, `LeaderScheduleCache`, and a test entry "`OneSlotLeader`" which will only be compiled in for tests.

Take a look at this commit here: 60ef08f17eb623e09bad0103be27716c21363689 where I show my approach. It would trivial to extend to the usecase in this PR.

I've often seen the argument of "tradeoffs" used, but here I think there is no excuse. This is incredibly confusing to someone who isn't very familiar with Zig, and better options exist.

src/replay/confirm_slot.zig
Rexicon226: I understand, but these tests don't fail (I hope). This is really a nit, I'm totally fine leaving these, but I personally prefer not to leave debug prints in the code.

src/replay/confirm_slot.zig
Rexicon226: It makes perfect sense! `@TypeOf` guarantees that evaluation of the body has no side-effects and we can take full advantage of that. If you take a step back and read the code, it becomes so much better. "this function returns the type returned by this function call".

src/replay/confirm_slot.zig
Rexicon226: We could add them, but we haven't yet. I always prefer the more conservative approach, YAGNI.

src/replay/resolve_lookup.zig
Rexicon226: I don't think that "it's another lifetime" is really a good argument. Yes you need to manage it, that's the entire point. We're trying to be careful with the ownership of the `Transaction`, and a pointer is a great way to describe that. Making said reference stable only provides benefits as far as I can see.

src/replay/scheduler.zig
Rexicon226: Having each argument on a separate line pretty clearly increases readability in my opinion. Because... they are on separate lines. There could be other function calls inline with the arguments, there could be more arguments, they could have long or short names (both have their downsides in this context).

With this approach we even the playing field. Each line is clearly a new argument and they don't become mixed when reading through it.

If someone isn't familiar enough with Zig to understand the version I suggested, they are going to have a hell of a time reading through the rest of the codebase, and the standard library. This is a very common pattern, used for the aforementioned reasons. Are there any other arguments against taking this approach?

src/replay/scheduler.zig
Rexicon226: That you're not creating what looks to be a tuple literal and then "magically" coercing it into a slice. Instead it's clear what the type we're creating and how it'll be used.

src/replay/confirm_slot.zig
Rexicon226: I really prefer we don't put allocators everywhere. I have a large branch WIP that removes a ton of managed stuff, since we are storing kilobytes of duplicate allocator interfaces. Not adding more just helps the cause :).

src/replay/resolve_lookup.zig
dnut: > a pointer is a great way to describe that

Your argument is that the pointer acts as a form of documentation to communicates that the memory is borrowed? I see what you mean, but I don't think it's a great way to communicate that idea. It can be a pointer and still be owned by this struct. Explicit documentation is clearer and more conclusive.

src/replay/confirm_slot.zig
dnut: I agree it has a cost but sometimes the cost is justified to ensure the correct allocator is used. In this case it wasn't necessary so I removed it. When I first wrote this struct, I wasn't sure how it would be used, and if it would be straightforward to ensure the correct allocator is in scope.

src/replay/scheduler.zig
dnut: I think this is a matter of personal preference and is pretty low consequence. I don't think we need to enforce any rules about this unless the team overall has a strong preference about how this should be written.

---------------

I don't think it's a question of whether someone understands zig enough to make sense of the code. I think anyone who understands one variant will be able to understand the other. My concern is how quickly and easily a reader can extract the information they are looking for.

I find it easier to understand code when the vertical direction indicates steps and the horizontal direction specifies inputs and outputs of those steps. In general, this is how most code is written, unless the function call would exceed 100 characters. So my approach is actually closer to the common pattern. When the vertical direction is cluttered with function arguments, I just find it more tedious to scan through the code. It's less organized and uses space less efficiently. I need to use more mental bandwidth to understand the same thing.

src/replay/resolve_lookup.zig
Rexicon226: > It can be a pointer and still be owned by this struct. 

It could be, but it isn't. It is the best way to communicate the idea when it comes to the type system.

Why not have both? We document that the pointer comes from an external source and we see that by reading the (de)init code.

src/replay/confirm_slot.zig
Rexicon226: The cases where you have the multiple allocators and the possibility of using the wrong one are very rare. In this case `ConformSlotFuture` is always managed by something that stores its own allocator, so there is practically speaking no possibility of this.

We currently have this double allocator abstraction in the gossip code, but I think that will go also, when I get the time to do some benchmarks on the real affects to performance it has to just share the same allocator. I understand the argument before was that the contention on a single allocator was too high, but now with the new smp allocator provided in 0.14 that shouldn't be an issue anymore.

Anyways, happy to resolve this!

src/replay/trackers.zig
```diff
@@ -17,27 +17,90 @@ const SlotState = sig.core.SlotState;
...
+        var slots = self.slots;
+        for (slots.values()) |v| allocator.destroy(v);
+        slots.deinit(allocator);
+    }
+
+    pub fn put(
+        self: *SlotTracker,
+        allocator: Allocator,
+        slot: Slot,
+        constants: SlotConstants,
```
dadepo, line 54: `SlotConstants` also has a `slot` field. Does that duplicate the `slot` parameter passed in the function? if so, perhaps the slot can be retrieved from `SlotConstants.slot`

src/replay/service.zig
```diff
@@ -91,6 +134,79 @@ fn advanceReplay(state: *ReplayState) !void {
...
+/// Analogous to
+/// [generate_new_bank_forks](https://github.com/anza-xyz/agave/blob/146ebd8be3857d530c0946003fcd58be220c3290/core/src/replay_stage.rs#L4149)
+fn trackNewSlots(
+    allocator: Allocator,
+    blockstore_db: *sig.ledger.BlockstoreDB,
+    slot_tracker: *SlotTracker,
+    epoch_tracker: *EpochTracker,
+    slot_leaders: SlotLeaders,
+    /// needed for update_fork_propagated_threshold_from_votes
+    _: *ProgressMap,
```
dadepo, line 149: Is it clear in which work the `update_fork_propagated_threshold_from_votes` is going to be implemented? It is not listed in the edge case bookkeeping, neither in the consensus track.

src/replay/service.zig
```diff
@@ -91,6 +134,79 @@ fn advanceReplay(state: *ReplayState) !void {
...
+) !void {
+    const root = slot_tracker.root.load(.monotonic);
+    var frozen_slots = try slot_tracker.frozenSlots(allocator);
+    defer frozen_slots.deinit(allocator);
+
+    var frozen_slots_since_root = try std.ArrayListUnmanaged(sig.core.Slot)
+        .initCapacity(allocator, frozen_slots.count());
+    defer frozen_slots_since_root.deinit(allocator);
+    for (frozen_slots.keys()) |slot| if (slot >= root) {
+        try frozen_slots_since_root.append(allocator, slot);
```
dadepo: Would it be better to use `appendAssumeCapacity` here given that the size would be less or greater to the init capacity of `frozen_slots.count()`?

src/replay/service.zig
```diff
@@ -91,6 +134,79 @@ fn advanceReplay(state: *ReplayState) !void {
...
+    };
+
+    var next_slots = try BlockstoreReader
+        .getSlotsSince(blockstore_db, allocator, frozen_slots_since_root.items);
```
dadepo: nit: I think the general convention is to have `allocator` as the first param.

src/ledger/reader.zig
dadepo: stylistic question. I have also seen this variant

```
var map: std.AutoArrayHashMapUnmanaged(Slot, ArrayList(Slot)) = .{};
```

Any reason for one over the other? the `... = .{}` looks a bit more compact though.

src/replay/scheduler.zig
Rexicon226: Agreed that this is of low consequence, but just for the interest of conversation,

I don't think anybody actually reads line by line. Unless you have one of those paper cutouts from elementary school that blocks off the rest words on the page, your brain will naturally start to search for patterns. At least for me personally, my brain immediately jumps to the opening and closing parenthesis or brackets, depending on the context. After I see the function call taking place, I naturally look at the code in-between the two parenthesis (this would be several lines of arguments in our case), and I come back to the point where having an argument on each line is clearer. This is a recurring pattern in Zig, with the same idea applying to function arguments, container initialization, switch cases with multiple matches, and many more. We don't place longer struct initialization all on the same line because it's an "input" to the assignment, we split it up into many lines to make it clear what values each field is being initialized to.

If we cared about using space efficiently, we wouldn't be using indents, or spaces between lines of code. With organization, I honestly do not see how the multi-line approach is less organized. You have an opening line which contains the information about the function we're calling (is it a method, what's the name, etc.), three lines with one for each argument, clearly detailing what they are and how many of them (it's much easier to count lines instead of counting arguments between commas), and the last line which acts as a deliminator for the whole function call.

src/ledger/reader.zig
InKryption: That exists temporarily for compatibility with older code that used it. The default field values will be removed in the next release cycle; the `.empty` declaration is the correct initialization going forward.

src/core/bank.zig
```diff
@@ -77,8 +81,33 @@ pub const SlotConstants = struct {
     /// Whether and how epoch rewards should be distributed in this slot.
     epoch_reward_status: EpochRewardStatus,
 
+    pub fn fromBankFields(bank_fields: *const BankFields) SlotConstants {
+        return .{
+            .slot = bank_fields.slot,
+            .parent_slot = bank_fields.parent_slot,
```
dadepo, line 83: The fields of `SlotConstants` were not added in this PR but should `.parent_slot` not be nullable? for cases where the slot has no reacheable parent?

src/core/bank.zig
dnut: I've thought about this and read some arguments about it and concluded that it doesn't really matter. But personally I prefer `= SlotState{` because I think it's just a tiny bit more intuitive and readable.

InKryption: I think the fact that andrewrk himself is strongly leaning towards removing `T{}` syntax altogether (https://github.com/ziglang/zig/issues/5038), with other supporting proposals facilitating the unification towards coercion in place of typed literals (https://github.com/ziglang/zig/issues/19674) is a pretty strong argument against using `T{}` syntax, plus the fact that with the advent of decl literals, placing explicit type annotations rather than inferring types from the expression will continue to become the dominant syntax form, and thus using coercion increasingly becomes the more consistent and uniform choice.

At any rate, I won't block the change set on this.

src/sync/mux.zig
dnut: e27f76bbf92ead57d74b51c268719910487ca387

I don't see the problem with this function, but I guess since there's only one usage of it, I'm fine removing and inlining it, if you think that improves the code quality. If I find a lot of other places where I need this behavior, I may want to add it back.

------------------------------

I think it's a nice little helper function. I don't understand why the optional contract is a problem. It's just a function you can use if it suits your needs. There's no requirement to conform to anything, unless you want to use the function. It's not adding constraints on any existing code.

If this function does exist, I think putting it inside the struct makes it more discoverable. I don't see the benefit of moving it to another location.

src/accountsdb/buffer_pool.zig
dnut: My understanding of this code is that it copies `@min(self.len() - start, buf.len)` bytes. I found it a little confusing at the call site to verify how much data was copied. Directly returning the value means the call site doesn't need to rely on as many assumptions about the function's behavior. It's a little easier to use and less likely to get misinterpreted.

src/utils/thread.zig
dnut: That's a good point about `anyerror`. I used it out of laziness. Actually the error type is known and can be easily returned.

Regarding how to provide that error to the caller, I would prefer not to add extra layers of indirection. All three statuses are equally important and need to be handled at the same time. Handling it with a single switch is simple and direct.

135c07b21a6fbba1fb39c5fd3cb2515a1663590a

src/core/transaction.zig
dnut: Yes that was the intention. This array is going to be used as a seed for the prng. The private key itself will become a different, non-homogeneous array.

This means our sample set of keypairs has a size of 256, which is quite small. Should be fine for tests, which is the only place where it would make sense to call this function. I'm not opposed to changing this to increase the size of the sample set. Should just be two more lines of code.

src/replay/confirm_slot.zig
dnut: I don't see the harm of including a bit more context that can easily be ignored or consolidated to a different error type in the context that's making use of it.

src/replay/scheduler.zig
dnut: f695746f6f2ab194f5bb14ec4a1af8b38bd4dbef

I usually try to avoid these superfluous blocks. Not sure what happened here. You're saying this block can actually cause harm for the compiler's ability to understand the code?

src/replay/scheduler.zig
dnut: I was aware of this when I wrote the code, but this is a test so it's not really of any consequence. I was just trying to keep the code simple and easy to read.

src/accountsdb/buffer_pool.zig
InKryption: After taking a second look, I see what you're saying; yeah, that's actually better. May be worth also adding a comment on this function briefly describing the return value's meaning, and its relation to `self` and `start`.

Also, it would be good to add an assertion to this function that `start <= self.len()`, since I think that would have helped in avoiding my misunderstanding.

src/replay/scheduler.zig
dnut: The line above is `allocator.free(transactions[5].signatures);`

src/replay/scheduler.zig
dnut: We don't need accountsdb since there are no lookup tables.

src/core/entry.zig
dnut: e366fe00e65ac2ab4293ad98731b0bcc6ad67404

src/sync/mux.zig
InKryption: It obfuscates the real operations of the code, which is acquiring a lock, calling a method called "clone" on the protected value, unlocking it, and then returning the result of `clone`. This assumes a lot about the protected value, in particular that it has a clone function, and that the clone function returns a fully deep copy that then doesn't need to be protected by the lock afterwards.

In the simple case, that's probably true, but I am not comfortable with ergonomic optimizations in the simple case leaving the door open to potential footguns in the less complicated cases.

Worse, this kind of design is often a slippery slope, with a tendency to build on them instead of peeling them back when necessary, crafting further complicated abstractions that obfuscate the semantics even more.

I don't think it's a fundamentally bad helper function, but I don't think it's one that deserves to live inside any of the synchronization primitive wrappers, because it should not their responsibility to understand any of the implementation details of the protected value, especially nothing about how it's allocated or cloned.

src/core/transaction.zig
InKryption: I think that's fine, just wanted to clarify, thank you for elaborating.

src/ledger/reader.zig
dnut: c602473b172ab5622cc9af5b3e4d9cadd288d36f

src/replay/scheduler.zig
InKryption: In the sense that `catch |err| switch (err)` is understood better by the compiler, and `catch |err| { switch (err)` is understood in a more generic sense, yeah.

The former becomes a single intermediate instruction, which the compiler can emit more specialized codegen for, whereas the latter results in a generic control flow split, where one branch is the success path, and the other is a block which happens to contain a switch.

It's in the same vein as `foo[a..][0..b]` compiling down as a single slicing operation, resulting in just a single bounds check.

src/replay/scheduler.zig
InKryption: I think it is of consequence for there to be potential leaks, which in a failure state would lead to confusion seeing both the main error and the leaks. I don't think we should be lazy about memory allocation just because it's a unit test.

src/replay/scheduler.zig
InKryption: You're right, my bad, don't know how I missed that -

src/utils/thread.zig
InKryption: That's fine, if there's a concrete error set, let's use that.

src/replay/confirm_slot.zig
dnut: Yeah it does make perfect sense. I like how the meaning is immediately intuitive. It's just weird to see the syntax for a function call, but the function is not actually called.

src/replay/confirm_slot.zig
InKryption: Is there anywhere where that's actually being taken advantage of? If not, I would personally prefer not to leave this open to the temptation of `.failed_to_load_entries => |err| return err`.
If it's just for optional error reporting, how would you feel about:
```zig
    /// Payload is a statically lived string that provides some context on the failure to load entries, normally the name of an error.
    failed_to_load_entries: []const u8,
```

src/replay/scheduler.zig
InKryption: Would be nice to encode this mode of operation with a dedicated noop provider, so as to trigger `error.PubkeyNotInIndex`, rather than illegal behavior.

src/replay/confirm_slot.zig
dnut: 1873e4cfd04d3751325d68f3073f54fe61b3b498

tests should output a helpful message when they fail. if it's safe to assume a test will never fail, you might as well delete the entire test.

src/replay/scheduler.zig
dnut: 0d27f8b7bc7baa41f9bdc67b643eef299e5dddcb

src/replay/confirm_slot.zig
dnut: I think it still serves a purpose of clarifying which item in this file is being tested, regardless of whether there are tests for the other items in this file. I don't think it really matters though tbh, so I'm happy to remove it if you prefer.

src/ledger/reader.zig
dnut: I prefer the current style and don't plan to change this due to the conclusion of https://github.com/Syndica/sig/pull/722#discussion_r2142703928

src/replay/execution.zig
dnut: 8e84374c45c0a57efb99e66bb3331ea4508c7d4e

src/replay/scheduler.zig
dnut: I prefer the current style and don't plan to change this due to the conclusion of https://github.com/Syndica/sig/pull/722#discussion_r2142703928

src/replay/scheduler.zig
dnut: I prefer the current style and don't plan to change this due to the conclusion of https://github.com/Syndica/sig/pull/722#discussion_r2142703928

src/core/poh.zig
dnut: e66276b363340afbe9e544e82e8756875125c457

src/core/entry.zig
dnut: I didn't notice this inconsistency. Something definitely needs to change here.

src/replay/confirm_slot.zig
dnut: bc707a92699367ff268052bc34c18ab47d67222e

src/accountsdb/buffer_pool.zig
dnut: e8f753ab687dd61dc582840229c6424eae7bd03d

src/replay/execution.zig
dnut: Maybe I changed this after your comment. The current code looks like this:
```zig
    var slot_statuses = std.ArrayListUnmanaged(struct { Slot, ReplaySlotStatus }).empty;
    defer {
        for (slot_statuses.items) |status| status[1].confirm.destroy(state.allocator);
        slot_statuses.deinit(state.allocator);
    }
    for (active_slots) |slot| {
        const result = try replaySlot(state, slot);
        try slot_statuses.append(state.allocator, .{ slot, result });
    }
```

dnut: c4474c95b830f0ea1bbfe0904a0d7151c8409f51

src/replay/scheduler.zig
dnut: 86283f5aff1b1433a689e3b822f45d2fbea3f5ec

src/replay/scheduler.zig
dnut: 06545f22ae7b3bd58f7c525f51f4de8759b552df

src/replay/scheduler.zig
dnut: 06545f22ae7b3bd58f7c525f51f4de8759b552df

src/replay/scheduler.zig
dnut: 90d26d4c6967b51e99c28bfe79470f56c58e9a5b

src/replay/trackers.zig
dnut: 7422ef3dd2850a9b4465a8705381fa20221fce20

src/replay/trackers.zig
dnut: 40aaa240b6ee9edf41bdc4347110194400bae5f8

src/replay/trackers.zig
dnut: my only concern with ensureTotalCapacity is that it may end up allocating way more memory than needed if only a small proportion of the slots are frozen.

my instinct is that the cost of repeated syscalls is worse than the cost of using more memory here, so i took your suggestion.

7293f576c29051e41e34365784397a34277130bd

src/replay/trackers.zig
dnut: That's a good point but actually I've been kind of thinking about using the opposite mitigation. Including `slot` in SlotConstants feels redundant since you should know what slot you're dealing with already.

src/core/bank.zig
dnut: The parent slot should be known for any slot that we have these constants for. I'm not sure what you mean by reachable.

src/replay/service.zig
dnut: bbf8ca4a06e82f179c1fe5ddd67fb738ba713392

src/replay/service.zig
dnut: b45d200002aedf265b421cfdbb9089b0e91b4a9c

src/replay/service.zig
dnut: `update_fork_propagated_threshold_from_votes` is needed for `compute_bank_stats` which is part of processing consensus in replay. We need all the consensus operations in replay to be implemented now, unless we explicitly review it and determine it not to be necessary. So we need compute_bank_stats implemented, and every function it calls.

Edge case bookkeeping is just the subset of functions in replay that call `check_slot_agrees_with_cluster`.

src/core/bank.zig
dnut: I wasn't intending to flesh out all the constants they use for this since I don't know if we actually need it. But I do see an advantage of this one const acting as a sort of documentation.

bb08b5075b441568d3a84f95dbcc7b8633e6f9e5

src/replay/confirm_slot.zig
dnut: This stuff is going to be used in my next task, which is processing the replay results. I'll switch to a string now, which will probably be fine, and I'll sort it out later.

28311c8cd58b191fd67a2ce8dff8e6efa4cbd517

src/core/entry.zig
dnut: having a function like this makes a specific test a lot cleaner. but i see how it's problematic when generally available. so i just defined it within the test so it's only used there.

f727292b719dd2818a280d7c876d7defcdd5f2fb

src/core/leader_schedule.zig
dnut: OK sure, you're talking about a union-based interface. I'm familiar with those. It's just not what I thought you meant when you said "comptime interface." I assumed you were referring to static dispatch. Union-based interfaces are still a form of dynamic dispatch, so I would be inclined to call it a "runtime interface."

Personally I don't think it makes much difference in this case whether we use a vtable or a union. So I'm fine with switching to a union if you have a strong preference.

src/core/entry.zig
InKryption: Good stuff, I do like that solution.

InKryption (approved):
All my feedback has been addressed, will be great to see this merged after remaining feedback from other reviewers has been addressed.

src/core/leader_schedule.zig
dnut: OK sure, you're talking about a union-based interface. I'm familiar with those. I can see using a union-based interface in cases like this where the flexibility of a vtable is not really needed.

When you said "comptime interface," I assumed you were referring to static dispatch. Union-based interfaces are still a form of dynamic dispatch, so I would be inclined to call it a "runtime interface."

To resolve this conversation, I deleted the code in question. There's no more `constCast` and no more conversion method from `LeaderSchedule` to `SlotLeaders`. I only added this for a test, where I can actually use pre-existing code instead. 064580a58b6f8cb3e8f693c25169bacdeb8da689

Switching this interface to a tagged union is not in scope for this pr. For that change, I have one concern about how it tangles up `core` with code outside of `core`, which is something I've been trying to avoid. I'll have to put a little more thought into that aspect of it.

src/replay/scheduler.zig
dnut: 5eb8971f275a72e80514660e88b3d76505715858

Hopefully we can resolve this conversation now that I've eliminated the need for *any* line breaks, so there's no decision to be made about how to do it.

continued bikeshedding below...

-------------------------


> I don't think anybody actually reads line by line. Unless you have one of those paper cutouts from elementary school that blocks off the rest words on the page, your brain will naturally start to search for patterns.

This is also the foundation of my argument. I use pattern recognition first from a zoomed out perspective to see the high level structure, then I zoom in on details and continue to use pattern recognition. I'd like the more prominent patterns - those which are more visible from a zoomed-out view - to represent high-level structure.

The sequence of function calls within a block is a higher-level structure relative to the list of arguments that are passed to any one of those functions. First I'd like to recognize a sequence of function calls. The list of arguments for one of those functions is a finer detail that I'm only interested in after I've honed in on a specific function call. I find it easier to identify a sequence of function calls when it isn't interspersed with visual noise of finely grained details that I don't care about yet.

> If we cared about using space efficiently, we wouldn't be using indents, or spaces between lines of code. 

We do care about using space efficiently, but it's not the only thing we care about, and often there are more important priorities.

src/replay/service.zig
dadepo: In Agave, after replay_active_banks, the following top level functions cover the implementation

| Function              |  Work stream |
| :---------------- |  :----: |
| Self::process_ancestor_hashes_duplicate_slots        |  Edge case book-keeping |
| Self::process_duplicate_confirmed_slots           |  Edge case book-keeping |
| Self::process_gossip_verified_vote_hashes   |  Edge case book-keeping |
| Self::process_popular_pruned_forks |  Edge case book-keeping |
| Self::process_duplicate_slots |  Edge case book-keeping |
| Self::compute_bank_stats |  TODO |
| Self:: check_for_vote_only_mode |  TODO |
| Self::maybe_refresh_last_vote | Consensus |
| Reset fork logic |  Consensus |
| Self::handle_votable_bank | Consensus |
| Self::dump_then_repair_correct_slots |  TODO |
| Self::retransmit_latest_unpropagated_leader_slot |  TODO |
| Self:: maybe_start_leader |  TODO |

And from the above it seems `Self::compute_bank_stats` and `Self:: check_for_vote_only_mode` has not been classified. Also the functions after handling the reset fork `Self::dump_then_repair_correct_slots`, ` Self::retransmit_latest_unpropagated_leader_slot` and `Self:: maybe_start_leader`.

I took the things that fall under edge case booking from the listing here https://github.com/orgs/Syndica/projects/2/views/25?pane=issue&itemId=108886678&issue=Syndica%7Csig%7C777 does that list need to be updated to make it clearer which implementation stays as part of edge case book-keeping? or it is fine as is.

src/core/genesis_config.zig
dnut: I took a stab at making this function more readable. What do you think? 

22b86dbf41912391cbad4608037e0ea2ff4c7d19

Overall, now that I understand how it works, I think it's kind of a dumb calculation. There are much better solutions for this kind of problem used in robotics for example. Anyway, we have to be conformant with agave, so we don't really have a choice.

I added this extensive test because I actually introduced several conformance bugs. The original port was correct, but I made some mistakes while refactoring to the current revision. It was helpful to ensure the calculation is exactly the same as agave.

Rexicon226: Hmm, yeah looks way better. Thank you for making the effort.

src/core/leader_schedule.zig
Rexicon226: > Union-based interfaces are still a form of dynamic dispatch, so I would be inclined to call it a "runtime interface."

It is a bit of a grey area, but I do think it is a comptime interface. While the tag may be runtime known, even though I think currently in all? use-cases we could know the tag at compile time, there is no dispatch happening here. No indirection that the optimizer has to de-virtualize. 

So I think it is static-dispatch in that sense, but a runtime interface in other senses. Either way, I think it's pretty much the best solution, unless we can actually eliminate all interface-like solutions and just take the underlying type at each callsite (which would be great). Happy to discuss this further in a separate PR.

src/replay/confirm_slot.zig
Rexicon226: Cool thanks, I'm not really pressed on making sure the tests are named perfectly, but just wanted to put it out there. Thanks for the change.

src/replay/scheduler.zig
Rexicon226: "Workaround" looks good to me, thank you.

---

Regarding the bikesheding, I think this is a purely subjective discussion. It will always boil down to "i like this and dislike that", and the goal is to find a middle ground. I like the solution to qualifying the function call less, when that was the main cause of the large column count, like in these cases. In other cases, such as a larger number of arguments, I would definitely prefer having them on separate lines considering my points above. Anyways, it's not a hard pressing issue.

src/replay/resolve_lookup.zig
dnut: You're arguing that if a struct contains a pointer to borrowed memory, that struct should not be passed by value - it should be passed as a borrowed pointer itself, as a form of documentation to indicate that it also contains a borrowed pointer. This approach can easily get out of hand as the effect propagates recursively. This is an unusual pattern I haven't seen before, so I don't think it has a clear documentation benefit.

I'm struggling to justify the added layer of indirection that introduces another region of memory that could potentially be freed or go out of scope and cause memory errors when a pointer to it is dereferenced.

dnut: Also btw, this is not part of the current pr. The ResolvedTransaction struct is are already on main as-is.

I made a tweak to this deinit function in the current pr, which is why you were able to comment here, but I have since reverted it back to what it has on main.

yewman: Just to throw my 2 cents in I agree with David here, I think using a pointer + good documentation is preferable. That being said, I don't think we should hold up this PR if we cant come to a resolution since as you said Drew, it is already on main. Perhaps we can discuss further and create a PR to change it if we decide thats the way to go.

Rexicon226: I'm fine discussing this later.

To answer your above comment, I'm not exactly sure what you're referring to here. The fact that we're storing a reference to it has nothing to do with the fact that it contains pointers itself, we don't care what the internals of it look like. What it does tell us is that somebody else owns the actual value (hence it being a constant pointer), and we won't be de-initializing it.

src/sync/mux.zig
```diff
@@ -125,6 +125,14 @@ pub fn Mux(comptime T: type) type {
                 .valid = true,
             };
         }
+
+        /// Acquires the lock just long enough to shallow copy the item, and
+        /// returns the copy.
+        pub fn readCopy(self: *Self) T {
```
yewman, line 135: Could we rename this to `readShallowCopy` to make it abundantly clear in call sites that this is not safe if `T` contains pointers? Otherwise a comptime check would be ideal because I can't think of a use case where it would make sense to use this function on a type which has pointers because it would defeat the purpose of the sync primitive. Currently it is only used for copying a `Hash` but it has potential to be misused

yewman (commented):
Just one comment otherwise looks really good, very clean and readable!

yewman (dismissed):

src/sync/mux.zig
dnut: e0f9a18af89de096c268ba8c453cbcbad62027cc

I went with the comptime check because I think it provides the best safety.

---------------------------------------------------

And here's a more long-winded explanation of my thought process...

Typically, I don't think it makes a lot of sense to contain pointers within a RwMux. The whole idea of the RwMux is that the state of the lock needs to be shared. It doesn't just wrap a pointer, it needs to be the pointer. Passing around a `RwMux(*Hash)` to different threads wouldn't work because you'd be copying the RwMux everywhere and none of the threads would respect each other's locks. You need to pass `*RwMux(Hash)` to those threads so the lock is respected.

That being said, I could imagine that hypothetically somebody might use a RwMux correctly to protect a type that also has some mutable pointers inside it. Something like to `*RwMux(struct { data: u64, state: *Something })`. In that case, it would be a misuse of the RwMux to copy the pointer, unlock the mux, then read from the pointer. To be clear, you don't need `readCopy` to perform this misuse. You can easily do this by just directly using `read` and `unlock`. That would be foolish but certainly it's easy to do. In my mind, using readCopy is exactly the same pattern. It's something foolish you can do, and it's on you to avoid that mistake.

The difference is that there's no way to implement `read` and `unlock` to prevent this footgun. But `readCopy` actually gives us a great opportunity. Adding the comptime check to disallow its usage with mutable pointers would *only* block the misuse scenario. So actually, encouraging the use of `readCopy` for this use case could improve the safety of our code, as compared to having people always use `read` and `unlock` and trying to use their best judgement about whether it's safe to read from copies after unlocking.

src/replay/trackers.zig
dnut: 19052ecfe1a99abda3faa8e428da8906ab1149df

InKryption (approved):

Rexicon226 (approved):

src/core/bank.zig
dadepo: The question arose while implementing the `parents` function in https://github.com/Syndica/sig/pull/762/files#diff-87e121578ca9adc5b886321addf8c514e88ba018158ff9bbeb1e3183a5c67a46R43 where `current.constants.parent_slot` is continually access to get the parents.

So I was wondering, till how far back will the `parent_slot` point? I doubt it would be up until genesis, so I assume there will be a point where `parent_slot` is not set?

Currently using `parent_slot == current_slot` to terminate the loop that continually retrieves the parent slots. Is there a better way?

_I'll unresolve this, as the conclusion might affect the implementation here or the implementation in the PR I was working on._

src/replay/execution.zig
```diff
@@ -0,0 +1,212 @@
...
+        // results. this will change once we call the svm and process the
+        // results of execution.
+        _, const status = slot_status;
+        while (status == .confirm and try status.confirm.poll() == .pending) {
+            std.time.sleep(std.time.ns_per_ms); // TODO: consider futex-based wait like ResetEvent
+        }
+        status.deinit(state.allocator);
+    }
+
+    // TODO: process_replay_results: https://github.com/anza-xyz/agave/blob/3f68568060fd06f2d561ad79e8d8eb5c5136815a/core/src/replay_stage.rs#L3443
```
dadepo, line 95: Digging deeper into the consensus related aspects of replay, finding a couple more operations that affects consensus. The `process_replay_results` is one of such where it contains logic that reads/updates consensus related datastructure like the forkchoice. (specifically via the `mark_dead_slot`)

Given that this PR contains a lot of changes already, this is probably something to push in a follow up PR, but needs some coordination.

dadepo (approved):

src/replay/execution.zig
dnut: Yes I was planning to work on this in the next PR and we should discuss it since it touches consensus.

src/core/bank.zig
dnut: > I doubt it would be up until genesis, so I assume there will be a point where parent_slot is not set?

The parent slot will always be set. If we have any of this data about a slot, we should know what its parent slot is. 

That doesn't mean the parent slot is in the map. So your loop needs to handle the possibility that the parent slot is set to a value that is not in the map. That's when it would terminate.

The slot tracker will never have the genesis slot unless it's bootstrapping a new cluster. When you start from a snapshot, that will be the oldest slot in here. Then as soon as a slot is rooted, all slots older than it should be removed from the slot tracker to avoid leaking memory.

Your loop can just continue until either it hits genesis (should basically never happen in practice) or the parent slot is not in the map. There is no need for the parent_slot to be null. You definitely wouldn't want to use that field to indicate whether or not the parent slot is in this map.

PR #723, opened 2025-05-07
feat(batch execution): status cache transaction checks

Sobeston:
Based on top of https://github.com/Syndica/sig/pull/689, will rebase when merged.

Implements part of transaction checking.


src/core/status_cache.zig
```diff
@@ -0,0 +1,71 @@
...
+    const statusValElement = struct { key_slice: KeySlice, maybe_err: T };
+    const StatusVal = ArrayList(statusValElement);
+    // TODO: might be able to get rid of this RwMux, agave has one here
+    const StatusKv = RwMux(HashMap(Hash, StatusVal));
+    const KeyMap = HashMap(KeySlice, ForkStatus);
+
+    const HighestFork = struct { slot: Slot, index: usize, key_map: KeyMap };
+
+    const KeyStatusMap = HashMap(Hash, HighestFork);
+    const SlotDeltaMap = HashMap(Slot, StatusKv);
```
yewman, line 34: Lets remove these if they are unused, the unaliased definition is lot that long

src/core/status_cache.zig
```diff
@@ -0,0 +1,71 @@
...
+            .cache = .{},
+            .roots = roots,
+            .slot_deltas = .{},
+        };
+    }
+
+    pub fn getStatus(
+        self: *const StatusCache,
+        key: []const u8,
+        tx_blockhash: *const Hash,
```
yewman, line 49:
```suggestion
        blockhash: *const Hash,
```

src/core/status_cache.zig
```diff
@@ -0,0 +1,71 @@
...
+const RwMux = sig.sync.RwMux;
+
+// StatusCache is only used with <Result<(), TransactionError>>
+const T = ?sig.ledger.transaction_status.TransactionError;
+const Hash = sig.core.Hash;
+const Slot = sig.core.Slot;
+
+const Fork = struct { slot: Slot, maybe_err: T };
+
+pub const StatusCache = struct {
```
yewman, line 15: Lets write some unit tests for coverage on this data structure

src/runtime/check_transactions.zig
```diff
@@ -0,0 +1,208 @@
...
+    blockhash_queue: *const BlockhashQueue,
+    max_age: u64,
+    last_blockhash: *const Hash,
+    next_durable_nonce: *const Hash,
+    next_lamports_per_signature: u64,
+) TransactionResult(?CachedAccount) {
+    if (blockhash_queue.getHashInfoIfValid(last_blockhash, max_age)) |hash_info| {
+        _ = hash_info;
+        return .{ .ok = null };
+    }
```
yewman, line 44:
```suggestion
    if (blockhash_queue.getHashInfoIfValid(last_blockhash, max_age) != null) {
        return .{ .ok = null };
    }
```

src/runtime/check_transactions.zig
```diff
@@ -0,0 +1,208 @@
...
+    if (checkLoadAndAdvanceMessageNonceAccount(
+        transaction,
+        next_durable_nonce,
+        next_lamports_per_signature,
+        batch_account_cache,
+    )) |nonce| {
+        const nonce_account, const previous_lamports_per_signature = nonce;
+        _ = previous_lamports_per_signature;
+        return .{ .ok = nonce_account };
+    }
```
yewman, line 55:
```suggestion
        const nonce_account, _ = nonce;
        return .{ .ok = nonce_account };
    }
```

src/runtime/check_transactions.zig
```diff
@@ -0,0 +1,208 @@
...
+}
+
+fn checkLoadAndAdvanceMessageNonceAccount(
+    transaction: *const RuntimeTransaction,
+    next_durable_nonce: *const Hash,
+    next_lamports_per_signature: u64,
+    batch_account_cache: *BatchAccountCache,
+) ?struct { CachedAccount, u64 } {
+    const nonce_is_advanceable = !transaction.recent_blockhash.eql(next_durable_nonce.*);
+    if (!nonce_is_advanceable) return null;
```
yewman, line 67:
```suggestion
    if (transaction.recent_blockhash.eql(next_durable_nonce.*)) return null;
```

src/runtime/transaction_execution.zig
```diff
@@ -350,22 +348,21 @@ pub fn executeTransaction(
...
-    _ = ancestors;
-    _ = batch_account_cache;
-    _ = blockhash_queue;
-    _ = max_age;
-    _ = last_blockhash;
-    _ = next_durable_nonce;
-    _ = next_lamports_per_signature;
-    @panic("not implemented");
+) TransactionResult(?CachedAccount) {
+    return sig.runtime.check_transactions.checkAge(
```
yewman, line 357: lets just call this function directly

src/runtime/transaction_execution.zig
```diff
@@ -375,11 +372,12 @@ pub fn checkStatusCache(
     ancestors: *const Ancestors,
     status_cache: *const StatusCache,
 ) ?TransactionError {
-    _ = msg_hash;
-    _ = recent_blockhash;
-    _ = status_cache;
-    _ = ancestors;
-    @panic("not implemented");
+    return sig.runtime.check_transactions.checkStatusCache(
```
yewman, line 375: Same with this one

yewman (commented):
A few comments from a first pass

src/runtime/check_transactions.zig
Sobeston: fixed in bf5e9ac9f652b24048bab8ef943f920ec25d2312

src/runtime/check_transactions.zig
Sobeston: fixed in bf5e9ac9f652b24048bab8ef943f920ec25d2312

src/core/status_cache.zig
Sobeston: fixed in bf5e9ac9f652b24048bab8ef943f920ec25d2312

src/core/status_cache.zig
Sobeston: removed in bf5e9ac9f652b24048bab8ef943f920ec25d2312

src/core/status_cache.zig
```diff
@@ -0,0 +1,71 @@
...
+        const stored_forks: ArrayList(Fork) = map.key_map.get(key_slice) orelse return null;
+        return for (stored_forks.items) |fork| {
+            if (ancestors.ancestors.contains(fork.slot) or self.roots.contains(fork.slot)) {
+                break fork;
+            }
+        } else null;
+    }
+};
+
+pub const Ancestors = struct {
```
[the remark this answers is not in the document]
Sobeston: I think the one in the bank is primarily for (de)serialisation, semantically I think it should be a set

src/runtime/check_transactions.zig
```diff
@@ -0,0 +1,211 @@
...
+            .Initialized = .{
+                .authority = nonce_data.authority,
+                .durable_nonce = next_durable_nonce.*,
+                .lamports_per_signature = next_lamports_per_signature,
+            },
+        },
+    };
+
+    // could probably be smaller
+    var serialize_buf: [@sizeOf(NonceData) * 2]u8 = undefined;
```
yewman, line 89: why not just `@sizeOf(NonceVersions)`? Am I missing something...

src/core/status_cache.zig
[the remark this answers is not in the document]
yewman: Hmm, would it be possible to write a custom ser/der for your implementation of Ancestors and update `core/bank.zig` accordingly?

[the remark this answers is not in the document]
Sobeston: Yeah maybe. Agave itself does have two types for it. @InKryption?

src/runtime/check_transactions.zig
Sobeston: In case serialised size is different. Corrected: 0fef86648b5e063cde3ecd4805d544c64089fa36

src/runtime/check_transactions.zig
```diff
@@ -0,0 +1,211 @@
...
+        if (signer.equals(&nonce_data.authority)) break;
+    } else return null;
+
+    return .{
+        .{ .pubkey = nonce_address, .account = nonce_account },
+        nonce_data,
+    };
+}
+
+const NonceData = struct {
```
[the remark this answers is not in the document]
Sobeston: Fixed in 6caffa5f4be393ab53386b80222879928a5caea9

src/core/status_cache.zig
Sobeston: added some tests: 88f3538fc5855f329dd22c490347bf0f213280af

src/core/status_cache.zig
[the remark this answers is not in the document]
InKryption: I think unifying it under one type would probably be a nice way to avoid needing to copy the state in between the types - afaict, the only reason there's two types in agave (for this, and for many other things) is because that's just how they separate the "serialization" layer from the internals, and I don't think we need to replicate that convention.

If you go ahead and do that, also make this unmanaged, and add a proper deinit method please.

src/runtime/check_transactions.zig
yewman: We are serializing nonce.Versions into the account data, not nonce.State right? So the size should be 88 if I am not mistaken. Why would the seralized size be different? The size of the nonce.Versions is comptime known

Sobeston: `@sizeOf(Versions)` is 88
but according to this test, it's serialised as 80 bytes:
https://github.com/anza-xyz/solana-sdk/blob/51e1da20ab83511563bd400cb448c2fee4ac4db6/nonce/src/state.rs#L104

Annoyingly agave attaches this size const to the State, rather than the Versions. However it's actually describing the Versions size.

Size:
32 (pubkey) + 32 (hash) + 8 (lamports per signature) + 4 + 4 (two bincoded enum values) = 80

(not sure why it's 88 in memory?)

cleaned up the const in 9fe77d1d5bc348b6dfb5bfde33ca5099d177b301

src/core/status_cache.zig
```diff
@@ -186,9 +187,71 @@ pub const StatusCache = struct {
...
+        // as if it's 0, and deserialize 0 as if it's void.
+        const VoidCfg = struct {
+            fn deserialize(alloc: std.mem.Allocator, reader: anytype, params: bincode.Params) !void {
+                _ = try bincode.read(alloc, usize, reader, params);
+            }
+            fn serialize(writer: anytype, data: anytype, params: bincode.Params) !void {
+                _ = data;
+                try bincode.write(writer, @as(usize, 0), params);
+            }
+        };
```
InKryption, line 208: Could move this namespace outside of the function, and can just inline the return value into where `mapConfig()` is, no point in having this wrapper function imo.

Sobeston: e27778a0f97665fe9e2b39185155e7a4a28a80e7

src/runtime/check_transactions.zig
yewman: the u64 in the FeeCalculator right?

yewman (approved):
Good to go once we have code cov :)

InKryption (approved):

PR #724, opened 2025-05-07
refactor(consensus, tower): Clearer Tower API

dadepo:
This refactor PR mostly moves things around. 

The objective is to simplify the Tower API and make it more explicit. 

Methods were extracted into a new structure (currently named `ReplayTower` - other suggestion for the new name is `Consensus`) which houses the extra methods that is used during Replay. Some of these methods depends on the extracted tower, while some don't.

### DIFF
--- a/src/consensus/tower.zig
+++ b/src/consensus/tower.zig
@@ -1979,8 +1979,8 @@ fn greatestCommonAncestor(
     slot_a: Slot,
     slot_b: Slot,
 ) ?Slot {
-    var ancestors_a = ancestors.get(slot_a) orelse return null;
-    var ancestors_b = ancestors.get(slot_b) orelse return null;
+    const ancestors_a = ancestors.get(slot_a) orelse return null;
+    const ancestors_b = ancestors.get(slot_b) orelse return null;
 
     var max_slot: ?Slot = null;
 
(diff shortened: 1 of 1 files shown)


src/consensus/tower.zig
```diff
@@ -1982,136 +1470,729 @@ pub const Tower = struct {
...
-                &ProgressMap.INIT,
-                &latest,
-                &fork_choice,
-                .{},
-                &slot_history,
-            ),
-        );
+    var max_slot: ?Slot = null;
+
+    var superset, const subset = if (ancestors_a.count() >= ancestors_b.count())
```
InKryption, line 1987: Should `const`ify some of these vars

src/consensus/tower.zig
```diff
@@ -1982,136 +1470,729 @@ pub const Tower = struct {
...
+// Optimistically skip the stake check if casting a vote would not increase
+// the lockout at this threshold. This is because if you bounce back to
+// voting on the main fork after not voting for a while, your latest vote
+// might pop off a lot of the votes in the tower. The stake from these votes
+// would have rolled up to earlier votes in the tower, which presumably
+// could have helped us pass the threshold check. Worst case, we'll just
+// recheck later without having increased lockouts.
+fn optimisticallyBypassVoteStakeThresholdCheck(
+    // Needs to be an iterator that produces Lockout
+    tower_before_applying_vote: anytype,
```
InKryption, line 2011: This comment appears to be inaccurate/misleading, and this type could maybe be changed to just be a `[]const Slot`?

src/consensus/tower.zig
```diff
@@ -1982,136 +1470,729 @@ pub const Tower = struct {
...
+    threshold_depth: usize,
+    threshold_size: f64,
+    slot: Slot,
+    voted_stakes: *const AutoHashMapUnmanaged(Slot, u64),
+    total_stake: u64,
+) ThresholdDecision {
+    const threshold_vote = maybe_threshold_vote orelse {
+        // Tower isn't that deep.
+        return ThresholdDecision{ .passed_threshold = {} };
```
InKryption, line 2036: Why not just `return .passed_threshold;`?

src/consensus/tower.zig
```diff
@@ -1982,136 +1470,729 @@ pub const Tower = struct {
...
+pub fn populateAncestorVotedStakes(
+    voted_stakes: *SortedSet(Slot),
+    vote_slots: []const Slot,
+    ancestors: *const AutoHashMapUnmanaged(Slot, SortedSet(Slot)),
+) !void {
+    // If there's no ancestors, that means this slot must be from before the current root,
+    // in which case the lockouts won't be calculated in bank_weight anyways, so ignore
+    // this slot
+    for (vote_slots) |vote_slot| {
+        if (ancestors.get(vote_slot)) |maybe_slot_ancestors| {
```
InKryption, line 2085: Should perhaps be using `ancestors.getPtr` so that the copy isn't necessary, and so that the sort set is actually be sorted?

InKryption (changes requested):
Just a few comments, otherwise overall it lgtm.

src/consensus/tower.zig
dadepo: [87a06af](https://github.com/Syndica/sig/pull/724/commits/87a06af898e3faa6cb3745d438fe4be1d43350cd)

src/consensus/tower.zig
dadepo: [2e2ebce](https://github.com/Syndica/sig/pull/724/commits/2e2ebce2b083c0bab7074c204a36b93212c68053)

InKryption (approved):
Changes LGTM

dadepo:
Closing in favour of #728

PR #729, opened 2025-05-09
refactor(consensus, forkchoice) Add `forkchoice.heaviestSlotOnSameVotedFork` method

dadepo:
This adds the implementation for heaviestSlotOnSameVotedFork.

This could not be implemented until now because it depends on Tower.

### DIFF
--- a/src/consensus/fork_choice.zig
+++ b/src/consensus/fork_choice.zig
@@ -11,7 +11,7 @@ const Pubkey = sig.core.Pubkey;
 const SortedMap = sig.utils.collections.SortedMap;
 const SlotAndHash = sig.core.hash.SlotAndHash;
 const Slot = sig.core.Slot;
-const ReplayTower = sig.consensus.tower.ReplayTower;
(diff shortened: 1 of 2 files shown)


dadepo:
The Agave version has two tests. One was implemented. The second test depends on `heaviest_subtree_fork_choice.add_votes` which was not implemented until now because it depends on input retrieved from `Bank`.

The `heaviest_subtree_fork_choice.add_votes` method can now be implemented, and would be done in a follow up PR. This is still possible even though there is no direct `Bank` struct, the convention of passing in the exact inputs directly would be followed (instead of via a Bank) and in this case the `epoch_stakes` and `epoch_schedule`.

dadepo:
> The second test depends on `heaviest_subtree_fork_choice.add_votes` which was not implemented until now because it depends on input retrieved from `Bank`

Further look into `heaviest_subtree_fork_choice.add_votes` shows it is only used in tests and in a struct called `RepairWeight` that has not popped up yet for implementation. In the meantime I manually added tests that covers the other code paths in the new code.

dnut (approved):

PR #730, opened 2025-05-09
fix(svm): cpi conformance

kprotty:
Fix cpi syscalls to pass conformance tests. Paired with https://github.com/Syndica/sig-fuzz/pull/8

### DIFF
--- a/src/vm/syscalls/cpi.zig
+++ b/src/vm/syscalls/cpi.zig
@@ -153,14 +153,14 @@ fn VmValue(comptime T: type) type {
             memory_map: *const MemoryMap,
             check_aligned: bool,
         },
-        translated_ptr: usize,
+        translated_addr: usize,
 
         pub fn get(self: Self, comptime state: memory.MemoryState) !(switch (state) {
             .constant => *align(1) const T,
             .mutable => *align(1) T,
         }) {
             switch (self) {
-                .translated_ptr => |ptr| return @ptrFromInt(ptr),
+                .translated_addr => |ptr| return @ptrFromInt(ptr),
                 .vm_address => |vma| {
                     return vma.memory_map.translateType(T, state, vma.vm_addr, vma.check_aligned);
                 },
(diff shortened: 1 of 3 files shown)


src/vm/syscalls/cpi.zig
```diff
@@ -150,14 +153,14 @@ fn VmValue(comptime T: type) type {
             memory_map: *const MemoryMap,
             check_aligned: bool,
         },
-        translated: *T,
+        translated_ptr: usize,
```
Rexicon226, line 156: nit: `translated_addr`

src/vm/syscalls/cpi.zig
```diff
@@ -623,6 +626,18 @@ fn translateAccounts(
         }
     }
 
+    var account_info_keys = std.ArrayList(*const Pubkey).init(allocator);
```
Rexicon226, line 629: We shouldn't be allocating here. Generally speaking, allocation inside syscalls is very rare. Either use a bounded array, since we know the theoretical maximum number of account infos, or only check the address in this loop, and then do the transaction later on.

src/vm/syscalls/cpi.zig
```diff
@@ -637,6 +652,13 @@ fn translateAccounts(
         var callee_account = try ic.borrowInstructionAccount(meta.index_in_caller);
         defer callee_account.release();
 
+        const account_key = blk: {
```
Rexicon226, line 655: Does this need to be an block? Just wondering why, is it somehow harmful to have `account_meta` in the parent scope?

yewman (approved):
lgtm

src/vm/syscalls/cpi.zig
kprotty: There's `meta` already in scope, so another `account_meta` seems extra

src/vm/syscalls/cpi.zig
kprotty: https://github.com/Syndica/sig/pull/730/commits/7b4b87863fc24a9668cb54065daa7651b8534c3c

src/vm/syscalls/cpi.zig
kprotty: https://github.com/Syndica/sig/pull/730/commits/7b4b87863fc24a9668cb54065daa7651b8534c3c

Rexicon226 (approved):

PR #731, opened 2025-05-09
refactor(consensus): split tower.zig into two files

dnut:
This just moves some code out of tower.zig into a new file called replay_tower.zig.

In #728, @dadepo split Tower into two structs: Tower and ReplayTower. This was a really nice clean division that makes it clear what's really the tower data structure versus the operations and state that build on top of it to make decisions in replay. I'd just like to take this a step further by splitting them into separate files to increase the clarity even more.

The conceptual division between tower/not-tower actually runs not only between the two structs, but actually between two collections of data types and functions. Every decl and test in these files can easily be classified as being for either tower or replay tower. Currently they're all just scattered arbitrarily in tower.zig, so it's not entirely clear what their purposes are. Giving each set its own file makes the distinction much more apparent.

Also the file was getting pretty massive and it's easy to get lost in files that large. Alternatively, having them side by side in your editor is nice for maximizing context, as you can see the caller and the callee at the same time.


dadepo (commented):
Looks good. Just not sure yet why the codecov needs to fail for this, since it is just moving stuff to a different file

dnut:
> Looks good. Just not sure yet why the codecov needs to fail for this, since it is just moving stuff to a different file

We can just ignore that, it's not changing anything from main. It's just a limitation of codecov.

Codecov isn't reporting lower test coverage for this code than before. We merged previous PRs with lower numbers for these files because the overall diff met the requirement. This time the change is honing in on code that has slightly below average coverage, and codecov doesn't recognize that it's the same code, so it's flagging it as a failure.

Maybe we should improve the coverage in these files, but I don't think it's in scope for this PR since there is no new or modified logic here.

dadepo (approved):

src/consensus/replay_tower.zig
```diff
@@ -0,0 +1,2735 @@
...
+    defer stakes.deinit(std.testing.allocator);
+    try stakes.ensureTotalCapacity(std.testing.allocator, 1);
+
+    stakes.putAssumeCapacity(0, 2);
+
+    const result = isSlotConfirmed(&tower, 0, &stakes, 2);
+    try std.testing.expect(result);
+}
+
+test "default tower has no stray last vote" {
```
dadepo, line 2486: Sorry I missed this in the PR. Only noticing now that the `test:` prefix were removed. 

Is there any benefit to not having the prefix?

One benefit to having it, is that it comes in handy when filtering tests to run. ie: `zig build test -Dfilter="tower:"`.

Also, although not exactly matching what is in the style guide, https://github.com/Syndica/sig/blob/f66a2c7a81b8052d4645c372371686a911b6e6d1/docs/docusaurus/docs/contributing/style-guide.mdx?plain=1#L142 but it does come close.

PR #733, opened 2025-05-14
feat(compute_budget): implement compute budget and related data structures

yewman:
- Implement Compute Budget Program
- Implement Builtin Costs 
- Refactor some program definitions


src/runtime/program/builtin_costs.zig
```diff
@@ -0,0 +1,141 @@
...
+    @setEvalBranchQuota(10_000);
+    const entries = MIGRATING_BUILTIN_COSTS ++ NON_MIGRATING_BUILTIN_COSTS;
+    return std.StaticStringMap(BuiltinCost).initComptime(&entries);
+}
+
+fn initMaybeBuiltinKey() [256]bool {
+    var table = [_]bool{false} ** 256;
+    for (BUILTIN_COSTS.keys()) |key| table[key[0]] = true;
+    return table;
+}
```
Sobeston, line 31: I know agave does this this, but is it really necessary? Feels confusing.

(not going to block approval over it, if agave does it then it works)

src/runtime/program/compute_budget/lib.zig
```diff
@@ -1,128 +1,798 @@
...
+const Pubkey = sig.core.Pubkey;
+const Packet = sig.net.Packet;
+const FeatureSet = sig.runtime.FeatureSet;
 const InstructionInfo = sig.runtime.InstructionInfo;
+const TransactionError = sig.runtime.transaction_error.TransactionError;
+const TransactionResult = sig.runtime.transaction_error.TransactionResult;
 
-pub const ID = sig.runtime.ids.COMPUTE_BUDGET_PROGRAM_ID;
+const MIGRATING_BUILTIN_COSTS = builtin_costs.MIGRATING_BUILTIN_COSTS;
+const MAX_PUBKEYS_PER_PACKET = Packet.DATA_SIZE / Pubkey.SIZE;
```
Sobeston, line 16: Cool constant. Does this still hold up after we've loaded accounts from the address lookup table?

Sobeston (commented):
A couple Qs:

src/runtime/program/builtin_costs.zig
yewman: I agree with your assessment here. It primarily exists for managing core bpf migrations and modifies the behaviour of the compute budget program. Since the core bpf migrations will not be complete by the time we are running on test net we will likely require the associated handling logic. We could probably implement something simpler however that would come with some risk which I would rather avoid right now given timelines

src/runtime/program/compute_budget/lib.zig
yewman: Hmm, thats a good point. I have changed it to use Transaction.MAX_ACCOUNTS instead which is 128
4436f8fb

Sobeston (dismissed):

Rexicon226:
I'd like to take a look at this after the conflicts are resolved :).

Rexicon226 (approved):

PR #745, opened 2025-05-22
chore: update to zig 0.14.1

Rexicon226:
We still need to wait for the 0.14.1 binaries to be uploaded to the website, after that the CI should (hopefully) pass. Should be good enough to review in the mean time!


InKryption (approved):
Been over most of it, everything looks nominal, great to have the updated formatting, removal of the `defaultTargetDetectM3` workaround, and the workarounds added in consideration for playing with the x86_64 backend.

I see some changes that are probably going to conflict with at least one of my PRs (#740), but such is the nature of big version updates.

InKryption (approved):

PR #762, opened 2025-05-27
feat(replay): consensus integration v1

dadepo:
Implements the consensus aspects in replay.

Features to be added in follow up PRs.
- In `handle_votable_bank` all logic that involves the following data structures. Once these data structures (or their analogous) lands on main, the implementation can be updated.
  - LeaderScheduleCache
  - AggregateCommitmentService
  - SnapshotController
  - RpcSubscriptions
  - BlockCommitmentCache
  - BankNotificationSenderConfig
  - DuplicateSlotsTracker
  - DuplicateConfirmedSlots
  - UnfrozenGossipVerifiedVoteHashes
  - EpochSlotsFrozenSlots
- Logic involved generating vote transactions
- Updates to blockstore's `slots_stats` (`blockstore.slots_stats.mark_rooted(new_root)`) will be done when analogous feature is added to sig's blockstore
- Pushing vote to the network
- Processing fork rest.

### DIFF
--- a/src/consensus/progress_map.zig
+++ b/src/consensus/progress_map.zig
@@ -463,7 +463,7 @@ pub const ForkStats = struct {
     vote_threshold: VoteThreshold,
     is_locked_out: bool,
     voted_stakes: consensus.VotedStakes,
-    duplicate_confirmed_hash: ?Hash,
+    duplicate_confirmed_hash: Hash,
     computed: bool,
     lockout_intervals: LockoutIntervals,
     bank_hash: Hash,
--- a/src/consensus/replay_tower.zig
+++ b/src/consensus/replay_tower.zig
@@ -3453,9 +3453,9 @@ fn fillProgressMapForkStats(
     }
 }
 
-const MAX_TEST_TREE_LEN = 100;
+pub const MAX_TEST_TREE_LEN = 100;
 const Tree = struct { root: SlotAndHash, data: std.BoundedArray(TreeNode, MAX_TEST_TREE_LEN) };
-const TestFixture = struct {
+pub const TestFixture = struct {
     fork_choice: HeaviestSubtreeForkChoice,
     ancestors: AutoArrayHashMapUnmanaged(Slot, SortedSet(Slot)) = .{},
     descendants: AutoArrayHashMapUnmanaged(Slot, SortedSet(Slot)) = .{},
--- a/src/consensus/replay_tower.zig
+++ b/src/consensus/replay_tower.zig
@@ -1504,7 +1504,7 @@ fn checkVoteStakeThreshold(
     threshold_depth: usize,
     threshold_size: f64,
     slot: Slot,
-    voted_stakes: *const AutoHashMapUnmanaged(Slot, u64),
+    voted_stakes: *const VotedStakes,
     total_stake: u64,
 ) ThresholdDecision {
     const threshold_vote = maybe_threshold_vote orelse {
@@ -1574,7 +1574,7 @@ test "check vote threshold without votes" {
     var tower = try createTestReplayTower(1, 0.67);
     defer tower.deinit(std.testing.allocator);
 
-    var stakes = AutoHashMapUnmanaged(u64, u64){};
+    var stakes = VotedStakes{};
     defer stakes.deinit(std.testing.allocator);
     try stakes.ensureTotalCapacity(std.testing.allocator, 1);
 
@@ -1594,7 +1594,7 @@ test "check vote threshold no skip lockout with new root" {
     var replay_tower = try createTestReplayTower(4, 0.67);
     defer replay_tower.deinit(std.testing.allocator);
 
-    var stakes = AutoHashMapUnmanaged(u64, u64){};
+    var stakes = VotedStakes{};
     defer stakes.deinit(std.testing.allocator);
     try stakes.ensureTotalCapacity(std.testing.allocator, MAX_LOCKOUT_HISTORY);
 
@@ -1819,7 +1819,7 @@ test "check vote threshold below threshold" {
     var replay_tower = try createTestReplayTower(1, 0.67);
     defer replay_tower.deinit(std.testing.allocator);
 
-    var stakes = AutoHashMapUnmanaged(u64, u64){};
+    var stakes = VotedStakes{};
     defer stakes.deinit(std.testing.allocator);
     try stakes.ensureTotalCapacity(std.testing.allocator, 1);
 
@@ -1845,7 +1845,7 @@ test "check vote threshold above threshold" {
     var replay_tower = try createTestReplayTower(1, 0.67);
     defer replay_tower.deinit(std.testing.allocator);
 
-    var stakes = AutoHashMapUnmanaged(u64, u64){};
+    var stakes = VotedStakes{};
     defer stakes.deinit(std.testing.allocator);
     try stakes.ensureTotalCapacity(std.testing.allocator, 1);
 
@@ -1871,7 +1871,7 @@ test "check vote thresholds above thresholds" {
     var tower = try createTestReplayTower(VOTE_THRESHOLD_DEPTH, 0.67);
     defer tower.deinit(std.testing.allocator);
 
-    var stakes = AutoHashMapUnmanaged(u64, u64){};
+    var stakes = VotedStakes{};
     defer stakes.deinit(std.testing.allocator);
     try stakes.ensureTotalCapacity(std.testing.allocator, 3);
 
@@ -1902,7 +1902,7 @@ test "check vote threshold deep below threshold" {
     var tower = try createTestReplayTower(VOTE_THRESHOLD_DEPTH, 0.67);
     defer tower.deinit(std.testing.allocator);
 
-    var stakes = AutoHashMapUnmanaged(u64, u64){};
+    var stakes = VotedStakes{};
     defer stakes.deinit(std.testing.allocator);
     try stakes.ensureTotalCapacity(std.testing.allocator, 2);
 
@@ -1932,7 +1932,7 @@ test "check vote threshold shallow below threshold" {
     var tower = try createTestReplayTower(VOTE_THRESHOLD_DEPTH, 0.67);
     defer tower.deinit(std.testing.allocator);
 
-    var stakes = AutoHashMapUnmanaged(u64, u64){};
+    var stakes = VotedStakes{};
     defer stakes.deinit(std.testing.allocator);
     try stakes.ensureTotalCapacity(std.testing.allocator, 2);
 
@@ -1962,7 +1962,7 @@ test "check vote threshold above threshold after pop" {
     var tower = try createTestReplayTower(1, 0.67);
     defer tower.deinit(std.testing.allocator);
 
-    var stakes = AutoHashMapUnmanaged(u64, u64){};
+    var stakes = VotedStakes{};
     defer stakes.deinit(std.testing.allocator);
     try stakes.ensureTotalCapacity(std.testing.allocator, 1);
 
@@ -1991,7 +1991,7 @@ test "check vote threshold above threshold no stake" {
     var tower = try createTestReplayTower(1, 0.67);
     defer tower.deinit(std.testing.allocator);
 
-    var stakes = AutoHashMapUnmanaged(u64, u64){};
+    var stakes = VotedStakes{};
     defer stakes.deinit(std.testing.allocator);
 
     _ = try tower.recordBankVote(
@@ -2015,7 +2015,7 @@ test "check vote threshold lockouts not updated" {
     var tower = try createTestReplayTower(1, 0.67);
     defer tower.deinit(std.testing.allocator);
 
-    var stakes = AutoHashMapUnmanaged(u64, u64){};
+    var stakes = VotedStakes{};
     defer stakes.deinit(std.testing.allocator);
     try stakes.ensureTotalCapacity(std.testing.allocator, 2);
 
@@ -2821,7 +2821,7 @@ test "is slot confirmed not enough stake failure" {
     var tower = try createTestReplayTower(1, 0.67);
     defer tower.deinit(std.testing.allocator);
 
-    var stakes = AutoHashMapUnmanaged(u64, u64){};
+    var stakes = VotedStakes{};
     defer stakes.deinit(std.testing.allocator);
     try stakes.ensureTotalCapacity(std.testing.allocator, 1);
 
@@ -2835,7 +2835,7 @@ test "is slot confirmed unknown slot" {
     var tower = try createTestReplayTower(1, 0.67);
     defer tower.deinit(std.testing.allocator);
 
-    var stakes = AutoHashMapUnmanaged(u64, u64){};
+    var stakes = VotedStakes{};
     defer stakes.deinit(std.testing.allocator);
 
     const result = isSlotConfirmed(&tower, 0, &stakes, 2);
@@ -2846,7 +2846,7 @@ test "is slot confirmed pass" {
     var tower = try createTestReplayTower(1, 0.67);
     defer tower.deinit(std.testing.allocator);
 
-    var stakes = AutoHashMapUnmanaged(u64, u64){};
+    var stakes = VotedStakes{};
     defer stakes.deinit(std.testing.allocator);
     try stakes.ensureTotalCapacity(std.testing.allocator, 1);
 
--- a/src/consensus/replay_tower.zig
+++ b/src/consensus/replay_tower.zig
@@ -1574,7 +1574,7 @@ test "check vote threshold without votes" {
     var tower = try createTestReplayTower(1, 0.67);
     defer tower.deinit(std.testing.allocator);
 
-    var stakes = VotedStakes{};
+    var stakes = VotedStakes.empty;
     defer stakes.deinit(std.testing.allocator);
     try stakes.ensureTotalCapacity(std.testing.allocator, 1);
 
@@ -1594,7 +1594,7 @@ test "check vote threshold no skip lockout with new root" {
     var replay_tower = try createTestReplayTower(4, 0.67);
     defer replay_tower.deinit(std.testing.allocator);
 
-    var stakes = VotedStakes{};
+    var stakes = VotedStakes.empty;
     defer stakes.deinit(std.testing.allocator);
     try stakes.ensureTotalCapacity(std.testing.allocator, MAX_LOCKOUT_HISTORY);
 
@@ -1819,7 +1819,7 @@ test "check vote threshold below threshold" {
     var replay_tower = try createTestReplayTower(1, 0.67);
     defer replay_tower.deinit(std.testing.allocator);
 
-    var stakes = VotedStakes{};
+    var stakes = VotedStakes.empty;
     defer stakes.deinit(std.testing.allocator);
     try stakes.ensureTotalCapacity(std.testing.allocator, 1);
 
@@ -1845,7 +1845,7 @@ test "check vote threshold above threshold" {
     var replay_tower = try createTestReplayTower(1, 0.67);
     defer replay_tower.deinit(std.testing.allocator);
 
-    var stakes = VotedStakes{};
+    var stakes = VotedStakes.empty;
     defer stakes.deinit(std.testing.allocator);
     try stakes.ensureTotalCapacity(std.testing.allocator, 1);
 
@@ -1871,7 +1871,7 @@ test "check vote thresholds above thresholds" {
     var tower = try createTestReplayTower(VOTE_THRESHOLD_DEPTH, 0.67);
     defer tower.deinit(std.testing.allocator);
 
-    var stakes = VotedStakes{};
+    var stakes = VotedStakes.empty;
     defer stakes.deinit(std.testing.allocator);
     try stakes.ensureTotalCapacity(std.testing.allocator, 3);
 
@@ -1902,7 +1902,7 @@ test "check vote threshold deep below threshold" {
     var tower = try createTestReplayTower(VOTE_THRESHOLD_DEPTH, 0.67);
     defer tower.deinit(std.testing.allocator);
 
-    var stakes = VotedStakes{};
+    var stakes = VotedStakes.empty;
     defer stakes.deinit(std.testing.allocator);
     try stakes.ensureTotalCapacity(std.testing.allocator, 2);
 
@@ -1932,7 +1932,7 @@ test "check vote threshold shallow below threshold" {
     var tower = try createTestReplayTower(VOTE_THRESHOLD_DEPTH, 0.67);
     defer tower.deinit(std.testing.allocator);
 
-    var stakes = VotedStakes{};
+    var stakes = VotedStakes.empty;
     defer stakes.deinit(std.testing.allocator);
     try stakes.ensureTotalCapacity(std.testing.allocator, 2);
 
@@ -1962,7 +1962,7 @@ test "check vote threshold above threshold after pop" {
     var tower = try createTestReplayTower(1, 0.67);
     defer tower.deinit(std.testing.allocator);
 
-    var stakes = VotedStakes{};
+    var stakes = VotedStakes.empty;
     defer stakes.deinit(std.testing.allocator);
     try stakes.ensureTotalCapacity(std.testing.allocator, 1);
 
@@ -1991,7 +1991,7 @@ test "check vote threshold above threshold no stake" {
     var tower = try createTestReplayTower(1, 0.67);
     defer tower.deinit(std.testing.allocator);
 
-    var stakes = VotedStakes{};
+    var stakes = VotedStakes.empty;
     defer stakes.deinit(std.testing.allocator);
 
     _ = try tower.recordBankVote(
@@ -2015,7 +2015,7 @@ test "check vote threshold lockouts not updated" {
     var tower = try createTestReplayTower(1, 0.67);
     defer tower.deinit(std.testing.allocator);
 
-    var stakes = VotedStakes{};
+    var stakes = VotedStakes.empty;
     defer stakes.deinit(std.testing.allocator);
     try stakes.ensureTotalCapacity(std.testing.allocator, 2);
 
@@ -2821,7 +2821,7 @@ test "is slot confirmed not enough stake failure" {
     var tower = try createTestReplayTower(1, 0.67);
     defer tower.deinit(std.testing.allocator);
 
-    var stakes = VotedStakes{};
+    var stakes = VotedStakes.empty;
     defer stakes.deinit(std.testing.allocator);
     try stakes.ensureTotalCapacity(std.testing.allocator, 1);
 
@@ -2835,7 +2835,7 @@ test "is slot confirmed unknown slot" {
     var tower = try createTestReplayTower(1, 0.67);
     defer tower.deinit(std.testing.allocator);
 
-    var stakes = VotedStakes{};
+    var stakes = VotedStakes.empty;
     defer stakes.deinit(std.testing.allocator);
 
     const result = isSlotConfirmed(&tower, 0, &stakes, 2);
@@ -2846,7 +2846,7 @@ test "is slot confirmed pass" {
     var tower = try createTestReplayTower(1, 0.67);
     defer tower.deinit(std.testing.allocator);
 
-    var stakes = VotedStakes{};
+    var stakes = VotedStakes.empty;
     defer stakes.deinit(std.testing.allocator);
     try stakes.ensureTotalCapacity(std.testing.allocator, 1);
 
--- a/src/consensus/replay_tower.zig
+++ b/src/consensus/replay_tower.zig
@@ -31,7 +31,7 @@ const ProgressMap = sig.consensus.ProgressMap;
 const Tower = sig.consensus.tower.Tower;
 const TowerError = sig.consensus.tower.TowerError;
 const VoteTransaction = sig.consensus.vote_transaction.VoteTransaction;
-const VotedStakes = sig.consensus.tower.VotedStakes;
+const VotedStakes = sig.consensus.progress_map.consensus.VotedStakes;
 
 const Stake = u64;
 
@@ -3353,7 +3353,7 @@ pub fn createTestReplayTower(
 fn isSlotConfirmed(
     replay_tower: *const ReplayTower,
     slot: Slot,
-    voted_stakes: *const sig.consensus.tower.VotedStakes,
+    voted_stakes: *const sig.consensus.progress_map.consensus.VotedStakes,
     total_stake: u64,
 ) bool {
     if (!builtin.is_test) {
--- a/src/consensus/replay_tower.zig
+++ b/src/consensus/replay_tower.zig
@@ -1391,7 +1391,7 @@ pub const ReplayTower = struct {
     }
 };
 
-const BlockhashStatus = union(enum) {
+pub const BlockhashStatus = union(enum) {
     /// No vote since restart
     uninitialized,
     /// Non voting validator
--- a/src/consensus/tower.zig
+++ b/src/consensus/tower.zig
@@ -10,7 +10,7 @@ const LatestValidatorVotesForFrozenBanks =
     sig.consensus.unimplemented.LatestValidatorVotesForFrozenBanks;
 const Lockout = sig.runtime.program.vote.state.Lockout;
 const LockoutIntervals = sig.consensus.unimplemented.LockoutIntervals;
-pub const VotedStakes = sig.consensus.progress_map.consensus.VotedStakes;
+const VotedStakes = sig.consensus.progress_map.consensus.VotedStakes;
 const Pubkey = sig.core.Pubkey;
 const Slot = sig.core.Slot;
 const SortedSet = sig.utils.collections.SortedSet;
--- a/src/consensus/tower.zig
+++ b/src/consensus/tower.zig
@@ -24,7 +24,7 @@ const StakeAndVoteAccountsMap = sig.core.stake.StakeAndVoteAccountsMap;
 const Logger = sig.trace.Logger;
 const ScopedLogger = sig.trace.ScopedLogger;
 
-const DUPLICATE_THRESHOLD = sig.consensus.unimplemented.DUPLICATE_THRESHOLD;
+const DUPLICATE_THRESHOLD = sig.replay.service.DUPLICATE_THRESHOLD;
 
 pub const MAX_LOCKOUT_HISTORY = sig.runtime.program.vote.state.MAX_LOCKOUT_HISTORY;
 
--- a/src/consensus/tower.zig
+++ b/src/consensus/tower.zig
@@ -508,7 +508,7 @@ pub fn isSlotDuplicateConfirmed(
 }
 
 test "is slot duplicate confirmed not enough stake failure" {
-    var stakes = VotedStakes{};
+    var stakes = VotedStakes.empty;
     defer stakes.deinit(std.testing.allocator);
     try stakes.ensureTotalCapacity(std.testing.allocator, 1);
 
@@ -523,7 +523,7 @@ test "is slot duplicate confirmed not enough stake failure" {
 }
 
 test "is slot duplicate confirmed unknown slot" {
-    var stakes = VotedStakes{};
+    var stakes = VotedStakes.empty;
     defer stakes.deinit(std.testing.allocator);
 
     const result = isSlotDuplicateConfirmed(
@@ -535,7 +535,7 @@ test "is slot duplicate confirmed unknown slot" {
 }
 
 test "is slot duplicate confirmed pass" {
-    var stakes = VotedStakes{};
+    var stakes = VotedStakes.empty;
     defer stakes.deinit(std.testing.allocator);
     try stakes.ensureTotalCapacity(std.testing.allocator, 1);
 
--- a/src/consensus/tower.zig
+++ b/src/consensus/tower.zig
@@ -10,6 +10,7 @@ const LatestValidatorVotesForFrozenBanks =
     sig.consensus.unimplemented.LatestValidatorVotesForFrozenBanks;
 const Lockout = sig.runtime.program.vote.state.Lockout;
 const LockoutIntervals = sig.consensus.unimplemented.LockoutIntervals;
+pub const VotedStakes = sig.consensus.progress_map.consensus.VotedStakes;
 const Pubkey = sig.core.Pubkey;
 const Slot = sig.core.Slot;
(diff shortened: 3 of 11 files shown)


src/consensus/tower.zig
```diff
@@ -10,6 +10,7 @@ const LatestValidatorVotesForFrozenBanks =
     sig.consensus.unimplemented.LatestValidatorVotesForFrozenBanks;
 const Lockout = sig.runtime.program.vote.state.Lockout;
 const LockoutIntervals = sig.consensus.unimplemented.LockoutIntervals;
+pub const VotedStakes = sig.consensus.progress_map.consensus.VotedStakes;
```
dnut: Can you make this as a pub const in only one place, either here or progress_map.consensus? It's confusing how it's being imported from both places. And would you mind keeping pub decls separate from the list of private imports?

src/core/bank.zig
```diff
@@ -113,7 +113,8 @@ pub const SlotState = struct {
     accounts_lt_hash: sig.sync.Mux(LtHash),
 
     pub fn isFrozen(self: *const SlotState) bool {
-        return self.hash.read().get() != null;
+        var mut_self = self.*;
+        return mut_self.hash.read().get().* != null;
```
dnut: Yes the original code was broken, but this is not the correct fix.

This is not safe because it copies the rwmux when it should be mutating the original to ensure the lock is used properly. This function definitely needs `*SlotTracker`.

potential problems you can run into with your approach:
- if another thread holds a write lock, you'll never see it get unlocked because your copy is a snapshot that is permanently locked. This function will block forever and never return.
- other threads will not see your read lock. another thread might mutate this data while you are reading it.

I also have a fix for this in my branch.

src/replay/service.zig
```diff
@@ -11,8 +11,16 @@ const BlockstoreReader = sig.ledger.BlockstoreReader;
...
+const processConsensus = sig.replay.consensus.processConsensus;
+
 /// Number of threads to use in replay's thread pool
 const NUM_THREADS = 4;
+// Give at least 4 leaders the chance to pack our vote
+const REFRESH_VOTE_BLOCKHEIGHT: usize = 16;
+
+pub const DUPLICATE_LIVENESS_THRESHOLD: f64 = 0.1;
+pub const SWITCH_FORK_THRESHOLD: f64 = 0.38;
+pub const DUPLICATE_THRESHOLD: f64 = 1.0 - SWITCH_FORK_THRESHOLD - DUPLICATE_LIVENESS_THRESHOLD;
```
dnut: Is this needed here? It's also in a file called unimplemented.zig, which is kind of confusing. Can you find one good place for this?

src/replay/consensus.zig
```diff
@@ -0,0 +1,1321 @@
...
+    slot_history: SlotHistory,
+    epoch_stakes: EpochStakeMap,
+    latest_validator_votes_for_frozen_banks: LatestValidatorVotesForFrozenBanks,
+};
+
+pub fn processConsensus(maybe_deps: ?ConsensusDependencies) !void {
+    const deps = if (maybe_deps) |deps|
+        deps
+    else
+        return error.Todo;
```
dnut, line 59: We'll need to get this actually running as part of replay very soon. Is there anything preventing us from assembling all the dependencies?

src/replay/consensus.zig
```diff
@@ -0,0 +1,1321 @@
...
+            deps.progress_map,
+            if (heaviest_slot_on_same_voted_fork) |h| h.slot else null,
+            &last_vote_refresh_time,
+        );
+    }
+
+    if (deps.replay_tower.tower.isRecent(heaviest_slot) and
+        heaviest_fork_failures.items.len != 0)
+    {
+        // Implemented the log
```
dnut: is this a TODO?

src/replay/consensus.zig
```diff
@@ -0,0 +1,1321 @@
...
+/// - `false` if any condition was not met or the refresh failed
+///
+/// Note: This is not simply resending the same vote, but creating a new distinct transaction that:
+/// - Votes for the same slot
+/// - Contains updated metadata (timestamp/blockhash)
+/// - Generates a new signature
+/// - Will be processed as a separate transaction by the network
+///
+/// Analogous to [maybe_refresh_last_vote](https://github.com/anza-xyz/agave/blob/ccdcdbe9b6ff7dbd583d2101fe57b7cc41a6f863/core/src/replay_stage.rs#L2606)
+fn maybeRefreshLastVote(
```
dnut, line 173: It's great that you already implemented this and wrote some extensive test coverage for it, because it's ultimately necessary for a full validator to execute this logic. 

For a non-voting validator, it seems that this function has no meaningful effects. Does that sound right to you?

src/replay/consensus.zig
```diff
@@ -0,0 +1,1321 @@
...
+            ),
+        };
+
+        try writer.setRoots(rooted_slots);
+    }
+
+    // Audit: The rest of the code maps to Self::handle_new_root in Agave.
+    // TODO
+    // - Prune program cache bank_forks.read().unwrap().prune_program_cache(new_root);
+    // - Set root on slot tracker: let removed_banks = bank_forks.write().unwrap().set_root(
```
dnut: Can you add `root: Slot` to SlotTracker and update it here?

src/replay/consensus.zig
```diff
@@ -0,0 +1,1321 @@
...
+            .db = blockstore_reader.db,
+            .logger = .noop,
+            .lowest_cleanup_slot = &lowest_cleanup_slot,
+            .max_root = &max_root,
+            .scan_and_fix_roots_metrics = try registry.initStruct(
+                sig.ledger.result_writer.ScanAndFixRootsMetrics,
+            ),
+        };
+
+        try writer.setRoots(rooted_slots);
```
dnut: setRoots updates the state behind the `max_root` pointer in LedgerResultWriter. This is supposed to be the max_root state that's shared with the rest of the validator. That value is initialized in cmd.zig. But instead you're just initializing a value here and discarding it immediately after mutating it. So this code is not having the intended effect.

The safest bet is probably to add a dependency on LedgerResultWriter, initialize that struct using the same max_root as BlockstoreReader in the validator function of cmd.zig, and pass it down to here.

Eventually, I'd like to improve the ledger structs. The shared state is confusing, and the function's dependencies are overkill. But I think for now we can live with a centrally constructed LedgerResultWriter.

src/replay/consensus.zig
```diff
@@ -0,0 +1,1321 @@
...
+    // TODO
+    // - Prune program cache bank_forks.read().unwrap().prune_program_cache(new_root);
+    // - Set root on slot tracker: let removed_banks = bank_forks.write().unwrap().set_root(
+    //        new_root,
+    //        snapshot_controller,
+    //        highest_super_majority_root,
+    //    )?
+    //
+    // Update the progress map.
+    // TODO Move to its own function?
```
dnut: Personally I think it's fine either way. Can you decide what you want to do here and either move the code or remove the TODO?

src/replay/consensus.zig
```diff
@@ -0,0 +1,1321 @@
...
+    {
+        var to_remove = try std.ArrayListUnmanaged(Slot).initCapacity(
+            allocator,
+            progress.map.count(),
+        );
+        defer to_remove.deinit(allocator);
+
+        var it = progress.map.iterator();
+        while (it.next()) |entry| {
+            // TODO should frozen state be taking into consideration.
```
dnut: If yes, then what? And do you know the answer? Maybe we should address this now.

src/replay/consensus.zig
```diff
@@ -0,0 +1,1321 @@
...
+    }
+
+    // Update forkchoice
+    try fork_choice.setTreeRoot(&.{
+        .slot = new_root,
+        .hash = root_hash,
+    });
+}
+
+fn resetFork(
```
dnut, line 416: Can you add a comment saying this is a placeholder for logic that is necessary for certain circumstances we're not currently handling, and mention those circumstances (voting and producing blocks?), and link to the analogous code in agave.

src/replay/consensus.zig
```diff
@@ -0,0 +1,1321 @@
...
+/// even if there are multiple competing versions of that slot.
+///
+/// This means the slot becomes a valid candidate for fork selection and
+/// can influence which chain the validator builds upon.
+///
+/// Note: 1. This is "duplicate confirmed", which is different from "regular" confirmation,
+/// where a slot is simply processed and frozen.
+/// Note: 2. The slot is skipped if it is already duplicate confirmed in the progress map's fork state
+///          or if the slot is not already frozen.
+fn towerDuplicateConfirmedForks(
```
dnut: When is this supposed to be called? I see no usages of it.

src/consensus/tower.zig
dadepo: [6908433](https://github.com/Syndica/sig/pull/762/commits/6908433315f0b36a5b94da47fb1ab2500f7c59f9)

src/replay/service.zig
dadepo: It should actually be here and dropped from the `unimplemented.zig`. The one in `unimplemented.zig` had a comment about it needing to be in replay. The `unimplemented.zig` is supposed to house things not implemented yet and should end up being deleted.

[617d989](https://github.com/Syndica/sig/pull/762/commits/617d9892452e9b4938dd596d21c2767737f70ced)

src/replay/consensus.zig
dadepo: > Is there anything preventing us from assembling all the dependencies?

Yeah. The part of the code that calls the `processConsensus` function still need to do some work to get all the consensus deps.

Currently this is 

```
pub const ConsensusDependencies = struct {
    allocator: Allocator,
    replay_tower: *ReplayTower,
    progress_map: *ProgressMap,
    slot_tracker: *SlotTracker,
    epoch_tracker: *EpochTracker,
    fork_choice: *ForkChoice,
    blockstore_reader: *BlockstoreReader,
    ancestors: std.AutoHashMapUnmanaged(u64, SortedSet(u64)),
    descendants: std.AutoArrayHashMapUnmanaged(u64, SortedSet(u64)),
    vote_account: Pubkey,
    slot_history: SlotHistory,
    epoch_stakes: EpochStakeMap,
    latest_validator_votes_for_frozen_banks: LatestValidatorVotesForFrozenBanks,
};
```

Some of this can be retrieved directly from `ReplayDependencies` with the current state where this PR branch from being

```
pub const ReplayDependencies = struct {
    /// Used for all allocations within the replay stage
    allocator: Allocator,
    logger: sig.trace.Logger,
    /// Tell replay when to exit
    exit: *std.atomic.Value(bool),
    /// Used in the EpochManager
    epoch_schedule: sig.core.EpochSchedule,
    /// Used to get the entries to validate them and execute the transactions
    blockstore_reader: *BlockstoreReader,
    /// Used to get the entries to validate them and execute the transactions
    accounts_db: *AccountsDB,
};
```


-while others might require having to extend `ReplayDependencies` or include logic that provides them - but given the implementations before hitting the `processConsensus` is currently being worked on, made it optional for now to get compilation passing.

src/replay/consensus.zig
dadepo: This was actually part of the functions that was initially implemented in this PR, but should actually be implemented in the edge cases PR. Removed. [5f40b6f](https://github.com/Syndica/sig/pull/762/commits/5f40b6facc67bd529dd285de89a11297015e2526)

src/replay/consensus.zig
dadepo: [c68558c](https://github.com/Syndica/sig/pull/762/commits/c68558cf95529d144b430985fca4329fbd05873c)

src/replay/consensus.zig
dadepo: That is indeed the case. These implementation were already added before the sync where it was agreed to limit things to non-voting flow. But since most of it was already done, and would eventually be needed, decided to leave it in, instead of removing it all together. Only hard code to `.non_voting` in places where needed.

src/replay/consensus.zig
dadepo: Yes. Made it explicit [bce9e46](https://github.com/Syndica/sig/pull/762/commits/bce9e46500f8cb351a6947f777a6329680445286)

src/replay/consensus.zig
dadepo: [527a4c2](https://github.com/Syndica/sig/pull/762/commits/527a4c21e1e6033fae23eabef8f4b34c5d74d79b)

src/replay/consensus.zig
dadepo: [bc35b76](https://github.com/Syndica/sig/pull/762/commits/bc35b7695d2b345737a717e5f8d99cc3afdbd410)

Fix build [2b12ada](https://github.com/Syndica/sig/pull/762/commits/2b12adac1e343066abd64a7920c2cee9867043ea)

src/replay/consensus.zig
dadepo: [bfdf54e](https://github.com/Syndica/sig/pull/762/commits/bfdf54e2082c6d6319419e5b31a7e7eedda64a23)

src/core/bank.zig
dadepo: [ae68626](https://github.com/Syndica/sig/pull/762/commits/ae68626eb7e59fdf1a15c2510a9b7f644b493a87)

src/replay/consensus.zig
dadepo: Cross checked and the Agave implementation, only keeps the slot in the progress map if it also found in the bank fork. Regardless of it is active or not.

[f3388ef](https://github.com/Syndica/sig/pull/762/commits/f3388ef590c6410198ea042433c6efe2b0f6b99d)

src/consensus/replay_tower.zig
```diff
@@ -31,7 +31,7 @@ const ProgressMap = sig.consensus.ProgressMap;
 const Tower = sig.consensus.tower.Tower;
 const TowerError = sig.consensus.tower.TowerError;
 const VoteTransaction = sig.consensus.vote_transaction.VoteTransaction;
-const VotedStakes = sig.consensus.tower.VotedStakes;
+const VotedStakes = sig.consensus.progress_map.consensus.VotedStakes;
```
InKryption, line 34: When I wrote this consensus namespace inside progress_map, it wasn't intended as a permanent home, I just didn't have a good understanding of where these things should live. Could you perhaps move them to a better home, so that we don't have this `sig.consensus.progress_map.consensus` silliness? I'm still not sure where it *should* live, but maybe you have a better idea than this?

src/consensus/replay_tower.zig
```diff
@@ -1574,7 +1574,7 @@ test "check vote threshold without votes" {
     var tower = try createTestReplayTower(1, 0.67);
     defer tower.deinit(std.testing.allocator);
 
-    var stakes = AutoHashMapUnmanaged(u64, u64){};
+    var stakes = VotedStakes{};
```
InKryption: Just to note on this and all of the other instances of this, default hashmap initialization is scheduled for deprecation, the `.empty` decl will become the canonical initialization procedure.

src/replay/consensus.zig
```diff
@@ -0,0 +1,1243 @@
...
+        // TODO Implemented the Self::log_heaviest_fork_failures
+    }
+
+    // Vote on the fork
+    if (maybe_voted_slot) |voted| {
+        const slot_tracker = deps.slot_tracker;
+        var found_slot = slot_tracker.slots.get(voted.slot) orelse
+            return error.MissingSlot;
+
+        const voted_hash = found_slot.state.hash.read().get().* orelse
```
InKryption: This should use `slot_tracker.get(voted.slot)` now, yeah?

On first glance I thought this was unintentionally copying the hashmap element, and the lock along with it, potentially resulting in a race, before I remembered that Drew's change made the elements into pointers.

Also it should be `const found_slot`, and maybe call it `found_slot_info` instead? The name did confuse me for a moment about what it was, thinking it was a slot value.

src/replay/consensus.zig
```diff
@@ -0,0 +1,1243 @@
...
+            deps.replay_tower,
+            deps.progress_map,
+            deps.fork_choice,
+        );
+    }
+
+    // Reset onto a fork
+    if (maybe_reset_slot) |reset_slot| {
+        // TODO implement
+        _ = &reset_slot;
```
InKryption: A word of caution, `_ = &foo;` prevents the compiler from seeing it as an unnecessary discard. Is this intentional here? Obviously I don't think there's any danger here that will be any problem here, just want to make sure this is intentional and not an oversight in some way.

src/replay/consensus.zig
```diff
@@ -0,0 +1,1243 @@
...
+    ledger_result_writer: *LedgerResultWriter,
+    slot_tracker: *SlotTracker,
+    progress: *ProgressMap,
+    fork_choice: *ForkChoice,
+    new_root: Slot,
+) !void {
+    // get the root bank before squash.
+    if (slot_tracker.slots.count() == 0) return error.EmptySlotTracker;
+    var root_tracker = slot_tracker.slots.get(new_root) orelse return error.MissingSlot;
+    const maybe_root_hash, var hash_lg = root_tracker.state.hash.readWithLock();
```
InKryption: Same comment as on the other instance of `slot_tracker.slots.get(slot)`.
Also, it should be `const root_tracker`, no reason for it to be var.

src/replay/consensus.zig
```diff
@@ -0,0 +1,1243 @@
...
+        progress.map.count(),
+    );
+    defer to_remove.deinit(allocator);
+
+    var it = progress.map.iterator();
+    while (it.next()) |entry| {
+        if (slot_tracker.slots.get(entry.key_ptr.*) == null) {
+            to_remove.appendAssumeCapacity(entry.key_ptr.*);
+        }
+    }
```
```suggestion
    for (progress.map.keys()) |progress_slot| {
        if (slot_tracker.get(progress_slot) == null) {
            to_remove.appendAssumeCapacity(progress_slot);
        }
    }
```

src/replay/consensus.zig
```diff
@@ -0,0 +1,1243 @@
...
+test "maybeRefreshLastVote - no heaviest slot on same fork" {
+    var prng = std.Random.DefaultPrng.init(91);
+    const random = prng.random();
+
+    const root = SlotAndHash{ .slot = 0, .hash = Hash.initRandom(random) };
+
+    var fixture = try TestFixture.init(testing.allocator, root);
+    defer fixture.deinit(testing.allocator);
+
+    var replay_tower = try createTestReplayTower(
```
InKryption, line 450: Missing a `defer replay_tower.deinit(std.testing.allocator)`? Both here and in other tests. Is there a particular reason for why the tests are structured in such a way? I've noticed this a few times.

src/replay/consensus.zig
```diff
@@ -0,0 +1,1243 @@
...
+        std.testing.allocator,
+        3,
+    );
+    defer expected_slots.deinit(std.testing.allocator);
+    var lockouts = [_]Lockout{
+        Lockout{ .slot = 3, .confirmation_count = 3 },
+        Lockout{ .slot = 4, .confirmation_count = 2 },
+        Lockout{ .slot = 5, .confirmation_count = 1 },
+    };
+    try expected_slots.appendSlice(std.testing.allocator, &lockouts);
```
InKryption: In reference to this, and the two other instances of this in a couple subsequent tests: there's no reason for `lockouts` to be a `var`, and it could very well be inlined.
You could even do something like this:
```suggestion
    var expected_slots: std.ArrayListUnmanaged(Lockout) =
        try .fromOwnedSlice(try std.testing.allocator.dupe(Lockout, &.{
            .{ .slot = 3, .confirmation_count = 3 },
            .{ .slot = 4, .confirmation_count = 2 },
            .{ .slot = 5, .confirmation_count = 1 },
        }));
    defer expected_slots.deinit(std.testing.allocator);
```

src/replay/trackers.zig
```diff
@@ -101,6 +101,25 @@ pub const SlotTracker = struct {
         }
         return frozen_slots;
     }
+
+    pub fn parents(
+        self: *const SlotTracker,
+        allocator: Allocator,
+        slot: Slot,
+    ) Allocator.Error![]const Slot {
+        var parents_list = std.ArrayListUnmanaged(Slot){};
```
InKryption: Same comment as on the hashmap case, this should use `.empty` initialization.

src/replay/trackers.zig
```diff
@@ -101,6 +101,25 @@ pub const SlotTracker = struct {
...
+        errdefer parents_list.deinit(allocator);
+
+        var current_slot = slot;
+        while (self.slots.get(current_slot)) |current| {
+            const parent_slot = current.constants.parent_slot;
+            try parents_list.append(allocator, parent_slot);
```
InKryption: Question: do you think it would be possible to `ensureTotalCapacity` on `parents_list` with some reasonable estimate? Ie `try parents_list.ensureTotalCapacity(allocator, self.slots.count() - (self.slots.getIndex(slot) orelse self.slots.count()))`? Would still need to `try append`, but may reduce the allocation overhead in common cases.

InKryption (changes requested):
A few small comments, some suggestions, and a couple questions. Overall the structure seems sound, very good work.

src/consensus/replay_tower.zig
dadepo: [4d1cc21](https://github.com/Syndica/sig/pull/762/commits/4d1cc217ec9f7f11fb52037568bdc370a3b2e008)

src/replay/consensus.zig
dadepo: [5bfc9cf](https://github.com/Syndica/sig/pull/762/commits/5bfc9cfcacddb3a85878774c40a6f56aebd5503d)

src/replay/consensus.zig
dadepo: It was an oversight [ba9586e](https://github.com/Syndica/sig/pull/762/commits/ba9586ee94ab899bce14495fea3ddfa53f17f8ca)

src/consensus/replay_tower.zig
dadepo: Looked into this and I see some other issues. For example `ThresholdDecision` is in this `sig.consensus.progress_map.consensus` namespace while also existing in `sig.consensus.tower.ThresholdDecision`

So I think it would require a proper cleanup that should ideally be done in a separate PR? 

This diff showed up because `sig.consensus.progress_map.consensus.VotedStakes` was being re-exposed in `sig.consensus.tower.VotedStakes` and Drew suggested to keep it in one place in the comment here https://github.com/Syndica/sig/pull/762#discussion_r2148306634

As per a better place? My first instincts is to put all these in a new `stakes.zig` file under `consensus`. If not that, then perhaps have them in `sig.consensus.replay_tower` instead.

InKryption: I think that makes sense, as long as there's a plan for how we can remove the redundancies. I'm happy to leave that work to a separate PR if you think it merits that - I'll leave that to you if that's alright, since it sounds like you have a clear vision for it (mark this as resolved after you read this).

src/replay/consensus.zig
dadepo: [0245b07](https://github.com/Syndica/sig/pull/762/commits/0245b075510c2b16402df21b442516033de01709)

src/replay/consensus.zig
dadepo: [501a90e](https://github.com/Syndica/sig/pull/762/commits/501a90ee0e049234ffa5402f973ae9df1f8edfe0)

src/replay/trackers.zig
dadepo: [726e5e5](https://github.com/Syndica/sig/pull/762/commits/726e5e5d8b28353bd631b5fe58ddcd736d59bc1a)

src/replay/trackers.zig
dadepo: I think that should work. Something like this? [a86bcf1](https://github.com/Syndica/sig/pull/762/commits/a86bcf1104a903eb000a0af6e65095f29f449c90)

src/replay/consensus.zig
dadepo: [b90deb9](https://github.com/Syndica/sig/pull/762/commits/b90deb9ee5133addc8e933914f9a9d6fe0ff84b8)

src/replay/consensus.zig
dadepo: Improved in [0ec9e81](https://github.com/Syndica/sig/pull/762/commits/0ec9e81405ea12771cb8ac8724a1f6257df6a080)

src/replay/trackers.zig
InKryption: Is that a hard invariant? As in, will the distance of a slot relative to the end always represent the upper bound of the number of parents?

dadepo: > As in, will the distance of a slot relative to the end always represent the upper bound of the number of parents?

I think so, but not totally sure. To be on the safe side, I switch to `append` instead of `appendAssumeCapacity`

dadepo: [38ed8d6](https://github.com/Syndica/sig/pull/762/commits/38ed8d6da8c95ac9b81502c8fb463f70914e4f64)

src/consensus/progress_map.zig
```diff
@@ -463,7 +463,7 @@ pub const ForkStats = struct {
     vote_threshold: VoteThreshold,
     is_locked_out: bool,
     voted_stakes: consensus.VotedStakes,
-    duplicate_confirmed_hash: Hash,
+    duplicate_confirmed_hash: ?Hash,
```
dnut: Why did you switch this to be optional?

src/replay/service.zig
```diff
@@ -13,14 +13,19 @@ const ProgressMap = sig.consensus.ProgressMap;
...
 const SlotTracker = replay.trackers.SlotTracker;
 const EpochTracker = replay.trackers.EpochTracker;
 
 /// Number of threads to use in replay's thread pool
 const NUM_THREADS = 4;
 
+const SWITCH_FORK_THRESHOLD: f64 = 0.38;
+const MAX_ENTRIES: u64 = 1024 * 1024; // 1 million slots is about 5 days
+const DUPLICATE_LIVENESS_THRESHOLD: f64 = 0.1;
+pub const DUPLICATE_THRESHOLD: f64 = 1.0 - SWITCH_FORK_THRESHOLD - DUPLICATE_LIVENESS_THRESHOLD;
```
dnut, line 27: My instinct would be for these to be in consensus.zig. Do you think they'll be used in this file?

src/replay/trackers.zig
```diff
@@ -101,6 +101,29 @@ pub const SlotTracker = struct {
...
+    pub fn parents(
+        self: *const SlotTracker,
+        allocator: Allocator,
+        slot: Slot,
+    ) Allocator.Error![]const Slot {
+        var parents_list = std.ArrayListUnmanaged(Slot).empty;
+        errdefer parents_list.deinit(allocator);
+
+        const max_possible_parent_count =
+            self.slots.count() - (self.slots.getIndex(slot) orelse self.slots.count());
```
dnut: This is clever, but I don't think it's safe to assume this is the "maximum possible." It assumes that the map is sorted, but if any items are removed with swapRemove then the map will not be sorted any more. Since the order can be random, you can only safely assume that `self.slots.count()` is the maximum until you walk through the map.

This would be a problem if you were using appendAssumeCapacity below, but you're not. You're just using append, which means it does not fully rely on this assumption. It will grow the list if necessary. So you don't have a bug here, but the name of this variable is misleading. It's not the max possible parent count. It's an estimated parent count, and it's only being treated as an estimate.

As an estimate, I think `self.slots.count()/2` would be equally accurate as this code, but it would have less chaotic behavior, resulting in a maximum of 2 allocations, whereas the current code could result in more.

I would lean towards using `self.slots.count()` as a more authoritative maximum, and then you could use appendAssumeCapacity.

src/replay/consensus.zig
```diff
@@ -0,0 +1,1213 @@
...
+    const heaviest_slot = deps.fork_choice.heaviestOverallSlot().slot;
+    const heaviest_slot_on_same_voted_fork =
+        (try deps.fork_choice.heaviestSlotOnSameVotedFork(deps.replay_tower)) orelse null;
+
+    const heaviest_epoch: Epoch = deps.epoch_tracker.schedule.getEpoch(heaviest_slot);
+
+    var last_vote_refresh_time: LastVoteRefreshTime = .{
+        .last_refresh_time = sig.time.Instant.now(),
+        .last_print_time = sig.time.Instant.now(),
+    };
```
dnut:
```suggestion
    const now = sig.time.Instant.now();
    var last_vote_refresh_time: LastVoteRefreshTime = .{
        .last_refresh_time = now,
        .last_print_time = now,
    };
```

This may have a performance benefit from eliminating a syscall. It's probably negligible though.

Also, the current code results in a tiny difference between these two numbers. I don't see any benefit for that since the difference has nothing to do with the meaning of the fields. I think the difference is more likely to cause confusion than anything else.

src/replay/consensus.zig
```diff
@@ -0,0 +1,1213 @@
...
+    epoch_tracker: *EpochTracker,
+    fork_choice: *ForkChoice,
+    blockstore_reader: *BlockstoreReader,
+    ledger_result_writer: *LedgerResultWriter,
+    ancestors: std.AutoHashMapUnmanaged(u64, SortedSet(u64)),
+    descendants: std.AutoArrayHashMapUnmanaged(u64, SortedSet(u64)),
+    vote_account: Pubkey,
+    slot_history: SlotHistory,
+    epoch_stakes: EpochStakeMap,
+    latest_validator_votes_for_frozen_banks: LatestValidatorVotesForFrozenBanks,
```
dnut: It seems that none of these things are mutated internally and they need to be maintained by the caller who owns the data, so my instinct is to pass them by const pointer.

```suggestion
    ancestors: *const std.AutoHashMapUnmanaged(u64, SortedSet(u64)),
    descendants: *const std.AutoArrayHashMapUnmanaged(u64, SortedSet(u64)),
    vote_account: Pubkey,
    slot_history: *const SlotHistory,
    epoch_stakes: *const EpochStakeMap,
    latest_validator_votes_for_frozen_banks: *const LatestValidatorVotesForFrozenBanks,
```

src/replay/consensus.zig
```diff
@@ -0,0 +1,1213 @@
...
+    for (progress.map.keys()) |progress_slot| {
+        if (slot_tracker.get(progress_slot) == null) {
+            to_remove.appendAssumeCapacity(progress_slot);
+        }
+    }
+
+    for (to_remove.items) |key| {
+        _ = progress.map.swapRemove(key);
+    }
+
```
dnut: Shouldn't this be updating the slot tracker here to remove these slots? In agave they call `BankForks::set_root` before updating the progress map. This removes all these slots from BankForks, which would be equivalent to us removing them from the slot tracker. Then you wouldn't need to iterate over to_remove for the progress map since it's already handled by the first loop. Also, when removing them from slot_tracker, that could produce the list of parents, so you don't need to do a separate call to `parents`

I wonder if we should just combine the progress map with the slot tracker.

src/replay/consensus.zig
dadepo: [0649012](https://github.com/Syndica/sig/pull/762/commits/064901215d6e70dab45e9fb2b4249cb9372f857f)

src/replay/consensus.zig
dadepo: [4c49988](https://github.com/Syndica/sig/pull/762/commits/4c49988a6ed2eb944684dc7dae9913dc0bbb2c51)

src/replay/trackers.zig
dadepo: [d46a10d](https://github.com/Syndica/sig/pull/762/commits/d46a10dfe3794e9e0026418d05fe9ec3b5137cbc) [87954f9](https://github.com/Syndica/sig/pull/762/commits/87954f9b61e9798f2a43e1152ae75105bff87840)

src/consensus/progress_map.zig
dadepo: It is optional in Agave https://github.com/anza-xyz/agave/blob/72c2cc48800ec5324bd2af2e70fc4abcbd792071/core/src/consensus/progress_map.rs#L191

And I must have updated it as part of initial implementations that needed it to be, that we later decided to move to follow up PRs, only did not revert this part then.

[f586f6c](https://github.com/Syndica/sig/pull/762/commits/f586f6ce4c23a0bb1b3c45d9c7991d773f7bb351)

src/replay/service.zig
dadepo: Moved them to `replay_tower` in another PR (draft) that makes use of the stake cache in the collectVoteLockout method https://github.com/Syndica/sig/blob/caed56f213f7d442e0fa717f6636ca8a5db3165b/src/consensus/replay_tower.zig#L46-L52 given this is where most of them seems to be used.

I think after all these pieces are assembled, these imports can be improved.

src/replay/consensus.zig
dadepo: Part of the `BankForks::set_root` operation is noted as TODO here https://github.com/Syndica/sig/blob/38ed8d6da8c95ac9b81502c8fb463f70914e4f64/src/replay/consensus.zig#L385

> I wonder if we should just combine the progress map with the slot tracker.

Not the first time the idea would be broached...and maybe the combination can be renamed to `replay_state_tracker`? As I think there is still some uncertainty to the `slot tracker` name?

But maybe these are optimization that can be made post having it run on test net?

dnut: Why is this a TODO? It should be easy to implement. Iterate over each slot and remove it if it's not in the root slot's ancestors.

Also regarding the highlighted code for this thread. Why are you using an array list for this? Why not just remove the items directly?

dadepo: > Why is this a TODO?

I can't remember... Probably got mistaken annotated as part of the things to be implemented later. 

I'll update this PR to include it

dadepo: > I can't remember... Probably got mistaken annotated as part of the things to be implemented later.

Looking at it again, it was mainly because the slot tracker currently does not have a way to list descendants and less so, because there is no implementation of `block_commitment_cache` from where `highest_super_majority_root` can be retrieved. See https://github.com/anza-xyz/agave/blob/1c776667cb7f9e4f20891c08c256e82136829465/runtime/src/bank_forks.rs#L591

> Also regarding the highlighted code for this thread. Why are you using an array list for this? Why not just remove the items directly?

[5de6505](https://github.com/Syndica/sig/pull/762/commits/5de6505489e03de7f081ebf82254f8b46d400032)

dnut: We need a way to clean up the slot tracker so it doesn't grow unbounded. But I don't think we actually need to keep old rooted slots in there. Why don't you just delete everything older than the root? That would be the simplest approach. If we need that information later, we can add it back.

------------------------------------------

If there's a real need to keep a history of metadata for rooted slots, there is another solution you could implement easily right now using Ancestors (instead of descendants) and hardcoding a maximum number of rooted slots to track (instead of using the block commitment cache). I added ancestors to my branch, it was only a few lines of code:

https://github.com/Syndica/sig/blob/747ec0b0e93bc581c0fe4d781b7bcd0923141e0a/src/replay/service.zig#L221-L223

https://github.com/Syndica/sig/blob/747ec0b0e93bc581c0fe4d781b7bcd0923141e0a/src/core/bank.zig#L85-L88

src/replay/consensus.zig
```diff
@@ -0,0 +1,1212 @@
...
+    last_reset_hash: Hash,
+    last_blockhash: Hash,
+    last_reset_bank_descendants: std.ArrayList(Slot),
+) !void {
+    _ = &progress;
+    _ = &blockstore;
+    _ = &reset_slot;
+    _ = &last_reset_hash;
+    _ = &last_blockhash;
+    _ = &last_reset_bank_descendants;
```
InKryption: Why do you default to `_ = &foo;` in many places in this code? Usually this style of unused ref is reserved for comptime blocks to force evaluation of a function, so it's jarring to see it used throughout here, and has the drawback that when you do start using the parameter, `_ = &foo;` won't be flagged as a redundant unused declaration.

dadepo: I think I had it this way, before you pointed the drawback in another PR.

[0a3fb8c](https://github.com/Syndica/sig/pull/762/commits/0a3fb8c8b527efd745026427e1f03d279fa2bfdb)

src/replay/consensus.zig
dadepo: > Why don't you just delete everything older than the root? 

That is mainly due to the fact that in Agave's, the pruning logic does not delete everything older the root. See https://github.com/anza-xyz/agave/blob/1c776667cb7f9e4f20891c08c256e82136829465/runtime/src/bank_forks.rs#L541-L590

It only removes things not in the rooted paths. I am not sure if deleting everything older than the root won't lead to a different behaviour.

---

I'll do that with a note to change this later to only prune things not in the rooted path.

InKryption (approved):

src/replay/consensus.zig
dadepo: [a3fa6a7](https://github.com/Syndica/sig/pull/762/commits/a3fa6a70a17c7921ee7cffd4aa37cdd2c7aaebae)

Tests updates/fixes [c5349a4](https://github.com/Syndica/sig/pull/762/commits/c5349a4df48aba6e7d2beb37563d8a3e87058293)

dadepo: > Also regarding the highlighted code for this thread. Why are you using an array list for this? Why not just remove the items directly?
> > [5de6505](https://github.com/Syndica/sig/pull/762/commits/5de6505489e03de7f081ebf82254f8b46d400032)

Revisiting [5de6505](https://github.com/Syndica/sig/pull/762/commits/5de6505489e03de7f081ebf82254f8b46d400032)

I am not 100% sure if it is safe to be removing from the same map being iterated.

dnut: Yeah I knew how agave works, but my point was that I don't think we need it right now. If we do need it, then it can be done using ancestors, which provides the same information as descendants but it's easier to initialize.

> I am not 100% sure if it is safe to be removing from the same map being iterated.

It can be if you are careful. This should work:

```zig
var keys = map.keys();
var index: usize = 0;
while (index < keys.len) {
    if (keys[index] < root) {
        map.swapRemoveAt(index);
        keys = map.keys();
    } else {
        index += 1;
    }
}
```

dnut: I went ahead and implemented it so we can get this merged asap. 031e0b8372db3c3aff24ddefcb6568b6dec19c78

dnut (approved):

PR #765, opened 2025-05-28
fix(svm): bpf-loader partial conformance 

kprotty:
Fix conformance for `bpf-loader-v{1,2,3}` instruction fixtures, `bpf-loader-v{1,2}-programs` fixtures, and `compute-budget` fixtures.

### DIFF
--- a/src/runtime/program/bpf_loader/execute.zig
+++ b/src/runtime/program/bpf_loader/execute.zig
@@ -2871,7 +2871,7 @@ test "executeV3ExtendProgram" {
 
     // Test extend_program disabled when ENABLE_EXTEND_PROGRAM_CHECKED is enabled
     {
-        var tc = try sig.runtime.testing.createTransactionContext(
+        var tx = try sig.runtime.testing.createTransactionContext(
             allocator,
             prng.random(),
             .{
@@ -2894,10 +2894,14 @@ test "executeV3ExtendProgram" {
                 },
             },
         );
-        defer sig.runtime.testing.deinitTransactionContext(allocator, tc);
+        const tc = &tx[1];
+        defer {
+            sig.runtime.testing.deinitTransactionContext(allocator, tc.*);
+            tx[0].deinit(allocator);
+        }
 
         const instruction_info = try sig.runtime.testing.createInstructionInfo(
-            &tc,
+            tc,
             bpf_loader_program.v3.ID,
             bpf_loader_program.v3.Instruction{
                 .extend_program = .{ .additional_bytes = 0 },
@@ -2908,14 +2912,14 @@ test "executeV3ExtendProgram" {
 
         try std.testing.expectError(
             InstructionError.InvalidInstructionData,
-            sig.runtime.executor.executeInstruction(allocator, &tc, instruction_info),
+            sig.runtime.executor.executeInstruction(allocator, tc, instruction_info),
         );
         try std.testing.expectEqual(tc.compute_meter, 0);
     }
 
     // Test extend_program_checked disabled when ENABLE_EXTEND_PROGRAM_CHECKED is not present.
     {
-        var tc = try sig.runtime.testing.createTransactionContext(
+        var tx = try sig.runtime.testing.createTransactionContext(
             allocator,
             prng.random(),
             .{
@@ -2932,10 +2936,14 @@ test "executeV3ExtendProgram" {
                 },
             },
         );
-        defer sig.runtime.testing.deinitTransactionContext(allocator, tc);
+        const tc = &tx[1];
+        defer {
+            sig.runtime.testing.deinitTransactionContext(allocator, tc.*);
+            tx[0].deinit(allocator);
+        }
 
         const instruction_info = try sig.runtime.testing.createInstructionInfo(
-            &tc,
+            tc,
             bpf_loader_program.v3.ID,
             bpf_loader_program.v3.Instruction{
                 .extend_program_checked = .{ .additional_bytes = 0 },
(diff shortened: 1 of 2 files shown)


src/runtime/instruction_info.zig
```diff
@@ -121,6 +121,23 @@ pub const InstructionInfo = struct {
         return data;
     }
 
+    /// [agave] https://github.com/anza-xyz/solana-sdk/blob/1276772ee61fbd1f8a60cfec7cd553aa4f6a55f3/bincode/src/lib.rs#L9
+    pub fn limitedDeserializeInstruction(
```
yewman, line 126: Was this implemented due to hitting a conformance issue with `deserializeInstruction`. If so, we should remove it and replace all usages with this implementation

src/runtime/program/bpf_loader/execute.zig
```diff
@@ -1427,9 +1512,12 @@ pub fn deployProgram(
     tc: *TransactionContext,
     program_id: Pubkey,
     owner_id: Pubkey,
+    account_size: u64,
     data: []const u8,
     slot: u64,
 ) (error{OutOfMemory} || InstructionError)!void {
+    _ = account_size;
```
yewman: Why add the arg only to discard it?

kprotty: would be passed into `load_program_from_bytes`, which passes & stores it into a ProgramCacheEntry, then mostly used to verify account data size in `run_program_checks`. I _think_ it could be omitted?

src/runtime/instruction_info.zig
kprotty: Not conformance related, but to match a nicety in avoiding heap alloc. Should we should switch all `deserializeInstruction` use-cases towards it? Or keep the old way as it's fine for now.

yewman: I would put a comment on this function to note that its just an optimisation to avoid a heap alloc, and otherwise identical to `deserializeInsturction`. Then create an issue to transition all invocations of `deserializeInstruction` to use the new implementation. Making the transition in a seperate PR would be easier for review :)

src/runtime/program/bpf_loader/execute.zig
yewman: Should we remove it for now and then add it later if we find its required then?

kprotty: https://github.com/Syndica/sig/pull/765/commits/185812497a5749d7878e38c85ad5e99462a5dfb0

src/runtime/instruction_info.zig
kprotty: https://github.com/Syndica/sig/pull/765/commits/185812497a5749d7878e38c85ad5e99462a5dfb0

yewman (approved):
lgtm

yewman (approved):

data/test-elfs/direct_mapping.zig
```diff
@@ -29,7 +29,7 @@ const log: *align(1) const fn (
 ) void = @ptrFromInt(0x207559bd);
 
 // hashed as `sol_panic_`
-const panic: *const fn (
+const panic: *align(1) const fn (
```
Sobeston, line 32: An align(1) function pointer does feel a bit weird? Wondering what happened with these

src/runtime/program/bpf/execute.zig
```diff
@@ -83,6 +83,7 @@ pub fn execute(
                 .aligned_memory_mapping = !direct_mapping,
                 .minimum_version = min_sbpf_version,
                 .maximum_version = max_sbpf_version,
+                .optimize_rodata = false,
```
Sobeston, line 86: What's this bit for?

src/runtime/program/bpf_loader/execute.zig
```diff
@@ -76,25 +81,55 @@ pub fn execute(
...
+    //     .set_program_length => |args| executeV4SetProgramLength(
+    //         allocator,
+    //         ic,
+    //         args.new_size,
+    //     ),
+    //     .deploy => executeV4Deploy(allocator, ic),
+    //     .retract => executeV4Retract(allocator, ic),
+    //     .transfer_authority => executeV4TransferAuthority(allocator, ic),
+    //     .finalize => executeV4Finalize(allocator, ic),
+    // };
```
Sobeston, line 120: is this supposed to be commented out?

Sobeston (commented):
Got a few questions, but overall looking good!

data/test-elfs/direct_mapping.zig
kprotty: Seems odd, but it's how BPF-compiled (zig/rust/c) programs link to the syscall definitions. The magic number are a [hash of the syscall name](https://docs.rs/solana-define-syscall/latest/src/solana_define_syscall/lib.rs.html#14-17) resolved by the [interpreter](https://github.com/Syndica/sig/blob/main/src/vm/executable.zig#L360C29-L360C42)/jit

src/runtime/program/bpf/execute.zig
kprotty: Just to match the [v1 environment](https://github.com/anza-xyz/agave/blob/master/programs/bpf_loader/src/syscalls/mod.rs#L366) config.

src/runtime/program/bpf_loader/execute.zig
kprotty: Yea - when porting changes from `2.2.13`, I implemented the types & dispatch needed for bpf-loader-v4 only to realize it's unused atm. There's a TODO at the top of the function left for it.

data/test-elfs/direct_mapping.zig
Sobeston: Weird!

Sobeston (approved):

PR #766, opened 2025-05-28
refactor(build): make install step depend on executable compilation

InKryption:
This way `-Dno-bin -Dno-run` works even when using the install step,
whether explicitly or by default - ie, `zig build -Dno-bin -Dno-run`
no longer ignores the existence of compile errors.

I missed this in #760.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -313,14 +313,16 @@ fn addInstallAndRun(
     exe: *Build.Step.Compile,
     config: Config,
 ) void {
+    const install_step = b.getInstallStep();
     var send_step: ?*Build.Step = null;
 
     step.dependOn(&exe.step);
+    install_step.dependOn(&exe.step);
 
     if (config.install or (config.ssh_host != null and config.run)) {
         const install = b.addInstallArtifact(exe, .{});
         step.dependOn(&install.step);
-        b.getInstallStep().dependOn(&install.step);
+        install_step.dependOn(&install.step);
 
         if (config.ssh_host) |host| {
             const install_dir = if (config.ssh_install_dir[0] == '/')


InKryption:
* **#766** <a href="https://app.graphite.dev/github/pr/Syndica/sig/766?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Syndica/sig/766?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Sobeston (approved):
:ok_hand:

build.zig
```diff
@@ -313,14 +313,16 @@ fn addInstallAndRun(
     exe: *Build.Step.Compile,
     config: Config,
 ) void {
+    const install_step = b.getInstallStep();
     var send_step: ?*Build.Step = null;
 
     step.dependOn(&exe.step);
+    install_step.dependOn(&exe.step);
```
dnut, line 320: What problem does this solve?

The root install step already depends on this artifact's install step if it exists, which depends on the executable. If you're not installing this artifact, then why should the root install step care about the executable?

dnut: I just noticed your comment which answers the question "What problem does this solve"

> This way -Dno-bin -Dno-run works even when using the install step,
> whether explicitly or by default - ie, zig build -Dno-bin -Dno-run
> no longer ignores the existence of compile errors.

dnut: My concern is that there will be unnecessary building going on. If all you want to do is build one executable, you shouldn't be forced into building all of them.

InKryption: The install step inherently implies "building" all things that are to be installed. This change just makes it so `install -Dno-bin -Dno-run` has behaviour consistent with `sig test [...] -Dno-bin -Dno-run`, ie running semantic analysis for all specified artifacts, without actually installing any of them, or running any of the binaries (thus avoiding wasted codegen).

Rexicon226: To be clear there is no "root" step. The install step is the one ran when either 
1. No step is provided (since it's the default).
2. `install` is specified.

> If all you want to do is build one executable, you shouldn't be forced into building all of them.

We have individual steps for each executable, `sig`, `vm`, etc. Those are not related to the `install` step in any way.

When `install` is invoked, it *already* has a transient dependency on the `exe` being compiled through the `InstallArtifact` Step.

dnut: OK I think understand what's going on now. `addInstallArtifact` automatically creates this dependency.

The implicit "install everything" step feels a little too magical. I think I would prefer to be required to define it explicitly.

Anyway, I'm happy to finally merge this pr.

Rexicon226: There is no implicit "install everything" step, the `install` step acts as every other step in the build system, it isn't unique. There just exists a step pre-defined for you called `install` so that there can be a standard way of "building" a project.

PR #771, opened 2025-06-04
chore(build, log): make some fixes and cleanup

Rexicon226:
- Moved away from using the `comptime {` block to reference tests, since this ends up pulling in all of the tests when compiling other stuffs and increases the compile times needlessly.
- Added a `check` step so we can use incremental compilation with ZLS smoother
- Updates some more of the package hashes to make sure they're using the updated package hash themselves. ZLS will try to build the dependency as the root, which will fail on 0.14 unless it uses the newer package hash. Also improves the incremental experience.
- Removed an extremely silly function that should never exist!

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -233,6 +233,8 @@ pub fn build(b: *Build) void {
             .error_tracing = config.error_tracing,
         }),
         .sanitize_thread = config.enable_tsan,
+        .filters = config.filters orelse &.{},
+        .use_llvm = config.use_llvm,
     });
     unit_tests_exe.linkLibC();
     switch (config.blockstore_db) {
--- a/src/runtime/log_collector.zig
+++ b/src/runtime/log_collector.zig
@@ -28,8 +28,9 @@ pub const LogCollector = struct {
     }
 
     pub fn deinit(self: LogCollector, allocator: std.mem.Allocator) void {
-        for (self.messages.items) |message| allocator.free(message);
-        allocator.free(self.messages.allocatedSlice());
+        var copy = self;
+        for (copy.messages.items) |message| allocator.free(message);
+        copy.messages.deinit(allocator);
     }
 
     /// [agave] https://github.com/anza-xyz/agave/blob/faea52f338df8521864ab7ce97b120b2abb5ce13/program-runtime/src/log_collector.rs#L43


yewman (dismissed):
lgtm

build.zig
```diff
@@ -303,6 +304,11 @@ pub fn build(b: *Build) void {
         .install_subdir = "docs",
     });
     docs_step.dependOn(&install_sig_docs.step);
+
+    const check_step = b.step("check", "Checks that sig compiles (useful for development)");
+    const check_test = b.addTest(.{ .root_module = unit_test_mod });
+    check_test.generated_bin = null;
+    check_step.dependOn(&check_test.step);
```
dnut: incremental compilation works with `zig build` now, so we don't need this

Rexicon226: I'm confused by what you mean.
- What do you mean by no longer needing?
- What do you mean by incremental working with `zig build`? It very much does not work; Zig is not able to incrementally link.

dnut: Why does your check step only include the unit tests?

Rexicon226: Because that covers the majority of the source code, and I've hit some compiler bugs when experimenting with incrementally compiling some of the other executables (mainly the `sig` one, but I've had issues with other ones).

This is purely a QOL improvement; if in the future somebody is working on another component, they can easily add it here.

dnut: What compiler bugs?

The problem I observed in the past was that compilation completely failed due to `usingnamespace`. You simply couldn't run `zig build -fincremental` at all without error. It failed with most build targets, but not `test` because it had no dependencies on code with `usingnamespace`. That's not a problem any more since I eliminated all usages of `usingnamespace` from our dependencies.

I use ZLS with just this, no changes to build.zig, and it works great:
```
    "zig.zls.buildOnSaveArgs": [
        "-fincremental",
        "-Dno-bin",
        "-Dno-run",
        "-freference-trace"
    ],
```
I get instant feedback from ZLS about problems with *any* build target. I think that's the best default.

My understanding is that adding a `check` target will make ZLS use that instead of just `zig build`, is that correct?

Rexicon226: Ah, I wasn't aware that you had removed the `usingnamespace`s and you didn't mention that you were providing extra flags to ZLS. In that case sure, the `check` step isn't necessary anymore. I can remove that part of the PR.

Rexicon226: 5dc4274fbfe326534b9e7b2773cdf66fedb9da83

dnut (approved):

dnut (approved):

PR #778, opened 2025-06-10
feat(replay): Edge cases

### LINKED ISSUE #777: edge case bookkeeping
Here are the functions from the replay stage that we need to support:

* process_ancestor_hashes_duplicate_slots
* process_duplicate_confirmed_slots
* process_duplicate_slots
* process_gossip_verified_vote_hashes
* process_popular_pruned_forks
* tower_duplicate_confirmed_forks + mark_slots_duplicate_confirmed

In agave, this code is a huge mess. They all call into a single function with an enum specifying what logic they want to run. Then that single function executes whatever logic is specified by the enum. In reality this should probably just be entirely separate functions where you specify the behavior by calling the desired function instead of by passing a specific enum.

The singular function also has a ridiculous number of dependencies listed, which are probably not needed in the majority of cases. Breaking this up into separate functions will give us clarity about dependencies are actually needed.

You'll need to carefully map out every single scenario before starting implementation. It is complex because single enums of one type are mapped to multiple enums of another type. Be sure that it actually makes sense to eliminate all the enums. My instinct is that it does make sense, but I can't be sure without enumerating every single case and the logic, and verifying they can be isolated.

InKryption:
Closes #777.

### DIFF
--- a/src/consensus/fork_choice.zig
+++ b/src/consensus/fork_choice.zig
@@ -416,7 +416,7 @@ pub const ForkChoice = struct {
     }
 
     pub fn isDuplicateConfirmed(
-        self: *ForkChoice,
+        self: ForkChoice,
         slot_hash_key: *const SlotAndHash,
     ) ?bool {
         if (self.fork_infos.get(slot_hash_key.*)) |fork_info| {
--- a/src/replay/service.zig
+++ b/src/replay/service.zig
@@ -91,6 +91,7 @@ fn advanceReplay(state: *ReplayState) !void {
     // TODO: maybe_start_leader
 }
 
+// -- handleEdgeCases START -- //
 fn handleEdgeCases() void {
     // TODO: process_ancestor_hashes_duplicate_slots
 
@@ -101,8 +102,8 @@ fn handleEdgeCases() void {
     // TODO: process_popular_pruned_forks
 
     // TODO: process_duplicate_slots
-
 }
+// -- handleEdgeCases END -- //
 
 fn processConsensus() void {
     // TODO: for each slot:
--- a/src/replay/service.zig
+++ b/src/replay/service.zig
@@ -93,7 +93,7 @@ fn advanceReplay(state: *ReplayState) !void {
 
 // -- handleEdgeCases START -- //
 fn handleEdgeCases() void {
-    // TODO: process_ancestor_hashes_duplicate_slots
+    _ = &process_ancestor_hashes_duplicate_slots; // TODO:
 
     // TODO: process_duplicate_confirmed_slots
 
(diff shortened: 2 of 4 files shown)


InKryption:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Syndica/sig/778?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#778** <a href="https://app.graphite.dev/github/pr/Syndica/sig/778?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Syndica/sig/778?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#785** <a href="https://app.graphite.dev/github/pr/Syndica/sig/785?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/replay/service.zig
```diff
@@ -91,19 +96,1269 @@ fn advanceReplay(state: *ReplayState) !void {
...
+        return .{
+            .epoch_slots_frozen_hash = epoch_slots_frozen_hash,
+            .duplicate_confirmed_hash = duplicate_confirmed_hash,
+            .bank_status = bank_status,
+            .is_popular_pruned = is_popular_pruned,
+        };
+    }
+};
+
+fn processAncestorHashesDuplicateSlots(
```
dadepo: I know this is generally not consistently done, but can you help add a link that points to the analogous function in Agave?

Also only looking at the input params, I noticed a couple are missing when compared to Agave:

- `duplicate_slots_tracker: &mut DuplicateSlotsTracker,`
- `ancestor_hashes_replay_update_sender: &AncestorHashesReplayUpdateSender`
- `purge_repair_slot_counter: &mut PurgeRepairSlotCounter`

Is this omission intentional?

src/replay/service.zig
```diff
@@ -91,19 +96,1269 @@ fn advanceReplay(state: *ReplayState) !void {
...
+    allocator: std.mem.Allocator,
+    logger: sig.trace.Logger,
+    pubkey: sig.core.Pubkey,
+    ancestor_duplicate_slots_receiver: *sig.sync.Channel(AncestorDuplicateSlotToRepair),
+    duplicate_confirmed_slots: *const DuplicateConfirmedSlots,
+    epoch_slots_frozen_slots: *EpochSlotsFrozenSlots,
+    progress: *const sig.consensus.ProgressMap,
+    fork_choice: *sig.consensus.HeaviestSubtreeForkChoice,
+    bank_forks_rwmux: *sig.sync.RwMux(sig.replay.trackers.SlotTracker),
```
dadepo: nit:
```suggestion
    slot_tracker_rwmux: *sig.sync.RwMux(sig.replay.trackers.SlotTracker),
```

src/replay/service.zig
```diff
@@ -91,19 +96,1269 @@ fn advanceReplay(state: *ReplayState) !void {
...
+    is_local_replay_duplicate_confirmed: bool,
+) ?sig.core.Hash {
+    // the following code is simply ported in order to match the logic of the equivalent match statement
+    // TODO: maybe simplify all of this if possible?
+
+    const maybe_local_duplicate_confirmed_hash = if (is_local_replay_duplicate_confirmed) blk: {
+        // If local replay has duplicate_confirmed this slot, this slot must have
+        // descendants with votes for this slot, hence this slot must be
+        // frozen.
+        break :blk maybe_bank_frozen_hash.?;
```
dadepo: Should the possibility for panic be removed here by using `.?` and returning either an error or null instead?

Although seeing the comment, it looks like that should never be the case?

src/replay/service.zig
```diff
@@ -91,19 +96,1269 @@ fn advanceReplay(state: *ReplayState) !void {
...
+            }
+        }
+    }
+
+    fn epochSlotsFrozen(
+        allocator: std.mem.Allocator,
+        logger: sig.trace.Logger,
+        slot: sig.core.Slot,
+        root: sig.core.Slot,
+        epoch_slots_frozen_slots: *EpochSlotsFrozenSlots,
```
dadepo: Nit: Maybe have `epoch_slots_frozen_slots` and  `epoch_slots_frozen_state` next to each other.

src/replay/service.zig
```diff
@@ -91,19 +96,1269 @@ fn advanceReplay(state: *ReplayState) !void {
...
+                // The cluster sample found the troublesome slot which caused this fork to be pruned
+                logger.warn().logf(
+                    "EpochSlots sample returned slot {} with hash {}, " ++
+                        "but we have pruned it due to incorrect ancestry",
+                    .{ slot, epoch_slots_frozen_hash },
+                );
+            },
+        }
+
+        try resulting_change.repairDuplicateConfirmedVersion(
```
dadepo: nit: Any reason not to have it as `duplicate_slots_to_repair.put(allocator, slot, epoch_slots_frozen_hash)`

I.e remove the need for the static function?

src/replay/service.zig
InKryption: Would rather not start renaming these yet; I'm not convinced on `SlotTracker` as the final name for the lightweight bankforks, given that an extremely similar name is currently in use in vote tracker "SlotVoteTracker", and is often referred to as just "SlotTracker".

src/replay/service.zig
InKryption: Yeah these omissions are all intentional, as these parameters are not necessary for this codepath, they were only there in agave because `apply_state_changes` required the sum of all the parameters for all possible codepaths.

InKryption: 115f85961231a55bc485639d4026ae631c828e07

src/replay/service.zig
InKryption: Yeah, shouldn't ever be the case, check out both of the callers: `getClusterConfirmedHashFromState` & `getDuplicateConfirmedHashFromState`.

InKryption: Though maybe it would be better to represent this dependence in the parameter type(s) themselves.

src/replay/service.zig
InKryption: 73c5ca430f4afbced10f1b83e56427b71e8944b5

src/replay/service.zig
InKryption: Mainly the fact that it helps to have all the possible state changes centralized in one place, gives a name to the action that is designated to occur in multiple places, also makes a good place to have the reference to the equivalent agave code.

src/consensus/fork_choice.zig
```diff
@@ -323,6 +336,10 @@ pub const ForkChoice = struct {
         return null;
     }
 
+    pub fn bestOverallSlot(self: *const ForkChoice) ?SlotAndHash {
```
dadepo: This is already implemented as https://github.com/Syndica/sig/blob/cbfbdc8f0ceef85928f246e72194f3108d8dc3e7/src/consensus/fork_choice.zig#L298

I remember during the review that added these methods, it was suggested to change the name. But maybe the original name should have been kept to avoid oversight like this...

src/consensus/fork_choice.zig
```diff
@@ -313,6 +318,14 @@ pub const ForkChoice = struct {
         return null;
     }
 
+    pub fn bestSlot(
```
dadepo: This is already implemented as https://github.com/Syndica/sig/blob/cbfbdc8f0ceef85928f246e72194f3108d8dc3e7/src/consensus/fork_choice.zig#L305

I remember during the review that added these methods, it was suggested to change the name. But maybe the original name should have been kept to avoid oversight like this...

dadepo (commented):
Still need to review `src/replay/service.zig` again, but took a quick look just now and thought to point out the re-implementation of existing methods on forkchoice.

src/consensus/fork_choice.zig
InKryption: The renames did make it harder to figure out what was equivalent where, though that could also just be down to me not thoroughly reading the comments.

Maybe in cases like these we could use an alias for functions, ie here: `pub const bestOverallSlot = heaviestOverallSlot;`, or `pub const bestOverallSlot = @compileError("Use heaviestOverallSlot instead");`

InKryption: 523c46d2167767976942308aff0823ed8575668c

src/consensus/fork_choice.zig
InKryption: 523c46d2167767976942308aff0823ed8575668c

dadepo (approved):

src/replay/service.zig
InKryption: 3744cd0f4789247b17f656293265dc87e3cf91e2

src/replay/service.zig
```diff
@@ -206,19 +211,2562 @@ fn trackNewSlots(
     }
 }
 
+// -- handleEdgeCases START -- //
```
dnut: Can you move this code to a separate file? service.zig is the top level entrypoint for replay and represents the category of code used for high level orchestration. The code you've implemented is a subset of the consensus logic. Dadepo has introduced a consensus.zig for the consensus code. This code would make a lot more sense inside a file called consensus.zig than it would inside service.zig. But actually I think this code is distinct and complex enough to warrant its own file.

src/replay/service.zig
dnut: I'd rather have a name that accurately reflects the actual type than a name that is intentionally wrong. We can rename the struct if we think of something better, but until then, let's have some consistency. Often things like this never get changed and then you're left with confusing inconsistencies littered around the code.

src/replay/service.zig
```diff
@@ -206,19 +211,2562 @@ fn trackNewSlots(
...
+        return switch (param) {
+            .is_dead => .dead,
+            .hash => |maybe_hash| .fromHash(maybe_hash),
+        };
+    }
+
+    /// Returns `.frozen` or `.unprocessed`.
+    pub fn fromHash(maybe_hash: ?sig.core.Hash) BankStatus {
+        if (maybe_hash) |hash| {
+            if (hash.eql(.ZEROES)) {
```
dnut: you don't need to handle this case because we use null to indicate unprocessed

src/replay/service.zig
```diff
@@ -206,19 +211,2562 @@ fn trackNewSlots(
...
+            else
+                .{ .hash = hash: {
+                    const bank_forks, var bank_forks_lg = bank_forks_rwmux.readWithLock();
+                    defer bank_forks_lg.unlock();
+
+                    const bank = bank_forks.slots.get(epoch_slots_frozen_slot) orelse
+                        break :hash null;
+                    break :hash bank.state.hash.readCopy();
+                } },
+        );
```
dnut: I think this code is easier to understand if you consolidate all the layers of initialization in one place.

```suggestion
        const bank_status: BankStatus =
            if (progress.isDead(epoch_slots_frozen_slot) orelse false)
                .dead
            else hash: {
                const bank_forks, var bank_forks_lg = bank_forks_rwmux.readWithLock();
                defer bank_forks_lg.unlock();

                break :hash if (bank_forks.slots.get(epoch_slots_frozen_slot)) |bank|
                    if (bank.state.hash.readCopy()) |hash|
                        .{ .frozen = hash }
                    else
                        .unprocessed
                else
                    .unprocessed;
            };
```

With this I think you can delete most declarations in BankStatus.

src/replay/service.zig
```diff
@@ -206,19 +211,2562 @@ fn trackNewSlots(
...
+                .{ slot, duplicate_confirmed_hash, local_duplicate_confirmed_hash },
+            );
+        }
+        return local_duplicate_confirmed_hash;
+    }
+    const bank_frozen_hash = local_duplicate_confirmed_hash;
+    return bank_frozen_hash;
+}
+
+/// AUDIT: https://github.com/anza-xyz/agave/blob/0315eb6adc87229654159448344972cbe484d0c7/core/src/repair/cluster_slot_state_verifier.rs#L848
```
dnut: In the rest of the codebase we use the pattern `Analogous to [item name](url)`. I think it would be good to be consistent so it's easy to search for. Also I think it's valuable to include the actual name of the analogous item so it's easy to search for. Personally I don't like to view code in github, and I usually just search for the name in my editor.

src/replay/service.zig
```diff
@@ -206,19 +211,2562 @@ fn trackNewSlots(
...
+        duplicate_confirmed_slots: *const DuplicateConfirmedSlots,
+        fork_choice: *const sig.consensus.HeaviestSubtreeForkChoice,
+        epoch_slots_frozen_slots: *const EpochSlotsFrozenSlots,
+    ) DeadState {
+        _ = slot; // autofix
+        _ = duplicate_slots_tracker; // autofix
+        _ = duplicate_confirmed_slots; // autofix
+        _ = fork_choice; // autofix
+        _ = epoch_slots_frozen_slots; // autofix
+    }
```
dnut: Missing implementation? Can we just remove this?

Going off on a tangent... I was thinking about adding a lint to `scripts/style.py` to disallow `// autofix` anywhere in the code. What do you think about that?

src/replay/service.zig
```diff
@@ -206,19 +211,2562 @@ fn trackNewSlots(
...
+    _ = &processPopularPrunedForks; // TODO:
+
+    _ = &processDuplicateSlots; // TODO:
+}
+
+const DuplicateSlotsToRepair = std.AutoArrayHashMapUnmanaged(
+    sig.core.Slot,
+    sig.core.Hash,
+);
+const DuplicateSlotsTracker = sig.utils.collections.SortedMapUnmanaged(
```
dnut: How about just `DuplicateSlots`? I agree the overuse of "tracker" is a bit of an issue. In this case simply omitting the word seems better.

src/replay/service.zig
```diff
@@ -206,19 +211,2562 @@ fn trackNewSlots(
...
+    );
+}
+
+/// Finds the duplicate confirmed hash for a slot.
+///
+/// 1) If `maybe_bank_frozen_hash != null and isDuplicateConfirmed(maybe_bank_frozen_hash.?)`, `return maybe_bank_frozen_hash.?`
+/// 2) If `maybe_duplicate_confirmed_hash != null`, `return maybe_duplicate_confirmed_hash.?`
+/// 3) Else return null
+fn getDuplicateConfirmedHash(
+    logger: sig.trace.Logger,
```
dnut: What do you think about consolidating these four functions to be just two functions?

1. BankFrozenState.fromState
2. getClusterConfirmedHashFromState
3. getDuplicateConfirmedHashFromState
4. getDuplicateConfirmedHash

I think getClusterConfirmedHashFromState can be absorbed into BankFrozenState.fromState, and getDuplicateConfirmedHashFromState can probably just be eliminated and replaced with direct calls to `get` and getDuplicateConfirmedHash.

src/replay/service.zig
```diff
@@ -206,19 +211,2562 @@ fn trackNewSlots(
...
+    }
+    const bank_frozen_hash = local_duplicate_confirmed_hash;
+    return bank_frozen_hash;
+}
+
+/// AUDIT: https://github.com/anza-xyz/agave/blob/0315eb6adc87229654159448344972cbe484d0c7/core/src/repair/cluster_slot_state_verifier.rs#L848
+const check_slot_agrees_with_cluster = struct {
+    fn bankFrozen(
+        allocator: std.mem.Allocator,
+        logger: sig.trace.Logger,
```
dnut: Can you scope all these loggers?

It's easy now to convert loggers to the correct scope when they have a different scope. You can pass it as `.from(logger)`

src/replay/service.zig
```diff
@@ -206,19 +211,2562 @@ fn trackNewSlots(
...
+        //     (
+        //         "my_hash",
+        //         bank_status
+        //             .bank_hash()
+        //             .unwrap_or_default()
+        //             .to_string(),
+        //         String
+        //     ),
+        // );
+
```
dnut: Could you replace comments like this with note saying this might be a good place to record metrics since agave does it here? We can always look at the agave code to see exactly how they do it.

src/replay/service.zig
```diff
@@ -206,19 +211,2562 @@ fn trackNewSlots(
...
+        }
+
+        // Needs to happen before the bank_frozen_hash.is_none() check below to account for duplicate
+        // signals arriving before the bank is constructed in replay.
+        if (try duplicate_slots_tracker.fetchPut(allocator, slot, {})) |_| {
+            // If this slot has already been processed before, return
+            return;
+        }
+
+        // datapoint_info!(
```
dnut: same as above

src/replay/service.zig
```diff
@@ -206,19 +211,2562 @@ fn trackNewSlots(
...
+
+    /// Ordered from strongest confirmation to weakest. Stronger
+    /// confirmations take precedence over weaker ones.
+    pub const Kind = enum {
+        duplicate_confirmed,
+        epoch_slots_frozen,
+    };
+};
+
+pub const BankStatus = union(enum) {
```
dnut: We don't have a "bank" so can you rename this to something that is relevant to sig? How about SlotStatus?

In general, I would probably remove all instances of the word "bank" from this code.

src/replay/service.zig
```diff
@@ -206,19 +211,2562 @@ fn trackNewSlots(
...
+            logger.err().logf(
+                "For slot {}, the gossip duplicate confirmed hash {}, is not equal" ++
+                    "to the confirmed hash we replayed: {}",
+                .{ slot, duplicate_confirmed_hash, local_duplicate_confirmed_hash },
+            );
+        }
+        return local_duplicate_confirmed_hash;
+    }
+    const bank_frozen_hash = local_duplicate_confirmed_hash;
+    return bank_frozen_hash;
```
dnut: I found this pretty confusing at first. I think it's easier to understand if you distinguish the conditional logging from the determination of the return value.

```suggestion
    if (maybe_local_duplicate_confirmed_hash) |local_duplicate_confirmed_hash| {
        if (maybe_duplicate_confirmed_hash) |duplicate_confirmed_hash| {
            if (!local_duplicate_confirmed_hash.eql(duplicate_confirmed_hash)) {
                logger.err().logf(
                    "For slot {}, the gossip duplicate confirmed hash {}, is not equal" ++
                        "to the confirmed hash we replayed: {}",
                    .{ slot, duplicate_confirmed_hash, local_duplicate_confirmed_hash },
                );
            }
        }
    }

    return maybe_local_duplicate_confirmed_hash orelse maybe_duplicate_confirmed_hash;
```

src/replay/service.zig
```diff
@@ -206,19 +211,2562 @@ fn trackNewSlots(
...
+            std.debug.assert(!self.finalized);
+            self.finalized = true;
+            if (self.frozen_hash) |frozen_hash| {
+                try blockstore.insertBankHash(slot, frozen_hash, false);
+            }
+        }
+    };
 
-    // TODO: process_duplicate_confirmed_slots
+    fn bankFrozen(
```
dnut: The meaning of this function name is a bit confusing. Can you change it to indicate that it's mutating some state? How about `freezeSlot` or `setSlotHash`?

A slot (bank) can't have a hash without being frozen, so maybe "frozen" is redundant.

Also this function seems like it's essentially a setter for NotDupeConfirmedFrozenHash. I think the intent would be clearer if this were a method on that struct.

src/replay/service.zig
```diff
@@ -206,19 +211,2562 @@ fn trackNewSlots(
...
-    // TODO: process_gossip_verified_vote_hashes
+    fn repairDuplicateConfirmedVersion(
+        allocator: std.mem.Allocator,
+        slot: u64,
+        duplicate_slots_to_repair: *DuplicateSlotsToRepair,
+        duplicate_confirmed_hash: sig.core.Hash,
+    ) !void {
+        try duplicate_slots_to_repair.put(allocator, slot, duplicate_confirmed_hash);
+    }
+
```
dnut: I would probably just remove this function. It's not clear what purpose it serves, and the indirection obfuscates what's actually going on.

src/replay/service.zig
```diff
@@ -206,19 +211,2562 @@ fn trackNewSlots(
...
+        state_change.sendAncestorHashesReplayUpdate(
+            ancestor_hashes_replay_update_sender,
+            .{ .kind = .popular_pruned_fork, .slot = slot },
+        );
+    }
+};
+
+/// AUDIT: https://github.com/anza-xyz/agave/blob/0315eb6adc87229654159448344972cbe484d0c7/core/src/repair/cluster_slot_state_verifier.rs#L793
+const state_change = struct {
+    const NotDupeConfirmedFrozenHash = struct {
```
dnut: I'm confused by this name. What is this struct for? Can you document its meaning? is there maybe a better name for it?

src/replay/service.zig
```diff
@@ -206,19 +211,2562 @@ fn trackNewSlots(
...
-    // TODO: process_duplicate_confirmed_slots
+    fn bankFrozen(
+        slot: u64,
+        fork_choice: *const sig.consensus.HeaviestSubtreeForkChoice,
+        not_dupe_confirmed_frozen_hash: *NotDupeConfirmedFrozenHash,
+        bank_frozen_hash: sig.core.Hash,
+    ) void {
+        const is_duplicate_confirmed =
+            fork_choice.isDuplicateConfirmed(&.{ .slot = slot, .hash = bank_frozen_hash }) orelse
+            std.debug.panic("frozen bank must exist in fork choice", .{});
```
dnut: Is this guaranteed by the other code in this file, or just an assumption?

src/replay/service.zig
```diff
@@ -206,19 +211,2562 @@ fn trackNewSlots(
...
+            allocator,
+            slot,
+            epoch_slots_frozen_hash,
+        )) |old_epoch_slots_frozen_hash_kv| {
+            const old_epoch_slots_frozen_hash = old_epoch_slots_frozen_hash_kv.value;
+            if (old_epoch_slots_frozen_hash.eql(epoch_slots_frozen_hash)) {
+                // If EpochSlots has already told us this same hash was frozen, return
+                return;
+            }
+        }
```
dnut: What if it's a different hash? We just discard it? A discrepancy here seems noteworthy.

src/replay/service.zig
```diff
@@ -206,19 +211,2562 @@ fn trackNewSlots(
...
+    bank_forks_rwmux: *sig.sync.RwMux(sig.replay.trackers.SlotTracker),
+    duplicate_slots_to_repair: *DuplicateSlotsToRepair,
+) !void {
+    const root = root: {
+        const bank_forks, var bank_forks_lg = bank_forks_rwmux.readWithLock();
+        defer bank_forks_lg.unlock();
+        break :root bank_forks.root;
+    };
+
+    while (ancestor_duplicate_slots_receiver.tryReceive()) |ancestor_dupe_slot_to_repair| {
```
dnut: Just a note for the future: we'll need to implement the source of this data, but it doesn't need to happen in this PR.

src/replay/service.zig
```diff
@@ -206,19 +211,2562 @@ fn trackNewSlots(
...
+        _ = purge_repair_slot_counter.orderedRemove(slot);
+    }
 
-    // TODO: process_duplicate_slots
+    fn sendAncestorHashesReplayUpdate(
+        ancestor_hashes_replay_update_sender: *sig.sync.Channel(AncestorHashesReplayUpdate),
+        ancestor_hashes_replay_update: AncestorHashesReplayUpdate,
+    ) void {
+        ancestor_hashes_replay_update_sender.send(ancestor_hashes_replay_update) catch {
+            // TODO: agave just ignores this error, is that alright?
```
dnut: I don't think so.

src/replay/service.zig
```diff
@@ -206,19 +211,2562 @@ fn trackNewSlots(
...
+        };
+    }
```
dnut: I would remove this function

src/replay/service.zig
```diff
@@ -206,19 +211,2562 @@ fn trackNewSlots(
...
+        blockstore: *sig.ledger.LedgerResultWriter,
+        purge_repair_slot_counter: *PurgeRepairSlotCounter,
+        not_dupe_confirmed_frozen_hash: *NotDupeConfirmedFrozenHash,
+        bank_frozen_hash: sig.core.Hash,
+    ) !void {
+        not_dupe_confirmed_frozen_hash.update(null);
+        // When we detect that our frozen slot matches the cluster version (note this
+        // will catch both bank frozen first -> confirmation, or confirmation first ->
+        // bank frozen), mark all the newly duplicate confirmed slots in blockstore
+        const new_duplicate_confirmed_slot_hashes = try fork_choice.markForkValidCandidate(&.{
```
dnut: I'm confused by this. I'm understanding "duplicate confirmed slot" to mean that a slot has been confirmed to be duplicated. In other words, the leader produced two blocks, and it's should no longer be treated as a valid slot. In fact, another function called `markSlotDuplicate` is implemented by simply marking the fork as invlaid. But here the fork is being marked as a valid candidate, which seems contradictory. Do you know what "duplicate confirmed slot" means exactly in this context and how it differs from the other context?

src/replay/service.zig
```diff
@@ -206,19 +211,2562 @@ fn trackNewSlots(
...
+    const root = root: {
+        const bank_forks, var bank_forks_lg = bank_forks_rwmux.readWithLock();
+        defer bank_forks_lg.unlock();
+        break :root bank_forks.root;
+    };
+    while (popular_pruned_forks_receiver.tryReceive()) |new_popular_pruned_slot| {
+        if (new_popular_pruned_slot <= root) {
+            continue;
+        }
+        check_slot_agrees_with_cluster.popularPrunedFork(
```
dnut: What do you think about inlining this function here and deleting it, since it's the only usage and doesn't do much.

dnut (changes requested):
I like how you've vastly simplified this code relative to the agave implementation, but it seems that there is still some residual agave-based structure in here that can be simplified easily.

Also I haven't noted every case with a comment, but a lot of the function names are ambiguous. They describe the condition when they would be called, but don't actually say anything about what the function is doing. There may be some room for improvement there.

src/replay/service.zig
InKryption: I think had a similar question to @dadepo a while ago in a meeting, and to my recollection I believe that they're related but not the same, in that you can have a duplicate slot, which is either valid or invalid, where the valid one is what we care about, and the invalid one(s) are what we'll eventually discard/prune. @dadepo could you confirm or elucidate?

src/replay/service.zig
InKryption: > Missing implementation? Can we just remove this?

I intend to implement this, it just wasn't relevant in this PR because nothing directly uses it, but it's used in other places during replay, e.g. see `fn mark_dead_slot`.
Should probably implement it in this PR.

> Going off on a tangent... I was thinking about adding a lint to scripts/style.py to disallow // autofix anywhere in the code. What do you think about that?

Not a bad idea. Only reason they're here right now is because I was using the warnings ZLS issues in the editor as a reminder to attend to them at some point, but probably a good idea to not allow that into changesets.

src/replay/service.zig
InKryption: Good call, no point in having this now that everything is already ported over.

InKryption: ff2b8b86f49f2b78a737f3b5894d237f2ac1ee65

src/replay/service.zig
InKryption: ff2b8b86f49f2b78a737f3b5894d237f2ac1ee65

src/replay/service.zig
InKryption: I actually do this in the PR stacked on top of this one, since that's where I started down the path of starting to connect this code to the existing code.

src/replay/service.zig
InKryption: 7d63191818fc90af35784c3112ee6b17a3972c03

src/replay/service.zig
InKryption: 2626e0142143ab6d6aa296ff3e0d7285f6106d28

src/replay/service.zig
InKryption: 70bdad79e161087becd9a4a77ff13b35f2cb34f9

src/replay/service.zig
InKryption: You actually made me realize an even better simplification that can be made here; agave does this weird monadic thing of carrying the None/null forwards through this logic, when an earlier return actually makes a lot more sense.

InKryption: a181ce2202731d70402a816807afaea265776f47

src/replay/service.zig
InKryption: It's an abstraction over some code that had to be repeated throughout a few different functions in specific ways which originally were consolidated by virtue of agave bundling the collective logic of apply_state_changes in one function loop.

In terms of what it's for, it's for keeping track of a potential confirmed frozen hash, which isn't a duplicate, in order to insert it into the blockstore if it's confirmed to not be duplicated.
In agave it's just this:
https://github.com/anza-xyz/agave/blob/0315eb6adc87229654159448344972cbe484d0c7/core/src/repair/cluster_slot_state_verifier.rs#L804

Not sure what exactly a better name would be, the name is about as literal as can be, just short of appending `ToInsertInLedger`.

Will add a doc comment giving some context.

src/replay/service.zig
InKryption: The hash will have been received through `ancestor_duplicate_slots_receiver`, which is captured as `ancestor_dupe_slot_to_repair`, with its origin being in `ancestor_hashes_service.rs`:
https://github.com/anza-xyz/agave/blob/0315eb6adc87229654159448344972cbe484d0c7/core/src/repair/ancestor_hashes_service.rs#L494-L502

I don't understand the broader mechanism at play here, but it seems to me based on that comment's wording, indicating we should be "dumping the fork", that discarding it is intentional.

src/replay/service.zig
InKryption: 71e17b8765493bcaf2c07c7ec7688ec3d39f841

Went a bit more descriptive in the comment for `state_change`.

src/replay/service.zig
dadepo: So my first encounter of the term "duplicate confirmed" was in the context of the fork choice, and my interpretation of what it means is that it is used to refer to a slot that was previously marked as a duplicate but has now been confirmed by the network as the canonical version.

And I think this is what is happening in this part of the code. The duplicate confirmed is more like this is a duplicate, that has now being confirmed to be canonical/valid.

I agree it is not the best of names, and the same questions came up when the method was being implemented in https://github.com/Syndica/sig/pull/505#discussion_r1962361481

A better name could be `confirmed_canonical_fork`? But the risk of renaming some of this variables/method now, while implementing based on Agave is that it might lead to some confusion where it is not exactly sure what concepts in Agave is being implemented in Sig. - but again, the current naming is not the best

InKryption: I would prefer to stick to the current agave naming conventions, at least until we ourselves have a better fundamental understanding of everything here, and we're no longer relying on the agave code as a source of truth for replay.

src/replay/service.zig
InKryption: f185742552ed9cf3acb6fac9a8eafc16ed304f7f

src/replay/service.zig
InKryption: 775f0ba630d97b33ac6baec7b3311daf0a15c1d6

src/replay/service.zig
InKryption: 775f0ba630d97b33ac6baec7b3311daf0a15c1d6

src/replay/service.zig
InKryption: I believe that it was guaranteed by what was the broader state machine, in the sense that this code would only be reached if the code that set up this function to be invoked observed the slothash inside fork_choice. But given that it's no longer part of a broader state machine like that, may be wise to instead log and return an error here

src/replay/service.zig
InKryption: The most accurate name would probably be something like `maybeUpdateNotDupeConfirmedFrozenHash`, seeing as all this is doing is updating the pending non-dupe confirmed frozen hash in the situation where it's been potentially changed.

I don't think it makes sense as a method on NotDupeConfirmedFrozenHash.

InKryption: 205ab3108a06644fc61a1891ad077bddcc9fa97c

src/replay/service.zig
InKryption: 205ab3108a06644fc61a1891ad077bddcc9fa97c

src/replay/service.zig
InKryption: 205ab3108a06644fc61a1891ad077bddcc9fa97c

src/replay/service.zig
InKryption: 205ab3108a06644fc61a1891ad077bddcc9fa97c

src/replay/service.zig
InKryption: 205ab3108a06644fc61a1891ad077bddcc9fa97c

src/replay/service.zig
InKryption: I took a slightly different approach in 5dd4b931bd64d0fbbeefeb0344873ed8b53c2c6a, because as of f185742552ed9cf3acb6fac9a8eafc16ed304f7f, `getClusterConfirmedHashFromState` is actually used in more than one place.
Went ahead and inlined the other one though.

src/replay/service.zig
InKryption: Changed "not" to "non", and added documentation detailing the motivation behind its existence: f308765585ee7d31a67cc0476ae8488c56df4d09

src/replay/service.zig
dadepo: Recently came across this doc in Agave https://github.com/anza-xyz/agave/blob/f39ea99f6c2382e15d82391647c1e0cb8e951234/docs/src/proposals/handle-duplicate-block.md?plain=1#L21 which might be illustrative.

src/replay/service.zig
dnut: The singular name is a little confusing. If you don't like `DuplicateSlots`, what about `DuplicateSlotSet`?

src/replay/service.zig
dnut: This comment helps a lot. But I think the name still has some room for improvement.

"non duplicate confirmed" implies to me that it's "confirmed as non-duplicate" or "not confirmed as a duplicate," but it's actually definitely a duplicate slot, it's just that it's also a non-confirmed slot. How about `DupeUnconfirmedFrozenHash`? This could still be misinterpreted in the same way, but I think it leans more towards the correct interpretation, where the slot is viewed as both duplicate and unconfirmed. It also hints that the "frozen hash" is unconfirmed, which is another correct interpretation. You could also expand on the documentation to clarify that "this slot is both duplicate and unconfirmed."

Also, can you constrain the comments to fit within the 100 character line limit unless it's strictly necessary (for example urls).

src/replay/service.zig
InKryption: Apologies, that was purely a typo, I fixed that up in a follow up commit

InKryption: 88497f8176c5d50a904ee7bf8402aa14a4ef425e

src/replay/service.zig
dnut: I understand this concept now so I'm glad I asked this question. I believe we should understand the code that we're writing and not just port it over blindly and hope that we can compare it to agave to figure out what's going on. I am in favor of renaming everything to be as clear as possible without worrying too much what names agave uses. Our code should be a reference implementation that stands on its own. We can use documentation to correlate the code to agave.

The name needs to represent two independent attributes: duplicate and confirmed. `confirmed_canonical_fork` doesn't indicate whether the slot is duplicate, so that's ambiguous. There are four variants. A slot can be:
- duplicate and confirmed
- duplicate and not confirmed
- not duplicate and confirmed
- not duplicate and not confirmed

The word "and" might bring some additional clarity here that these are independent attributes. Those four bullet points can be the exact names we use. So for example, the function could be called `duplicateAndConfirmedSlotMatchesCluster`.

It is tricky to name this concept clearly, and we should continue to give it more thought after merging this PR.

src/replay/service.zig
dnut: bump

src/replay/service.zig
dnut: I think I get the idea now. It looks like we're only interested in repairing whatever is the latest version that was decided as the "correct" hash and we don't need to keep trying to repair old versions. This logic seems reasonable.

src/replay/service.zig
InKryption: My understanding differs, it isn't a duplicate, rather it's a non-duplicate and confirmed frozen hash, ie it is canonical and not duplicated, reflected in the fact that `finalize` inserts the bank hash with `is_duplicate_confirmed = false`. In boolean terms, it is `!duplicate and confirmed`, which for the purposes of the ledger, is categorically the same as `!is_duplicate_confirmed`, since `is_duplicate_confirmed = duplicate and confirmed`.

InKryption: Perhaps `ConfirmedNonDupeFrozenHash` would be more clear?

InKryption: 788741f69c41f285cc32a070db8e726dfc96f1ec

src/replay/service.zig
InKryption: cfecae1ec005615a49acbefa75164aca679893ce

src/replay/service.zig
InKryption: 6ebda515e793c2f0c5856f2116c51e8dcee8080e

took a minute on this because I wanted to make exhaustively sure it wasn't actually going to be used elsewhere at some point

src/replay/service.zig
dnut: @dadepo any more thoughts on this?

dnut (approved):

src/consensus/latest_validator_votes.zig
```diff
@@ -148,11 +157,15 @@ const builtin = @import("builtin");
 pub fn latestVote(
     lastest_votes: *const LatestValidatorVotes,
     pubkey: Pubkey,
-    is_replay_vote: bool,
+    vote_kind: LatestValidatorVotes.VoteKind,
 ) ?struct { slot: Slot, hashes: []const Hash } {
     if (!builtin.is_test) {
         @compileError("latestVote should only be called in test mode");
     }
+    const is_replay_vote = switch (vote_kind) {
```
dadepo, line 165: nit: `const is_replay_vote = vote_kind == .replay;`

dadepo (approved):

src/consensus/latest_validator_votes.zig
InKryption: If it's alright I'd rather this get addressed as part of some refactor in a future PR so we can merge this now while codecov is working, otherwise the added commit may invalidate it and put us back to waiting.

PR #781, opened 2025-06-12
fix: incremental compilation

dnut:
### Problem
Incremental compilation doesn't work for sig because one of our dependencies (libxev) uses the `usingnamespace` keyword.

### Fix
I forked libxev and removed usingnamespace here: https://github.com/dnut/libxev and switched to that as sig's dependency.

### Verify
Run `zig build -fincremental`. It panics on main with this error:

> error: thread 3490520122 panic: 'usingnamespace' is not supported by incremental compilation

It runs without error on this branch.

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -30,8 +30,8 @@
             .hash = "lsquic-0.1.0-oZGGDUqdHQGUkmpUvZT5RmJ07lE-EoX7vfYFVFePmS9P",
         },
         .xev = .{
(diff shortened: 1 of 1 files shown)


InKryption:
Should this rather be under the Syndica org?

dnut:
I don't think it makes a difference, since either way you need a new commit in the sig repo to change it. But sure I can move it to Syndica.

dnut:
Actually I was able to fix this by deleting my own repo. Now the Syndica fork points correctly upstream to mitchell's repo and we don't need to delete/recreate it.

InKryption (approved):

yewman (approved):

PR #782, opened 2025-06-12
refactor(runtime): change transaction context account to take mutable pointer to AccountSharedData

yewman:
- Most changes in this PR relate to managing pointer lifetimes in testing contexts
- When creating the trancation context, we now create and return an account cache which stores the AccountSharedData's
- The accounts within this cache must live as long as the transaction context accounts

### DIFF
--- a/src/runtime/program/bpf/serialize.zig
+++ b/src/runtime/program/bpf/serialize.zig
@@ -886,16 +886,16 @@ test "serializeParameters" {
             };
 
             const pre_accounts = blk: {
-                var accounts = std.ArrayList(struct {
+                var accounts = try std.ArrayListUnmanaged(struct {
                     pubkey: Pubkey,
                     account: AccountSharedData,
-                }).init(allocator);
+                }).initCapacity(allocator, tc.accounts.len);
                 errdefer {
                     for (accounts.items) |account| allocator.free(account.account.data);
-                    accounts.deinit();
+                    accounts.deinit(allocator);
                 }
                 for (tc.accounts) |account| {
-                    try accounts.append(.{
+                    accounts.appendAssumeCapacity(.{
                         .pubkey = account.pubkey,
                         .account = .{
                             .lamports = account.account.lamports,
@@ -906,7 +906,7 @@ test "serializeParameters" {
                         },
                     });
                 }
-                break :blk try accounts.toOwnedSlice();
+                break :blk try accounts.toOwnedSlice(allocator);
             };
             defer {
                 for (pre_accounts) |account| allocator.free(account.account.data);
--- a/src/runtime/program/bpf/serialize.zig
+++ b/src/runtime/program/bpf/serialize.zig
@@ -726,7 +726,7 @@ fn deserializeParametersAligned(
 
 // [agave] https://github.com/anza-xyz/agave/blob/108fcb4ff0f3cb2e7739ca163e6ead04e377e567/program-runtime/src/serialization.rs#L778
 test "serializeParameters" {
-    const TransactionContextAccount = sig.runtime.TransactionContextAccount;
+    const AccountSharedData = sig.runtime.AccountSharedData;
     const createTransactionContext = sig.runtime.testing.createTransactionContext;
     const deinitTransactionContext = sig.runtime.testing.deinitTransactionContext;
     const createInstructionInfo = sig.runtime.testing.createInstructionInfo;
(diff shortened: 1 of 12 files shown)


src/runtime/program/bpf/serialize.zig
```diff
@@ -883,24 +886,30 @@ test "serializeParameters" {
...
                 }
                 for (tc.accounts) |account| {
-                    try accounts.append(TransactionContextAccount.init(account.pubkey, .{
-                        .lamports = account.account.lamports,
-                        .owner = account.account.owner,
-                        .data = try allocator.dupe(u8, account.account.data),
-                        .executable = account.account.executable,
-                        .rent_epoch = account.account.rent_epoch,
-                    }));
+                    try accounts.append(.{
```
InKryption: This could be made into `accounts.appendAssumeCapacity` by first doing `try accounts.ensureTotalCapacityPrecise(allocator, tc.accounts.len)`.
Also might be wise to get used to using an unmanaged arraylist even in these situations, as managed data structures will likely be removed in the next or subsequent release cycle.

src/runtime/testing.zig
```diff
@@ -108,21 +109,35 @@ pub fn createTransactionContext(
...
+                .data = try allocator.dupe(u8, account_params.data),
+                .owner = account_params.owner orelse Pubkey.initRandom(random),
+                .executable = account_params.executable,
+                .rent_epoch = account_params.rent_epoch,
+            },
         );
     }
 
+    for (account_keys.items) |key| {
+        try accounts.append(TransactionContextAccount.init(
```
InKryption:
```suggestion
    try accounts.ensureUnusedCapacity(account_keys.items.len);
    for (account_keys.items) |key| {
        accounts.appendAssumeCapacity(TransactionContextAccount.init(
```

src/runtime/testing.zig
```diff
@@ -108,21 +109,35 @@ pub fn createTransactionContext(
...
-                    .lamports = account_params.lamports,
-                    .data = try allocator.dupe(u8, account_params.data),
-                    .owner = account_params.owner orelse Pubkey.initRandom(random),
-                    .executable = account_params.executable,
-                    .rent_epoch = account_params.rent_epoch,
-                },
-            ),
+        const key = account_params.pubkey orelse Pubkey.initRandom(random);
+        try account_keys.append(key);
+        try account_cache.account_cache.put(
```
InKryption: Should do `try account_keys.ensureUnusedCapacity(params.accounts.len)` and `try account_cache.account_cache.ensureUnusedCapacity(params.accounts.len)` beforehand and *`AssumeCapacity` for this append and put.

src/runtime/transaction_context.zig
```diff
@@ -86,7 +86,6 @@ pub const TransactionContext = struct {
     }, MAX_INSTRUCTION_TRACE_LENGTH);
 
     pub fn deinit(self: TransactionContext) void {
-        for (self.accounts) |account| account.deinit(self.allocator);
```
InKryption, line 89: I don't quite understand this change, does the context not own these? If not, could we have a comment on why, and/or who actually owns them? So as to avoid some well-intentioned but misinformed refactor adding this back.

yewman: The TransactionContextAccount no longer owns the AccountSharedData so there is nothing to deinit, I think the fact that the AccountSharedData in the TransactionContextAccount is now a pointer is evidence enough that it does not own the AccountSharedData. I can add a comment if you feel strongly though :)

src/runtime/program/bpf/serialize.zig
yewman: f68320a6

src/runtime/testing.zig
yewman: f68320a6

src/runtime/testing.zig
yewman: f68320a6

src/runtime/transaction_context.zig
InKryption: Yeah I would prefer a comment; I see what you're saying now that you've pointed it out, but that wasn't immediately obvious to me otherwise. Some simple comment on the field that says "each `TransactionContextAccount` is a non-owning reference".

yewman: f0fe0f0f

InKryption (approved):

PR #783, opened 2025-06-13
eliminate `src/time/time.zig`

alichraghi:
See the commit message for details


dnut (changes requested):
I don't think sig's Duration, Instant, and Timer types are redundant or silly wrappers. Strong types are valuable. They communicate intent and reduce the likelihood of confusion or misuse.

Is there a specific problem caused by these types? If so, maybe there's another way to address it rather than switching all time values to be represented as plain integer types.

alichraghi:
I can see the concern about `Duration` no longer being a strong type. I'll address that with a tagged enum `enum(u64)` instead.

However, leaky abstractions like `Instant` and `Timer` only add maintenance burden and hide stdlib API documentation with no benefits. Both `Instant.UNIX_EPOCH` and `Instant.plus()` are already covered by `std.time.instantFromEpoch`, and I see `Instant.elapsed()` as poor API design because it forces you to call `Instant.now()` in [some cases](https://github.com/Syndica/sig/pull/783/files#diff-56b9278a4ecf8ab2a0a52baee3072ede5acc35509654e436dda14682a72522f2R432-R444), creating unnecessary coupling.

PR #786, opened 2025-06-16
feat(runtime): program map

yewman:
Implement program map for bpf execution
sig-fuzz: https://github.com/Syndica/sig-fuzz/pull/15

Failing conformance fixtures:
vm_interp/v3: 9/36261 failed, 0 skipped,
syscall/alt_bn128: PANICKED,
syscall/secp256k1: PANICKED,
instr/bpf-loader-upgradeable-v1-programs: 75/371 failed, 0 skipped,
instr/bpf-loader-v3-programs: 456/509 failed, 0 skipped,
instr/stake: 6934/6934 failed, 0 skipped,
instr/zk_sdk: 168/168 failed, 0 skipped

### DIFF
--- a/src/runtime/program_loader.zig
+++ b/src/runtime/program_loader.zig
@@ -67,12 +67,12 @@ pub fn loadProgram(
     slot: u64,
 ) !LoadedProgram {
     // executable bytes are owned by the entry in the accounts map and should not be freed
-    const deployment_slot, const executable_bytes = loadDeploymentSlotAndExecutableBytes(
+    const maybe_deployment_slot, const executable_bytes = loadDeploymentSlotAndExecutableBytes(
         account,
         accounts,
     ) orelse return .failed;
 
-    if (deployment_slot >= slot) return .failed;
+    if (maybe_deployment_slot) |ds| if (ds >= slot) return .failed;
 
     const source = try allocator.dupe(u8, executable_bytes);
     var executable = Executable.fromBytes(
@@ -101,9 +101,9 @@ pub fn loadProgram(
 pub fn loadDeploymentSlotAndExecutableBytes(
     account: *const AccountSharedData,
     accounts: *const std.AutoArrayHashMapUnmanaged(Pubkey, AccountSharedData),
-) ?struct { u64, []const u8 } {
+) ?struct { ?u64, []const u8 } {
     if (account.owner.equals(&bpf_loader.v1.ID) or account.owner.equals(&bpf_loader.v2.ID)) {
-        return .{ 0, account.data };
+        return .{ null, account.data };
     } else if (account.owner.equals(&bpf_loader.v3.ID)) {
         const program_state = sig.bincode.readFromSlice(
             failing_allocator,
@@ -147,6 +147,77 @@ pub fn loadDeploymentSlotAndExecutableBytes(
         return null;
     } else {
         return null;
+    }
+}
+
+test "loadPrograms: load v1, v2 program" {
+    const allocator = std.testing.allocator;
+    var prng = std.Random.DefaultPrng.init(0);
+
+    const program_elf = try std.fs.cwd().readFileAlloc(
+        allocator,
+        sig.ELF_DATA_DIR ++ "hello_world.so",
+        std.math.maxInt(usize),
+    );
+    defer allocator.free(program_elf);
+
+    var accounts = std.AutoArrayHashMapUnmanaged(Pubkey, AccountSharedData){};
+    defer accounts.deinit(allocator);
+
+    const program_v1_key = Pubkey.initRandom(prng.random());
+    const program_v2_key = Pubkey.initRandom(prng.random());
+
+    try accounts.put(
+        allocator,
+        program_v1_key,
+        .{
+            .lamports = 1,
+            .owner = bpf_loader.v1.ID,
+            .data = program_elf,
+            .executable = true,
+            .rent_epoch = std.math.maxInt(u64),
+        },
+    );
+
+    try accounts.put(
+        allocator,
+        program_v2_key,
+        .{
+            .lamports = 1,
+            .owner = bpf_loader.v2.ID,
+            .data = program_elf,
+            .executable = true,
+            .rent_epoch = std.math.maxInt(u64),
+        },
+    );
+
+    const environment = vm.Environment{
+        .loader = .{},
+        .config = .{},
+    };
+    defer environment.deinit(allocator);
+
+    { // Success
+        var loaded_programs = try loadPrograms(
+            allocator,
+            &accounts,
+            &environment,
+            0,
+        );
+        defer {
+            for (loaded_programs.values()) |*v| v.deinit(allocator);
+            loaded_programs.deinit(allocator);
+        }
+
+        switch (loaded_programs.get(program_v1_key).?) {
+            .failed => std.debug.panic("Program failed to load!", .{}),
+            .loaded => {},
+        }
+
+        switch (loaded_programs.get(program_v2_key).?) {
+            .failed => std.debug.panic("Program failed to load!", .{}),
+            .loaded => {},
+        }
     }
 }
 
@@ -281,6 +352,53 @@ test "loadPrograms: load v3 program" {
         );
         defer {
             for (loaded_programs.values()) |*loaded_program| loaded_program.deinit(allocator);
+            loaded_programs.deinit(allocator);
+        }
+
+        switch (loaded_programs.get(program_key).?) {
+            .failed => {},
+            .loaded => std.debug.panic("Program should not load!", .{}),
+        }
+    }
+}
+
+test "loadPrograms: bad owner" {
+    const allocator = std.testing.allocator;
+    var prng = std.Random.DefaultPrng.init(0);
+
+    var accounts = std.AutoArrayHashMapUnmanaged(Pubkey, AccountSharedData){};
+    defer {
+        for (accounts.values()) |account| allocator.free(account.data);
+        accounts.deinit(allocator);
+    }
+
+    const program_key = Pubkey.initRandom(prng.random());
+    try accounts.put(
+        allocator,
+        program_key,
+        .{
+            .lamports = 1,
+            .owner = Pubkey.initRandom(prng.random()),
+            .data = try allocator.dupe(u8, &.{ 10, 0, 10, 3 }),
+            .executable = true,
+            .rent_epoch = std.math.maxInt(u64),
+        },
+    );
+
+    const environment = vm.Environment{
+        .loader = .{},
+        .config = .{},
+    };
+
+    { // Failed to load program with bad owner
+        var loaded_programs = try loadPrograms(
+            allocator,
+            &accounts,
+            &environment,
+            prng.random().int(u64),
+        );
+        defer {
+            for (loaded_programs.values()) |*v| v.deinit(allocator);
             loaded_programs.deinit(allocator);
         }
 


src/runtime/sysvar_cache.zig
```diff
@@ -35,8 +35,8 @@ pub const SysvarCache = struct {
     stake_history_obj: ?StakeHistory = null,
 
     // deprecated sysvars, these should be removed once practical
-    fees: ?Fees = null,
-    recent_blockhashes: ?RecentBlockhashes = null,
+    fees_obj: ?Fees = null,
+    recent_blockhashes_obj: ?RecentBlockhashes = null,
```
Sobeston, line 39: Missing something, what exactly is an "object" in this context?

src/runtime/testing.zig
```diff
@@ -29,6 +31,13 @@ pub const ExecuteContextsParams = struct {
     feature_set: []const FeatureParams = &.{},
     epoch_stakes: []const EpochStakeParam = &.{},
 
+    // Programs to be inseted into the program map.
+    program_map: *const ProgramMap = &.{},
```
Sobeston: inseted -> inserted

src/runtime/transaction_context.zig
```diff
@@ -58,18 +58,18 @@ pub const TransactionContext = struct {
...
-    instruction_trace: InstructionTrace,
+    instruction_stack: InstructionStack = .{},
+    instruction_trace: InstructionTrace = .{},
     top_level_instruction_index: u16 = 0,
-    return_data: TransactionReturnData,
+    return_data: TransactionReturnData = .{},
 
     /// Total change to account data size within transaction
-    accounts_resize_delta: i64,
+    accounts_resize_delta: i64 = 0,
```
Sobeston, line 72: might be an anti-pattern to have default inits for all these fields?

src/runtime/program/bpf_loader/execute.zig
```diff
@@ -1431,54 +1431,42 @@ pub fn deployProgram(
...
+        &environment.loader,
+        environment.config,
     ) catch |err| {
         try tc.log("{s}", .{@errorName(err)});
         return InstructionError.InvalidAccountData;
     };
     defer executable.deinit(allocator);
 
+    executable.verify(&environment.loader) catch |err| {
+        try tc.log("{s}", .{@errorName(err)});
```
Sobeston: Some of these logs could be better, e.g. why not "executable (pubkey) failed to verify: (err)"

src/runtime/program_loader.zig
```diff
@@ -0,0 +1,261 @@
...
+
+    for (accounts.keys(), accounts.values()) |pubkey, account| {
+        if (account.executable)
+            try programs.put(allocator, pubkey, try loadProgram(
+                allocator,
+                &account,
+                accounts,
+                enviroment,
+                slot,
+            ));
```
Sobeston: nit: `if (!account.executable) continue` preferable

src/runtime/sysvar/recent_blockhashes.zig
```diff
@@ -23,6 +23,13 @@ pub const RecentBlockhashes = struct {
         allocator.free(self.entries);
     }
 
+    pub fn last(self: RecentBlockhashes) ?Entry {
+        return if (self.entries.len > 0)
+            self.entries[self.entries.len - 1]
+        else
+            null;
+    }
```
Sobeston: nit, could be:
```
if (self.entries.len == 0) return null;
return self.entries[self.entries.len - 1]
```

src/runtime/transaction_execution.zig
```diff
@@ -192,6 +196,7 @@ pub fn loadAndExecuteTransactions(
     environment: *const TransactionExecutionEnvironment,
     config: *const TransactionExecutionConfig,
 ) error{OutOfMemory}![]TransactionResult(ProcessedTransaction) {
+    const program_map = ProgramMap{};
```
Sobeston: could do with a comment for where the "real" program_map is supposed to come from

Sobeston (changes requested):
Broad strokes looking good. Got a couple Qs and nits

src/runtime/sysvar_cache.zig
yewman: just used to differentiate between sysvars contained in the cache which are stored as raw bytes vs deserialized type. Perhaps a `_bytes` suffix for those stored in raw form would be better, I am impartial

src/runtime/transaction_context.zig
yewman: Initialising these fields with anything other than the default is a non-standard use case so imo it kind of makes sense. If you think we should have no defaults for the transaction context then I can create an init fn with options

src/runtime/transaction_execution.zig
yewman: 3f67b89

src/runtime/sysvar/recent_blockhashes.zig
yewman: 3f67b89

src/runtime/program_loader.zig
yewman: 3f67b89

src/runtime/program/bpf_loader/execute.zig
yewman: 3f67b89

src/runtime/testing.zig
yewman: 3f67b89

src/runtime/sysvar_cache.zig
Sobeston: Yeah I see, I think this is fine as is

Sobeston (dismissed):

kprotty (approved):

src/runtime/executor.zig
```diff
@@ -134,24 +134,25 @@ fn processNextInstruction(
 
     // Lookup the program id
     // [agave] https://github.com/anza-xyz/agave/blob/1ceeab0548d5db0bbcf7dab7726c2ffb4180fa76/program-runtime/src/invoke_context.rs#L509-L533
-    const program_id = blk: {
+    const native_program_id, const program_id = blk: {
```
dnut, line 137: This makes sense now, but it seems like it was completely broken before.

Below, we have this:

```zig
    try stable_log.programInvoke(
        ic.tc,
        program_id,
        ic.tc.instruction_stack.len,
    );
```

How did this work at all for bpf programs when program_id was the bpf loader program?

src/runtime/transaction_execution.zig
```diff
@@ -67,7 +70,10 @@ pub const TransactionExecutionEnvironment = struct {
     rent_collector: *const RentCollector,
     blockhash_queue: *const BlockhashQueue,
     epoch_stakes: *const EpochStakes,
+    vm_environment: *const VmEnvironment,
+    next_vm_environment: ?*const VmEnvironment,
```
dnut: Do you see any downsides to this approach?

```suggestion
    vm_environment: *const vm.Environment,
    next_vm_environment: ?*const vm.Environment,
```

I think using a consistent name for this struct will help clarity.

src/runtime/program_loader.zig
```diff
@@ -0,0 +1,265 @@
...
+
+    for (accounts.keys(), accounts.values()) |pubkey, account| {
+        if (!account.executable) continue;
+        try programs.put(allocator, pubkey, try loadProgram(
+            allocator,
+            &account,
+            accounts,
+            enviroment,
+            slot,
+        ));
```
dnut: Not important, but I kinda prefer it this way. What do you think?
```suggestion
        try programs.put(
            allocator,
            pubkey,
            try loadProgram(allocator, &account, accounts, enviroment, slot),
        );
```

Sobeston (dismissed):

src/runtime/program_loader.zig
```diff
@@ -0,0 +1,265 @@
...
+    allocator: std.mem.Allocator,
+    accounts: *const std.AutoArrayHashMapUnmanaged(Pubkey, AccountSharedData),
+    enviroment: *const VmEnvironment,
+    slot: u64,
+) error{OutOfMemory}!ProgramMap {
+    var programs = ProgramMap{};
+    errdefer programs.deinit(allocator);
+
+    for (accounts.keys(), accounts.values()) |pubkey, account| {
+        if (!account.executable) continue;
```
dnut, line 44: Is this the same way agave decides whether to load a program? It seems fine, but I wonder if there's a leaner approach that can distinguish whether the program will actually be executed.

There are definitely going to be times when an executable program id is included in a transaction where it would never be invoked. For example, sometimes the program id is just an input for another program to write that program id into an account, and it has no CPIs.

It seems easy enough to identify whether a program is being invoked as a top-level instruction. But I'm not sure if there's a way to detect whether the program will be invoked via cpi. Do you think that could be possible without executing the transaction?

src/runtime/program_loader.zig
```diff
@@ -0,0 +1,265 @@
...
+            accounts,
+            enviroment,
+            slot,
+        ));
+    }
+
+    return programs;
+}
+
+pub fn loadProgram(
```
dnut, line 62: This function assumes that the account is executable. It's fine because the calling function is already checking this. But this could potentially return a LoadedProgram for an account that's not actually executable. Granted,, I'm pretty sure the program would be all zeros in that scenario since the owner needs to be a bpf loader program, and you can't change the owner without zeroing out the account. But still there is this little loophole.

I would normally expect this function to check if the account is actually executable and return an error if not. I get that this is redundant with loadPrograms, so maybe it's just a fair assumption to expect the caller to have verified this already. What do you think about adding a comment to indicate this?

```suggestion
/// This function assumes the account is executable and does not check the executable flag.
/// Verify that the account is executable before calling this function.
pub fn loadProgram(
```

Or maybe it's cheap enough to check twice.

src/runtime/program_loader.zig
```diff
@@ -0,0 +1,265 @@
...
+        &environment.loader,
+        environment.config,
+    ) catch {
+        allocator.free(source);
+        return .failed;
+    };
+    errdefer executable.deinit(allocator);
+
+    executable.verify(&environment.loader) catch return .failed;
+
```
dnut: The errdefer will not be triggered because errors are being replaced with an enum. I think you need a similar catch block to above,
```suggestion

    executable.verify(&environment.loader) catch {
        executable.deinit(allocator);
        return .failed;
    };

```

Alternatively you could return error.Failed to avoid this footgun, and convert error.Failed to LoadedProgram.failed in the calling function.

src/runtime/program_loader.zig
```diff
@@ -0,0 +1,265 @@
...
+pub fn loadDeploymentSlotAndExecutableBytes(
+    allocator: std.mem.Allocator,
+    account: *const AccountSharedData,
+    accounts: *const std.AutoArrayHashMapUnmanaged(Pubkey, AccountSharedData),
+) ?struct { u64, []const u8 } {
+    if (account.owner.equals(&bpf_loader.v1.ID) or account.owner.equals(&bpf_loader.v2.ID)) {
+        return .{ 0, account.data };
+    } else if (account.owner.equals(&bpf_loader.v3.ID)) {
+        const program_state = sig.bincode.readFromSlice(
```
dnut, line 108: This leaks if you return null in any of the positions below. You might want to use errdefer and return an error instead of null.

src/runtime/program_loader.zig
```diff
@@ -0,0 +1,265 @@
...
+        if (program_data_account.data.len < bpf_loader.v3.State.PROGRAM_DATA_METADATA_SIZE) {
+            return null;
+        }
+
+        const program_metadata_bytes =
+            program_data_account.data[0..bpf_loader.v3.State.PROGRAM_DATA_METADATA_SIZE];
+        const program_elf_bytes =
+            program_data_account.data[bpf_loader.v3.State.PROGRAM_DATA_METADATA_SIZE..];
+
+        const program_metadata = sig.bincode.readFromSlice(
```
dnut, line 132: same as last comment

src/vm/interpreter.zig
```diff
@@ -19,12 +20,13 @@ pub const Vm = struct {
...
     registers: RegisterMap,
     memory_map: MemoryMap,
-    loader: *const BuiltinProgram,
+    loader: *const Registry(Syscall),
 
     vm_addr: u64,
     call_frames: std.ArrayListUnmanaged(CallFrame),
     depth: u64,
     instruction_count: u64,
+    prev_instruction_meter: u64 = 0,
```
dnut: What's the difference between this and TransactionContext.compute_meter? _(the next comment is also related to this)_

src/vm/interpreter.zig
```diff
@@ -90,10 +93,33 @@ pub const Vm = struct {
         return .{ self.result, instruction_count };
     }
 
+    fn dispatchSyscall(self: *Vm, entry: anytype) !void {
+        self.instruction_count = self.prev_instruction_meter - self.instruction_count;
+        if (self.executable.config.enable_instruction_meter) {
+            self.transaction_context.consumeUnchecked(
+                self.prev_instruction_meter - self.instruction_count,
+            );
+        }
```
dnut: This logic is pretty confusing to me. Can you explain what this is supposed to be doing? For one, I don't really understand the meaning of all these three different ways of tracking instructions. Can you clarify the distinct purpose of each?

Also, the math seems redundant. Once you do this:
```zig
self.instruction_count = self.prev_instruction_meter - self.instruction_count
```
That means that this expression:
```zig
self.prev_instruction_meter - self.instruction_count
```
is equal to the previous instruction_count. Is that intended?

In other words, assuming there is no overflow, I think your code is equivalent to this:

```zig
        if (self.executable.config.enable_instruction_meter) {
            self.transaction_context.consumeUnchecked(
                self.instruction_count,
            );
        }
        self.instruction_count = self.prev_instruction_meter - self.instruction_count;
```

What's the purpose of the flip-flopping of instruction count? For example lets say prev_instruction_meter is 5 and instruction_count is 3. Each time this function is called, it will alternating instruction_count back and forth between 2 and 3. I realize these numbers will change over time, so it won't be quite that simple, but there still seems to be this flip-flopping effect where instruction_count reflects an alternating portion of prev_instruction_meter, which doesn't seem very meaningful. What am I missing here?

dnut (changes requested):
I like how simple and effective this solution is.

Everything looks good except I think you need to fix error handling in a couple places, and I could use some clarification on the changes to instruction counting.

Otherwise, the stylistic comments are just suggestions - feel free to decline those if you disagree.

src/runtime/transaction_execution.zig
yewman: as in package current and next vm in a single struct?

src/runtime/executor.zig
yewman: it was broken for bpf v3 programs, without a program map we would have needed to introduce a bunch of additional logic to pass the program_data account through to this stage. Instead previous tests just included the v3 elf in the program account within the transaction context similar to v1 and v2 programs.

src/runtime/program_loader.zig
yewman: 644c3e97

src/runtime/program_loader.zig
yewman: Agave checks program owners which is a better approach. Perhaps there is a way to determine whether a program is invoked but at this point I think the complexity would outweigh a potential performance improvement. Also the complexity might actually be worse than just accepting some programs will be loaded which are not invoked
644c3e97

src/runtime/program_loader.zig
yewman: `loadDeploymentSlotAndExecutableBytes` will fail if the account owner is not a bpf program which is sufficient. Also makes more sense with the updated conditional in `loadPrograms`

src/runtime/program_loader.zig
yewman: Good catch thank you, I have put the deinit in the verify catch which is okay imo. Let me know if you'd still prefer me to change to using an error.

yewman: 644c3e97

src/runtime/program_loader.zig
yewman: 644c3e97

src/runtime/program_loader.zig
yewman: This is misleading since there are actually no allocations under the hood. To make this more clear I have implemented a deserialise method on the `State` which does not take an allocator. Perhaps a `bincode.readFromSliceNoAlloc` which uses a failing allocator would be a more general solution to this problem but I personally like explicit deserialisers anyway. Lmk if you'd prefer the bincode solution

src/runtime/program_loader.zig
yewman: ^ 644c3e97

src/runtime/program_loader.zig
yewman: 644c3e97

src/runtime/transaction_execution.zig
dnut: No I'm not talking about changing any structs. I'm just suggesting to refer to this struct using its actual name, and not importing it with a different name than it has where it's defined. You can add an import `const vm = ...` to the top of this file and then apply my suggestion.

yewman: c2c9b6ff

src/runtime/program_loader.zig
dnut: I would pass the failing allocator in here at this call site. You don't need to make any changes to bincode.

I don't agree with re-implementing the bincode serialization protocol from scratch for each struct. I don't see how that has any advantage over using the existing bincode implementation. It just seems like more code to write and maintain. Would you also argue against using `std.json`?

src/runtime/program_loader.zig
dnut: 14a88891ed513b1de679af75d63b8b62b8b6dfc8

src/runtime/program_loader.zig
dnut: 14a88891ed513b1de679af75d63b8b62b8b6dfc8

src/runtime/program_loader.zig
dnut: I think it would be cleaner to use an error here but it's not important. This is fine

src/vm/interpreter.zig
dnut: 14a88891ed513b1de679af75d63b8b62b8b6dfc8

src/runtime/program_loader.zig
yewman: yeah fair enough 7320f2b5

dnut (approved):

src/runtime/program/bpf/tests.zig
```diff
@@ -25,39 +113,45 @@ test "hello_world" {
...
-            .owner = program.bpf_loader.v3.ID,
-            .executable = true,
-            .rent_epoch = 0,
-            .data = program_bytes,
-        },
+    const feature_params = &[_]FeatureParams{
+        .{ .pubkey = sig.runtime.features.ENABLE_SBPF_V3_DEPLOYMENT_AND_EXECUTION },
     };
 
+    const program_account, const environment, const program_map = try prepareBpfV3Test(
```
kprotty: nit: `var program_map` to avoid copy in defer

src/runtime/program_loader.zig
```diff
@@ -0,0 +1,315 @@
...
+pub const ProgramMap = std.AutoArrayHashMapUnmanaged(Pubkey, LoadedProgram);
+
+pub const LoadedProgram = union(enum(u8)) {
+    failed,
+    loaded: struct {
+        executable: Executable,
+        source: []const u8,
+    },
+
+    pub fn deinit(self: LoadedProgram, allocator: std.mem.Allocator) void {
```
kprotty: would it make sense to `deinit(*Self)`?

src/vm/syscalls/lib.zig
```diff
@@ -915,7 +638,7 @@ fn callProgramAddressSyscall(
     program_id: Pubkey,
     overlap_outputs: bool,
 ) !struct { Pubkey, u8 } {
-    comptime std.debug.assert(builtin.is_test);
+    comptime std.debug.assert(@import("builtin").is_test);
```
kprotty: style: make import again

kprotty (approved):
good to merge, just some optional style nits

dnut (approved):

kprotty (approved):

PR #787, opened 2025-06-16
feat(ledger,consensus): Add incremental setters to ledger writer

InKryption:
Also some small fork choice refactors

### DIFF
--- a/src/ledger/result_writer.zig
+++ b/src/ledger/result_writer.zig
@@ -31,14 +31,12 @@ const schema = ledger.schema.schema;
 /// or reaching consensus on a block.
 pub const LedgerResultWriter = struct {
     allocator: Allocator,
-    logger: ScopedLogger(@typeName(Self)),
+    logger: ScopedLogger(@typeName(LedgerResultWriter)),
     db: BlockstoreDB,
     // TODO: change naming to 'highest_slot_cleaned'
     lowest_cleanup_slot: *RwMux(Slot),
     max_root: *std.atomic.Value(Slot),
     scan_and_fix_roots_metrics: ScanAndFixRootsMetrics,
-
-    const Self = @This();
 
     pub fn init(
         allocator: Allocator,
@@ -50,7 +48,7 @@ pub const LedgerResultWriter = struct {
     ) !LedgerResultWriter {
         return .{
             .allocator = allocator,
-            .logger = logger.withScope(@typeName(Self)),
+            .logger = logger.withScope(@typeName(LedgerResultWriter)),
             .db = db,
             .lowest_cleanup_slot = lowest_cleanup_slot,
             .max_root = max_root,
@@ -60,7 +58,7 @@ pub const LedgerResultWriter = struct {
 
     /// agave: write_transaction_status
     pub fn writeTransactionStatus(
-        self: *Self,
+        self: *LedgerResultWriter,
         slot: Slot,
         signature: Signature,
         writeable_keys: ArrayList(Pubkey),
@@ -87,7 +85,7 @@ pub const LedgerResultWriter = struct {
 
     /// agave: insert_bank_hash
     pub fn insertBankHash(
-        self: *Self,
+        self: *LedgerResultWriter,
         slot: Slot,
         frozen_hash: Hash,
         is_duplicate_confirmed: bool,
@@ -109,7 +107,7 @@ pub const LedgerResultWriter = struct {
 
     /// agave: set_duplicate_confirmed_slots_and_hashes
     pub fn setDuplicateConfirmedSlotsAndHashes(
-        self: *Self,
+        self: *LedgerResultWriter,
         duplicate_confirmed_slot_hashes: []const struct { Slot, Hash },
     ) !void {
         var setter = try self.setDuplicateConfirmedSlotsAndHashesIncremental();
@@ -126,7 +124,7 @@ pub const LedgerResultWriter = struct {
 
     /// Returns a struct which can be used to enact the same operation as `setDuplicateConfirmedSlotsAndHashes`, incrementally.
     pub fn setDuplicateConfirmedSlotsAndHashesIncremental(
-        self: *Self,
+        self: *LedgerResultWriter,
     ) !SetDuplicateConfirmedSlotsAndHashesIncremental {
         return .{
             .result_writer = self,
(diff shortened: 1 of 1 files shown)


src/ledger/result_writer.zig
```diff
@@ -110,58 +110,143 @@ pub const LedgerResultWriter = struct {
...
-            const slot, const frozen_hash = slot_hash;
-            const data = FrozenHashVersioned{ .current = FrozenHashStatus{
+            const slot, const hash = slot_hash;
+            try setter.addSlotAndHash(slot, hash);
+        }
+
+        try setter.commit();
+    }
+
+    /// Returns a struct which can be used to enact the same operation as `setDuplicateConfirmedSlotsAndHashes`, incrementally.
```
dadepo, line 125: Nit: Is this comment correct? or `setDuplicateConfirmedSlotsAndHashes` is meant to mean Agave's `set_duplicate_confirmed_slots_and_hashes`? Because if not, then it is confusing as Sig `setDuplicateConfirmedSlotsAndHashes` makes use of it and hence `setDuplicateConfirmedSlotsAndHashes` in Sig is incremental

InKryption: I'm not sure what you mean exactly, "incrementally" is referring to the fact that it doesn't take a slice parameter, and instead returns a struct that allows you to do the operation multiple times before committing it.

src/ledger/result_writer.zig
```diff
@@ -110,58 +110,143 @@ pub const LedgerResultWriter = struct {
...
                 .frozen_hash = frozen_hash,
                 .is_duplicate_confirmed = true,
             } };
-            try write_batch.put(schema.bank_hash, slot, data);
+            try self.write_batch.put(schema.bank_hash, slot, data);
+        }
+
+        pub fn cancel(self: *SetDuplicateConfirmedSlotsAndHashesIncremental) void {
+            std.debug.assert(!self.is_committed);
+            self.is_committed = true;
```
dadepo: Should `is_committed` be set to true on cancel?  especially given it is called in an  `errdefer`.

InKryption: It has to be, otherwise the assertion in the deinit will be tripped on the error path. That's why this function exists to be errdefer'd.

src/ledger/result_writer.zig
dadepo: > to enact the same operation as `setDuplicateConfirmedSlotsAndHashes`, incrementally

Unless I read this wrong, but the above alludes to  `setDuplicateConfirmedSlotsAndHashes` not being an incremental operation...and this comment says this method makes it possible to do same thing as `setDuplicateConfirmedSlotsAndHashes` but incrementally.

But it looks like `setDuplicateConfirmedSlotsAndHashes` is incremental because it's implementation depends on `setDuplicateConfirmedSlotsAndHashesIncremental`

InKryption: `setDuplicateConfirmedSlotsAndHashes` is a single function call, which takes a parameter with all the duplicate confirmed slothashes to set in one go. It is implemented in terms of its incremental counterpart, but it does not allow the caller to incrementally set the slothashes, ergo it is not itself incremental.

dadepo: You are right. I see where I got myself confused by the comments.

src/ledger/result_writer.zig
dadepo: is there a scenario where `cancel` can be called outside of `errdefer`?   Because then, it is not obvious why a `cancel` call sets `is_committed` to true?

src/ledger/result_writer.zig
```diff
@@ -110,58 +110,143 @@ pub const LedgerResultWriter = struct {
...
+        self: *Self,
+    ) !SetDuplicateConfirmedSlotsAndHashesIncremental {
+        return .{
+            .result_writer = self,
+            .write_batch = try self.db.initWriteBatch(),
+            .is_committed = false,
+        };
+    }
+    pub const SetDuplicateConfirmedSlotsAndHashesIncremental = struct {
+        result_writer: *Self,
```
dadepo: Any reason against this:

```suggestion
        result_writer: *LedgerResultWriter,
```

Was initially confused by this: having the `Self` in a struct, refer to an outer `Self` definition that points to another struct.

src/ledger/result_writer.zig
```diff
@@ -110,58 +110,143 @@ pub const LedgerResultWriter = struct {
...
+        defer setter.deinit();
+        errdefer setter.cancel();
+        for (rooted_slots) |rooted_slot| try setter.addRoot(rooted_slot);
+        try setter.commit();
+    }
 
-        try self.db.commit(&write_batch);
-        _ = self.max_root.fetchMax(max_new_rooted_slot, .monotonic);
+    /// Returns a struct which can be used to enact the same operation as `setRoots`, incrementally.
+    pub fn setRootsIncremental(self: *Self) !SetRootsIncremental {
```
dadepo: What about:?

```suggestion
    pub fn setRootsIncremental(self: *LedgerResultWriter) !SetRootsIncremental {
```

Given the recent convention is to use the structs name. It also makes it a tad clearer what `.result_writer` is being set to.

src/ledger/result_writer.zig
InKryption: The purpose of it setting `is_committed` to true is to avoid setting off the assertion in `deinit`, which asserts `is_committed` is true.

dadepo: If that is it's sole purpose, is it possible to have it inlined as a `errdefer` block instead? As i think it can be confusing to have the method called `cancel` which set `is_committed` to true, with nothing stopping it from being called outside of the `errdefer` path.

InKryption: I mean, by that same token, nothing stops anything else from setting `is_committed` to true outside the error path, the idea is that nothing has to touch the internals in order to use it properly, they simply have to interact with the API.

InKryption: I went ahead and elaborated further documentation for these items, does this resolve your doubts? 67b3ec99e92caf85b87ce1972b0f4133f6fa3ffc

src/ledger/result_writer.zig
InKryption: f9ee800214ea46aee9349fb789bd70d92659de32

src/ledger/result_writer.zig
dadepo: Yup. The rename + the comments is definitely an improvement.

dadepo (approved):

src/ledger/result_writer.zig
InKryption: f9ee800214ea46aee9349fb789bd70d92659de32

PR #794, opened 2025-06-25
fix(vm): more bpf v3 program conformance 

kprotty:
Address more `bpf-loader-v3-programs` fixture conformance, excluding mostly those missing syscalls


src/vm/elf.zig
```diff
@@ -943,14 +988,35 @@ pub const Elf = struct {
         bytes: []u8,
         headers: Headers,
         data: Data,
-        sbpf_version: sbpf.Version,
+        real_sbpf_version: sbpf.Version,
```
yewman: Could you add a comment on why there is a distinction between `real_sbpf_version` and an `sbf_version` in this function?

kprotty: https://github.com/Syndica/sig/pull/794/commits/5b2c6ce61b45474dfce8b2925bf3aec0dc7baf97

PR #795, opened 2025-06-25
feat(replay): execute and commit transactions

dnut:
- call the svm using SvmGateway
- commit transaction results using Committer

I squashed this branch to declutter the github PR. The detailed commit history still exists on branch [dnut/replay/commit-bak](https://github.com/Syndica/sig/tree/dnut/replay/commit-bak)

### DIFF
--- a/conformance/src/bank_methods.zig
+++ b/conformance/src/bank_methods.zig
@@ -167,7 +167,7 @@ fn applyBuiltinProgramFeatureTransitions(
         const feature_id = precompile.required_feature orelse continue;
         if (!feature_set.active(feature_id, 0)) continue;
 
-        const maybe_account = accounts_db.getAccount(
+        const maybe_account = accounts_db.getAccountLatest(
             &precompile.program_id,
         ) catch |err| switch (err) {
             error.PubkeyNotInIndex => null,


src/core/stake.zig
```diff
@@ -162,21 +170,38 @@ pub fn Stakes(comptime stakes_type: StakesType) type {
...
-            for (self.stake_delegations.keys(), self.stake_delegations.values()) |key, value|
-                try stake_delegations.put(
-                    allocator,
-                    key,
-                    if (is_account_type) value.clone(allocator) else value,
-                );
+            for (self.stake_delegations.keys(), self.stake_delegations.values()) |key, value| {
+                const new_value =
+                    if (stakes_type == .account and output_type == .account)
+                        try value.clone(allocator)
```
Sobeston, line 196: risk of a leak?

src/core/status_cache.zig
```diff
@@ -15,73 +15,85 @@ const T = ?sig.ledger.transaction_status.TransactionError;
...
+    };
 
     pub fn deinit(self: *StatusCache, allocator: std.mem.Allocator) void {
-        self.roots.deinit(allocator);
+        var roots = self.roots.tryWrite() orelse unreachable;
+        var cache = self.cache.tryWrite() orelse unreachable;
+        var slot_deltas = self.slot_deltas.tryWrite() orelse unreachable;
+        defer roots.unlock();
+        defer cache.unlock();
+        defer slot_deltas.unlock();
```
Sobeston: not a fan of the orelse unreachables here - would it be better to warn and return instead?

src/replay/confirm_slot.zig
```diff
@@ -196,11 +235,27 @@ pub const ConfirmSlotFuture = struct {
...
+            self.entries,
+            self.allocator,
+            self.initial_hash,
+            .{ .exit = self.exit },
+        ) catch |e| {
+            if (e != error.Exit) {
+                self.logger.err().logf("poh verification failed with error: {}", .{e});
+            }
+            self.exit.store(true, .monotonic);
+            return e;
```
Sobeston, line 254: just a nit, error.Exit doesn't sit that well with me?

src/runtime/transaction_execution.zig
```diff
@@ -122,14 +122,21 @@ pub const TransactionRollbacks = union(enum(u8)) {
...
 
+    pub fn accounts(self: *const TransactionRollbacks) []const CopiedAccount {
+        switch (self.*) {
+            .fee_payer_only, .same_nonce_and_fee_payer => |*item| {
+                return @as([*]const CopiedAccount, @ptrCast(item))[0..1];
+            },
+            .separate_nonce_and_fee_payer => |*items| return items,
+        }
+    }
+
```
Sobeston: Can clean it up to
```zig
    pub fn accounts(self: *const TransactionRollbacks) []const CopiedAccount {
        return switch (self.*) {
            .fee_payer_only, .same_nonce_and_fee_payer => |*item| item[0..1],
            .separate_nonce_and_fee_payer => |*items| items,
        };
    }

File: src/cmd.zig [hunk 5]
```diff
@@ -1121,22 +1130,28 @@ fn validator(
...
 
-    var epoch_context_manager = try sig.adapter.EpochContextManager.init(
-        allocator,
-        epoch_schedule,
-    );
-    try epoch_context_manager.put(epoch, .{
-        .staked_nodes = try staked_nodes.clone(allocator),
-        .leader_schedule = try LeaderSchedule.fromStakedNodes(
+    const staked_nodes = try collapsed_manifest.getStakedNodes(allocator, epoch);
+    var epoch_context_manager = try sig.adapter.EpochContextManager.init(allocator, epoch_schedule);
```
Sobeston: leak risk?

src/accountsdb/db.zig
```diff
@@ -1799,14 +1800,14 @@ pub const AccountsDB = struct {
         }
     }
 
-    pub const GetAccountError = GetFileFromRefError || error{PubkeyNotInIndex};
-    /// gets an account given an associated pubkey. mut ref is required for locks.
-    pub fn getAccount(
+    /// gets the latest version of an account at the provided address.
+    /// mut ref is required for locks.
+    pub fn getAccountLatest(
```
yewman, line 1811: Could this result in reading an account from a competing fork?

src/core/status_cache.zig
yewman: I am not a fan of unreachable here, but don't like the idea of warn and return either. I would prefer an explicit panic since the path is reachable.

src/core/stake.zig
yewman: yeah we will need an errdefer for this branch

src/core/bank.zig
```diff
@@ -176,38 +175,48 @@ pub const SlotState = struct {
...
-        .accounts_lt_hash = .init(.IDENTITY),
-        .collected_transaction_fees = .init(0),
-        .collected_priority_fees = .init(0),
-    };
-
     pub fn deinit(self: *SlotState, allocator: Allocator) void {
-        var blockhash_queue = self.blockhash_queue.write();
+        self.stakes_cache.deinit(allocator);
+
+        var blockhash_queue = self.blockhash_queue.tryWrite() orelse unreachable;
```
yewman: why not just use a panic here?

src/replay/confirm_slot.zig
```diff
@@ -130,30 +153,47 @@ pub const ConfirmSlotFuture = struct {
...
         return future;
     }
 
     pub fn destroy(self: *ConfirmSlotFuture, allocator: Allocator) void {
+        self.exit.store(true, .monotonic);
+        const exited_scheduler = self.scheduler.thread_pool.joinForDeinit(.fromMillis(100));
+        const exited_poh = self.poh_verifier.joinForDeinit(.fromMillis(100));
+        if (!exited_scheduler or !exited_poh) {
+            @panic("Failed to deinit ConfirmSlotFuture due to hanging threads.");
+        }
```
yewman, line 196: I like this approach and think it could be used instead of the `unreachable`'s in the deinit methods above.

src/replay/confirm_slot.zig
```diff
@@ -595,3 +691,123 @@ pub fn testAwait(future: anytype) !@TypeOf(future.poll()) {
...
+            .status_cache = &self.status_cache,
+            .stakes_cache = &self.stakes_cache,
+            .new_rate_activation_epoch = null,
+        };
+    }
+
+    /// This makes it so the svm will not return an error for these
+    /// transactions. It doesn't mean the transactions themselves will succeed,
+    /// but they will at least execute as if they are legally allowed in the
+    /// block.
```
yewman: Could you add a permalink to where this ensures transaction execution will reach? I am a little confused about what you mean by `will at least execute`

src/replay/execution.zig
```diff
@@ -211,21 +216,76 @@ fn replaySlot(state: *ReplayExecutionState, slot: Slot) !ReplaySlotStatus {
...
+            state.allocator.free(entries);
+        }
+        state.logger.info().logf("got {} entries for slot {}", .{ entries.len, slot });
 
-    confirmation_progress.num_shreds += num_shreds;
-    confirmation_progress.num_entries += entries.len;
-    for (entries) |e| confirmation_progress.num_txs += e.transactions.len;
+        if (entries.len == 0) {
+            for (entries) |entry| entry.deinit(state.allocator);
+            state.allocator.free(entries);
```
yewman: We don't need these deinits when entries.len == 0

src/replay/svm_gateway.zig
```diff
@@ -0,0 +1,192 @@
...
+        ancestors: *const Ancestors,
+        feature_set: FeatureSet,
+        rent_collector: *const RentCollector,
+        epoch_stakes: *const sig.core.EpochStakes,
+        status_cache: *StatusCache,
+    };
+
+    pub fn init(
+        allocator: Allocator,
+        batch: []const replay.resolve_lookup.ResolvedTransaction,
```
yewman:
```suggestion
        batch: []const ResolvedTransaction,
```

src/replay/svm_gateway.zig
```diff
@@ -0,0 +1,192 @@
...
+
+        const vm_environment = try vm.Environment.initV1(
+            allocator,
+            &params.feature_set,
+            // This does not actually set the compute budget. it's only used to
+            // set that max call depth and stack frame size. the actual compute
+            // budgets are determined per transaction.
+            &ComputeBudget.default(1_400_000),
+            false,
+            true, // TODO: should this be false?
```
yewman: Yes it should be false, see https://github.com/firedancer-io/agave/blob/10fe1eb29aac9c236fd72d08ae60a3ef61ee8353/runtime/src/bank.rs#L4950
To prevent deployment of broken or non conformant elfs, when a program is deployed by the bpf loader https://github.com/Syndica/sig/blob/396ae1904e9ecb1c2a9cb175d1f96fa3cbb472bf/src/runtime/program/bpf_loader/execute.zig#L1515-L1538 a special environment is created which has this set to true and also ommits syscalls which are no longer supported for deployment

src/replay/update_sysvar.zig
```diff
@@ -367,13 +425,13 @@ fn nextClock(
     feature_set: *const FeatureSet,
     epoch_schedule: *const EpochSchedule,
     stakes_cache: *StakesCache,
-    epoch_stakes_map: *const EpochStakesMap,
+    maybe_epoch_stakes: ?*const EpochStakes,
```
yewman, line 429: why make this nullable?

src/runtime/account_loader.zig
```diff
@@ -47,6 +47,10 @@ pub const LoadedTransactionAccounts = struct {
 pub const CachedAccount = struct {
     pubkey: Pubkey,
     account: *AccountSharedData,
+
```
yewman, line 50: Cached and Copied account are kind of related, I think it would make sense to define them next to each other. What are your thoughts?

src/runtime/account_loader.zig
```diff
@@ -174,16 +197,15 @@ pub const BatchAccountCache = struct {
...
 
-        // load v3 loader's ProgramData accounts.
-        for (transactions) |tx| {
-            for (tx.instruction_infos) |instr| {
-                const program_key = tx.accounts.items(
+        { // load v3 loader's ProgramData accounts.
+            for (instructions) |instr| {
+                const program_key = accounts.items(
                     .pubkey,
                 )[instr.program_meta.index_in_transaction];
```
yewman:
```suggestion
                const program_key = instr.program_meta.pubkey;
```

src/accountsdb/db.zig
dnut: yes, basically

This method is not fork-aware. There's no concept of being on a particular fork. It just gets whatever is the most recent version. So if you're on a particular fork, and you expect to get an account from the same fork you're on, using this method is wrong because you might get an account from a competing fork.

src/replay/execution.zig
dnut: Sure, but that case is being handled correctly, isnt it? Are you suggesting a change?

src/replay/svm_gateway.zig
dnut: I saw this code, but they actually create *two* environments (v1 and v2) in agave, v1 as false and v2 as true. We're just creating a single environment (v1). It feels like this is not going to be consistent with agave unless we support both.

src/replay/update_sysvar.zig
dnut: Well it was already nullable. An empty map is analogous to a null optional. The transaction harness relies on this. I tried to make it non-nullable but it broke conformance.

src/core/status_cache.zig
dnut: In my mind this is analogous to freeing memory in the deinit function. There's an implicit contract that the item is no longer in use. If you free memory that's still in use, that has undefined behavior. If you deinit an internally locked structure that is still in use, that has undefined behavior. It's consistent with the normal approach to deiniting things and I think it's consistent with other uses of assertions/unreachable.

Anyway I'm happy to improve this further.

Both of you are suggesting a solution that includes providing an error message to the user. That sounds good to me.

Beyond that, Sebastian is suggesting to leak and continue, whereas Harold is suggesting to panic (similar to unreachable). I lean towards the panic since the code already has UB at this point.

src/replay/execution.zig
yewman: Well these deinits are inside a block predicated on entries.len == 0

src/accountsdb/db.zig
yewman: Is that the expected behaviour for the current usages? I am worried this could end up causing issues

src/replay/svm_gateway.zig
yewman: Ahh yes, but environment v2 is only used in the context of creating a vm for executing a native program via the vm.invoke_function call (I have no idea why they went with this design, something about a consistent api for program invocation...). Since we do not follow this approach, and only create and use the vm for executing bpf programs, we only need environment v1.

src/replay/update_sysvar.zig
yewman: The fact that an empty map is analogous to a null optional is why I was surprised by this change. Given that they are analogous I am confused how making it non nullable would break conformance

src/core/status_cache.zig
yewman: Let's go with the panic, I am very much against leak and continue. As for `unreachable` vs `panic`, in my opinion unless a block is **strictly** unreachable we should be using panic.

src/replay/update_sysvar.zig
dnut: Uhh what? I don't follow your reasoning at all haha. 

The function already allowed it to be optional. For compatibility, I needed to keep it optional.

I would have liked to make it required, but that would have been an even bigger change. I tried, but it didn't work out because there were already examples in the code that were not passing in the epoch stakes. The transaction harness takes advantage of the fact that this is optional. So I kept it optional

All I did was move the `map.get` call from inside this function to outside the function. `map.get` returns an optional.

src/replay/execution.zig
dnut: Ah ok, I don't see that here in the conversation view on github.

src/replay/update_sysvar.zig
yewman: haha yeah I am with you my bad :)

src/core/status_cache.zig
dnut: > in my opinion unless a block is strictly unreachable we should be using panic

This contrasts with the typical usage of `assert`, which is implemented with `unreachable`. Do you think we should write our own assert function that uses `panic` instead?

src/accountsdb/db.zig
dnut: What usages? I think it's mainly used in tests. Typically, you shouldn't use this function, unless you really just want the absolute latest version of an account, and don't care about forking.

For context, I didn't create this function. It's been around for a long time. I only renamed it because "getAccount" is too general and makes it seem like this is the standard way of getting an account, when it's more of a special case. I'd like to revisit this to make further improvements, but I don't think it's in scope for this PR. At this point, I was just hoping to make a tiny easy incremental improvement.

src/core/stake.zig
dnut: f8cb45b2067982d9c437163cb5eda00e5c3b7b50

src/runtime/transaction_execution.zig
dnut: 98349ba779b42d6600b834ab76f9b604e128a43c

src/core/status_cache.zig
yewman: Yeah, that’s a fair point. From my understanding in optimisation modes Debug and ReleaseSafe, `unreachable` and `panic` both produce the same effect — they panic. If they produce the same effect then I see no reason to use `unreachable` on paths which are reachable as its semantics imply a condition which is not true. I guess from this perspective I seem to disagree with many of its usages in `std`. Is there some difference between `unreachable` and `panic` in ReleaseSafe that I am unaware of? If not it seems like `unreachable` is often used as semantically incorrect shorthand for an explicit `@panic(..)`. 

As for the debug assert I do think an argument could be made for defining our own function that uses a panic. Generally, debug assertions are intended to perform checks in debug builds and be compiled away in release builds. Since we only intend to run in ReleaseSafe using `debug.assert` may mislead readers into thinking that checks within debug assertions have no runtime cost in production which is not true.

src/accountsdb/db.zig
yewman: Yeah that is fair, I was actually thinking we could add a comptime builtin.is_test check. I understand that this falls outside the scope of this PR and I don't want to hold anything up unnecessarily. A comment / todo to warn users would be good for now I think.

src/runtime/account_loader.zig
dnut: i pulled this in from main in 7e97e1a580749c049a6d07d66ea08250f75c9959

src/replay/svm_gateway.zig
dnut: 064e58072186761d7a146216afaf0a17fd71ed64

src/core/bank.zig
dnut: because `unreachable` is a debug assertion and panic is not. but anyway i'm fine with a panic here

dnut: f43efc5b7f433aa0f5b1aaa75d14b8164eb893a5

src/core/status_cache.zig
dnut: f43efc5b7f433aa0f5b1aaa75d14b8164eb893a5

dnut: I agree with some of your points, but I also think it's reasonable to expect people to understand that ReleaseSafe does not optimize away safety checks.

I can definitely see some room for improvement here though regarding the inaccurate terminology and lack of error messages when using `unreachable`.

src/cmd.zig
dnut: I don't understand, where is the leak risk? There's a defer deinit right below this line.

Also I didn't add this in the current PR, not sure why it's showing up in the diff, maybe I shuffled some things around. But I'm happy to fix a leak here if you can explain it

src/accountsdb/db.zig
dnut: 967e37c97903516cd3b5317dc54ac5bab6ab24ae

this function is used outside of tests and we should probably go through all the usages and fix them

src/replay/confirm_slot.zig
dnut: do you have an alternative suggestion? I was hoping verifyPoh would distinguish between successful verification and early exit in its return value

src/replay/confirm_slot.zig
dnut: I don't know where the execution will reach, and I don't think it's relevant in this context. The idea here is to make it so the transactions would be legal for a leader to include in a block. I tried to make the comment clearer.

34b654a47c4d9505b2bc5bb8da1859922e2c60a0

src/replay/execution.zig
dnut: 14e24744d13c93e896afe5ef9ff8be73cb8ce922

src/replay/svm_gateway.zig
dnut: 4dacb7d774b353d4d0dd02dafe7499540c388554

src/runtime/account_loader.zig
dnut: I don't have a strong opinion about it. I think where they're located now is reasonable based on the fact that they are defined for their particular scopes and used in those scopes. I also see your point about them being similar, but it's not clear to me that either of the existing locations would make a lot of sense for the other one to be there, other than the fact that it brings the two related structs closer in proximity.

src/cmd.zig
Sobeston: Not sure if I misread this, or if it changed since I last read it, but this looks fine

src/replay/confirm_slot.zig
Sobeston: Might be missing something, what does an early exit mean in this context?

dnut: it means the `exit: *Atomic(bool)` parameter was set to true and verifyPoh exited before completing the verification

Sobeston (approved):

PR #796, opened 2025-06-26
feat(accountsdb): get with ancestors, and single-account push

Sobeston:


src/accountsdb/db.zig
```diff
@@ -1848,6 +1869,95 @@ pub const AccountsDB = struct {
         }
     }
 
+    // writes one account to storage
+    pub fn putAccount(self: *AccountsDB, account: Account, pubkey: Pubkey, slot: Slot) !void {
```
Sobeston: @yewman was thinking this should take an AccountSharedData. I think it's reasonable, I think this function would only be called from the runtime anyway. Any thoughts?

dnut: That makes sense to me. I have no idea how a caller would be able to construct an instance of `Account`. It seems like it only makes sense when it comes from AccountsDB.

Actually I think `AccountSharedData` should probably be the standard representation of accounts, and `Account` is just a specific thing that's used for reading accounts out of AccountsDB. Likewise I would probably rename `AccountSharedData` to `Account` and move it to `core`, and rename to `Account` to something like `AccountReadHandle` and move it to `accountsdb`.

dnut: Also minor nitpick but I would order the parameters differently to make it consistent with other kv stores, like hashmaps.

```suggestion
    pub fn putAccount(self: *AccountsDB, pubkey: Pubkey, slot: Slot, account: Account) !void {
```

Sobeston: Gone for AccountSharedData + slot as 1st param:
3ea90f21c4fabb99a1320a6b1204fb44f1c9dee5

I think eventually we should be able to merge Account and AccountSharedData, but requires some effort

src/accountsdb/db.zig
```diff
@@ -138,7 +138,10 @@ pub const AccountsDB = struct {
     // TODO: move to Bank struct
     bank_hash_stats: RwMux(BankHashStatsMap),
 
-    pub const PubkeysAndAccounts = struct { []const Pubkey, []const Account };
+    const PubkeyAndAccount = struct { pubkey: Pubkey, account: Account };
+
+    pub const PubkeysAndAccounts = std.MultiArrayList(PubkeyAndAccount);
+    // pub const PubkeysAndAccounts = struct { []const Pubkey, []const Account };
```
yewman: lets remove this guy

src/accountsdb/db.zig
```diff
@@ -1848,6 +1869,113 @@ pub const AccountsDB = struct {
         }
     }
 
+    /// writes one account to storage
+    /// intended for use from runtime
+    pub fn putAccount(
```
yewman, line 1926: Do we need this function or could we just call `putAccountSlice` with slice of length 1? Seems like there is a lot of duplicated logic?

src/accountsdb/db.zig
```diff
@@ -1669,6 +1670,26 @@ pub const AccountsDB = struct {
...
     }
 
+    /// gets an account given an associated pubkey. mut ref is required for locks.
+    /// Will only find rooted accounts, or unrooted accounts from a slot in ancestors.
+    pub fn getAccountWithAncestors(
+        self: *AccountsDB,
+        pubkey: *const Pubkey,
+        ancestors: *const sig.core.Ancestors,
+        /// below this slot, consider totally rooted/finalised (=> ancestors irrelevant)
+        min_ancestors_slot: Slot,
```
yewman: This seems like something we may want to set a default for, rather than requiring the user to provide it?  What are you're thoughts @dnut

src/accountsdb/db.zig
```diff
@@ -1669,6 +1670,26 @@ pub const AccountsDB = struct {
...
+    ) GetAccountError!Account {
```
yewman: I thought we were going with returning AccountSharedData?

src/accountsdb/db.zig
Sobeston: I think that for replay what we'd want to do is pass in the lowest ancestor. But I think it could also be the latest rooted slot?

src/accountsdb/db.zig
dnut: I think we should return Account for any reads from accountsdb to avoid unnecessary allocations and copying of data out of the buffer pool. Callers can do the conversion if/when they want.

src/accountsdb/db.zig
dnut: putAccountSlice is for tests. I thought we discussed this during our call? putAccountSlice assumes the database is empty. We can't use that function. We needed to add a new one that works for a live validator.

Are you arguing that the new one should also take a slice instead of a single account? I don't think it really matters, but I would lean towards a simpler API that only takes one at a time unless there's a compelling reason to batch it.

src/accountsdb/db.zig
dnut: Are you saying this commit changes the order of the parameters? Doesn't look like it

I think I'd rather keep the structs separate to avoid unnecessary complexity/coupling in code that just needs a slice for the data field. But I'm open to discussing combining them if you think there's an advantage of that.

src/accountsdb/db.zig
Sobeston: Yeah there's definitely some duplication here that isn't ideal. Idk, preferring to have some duplication for now than risk making a weird (or slower) function with a generic account

src/accountsdb/db.zig
Sobeston: 448bc1df1fd1ac7cd618a03ce58a78e642b72478

src/accountsdb/db.zig
Sobeston: Yeah separate for now is definitely the move.

I think I misread your previous comment, I changed it from account/pubkey/slot to account/slot/pubkey. 

Thinking it should now be slot/pubkey/account, changed in b4441d16cfd520123c27a22f836690f0fc162b98

src/accountsdb/db.zig
yewman: we are already duplicating the data the the returned Account see:

- For the unrooted map case: https://github.com/Syndica/sig/blob/0c3ccdfe3ccbf52ee578c5c3619beebee1d4742e/src/accountsdb/db.zig#L1499-L1500
- For the file case: https://github.com/Syndica/sig/blob/0c3ccdfe3ccbf52ee578c5c3619beebee1d4742e/src/accountsdb/db.zig#L1576

If we are already duping, we may as well dupe into an AccountSharedData imo. Maybe I am missing something but  I thought we chatted about this right Sebastian?

Sobeston: That method in the file case doesn't duplicate the underlying data, it effectively puts another borrow on the buffer pool read. Maybe I named it badly

> but I thought we chatted about this right Sebastian?

I think we talked about the parameter. The case of the return value is a little different? idk. I don't mind much either way

src/accountsdb/db.zig
yewman: `/// NOTE: only currently used in benchmarks and tests`

I don't think it is 'for' tests, just that at the time of implementation thats all it was used for. If there is zero benefit of `putAccountsSlice` existing then we should remove it in favour of iterating and calling `putAccount` in call sites, otherwise if it needs to exist I don't see the point in having two functions with near identical logic. 

I am happy to move on though for now as these changes are blocking short term goals but I think we should address it at some point.

Feel free to resolve this comment.

src/accountsdb/db.zig
yewman: I think using the largest rooted slot from accounts db makes sense, if down the line we find a use case for specifying it we can create an associated function. Otherwise it adds more dependencies to call sites which don't car or need to know the min ancestors slot, or incorrect/meaningless values may be used

src/accountsdb/db.zig
yewman: Can a caller distinguish between the `Account` returned via the file case vs the unrooted map case? If not then maybe returning an enum of `Account` and `AccountSharedData` for the file and unrooted map case respectively would be good to prevent allocating the data in an account twice if the caller wants to create an `AccountSharedData`

src/accountsdb/db.zig
Sobeston: True. My only issue is that I'm not sure if we need
a) the latest rooted slot
or b) the lowest ancestor slot

I think a is probably fine but b is technically correct? @dnut ?

dnut: I would just go until the root slot.

Technically, the safest approach is the strictest: use ancestors until we hit the limit of its history. This requies you to make the least number of assumptions. But that's overkill. We'd have to add a mechanism to track the edge of its history. It's not too hard but I think it's unnecessary.

Once a slot is rooted, replay prunes any branches that don't contain that slot. It's safe to assume that replay won't ask for an Account on a fork that conflicts with the root.

Maybe you can make a comment that this assumption is being made here?

How does agave do it?

src/accountsdb/db.zig
dnut: I thought we decided to just add the function we need and not worry about refactoring existing tests to use the new code until later. It was just a time saving measure. I agree ideally we can consolidate these functions somehow

src/accountsdb/db.zig
dnut: `Account.data` is already a union that can contain a buffer pool handle, an owned slice, or a borrowed slice. I think adding this extra layer would be redundant.

src/accountsdb/db.zig
Sobeston: Hm its usage varies a bit. What function in agave do you have in mind?

dnut: `AccountsDB.load`

After navigating through many layers of function calls deep from `load`, you eventually run into the `latest_slot` function: https://github.com/anza-xyz/agave/blob/b948b97d2a08850f56146074c0be9727202ceeff/accounts-db/src/accounts_index.rs#L1024-L1063

Note where it says "the slot `S` is an ancestor or root"

The way this function seems to work is that it iterates over all accounts, looking for one with a slot in ancestors. If none of them are in ancestors, it iterates from the beginning again until it finds a slot that's `<= max_root` and contained in the set of stored roots.

```rust
    // Given a SlotList `L`, a list of ancestors and a maximum slot, find the latest element
    // in `L`, where the slot `S` is an ancestor or root, and if `S` is a root, then `S <= max_root`
    pub(crate) fn latest_slot(
        &self,
        ancestors: Option<&Ancestors>,
        slot_list: &[(Slot, T)],
        max_root_inclusive: Option<Slot>,
    ) -> Option<usize> {
        let mut current_max = 0;
        let mut rv = None;
        if let Some(ancestors) = ancestors {
            if !ancestors.is_empty() {
                for (i, (slot, _t)) in slot_list.iter().rev().enumerate() {
                    if (rv.is_none() || *slot > current_max) && ancestors.contains_key(slot) {
                        rv = Some(i);
                        current_max = *slot;
                    }
                }
            }
        }

        let max_root_inclusive = max_root_inclusive.unwrap_or(Slot::MAX);
        let mut tracker = None;

        for (i, (slot, _t)) in slot_list.iter().rev().enumerate() {
            if (rv.is_none() || *slot > current_max) && *slot <= max_root_inclusive {
                let lock = match tracker {
                    Some(inner) => inner,
                    None => self.roots_tracker.read().unwrap(),
                };
                if lock.alive_roots.contains(slot) {
                    rv = Some(i);
                    current_max = *slot;
                }
                tracker = Some(lock);
            }
        }

        rv.map(|index| slot_list.len() - 1 - index)
    }
```

dnut: 8adfa685cff1259453f9d963586325e945af5b57

dnut: I went with an approach similar to agave.

src/accountsdb/db.zig
```diff
@@ -1848,6 +1910,259 @@ pub const AccountsDB = struct {
...
+
+        if (self.geyser_writer) |geyser_writer| {
+            const data_versioned = sig.geyser.core.VersionedAccountPayload{
+                .AccountPayloadV1 = .{
+                    .accounts = &.{duplicated},
+                    .pubkeys = &.{pubkey},
+                    .slot = slot,
+                },
+            };
+            try geyser_writer.writePayloadToPipe(data_versioned);
```
InKryption:
```suggestion
            try geyser_writer.writePayloadToPipe(.{
                .AccountPayloadV1 = .{
                    .accounts = &.{duplicated},
                    .pubkeys = &.{pubkey},
                    .slot = slot,
                },
            });
```

src/accountsdb/db.zig
```diff
@@ -753,21 +766,45 @@ pub const AccountsDB = struct {
...
+            try reference_bufs.append(references_buf);
+            try global_indices.append(ref_global_index);
+
             // index the account file
-            var slot_references = std.ArrayListUnmanaged(AccountRef).initBuffer(
-                references_buf[n_accounts_total..],
-            );
+            var slot_references = AccountIndex.SlotRefMapValue{
+                .global_index = ref_global_index,
+                .refs = std.ArrayListUnmanaged(AccountRef).initBuffer(references_buf),
```
InKryption:
```suggestion
                .refs = .initBuffer(references_buf),
```

src/accountsdb/db.zig
```diff
@@ -1848,6 +1910,259 @@ pub const AccountsDB = struct {
...
+
+            const min_slot = if (slot == 0) null else slot - 1;
+            const ref = slotListMaxWithinBounds(head_ref.ref_ptr, min_slot, slot) orelse
+                break :search_and_overwrite;
+
+            if (ref.location != .UnrootedMap) {
+                return error.CannotWriteRootedSlot;
+            }
+
+            const index = ref.location.UnrootedMap.index;
```
InKryption:
```suggestion
            const index = switch (ref.location) {
                .UnrootedMap => |index| index,
                else => return error.CannotWriteRootedSlot,
            };
```

src/accountsdb/db.zig
```diff
@@ -1848,6 +1910,259 @@ pub const AccountsDB = struct {
...
+        }
+
+        {
+            const unrooted_accounts, var unrooted_accounts_lg =
+                self.unrooted_accounts.writeWithLock();
+            defer unrooted_accounts_lg.unlock();
+
+            const entry = try unrooted_accounts.getOrPut(slot);
+
+            if (!entry.found_existing) entry.value_ptr.* = .{};
```
InKryption:
```suggestion
            if (!entry.found_existing) entry.value_ptr.* = .empty;
```

src/accountsdb/db.zig
```diff
@@ -1848,6 +1910,259 @@ pub const AccountsDB = struct {
...
+        std.debug.assert(pubkeys.len > 0);
+
+        const slot_ref_map, var lock = self.account_index.slot_reference_map.writeWithLock();
+        defer lock.unlock();
+
+        const slot_entry = try slot_ref_map.getOrPut(slot);
+        if (!slot_entry.found_existing) slot_entry.value_ptr.* = .{
+            .refs = .{},
+            .global_index = undefined,
+        };
```
InKryption: Is it fine for this global_index to be undefined if we hit an error? I would would heed caution and either set it to some reliably pathological value like `maxInt` (guaranteeing index out of bounds if for some reason we happen to use it in or after the recovery path), or popping this entry in an errdefer, ie `errdefer if (!slot_entry.found_existing) _ = slot_ref_map.pop().?;`.
Any thoughts?

src/accountsdb/db.zig
```diff
@@ -1887,14 +2202,16 @@ pub const AccountsDB = struct {
...
-            errdefer self.allocator.free(pubkeys_duped);
-
             const unrooted_accounts, var unrooted_accounts_lg =
                 self.unrooted_accounts.writeWithLock();
             defer unrooted_accounts_lg.unlock();
-            // NOTE: there should only be a single state per slot
-            try unrooted_accounts.putNoClobber(slot, .{ pubkeys_duped, accounts_duped });
+
+            const entry = try unrooted_accounts.getOrPut(slot);
+            if (!entry.found_existing) entry.value_ptr.* = .{};
```
InKryption:
```suggestion
            if (!entry.found_existing) entry.value_ptr.* = .empty;
```

src/accountsdb/db.zig
```diff
@@ -3134,6 +3429,116 @@ test "write and read an account" {
...
+
+    // slot is not in ancestors
+    try std.testing.expectEqual(null, accounts_db.getAccountWithAncestors(&pubkey, &.{}));
+
+    // write account to the same pubkey in the next slot (!)
+    {
+        var data_2 = [_]u8{ 0, 1, 0, 1 };
+
+        const test_account_2 = Account{
+            .data = AccountDataHandle.initAllocated(&data_2),
```
InKryption:
```suggestion
            .data = .initAllocated(&data_2),
```

src/accountsdb/db.zig
```diff
@@ -4016,3 +4423,324 @@ test "read/write benchmark disk" {
...
+        const num_slots_to_insert = random.uintAtMost(usize, slots.len);
+
+        for (0..num_slots_to_insert) |j| {
+            const slot = slots[random.uintLessThan(u64, slots.len)];
+
+            var ancestors = Ancestors{};
+            defer ancestors.deinit(allocator);
+            try ancestors.ancestors.put(allocator, slot, {});
+
+            errdefer std.debug.print(
```
InKryption: For integration with the test runner (to avoid it being truncated):
```suggestion
            errdefer std.log.err(
```

src/accountsdb/db.zig
```diff
@@ -4016,3 +4423,324 @@ test "read/write benchmark disk" {
...
+    var allocated_accounts = ArrayList(sig.runtime.AccountSharedData).init(allocator);
+    defer {
+        for (allocated_accounts.items) |account| allocator.free(account.data);
+        allocated_accounts.deinit();
+    }
+
+    var expected_latest: [pubkey_count]?struct {
+        slot: Slot,
+        account: sig.runtime.AccountSharedData,
+    } = .{null} ** pubkey_count;
```
InKryption: I believe you can use `@splat(null)` here now.

src/accountsdb/manager.zig
```diff
@@ -244,21 +244,20 @@ fn flushSlot(db: *AccountsDB, slot: Slot) !FileId {
 
     defer db.metrics.number_files_flushed.inc();
 
-    const pubkeys, const accounts: []const Account = blk: {
+    var pubkeys_and_accounts = blk: {
```
InKryption, line 247: Can this be `const`?

Sobeston: Can't be, there's a deinit near the bottom of the function

src/accountsdb/db.zig
Sobeston: e17527c22f886e4268e1a1f7a7c4504f3f7485e9

src/accountsdb/db.zig
Sobeston: e17527c22f886e4268e1a1f7a7c4504f3f7485e9

src/accountsdb/db.zig
Sobeston: e17527c22f886e4268e1a1f7a7c4504f3f7485e9

src/accountsdb/db.zig
Sobeston: e17527c22f886e4268e1a1f7a7c4504f3f7485e9

src/accountsdb/db.zig
Sobeston: e17527c22f886e4268e1a1f7a7c4504f3f7485e9

src/accountsdb/db.zig
Sobeston: e17527c22f886e4268e1a1f7a7c4504f3f7485e9

src/accountsdb/db.zig
Sobeston: e17527c22f886e4268e1a1f7a7c4504f3f7485e9

src/accountsdb/db.zig
Sobeston: e17527c22f886e4268e1a1f7a7c4504f3f7485e9

src/accountsdb/db.zig
Sobeston: Cleared it up in cc254b323144b75e536b8e3ae7ddaeeafd5f4857

src/accountsdb/manager.zig
InKryption: Is there a reason it isn't in a `defer`?

Sobeston: Yeah I'm not sure about it either way

e.g. if we got the unrooted accounts for a slot, and then a file access failed, we'd have removed the accounts from memory without writing them to disk. Or we fail to add the accountfile to the file map. 

But in these cases, either option isn't any good... something will go very wrong anyway. Maybe if we leave the deinit out, we could make this function retry-able later down the line? But even then I'm not sure if there's a case where this would fail but the retry would succeed.

So defer or not, a failure in this function will probably cause a validator-stopping failure :thinking:

InKryption: Suppose that's fair enough. I guess if it becomes of import down the line we'll hopefully catch it via leak detection or what-have-you.

InKryption (approved):

dnut (approved):

PR #798, opened 2025-06-30
feat(consensus) Build vote lockout towers from stakes in collectVoteLockouts

dadepo:
wip


src/consensus/replay_tower.zig
```diff
@@ -1579,6 +1613,622 @@ fn optimisticallyBypassVoteStakeThresholdCheck(
...
+        var vote_state = try TowerVoteState.fromAccount(
+            allocator,
+            &vote_account,
+        );
+
+        for (vote_state.votes.constSlice()) |vote| {
+            const interval = try lockout_intervals
+                .getOrPut(allocator, vote.lastLockedOutSlot());
+            if (!interval.found_existing) {
+                interval.value_ptr.* = std.ArrayList(VotedSlotAndPubkey).init(allocator);
```
```suggestion
                interval.value_ptr.* = .init(allocator);
```

src/consensus/replay_tower.zig
```diff
@@ -1579,6 +1613,622 @@ fn optimisticallyBypassVoteStakeThresholdCheck(
...
+            }
+            try interval.value_ptr.*.append(.{ .slot = vote.slot, .pubkey = key });
```
```suggestion
            try interval.value_ptr.append(.{ .slot = vote.slot, .pubkey = key });
```

src/consensus/replay_tower.zig
```diff
@@ -1579,6 +1613,622 @@ fn optimisticallyBypassVoteStakeThresholdCheck(
...
+        // We don't want to update the ancestors stakes of this vote b/c it does not
+        // represent an actual vote by the validator.
+
+        // Note: It should not be possible for any vote state in this bank to have
+        // a vote for a slot >= bank_slot, so we are guaranteed that the last vote in
+        // this vote stack is the simulated vote, so this fetch should be sufficient
+        // to find the last unsimulated vote.
+        std.debug.assert(
+            if (vote_state.nthRecentLockout(0)) |l| l.slot == bank_slot else false,
+        );
```
InKryption: I think this assertion would be clearer like this:
```suggestion
        std.debug.assert(vote_state.nthRecentLockout(0).?.slot == bank_slot);
```
That way if `nthRecentLockout` returns null, it trips the null unwrap; otherwise, it will trip the assert - thus distinguishing these two illegal invariants.

src/consensus/replay_tower.zig
```diff
@@ -1579,6 +1613,622 @@ fn optimisticallyBypassVoteStakeThresholdCheck(
...
+            if (max_parent == null or slot > max_parent.?) {
+                max_parent = slot;
+            }
+        }
+        if (max_parent) |parent| {
+            break :blk voted_stakes.get(parent) orelse @as(u64, 0);
+        } else {
+            break :blk @as(u64, 0);
+        }
+    };
```
```suggestion
    const fork_stake: u64 = blk: {
        var bank_ancestors = ancestors.get(bank_slot) orelse break :blk 0;
        var max_parent: ?Slot = null;
        for (bank_ancestors.items()) |slot| {
            if (max_parent == null or slot > max_parent.?) {
                max_parent = slot;
            }
        }
        if (max_parent) |parent| {
            break :blk voted_stakes.get(parent) orelse 0;
        } else {
            break :blk 0;
        }
    };
```

src/consensus/replay_tower.zig
```diff
@@ -1579,6 +1613,622 @@ fn optimisticallyBypassVoteStakeThresholdCheck(
...
+    allocator: std.mem.Allocator,
+    accounts_db: *AccountsDB,
+    vote_account_pubkey: *const Pubkey,
+) ?Slot {
+    const vote_account = accounts_db.getAccount(vote_account_pubkey) catch return null;
+    const vote_state = stateFromAccount(
+        allocator,
+        &vote_account,
+        vote_account_pubkey,
+    ) catch return null;
```
InKryption: I don't think it is wise to allow ignoring potential allocation errors like this.
I think instead `stateFromAccount` should be made to return only `std.mem.Allocator.Error || error{BincodeError}`, avoiding the fact that bincode currently propagates a bunch of `anyerror` silliness - that would in turn allow this call site to safely do:
```zig
const vote_state = stateFromAccount(
    allocator,
    &vote_account,
    vote_account_pubkey,
) catch |err| switch (err) {
    error.OutOfMemory => |e| return e,
    error.BincodeError => return null,
};
```

src/consensus/tower_state.zig
```diff
@@ -4,10 +4,39 @@ const sig = @import("../sig.zig");
...
+        defer allocator.free(owned_slice);
+
+        return .{
+            .root_slot = vote_state.root_slot,
+            .votes = try std.BoundedArray(
+                Lockout,
+                MAX_LOCKOUT_HISTORY,
+            ).fromSlice(owned_slice),
+        };
+    }
```
InKryption:
```suggestion
    pub fn fromAccount(account: *const VoteAccount) !TowerVoteState {
        const vote_state = account.state;
        var lockouts: std.BoundedArray(Lockout, MAX_LOCKOUT_HISTORY) = .{};
        for (vote_state.votes.items) |landed| {
            try lockouts.append(.{
                .slot = landed.lockout.slot,
                .confirmation_count = landed.lockout.confirmation_count,
            });
        }

        return .{
            .root_slot = vote_state.root_slot,
            .votes = lockouts,
        };
    }
```

InKryption (changes requested):
I have a few suggestions, otherwise it's looking good.

src/consensus/replay_tower.zig
dadepo: [1fc358c](https://github.com/Syndica/sig/pull/798/commits/1fc358cd49aa8da5b09a1a3df7a623ed9c1e787c)

src/consensus/replay_tower.zig
dadepo: [1fc358c](https://github.com/Syndica/sig/pull/798/commits/1fc358cd49aa8da5b09a1a3df7a623ed9c1e787c)

src/consensus/replay_tower.zig
dadepo: [5f62fa2](https://github.com/Syndica/sig/pull/798/commits/5f62fa2a26303a876bcec1b00d803f9fddf2da43)

src/consensus/replay_tower.zig
dadepo: [4e5371f](https://github.com/Syndica/sig/pull/798/commits/4e5371f5a85a89293ff0cd3eec657fa4d79647ce)

src/consensus/replay_tower.zig
dadepo: [460e2d9](https://github.com/Syndica/sig/pull/798/commits/460e2d930f60ae9794d1690c54d2dd00ec21d225)

src/consensus/tower_state.zig
dadepo: [e9c9498](https://github.com/Syndica/sig/pull/798/commits/e9c9498f4b7c6c25c7306b4019095dd5fff6ed00)

InKryption (approved):

InKryption (dismissed):

src/consensus/replay_tower.zig
```diff
@@ -1579,6 +1613,622 @@ fn optimisticallyBypassVoteStakeThresholdCheck(
...
+            try interval.value_ptr.append(.{ .slot = vote.slot, .pubkey = key });
+        }
+
+        // Vote account for this validator
+        if (key.equals(vote_account_pubkey)) {
+            my_latest_landed_vote = if (vote_state.nthRecentLockout(0)) |l| l.slot else null;
+            logger.debug().logf("vote state {any}", .{vote_state});
+            const observed_slot = if (vote_state.nthRecentLockout(0)) |l| l.slot else 0;
+
+            logger.debug().logf("observed slot {any}", .{observed_slot});
```
dnut, line 1671: why are we logging this? debug is the default log level fyi

src/consensus/replay_tower.zig
```diff
@@ -1579,6 +1613,622 @@ fn optimisticallyBypassVoteStakeThresholdCheck(
...
+            .{ vote_account_pubkey, key, voted_stake },
+        );
+
+        var vote_state = try TowerVoteState.fromAccount(
+            &vote_account,
+        );
+
+        for (vote_state.votes.constSlice()) |vote| {
+            const interval = try lockout_intervals
+                .getOrPut(allocator, vote.lastLockedOutSlot());
```
dnut: unnecessary wrapping
```suggestion
        var vote_state = try TowerVoteState.fromAccount(&vote_account);

        for (vote_state.votes.constSlice()) |vote| {
            const interval = try lockout_intervals.getOrPut(allocator, vote.lastLockedOutSlot());
```

src/consensus/replay_tower.zig
```diff
@@ -1579,6 +1613,622 @@ fn optimisticallyBypassVoteStakeThresholdCheck(
     return false;
 }
 
+pub fn collectVoteLockouts(
+    allocator: std.mem.Allocator,
+    logger: Logger,
```
dnut: I think we should probably stop using unscoped loggers since it's so easy to convert the loggers now

```suggestion
    logger: ScopedLogger("replay_tower"),
```

src/consensus/replay_tower.zig
```diff
@@ -1579,6 +1613,622 @@ fn optimisticallyBypassVoteStakeThresholdCheck(
...
+    // Tree of intervals of lockouts of the form [slot, slot + slot.lockout],
+    // keyed by end of the range
+    var lockout_intervals = LockoutIntervals.empty;
+    errdefer lockout_intervals.deinit(allocator);
+
+    var my_latest_landed_vote: ?Slot = null;
+
+    for (vote_accounts.keys(), vote_accounts.values()) |key, value| {
+        const voted_stake = value.stake;
+        const vote_account = value.account;
```
dnut: you don't need to declare separate variables if you use clearer names to begin with

```suggestion
    for (vote_accounts.keys(), vote_accounts.values()) |vote_address, vote| {
```

src/consensus/replay_tower.zig
```diff
@@ -1579,6 +1613,622 @@ fn optimisticallyBypassVoteStakeThresholdCheck(
...
+        try vote_state.processNextVoteSlot(bank_slot);
+
+        for (vote_state.votes.constSlice()) |vote| {
+            try vote_slots.put(vote.slot);
+        }
+
+        if (start_root != vote_state.root_slot) {
+            if (start_root) |root| {
+                const vote = Lockout{ .slot = root, .confirmation_count = MAX_LOCKOUT_HISTORY };
+                logger.trace().logf("ROOT: {}", .{vote.slot});
```
dnut: This log message is unclear and I don't understand what value this provides. It's going to print this for each vote account. Should it include the vote account address in the log message? Maybe indicate that it's a simulated root?

src/consensus/replay_tower.zig
```diff
@@ -1579,6 +1613,622 @@ fn optimisticallyBypassVoteStakeThresholdCheck(
...
+        }
+        total_stake += voted_stake;
+    }
+
+    try populateAncestorVotedStakes(
+        allocator,
+        &voted_stakes,
+        vote_slots.items(),
+        ancestors,
+    );
```
dnut, line 1734:
```suggestion
    try populateAncestorVotedStakes(allocator, &voted_stakes, vote_slots.items(), ancestors);
```

is this an accident, or do you prefer to wrap some code when it's not necessary?

src/consensus/replay_tower.zig
```diff
@@ -1579,6 +1613,622 @@ fn optimisticallyBypassVoteStakeThresholdCheck(
...
+        // Add the last vote to update the `heaviest_subtree_fork_choice`
+        if (vote_state.lastVotedSlot()) |last_landed_voted_slot| {
+            if (progress_map.getHash(last_landed_voted_slot)) |frozen_hash| {
+                _ = try latest_validator_votes_for_frozen_banks.checkAddVote(
+                    allocator,
+                    key,
+                    last_landed_voted_slot,
+                    frozen_hash,
+                    .replay,
+                );
```
dnut: Is this comment accurate? I don't see the fork choice being updated.

src/consensus/replay_tower.zig
```diff
@@ -1579,6 +1613,622 @@ fn optimisticallyBypassVoteStakeThresholdCheck(
...
+            }
+        }
+
+        // Simulate next vote and extract vote slots using the provided bank slot.
+        try vote_state.processNextVoteSlot(bank_slot);
```
dnut, line 1690: Why simulate voting on the slot? Doesn't this change the lockouts? What lockouts are we actually "collecting" in this function? Not real ones?

src/consensus/replay_tower.zig
```diff
@@ -1579,6 +1613,622 @@ fn optimisticallyBypassVoteStakeThresholdCheck(
     return false;
 }
 
+pub fn collectVoteLockouts(
```
dnut, line 1619: This returns more than just lockouts. Is there a better name for this function? Can you describe in one or two sentences what this function is doing (maybe as a doc comment would help)?

src/consensus/replay_tower.zig
```diff
@@ -1579,6 +1613,622 @@ fn optimisticallyBypassVoteStakeThresholdCheck(
...
+    total_stake: Stake,
+) bool {
+    if (voted_stakes.get(slot)) |stake| {
+        return (@as(f64, @floatFromInt(stake)) / @as(f64, @floatFromInt(total_stake))) >
+            DUPLICATE_THRESHOLD;
+    } else {
+        return false;
+    }
+}
+
```
dnut: How does this function know that the slot is duplicate? The inputs don't seem sufficient to draw that conclusion.

It seems like the goal here is to pass in a slot that is already known to be duplicate, and to check if it should be considered "confirmed" despite being a duplicate. Am I getting it?

Maybe this would be clearer with some name changes.


```suggestion
pub fn isDuplicateSlotConfirmed(
    duplicate_slot: Slot,
    voted_stakes: *const VotedStakes,
    total_stake: Stake,
) bool {
    if (voted_stakes.get(duplicate_slot)) |stake| {
        return (@as(f64, @floatFromInt(stake)) / @as(f64, @floatFromInt(total_stake))) >
            DUPLICATE_THRESHOLD;
    } else {
        return false;
    }
}

```

src/consensus/replay_tower.zig
dadepo: [80e0b67](https://github.com/Syndica/sig/pull/798/commits/80e0b679f0a3c4835fbb5b846d11a9690026d94f)

src/consensus/replay_tower.zig
dadepo: It is this one that was actually an accident https://github.com/Syndica/sig/pull/798#discussion_r2226298255. I generally format into newlines when the input params is 2 or more.

src/consensus/replay_tower.zig
dadepo: [4839719](https://github.com/Syndica/sig/pull/798/commits/48397198e9d0d6d499363f3d5105899c6770cef1)

src/consensus/replay_tower.zig
dadepo: [7530482](https://github.com/Syndica/sig/pull/798/commits/75304829db9c578dad873d45f45bbbb7a1ee92c5)

src/consensus/replay_tower.zig
dnut: Interesting ok. Normally I prefer to keep these on one line, but I guess it's a matter of personal preference. I'm fine with this as is

src/consensus/replay_tower.zig
dadepo: So, not sure you are aware but this comments (and most of the new ones you added) applies to already existing code - the reason why this is showing as a diff and hard to tell what exactly changed in the implementation of `collectVoteLockouts` was that it was moved from tower.zig to replay_tower.zig. 

See it on main https://github.com/Syndica/sig/blob/41bba8ce9188c92ee9dd4c54042f5f283f64f318/src/consensus/tower.zig#L286-L426

I will try and respond to them, but maybe it should not be considered as blockers to moving these PR forward.

dadepo: Actually this made me realise the the removal from the `tower.zig` must have sneaked back in, during one of the merges from the target branch. I'll update that now

dadepo: Removed the moved definitions here [ea876b2](https://github.com/Syndica/sig/pull/798/commits/ea876b2e19eca15689fba04f409dcc480689b155)

src/consensus/replay_tower.zig
dadepo: > It seems like the goal here is to pass in a slot that is already known to be duplicate, and to check if it should be considered "confirmed" despite being a duplicate. Am I getting it?

That is indeed the case.  But given that in Agave's lingo, "duplicate confirmed" seems to be a distinct phrase that pops up in a lot of places. Hence why the name of the method.

The suggested renaming is indeed clearer. 

[11f454d](https://github.com/Syndica/sig/pull/798/commits/11f454d1bb0bf533a1dcc05a848a6919a6f64d35)

src/consensus/replay_tower.zig
dadepo: [12052cf](https://github.com/Syndica/sig/pull/798/commits/12052cfe1860672c955f06aa8dffac1fe1b4fc29)

src/consensus/replay_tower.zig
dadepo: It is correct but not the most clear. The votes added here is retrieved and used to update the fork choice here (in the PR right above this PR)

https://github.com/Syndica/sig/blob/c7ffab18d139e86547dfa338efa39b9d2dcd52b1/src/consensus/fork_choice.zig#L1447-L1467

Which is called in computeBankStats here https://github.com/Syndica/sig/blob/0234aee648203e65dec8d30ec9580a0737b18b23/src/replay/consensus.zig#L496 (2 PRs right above this PR)

I have tried to improve the comment here [512275e](https://github.com/Syndica/sig/pull/798/commits/512275ee757b23eef0373589516110a0ddae080c)

src/consensus/replay_tower.zig
dadepo: I agree. it does not look like it is doing much. I've removed it now [0a2aa19](https://github.com/Syndica/sig/pull/798/commits/0a2aa194bbfc0c298ef89d4acd4c24423f89c21a)

If in practice, the need to see this information proves important, then it can easily be added.

src/consensus/replay_tower.zig
dadepo: I had to dig up my notes on this one as I had similar questions when it was first added. My understanding is this. 

The simulation occurs to determine, if all validator vote on this `bank_slot`, what would their lockout state be - ie what votes would have been expired. Basically what will the consensus state be. It then uses this to calculate what the `voted_stakes` will be.

I imagine it as if I am a validator deciding whether to vote on slot 100. I need to know:

- Which other validators would still be "locked" to their previous votes if they voted on slot 100
- How much stake would support slot 100's ancestors after expiry slots has been removed based on above.

So I think this answers the question `> Why simulate voting on the slot?`

> Doesn't this change the lockouts?
It does but for the one used to calculate what the stake would be if everybody voted on `bank_fork`

> What lockouts are we actually "collecting" in this function? Not real ones?

So the ones not affected via the simulation is collected before the simulation and returned as `lockout_intervals` in https://github.com/Syndica/sig/blob/00ed0c34fdaa3fe32a86e709ad9d837a4fef9383/src/consensus/replay_tower.zig#L1657-L1664

src/consensus/replay_tower.zig
```diff
@@ -1579,6 +1613,622 @@ fn optimisticallyBypassVoteStakeThresholdCheck(
...
+        if (start_root != vote_state.root_slot) {
+            if (start_root) |root| {
+                const lockout = Lockout{ .slot = root, .confirmation_count = MAX_LOCKOUT_HISTORY };
+                try vote_slots.put(lockout.slot);
+            }
+        }
+        if (vote_state.root_slot) |root| {
+            const lockout = Lockout{ .slot = root, .confirmation_count = MAX_LOCKOUT_HISTORY };
+            try vote_slots.put(lockout.slot);
```
dnut: `lockout` is unnecessary

src/consensus/replay_tower.zig
dnut: I see what you mean. It still feels very strange to me. But I guess this is exactly how agave works, so it makes sense to follow the same consensus protocol. Do you know why we care about these things? How is this information used?

> - Which other validators would still be "locked" to their previous votes if they voted on slot 100
> - How much stake would support slot 100's ancestors after expiry slots has been removed based on above.

src/consensus/replay_tower.zig
dadepo: [c7fcd21](https://github.com/Syndica/sig/pull/798/commits/c7fcd21b86c8cdbdb56fc0b101632b99827bb6aa)

dnut (approved):

src/consensus/replay_tower.zig
dadepo: Not from the top of my head (I'll have to trace things a bit). I know these values end up in the progress map in `computeBankStats` and values from the progress map get retrieved at various point during replay/consensus - this would be the place i'll start if I want to trace.

PR #800, opened 2025-06-30
feat(replay): main entrypoint into consensus edge cases

InKryption:
Also modify code to use a central replay-scoped logger, and change from "handleEdgeCases" to "processEdgeCases", since the entrypoint is essentially entirely comprised of calls to functions which are also prefixed with "process", facilitating some prefix stripping for the return value's fields.

### DIFF
--- a/src/replay/service.zig
+++ b/src/replay/service.zig
@@ -53,6 +53,15 @@ pub const ReplayDependencies = struct {
 
 pub const Logger = sig.trace.ScopedLogger("replay");
 
+pub const SlotData = struct {
+    duplicate_confirmed_slots: replay.edge_cases.DuplicateConfirmedSlots,
+    epoch_slots_frozen_slots: replay.edge_cases.EpochSlotsFrozenSlots,
+    duplicate_slots_to_repair: replay.edge_cases.DuplicateSlotsToRepair,
+    purge_repair_slot_counter: replay.edge_cases.PurgeRepairSlotCounters,
+    unfrozen_gossip_verified_vote_hashes: replay.edge_cases.UnfrozenGossipVerifiedVoteHashes,
+    duplicate_slots: replay.edge_cases.DuplicateSlots,
+};
+
 const ReplayState = struct {
     allocator: Allocator,
     logger: Logger,
--- a/src/replay/service.zig
+++ b/src/replay/service.zig
@@ -51,9 +51,11 @@ pub const ReplayDependencies = struct {
     root_slot_state: sig.core.SlotState,
 };
 
+pub const Logger = sig.trace.ScopedLogger("replay");
+
 const ReplayState = struct {
     allocator: Allocator,
-    logger: sig.trace.ScopedLogger("replay"),
+    logger: Logger,
     thread_pool: *ThreadPool,
     slot_leaders: SlotLeaders,
     slot_tracker: *SlotTracker,
@@ -135,7 +137,7 @@ fn advanceReplay(state: *ReplayState) !void {
 
     _ = try replay.execution.replayActiveSlots(&state.execution);
 
-    replay.edge_cases.handleEdgeCases();
+    _ = &replay.edge_cases.processEdgeCases;
 
     processConsensus();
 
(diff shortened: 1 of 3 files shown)


InKryption:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Syndica/sig/800?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#800** <a href="https://app.graphite.dev/github/pr/Syndica/sig/800?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Syndica/sig/800?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#778** <a href="https://app.graphite.dev/github/pr/Syndica/sig/778?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#787** <a href="https://app.graphite.dev/github/pr/Syndica/sig/787?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/replay/service.zig
```diff
@@ -130,7 +132,7 @@ fn advanceReplay(state: *ReplayState) !void {
 
     _ = try replay.execution.replayActiveSlots(&state.execution);
 
-    handleEdgeCases();
+    _ = &processEdgeCases;
```
dadepo: Question for my own understanding, any reason why not:
 
```suggestion
    _ = processEdgeCases;
```

As you have often recommended? (I changed it locally and it seems to compile). Is this an omission or something warrants this form here?

src/replay/service.zig
```diff
@@ -211,17 +213,136 @@ fn trackNewSlots(
     }
 }
 
-// -- handleEdgeCases START -- //
-fn handleEdgeCases() void {
-    _ = &processAncestorHashesDuplicateSlots; // TODO:
+// -- processEdgeCases START -- //
```
dadepo: A bit tangential, but not sure there is a game plan yet on where/when things like the `VoteListener` will be kickstarted? 

I believe this is another category of "inputs" - components running alongside the replay loop and is necessary for consensus to work properly - that needs to be accounted for.

src/replay/service.zig
InKryption: In this case it is intentional: by default, the way references to functions work is that direct references merely force their signature to be evaluated (excluding the interdependent generic parameters), whilst forcing evaluation of the (non-generic) body requires either calling it, or otherwise creating a function pointer from it (ie by taking its address or assigning it to a function pointer binding).
I want the body to be evaluated so any type errors are caught without needing to call it.

src/replay/service.zig
```diff
@@ -211,17 +213,136 @@ fn trackNewSlots(
...
+        duplicate_slots_receiver: *sig.sync.Channel(sig.core.Slot),
+        ancestor_hashes_replay_update_sender: *sig.sync.Channel(AncestorHashesReplayUpdate),
+
+        duplicate_confirmed_slots: *DuplicateConfirmedSlots,
+        epoch_slots_frozen_slots: *EpochSlotsFrozenSlots,
+        duplicate_slots_to_repair: *DuplicateSlotsToRepair,
+        purge_repair_slot_counter: *PurgeRepairSlotCounters,
+        unfrozen_gossip_verified_vote_hashes: *UnfrozenGossipVerifiedVoteHashes,
+        latest_validator_votes_for_frozen_banks: *LatestValidatorVotesForFrozenSlots,
+        duplicate_slots_tracker: *DuplicateSlots,
```
dnut: I would probably have these owned by a single struct, and pass that struct by pointer.

InKryption: 17a5bbbbf5385cdc7a46566d9ad0e42cd60ced84

dnut (approved):

dadepo (dismissed):

PR #801, opened 2025-07-01
feat(sysvars): sysvar update functions and sysvar cache population

yewman:
- Change, RecentBlockhashes, SlotHashes, and StakeHistory to use ArrayListUnmanged
- Implement sysvar update functions - missing some functionality dependent on the stakes cache
    - update functions have many args, some of which may end up being passed in by contexts such SloteState, SlotConstants, EpochConstants etc. @dnut I would like to discuss this with you so that the implementation remains inline with your vision for dependency management. 
- Implement fillMissingEntries to populate missing entries in the sysvar cache from accounts db

### DIFF
--- a/src/replay/update_sysvar.zig
+++ b/src/replay/update_sysvar.zig
@@ -117,10 +117,7 @@ pub const UpdateClockDeps = struct {
     update_sysvar_deps: UpdateSysvarAccountDeps,
 };
 
-pub fn updateClock(
-    allocator: std.mem.Allocator,
-    deps: UpdateClockDeps,
-) !void {
+pub fn updateClock(allocator: std.mem.Allocator, deps: UpdateClockDeps) !void {
     const clock = try nextClock(
         allocator,
         deps.feature_set,
@@ -170,10 +167,7 @@ pub fn updateLastRestartSlot(
     );
 }
 
-pub fn updateSlotHistory(
-    allocator: std.mem.Allocator,
-    deps: UpdateSysvarAccountDeps,
-) !void {
+pub fn updateSlotHistory(allocator: std.mem.Allocator, deps: UpdateSysvarAccountDeps) !void {
     var slot_history: SlotHistory = try getSysvarFromAccount(
         allocator,
         deps.accounts_db,
@@ -206,11 +200,7 @@ pub fn updateSlotHashes(
     try updateSysvarAccount(allocator, SlotHashes, slot_hashes, deps);
 }
 
-pub fn updateRent(
-    allocator: std.mem.Allocator,
-    rent: Rent,
-    deps: UpdateSysvarAccountDeps,
-) !void {
+pub fn updateRent(allocator: std.mem.Allocator, rent: Rent, deps: UpdateSysvarAccountDeps) !void {
     try updateSysvarAccount(allocator, Rent, rent, deps);
 }
 


src/core/blockhash_queue.zig
```diff
@@ -0,0 +1,360 @@
...
+        return hash_info;
+    }
+
+    pub fn insertGenesisHash(
+        self: *BlockhashQueue,
+        allocator: Allocator,
+        hash: Hash,
+        lamports_per_signature: u64,
+    ) Allocator.Error!void {
+        std.debug.assert(self.last_hash_index == 0, "Genesis hash should be the first entry");
```
kprotty: assert takes no message

src/core/epoch_schedule.zig
```diff
@@ -86,12 +88,40 @@ pub const EpochSchedule = extern struct {
...
+                self.slots_per_epoch) +|
+                self.first_normal_slot;
+        }
+    }
+
+    pub fn getLastSlotInEpoch(self: *const EpochSchedule, epoch: Epoch) Slot {
+        return self.getFirstSlotInEpoch(epoch) +| self.getSlotsInEpoch(epoch) -| 1;
+    }
+
+    pub fn getLeaderScheduleEpoch(self: *const EpochSchedule, slot: Slot) u64 {
```
dnut: What does this mean? Can you document it?

I assume it has something to do with the fact that the current epoch's leader schedule is derived from staked nodes from another epoch, but that should be a *prior* epoch, whereas here you're *adding* to the current epoch. So I'm very confused by what this is supposed to be returning.

src/core/hard_forks.zig
```diff
@@ -1,39 +1,60 @@
...
+        } else {
+            const index = std.sort.lowerBound(
+                HardFork,
+                self.entries.items,
+                new_slot,
+                HardFork.searchCmp,
+            );
+            try self.entries.insert(allocator, index, .{ .slot = new_slot, .count = 1 });
+        }
     }
```
dnut: I think you can get rid of the call to `binarySearch` and just use `lowerBound` here.

src/core/hard_forks.zig
```diff
@@ -1,39 +1,60 @@
...
+            };
+        } else {
+            const index = std.sort.lowerBound(
+                HardFork,
+                self.entries.items,
+                new_slot,
+                HardFork.searchCmp,
+            );
+            try self.entries.insert(allocator, index, .{ .slot = new_slot, .count = 1 });
+        }
```
kprotty: optimization opportunity, but makes sense.

src/core/hard_forks.zig
```diff
@@ -1,39 +1,60 @@
 const std = @import("std");
+
+const Allocator = std.mem.Allocator;
 const Slot = @import("time.zig").Slot;
 
 pub const HardForks = struct {
-    forks: std.ArrayListUnmanaged(Fork) = .{},
+    entries: std.ArrayListUnmanaged(HardFork) = .{},
```
kprotty: could this have been a ArrayHashMap (using its `.sort()` when needed)?

src/replay/update_sysvar.zig
```diff
@@ -0,0 +1,1435 @@
...
+
+    try accounts_db.putAccountSlice(
+        sysvar_accounts.slice().items(.account),
+        sysvar_accounts.slice().items(.pubkey),
+        slot,
+    );
+}
+
+test "updateClock" {
+    // TODO
```
kprotty: double check if saving for later

src/runtime/sysvar/recent_blockhashes.zig
```diff
@@ -1,34 +1,181 @@
...
+
+    pub fn default(allocator: Allocator) Allocator.Error!RecentBlockhashes {
+        return .{ .entries = try .initCapacity(allocator, MAX_ENTRIES) };
     }
 
-    pub fn last(self: RecentBlockhashes) ?Entry {
-        if (self.entries.len == 0) return null;
-        return self.entries[self.entries.len - 1];
+    pub fn deinit(self: RecentBlockhashes, allocator: Allocator) void {
+        allocator.free(self.entries.allocatedSlice());
```
kprotty: could it be entries.deinit()?

kprotty (commented):
nice. mostly nitpicks and questions

src/core/blockhash_queue.zig
```diff
@@ -0,0 +1,360 @@
...
+        return isHashIndexValid(self.last_hash_index, max_age, hash_info.index);
+    }
+
+    pub fn getHashInfoIfValid(self: BlockhashQueue, hash: Hash, max_age: usize) ?BlockhashInfo {
+        const hash_info = self.hash_infos.get(hash) orelse return null;
+        if (!isHashIndexValid(self.last_hash_index, max_age, hash_info.index)) return null;
+        return hash_info;
+    }
+
+    pub fn insertGenesisHash(
```
dadepo, line 77: More of a question. This method does not seem to be used anywhere in this PR, and I could not find it's counterpart in Agave.

src/core/blockhash_queue.zig
dadepo: Should `last_hash_index` not be updated after this function call?

src/core/blockhash_queue.zig
```diff
@@ -0,0 +1,360 @@
...
+        });
+    }
+
+    pub fn insertHash(
+        self: *BlockhashQueue,
+        allocator: Allocator,
+        hash: Hash,
+        lamports_per_signature: u64,
+    ) Allocator.Error!void {
+        self.last_hash_index += 1;
```
dadepo: Should updating of this field not be done at the end? If it is left at the top, should there not be an `errdefer` block? to revert the changes in case any of the `try` leads to an error?

src/core/blockhash_queue.zig
```diff
@@ -0,0 +1,360 @@
...
+        }
+    }
+
+    pub fn initRandom(
+        allocator: Allocator,
+        random: std.Random,
+        max_list_entries: usize,
+    ) Allocator.Error!BlockhashQueue {
+        // Used by BankFeilds.initRandom inside accounts_db.manager.runLoop, should be made test only when possible.
+        // if (!builtin.is_test) @compileError("only for testing");
```
dadepo, line 137: Uncomment now?

```suggestion
        if (!builtin.is_test) @compileError("only for testing");
```

src/runtime/sysvar/rent.zig
```diff
@@ -65,4 +60,14 @@ pub const Rent = extern struct {
 
         return @intFromFloat((@as(f64, @floatFromInt(lamports_per_year)) * years_elapsed));
     }
+
+    pub fn initRandom(random: std.Random) Rent {
+        // Used by BankFeilds.initRandom inside accounts_db.manager.runLoop, should be made test only when possible.
+        // if (!builtin.is_test) @compileError("only for testing");
```
dadepo, line 66: Uncomment now?

```suggestion
        if (!builtin.is_test) @compileError("only for testing");
```

src/replay/update_sysvar.zig
```diff
@@ -0,0 +1,1435 @@
...
+    old_account: ?*const Account,
+) Allocator.Error!AccountSharedData {
+    // This should NEVER happen, dynamiclly sized sysvars manage there max size.
+    if (bincode.sizeOf(sysvar, .{}) > Sysvar.SIZE_OF)
+        std.debug.panic("sysvar data size exceeds maximum allowed size: sysvar={s}, size={}", .{
+            @typeName(Sysvar),
+            bincode.sizeOf(sysvar, .{}),
+        });
+
+    const sysvar_data = try allocator.alloc(u8, Sysvar.SIZE_OF);
```
dnut: I think the name SIZE_OF is misleading because it implies it's identical to `@sizeOf(Sysvar)` but it's not always. I think the name should indicate what it's actually supposed to be used for.

Here's one potential option to illustrate my point.

```suggestion
    const sysvar_data = try allocator.alloc(u8, Sysvar.STORAGE_SIZE);
```

src/core/hard_forks.zig
dadepo: nit:

```suggestion
    entries: std.ArrayListUnmanaged(HardFork) = .empty,
```

src/core/hard_forks.zig
```diff
@@ -1,39 +1,60 @@
...
-    pub const Fork = extern struct { slot: Slot, count: u64 };
+    pub const HardFork = extern struct {
+        slot: Slot,
+        count: u64,
 
-    pub fn register(self: *HardForks, allocator: std.mem.Allocator, new_slot: Slot) !void {
-        const maybe_index: ?u64 = for (self.forks.items, 0..) |hard_fork, index| {
-            if (hard_fork.slot == new_slot) break index;
-        } else null;
+        pub fn sortCmp(_: void, a: HardFork, b: HardFork) bool {
```
dadepo, line 14:
```suggestion
        pub fn sortCmp(a: HardFork, b: HardFork) bool {
```

src/replay/update_sysvar.zig
```diff
@@ -0,0 +1,1435 @@
...
+    defer recent_blockhashes.deinit(allocator);
+
+    try updateSysvarAccount(
+        allocator,
+        RecentBlockhashes,
+        recent_blockhashes,
+        deps,
+    );
+}
+
```
dnut: There are a lot of unnecessary wraps in this file. I think if you collapse them, the code becomes a lot more readable.

```suggestion
pub fn fillMissingEntries(
    allocator: Allocator,
    db: *AccountsDb,
    ancestors: *const Ancestors,
    sysvar_cache: *SysvarCache,
) !void {
    if (sysvar_cache.clock == null)
        if (getSysvarAndDataFromAccount(allocator, db, ancestors, Clock)) |sysvar_data| {
            sysvar_cache.clock = sysvar_data.data;
        };

    if (sysvar_cache.epoch_schedule == null)
        if (getSysvarAndDataFromAccount(allocator, db, ancestors, EpochSchedule)) |sysvar_data| {
            sysvar_cache.epoch_schedule = sysvar_data.data;
        };

    if (sysvar_cache.epoch_rewards == null)
        if (getSysvarAndDataFromAccount(allocator, db, ancestors, EpochRewards)) |sysvar_data| {
            sysvar_cache.epoch_rewards = sysvar_data.data;
        };

    if (sysvar_cache.rent == null)
        if (getSysvarAndDataFromAccount(allocator, db, ancestors, Rent)) |sysvar_data| {
            sysvar_cache.rent = sysvar_data.data;
        };

    if (sysvar_cache.last_restart_slot == null)
        if (getSysvarAndDataFromAccount(allocator, db, ancestors, LastRestartSlot)) |sysvar_data| {
            sysvar_cache.last_restart_slot = sysvar_data.data;
        };

    if (sysvar_cache.slot_hashes == null)
        if (getSysvarAndDataFromAccount(allocator, db, ancestors, SlotHashes)) |sysvar_data| {
            sysvar_cache.slot_hashes = sysvar_data.data;
            sysvar_cache.slot_hashes_obj = sysvar_data.sysvar;
        };

    if (sysvar_cache.stake_history == null)
        if (getSysvarAndDataFromAccount(allocator, db, ancestors, StakeHistory)) |sysvar_data| {
            sysvar_cache.stake_history = sysvar_data.data;
            sysvar_cache.stake_history_obj = sysvar_data.sysvar;
        };

    if (sysvar_cache.fees_obj == null)
        if (getSysvarFromAccount(allocator, db, ancestors, Fees)) |fees| {
            sysvar_cache.fees_obj = fees;
        };

    if (sysvar_cache.recent_blockhashes_obj == null)
        if (getSysvarFromAccount(allocator, db, ancestors, RecentBlockhashes)) |blockhashes| {
            sysvar_cache.recent_blockhashes_obj = blockhashes;
        };
}

pub const UpdateClockDeps = struct {
    feature_set: *const FeatureSet,
    epoch_schedule: *const EpochSchedule,
    epoch_stakes_map: *const EpochStakesMap,
    stakes_cache: *StakesCache,

    epoch: Epoch,
    parent_epoch: ?Epoch,
    genesis_creation_time: i64,
    ns_per_slot: u64,

    update_sysvar_deps: UpdateSysvarAccountDeps,
};

pub fn updateClock(allocator: std.mem.Allocator, deps: UpdateClockDeps) !void {
    const clock = try nextClock(
        allocator,
        deps.feature_set,
        deps.update_sysvar_deps.ancestors,
        deps.epoch_schedule,
        deps.stakes_cache,
        deps.epoch_stakes_map,
        deps.ns_per_slot,
        deps.genesis_creation_time,
        deps.update_sysvar_deps.accounts_db,
        deps.update_sysvar_deps.slot,
        deps.epoch,
        deps.parent_epoch,
    );
    try updateSysvarAccount(allocator, Clock, clock, deps.update_sysvar_deps);
}

pub fn updateLastRestartSlot(
    allocator: std.mem.Allocator,
    feature_set: *const FeatureSet,
    hard_forks: *const HardForks,
    deps: UpdateSysvarAccountDeps,
) !void {
    if (!feature_set.active.contains(features.LAST_RESTART_SLOT_SYSVAR)) return;

    const new_last_restart_slot = blk: {
        var iter = std.mem.reverseIterator(hard_forks.entries.items);
        while (iter.next()) |hf| if (hf.slot <= deps.slot) break :blk hf.slot;
        break :blk 0;
    };

    if (getSysvarFromAccount(
        allocator,
        deps.accounts_db,
        deps.ancestors,
        LastRestartSlot,
    )) |current| {
        // Only write a new LastRestartSlot if it has changed.
        if (new_last_restart_slot == current.last_restart_slot) return;
    }

    try updateSysvarAccount(
        allocator,
        LastRestartSlot,
        .{ .last_restart_slot = new_last_restart_slot },
        deps,
    );
}

pub fn updateSlotHistory(allocator: std.mem.Allocator, deps: UpdateSysvarAccountDeps) !void {
    var slot_history: SlotHistory = getSysvarFromAccount(
        allocator,
        deps.accounts_db,
        deps.ancestors,
        SlotHistory,
    ) orelse try SlotHistory.default(allocator);
    defer slot_history.deinit(allocator);

    slot_history.add(deps.slot);

    try updateSysvarAccount(allocator, SlotHistory, slot_history, deps);
}

pub fn updateSlotHashes(
    allocator: std.mem.Allocator,
    parent_slot: Slot,
    parent_hash: Hash,
    deps: UpdateSysvarAccountDeps,
) !void {
    var slot_hashes: SlotHashes = getSysvarFromAccount(
        allocator,
        deps.accounts_db,
        deps.ancestors,
        SlotHashes,
    ) orelse try SlotHashes.default(allocator);
    defer slot_hashes.deinit(allocator);

    slot_hashes.add(parent_slot, parent_hash);

    try updateSysvarAccount(allocator, SlotHashes, slot_hashes, deps);
}

pub fn updateRent(allocator: std.mem.Allocator, rent: Rent, deps: UpdateSysvarAccountDeps) !void {
    try updateSysvarAccount(allocator, Rent, rent, deps);
}

pub fn updateEpochSchedule(
    allocator: std.mem.Allocator,
    epoch_schedule: EpochSchedule,
    deps: UpdateSysvarAccountDeps,
) !void {
    try updateSysvarAccount(allocator, EpochSchedule, epoch_schedule, deps);
}

// TODO: Requires StakesCache
// pub fn updateStakeHistory(
//     allocator: std.mem.Allocator,
//     ancestors: *const Ancestors,
//     rent: *const Rent,
//     state: *SlotState,
//     accounts_db: *AccountsDb,
//     slot: Slot,
//     epoch: Epoch,
//     maybe_epoch: ?Epoch,
//     stakes_cache: *const StakesCache,
// ) !void {
//     if (maybe_epoch) |e| if (e == epoch) return;
//     try updateSysvarAccount(
//         allocator,
//         state,
//         accounts_db,
//         ancestors,
//         rent,
//         slot,
//         StakeHistory
//         stakes_cache.stakes().history(),
//     );
// }

pub fn updateRecentBlockhashes(
    allocator: std.mem.Allocator,
    blockhash_queue: *const BlockhashQueue,
    deps: UpdateSysvarAccountDeps,
) !void {
    const recent_blockhashes = try RecentBlockhashes.fromBlockhashQueue(allocator, blockhash_queue);
    defer recent_blockhashes.deinit(allocator);

    try updateSysvarAccount(allocator, RecentBlockhashes, recent_blockhashes, deps);
}

```

dadepo (commented):
About a quarter done. Submitting the reviews so those can be procossed

src/runtime/sysvar/slot_hashes.zig
```diff
@@ -1,42 +1,192 @@
...
+        Pubkey.parseBase58String("SysvarS1otHashes111111111111111111111111111") catch unreachable;
 
-    // [agave] https://github.com/anza-xyz/solana-sdk/blob/9148b5cc95b43319f3451391ec66d0086deb5cfa/slot-hashes/src/lib.rs#L21
     pub const MAX_ENTRIES: usize = 512;
 
-    fn compareFn(key: Slot, mid_item: Entry) std.math.Order {
-        return std.math.order(key, mid_item[0]);
+    pub const SIZE_OF: usize = 20_488;
+
+    pub fn default(allocator: Allocator) Allocator.Error!SlotHashes {
```
dnut: `default` is widely used in rust which is unfortunate because I don't think it's a great name for this kind of thing.

```suggestion
    pub fn empty(allocator: Allocator) Allocator.Error!SlotHashes {
```

A similar suggestion applies to the other sysvars

src/replay/update_sysvar.zig
```diff
@@ -0,0 +1,1435 @@
...
+    allocator: std.mem.Allocator,
+    comptime Sysvar: type,
+    sysvar: Sysvar,
+    deps: UpdateSysvarAccountDeps,
+) !void {
+    // TODO: handle errors directly
+    const maybe_old_account = deps.accounts_db.getAccountWithAncestors(
+        &Sysvar.ID,
+        deps.ancestors,
+    ) catch null;
```
dnut:
```suggestion
    const maybe_old_account = try deps.accounts_db.getAccountWithAncestors(
        &Sysvar.ID,
        deps.ancestors,
    );
```

src/replay/update_sysvar.zig
```diff
@@ -0,0 +1,1435 @@
...
+        if (new_account.lamports > old_account.lamports)
+            _ = deps.capitalization.fetchAdd(
+                new_account.lamports - old_account.lamports,
+                .monotonic,
+            )
+        else if (new_account.lamports < old_account.lamports)
+            _ = deps.capitalization.fetchSub(
+                old_account.lamports - new_account.lamports,
+                .monotonic,
+            );
```
dnut:
```suggestion
        if (new_account.lamports != old_account.lamports)
            _ = deps.capitalization.fetchAdd(
                new_account.lamports -% old_account.lamports,
                .monotonic,
            );
```

src/replay/update_sysvar.zig
```diff
@@ -0,0 +1,1435 @@
...
+            @typeName(Sysvar),
+            bincode.sizeOf(sysvar, .{}),
+        });
+
+    const sysvar_data = try allocator.alloc(u8, Sysvar.SIZE_OF);
+    errdefer allocator.free(sysvar_data);
+    @memset(sysvar_data, 0);
+
+    // writeToSlice may return a 'NoSpaceLeft' error. The above allocation ensures
+    // that there is enough space for serialisation.
```
dnut: Are you sure this is true? You need to anticipate how large the item is, which it's not clear that SIZE_OF does correctly for dynamically sized items. Why not just use bincode.writeAlloc?

src/replay/update_sysvar.zig
```diff
@@ -0,0 +1,1435 @@
...
+
+    const sysvar = bincode.readFromSlice(
+        allocator,
+        Sysvar,
+        data.owned_allocation,
+        .{},
+    ) catch {
+        data.deinit(allocator);
+        return null;
+    };
```
dnut:
```suggestion
    const data = account.data.iterator();
    const sysvar = bincode.read(allocator, Sysvar, data.reader(), .{}) catch return null;
```

src/replay/update_sysvar.zig
```diff
@@ -0,0 +1,1435 @@
...
+) ?Sysvar {
+    const maybe_account = accounts_db.getAccountWithAncestors(
+        &Sysvar.ID,
+        ancestors,
+    ) catch return null;
+
+    const account = maybe_account orelse return null;
+    defer account.deinit(allocator);
+
+    const data = account.data.dupeAllocatedOwned(allocator) catch
```
dnut: same as above

src/runtime/sysvar/recent_blockhashes.zig
```diff
@@ -1,34 +1,181 @@
...
+
+    pub fn initWithEntries(
+        allocator: Allocator,
+        entries: []const Entry,
+    ) Allocator.Error!RecentBlockhashes {
+        if (!builtin.is_test) @compileError("only for tests");
+        std.debug.assert(entries.len <= MAX_ENTRIES);
+        var self = try RecentBlockhashes.default(allocator);
+        for (entries) |entry| self.entries.appendAssumeCapacity(entry);
+        return self;
```
dnut:
```suggestion
        const dupe = allocator.dupe(Entry, entries);
        return .{ .entries = std.ArrayListUnmanaged(Entry).fromOwnedSlice(dupe) };
```

src/runtime/sysvar/slot_history.zig
```diff
@@ -22,8 +24,19 @@ pub const SlotHistory = struct {
     pub const ID =
         Pubkey.parseBase58String("SysvarS1otHistory11111111111111111111111111") catch unreachable;
 
-    pub fn deinit(self: SlotHistory, allocator: std.mem.Allocator) void {
-        sig.bincode.free(allocator, self);
+    pub const SIZE_OF: u64 = 131_097;
+
+    pub fn default(allocator: Allocator) Allocator.Error!SlotHistory {
+        var bits = try DynamicArrayBitSet(u64).initEmpty(allocator, MAX_ENTRIES);
+        bits.set(0);
```
dnut, line 33: why is this the default?

src/core/blockhash_queue.zig
yewman: 41c5ad

src/core/hard_forks.zig
yewman: 41c5ad0d

src/core/hard_forks.zig
yewman: 41c5ad0d
Added a comment, can consider a hash map depending on usage in future

src/replay/update_sysvar.zig
yewman: filled in next pr

src/runtime/sysvar/recent_blockhashes.zig
yewman: would require passing `*RecentBlockhashes`,  based on previous conversations my understanding is that this is the preferred approach

src/core/blockhash_queue.zig
yewman: We don't have start up from genesis completed yet, but will want it in the future. I have added a test case demonstrating usage. It exists specifically to add an initial hash without incrementing the last_hash_index

src/core/blockhash_queue.zig
yewman: ^

yewman: https://github.com/firedancer-io/agave/blob/10fe1eb29aac9c236fd72d08ae60a3ef61ee8353/accounts-db/src/blockhash_queue.rs#L96-L97

src/core/blockhash_queue.zig
yewman: 41c5ad

yewman: 41c5ad0d1

src/core/blockhash_queue.zig
yewman: As per comment, this will break compilation

src/runtime/sysvar/rent.zig
yewman: ^

src/core/hard_forks.zig
yewman: 41c5ad0d

src/core/hard_forks.zig
yewman: comparator for std.sort functions require a context argument, in this case we do not use a context hence it is type void

src/core/hard_forks.zig
yewman: 41c5ad0d

src/replay/update_sysvar.zig
yewman: b5225ed0

src/replay/update_sysvar.zig
yewman: b5225ed0

src/replay/update_sysvar.zig
yewman: ad9034e1

yewman: b37e2bd3

src/replay/update_sysvar.zig
yewman: b37e2bd3

src/replay/update_sysvar.zig
yewman: Bad assumption, changed to try here b37e2bd3.  With regard to using `writeAlloc`, all allocated sysvars have a fixed max size and when serialised for accounts db require that their data field contains a slice with length equal to their fixed max size, if a sysvar does not occupy its full capacity the remaining bytes should be zeroed

src/replay/update_sysvar.zig
yewman: b37e2bd3, left the comment in the wrong function, we need to dupe here since the bytes are subsequently stored in the sysvar cache

src/replay/update_sysvar.zig
yewman: b37e2bd3

src/runtime/sysvar/recent_blockhashes.zig
yewman: b37e2bd3

src/runtime/sysvar/slot_hashes.zig
yewman: For the allocated sysvars the default initialises them with the default max entries, and they are then not allowed to be resized. Perhaps using a BoundedArray containing an allocated slice would be preferable to enforce this constraint. As for the usage of default, I tend to agree with you and don't think its the best name either, however in this instance since it is associated with a fixed max size I think it kind of makes sense. Happy to run with empty if you prefer though, so lmk

src/runtime/sysvar/slot_history.zig
yewman: It presume it is genesis related, this method is a direct port from agave. The first call to update_slot_history uses an unwrap_default() on the slot history sysvar, adds the bank slot and writes to accounts db. If the bank slot at genesis is always zero then it is kind of redundant, otherwise it will have effects. Removing it does not affect any existing tests, so I think it is probably fine to remove. Lmk if you have a preference

src/core/epoch_schedule.zig
yewman: It is effectively used to control 'when' stake information is saved into the epoch stakes map, from which the leader schedule is computed. `Bank::process_new_epoch` and `Bank::_new_from_parent` both get the leader schedule epoch and call `Bank::update_epoch_stakes` which adds a new entry to the epoch_stakes_map using the current stakes cache. When the leader schedule slot offset is equal to the slots per epoch then this function will always return 1 (this is the case for mainnet and epoch schedule default). I have added a comment and test case to reflect this. Potentially we could remove it, however, since the fuzzing framework can input any epoch schedule I cannot be certain that it will not cause issues. 8037bd19

src/core/blockhash_queue.zig
dadepo: I also see the only place that index get's updated is via https://github.com/firedancer-io/agave/blob/10fe1eb29aac9c236fd72d08ae60a3ef61ee8353/accounts-db/src/blockhash_queue.rs#L113 still not sure how it fits together.

dadepo: > I also see the only place...

Not the only place.

src/runtime/sysvar/slot_hashes.zig
dnut: Let's make a distinction between size (capacity) and usage (length).

This is effectively a fixed size data structure. There's only one size this struct is supposed to allocate. I would expect any init function to create it with the same size. I don't think it's meaningful to say there is a "default" size since the size is not supposed to be adjustable. It's just the "correct" size. So from the perspective of size, this is just a regular initialization function. I think `init` would be a better name because you're simply initializing the size to the only valid state.

From the perspective of usage, I think `empty` is a clearer way to indicate the emptiness than `default`. `init` would also be fine since it's unopinionated.

dnut: Also I like the bounded array idea. I think that makes a bit more sense.

src/core/epoch_schedule.zig
dnut: This comment plus the code is inconsistent with my prior understanding:

> Gets the epoch for which the given slot should save information about stakers.

The function returns an epoch for any possible input slot.

But I thought the leader schedule is derived from only one specific slot in the epoch, not any slot. If the comment is correct, I would have thought it would return `null` for the majority of inputs. Do you have any insight into this? Can you actually observe the stake accounts at any slot within an epoch, and it will always represent the same exact stake per node? I thought it was variable and you had to be careful about the which slot you use to snapshot the stakes to determine the epoch staked nodes for the next epoch.

Also can you use the `Epoch` alias here if it's returning an epoch number?

src/replay/update_sysvar.zig
```diff
@@ -0,0 +1,1336 @@
...
+        data.owned_allocation,
+        .{},
+    ) catch {
+        data.deinit(allocator);
+        return null;
+    };
+
+    return .{ sysvar, .{
+        .lamports = account.lamports,
+        .data = try allocator.dupe(u8, data.owned_allocation),
```
dnut: data is being duped twice

src/runtime/sysvar/slot_history.zig
dnut: Maybe the genesis slot should explicitly be added by the user when processing slots starting from genesis. What do you think? I don't have enough context to have a preference. I trust your judgment

src/core/epoch_schedule.zig
yewman: Ahh yes, so the epoch stakes are only saved if the `leader_schedule_epoch` provided to the `update_epoch_stakes` method is not already in the epoch stakes map, thus the epoch stakes are always saved at a specific slot, which will be the first slot for which this function returns an epoch which is not in the current epoch stakes map. With current configuration, this will be the first slot of a new epoch

src/runtime/sysvar/slot_hashes.zig
yewman: I think the bounded array idea makes more sense as well, would you be okay with me renaming these to `init` and transitioning these structs to use the bounded array in a follow up PR. They touch a lot of code and it may create more headaches with so many open dependent branches right now. Otherwise let me know and I will get straight on it now.

src/replay/update_sysvar.zig
yewman: 9da2ff65

src/runtime/sysvar/slot_hashes.zig
dnut: I would prefer not to introduce technical debt, but if that overall makes the whole process easier, then the debt has negative interest, which sounds good to me, as long as it isn't forgotten about.

src/runtime/sysvar/slot_history.zig
yewman: I have a very slight preference to leave it, I have added a comment for clarity

src/core/epoch_schedule.zig
dnut: OK I get it. This is confusing, but I see what's going on now.

Ideally, I think this function would be clearer if it returns null for any slot that isn't actually the one being used to calculate the staked nodes. But I see why that's not practical. This function would need to be fork aware to do that. For example if the first slot of an epoch is skipped, then this would need to return non-null for the second slot. So that approach doesn't really work unless you pass Ancestors into this function, which is messy.

Ultimately I would prefer a less confusing approach overall, but I see that this couples with sysvars, which seems tricky to untangle.

Maybe the comment can be expanded a bit more to explain the whole picture, because it is so confusing how all of this code interacts.

What do you think about this as the method's doc:
```
    /// Gets the epoch for which the stakes from the current slot could
    /// potentially be used to defined the "epoch staked nodes," which are used
    /// to calculate the leader schedule.
    ///
    /// In agave this occurs during `Bank::process_new_epoch` and
    /// `Bank::_new_from_parent`. If the leader schedule slot offset equals the
    /// number of slots per epoch (current default) this function will always
    /// return the epoch immediately after the epoch of the given slot.
    ///
    /// If this slot is the first slot on its fork that could potentially be
    /// used to defined a particular epoch's "epoch staked nodes," then this
    /// will be the slot that is used for that purpose. All future slots on the
    /// same fork will *not* be used for that epoch, even if they do return the
    /// same Epoch number from this function.
```

dnut: not sure if you saw this

> Also can you use the Epoch alias here if it's returning an epoch number?

src/runtime/sysvar/slot_hashes.zig
dnut: where would you like to do this? in one of the stacked prs?

dnut: what do you think about committing it in another pr and then we can resolve this?

src/core/epoch_schedule.zig
yewman: Oh yeah I missed that. I think your proposed method doc is great. 6f1268c0

src/runtime/sysvar/slot_hashes.zig
yewman: Yeah the plan would be to implement it on top of harnew/sysvars-with-stakes-cache, I will create a new PR now

src/runtime/sysvar/recent_blockhashes.zig
kprotty: what about `var entries = self.entries; entries.deinit(allocator);`

src/runtime/sysvar/slot_hashes.zig
yewman: 69ed3dc

src/runtime/sysvar/recent_blockhashes.zig
yewman: 69ed3dc

dnut (dismissed):

dnut (dismissed):

kprotty (approved):

dadepo (dismissed):

PR #803, opened 2025-07-02
refactor(consensus:vote_listener): refactor the vote listener unit test to better "parameterize" it

InKryption:
Expands the inputs, and makes the logic more easily parameterizable.
Also adds comments referencing the relevant logic.
Also attempts to Makes it less flakey, though this seems unsuccessful. This will be properly addressed in a follow up change.

### DIFF
--- a/src/consensus/vote_listener.zig
+++ b/src/consensus/vote_listener.zig
@@ -140,6 +140,7 @@ pub const Senders = struct {
 };
 
 pub const VoteListener = struct {
+    allocator: std.mem.Allocator,
     verified_vote_transactions: *sig.sync.Channel(Transaction),
     recv: std.Thread,
     process_votes: std.Thread,
@@ -196,6 +197,7 @@ pub const VoteListener = struct {
         process_votes_thread.setName("solCiProcVotes") catch {};
 
         return .{
+            .allocator = allocator,
             .verified_vote_transactions = verified_vote_transactions,
             .recv = recv_thread,
             .process_votes = process_votes_thread,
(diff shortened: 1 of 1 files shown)


InKryption:
* **#805** <a href="https://app.graphite.dev/github/pr/Syndica/sig/805?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#804** <a href="https://app.graphite.dev/github/pr/Syndica/sig/804?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#803** <a href="https://app.graphite.dev/github/pr/Syndica/sig/803?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Syndica/sig/803?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/consensus/vote_listener.zig
```diff
@@ -1689,26 +1615,163 @@ test VoteListener {
...
-        try std.testing.expectEqualDeep(
-            &expected_trackers,
-            actual_trackers.items,
+    try std.testing.expectEqualDeep(
+        expected_trackers.items,
+        actual_trackers.items,
+    );
+}
+
+const TestSlotVoteTracker = struct {
```
dadepo, line 1656: Is this not testing a different implementation than the real vote tracker? Is there a risk of drift, where changes are made in `SlotVoteTracker` but not reflected in this other struct used in the test?

InKryption: This data structure just makes it easier to test the contents of the real slot vote tracker, by being properly usable in `expectEqual{Deep}`.
I don't think there's a serious risk of them drifting in a way that would be harmful: the `from` function expects the real `SlotVoteTracker` to be of a certain shape, so if any of the relevant parts of that shape changes, it would be a compile error.
If new fields are added to `SlotVoteTracker`, then there wouldn't be any error, but it also wouldn't change the behavior of any of the tests; if the changes need to be tested, then whoever is writing or modifying the test would realize that it lacks certain fields, and add them so they can be tested properly, or use something else specific to what they're testing.

dnut (approved):

PR #804, opened 2025-07-02
refactor(consensus:vote_listener): Begin gutting and overhauling `BankForksStub`

InKryption:
This change set begins to replace the current internals of BankForksStub with a slot tracker and epoch tracker, setting it up for proper integration.

### DIFF
--- a/src/consensus/vote_listener.zig
+++ b/src/consensus/vote_listener.zig
@@ -12,49 +12,63 @@ const TransactionMessage = sig.core.transaction.Message;
 const VoteTransaction = sig.consensus.vote_transaction.VoteTransaction;
 const VoteTracker = sig.consensus.VoteTracker;
 
+const SlotTracker = sig.replay.trackers.SlotTracker;
+const EpochTracker = sig.replay.trackers.EpochTracker;
+
 pub const BankForksStub = struct {
-    root_slot: Slot,
-    banks: std.AutoArrayHashMapUnmanaged(Slot, BankStub),
+    slot_tracker: SlotTracker,
+    epoch_tracker: EpochTracker,
 
     pub fn deinit(self: BankForksStub, allocator: std.mem.Allocator) void {
-        var copy = self;
-        for (self.banks.values()) |bank| {
-            bank.deinit(allocator);
-        }
-        copy.banks.deinit(allocator);
+        self.slot_tracker.deinit(allocator);
+        self.epoch_tracker.deinit(allocator);
     }
 
     pub fn init(
         allocator: std.mem.Allocator,
+        epoch_schedule: sig.core.EpochSchedule,
         root: struct {
             slot: Slot,
-            /// Gets cloned.
-            bank: BankStub,
+            constants: sig.core.SlotConstants,
+            state: sig.core.SlotState,
+            epoch_constants: sig.core.EpochConstants,
         },
     ) std.mem.Allocator.Error!BankForksStub {
         var self: BankForksStub = .{
-            .root_slot = root.slot,
-            .banks = .{},
+            .slot_tracker = try .initWithRoot(allocator, root.slot, root.constants, root.state),
+            .epoch_tracker = .{
+                .schedule = epoch_schedule,
+                .epochs = .{},
+            },
         };
         errdefer self.deinit(allocator);
 
-        try self.banks.ensureUnusedCapacity(allocator, 1);
-        self.banks.putAssumeCapacity(self.root_slot, try root.bank.clone(allocator));
+        try self.epoch_tracker.epochs.ensureUnusedCapacity(allocator, 1);
+        self.epoch_tracker.epochs.putAssumeCapacity(
+            epoch_schedule.getEpoch(root.slot),
+            try root.epoch_constants.clone(allocator),
+        );
 
         return self;
     }
 
     pub fn getBank(self: *const BankForksStub, slot: Slot) ?BankStub {
-        return self.banks.get(slot);
+        const st_elem = self.slot_tracker.get(slot) orelse return null;
+        const et_elem = self.epoch_tracker.getForSlot(slot);
+        return .{
+            .slot = slot,
+            .st_elem = st_elem,
+            .et_elem = et_elem,
+        };
     }
 
     pub fn rootBank(self: *const BankForksStub) BankStub {
-        return self.banks.get(self.root_slot).?; // root slot's bank must exist
+        return self.getBank(self.slot_tracker.root).?; // root slot's bank must exist
     }
 
     pub fn bankHash(self: *const BankForksStub, slot: Slot) ?Hash {
         const bank = self.getBank(slot) orelse return null;
-        return bank.hash;
+        return bank.st_elem.state.hash.readCopy();
     }
 
     pub fn bankHashOrNullIfFrozen(self: *const BankForksStub, slot: Slot) ?Hash {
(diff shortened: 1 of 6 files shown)


src/consensus/vote_listener.zig
```diff
@@ -12,49 +12,63 @@ const TransactionMessage = sig.core.transaction.Message;
...
+            epoch_constants: sig.core.EpochConstants,
         },
     ) std.mem.Allocator.Error!BankForksStub {
         var self: BankForksStub = .{
-            .root_slot = root.slot,
-            .banks = .{},
+            .slot_tracker = try .initWithRoot(allocator, root.slot, root.constants, root.state),
+            .epoch_tracker = .{
+                .schedule = epoch_schedule,
+                .epochs = .{},
```
dadepo: ?
```suggestion
                .epochs = .empty,
```

src/replay/trackers.zig
```diff
@@ -104,9 +123,16 @@ pub const SlotTracker = struct {
...
+    epochs: std.AutoArrayHashMapUnmanaged(Epoch, EpochConstants),
     schedule: EpochSchedule,
 
+    pub fn init(schedule: EpochSchedule) EpochTracker {
+        return .{
+            .epochs = .empty,
+            .schedule = schedule,
+        };
+    }
+
```
dnut: What's the point of adding more layers? This is superfluous with struct initialization syntax.

```suggestion
    epochs: std.AutoArrayHashMapUnmanaged(Epoch, EpochConstants) = .empty,
    schedule: EpochSchedule,

```

src/replay/trackers.zig
```diff
@@ -24,23 +24,38 @@ pub const SlotTracker = struct {
     slots: std.AutoArrayHashMapUnmanaged(Slot, *Element),
     root: Slot,
 
-    const Element = struct {
+    pub const Element = struct {
```
dnut, line 27: Can we keep this private since it's only used for internal storage?

src/replay/trackers.zig
```diff
@@ -68,6 +83,10 @@ pub const SlotTracker = struct {
         };
     }
 
+    pub fn getRoot(self: *const SlotTracker) Reference {
+        return self.get(self.root).?; // root slot's bank must exist
```
dnut, line 106: I don't think this is a safe assumption unless we delete `init` and rename `initWithRoot` to `init`

src/core/bank.zig
```diff
@@ -231,22 +231,31 @@ pub const EpochConstants = struct {
...
 
-    pub fn deinit(self: EpochConstants, allocator: Allocator) void {
-        self.stakes.deinit(allocator);
+    pub fn clone(
+        self: EpochConstants,
+        allocator: std.mem.Allocator,
+    ) std.mem.Allocator.Error!EpochConstants {
+        var cloned = self;
+        cloned.stakes = try self.stakes.clone(allocator);
+        return cloned;
```
dnut: This works fine, but usually I prefer to construct the entire struct from scratch so the compiler will force people who add new fields to think about how to clone them.

src/consensus/vote_listener.zig
```diff
@@ -65,69 +79,8 @@ pub const BankForksStub = struct {
...
-            return .{
-                .slot = self.slot,
-                .hash = self.hash,
-                .ancestors = ancestors,
-                .epoch_schedule = self.epoch_schedule,
-                .epoch_stakes = epoch_stakes,
-            };
-        }
+        st_elem: SlotTracker.Reference,
+        et_elem: ?sig.core.EpochConstants,
```
dnut, line 83: Why did you expand this to hold more state than necessary? If it worked fine with only ancestors, epoch_schedule, and epoch_stakes, why not keep it that way?

src/consensus/vote_listener.zig
```diff
@@ -65,69 +79,8 @@ pub const BankForksStub = struct {
 
     pub const BankStub = struct {
```
dnut, line 80: Why don't you add two methods to this struct: `epochStakes(Slot) EpochStakes` and `rootSlot() Slot`, and call them from throughout the code in this file instead of directly accessing the fields?

src/replay/trackers.zig
InKryption: Just following the rule of thumb presented by https://ziglang.org/documentation/master/#toc-Faulty-Default-Field-Values, as I try to do in most other code. Default values only make sense if leaving them in their default state is coherent with every other field's possible invariants.

src/replay/trackers.zig
InKryption: ac077445620091aa49c8a894b147a1956b690427

src/replay/trackers.zig
InKryption: I think the latter would be preferable, I'll go ahead and try to do that.

src/replay/trackers.zig
InKryption: Actually I was going to do some other things with this, specifically change the API to use this for the initialization of slot elements, ie in `put`, and now also in `init`.

src/replay/trackers.zig
dnut: The data invariants of the struct cannot be violated by omitting epochs from initialization.

src/replay/trackers.zig
InKryption: e842fe0989dafcaa6c1be8997b1a7bd3daa8811a

src/core/bank.zig
InKryption: 3164ef607a5541f0fd064b8a1a409111e3e1ad36

src/consensus/vote_listener.zig
InKryption: Just to leave an answer here, this change was made before we had that discussion where it was made clear that this state isn't all needed, and that the methods for querying the relevant data can be made more narrow in scope. Am in the process of refactoring these things.

src/consensus/vote_listener.zig
InKryption: Same answer as on the fields comment

src/replay/trackers.zig
InKryption: Re-publicized in e842fe0989dafcaa6c1be8997b1a7bd3daa8811a

src/replay/trackers.zig
InKryption: bc435de0981c92b82392fa244d1c763c7bfcbdb5

src/consensus/vote_listener.zig
InKryption: I think it will be easier to do this as another PR, which I have: #815

src/replay/trackers.zig
dnut: I think the whole slot tracker idea may need to be redesigned after we get replay working. That's why I'm trying to avoid propagating it and its internal data structures. Anyway your change doesn't introduce too much coupling so I'm fine with it.

src/consensus/vote_listener.zig
dnut: It's not a high priority. let's just leave it as is and come back to it another time.

src/consensus/vote_listener.zig
dnut: It's not a high priority. let's just leave it as is and come back to it another time.

dnut (approved):

src/consensus/vote_listener.zig
```diff
@@ -12,49 +12,63 @@ const TransactionMessage = sig.core.transaction.Message;
...
+        return .{
+            .slot = slot,
+            .st_elem = st_elem,
+            .et_elem = et_elem,
+        };
     }
 
     pub fn rootBank(self: *const BankForksStub) BankStub {
-        return self.banks.get(self.root_slot).?; // root slot's bank must exist
+        return self.getBank(self.slot_tracker.root).?; // root slot's bank must exist
```
dadepo, line 66: nit: Is right time to update the terminology away from Bank?

src/consensus/vote_listener.zig
```diff
@@ -314,16 +257,9 @@ fn getVoteTransactionsAfterCursor(
 fn verifyVoteTransaction(
     allocator: std.mem.Allocator,
     vote_tx: Transaction,
-    bank_forks: struct {
-        /// Should be `bank_forks.rootBank()`
-        root_bank: BankForksStub.BankStub,
-        /// Should be `bank_forks.rootBank().epoch_schedule`
-        epoch_schedule: sig.core.EpochSchedule,
-    },
+    /// Should be associated with the root bank.
```
dadepo, line 260: Same question about moving away from the `bank` terminology

src/replay/edge_cases.zig
```diff
@@ -1313,34 +1315,67 @@ const TestData = struct {
...
-            ) @This() {
-                return .{
-                    .parent_slot = parent_slot,
-                    .slot = slot,
-                    .hash = .initRandom(_random),
-                    .fork_progress_init = fork_progress_init,
-                };
-            }
-        };
+        fn toDummyElem(
```
dadepo, line 1348: Maybe some comment on why this is a dummyElem? and also a `is_test` check that this is meant to be used during test

src/consensus/vote_listener.zig
InKryption: I believe this is already addressed in more recent commits

src/consensus/vote_listener.zig
InKryption: I would say yes, I've done so in #815, with the main remnants of the terminology being `BankForksStub` which just needs a brand new name entirely with its purpose now being different, instances of that being called `bank_forks`, `progressWithNewRootBank`, `bank_notification` field of `Senders`, and `BankNotification` - these shouldn't be hard, but would rather leave that to another PR making incremental progress on this.

src/replay/edge_cases.zig
InKryption: 50a572702fa537ab9ac27daf52b34a76503a2271

src/consensus/vote_listener.zig
dadepo: Yeah. This was a review I had since last week and just got submitted now.

PR #806, opened 2025-07-07
feat(replay): stakes cache 

yewman:


src/core/stakes_cache.zig
```diff
@@ -0,0 +1,748 @@
...
+            .eq => _ = self.staked_nodes.swapRemove(pubkey),
+            .gt => current_stake.* -= stake,
+        }
+    }
+};
+
+/// NOTE: This struct is wrapped in an `Arc` in agave
+pub const VoteAccount = struct {
+    account: AccountSharedData,
+    state: VoteState,
```
dadepo: I know this is still in draft, but it did come up last week that maybe this should not be the full `VoteState` struct? The other question was about using versioned instead.

So might be an idea to clarify that without going ahead with full implementation + tests etc.

@dnut may have some more thoughts around this.

yewman: Yeah I am open to it being minimal, what do you require from the vote state for your components? If we don't store the entire vote state I see no reason to handle versioning

dadepo: > what do you require from the vote state for your components?

That would be the votes, and the root slot

src/core/stake.zig
```diff
@@ -1,828 +1,867 @@
...
-        /// Should be a PRNG, not a true RNG. See the documentation on `std.Random.uintLessThan`
-        /// for commentary on the runtime of this function.
-        random: std.Random,
-        max_list_entries: usize,
-    ) Allocator.Error!EpochStakes {
-        var result_stakes = try Stakes(.delegation).initRandom(allocator, random, max_list_entries);
-        errdefer result_stakes.deinit(allocator);
+        pub fn default() Self {
+            return .{ .stakes = RwMux(StakesT).init(StakesT.DEFAULT) };
+        }
```
dnut:
```suggestion
        pub const empty = Self{
            .stakes = RwMux(StakesT).init(StakesT.DEFAULT),
        };
```
also would rename StakesT.DEFAULT

src/accountsdb/db.zig
```diff
@@ -1967,26 +1974,37 @@ pub const AccountsDB = struct {
...
+        std.debug.assert(pubkeys.len > 0);
+
         const slot_ref_map, var lock = self.account_index.slot_reference_map.writeWithLock();
         defer lock.unlock();
 
         const slot_entry = try slot_ref_map.getOrPut(slot);
-        if (!slot_entry.found_existing) slot_entry.value_ptr.* = .{};
+        if (!slot_entry.found_existing) slot_entry.value_ptr.* = .{
+            .refs = .{},
+            .global_index = undefined,
```
dadepo: More of a question. Will it be better to have .global_index as an optional instead?

src/core/stake.zig
```diff
@@ -1,828 +1,869 @@
...
 
         const Self = @This();
-        pub const DelegationsMap = std.AutoArrayHashMapUnmanaged(Pubkey, Element);
 
-        pub fn deinit(stakes: Self, allocator: Allocator) void {
-            var copy = stakes;
-            copy.vote_accounts.deinit(allocator);
-            copy.delegations.deinit(allocator);
-            copy.history.deinit(allocator);
+        pub const DEFAULT: Self = .{
```
dadepo: I think `empty` is the convention now

```suggestion
        pub const EMPTY: Self = .{
```

src/core/stake.zig
```diff
@@ -1,828 +1,869 @@
...
-        ) Allocator.Error!Self {
-            const vote_accounts = try stakes.vote_accounts.clone(allocator);
-            errdefer vote_accounts.deinit(allocator);
+        pub fn clone(self: *const Self, allocator: Allocator) Allocator.Error!Self {
+            const vote_accs = try self.vote_accounts.clone(allocator);
+            errdefer vote_accs.deinit(allocator);
 
-            var delegations = try stakes.delegations.clone(allocator);
-            errdefer delegations.deinit(allocator);
+            var stake_delegations = std.AutoArrayHashMapUnmanaged(Pubkey, T){};
```
dadepo:
```suggestion
            var stake_delegations = std.AutoArrayHashMapUnmanaged(Pubkey, T).empty;
```

src/core/stake.zig
```diff
@@ -1,828 +1,869 @@
...
+            errdefer {
+                if (is_account_type) for (stake_delegations.values()) |*v| v.deinit(allocator);
```
dadepo, line 167: A comment here might be helpful as the reason why the deinit is only being called when account type is `.account`

src/core/stake.zig
```diff
@@ -1,828 +1,869 @@
...
-pub fn Stakes(comptime delegation_type: enum {
+pub const StakesType = enum {
     delegation,
     stake,
-}) type {
-    const Element = switch (delegation_type) {
+    account,
+};
+
+// pub const Stakes = StakesGeneric(.stake);
```
dadepo: to delete?
```suggestion

```

src/core/stake.zig
```diff
@@ -1,828 +1,869 @@
...
+        pub fn deinit(self: *Self, allocator: Allocator) void {
+            var stakes: *StakesT, var stakes_guard = self.stakes.writeWithLock();
+            defer stakes_guard.unlock();
+            stakes.deinit(allocator);
+        }
 
-        var epoch_authorized_voters =
-            try epochAuthorizedVotersRandom(allocator, random, max_list_entries);
-        errdefer epoch_authorized_voters.deinit(allocator);
+        pub fn checkAndStore(
```
dadepo, line 48: Nit: ref to agave

dadepo, line 48: What would be the argument of taking the lock at the beginning and using that through the function?

An attempt to flatten the if/else a bit

```
        pub fn checkAndStore(
            self: *Self,
            allocator: Allocator,
            pubkey: Pubkey,
            account: AccountSharedData,
            new_rate_activation_epoch: ?Epoch,
        ) Allocator.Error!void {
            var stakes: *StakesT, var stakes_guard = self.stakes.writeWithLock();
            defer stakes_guard.unlock();

            // Handle zero lamports
            if (account.lamports == 0) {
                if (vote_program.ID.equals(&account.owner)) {
                    stakes.removeVoteAccount(allocator, pubkey);
                } else if (stake_program.ID.equals(&account.owner)) {
                    stakes.removeStakeAccount(allocator, pubkey, new_rate_activation_epoch);
                }
                return;
            }

            // Handle vote accounts
            if (vote_program.ID.equals(&account.owner)) {
                if (!VersionedVoteState.isCorrectSizeAndInitialized(account.data)) {
                    stakes.removeVoteAccount(allocator, pubkey);
                    return;
                }

                const vote_account = VoteAccount.fromAccountSharedData(
                    allocator,
                    try account.clone(allocator),
                ) catch {
                    stakes.removeVoteAccount(allocator, pubkey);
                    return;
                };
                try stakes.upsertVoteAccount(
                    allocator,
                    pubkey,
                    vote_account,
                    new_rate_activation_epoch,
                );
                return;
            }

            // Handle stake accounts
            if (stake_program.ID.equals(&account.owner)) {
                const stake_account = StakeAccount.fromAccountSharedData(
                    allocator,
                    try account.clone(allocator),
                ) catch {
                    stakes.removeStakeAccount(allocator, pubkey, new_rate_activation_epoch);
                    return;
                };
                try stakes.upsertStakeAccount(
                    allocator,
                    pubkey,
                    stake_account,
                    new_rate_activation_epoch,
                );
            }
        }
```

src/runtime/sysvar/stake_history.zig
```diff
@@ -11,7 +11,7 @@ const Pubkey = sig.core.Pubkey;
 
 /// [agave] https://github.com/anza-xyz/agave/blob/8db563d3bba4d03edf0eb2737fba87f394c32b64/sdk/sysvar/src/stake_history.rs#L67
 pub const StakeHistory = struct {
-    entries: std.ArrayListUnmanaged(Entry),
+    entries: std.ArrayListUnmanaged(Entry) = .{},
```
dadepo:
```suggestion
    entries: std.ArrayListUnmanaged(Entry) = .empty,
```

src/runtime/sysvar/stake_history.zig
```diff
@@ -46,6 +58,8 @@ pub const StakeHistory = struct {
 
     pub const SIZE_OF: u64 = 16_392;
 
+    pub const EMPTY: StakeHistory = .{ .entries = .{} };
```
dadepo:
```suggestion
    pub const EMPTY: StakeHistory = .{ .entries = .empty };
```

src/runtime/program/vote/state.zig
```diff
@@ -842,11 +882,66 @@ pub const VoteState = struct {
...
+            .last_timestamp = self.last_timestamp,
+        };
+    }
+
+    pub fn equals(self: *const VoteState, other: *const VoteState) bool {
+        if (self.votes.items.len != other.votes.items.len) return false;
+        for (self.votes.items, other.votes.items) |a, b|
+            if (!std.meta.eql(a, b)) return false;
+
+        if (self.voters.count() != other.voters.count()) return false;
```
dadepo: What do you think about moving this to the ``AuthorizedVoters`` type and have it used in here, similar to the `equals` on `CircBufV1`?

src/core/vote_accounts.zig
```diff
@@ -0,0 +1,848 @@
...
+    }
+
+    pub fn getDelegatedStake(self: *const VoteAccounts, pubkey: Pubkey) u64 {
+        const entry = self.vote_accounts.getPtr(pubkey) orelse return 0;
+        return entry.stake;
+    }
+
+    /// NOTE: If we make `staked_nodes` nullable, this function should be changed to compute
+    /// the staked nodes if `staked_nodes` is null.
+    pub fn getStakedNodes(self: *const VoteAccounts, allocator: Allocator) !*const StakedNodesMap {
```
dadepo: Maybe the method is not needed all together for now? Given that there is a preference to avoid these kind of getter's that does nothing more but access fields.

src/core/vote_accounts.zig
```diff
@@ -0,0 +1,848 @@
...
+    /// If the vote account is new, it will calculate the stake and add it to the `staked_nodes` map.
+    /// If the vote account already exists, and the node pubkey has changed, it will move the stake
+    /// from the old node to the new node in the `staked_nodes` map.
+    /// Takes ownership of `account` and returns the previous value if it existed.
+    pub fn insert(
+        self: *VoteAccounts,
+        allocator: Allocator,
+        pubkey: Pubkey,
+        account: VoteAccount,
+        // TODO: pass deps to compute stake
```
dadepo: Can be done now?

dadepo (commented):
Halfway done. Looks good. Just some comments/question. There is also the general observation of missing ref to agave in some key places

src/core/vote_accounts.zig
```diff
@@ -0,0 +1,848 @@
...
+        else
+            entry.value_ptr.* = stake;
+    }
+
+    /// Subtracts `stake` from an entry in `staked_nodes`. If the entry does not exist, it will panic.
+    fn subNodeStake(self: *VoteAccounts, pubkey: Pubkey, stake: u64) void {
+        if (stake == 0) return;
+
+        const current_stake = self.staked_nodes.getPtr(pubkey) orelse
+            @panic("pubkey not present in staked_nodes");
```
dadepo: Should the panics also be copied over? Any reason not to replace these with errors?

src/core/epoch_stakes.zig
```diff
@@ -0,0 +1,219 @@
...
+}
+
+pub fn epochStakeMapRandom(
+    allocator: Allocator,
+    random: std.Random,
+    comptime stakes_type: StakesType,
+    min_list_entries: usize,
+    max_list_entries: usize,
+) Allocator.Error!EpochStakesMap(stakes_type) {
+    var map: EpochStakesMap(stakes_type) = .{};
```
dadepo:
```suggestion
    var map: EpochStakesMap(stakes_type) = .empty;
```

dadepo (changes requested):
Looks good. Some req for change

src/core/stakes_cache.zig
yewman: lets leave this for now and re-evaluate its simplification later

src/core/stake.zig
yewman: made it `init` inline with changes to sysvars PR

src/accountsdb/db.zig
yewman: this diff should not be here, it is part of https://github.com/Syndica/sig/pull/796
@Sobeston

src/core/stake.zig
yewman: Why hold the lock for longer than needed? Or even worse acquire it without the needing to

yewman: a59c79fa

src/core/stake.zig
yewman: a59c79fa

src/core/stake.zig
yewman: a59c79fa

src/core/stake.zig
yewman: a59c79fa

src/core/stake.zig
yewman: a59c79fa

src/runtime/sysvar/stake_history.zig
yewman: a59c79fa

src/runtime/sysvar/stake_history.zig
yewman: a59c79fa

src/core/vote_accounts.zig
yewman: a59c79fa

src/core/vote_accounts.zig
yewman: a59c79fa

src/core/epoch_stakes.zig
yewman: a59c79fa

src/core/vote_accounts.zig
yewman: 2dada143

src/runtime/program/vote/state.zig
yewman: 6ba0b15b

src/core/stake.zig
dadepo: So, that question came from the observation it seems ~all~ most of the if/else branches included acquiring the locks, and perhaps it might be more efficient to acquire it once.

Keeping the re-acquiring of the locks it may be still possible to flatten this a bit

```
        pub fn checkAndStore(
            self: *Self,
            allocator: Allocator,
            pubkey: Pubkey,
            account: AccountSharedData,
            new_rate_activation_epoch: ?Epoch,
        ) !void {
            if (account.lamports == 0) {
                if (vote_program.ID.equals(&account.owner)) {
                    var stakes, var stakes_guard = self.stakes.writeWithLock();
                    defer stakes_guard.unlock();
                    try stakes.removeVoteAccount(allocator, pubkey);
                } else if (stake_program.ID.equals(&account.owner)) {
                    var stakes: *StakesT, var stakes_guard = self.stakes.writeWithLock();
                    defer stakes_guard.unlock();
                    try stakes.removeStakeAccount(allocator, pubkey, new_rate_activation_epoch);
                }
                return;
            }

            // Handle vote accounts
            if (vote_program.ID.equals(&account.owner)) {
                if (!VersionedVoteState.isCorrectSizeAndInitialized(account.data)) {
                    var stakes: *StakesT, var stakes_guard = self.stakes.writeWithLock();
                    defer stakes_guard.unlock();
                    try stakes.removeVoteAccount(allocator, pubkey);
                    return;
                }

                const vote_account = VoteAccount.fromAccountSharedData(
                    allocator,
                    try account.clone(allocator),
                ) catch {
                    var stakes: *StakesT, var stakes_guard = self.stakes.writeWithLock();
                    defer stakes_guard.unlock();
                    try stakes.removeVoteAccount(allocator, pubkey);
                    return;
                };

                var stakes: *StakesT, var stakes_guard = self.stakes.writeWithLock();
                defer stakes_guard.unlock();
                try stakes.upsertVoteAccount(
                    allocator,
                    pubkey,
                    vote_account,
                    new_rate_activation_epoch,
                );
                return;
            }

            // Handle stake accounts
            if (stake_program.ID.equals(&account.owner)) {
                const stake_account = StakeAccount.fromAccountSharedData(
                    allocator,
                    try account.clone(allocator),
                ) catch {
                    var stakes: *StakesT, var stakes_guard = self.stakes.writeWithLock();
                    defer stakes_guard.unlock();
                    try stakes.removeStakeAccount(allocator, pubkey, new_rate_activation_epoch);
                    return;
                };

                var stakes: *StakesT, var stakes_guard = self.stakes.writeWithLock();
                defer stakes_guard.unlock();
                try stakes.upsertStakeAccount(
                    allocator,
                    pubkey,
                    stake_account,
                    new_rate_activation_epoch,
                );
            }
        }
```

dadepo (approved):

src/core/stake.zig
```diff
@@ -1,828 +1,1105 @@
...
-            try allocator.alloc(Pubkey, random.uintLessThan(usize, max_list_entries));
-        errdefer allocator.free(vote_accounts);
-        for (vote_accounts) |*vote_account| vote_account.* = Pubkey.initRandom(random);
-        return .{
-            .vote_accounts = vote_accounts,
-            .total_stake = random.int(u64),
-        };
+    pub fn getDelegation(self: StakeAccount) Delegation {
+        return self.state.getDelegation() orelse
+            @panic("StakeAccount does not have a delegation");
```
kprotty: are the panics for testing, or possible codepaths on misconfigured user input?

yewman: `StakeAccount` should only be constructed using `fromAccountSharedData` which requires that the deserialised state has a Delegation, and thus also a Stake. So if someone manually constructs an invalid StakeAccount then this panic will be hit, but otherwise it will never be reached. Either I can change these to errors, add a comment on why the panic is okay, or store the the stake state variant `stake: struct { meta: Meta, stake: Stake, flags: StakeFlags },` directly instead of `StakeStateV2`. Also perhaps renaming `fromAccountSharedData` to `init` would make it more clear that this validation occurs on initialisation of the `StakeAccount`. Lmk your thoughts

src/core/stake.zig
yewman: The current implementation only ever acquires the stakes cache lock once. I have flattened the vote program block as you suggested  6a4c23fa

src/core/stake.zig
kprotty: I'd vote on returning error (or optional) as then it can be handled at the callsite which provides more context on why a Delegation is a hard requirement from the parsed StakeStateV2. Thoughts?

yewman: I'd prefer to eliminate errors if possible, thoughts on this approach: 41c429a4

dnut (approved):

dadepo (approved):

PR #810, opened 2025-07-11
feat(replay): Initialize consensus state

InKryption:
Note: at the time of writing, there are changes included that remove the RwMux wrappers from the slot and epoch trackers, and in general it's currently presuming single-threaded execution, and this is largely for simplicity. Later on when the vote listener PR stack is merged, I'll devise a strategy for synchronizing access properly, with a full view of the design, in such a way that it leaves open to us the possibility of merging the thread of execution of the vote listener and replay, rather than locking us into the architecture prescribed by agave.


src/accountsdb/account_store.zig
```diff
@@ -211,3 +211,81 @@ pub const ThreadSafeAccountMap = struct {
...
+    try expectAccount(account_reader, addr1, ancestors1, sharedToCoreAccount(expected_account));
+}
+
+fn expectAccount(
+    account_reader: AccountReader,
+    address: Pubkey,
+    maybe_ancestors: ?Ancestors,
+    expected: ?sig.core.Account,
+) !void {
+    const actual = if (maybe_ancestors) |ancestors|
```
dadepo:
```suggestion
    if (!@import("builtin").is_test) @compileError("Not allowed outside of tests.");
    const actual = if (maybe_ancestors) |ancestors|
```

src/cmd.zig
```diff
@@ -1080,32 +1080,32 @@ fn validator(
...
+        ledger_db,
         app_base.metrics_registry,
         lowest_cleanup_slot,
         max_root,
     );
 
-    const ledger_result_writer = try allocator.create(LedgerResultWriter);
-    defer allocator.destroy(ledger_result_writer);
-    ledger_result_writer.* = try LedgerResultWriter.init(
+    const ledger_writer = try allocator.create(LedgerResultWriter);
```
dadepo: I vaguely remember there was an argument to having the variable include `result` something around it is the results of the execution that is being persisted. Maybe @dnut remembers.

Based off that, this renaming, may or may not be necessary.

src/cmd.zig
```diff
@@ -1200,17 +1206,26 @@ fn validator(
             sig.replay.service.run,
             .{sig.replay.service.ReplayDependencies{
                 .allocator = allocator,
-                .logger = app_base.logger.unscoped(),
-                .my_identity = .{ .data = app_base.my_keypair.public_key.bytes },
+                .logger = .from(app_base.logger),
+                .my_identity = .fromPublicKey(&app_base.my_keypair.public_key),
+                .vote_identity = .fromPublicKey(&app_base.my_keypair.public_key), // TODO: is this fine, or do we need a separate identity for the vote account?
```
dadepo, line 1229: I think there is the need to have separate keys for the identity and voting. Although in practice the operator may choose to use the same keys. According to https://github.com/anza-xyz/agave/blob/07c4a88cdbac5dc39010edea022ec3c724214d1e/docs/src/operations/guides/vote-accounts.md?plain=1#L74-L75

src/consensus/replay_tower.zig
```diff
@@ -135,7 +134,7 @@ pub const ReplayTower = struct {
         node_pubkey: Pubkey,
         vote_account_pubkey: Pubkey,
         fork_root: Slot,
-        accounts_db: *AccountsDB,
+        accounts_db: sig.accounts_db.AccountReader,
```
dadepo: nit:

```suggestion
        accounts_reader: AccountReader,
```

With AccountReader aliased properly at the top of the file

src/consensus/tower.zig
```diff
@@ -92,11 +92,23 @@ pub const Tower = struct {
         allocator: std.mem.Allocator,
         vote_account_pubkey: *const Pubkey,
         fork_root: Slot,
-        accounts_db: *AccountsDB,
+        accounts_db: sig.accounts_db.AccountReader,
```
dadepo: Same suggestion regarding naming and shortening the import with an aliases.

src/core/bank.zig
```diff
@@ -284,14 +284,17 @@ pub const EpochConstants = struct {
...
+        bank_fields: *const BankFields,
+    ) Allocator.Error!EpochConstants {
         return .{
             .hashes_per_tick = bank_fields.hashes_per_tick,
             .ticks_per_slot = bank_fields.ticks_per_slot,
             .ns_per_slot = bank_fields.ns_per_slot,
             .genesis_creation_time = bank_fields.genesis_creation_time,
             .slots_per_year = bank_fields.slots_per_year,
-            .stakes = bank_fields.epoch_stakes,
+            .stakes = try bank_fields.epoch_stakes.get(bank_fields.epoch).?.clone(allocator), // TODO: is this the right thing to do?
```
dadepo: bumping the TODO

src/core/hash.zig
```diff
@@ -10,6 +10,10 @@ pub const SlotAndHash = struct {
     slot: Slot,
     hash: Hash,
 
+    pub fn tuple(self: SlotAndHash) struct { Slot, Hash } {
```
dadepo, line 13: I have generally found the named field easier to work with than the tuple which require indices to access. hence the question: what will you say is the benefit of this?

Also it does not look like it is used anywhere.

src/core/vote_accounts.zig
```diff
@@ -341,12 +341,12 @@ pub const VoteAccount = struct {
...
-            node_pubkey orelse Pubkey.initRandom(random),
-            Pubkey.initRandom(random),
-            Pubkey.initRandom(random),
+            node_pubkey orelse .initRandom(random),
+            .initRandom(random),
+            .initRandom(random),
             random.int(u8),
             random.intRangeAtMost(u64, 1, 1_000_000),
-            Clock.initRandom(random),
+            .initRandom(random),
```
dadepo, line 361: I have an ambivalent feelings towards the "dot" syntax. 

I think it prioritize writing over reading code and it may hurt readability. This diff is an instance of such. Previously it is clear what is being initiated...ie `Clock`, `Pubkey` etc but now it is only a bunch of `.initRandom` calls

src/accountsdb/account_store.zig
InKryption: 72f17305c109c40770c3b300bb08dd9aa8643127

src/core/hash.zig
InKryption: It actually is used, specifically in `edge_cases.processAncestorHashesDuplicateSlots`.
To answer: as demonstrated in that function it is just an easy way to have the slot hash pair named one thing, and then be able to also extract it into a slot and hash variable in one line, with the static assurance that you're extracting all the data.

src/core/vote_accounts.zig
InKryption: I don't think it is a particularly important detail, no more than, for example, the names of some of the previous parameters, which we otherwise only know to be a bunch of random bytes/integers.

src/consensus/replay_tower.zig
InKryption: I'm not convinced the alias would be that helpful; it is used exactly once in this entire file.
55a708f174152d359d406ff5ed95b41c0b8c30ac

src/consensus/tower.zig
InKryption: Same response to other comment 55a708f174152d359d406ff5ed95b41c0b8c30ac

src/core/hash.zig
dadepo: I see it now. Don't know why the search usage did not pick this when I tried earlier.

src/cmd.zig
dadepo: On second thought, I think this should be resolved, either by updating the code, or confirming this is fine.

Reason being that, if it is not fine, then this is probably introducing an issue. An operator may set up a vote identity with the expectation this would be used to sign the transaction, but the validator identity will be used instead.

I'll dismiss my approval based on the above.

src/replay/service.zig
```diff
@@ -31,128 +39,471 @@ const MAX_ENTRIES: u64 = 1024 * 1024; // 1 million slots is about 5 days
 const DUPLICATE_LIVENESS_THRESHOLD: f64 = 0.1;
 pub const DUPLICATE_THRESHOLD: f64 = 1.0 - SWITCH_FORK_THRESHOLD - DUPLICATE_LIVENESS_THRESHOLD;
 
+pub const Logger = sig.trace.ScopedLogger("replay");
+
 pub const ReplayDependencies = struct {
```
dadepo: Nit: I think a comment on here would be useful to distinguish the difference between this (`ReplayDependencies`)  and `ReplayState` - something along the line of: ReplayState is initialized based off `ReplayDependencies` and `ReplayState` is what the various operations during replay updates - Although the `senders` and `receivers` does not fit that mould perfectly.

src/replay/service.zig
```diff
@@ -31,128 +39,471 @@ const MAX_ENTRIES: u64 = 1024 * 1024; // 1 million slots is about 5 days
...
+        //     - from validator [to tvu](https://github.com/anza-xyz/agave/blob/0315eb6adc87229654159448344972cbe484d0c7/core/src/validator.rs#L1486)
+        //     - from tvu to [replay_config](https://github.com/anza-xyz/agave/blob/0315eb6adc87229654159448344972cbe484d0c7/core/src/tvu.rs#L311)
+        //     - replay_config to [ReplayStage](https://github.com/anza-xyz/agave/blob/0315eb6adc87229654159448344972cbe484d0c7/core/src/replay_stage.rs#L563)
+        const replay_tower: sig.consensus.ReplayTower = try .init(
+            allocator,
+            .from(deps.logger),
+            deps.my_identity,
+            deps.vote_identity,
+            deps.root.slot,
+            deps.account_store.reader(),
```
dadepo: Is the account data used to init the tower guaranteed to be the one associated with the heaviest based off consensus?

If not it seems this needs to be the case based off the logic in Agave here https://github.com/anza-xyz/agave/blob/7baf884e6b87045b3b9b2af1e77650a58c25ae90/core/src/consensus.rs#L369-L386

The tower is init based on the heaviest bank gotten from fork choice.

src/replay/service.zig
InKryption: I'm not sure that's necessary, `ReplayState` is a structure which is private and scoped to service.zig, and the names appear to illustrate this dichotomy already. Is there a situation you forsee in which they would be confused?

src/replay/service.zig
InKryption: The actual code for this is a fair bit more complicated actually, I forgot to flag this specifically. The actual initialization in the validator is here:
https://github.com/anza-xyz/agave/blob/0315eb6adc87229654159448344972cbe484d0c7/core/src/validator.rs#L1451-L1463
which attempts to restore from an existing tower, but otherwise falls into this codepath:
https://github.com/anza-xyz/agave/blob/0315eb6adc87229654159448344972cbe484d0c7/core/src/validator.rs#L2209-L2228
which, under what I assume to be normal conditions, will follow to this codepath, calling into the function you've referenced:
https://github.com/anza-xyz/agave/blob/0315eb6adc87229654159448344972cbe484d0c7/core/src/validator.rs#L1948
which, iiuc, appears to be using the same unmodified bank_forks if you follow the call-tree, which indicate to me that in our current context, it should be equivalent.

src/core/vote_accounts.zig
dnut: In this case, I agree with trevor, but in the general case, I agree with dadepo. There's definitely information loss which can cause confusion in some cases. But I also appreciate having the option to use this syntax. It makes it a lot easier to *write* code (though readability is more important than writability), plus it tidies things up. Sometimes the lost information is just redundant and pointless, and it's better to reduce the noise.

src/cmd.zig
dnut: There are two ledger writers:
- ShredInserter: Writes the actual shreds/blocks/transactions to the ledger
- LedgerResultWriter: Writes the results of executing blocks/transactions

When it's implied by the context, maybe it would be fine to call either of these a "ledger writer." But in a general context like cmd.zig, I think it's worth distinguishing them.

src/cmd.zig
dnut: Ideally we would support the same distinction as agave but I don't think it's strictly necessary. The identity and the vote authority can be the same key. It's a nice-to-have to allow the operator to use different keypairs, but we can live without it for now.

I was initially confused by the word "identity." I think "vote_authority" would be a better name

src/core/bank.zig
dnut: This is not correct. The correct implementation is more complicated, and I've implemented it in another branch. I would expect this to panic if you run it on any cluster

dnut: I would expect this to panic if you run it on any cluster

src/replay/execution.zig
```diff
@@ -4,8 +4,6 @@ const replay = @import("lib.zig");
 
 const core = sig.core;
 
-const Allocator = std.mem.Allocator;
```
dnut: Why did you remove this?

src/replay/service.zig
```diff
@@ -2,27 +2,32 @@ const std = @import("std");
...
 const replay = @import("lib.zig");
 
-const Allocator = std.mem.Allocator;
+const collections = sig.utils.collections;
 
+const Allocator = std.mem.Allocator;
 const ThreadPool = sig.sync.ThreadPool;
 
-const AccountStore = sig.accounts_db.account_store.AccountStore;
-const SlotAccountReader = sig.accounts_db.account_store.SlotAccountReader;
```
dnut: I noticed you remove a lot of imports and replace them with fully qualified usages throughout the code. I think the code is clearer when each file lists its dependencies on top. I'd rather not have fully qualified usages littered throughout the code because it's not as easily discoverable.

src/replay/service.zig
```diff
@@ -162,16 +510,30 @@ fn advanceReplay(state: *ReplayState) !void {
...
+        .slot_tracker = &state.slot_tracker,
+        .progress = &state.progress,
+        .latest_validator_votes = &state.latest_validator_votes,
+        .slot_data = &state.slot_data,
+
+        .senders = state.senders,
+        .receivers = state.receivers,
+    });
 
     try replay.consensus.processConsensus(null);
```
dnut: Why is this null? Can you add the inputs here? Finishing processConsensus is the top priority for consensus.

src/replay/service.zig
```diff
@@ -31,128 +36,471 @@ const MAX_ENTRIES: u64 = 1024 * 1024; // 1 million slots is about 5 days
...
+
+            const htv_gop = try hash_to_votes.getOrPut(allocator, hash);
+            errdefer if (!htv_gop.found_existing) {
+                std.debug.assert(hash_to_votes.swapRemove(htv_gop.key_ptr.*));
+            };
+
+            try htv_gop.value_ptr.append(allocator, vote_pubkey);
+        }
+    }
 };
```
dnut: Is any of this going to be used outside edge_cases.zig? I think anything that's exclusive to the edge cases code makes sense to live within edge_cases.zig, just like ConsensusDependencies and ReplayExecutionState live in their respective files.

dnut (approved):

dnut (changes requested):
accidentally clicked approve on the last review

src/replay/service.zig
InKryption: I find imports help when they are few and used frequently; the sheer mass of imports in many files does not give me any extra context I can work with, and more often than not seems to just lead to inconsistent usage, where a person unfamiliar with a particular part of the code fully qualifies the type rather than using the existing import, because the existing import is effectively invisible in the huge header at the top of the file.

src/replay/execution.zig
InKryption: Was removed by the style script, probably because I replaced the few instances of Allocator with std.mem.Allocator while I was editing some of the code, because I just tend to fully qualify the allocator type.

src/replay/service.zig
InKryption: 11cf999d1533c62ed1bd71e50f090d98e3734d9d

src/replay/service.zig
InKryption: I moved this originally for some specific organizational limitation I encountered, but isn't a problem now after something I did.

InKryption: Though to reply to your question directly: at least in agave, `replay_active_banks` mutates a lot of the structures in this `SlotData` struct, including unfrozen gossip verified vote hashes. Would you still prefer it be in edge cases?

src/replay/service.zig
dnut: > the sheer mass of imports in many files does not give me any extra context I can work with

It definitely gives me context, especially if it's well organized.

We can leave the code as-is for now. Importing style is not our top priority, and with you pivoting to the transaction harness I'd just like to get this merged. We have a lot of disagreement about how to import, which means we need to spend a while learning about each other's values. Ultimately I think we should standardize our approach, but we can revisit it later to make a decision.

src/replay/execution.zig
dnut: Allocator is the one type that I'm always super annoyed about when it doesn't already have an import in a file. Anyway don't worry about it.

src/replay/service.zig
dnut: If it's used in the broader context of replay, then I think it's reasonable to keep it defined here. I just mean anything that's used exclusively in edge_cases.zig (except for initialization and passing down) should probably live in edge_cases.zig. But I'm not sure about how everything is used.

src/replay/service.zig
dnut: I also feel that the names speak for themselves. But I also like the idea of every struct having a doc comment. I'm fine with either way and would lean towards simply resolving this conversation so we can move on.

Dadepo if you have an objection to that, please reply soon so we can address it.

Trevor, if you have nothing else to do, feel free to add a doc to avoid more back and forth.

src/replay/service.zig
InKryption: I'm not too sure either; I think it'll be easier to just have it live in service for now, and if it's still only used by edge_cases as a whole, then we can move it there - let practice define our decision.

dnut: Either way, we're making a decision now. It's just an arbitrary, ill-informed decision that we'll likely change in the future. So we don't need to waste a lot of time trying to perfect it right now.

Personally, I prefer to keep things in the minimal scope where they are needed, and then broaden their scope as the need arises. I also have an aversion to throwing random things in a big catch-all bucket because it usually ends up turning into a huge mess. I prefer to keep things organized and segregated from the beginning because it's easier to combine than it is to separate.

src/replay/service.zig
```diff
@@ -535,7 +549,60 @@ fn advanceReplay(state: *ReplayState) !void {
...
+            const descendants_gop = try descendants.getOrPutValue(arena, ancestor_slot, .empty);
+            try descendants_gop.value_ptr.put(arena, slot);
+        }
+    }
+
+    const SlotHistory = sig.runtime.sysvar.SlotHistory;
+    const slot_history: SlotHistory = sh: {
+        const account_reader = state.account_store.reader();
+
+        const maybe_account = try account_reader.getLatest(SlotHistory.ID); // TODO: do we need to scope this with `forSlot` or anything?
```
dnut: `getLatest` is usually not what you want, and you should never use it in replay. In replay, we only ever care about the state for a particular slot. Actually I'm not sure if `getLatest` is ever appropriate, and maybe we should just remove it. The only reason I added `getLatest` is because it's the equivalent of the original `getAccount` function we had in AccountsDB. But then we realized that's not the behavior we want, which is why Sebastian added `getAccountWithAncestors`, which is the equivalent of agave's `AccountsDB::load`

Also, in the future, I think we should probably just get this from the sysvar cache, not accountsdb. But that'll require some changes that I have in another branch. I would add a TODO here to say switch to getting this from the sysvar cache.

src/replay/service.zig
InKryption: 8b1109e2e853d4ec17ec272b9d52992503369023

src/cmd.zig
dadepo: Given that the support of two keys is listed in the documentation of the only Solana client, and hence folks who may want to try out Sig would be familiar with that setup, could it be a point of confusion/contention? 

If not then I think, indeed it can be classified as a nice to have.

src/replay/service.zig
dadepo: I may be reading it wrong, but it seems the code path that does not end up calling `Tower::new_from_bankforks` end up doing the following instead:

- loading the tower from persistence https://github.com/anza-xyz/agave/blob/3e2d6a93cc9b0e313807657542663707a6edc165/core/src/validator.rs#L2231 (which we currently don't have)
- after loading, seems to do a bunch of stuff https://github.com/anza-xyz/agave/blob/3e2d6a93cc9b0e313807657542663707a6edc165/core/src/validator.rs#L1912-L1944 which I think the most relevant to the `tower` is `tower.adjust_lockouts_after_replay`

Looks to me like different operations?

src/cmd.zig
dadepo: It is okay to have the keys separated as it is now, but only accept one via the cli. Later we can extend the cli to support passing different keys for voting/identity

src/replay/service.zig
dadepo: What we have here should be analogous to instantiating the tower from the snapshot (where fork = 1) so no need to make use of the fork choice here.

src/cmd.zig
dadepo: Resolving this as this is a nit. If any improvement to the naming is needed it can be done later.

src/consensus/tower.zig
```diff
@@ -83,11 +83,23 @@ pub const Tower = struct {
...
         fork_root: Slot,
-        accounts_db: *AccountsDB,
+        account_reader: sig.accounts_db.AccountReader,
     ) !void {
-        const vote_account = accounts_db.getAccount(vote_account_pubkey) catch {
-            self.initializeRoot(fork_root);
-            return;
+        const vote_account = blk: {
+            const maybe_vote_account =
+                account_reader.getLatest(vote_account_pubkey.*) catch |err| switch (err) {
```
dadepo: Another place `getLatest` is being used that should be re-evaluated.

src/replay/service.zig
```diff
@@ -31,151 +36,492 @@ const MAX_ENTRIES: u64 = 1024 * 1024; // 1 million slots is about 5 days
...
+            );
+            errdefer slot_history.deinit(allocator);
+
+            if (data_iter.bytesRemaining() != 0) {
+                return error.TrailingBytesInSlotHistory;
+            }
+
+            break :sh slot_history;
+        };
+        return slot_history;
```
dnut: You can flatten this block to be the same as the function body with only a return and no break.

src/replay/service.zig
```diff
@@ -31,151 +36,492 @@ const MAX_ENTRIES: u64 = 1024 * 1024; // 1 million slots is about 5 days
...
 };
 
-pub const Logger = sig.trace.ScopedLogger("replay");
+pub const LedgerRef = struct {
+    db: BlockstoreDB,
+    reader: *sig.ledger.BlockstoreReader,
+    writer: *sig.ledger.LedgerResultWriter,
+};
+
+pub const SlotHistoryAccessor = struct {
```
dnut: My instinct would be to define this where it's used (replay_tower.zig). What do you think?

src/consensus/replay_tower.zig
```diff
@@ -1374,10 +1367,24 @@ pub const ReplayTower = struct {
...
 
+        const maybe_slot_history: ?SlotHistory =
+            if (heaviest_slot_on_same_voted_fork) |heaviest| blk: {
+                var ancestors_for_heaviest = ancestors.get(heaviest) orelse
+                    return error.MissingAncestors;
+                var slot_ancestors = sig.core.Ancestors.EMPTY;
+                for (ancestors_for_heaviest.items()) |ancestor_slot| {
+                    try slot_ancestors.addSlot(allocator, ancestor_slot);
+                }
+                break :blk try slot_history_accessor.getSlotHistory(allocator, &slot_ancestors);
```
dnut: It's unfortunate that you need to allocate to clone the ancestors only to immediately discard it. Also, this leaks.

Ideally you could just use a pointer to the existing data. I see two options:
1. shallow-copy `ancestors_for_heaviest` into Ancestors - this has ambiguous ownership semantics, but since it's only passed by const pointer and then immediately discarded, it's not a big problem.
2. change the ancestors map at the source to be a map from Slot to Ancestors instead of a SlotSet, and get it with `getPtr`

My instinct would be 2 unless that's impractical for some reason, then maybe I would go with 1. 1 is kind of a hack, so I would leave a comment there about the ownership and lifetimes if you take that approach.

src/replay/service.zig
dadepo: [b6b24da](https://github.com/Syndica/sig/pull/810/commits/b6b24dac5e2e87745c604127f9e8f70b16419fbe)

src/replay/service.zig
dadepo: Actually had a TODO on this regarding what the best place might. `replay_tower.zig` sounds ok.

[d5af195](https://github.com/Syndica/sig/pull/810/commits/d5af1950c9c25c61124df41510c5ea749db4cd84)

src/consensus/replay_tower.zig
dadepo: Went with 2 as that feels more correct. Although it ended up being more intrusive [5c6268d](https://github.com/Syndica/sig/pull/810/commits/5c6268d3fe7b3d9d15ca39614b736cd436dd32ae)
[8da86ea](https://github.com/Syndica/sig/pull/810/commits/8da86ea2b708c4677405703cfc1abb61d0cb988d)

src/replay/service.zig
```diff
@@ -38,162 +37,435 @@ const MAX_ENTRIES: u64 = 1024 * 1024; // 1 million slots is about 5 days
...
+            .num_blocks_on_fork = 0,
+            .num_dropped_blocks_on_fork = 0,
+        }));
+    }
+
+    // Given a root and a list of `frozen_slots` sorted smallest to greatest by slot,
+    // initialize a new HeaviestSubtreeForkChoice
+    //
+    // Analogous to [`new_from_frozen_banks`](https://github.com/anza-xyz/agave/blob/0315eb6adc87229654159448344972cbe484d0c7/core/src/consensus/heaviest_subtree_fork_choice.rs#L235)
+    var heaviest_subtree_fork_choice = fork_choice: {
```
dadepo, line 350: Might be an idea to move this to a function on 

I can see this functionality being used in two places in Agave

https://github.com/anza-xyz/agave/blob/0d78d8f4814ff7f8d4b9152beed2839a397991c7/core/src/consensus.rs#L370

and here

https://github.com/anza-xyz/agave/blob/0d78d8f4814ff7f8d4b9152beed2839a397991c7/core/src/consensus/heaviest_subtree_fork_choice.rs#L256-L270

Can be a follow up pr.

src/cmd.zig
```diff
@@ -1113,7 +1113,7 @@ fn validator(
     );
 
     var cleanup_service_handle = try std.Thread.spawn(.{}, sig.ledger.cleanup_service.run, .{
-        app_base.logger.unscoped(),
+        sig.ledger.cleanup_service.Logger.from(app_base.logger),
```
InKryption: should just be `.from(app_base.logger)`.

src/replay/execution.zig
```diff
@@ -37,14 +39,42 @@ pub const ReplayExecutionState = struct {
...
+            .vote_account = null, // voting not currently supported
+            .account_store = account_store,
+            .thread_pool = thread_pool,
+            .ledger_reader = ledger_reader,
+            .slot_tracker = slot_tracker,
+            .epochs = epochs,
+            .progress_map = progress_map,
+            .status_cache = .DEFAULT,
+        };
+    }
```
InKryption: Was this accidentally added back? Take note that this execution state does not actually own any state, it is merely a reference to data in the main replay state struct.

src/replay/execution.zig
```diff
@@ -37,14 +39,42 @@ pub const ReplayExecutionState = struct {
...
 };
 
 /// 1. Replays transactions from all the slots that need to be replayed.
 /// 2. Store the replay results into the relevant data structures.
 ///
 /// Analogous to [replay_active_banks](https://github.com/anza-xyz/agave/blob/3f68568060fd06f2d561ad79e8d8eb5c5136815a/core/src/replay_stage.rs#L3356)
-pub fn replayActiveSlots(state: ReplayExecutionState) !bool {
+pub fn replayActiveSlots(state: *ReplayExecutionState) !bool {
```
InKryption: Related to the other comment on ReplayExecutionState, it doesn't own anything, why is this now a mutable pointer?

src/replay/execution.zig
```diff
@@ -145,7 +175,7 @@ const ReplaySlotStatus = union(enum) {
 /// - [replay_active_bank](https://github.com/anza-xyz/agave/blob/161fc1965bdb4190aa2d7e36c7c745b4661b10ed/core/src/replay_stage.rs#L2979)
 /// - [replay_blockstore_into_bank](https://github.com/anza-xyz/agave/blob/161fc1965bdb4190aa2d7e36c7c745b4661b10ed/core/src/replay_stage.rs#L2232)
 /// - [confirm_slot](https://github.com/anza-xyz/agave/blob/161fc1965bdb4190aa2d7e36c7c745b4661b10ed/ledger/src/blockstore_processor.rs#L1494)
-fn replaySlot(state: ReplayExecutionState, slot: Slot) !ReplaySlotStatus {
+fn replaySlot(state: *ReplayExecutionState, slot: Slot) !ReplaySlotStatus {
```
InKryption: Related to the other comment on ReplayExecutionState, it doesn't own anything, why is this now a mutable pointer?

src/replay/service.zig
```diff
@@ -148,16 +155,18 @@ pub const Receivers = struct {
...
+    slot_tracker: *SlotTracker,
+    epochs: *EpochTracker,
+    hard_forks: sig.core.HardForks,
+    account_store: AccountStore,
+    progress_map: *ProgressMap,
     ledger: LedgerRef,
-    account_store: sig.accounts_db.AccountStore,
-
-    my_identity: Pubkey,
+    execution: ReplayExecutionState,
```
InKryption: It appears many changes I made to the replaystate struct were reverted here. This is supposed to be returned by a method instead, making this replaystruct state able to embed many fields directly without being self-referential.

InKryption (commented):
Some concerns @dadepo

src/consensus/tower.zig
dadepo: Since this is being used during initialisations, not sure it is possible to have ancestors at hand.

src/replay/service.zig
dadepo: [8685a8](https://github.com/Syndica/sig/pull/810/commits/48685a83009feff734747a457db0f03af6833a9c)

src/replay/execution.zig
dadepo: [8685a8](https://github.com/Syndica/sig/pull/810/commits/48685a83009feff734747a457db0f03af6833a9c)

src/replay/execution.zig
dadepo: [8685a8](https://github.com/Syndica/sig/pull/810/commits/48685a83009feff734747a457db0f03af6833a9c)

src/replay/execution.zig
dadepo: [8685a8](https://github.com/Syndica/sig/pull/810/commits/48685a83009feff734747a457db0f03af6833a9c)

src/cmd.zig
dadepo: [8685a8](https://github.com/Syndica/sig/pull/810/commits/48685a83009feff734747a457db0f03af6833a9c)

dadepo (dismissed):

src/replay/edge_cases.zig
```diff
@@ -138,30 +139,133 @@ pub fn processEdgeCases(
...
+    pub const DuplicateSlotsToRepair = std.AutoArrayHashMapUnmanaged(Slot, Hash);
+
+    /// Analogous to [PurgeRepairSlotCounter](https://github.com/anza-xyz/agave/blob/0315eb6adc87229654159448344972cbe484d0c7/core/src/repair/cluster_slot_state_verifier.rs#L20)
+    pub const PurgeRepairSlotCounters = collections.SortedMapUnmanaged(Slot, usize);
+
+    /// Analogous to [EpochSlotsFrozenSlots](https://github.com/anza-xyz/agave/blob/0315eb6adc87229654159448344972cbe484d0c7/core/src/repair/cluster_slot_state_verifier.rs#L22)
+    pub const EpochSlotsFrozenSlots = collections.SortedMapUnmanaged(Slot, Hash);
+
+    /// Analogous to [DuplicateConfirmedSlots](https://github.com/anza-xyz/agave/blob/0315eb6adc87229654159448344972cbe484d0c7/core/src/repair/cluster_slot_state_verifier.rs#L24)
+    pub const DuplicateConfirmedSlots = collections.SortedMapUnmanaged(Slot, Hash);
```
dnut, line 164: In my opinion, these aliases are pointless because they are redundant with the field and parameter names that they always go along with. I think they are actually harmful because they obscure the actual type with a layer of indirection.

I understand this code was only moved, and it's already like this on main. Just wanted to voice my thoughts. We don't need to block the pr for this.

src/consensus/tower.zig
dnut: Ancestors are in BankFields in the snapshot and loaded into SlotConstants before starting replay. This function could accept SlotAccountReader instead of AccountReader.

src/replay/service.zig
```diff
@@ -38,162 +41,438 @@ const MAX_ENTRIES: u64 = 1024 * 1024; // 1 million slots is about 5 days
...
+            &slot_tracker,
+            &epoch_tracker,
+            deps.my_identity,
+            deps.vote_identity,
+            deps.ledger.reader.*,
+        );
+        errdefer progress_map.deinit(deps.allocator);
+        errdefer fork_choice.deinit();
+
+        // TODO(ink): in agave replay_tower isn't created directly in replay,
```
dnut: This is not really a TODO because there's nothing we actually need to do. It's just an observation that agave has different requirements and a different design. I'm sure a lot of data is going to need to be shuffled around as sig's requirements evolve, but I don't think we can reasonably dictate what actually needs to be done at this point. It's helpful to have some notes in the code for this, but it's not really an actionable task at this time. I would change this to NOTE or unqualified.

dnut (changes requested):
almost there, just two (or three) things i would change

src/replay/service.zig
dadepo: [519e304](https://github.com/Syndica/sig/pull/810/commits/519e3047d02aa8e44a289b93c8b54dd8a8941668)

src/replay/edge_cases.zig
dadepo: I also have doubts about the usefulness of some the type aliases in the codebase, especially the ones that aliases collections like types from the std. Worth discussing.

dnut (dismissed):

dadepo (dismissed):

dadepo (approved):

PR #811, opened 2025-07-14
feat(consensus, forkchoice) Add more forkchoice methods

dadepo:
This PR adds missing methods on ForkChoice that is needed to implement the computeBankStats in replay.

It includes:
- computeBankStats on fork choice
   - needed addVotes
       - needed generateUpdateOperations
- slight refactor of splitOff
- port of some more tests
   - added addRootParent


src/consensus/fork_choice.zig
```diff
@@ -420,6 +456,48 @@ pub const ForkChoice = struct {
...
+            // The `heaviest_subtree_slot` and `deepest_slot` do not change
+            .deepest_slot = root_info.deepest_slot,
+            .heaviest_subtree_slot = root_info.heaviest_subtree_slot,
+            .children = root_parent_children,
+            .parent = null,
+            .latest_duplicate_ancestor = null,
+            .is_duplicate_confirmed = root_info.is_duplicate_confirmed,
+        };
+
+        try self.fork_infos.put(root_parent, root_parent_info);
```
InKryption: Refactor suggestion to avoid leak in error path, and to make use of `.from` for the logger:
```suggestion
        try self.fork_infos.ensureUnusedCapacity(1);
        var root_parent_children = SortedMap(SlotAndHash, void).init(self.allocator);
        try root_parent_children.put(self.tree_root, {});
        self.fork_infos.putAssumeCapacityNoClobber(root_parent, .{
            .logger = .from(self.logger),
            .stake_for_slot = 0,
            .stake_for_subtree = root_info.stake_for_subtree,
            .height = root_info.height + 1,
            // The `heaviest_subtree_slot` and `deepest_slot` do not change
            .deepest_slot = root_info.deepest_slot,
            .heaviest_subtree_slot = root_info.heaviest_subtree_slot,
            .children = root_parent_children,
            .parent = null,
            .latest_duplicate_ancestor = null,
            .is_duplicate_confirmed = root_info.is_duplicate_confirmed,
        });
```

src/consensus/fork_choice.zig
```diff
@@ -420,6 +456,48 @@ pub const ForkChoice = struct {
...
+    /// Analogous to [add_root_parent](https://github.com/anza-xyz/agave/blob/92b11cd2eef1d3f5434d6af702f7d7a85ffcfca9/core/src/consensus/heaviest_subtree_fork_choice.rs#L421)
+    pub fn addRootParent(self: *ForkChoice, root_parent: SlotAndHash) !void {
+        // Assert that the new root parent has a smaller slot than the current root
+        std.debug.assert(root_parent.slot < self.tree_root.slot);
+        // Assert that the root parent doesn't already exist
+        std.debug.assert(!self.fork_infos.contains(root_parent));
+
+        // Get the current root's fork info
+        const root_info = self.fork_infos.getPtr(self.tree_root) orelse
+            return error.MissingForkInfo;
```
InKryption: I think this should either just be an assertion like the other assumptions, or the other assumptions should become errors instead of assertions - personally I prefer the former, but I leave that to your judgement. Also it would be good to have this as part of an explicit error set, ie `(std.mem.Allocator.Error || error{MissingFork})!void`, or just `std.mem.Allocator.Error!void` if you go the assertion route.

src/consensus/fork_choice.zig
```diff
@@ -857,6 +935,160 @@ pub const ForkChoice = struct {
...
+            // If the new vote is for the same slot, but a different, smaller hash,
+            // then allow processing to continue as this is a duplicate version
+            // of the same slot.
+            if (pubkey_latest_vote) |latest_vote| {
+                if (new_vote_slot < latest_vote.slot or
+                    (new_vote_slot == latest_vote.slot and
+                        new_vote_hash.order(&latest_vote.hash) != .lt))
+                {
+                    continue;
+                }
```
InKryption: I sometimes find that for very simple outcomes, it makes sense to break the rule of DRY:
```suggestion
                if (new_vote_slot < latest_vote.slot) {
                    continue;
                }
                if (new_vote_slot == latest_vote.slot and
                    new_vote_hash.order(&latest_vote.hash) != .lt)
                {
                    continue;
                }
```

src/consensus/fork_choice.zig
```diff
@@ -857,6 +935,160 @@ pub const ForkChoice = struct {
...
+
+                const epoch = epoch_schedule.getEpoch(old_latest_vote_slot);
+                const stake_update = if (epoch_stakes.get(epoch)) |stakes| blk: {
+                    const stake_and_vote_account =
+                        stakes.current.stakes.vote_accounts.vote_accounts.get(pubkey) orelse
+                        break :blk 0;
+                    break :blk stake_and_vote_account.stake;
+                } else blk: {
+                    break :blk 0;
+                };
```
InKryption:
```suggestion
                const stake_update = stake_update: {
                    const stakes = epoch_stakes.get(epoch) orelse
                        break :stake_update 0;
                    const stake_and_vote_account =
                        stakes.current.stakes.vote_accounts.vote_accounts.get(pubkey) orelse
                        break :stake_update 0;
                    break :stake_update stake_and_vote_account.stake;
                };
```

src/consensus/fork_choice.zig
```diff
@@ -857,6 +935,160 @@ pub const ForkChoice = struct {
...
+            // In all cases, we need to remove this pubkey stake from the previous fork
+            if (try self.latest_votes.fetchPut(pubkey, new_vote_slot_hash)) |old_latest_vote| {
+                const old_latest_vote_slot = old_latest_vote.value.slot;
+                const old_latest_vote_hash = old_latest_vote.value.hash;
+                std.debug.assert(if (new_vote_slot == old_latest_vote_slot)
+                    // If the slots are equal, then the new
+                    // vote must be for a smaller hash
+                    new_vote_hash.order(&old_latest_vote_hash) == .lt
+                else
+                    new_vote_slot > old_latest_vote_slot);
```
InKryption:
```suggestion
                if (new_vote_slot == old_latest_vote_slot)
                    // If the slots are equal, then the new
                    // vote must be for a smaller hash
                    std.debug.assert(new_vote_hash.order(&old_latest_vote_hash) == .lt)
                else
                    std.debug.assert(new_vote_slot > old_latest_vote_slot);
```

src/consensus/fork_choice.zig
```diff
@@ -857,6 +935,160 @@ pub const ForkChoice = struct {
...
+                // Add this pubkey stake to new fork
+                const epoch = epoch_schedule.getEpoch(new_vote_slot_hash.slot);
+                const stake_update = if (epoch_stakes.get(epoch)) |stakes| blk: {
+                    const stake_and_vote_account =
+                        stakes.current.stakes.vote_accounts.vote_accounts.get(pubkey) orelse
+                        break :blk 0;
+                    break :blk stake_and_vote_account.stake;
+                } else blk: {
+                    break :blk 0;
+                };
```
InKryption:
```suggestion
                const stake_update = stake_update: {
                    const stakes = epoch_stakes.get(epoch) orelse
                        break :stake_update 0;
                    const stake_and_vote_account =
                        stakes.current.stakes.vote_accounts.vote_accounts.get(pubkey) orelse
                        break :stake_update 0;
                    break :stake_update stake_and_vote_account.stake;
                };
```

src/consensus/fork_choice.zig
```diff
@@ -1194,6 +1426,133 @@ pub const ForkChoice = struct {
...
+            root,
+        );
+        defer allocator.free(news);
+
+        for (news) |vote_tuple| {
+            try new_votes.append(allocator, .{
+                .pubkey = vote_tuple[0],
+                .slot_hash = vote_tuple[1],
+            });
+        }
```
InKryption:
```suggestion
        try new_votes.ensureUnusedCapacity(allocator, news.len);
        for (news) |vote_tuple| {
            const pubkey, const slot_hash = vote_tuple;
            new_votes.appendAssumeCapacity(.{
                .pubkey = pubkey,
                .slot_hash = slot_hash,
            });
        }
```

src/consensus/fork_choice.zig
```diff
@@ -1194,6 +1426,133 @@ pub const ForkChoice = struct {
...
+    ) !ForkChoice {
+        if (!builtin.is_test) {
+            @compileError("splitOff should only be used in test");
+        }
+        std.debug.assert(!self.tree_root.equals(slot_hash_key));
+
+        const node_to_split_at = self.fork_infos.getPtr(slot_hash_key) orelse
+            return error.SlotHashKeyNotFound;
+        var split_tree_root = node_to_split_at.*;
+        const parent = node_to_split_at.parent orelse return error.SplitNodeIsRoot;
```
InKryption: Why `getPtr` if it's immediately copied? Why not just `get` and assign directly to `split_tree_root`.

src/consensus/fork_choice.zig
```diff
@@ -2219,37 +2904,1856 @@ test "HeaviestSubtreeForkChoice.heaviestSlotOnSameVotedFork_missing_candidate" {
...
+        Pubkey.initRandom(random),
+    };
+
+    const duplicate_fork = try setupDuplicateForks();
+    const stake: u64 = 10;
+
+    defer test_allocator.destroy(duplicate_fork.fork_choice);
+    defer test_allocator.free(duplicate_fork.duplicate_leaves_descended_from_4);
+    defer test_allocator.free(duplicate_fork.duplicate_leaves_descended_from_5);
+    defer test_allocator.free(duplicate_fork.duplicate_leaves_descended_from_6);
```
InKryption: I think it would make sense to add a `deinit` method to the struct returned by `setupDuplicateForks`.

InKryption (changes requested):
Overall good work, but I have some questions and suggestions.

src/consensus/fork_choice.zig
dadepo: [4f93c08](https://github.com/Syndica/sig/pull/811/commits/4f93c081f293c5c95b5dc7006f32cb85c4442b67)

src/consensus/fork_choice.zig
dadepo: [202e274](https://github.com/Syndica/sig/pull/811/commits/202e27420feded1a6592e38762dbf8ccb7ee905d)

src/consensus/fork_choice.zig
dadepo: [77073be](https://github.com/Syndica/sig/pull/811/commits/77073bee5dbec6706db3744528e935890eb51d82)

src/consensus/fork_choice.zig
dadepo: [a37ee6d](https://github.com/Syndica/sig/pull/811/commits/a37ee6dacdd0bf265f0b1b24510cbe58906f2b21)

src/consensus/fork_choice.zig
dadepo: [413164d](https://github.com/Syndica/sig/pull/811/commits/413164d45231ccdd867c340bdae2c3ad4ca6b0c4)

src/consensus/fork_choice.zig
dadepo: [2513720](https://github.com/Syndica/sig/pull/811/commits/251372061c63323403b8810872f10c875eb860d2)

src/consensus/fork_choice.zig
dadepo: [77073be](https://github.com/Syndica/sig/pull/811/commits/77073bee5dbec6706db3744528e935890eb51d82)

src/consensus/fork_choice.zig
dadepo: [088f97b](https://github.com/Syndica/sig/pull/811/commits/088f97b975c4b8af139c1837dd765865fca32d39)

src/consensus/fork_choice.zig
dadepo: [c19180c](https://github.com/Syndica/sig/pull/811/commits/c19180cb815bc16cdf026d0c86f974d95145cb74)

InKryption (dismissed):

src/consensus/fork_choice.zig
```diff
@@ -857,6 +934,163 @@ pub const ForkChoice = struct {
...
+    /// 3) Generate aggregate operations for affected forks
+    ///
+    /// Key invariants:
+    /// - Votes older than the current tree root are ignored (they don't affect fork choice)
+    /// - Each pubkey can only appear once in the input batch
+    /// - Only the latest vote for each validator is considered (by slot, then by smallest hash)
+    /// - Stake is only updated if the validator has stake in the vote's epoch
+    ///
+    /// Analogous to [generate_update_operations](https://github.com/anza-xyz/agave/blob/92b11cd2eef1d3f5434d6af702f7d7a85ffcfca9/core/src/consensus/heaviest_subtree_fork_choice.rs#L969)
+    fn generateUpdateOperations(
```
dnut, line 951: What's the purpose of separately generating and applying operations? Why not apply them directly?

Also the name of this function doesn't tell you anything about what kind of updates its doing. Is this supposed to represent every possible change that could ever happen to ForkChoice? What do you think about `generateUpdatesForNewVotes`?

src/consensus/fork_choice.zig
```diff
@@ -857,6 +934,163 @@ pub const ForkChoice = struct {
...
+        self: *ForkChoice,
+        allocator: std.mem.Allocator,
+        pubkey_votes: []const PubkeyVote,
+        epoch_stakes: *const std.AutoHashMapUnmanaged(Epoch, VersionedEpochStakes),
+        epoch_schedule: *const EpochSchedule,
+    ) !UpdateOperations {
+        var update_operations = UpdateOperations.init(self.allocator);
```
dnut, line 958: Why not use the unmanaged variant?

Also, do you think the type alias is helpful?

src/consensus/fork_choice.zig
```diff
@@ -857,6 +934,163 @@ pub const ForkChoice = struct {
...
+                // If the new vote is less than the root we can ignore it. This is because there
+                // are two cases. Either:
+                // 1) The validator's latest vote was bigger than the new vote, so we can ignore it
+                // 2) The validator's latest vote was less than the new vote, then the validator's latest
+                // vote was also less than root. This means either every node in the current tree has the
+                // validators stake counted toward it (if the latest vote was an ancestor of the current root),
+                // OR every node doesn't have this validator's vote counting toward it (if the latest vote
+                // was not an ancestor of the current root). Thus this validator is essentially a no-op
+                // and won't affect fork choice.
+                continue;
```
dnut: Would you mind keeping comments less than 100 characters? These are only allowed by our linting because the linting script is not very smart right now, but I think we should change that.

Also I'm puzzled about the purpose of this comment. Isn't the conclusion here just that we simply do not care about votes for slots older than our current root. If we've already rooted a slot, nothing about older slots can be revisited. That's the whole idea of the slot being rooted. Does that not sum up everything we care about here?

This comment aims to prove to the reader that this vote's relationship to the validator's "latest" vote is irrelevant. But isn't that obvious once you understand that the slot is too old to be considered for consensus? I feel like this comment itself is also kind of irrelevant, or maybe I'm missing something.

src/consensus/fork_choice.zig
```diff
@@ -857,6 +934,163 @@ pub const ForkChoice = struct {
...
+            // In all cases, we need to remove this pubkey stake from the previous fork
+            if (try self.latest_votes.fetchPut(pubkey, new_vote_slot_hash)) |old_latest_vote| {
+                const old_latest_vote_slot = old_latest_vote.value.slot;
+                const old_latest_vote_hash = old_latest_vote.value.hash;
+                if (new_vote_slot == old_latest_vote_slot)
+                    // If the slots are equal, then the new
+                    // vote must be for a smaller hash
+                    std.debug.assert(new_vote_hash.order(&old_latest_vote_hash) == .lt)
+                else
+                    std.debug.assert(new_vote_slot > old_latest_vote_slot);
```
dnut: This code is redundant. It's executing the same logic twice. I think it can be consolidated if you use `getOrPut` instead of `getPtr` and `fetchPut`.

src/consensus/fork_choice.zig
```diff
@@ -857,6 +934,163 @@ pub const ForkChoice = struct {
...
+
+                const epoch = epoch_schedule.getEpoch(old_latest_vote_slot);
+                const stake_update = stake_update: {
+                    const stakes = epoch_stakes.get(epoch) orelse
+                        break :stake_update 0;
+                    const stake_and_vote_account =
+                        stakes.current.stakes.vote_accounts.vote_accounts.get(pubkey) orelse
+                        break :stake_update 0;
+                    break :stake_update stake_and_vote_account.stake;
+                };
```
dnut, line 1010: I understand this is a matter of personal preference so feel free to reject my suggestion.

I find it a little confusing/chaotic using labelled blocks with a bunch of breaks. I think the `orelse-based` expression is more readable, but it's up to you.

```suggestion
                const stake_update = ((epoch_stakes.get(epoch) orelse 0)
                    .current.stakes.vote_accounts.vote_accounts.get(pubkey) orelse 0)
                    .stake;
```

src/consensus/fork_choice.zig
InKryption: I initially was also put off by this, since it seemed like a repeat of the edge case's execution-as-data silliness, but an important distinction I noticed was the fact that `UpdateOperations` is a sorted map, meaning that the order in which code issues the operations isn't necessarily the same order in which they are intended to be executed.

src/consensus/fork_choice.zig
dadepo: So the `UpdateOperations` (and it being an alias) was added months ago, and used for these operations, before we had the unmanaged variant. 

I won't be able to use them here because doing so will involve update existing methods like `doInsertAggregateOperationsAcrossAncestors`.

I agree that these usage of `SortedMap` (here and other places) should be replaced by the unmanaged variant, but did not see it as the core of the changes being added in this PR which is mainly to support the implementation of the `computeBankStats`

src/consensus/fork_choice.zig
InKryption: That suggestion will not compile, `(a orelse 0).c` doesn't make sense.

src/consensus/fork_choice.zig
dadepo: [1e96f1c](https://github.com/Syndica/sig/pull/811/commits/1e96f1c4ec59da04eaf19d69a9ac112f016ad1ed)

src/consensus/fork_choice.zig
dadepo: [1e96f1c](https://github.com/Syndica/sig/pull/811/commits/1e96f1c4ec59da04eaf19d69a9ac112f016ad1ed)

src/consensus/fork_choice.zig
dnut: It's still redundant. You switched to getOrPut but the logic is still duplicated.

```zig
                if (new_vote_slot < old_latest_vote_slot) {
                    continue;
                }

                if (new_vote_slot == old_latest_vote_slot and
                    new_vote_hash.order(&old_latest_vote_hash) != .lt)
                {
                    continue;
                }

                if (new_vote_slot == old_latest_vote_slot) {
                    // If the slots are equal, then the new
                    // vote must be for a smaller hash
                    std.debug.assert(new_vote_hash.order(&old_latest_vote_hash) == .lt);
                } else {
                    std.debug.assert(new_vote_slot > old_latest_vote_slot);
                }
```

dnut: fd183cfe71012fd092d17d1820f6a358bbffd612

dnut (approved):

src/consensus/fork_choice.zig
dadepo: Oh this was actually the way it was before. But changed it here 202e27420feded1a6592e38762dbf8ccb7ee905d based off the suggestion from Trevor (https://github.com/Syndica/sig/pull/811#discussion_r2219824158), which I agree with.

PR #812, opened 2025-07-14
feat(replay,consensus): Implement compute bank stats

dadepo:


src/consensus/replay_tower.zig
```diff
@@ -4170,24 +4200,104 @@ pub const TestFixture = struct {
...
+            num_keypairs,
+        );
+
+        for (0..num_keypairs) |_| {
+            try node_pubkeys.append(allocator, Pubkey.initRandom(random));
+            try vote_pubkeys.append(allocator, Pubkey.initRandom(random));
+        }
+
+        self.node_pubkeys = node_pubkeys;
+        self.vote_pubkeys = vote_pubkeys;
```
```suggestion
        self.node_pubkeys.clearRetainingCapacity();
        self.vote_pubkeys.clearRetainingCapacity();

        try self.node_pubkeys.ensureTotalCapacityPrecise(allocator, num_keypairs);
        try self.vote_pubkeys.ensureTotalCapacityPrecise(allocator, num_keypairs);
        for (0..num_keypairs) |_| {
            self.node_pubkeys.appendAssumeCapacity(.initRandom(random));
            self.vote_pubkeys.appendAssumeCapacity(.initRandom(random));
        }
```

src/consensus/replay_tower.zig
```diff
@@ -4170,24 +4200,104 @@ pub const TestFixture = struct {
...
+            allocator,
+            input_tree.root.slot,
+        );
+        if (gop_root.found_existing) {
+            gop_root.value_ptr.deinit(allocator);
+        }
+        var root_fp = try ForkProgress.zeroes(allocator);
+        root_fp.fork_stats.computed = true;
+        root_fp.fork_stats.my_latest_landed_vote = null;
+        gop_root.value_ptr.* = root_fp;
```
```suggestion
        try self.progress.map.ensureUnusedCapacity(allocator, 1);
        const prev_root = self.progress.map.fetchPutAssumeCapacity(input_tree.root.slot, fp: {
            var root_fp = try ForkProgress.zeroes(allocator);
            root_fp.fork_stats.computed = true;
            root_fp.fork_stats.my_latest_landed_vote = null;
            break :fp root_fp;
        });
        if (prev_root) |kv| kv.value.deinit(allocator);
```

src/consensus/replay_tower.zig
```diff
@@ -4170,24 +4200,104 @@ pub const TestFixture = struct {
...
-            var fp = try ForkProgress.zeroes(allocator);
-            fp.fork_stats.computed = true;
-            fp.fork_stats.my_latest_landed_vote = null;
-            try self.progress.map.putNoClobber(
+            const gop = try self.progress.map.getOrPut(
                 allocator,
                 tree[0].slot,
-                fp,
             );
+            if (gop.found_existing) {
```
InKryption: Similar suggestion as above to use `fetchPut` here instead (along with an appropriate `ensureUnusedCapacity(1)` to avoid error paths with redundantly instantiated resources).

src/replay/consensus.zig
```diff
@@ -442,8 +458,68 @@ fn resetFork(
...
+    my_vote_pubkey: Pubkey,
+    ancestors: std.AutoHashMapUnmanaged(u64, SortedSet(u64)),
+    slot_tracker: *SlotTracker,
+    epoch_stakes: *const std.AutoHashMapUnmanaged(Epoch, VersionedEpochStakes),
+    epoch_schedule: *const EpochSchedule,
+    progress: *ProgressMap,
+    fork_choice: *ForkChoice,
+    latest_validator_votes: *LatestValidatorVotesForFrozenBanks,
+) ![]Slot {
+    var new_stats = std.ArrayListUnmanaged(Slot).empty;
```
InKryption, line 483: Missing `errdefer new_stats.deinit(allocator);`

src/consensus/replay_tower.zig
```diff
@@ -4170,24 +4200,104 @@ pub const TestFixture = struct {
...
+
+        self.node_pubkeys = node_pubkeys;
+        self.vote_pubkeys = vote_pubkeys;
+    }
+
     pub fn fill_fork(
         self: *TestFixture,
         allocator: std.mem.Allocator,
         input_tree: Tree,
+        is_frozen: bool,
```
InKryption: I'm not a huge fan of unnamed boolean parameters, it's not clear at the call-sites in the tests what `fixture.fill_fork(allocator, tree, true)` actually means. Would you be open to making it an enum? ie `state: enum { frozen, unfrozen }`, making the call-sites look like `fixture.fill_fork(allocator, tree, .frozen)`.

src/replay/consensus.zig
```diff
@@ -1234,3 +1316,200 @@ test "checkAndHandleNewRoot - success" {
...
+        random,
+    );
+    defer versioned_stakes.deinit(testing.allocator);
+
+    var epoch_stakes = std.AutoHashMapUnmanaged(Epoch, VersionedEpochStakes).empty;
+    defer epoch_stakes.deinit(testing.allocator);
+    try epoch_stakes.put(testing.allocator, 0, versioned_stakes);
+    try epoch_stakes.put(testing.allocator, 1, versioned_stakes);
+
+    var epoch_schedule = EpochSchedule.DEFAULT;
```
InKryption: does this need to be var? if not it should be const.

src/replay/consensus.zig
```diff
@@ -1234,3 +1316,200 @@ test "checkAndHandleNewRoot - success" {
...
+        &fixture.slot_tracker,
+        &epoch_stakes,
+        &epoch_schedule,
+        &fixture.progress,
+        &fixture.fork_choice,
+        &fixture.latest_validator_votes_for_frozen_banks,
+    );
+
+    // Sort frozen slots by slot number
+    var slot_list = try testing.allocator.alloc(u64, frozen_slots.count());
```
InKryption: This should be `const`.

src/replay/consensus.zig
```diff
@@ -1234,3 +1316,200 @@ test "checkAndHandleNewRoot - success" {
...
+
+    // Check that the heaviest slot is always the leaf (slot 3)
+    for (slot_list) |slot| {
+        const slot_info = fixture.slot_tracker.get(slot) orelse
+            return error.MissingSlot;
+        const best = fixture.fork_choice.heaviestSlot(
+            .{ .slot = slot, .hash = slot_info.state.hash.readCopy().? },
+        ) orelse
+            return error.MissingSlot;
+        try testing.expectEqual(@as(u64, 3), best.slot);
```
InKryption: I don't think the `@as` here is necessary?

src/replay/consensus.zig
```diff
@@ -1234,3 +1316,200 @@ test "checkAndHandleNewRoot - success" {
...
+        random,
+    );
+    defer versioned_stakes.deinit(testing.allocator);
+
+    var epoch_stakes = std.AutoHashMapUnmanaged(Epoch, VersionedEpochStakes).empty;
+    defer epoch_stakes.deinit(testing.allocator);
+    try epoch_stakes.put(testing.allocator, 0, versioned_stakes);
+    try epoch_stakes.put(testing.allocator, 1, versioned_stakes);
+
+    var epoch_schedule = EpochSchedule.DEFAULT;
```
InKryption: Same comment about constness on other instance of epoch schedule.

src/replay/consensus.zig
```diff
@@ -1234,3 +1316,200 @@ test "checkAndHandleNewRoot - success" {
...
+        @compileError("convertAncestorsMap should only be called in test mode");
+    }
+    var dst = std.AutoHashMapUnmanaged(u64, SortedSet(u64)).empty;
+    try dst.ensureTotalCapacity(allocator, @intCast(src.count()));
+    var it = src.iterator();
+    while (it.next()) |entry| {
+        try dst.put(allocator, entry.key_ptr.*, entry.value_ptr.*);
+    }
+    return dst;
+}
```
InKryption: What does this function actually do? It appears that the input and output are exactly identical.

InKryption (changes requested):

src/consensus/replay_tower.zig
dadepo: [fd1ae3f](https://github.com/Syndica/sig/pull/812/commits/fd1ae3f9ce599de47dbc003a24b6f0ae68082e73)

src/consensus/replay_tower.zig
dadepo: [e7ed897](https://github.com/Syndica/sig/pull/812/commits/e7ed897daf8feeae1c484e9abfbfe5ec024ee286)

src/consensus/replay_tower.zig
dadepo: [9489ab2](https://github.com/Syndica/sig/pull/812/commits/9489ab2eef84ff46f7efdc1d0caf175b9d7a39da)

src/replay/consensus.zig
dadepo: [b789fe2](https://github.com/Syndica/sig/pull/812/commits/b789fe2297419c1c0adbcf8f129643977b2b9201)

src/replay/consensus.zig
dadepo: [b789fe2](https://github.com/Syndica/sig/pull/812/commits/b789fe2297419c1c0adbcf8f129643977b2b9201)

src/replay/consensus.zig
dadepo: [b789fe2](https://github.com/Syndica/sig/pull/812/commits/b789fe2297419c1c0adbcf8f129643977b2b9201)

src/replay/consensus.zig
dadepo: [b789fe2](https://github.com/Syndica/sig/pull/812/commits/b789fe2297419c1c0adbcf8f129643977b2b9201)

src/replay/consensus.zig
dadepo: [b789fe2](https://github.com/Syndica/sig/pull/812/commits/b789fe2297419c1c0adbcf8f129643977b2b9201)

src/consensus/replay_tower.zig
dadepo: [4f96c1d](https://github.com/Syndica/sig/pull/812/commits/4f96c1de45367d349ef89c44964b2af13378729d) 
[4bf4296](https://github.com/Syndica/sig/pull/812/commits/4bf4296fbc7722dbb567e8cdab588ede86cda5b4)

src/replay/consensus.zig
dadepo: This is to help convert `std.AutoArrayHashMapUnmanaged` to `std.AutoHashMapUnmanaged` which came up in the tests that required ancestors which in the `fixture` is defined as `std.AutoArrayHashMapUnmanaged` but the new methods being tested expects `std.AutoHashMapUnmanaged`

InKryption: I see, didn't notice that small difference. Any reason the new methods can't just also use array hash maps? If not, I think changing them to that seems preferable to this.

Opinionated side note:
In general, I think non-array hashmaps are kind of a niche data structure, given the sheer amount of code that ends up iterating over hashmaps in practice. I think the main situations that justify non-array hashmaps are those where lookup performance is critical.

dadepo: > In general, I think non-array hashmaps are kind of a niche data structure, given the sheer amount of code that ends up iterating over hashmaps in practice.

This is one of the reason why I thought of keeping it as `std.AutoHashMapUnmanaged` in the new methods given all it's usage are lookups, and my understanding is that `std.AutoHashMapUnmanaged` is preferable for lookups while `std.AutoArrayHashMapUnmanaged` is preferably for iterations. - hence why I hesitated from switching the type, but convert instead in the tests, because if I had switched, it would mean the switch was driven by testing, but not the actual runtime behaviour.

InKryption: I would hesitate to make the decision to use non-array hashmap for performance reasons without some empirical data backing up the decision, but I suppose there's not a strong argument against it in this case  either beyond the awkwardness of this facet of testing.

I'm fine to leave it as is if that's how you prefer it, but I would suggest some refactors:
```suggestion
const builtin = @import("builtin");
fn convertAncestorsMap(
    allocator: std.mem.Allocator,
    src: *const std.AutoArrayHashMapUnmanaged(Slot, SortedSet(Slot)),
) !std.AutoHashMapUnmanaged(Slot, SortedSet(Slot)) {
    if (!builtin.is_test) {
        @compileError("convertAncestorsMap should only be called in test mode");
    }
    var dst = std.AutoHashMapUnmanaged(u64, SortedSet(u64)).empty;
    errdefer dst.deinit(allocator);

    try dst.ensureTotalCapacity(allocator, src.count());
    for (src.keys(), src.values()) |key, value| {
        dst.putAssumeCapacity(key, value);
    }
    return dst;
}
```
Addendum: what's the ownership for those values look like? Is it fine that they're not being cloned?

dadepo: Ended up dropping the `convertAncestorsMap` and switched `ancestors` to `AutoArrayHashMapUnmanaged`. I agree that the conversion function looks a bit clunky and yeah better to back the decision to go with non-array hashmap with some empirical data.
 
[0234aee](https://github.com/Syndica/sig/pull/812/commits/0234aee648203e65dec8d30ec9580a0737b18b23)

InKryption (approved):

src/replay/consensus.zig
```diff
@@ -443,8 +460,68 @@ fn resetFork(
...
+            fork_stats.total_stake = computed_bank_state.total_stake;
+            fork_stats.voted_stakes = computed_bank_state.voted_stakes;
+            fork_stats.lockout_intervals =
+                try computed_bank_state.lockout_intervals.clone(allocator);
+            fork_stats.block_height = blk: {
+                const slot_info = slot_tracker.get(slot) orelse return error.MissingSlots;
+                break :blk slot_info.constants.block_height;
+            };
+            fork_stats.my_latest_landed_vote = computed_bank_state.my_latest_landed_vote;
+            fork_stats.my_latest_landed_vote = computed_bank_state.my_latest_landed_vote;
```
dnut: duplicated code

src/replay/consensus.zig
```diff
@@ -443,8 +460,68 @@ fn resetFork(
...
+                epoch_stakes,
+                epoch_schedule,
+                latest_validator_votes,
+            );
+            const fork_stats = progress.getForkStats(slot) orelse return error.MissingForkStats;
+            fork_stats.fork_stake = computed_bank_state.fork_stake;
+            fork_stats.total_stake = computed_bank_state.total_stake;
+            fork_stats.voted_stakes = computed_bank_state.voted_stakes;
+            fork_stats.lockout_intervals =
+                try computed_bank_state.lockout_intervals.clone(allocator);
```
dnut: You don't need to clone this since computed_bank_state is being discarded in this loop. You should be able to just transfer ownership here, similar to voted_stakes.

Actually, it's strange that this does not show up as a leak in your tests, since computed_bank_state is never deinited. I just ran all of the tests for computeBankStats, and it looks like lockout_intervals is empty in every test. What do you think about modifying one of the tests to make it non-empty? Maybe that would improve test coverage.

src/replay/consensus.zig
dadepo: [94dadca](https://github.com/Syndica/sig/pull/812/commits/94dadca69de5323ae749318d67d647964c31ce7b)

src/replay/consensus.zig
dadepo: Removed the clone: [8e94ea2](https://github.com/Syndica/sig/pull/812/commits/8e94ea2cd79e77b58a5bb4dcb6f74bff94d8de67)

I'll look into making `lockout_intervals` non-empty but I am not sure if that will have much effect on the coverage, but I'll see.

dadepo: This ensure at least one lockout [918d2ed](https://github.com/Syndica/sig/pull/812/commits/918d2edb3774ef1a809f1713cb537c6a46ab28a3)

Revealed a leak in the tests that was handled in the deinit in the test fixture.

src/core/vote_accounts.zig
```diff
@@ -394,6 +396,16 @@ pub fn createVoteAccount(
     );
     defer vote_state.deinit();
 
+    try vote_state.votes.append(
+        LandedVote{
```
dadepo: Currently re-reviewing this, if this is the best spot or if it can be moved to the test fixture.

dnut (approved):

PR #814, opened 2025-07-15
feat(replay,accountsdb,core): deps for svm & commit

dnut:
### changes
- add dependencies that are needed in replay to invoke the svm and commit transactions. this is added to replay state and the slot constants/state
- implement AccountStore and AccountReader interfaces for AccountsDB
- move features.zig to core.
- add getActiveFeatures to replay
- implement LtHash mixin methods
- genericize account hashing to work with either Hash or LtHash

### context
I'm splitting some changes out from `dnut/replay/commit` because dnut/replay/commit can't merge until 4 other branches merge that are still in progress. Also smaller prs may be easier to review. Other people are going to be interacting with or using the code I have in my branch, so I'd like to prioritize merging those pieces


src/accountsdb/account_store.zig
```diff
@@ -0,0 +1,211 @@
...
+    ) !void {
+        const map, var lock = self.map.writeWithLock();
+        defer lock.unlock();
+
+        const gop = try map.getOrPut(self.allocator, address);
+        if (!gop.found_existing) {
+            gop.value_ptr.* = .empty;
+        }
+
+        const data = try self.allocator.dupe(u8, account.data);
```
yewman, line 182: What do you think about `put` taking ownership of the account, so that duplication becomes the responsibility of the caller and only occurs when strictly necessary. I get right now this function was implemented to mirror the behaviour of accounts db `putAccouount` but perhaps both could be updated

src/accountsdb/account_store.zig
```diff
@@ -0,0 +1,211 @@
...
+        return .{
+            .lamports = account.lamports,
+            .data = .{ .owned_allocation = data },
+            .owner = account.owner,
+            .executable = account.executable,
+            .rent_epoch = account.rent_epoch,
+        };
+    }
+
+    pub fn put(
```
yewman, line 168: Should this function should handle overwriting accounts inserted with an existing entry for the given slot?

src/accountsdb/account_store.zig
```diff
@@ -0,0 +1,211 @@
...
+
+    pub fn accountStore(self: *ThreadSafeAccountMap) AccountStore {
+        return .{ .thread_safe_map = self };
+    }
+
+    pub fn accountReader(self: *ThreadSafeAccountMap) AccountReader {
+        return .{ .thread_safe_map = self };
+    }
+
+    pub fn get(
```
yewman, line 128: I think we should be checking that account lamports are non-zero an returning null accordingly. Same for the accounts db get method
https://github.com/firedancer-io/agave/blob/f5bed6630a541549f8c91b4c6fe5adf1f07cf594/accounts-db/src/accounts.rs#L325

src/replay/resolve_lookup.zig
```diff
@@ -486,3 +462,25 @@ test resolveBatchGeneric {
...
+) !void {
+    var buf: [AddressLookupTable.MAX_SERIALIZED_SIZE]u8 = undefined;
+
+    try AddressLookupTable.overwriteMetaData(&buf, lookup_table.meta);
+
+    const src = std.mem.sliceAsBytes(lookup_table.addresses);
+    const dst = buf[sig.runtime.program.address_lookup_table.state.LOOKUP_TABLE_META_SIZE..];
+    @memcpy(dst[0..src.len], src);
+
+    try map.put(0, address, .{
```
yewman, line 479: From my understanding accounts with zero lamports should not be loadable, perhaps use 1 here

yewman (changes requested):

src/accountsdb/account_store.zig
```diff
@@ -0,0 +1,211 @@
...
+        };
+    }
+
+    pub fn getLatest(self: AccountStore, address: Pubkey) !?Account {
+        return switch (self) {
+            .accounts_db => |db| try db.getAccount(&address),
+            .thread_safe_map => |map| try map.getLatest(address),
+            .noop => null,
+        };
+    }
```
InKryption: Any reason these are duplicated in both this and the reader? I think we could delete these; reading from the store could be accomplished with `store.accountReader().get()`.

src/accountsdb/account_store.zig
```diff
@@ -0,0 +1,211 @@
...
+/// Interface for both reading and writing accounts.
+///
+/// Do not use this unless you need to *write* into the database generically.
+/// Otherwise use AccountReader if you only need to read accounts.
+pub const AccountStore = union(enum) {
+    accounts_db: *AccountsDB,
+    thread_safe_map: *ThreadSafeAccountMap,
+    noop,
+
+    pub fn accountReader(self: AccountStore) AccountReader {
```
InKryption: `accountReader` seems a bit redundant in context, how about just `reader()`.

InKryption (changes requested):
A couple of extremely minor comments, everything else looks great; I like the use of the tagged union for the account reader & store, and how this paves the way for more granular capabilities-based parameterization of accountsdb.

src/accountsdb/account_store.zig
dnut: Funny how I didn't put any thought into that when it's exactly the problem we're seeing in accountsdb.

c97ef2281eb937421ed0d79b46c283c234b08e98

src/accountsdb/account_store.zig
dnut: e0004f4c1b1504d4181f46ae2695e43f315ef14c

src/accountsdb/account_store.zig
dnut: e0004f4c1b1504d4181f46ae2695e43f315ef14c

src/accountsdb/account_store.zig
dnut: That's a substantial change to accountsdb that I wasn't planning to tackle with this PR. I can set up the interface here to have that behavior. Handling this properly in accountsdb is going to require changes to a lot of functions.

592932f7f85a62b422cbb455e9f6060b14c504de

src/replay/resolve_lookup.zig
dnut: c702df68ccb84e305a03553312d62812ae170507

src/accountsdb/account_store.zig
dnut: Let's say hypothetically the database actually needs to make a copy regardless. For example if we decided to use RocksDB for storage, then RocksDB is going to copy the data regardless. It would be a waste to clone it before, since it's going to stop using the pointer you pass it immediately after it copies the data out of it.

Granted with how accountsdb works now, that's not a problem. We can just remove the call to dupe inside putAccount. So your approach would be fine with our current code.

I'm not sure if your suggested approach helps in any specific scenario, but I'm guessing you have some thoughts on that.

You would also need to make sure the same exact allocator is used for initialization and free. That can be handled with an `allocator()` method on the interface, but it is more to juggle for users of the interface.

I think the optimal solution would be to accept a tagged union:

```zig
const AccountToPut = struct {
    data: union(enum) {
        /// Owned by the caller to `AccountStore.put`, and the database 
        /// must release its reference before returning from `put`
        borrowed: []const u8,
        /// Was allocated by `AccountStore.allocator` and the ownership
        /// is being transferred to the database upon calling `put`
        owned: []u8,
    },
    owner: Pubkey,
    lamports: u64,
    ...
};
```

Then the caller and the callee both have the flexibility to do the most efficient operation possible depending on their implementation details.

yewman: Ahh that is a good point, I overlooked the fact that we are duping in `putAccount` with the accounts db allocator. I think the `AccountToPut` tagged union could be a good move, however, I don't think it is worth holding up this PR for.

yewman (approved):
Looks good to me!

InKryption (approved):

yewman (approved):

PR #815, opened 2025-07-15
feat(consensus:vote_listener): Integrate ledger, solidify API boundary

InKryption:
Integrate ledger, flesh out "stubs"
* Remove `TODO_CONFIRMATION_VERIFIER`, refactor related code.
* Actually add ledger code usages.
* Implement `OptimisticConfirmationVerifier` as a stub.
* Start using `BankForksStub` as an API boundary.
  * Remove concept of individual banks, all querying is done through the API boundary.
* Remove a bunch of RwMux layers, can be added back later as needed.
* Remove dead comment.

Misc tweaks and additions:
* Make ledger `put` no longer require a mutable pointer to self.
* Implement `insertOptimisticSlot`.
* Add some ordered set/map methods.
* Use `sig.core.hash.SlotAndHash` in place of ad-hoc tuples.
* Remove left-over comment.
* Remove unused todo error.


src/consensus/vote_listener.zig
```diff
@@ -18,13 +18,15 @@ const EpochTracker = sig.replay.trackers.EpochTracker;
 pub const BankForksStub = struct {
```
dadepo: Motion to rename this away from BankForksStub, because it is no longer a stub but a struct that houses dependencies needed by vote processing.

Maybe `ProcessorDependency`? Suggestion for better naming is welcomed

src/replay/edge_cases.zig
```diff
@@ -623,6 +623,7 @@ fn processGossipVerifiedVoteHashes(
 /// Analogous to [process_popular_pruned_forks](https://github.com/anza-xyz/agave/blob/0315eb6adc87229654159448344972cbe484d0c7/core/src/replay_stage.rs#L1828)
 fn processPopularPrunedForks(
     logger: replay.service.Logger,
+    // TODO: ideally this should be renamed to pruned_but_popular_forks_receiver
```
dnut: addressing this now seems as easy as adding the todo. is there a reason not to do it now?

src/consensus/vote_listener.zig
dnut: I see this as the sole implementation of an implicit interface that describes the vote listener's dependency on a data source. So my instinct for naming is to imagine I'm naming an interface being used for dependency inversion.

Concretely, that results in me asking myself the question, "what role does this play in the context of the vote listener?" Its role is to dynamically provide specific pieces of data about arbitrary slots. What's the best name for this role? I'm not sure, but my first thought is `SlotDataProvider`.

I lean towards excluding the word "dependency" unless the struct's role is to contain direct dependencies, whereas in this case I think the struct itself is the abstraction that acts as the direct dependency. It's a subtle and perhaps flawed distinction, so I don't feel that strongly about it.

src/replay/edge_cases.zig
dadepo: >  is there a reason not to do it now?

Nope. Left it as a TODO as the observation about the naming felt like it was outside the scope of the PR. But the changes is indeed contained enough.

35628c0

src/consensus/vote_listener.zig
dadepo: >  Its role is to dynamically provide specific pieces of data about arbitrary slots. What's the best name for this role? I'm not sure, but my first thought is SlotDataProvider.

It does house the `epoch_tracker` tracker too, that is used to provide data relating to total stakes, delegated stakes  and the authorized voter's pub key, so I think `SlotDataProvider` might be too narrow.

Maybe `EpochDataProvider`? Not 100% sure if that also captures it well. 

I'll let it marinate for a bit more and maybe a better suggestion comes to mind.

> I lean towards excluding the word "dependency" unless the struct's role is to contain direct dependencies

Agreed.

dadepo: Felt `EpochDataProvider` might also be misleading, but initially justified that with the fact that it can be argued slots data is a subset of epoch data.

Ended up going with `VoteListenerData` [a1d5851](https://github.com/Syndica/sig/pull/815/commits/a1d5851644d7ef23b9e28d540364ae37b9faac53)

I think it is an improvement to what it is now (BankForksStub) and it can be improved later if something better comes up.

dnut: Yes I realized that it houses epoch_tracker, but it's only there for the purpose of accessing epoch-scoped information that's relevant for a particular slot. Note that all the functions accept slot as a parameter, not epoch (except getEpochTotalStake which is unused). This is where my question is salient "what role does this play in the context of the vote listener?" In my opinion, the name should describe what role it serves for the vote listener. Its role is serving data about a slot. The fact that it uses epoch tracker under the hood is not relevant to the abstraction.

I prefer `SlotDataProvider` because I think it's clearer within the current context about what it does. "Vote listener" is implied by the context, so the name VoteListenerData doesn't tell us much about what this struct is for. Anyway it's an improvement over the existing name, "BankForksStub" so I'm happy to merge with this name rather than get mired in unnecessary debate.

dnut (approved):

src/consensus/vote_listener.zig
dnut: feel free to resolve this

dadepo: > so the name VoteListenerData doesn't tell us much about what this struct is for

True I was hoping to alleviate that with the comment.

> Note that all the functions accept slot as a parameter, not epoch (except getEpochTotalStake which is unused)

I think that is a good reason to stick with `SlotDataProvider`  [34d7b81](https://github.com/Syndica/sig/pull/815/commits/34d7b81d6844d60bd8ef49066171749c0cc6ecf2)

InKryption (commented):
LGTM, great work @dadepo

PR #816, opened 2025-07-16
feat(replay): hash and freeze slot. lthash + old delta hash

dnut:
This adds `freezeSlot` and all the necessary supporting methods. Currently all that freezing the slot does is calculate the slot hash and update the slot state. Updating sysvars and hard forks will come in future prs.

### DIFF
--- a/src/utils/merkle_tree.zig
+++ b/src/utils/merkle_tree.zig
@@ -37,10 +37,11 @@ pub const NestedHashTree = NestedList(Hash);
 pub fn computeMerkleRoot(self: *const NestedHashTree, fanout: usize) !Hash {
     var length = self.len();
 
-    if (length == 0) {
+    if (length == 0) return comptime empty: {
+        @setEvalBranchQuota(3187);
         var hasher = Sha256.init(.{});
-        return .{ .data = hasher.finalResult() };
-    }
+        break :empty .{ .data = hasher.finalResult() };
+    };
 
     while (true) {
         const chunks = try std.math.divCeil(usize, length, fanout);


src/replay/freeze.zig
```diff
@@ -0,0 +1,198 @@
...
+        }) };
+    } else {
+        const accounts_hash = try deltaMerkleHash(params.account_reader, allocator, params.slot);
+        return .{ null, Hash.generateSha256(.{
+            &params.parent_slot_hash.data,
+            &accounts_hash.data,
+            &signature_count_bytes,
+            &params.blockhash.data,
+        }) };
+    }
```
yewman: I presume this function is analogous to `Bank::hash_internal_state`, in which case there are a few potentially important differences. In Agave the incorporation of the accounts hash is dependent on REMOVE_ACCOUNTS_DELTA_HASH feature, not the ACCOUNTS_LT_HASH feature. Subsequently, if ACCOUNTS_LT_HASH is active the accounts_lt_hash is incorporated, otherwise an epoch_accounts hash is incorporated if `Bank::wait_get_epoch_accounts_hash` returns Some. In psuedo code
```
hash = if (REMOVE_ACCOUNTS_DELTA_HASH) { hashv([parent,  sig_count, blockhash]) } else { hashv([parent, accounts_hash, sig_count, blockhash])}
hash = if (ACCOUNTS_LT_HASH) { hashv([hash, accounts_lt_hash]) } else { if (epoch_accounts_hash.is_some()) { hashv([hash, epoch_accounts_hash]) } else { hash } }
```
It would seem that ACCOUNTS_LT_HASH would imply REMOVE_ACCOUNTS_DELTA_HASH but I am not certain, a debug assert could be used to check this. There is still the question of the incorporation of the epoch accounts hash, are you familiar with this? 

My understanding of slot hashing is limited so perhaps the above is a none issue, if so I am curious to learn what I am missing!

dnut: c80d9068a96aa13eae84de2cb6da08c6b575cd7b

yewman: Looks good. Do you know what the go is with the epoch_accounts_hash and why we can ignore that?

dnut: It is complicated to calculate, and it is not used on mainnet/testnet/devnet any more since lthash is enabled. 

For conformance tests I don't think it would be needed since it's only used once per epoch, unless there is a conformance test that explicitly targets that code, which actually wouldn't surprise me. Do transaction harness tests validate the bank hash?

yewman: Okay that makes sense. I don't think it will be needed, but can't yet guarantee it. It makes sense to leave out for now, I'll let you know if I come across a need for it during fuzzing

yewman (approved):

src/replay/freeze.zig
```diff
@@ -0,0 +1,563 @@
...
+        collector_id,
+        payout,
+    ) catch |e|
+        switch (e) {
+            error.InvalidAccountOwner,
+            error.LamportOverflow,
+            error.InvalidRentPayingAccount,
+            => burn = total_fees,
+            else => return e,
+        };
```
InKryption: Interesting choice of indentation, any particular reason for this, or just aesthetics?

InKryption (commented):
Everything lgtm, just a curious question

src/replay/freeze.zig
dnut: it was a gradual evolution of changes. first i kept the arguments all on one line and moved the switch to the next line. then i added more arguments and needed to break the function call up into multiple lines, and overlooked that the catch and switch were still separated

62c929fc79624c14826336c2c9012386924f2342

InKryption (approved):

src/utils/merkle_tree.zig
```diff
@@ -34,9 +34,13 @@ pub fn NestedList(comptime T: type) type {
 
 pub const NestedHashTree = NestedList(Hash);
 
-pub fn computeMerkleRoot(self: *const NestedHashTree, fanout: usize) !*Hash {
+pub fn computeMerkleRoot(self: *const NestedHashTree, fanout: usize) !Hash {
     var length = self.len();
-    if (length == 0) return error.EmptyHashList;
+
+    if (length == 0) {
+        var hasher = Sha256.init(.{});
```
Rexicon226, line 42: There is no need to pay this runtime cost. The sha256 of a empty string is 
`e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`

You could just make that a constant on `Hash` or you could just compute it right above this function,

```zig
const empty_merkle_hash: Hash = b: {
    @setEvalBranchQuota(3187);
    var hasher: Sha256 = .init(.{});
    const result = hasher.finalResult();
    break :b .{ data = result };
};
```

dnut: how about this 7d668907ccfab0611ad3002193e22077654c8221

InKryption (approved):

src/utils/merkle_tree.zig
Rexicon226: Looks good, thanks.

ultd (approved):

PR #817, opened 2025-07-16
fix(accountsdb): overwrite accounts in-place instead of duplicating

dnut:


src/accountsdb/index.zig
```diff
@@ -294,35 +294,13 @@ pub const AccountIndex = struct {
...
-                account_ref.prev_ptr = prev;
-
+                // found a duplicate => dont do the insertion
                 return false;
             }
             const next_ptr = curr_ref.next_ptr orelse {
                 // end of the list => insert it here
-                account_ref.prev_ptr = curr_ref;
                 curr_ref.next_ptr = account_ref;
                 curr_ref.next_index = index;
```
Sobeston: We still need to update prev_ptr here

Sobeston (changes requested):
Looking good + nice comments, think one line was accidentally reverted

src/accountsdb/index.zig
dnut: 10f9062b4f83726084b94d75a1f939cbb6c03d8c

PR #819, opened 2025-07-22
fix(svm,accountsdb): account loader fork-awareness and 0-lamport handling

dnut:
This PR makes the svm's account loader fork-aware and fixes the handling of 0-lamport accounts by switching the account loader to use one of the new interfaces for accountsdb.

## Motivation

Replay will not work correctly unless the account loader uses fork-aware (ancestors-based) accountsdb methods. If it's not fork-aware it will provide incorrect account state into the SVM, resulting in incorrect transaction outputs.

This also fixes a bug that's causing problems in the transaction harness conformance tests. In the account loader, 0-lamport accounts are not supposed to be included, but currently they are being returned by accountsdb. The new interfaces handle that scenario properly. AccountsDB should eventually be changed to do this, but it's tedious to make it work in the most performant way. Fixing it through the interface is sufficient for now.

Why switch to the new interface:
- New interfaces already include both fork-awareness and the 0-lamport bugfix.
- Easier compatibility with replay, which uses these interfaces.
- Any other solution would be just as much work since it would require reworking the existing comptime interface to add the functionality supported by the new interfaces.
- I believe the new interfaces are also a better long-term approach. Compared to the existing comptime interface, they are more readable and more widely applicable to other areas of code, making it easier to manage dependencies and interoperability of components.

## Details

`AccountStore` and `AccountReader` were added in #814 as interfaces for `AccountsDB`. This PR also adds a new one called `SlotAccountReader`.

The git diff in this PR is a bit misleading about what code is actually new. The `AccountStore` struct appears to be newly added in this PR because I've moved it up from a lower location in the file. It's otherwise identical. `SlotAccountReader` is the entirely new addition, even though in the diff it appears to be an evolution of AccountStore.

The new `SlotAccountReader` interface is split off from `AccountReader` to serve the typical use case where you're only concerned about the state for one particular slot. Typically ancestors are passed around only for this use case. Instead, you can simply depend on an account reader that ensures it will only return accounts that should be visible during that slot.


src/accountsdb/account_store.zig
```diff
@@ -18,6 +18,32 @@ const AccountSharedData = sig.runtime.AccountSharedData;
 
 const AccountsDB = accounts_db.AccountsDB;
 
+/// Interface for both reading and writing accounts.
+///
+/// Do not use this unless you need to *write* into the database generically.
+/// Otherwise use AccountReader if you only need to read accounts.
+pub const AccountStore = union(enum) {
```
InKryption, line 25: I think it would make sense to make these unions all share a tag, both to illustrate the fact that they share a similar purpose in acting as "views" into an accountsdb-looking-shape, and also for the bonus that it keeps them in sync, both in terms of the names and the order of declaration.
```zig
pub const ViewKind = enum { accounts_db, thread_safe_map, noop };
pub const AccountStore = union(ViewKind) {...};
pub const AccountReader = union(ViewKind) {...};
pub const SlotAccountReader = union(ViewKind) {...};
```

src/accountsdb/account_store.zig
```diff
@@ -64,28 +83,37 @@ pub const AccountReader = union(enum) {
...
-/// Do not use this unless you need to *write* into the database generically.
-/// Otherwise use AccountReader if you only need to read accounts.
-pub const AccountStore = union(enum) {
-    accounts_db: *AccountsDB,
-    thread_safe_map: *ThreadSafeAccountMap,
+/// Interface for only reading accounts that should be visible for a particular slot.
+pub const SlotAccountReader = union(enum) {
+    accounts_db: struct { *AccountsDB, *const Ancestors },
+    thread_safe_map: struct { *ThreadSafeAccountMap, *const Ancestors },
+    simple_map: struct { Allocator, *const std.AutoArrayHashMapUnmanaged(Pubkey, Account) },
```
InKryption: Would it make sense to add this `simple_map` variant to the views? (related: shared enum tag comment).

src/runtime/account_loader.zig
```diff
@@ -982,13 +930,13 @@ test "loadAccount allocations" {
         }
     }.f;
 
-    try std.testing.checkAllAllocationFailures(allocator, checkFn, .{});
+    try std.testing.checkAllAllocationFailures(std.testing.allocator, checkFn, .{});
```
InKryption: Small suggestion, the `struct { fn f() void {} }.f` stuff looks pretty awkward, it can be made nicer by instead embracing namespacing:
```zig
const helper = struct {
    fn checkFn(allocator: Allocator) !void {...}
};

try std.testing.checkAllAllocationFailures(std.testing.allocator, helper.checkFn, .{});
```
Also opens the door to more easily adding more local functions without needing to refactor a huge amount or add multiple `struct {...}.f` warts.

InKryption (commented):
Great improvements here, just a few minor comments.

src/accountsdb/account_store.zig
dnut: I suppose it could be made to work with the other interfaces, but it would be strange. 

AccountsDB and the other map are specifically intended to store multiple versions of accounts (different slots have different versions). This is necessary for working with AccountStore and AccountReader, since those interfaces need to be able to manage multiple slots.

Once you narrow down to SlotAccountReader, you're just interested in getting a view of accountsdb for a single individual slot. Now it's fine to represent each account with just a single version that's valid for every slot it handles, since it only handles one slot. That's why the simple map is suitable in there.

When reading accounts with `AccountReader`, the simple map would just return whatever account is present, and ignore the slots that were requested by the caller. In the existing code, that means it would discard the pointer to ancestors passed into `forSlot` (in my other branch, AccountReader has another method where you ask for all the accounts from a specific slot). I suppose this could be fine when you're trying to keep your tests simple and you just want to populate the data to return, and don't care about obscuring some accounts.

When storing accounts with `AccountStore` is where it starts to get weirder. What is `AccountStore.put` supposed to do? Let's say you put an account from slot 9 and then you put the same account from slot 10. Should it overwrite the old account? What if you put 10 first then 9? I think overwriting data from other slots is more likely to cause confusion than be helpful.

src/runtime/account_loader.zig
dnut: This wasn't part of my change, and I'm not sure I agree this is an improvement. I'd rather avoid declaring pointless namespaces. I would like to just declare the function name in the current scope, and that's it. That being said, I also agree the way to achieve this in zig is pretty ugly.

I don't have a decisive enough preference here, and it's easy enough for me to change it to the way you like.

c2eb6e676e3585793e28f215c9055b06a5f8b9eb

src/accountsdb/account_store.zig
InKryption: All fair points, I guess I didn't think very deeply about the implications for the API in terms of shared representation. I was thinking it could just be like `thread_safe_map`, but without the thread safety, ie `simple_map: *Map(Pubkey, ArrayList(struct { Slot, Account }))`.
I'm not hard-set on this idea, but if you think it wouldn't be unreasonable to define it this way, I think that would be nice - I'll leave that decision up to you.

Feel free to resolve this comment at your discretion.

dnut: I haven't selected the best names for these things

src/runtime/account_loader.zig
InKryption: I'd concur it would be nice to be able to just declare functions inside functions directly, but it is what it is. I think it's harsh to call the namespace pointless, because viewed differently, the `}.f` is the non-zig way to accomplish this, whilst using a dedicated namespace for local functions is the intended way of doing it.

Again, still not ideal, but I think it has its merits.

dnut: what makes it non-zig if the syntax is valid and it works as intended?

InKryption: I mean non-zig in the sense that it doesn't follow the path of least resistance; it comes from a place of being used to do similar things in other languages, and trying to force zig features to do the same (regardless of whether one would argue that it *should* be able to do the same), where in the end it both reads awkwardly, and is awkward to write. It also offers resistance when refactoring, whether you're changing it to use a dedicated namespace, or you're adding a new function in the same way.

It's not an objectively right or wrong thing, it's moreso just that there's a way the creators of the tool intended it to be used, vs how you want to use it. The subpar aesthetics of the latter serve as an indicator of the former.

src/accountsdb/account_store.zig
Sobeston: Not sure about the simple_map allocator. It looks like get is used on the map directly, which wouldn't duplicate the account's memory (meaning we wouldn't need the allocator on deiniting it).

src/replay/service.zig
```diff
@@ -271,13 +270,12 @@ fn trackNewSlots(
...
     allocator: Allocator,
-    account_reader: AccountReader,
+    account_reader: SlotAccountReader,
     slot: Slot,
-    ancestors: *const sig.core.Ancestors,
 ) !sig.core.FeatureSet {
     var features = std.AutoArrayHashMapUnmanaged(Pubkey, Slot).empty;
     for (sig.core.FEATURES) |pubkey| {
-        const feature_account = try account_reader.get(pubkey, ancestors) orelse continue;
+        const feature_account = try account_reader.get(pubkey) orelse continue;
```
Sobeston, line 278: bit of a nit: is it really better to move the ancestors into the struct?
Also there might be a better name than SlotAccountReader :thinking:

Sobeston (changes requested):
Overall like this, just got some Qs

src/replay/service.zig
InKryption: My two cents: I personally think it's a marked improvement to have the caller provide the scope of accessible accounts rather than add that to the list of responsibilities of the function.

dnut: There are a bunch of contexts that need to get an account as it appears for a particular slot. The entire context only cares about that one slot, and there's only one valid version of each account for that slot. Ancestors are only here to help the account reader to do its job. But otherwise it's not really relevant to the context.

SlotAccountReader is meant to be the cleanest abstraction. It allows the code here to simply express a dependency on what it actually needs (account reader for a particular slot) and nothing else (like implementation details about how a particular account reader should find the correct version of an account).

If you need to provide ancestors at every call site, you intertwine that mechanism with every layer of the code, since it needs to propagate through every function signature. I don't see the advantage of that level of coupling. It's less flexible and can't be changed as easily.

Can you think of a better name for SlotAccountReader?

Sobeston: Okay yeah sounds good. My only peeve is that "SlotAccountReader" kinda sounds like it can only read accounts from one slot, but really it can read from a set of slots? Can't think of a better name so feel free to dismiss this if you can't either

src/accountsdb/account_store.zig
dnut: good point. is this better? 873fb8df3e757edb15492930ab6f4acfeaed50d0

Sobeston: Yep. Alternatively it could be a no-op allocator? But this is good

src/replay/service.zig
dnut: I see what you mean. You're saying that an account comes from a particular slot because that's where it was modified. You'd expect a "slot" reader should only read accounts that were modified in that particular slot.

I was thinking about it a little differently. Each slot has its own version of *all* accounts. In my mind, all the accounts roll-over from slot to slot. If slot 11 builds off slot 10, all of slot 10's accounts also exist in slot 11. So you should be able to read those accounts from slot 11 too.

Imagine a different approach to accountsdb where it snapshots the entire database for each slot. The moment 10 is frozen and you decide to start executing slot 11, then you copy a snapshot of the entire database from slot 10 into another database for slot 11. Then you can use slot 11's database to process the transactions in slot 11.

Of course that approach would have absurdly bad performance but I think it describes how accountsdb is supposed to *feel* from the user perspective. Under the hood it uses a copy-on-write approach that currently relies on Ancestors, but ultimately it's trying to achieve this feeling like you have a distinct database for each slot.

src/accountsdb/account_store.zig
dnut: I made it panic because I'd like be alerted if it's being misused.

src/replay/service.zig
Sobeston: Yeah that makes a lot of sense :+1: 

A clarifying comment would be nice, but feel free to skip it

dnut: 90b7883f63e5ad717ebc618a653f57cd2f096382

src/accountsdb/account_store.zig
dnut: This is elaborated in our other conversation, but I just wanted to conclude things here.

I see the appeal of this, but I think one of the reasons for having separate interfaces represented by different unions that have different types is because they can be satisfied in different ways. Using the same enum provides a strong assertion about consistency, but I think it would be misleading to take this approach here because it asserts a higher degree of consistency than what actually exists.

src/accountsdb/account_store.zig
dnut: I tried to improve naming/docs: 3b4d40c39e689c84bdf4a0e73a2b29f5028b4e00

InKryption (dismissed):

src/replay/service.zig
Sobeston: awesome

Sobeston (dismissed):
:+1:

Sobeston (dismissed):

InKryption (approved):

ultd (dismissed):

PR #821, opened 2025-07-25
feat(accountsdb): improve validator startup from snapshot

Sobeston:
- [guard against zero-sized snapshots](https://github.com/Syndica/sig/pull/821/commits/14a16a50da5cac2fab6f4657f41a747935e55b0c)
- [reduce allocs when loading snapshots](https://github.com/Syndica/sig/pull/821/commits/e4f669d254c51a1874e5643231a77c7b48dc9dd0)
- [reduce computeAccountHashesAndLamports threads](https://github.com/Syndica/sig/pull/821/commits/d18813233e51b6e7b7549333d887198eb848a3ab)

### DIFF
--- a/src/accountsdb/db.zig
+++ b/src/accountsdb/db.zig
@@ -802,11 +802,7 @@ pub const AccountsDB = struct {
             const n_accounts_this_slot = blk: {
                 var n_accounts: usize = 0;
 
-                var buffer_pool_frame_buf: [@sizeOf(u32) * 3]u8 = undefined;
-                var fba = std.heap.FixedBufferAllocator.init(&buffer_pool_frame_buf);
-                const frame_allocator = fba.allocator();
-
-                var iter = accounts_file.iterator(frame_allocator, &self.buffer_pool);
+                var iter = accounts_file.iterator(&self.buffer_pool);
                 while (try iter.nextNoData()) |account| {
                     n_accounts += 1;
                     shard_counts[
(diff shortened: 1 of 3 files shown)


src/accountsdb/snapshots.zig
```diff
@@ -1467,6 +1467,9 @@ pub fn parallelUnpackZstdTarBall(
 
     const file_size = (try file.stat()).size;
 
+    // calling posix.mmap on a zero-sized file will cause illegal behaviour
+    if (file_size == 0) return error.ZeroSizedTarball;
```
dnut, line 1471: I wish `std.posix.mmap` would do this. Maybe we should just write our own mmap function. Not now though. Feel free to resolve this comment.

src/accountsdb/db.zig
```diff
@@ -802,7 +802,11 @@ pub const AccountsDB = struct {
             const n_accounts_this_slot = blk: {
                 var n_accounts: usize = 0;
 
-                var iter = accounts_file.iterator(self.allocator, &self.buffer_pool);
+                var buffer_pool_frame_buf: [@sizeOf(u32) * 3]u8 = undefined;
+                var fba = std.heap.FixedBufferAllocator.init(&buffer_pool_frame_buf);
+                const frame_allocator = fba.allocator();
+
+                var iter = accounts_file.iterator(frame_allocator, &self.buffer_pool);
```
dnut: How do you know 12 bytes is enough?

Why didn't you put the fba in `readAccountNoData` and remove the allocator parameter?

dnut (dismissed):
Looks good and we can probably merge it as is. I just have one question I'd like to address first

InKryption (approved):
lgtm

src/accountsdb/snapshots.zig
Sobeston: Yeah, I'm tempted to make a Zig issue about std.posix UB stuff, it's annoying

src/accountsdb/db.zig
Sobeston: > How do you know 12 bytes is enough?

The allocation is based on how many buffer pool frames are required, 3 allows for us to always be able to read the account header.

> Why didn't you put the fba in readAccountNoData and remove the allocator parameter?

Good point, I should probably change the iterator a little to accommodate this

Sobeston: b17e24e0e3cad7fed49f8c3056e24d9998307d37

dnut (approved):

PR #824, opened 2025-07-28
refactor(consensus:ProgressMap): delete stubs, unify types, re-house some decls

InKryption:
* SortedMap: add/refactor some methods
* Unify types for progress map type
* Delete arc rwlock stubs
  It turns out there isn't really a broad need for fine-grained synchronization over these parts of the code, not currently or soon anyway.


src/accountsdb/account_store.zig
dadepo, line 1: This PR in general (not just this file) looks more like a refactoring one rather than one where a new feature is being added. Will that be a correct assessment? @InKryption 

I think this should be able to get merged in after [ink/replay-state-init](https://github.com/Syndica/sig/tree/ink/replay-state-init) - only thing stopping that would be to resolve all the current conflicts

PR #825, opened 2025-07-28
refactor(sysvars): use account store interfaces instead of accountsdb

dnut:
The sysvar code is the last remaining thing that's used by replay which still depends directly on accountsdb. In this change, I switch it over to using the same interfaces used in the rest of replay. This will simplify the replay code and make it easier to test everything.


InKryption (approved):
LGTM

dnut:
superseded by #816

PR #826, opened 2025-07-29
feat(ci): migrate sig-fuzz into the sig repo 

Rexicon226:
This PR moves over the main harness implementations from https://github.com/Syndica/sig-fuzz and adds some infrastructure for running https://github.com/firedancer-io/test-vectors in the CI. There are a few harnesses which are currently disabled, either because they haven't been implemented yet or because there are bugs present in the implementations.


conformance/scripts/do_fixtures.sh
```diff
@@ -0,0 +1,17 @@
...
+
+echo "Generating fixtures"
+solana-test-suite create-fixtures \
+-s ../libsolfuzz_agave.so \
+-i "../test-inputs"
+
+echo "Running fixtures"
+solana-test-suite exec-fixtures \
+-t ../conformance/zig-out/lib/libsolfuzz_sig.so \
+-i test_fixtures/ | tee /dev/tty | grep "Failed: 0,"
```
dnut: What's going on here? Does this duplicate the summary at the end to make it more visible? Why only do this if there are no failures?

.circleci/config.yml
```diff
@@ -241,6 +253,74 @@ jobs:
...
+            - test-vectors/
+      - run:
+          name: Run Fixtures
+          command: |
+            split_tests=$(./conformance/scripts/get_fixtures.sh | circleci tests split)
+            echo "$split_tests" > tests-to-run.txt
+            mkdir test-inputs/
+            xargs -a tests-to-run.txt cp -t test-inputs/
+            ./conformance/scripts/do_fixtures.sh
+
```
dnut: It seems like this could just be one shell script but you have it split across three files. Is there a specific reason for that? are those scripts independently useful?

dnut (dismissed):
Looks good, just a few questions

scripts/download_artifacts.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
```
dnut: Why is this in the main repo's script folder whereas the other scripts are nested under conformance?

conformance/scripts/do_fixtures.sh
Rexicon226: the `tee` duplicates that output to tty so we can see it in the console as well as to grep so it can match for `Failed: 0,` (which could only appear in the stdout if no tests failed).

.circleci/config.yml
Rexicon226: Originally, it was split up in a completely different way, before I recently simplified it. That required different parts of the command to run in different steps. I will move everything into the same script now.

scripts/download_artifacts.sh
Rexicon226: Also just has to do with it being an older part of the code, before I decided where to place the harness files. Will refactor the scripts in a minute.

conformance/scripts/do_fixtures.sh
dnut: I understood this much already, so I'm still curious about my original questions. To elaborate on that...

What's the point of duplicating the output to the same place it's already going? My assumption was that the grep output would be delayed relative to the tee output, which puts the message all the way at the end, so the summary is more prominent. Is that the idea?

And why only when there are no failures? Isn't the summary always good to see?

scripts/download_artifacts.sh
Rexicon226: [57ae087](https://github.com/Syndica/sig/pull/826/commits/57ae08766347756e7d3f7cc009c2cf344cc37384)

.circleci/config.yml
Rexicon226: [57ae087](https://github.com/Syndica/sig/pull/826/commits/57ae08766347756e7d3f7cc009c2cf344cc37384)

dnut (approved):

PR #828, opened 2025-07-30
feat(replay, consensus): Process replay results

dadepo:


src/replay/execution.zig
```diff
@@ -297,3 +302,777 @@ fn replaySlot(state: *ReplayExecutionState, slot: Slot) !ReplaySlotStatus {
...
+                    replay_state.account_store,
+                    &(replay_state.epochs.getForSlot(slot) orelse return error.MissingEpoch),
+                    slot_info.state,
+                    slot_info.constants,
+                    slot,
+                    entries[entries.len - 1].hash,
+                ));
+            }
+
+            processed_a_slot = true;
```
dnut: This entire function is one big loop. Why not instead make this function handle a single status, and call it in a loop? This will reduce the nesting a bit which helps readability. Also, this enables you to await the future in the caller instead of coloring this function with the async/await pattern. This simplification will make the function easier to understand and easier to use. I think it might simplify your tests too.

I think you could leave `replayActiveSlots` exactly as it was before this PR, except you can just add one line of code after freezing the slot to call this function. Likewise you could remove most of the code I highlighted here

dadepo: An attempt to simplify the loop [9ebf31b](https://github.com/Syndica/sig/pull/828/commits/9ebf31b1cdafc1b772ca585c3cb9d4d1aa966979)

Still left `processReplayResults` as the overarching method for processing the replay results. But within it, moved the awaiting, and the updates post freezing.

Mocking the future made the tests a bit harder to write, but kept it, as I did not see the code path around the future being tested.

src/replay/execution.zig
```diff
@@ -303,63 +303,160 @@ fn replaySlot(state: *ReplayExecutionState, slot: Slot) !ReplaySlotStatus {
...
+/// should be skipped (non-confirm status, pending after retries, or error).
+fn awaitConfirmedEntriesForSlot(
+    replay_state: *ReplayExecutionState,
+    slot: Slot,
+    status: ReplaySlotStatus,
+) !?[]const sig.core.Entry {
+    return switch (status) {
+        .confirm => |confirm_slot_future| blk: {
+            var attempts: u8 = 0;
+            while (attempts < 3) : (attempts += 1) {
```
dadepo: Still not sure if there is any benefit to having an upper limit for polling before giving up.

dnut: This is definitely not going to work because it only sleeps 1 nanosecond on every iteration and you're only giving it 3 chances. It's going to fail immediately, because execution is not instantaneous. I think it would be fine to just remove this check and we can revisit it later

src/replay/execution.zig
```diff
@@ -297,3 +302,785 @@ fn replaySlot(state: *ReplayExecutionState, slot: Slot) !ReplaySlotStatus {
...
+                );
+            }
+        }
+        // If `slot_hashes` becomes empty, it'll be removed by `setRoot()` later
+    }
+}
+
+pub fn processReplayResults(
+    replay_state: *ReplayExecutionState,
+    slot_statuses: *const std.ArrayListUnmanaged(struct { Slot, ReplaySlotStatus }),
```
dnut: code smell: passing a `*const ArrayList`

```suggestion
    slot_statuses: []const struct { Slot, ReplaySlotStatus },
```

src/replay/execution.zig
```diff
@@ -93,42 +133,7 @@ pub fn replayActiveSlots(state: *ReplayExecutionState) !bool {
...
-                future.entries[future.entries.len - 1].hash,
-            ));
-        } else {
-            state.logger.info().logf("partially confirmed slot {}", .{slot});
-        }
-    }
-
-    // TODO: process_replay_results: https://github.com/anza-xyz/agave/blob/3f68568060fd06f2d561ad79e8d8eb5c5136815a/core/src/replay_stage.rs#L3443
+    processed_a_slot = try processReplayResults(state, &slot_statuses);
     return processed_a_slot;
```
dnut:
```suggestion
    return try processReplayResults(state, &slot_statuses);
```

dnut: github is being glitchy and it's not displaying correctly here when I look at it. my suggestion is to replace this:
```zig
    var processed_a_slot = false;
    processed_a_slot = try processReplayResults(state, &slot_statuses);
    return processed_a_slot;
```
with this:
```zig
    return try processReplayResults(state, &slot_statuses);
```

src/replay/execution.zig
```diff
@@ -297,3 +302,785 @@ fn replaySlot(state: *ReplayExecutionState, slot: Slot) !ReplaySlotStatus {
...
+            }
+            // Return null if all attempts failed
+            break :blk null;
+        },
+        else => null,
+    };
+}
+
+/// Applies fork-choice and vote updates after a slot has been frozen.
+fn finalizeSlotPostFreeze(
```
dnut: I know I've been using the word "finalize" for this but actually I think it can cause some confusion because these state changes are mostly important for consensus, meanwhile in consensus "finalize" has a special meaning (which is not what this function does).

It's hard to think of a good name but how about something like `processSlotPostFreeze` or `accountForFrozenSlot`. Can you think of something that doesn't use the word "finalize"?

src/replay/execution.zig
dnut: I like it

src/replay/execution.zig
dadepo: [6c80766](https://github.com/Syndica/sig/pull/828/commits/6c8076655531719e9d391763af5608be897cb735)

src/replay/execution.zig
dadepo: [e12375b](https://github.com/Syndica/sig/pull/828/commits/e12375ba87159e4b39926b971adf6823f5079019)

src/replay/execution.zig
dadepo: [66d7212](https://github.com/Syndica/sig/pull/828/commits/66d72125ad958beae8026948e5dbbf76d9c54219)

src/replay/execution.zig
dadepo: What about `updateConsensusForFrozenSlot`? Since the updates are consensus related

 [736ae21](https://github.com/Syndica/sig/pull/828/commits/736ae213ebed43458006c152841509f47805b913)

dnut: I think that's better than either of my suggestions

dnut (approved):

InKryption (approved):

PR #831, opened 2025-07-31
refactor(features): rework the feature set abstraction

Rexicon226:
A follow-up change will work on refactoring the feature set higher up, storing it in the epoch constants rather than the slot constants. This PR amends the call sites to ensure the slot is being passed in properly, but it isn't necessary yet as the feature set is per-slot atm.

- Removes allocations from the feature set. There should never be a need to do so, since the number of features is well known at compile time.
- Provides a single source of truth regarding the name of the feature, its pubkey, and whether it has been activated on all clusters.
- Makes referencing the features more type-safe as well.


InKryption (dismissed):

src/core/features.zig
```diff
@@ -4,992 +4,197 @@ const sig = @import("../sig.zig");
...
+pub const map: std.EnumArray(Feature, Pubkey) = map: {
+    @setEvalBranchQuota(NUM_FEATURES * 1000);
+    var s: std.enums.EnumFieldStruct(Feature, Pubkey, null) = undefined;
+    for (@typeInfo(Feature).@"enum".fields, features) |field, feature| {
+        @field(s, field.name) = .parse(feature.pubkey);
+    }
+    break :map .init(s);
+};
+
+/// Represents the set of currently enabled feature flats.
```
yewman:
```suggestion
/// Represents the set of currently enabled feature flags.
```

conformance/src/bank_methods.zig
```diff
@@ -164,14 +158,14 @@ fn applyBuiltinProgramFeatureTransitions(
 
     for (builtins.STATELESS_BUILTINS) |builtin_program| {
         const core_bpf_config = builtin_program.core_bpf_migration_config orelse continue;
-        if (new_feature_activations.contains(core_bpf_config.enable_feature_id)) {
+        if (new_feature_activations.active(core_bpf_config.enable_feature_id, 0)) {
```
yewman, line 161: Why use 0 here and below, but use slot elsewhere in this method?

conformance/src/builtins.zig
```diff
@@ -3,17 +3,14 @@ const sig = @import("sig");
...
 // Live source buffer accounts for builtin migrations
-const STAKE_PROGRAM_SOURCE_BUFFER_ADDRESS =
-    Pubkey.parseBase58String("8t3vv6v99tQA6Gp7fVdsBH66hQMaswH5qsJVqJqo8xvG") catch unreachable;
-const CONFIG_PROGRAM_SOURCE_BUFFER_ADDRESS =
-    Pubkey.parseBase58String("BuafH9fBv62u6XjzrzS4ZjAE8963ejqF5rt1f8Uga4Q3") catch unreachable;
-const ADDRESS_LOOKUP_TABLE_SOURCE_BUFFER_ADDRESS =
-    Pubkey.parseBase58String("AhXWrD9BBUYcKjtpA3zuiiZG4ysbo6C6wjHo1QhERk6A") catch unreachable;
-const FEATURE_PROGRAM_SOURCE_BUFFER_ADDRESSS =
-    Pubkey.parseBase58String("3D3ydPWvmEszrSjrickCtnyRSJm1rzbbSsZog8Ub6vLh") catch unreachable;
+const STAKE_PROGRAM: Pubkey = .parse("8t3vv6v99tQA6Gp7fVdsBH66hQMaswH5qsJVqJqo8xvG");
```
yewman: Could we keep some meaningful suffix here? just STAKE_PROGRAM is not great imo.

conformance/src/txn_execute.zig
```diff
@@ -1036,38 +1039,28 @@ fn loadSlot(pb_txn_ctx: *const pb.TxnContext) u64 {
...
-        var feature_set = features.FeatureSet.EMPTY;
+        const pb_features = maybe_pb_features orelse break :set .ALL_DISABLED;
 
+        var feature_set: FeatureSet = .ALL_DISABLED;
         for (pb_features.features.items) |id| {
-            if (indexed_features.get(id)) |pubkey| {
-                try feature_set.active.put(allocator, pubkey, 0);
-            }
+            // only way for `setId` to fail is if the `id` doesn't exist.
+            feature_set.setSlotId(id, 0) catch continue;
```
yewman: Maybe we don't want to just continue here? It a fixture specifies a feature which we are missing we probably want to be aware of it

src/core/features.zig
```diff
@@ -4,992 +4,197 @@ const sig = @import("../sig.zig");
...
 /// [agave] https://github.com/anza-xyz/agave/blob/8db563d3bba4d03edf0eb2737fba87f394c32b64/sdk/feature-set/src/lib.rs#L1188
-pub const FeatureSet = struct {
-    active: std.AutoArrayHashMapUnmanaged(Pubkey, Slot),
-
-    pub const EMPTY: FeatureSet = .{ .active = .{} };
-
-    pub fn deinit(self: FeatureSet, allocator: std.mem.Allocator) void {
-        var active_ = self.active;
-        active_.deinit(allocator);
+pub const Set = struct {
```
yewman, line 42:
```suggestion
pub const FeatureSet = struct {
```

src/core/features.zon
```diff
@@ -0,0 +1,242 @@
+.{
+    .{ .name = "deprecate_rewards_sysvar", .pubkey = "GaBtBJvmS4Arjj5W1NmFcyvPjsHN38UGYDq2MDwbs9Qu" },
```
yewman, line 2: Just out of curiosity how did you generate this / where did you pull this info from?

src/core/hash.zig
```diff
@@ -93,7 +93,13 @@ pub const Hash = extern struct {
         return &self.data;
     }
 
-    pub fn parseBase58String(str: []const u8) error{InvalidHash}!Hash {
+    pub inline fn parse(comptime str: []const u8) Hash {
```
yewman, line 96: I'd prefer comptime distinction over runtime distinction i.e. `parseComptime` and `parse` vs. `parse` and `parseRuntime`

src/net/socket_utils.zig
```diff
@@ -621,15 +621,21 @@ pub const BenchmarkPacketProcessing = struct {
 
         // run incoming until received n_packets
 
+        std.debug.print("reciving: {}\n", .{n_packets});
```
yewman: Are these changes intended for this PR? And are these debug statements desired?

src/replay/update_sysvar.zig
```diff
@@ -863,21 +862,29 @@ test "update all sysvars" {
...
+            allocator,
+            account_reader,
+        )).?;
         defer allocator.free(old_account.data);
 
-        try updateLastRestartSlot(allocator, &feature_set, &hard_forks, update_sysvar_deps);
+        try updateLastRestartSlot(
+            allocator,
+            &feature_set,
+            slot,
```
yewman, line 884: Rather than passing a feature set and slot everywhere we want to check activations, what do you think about creating a struct which wraps the feature set and slot and just provides an `active(.variant)` method. Such a struct could replace the `FeatureSet` in the TransactionContext as the slot is constant.

conformance/src/bank_methods.zig
Rexicon226: Must have just missed it, will use the slot here.

conformance/src/builtins.zig
Rexicon226: I'd rather not. We could place them into a `feature_program` namespace if you'd like.

conformance/src/txn_execute.zig
Rexicon226: Probably a good point, will experiment to see if we hit this continue and for what.

src/core/features.zig
Rexicon226: I don't see why we would do this, it's clear that it's a feature set given it's under `features.zig`. This should be a file-as-struct in the first place, but for now `Set` is the second best.

src/core/features.zon
Rexicon226: I got it from here: https://github.com/firedancer-io/firedancer/blob/2903ecc00d6276179a7105c13a2c70f14212e70b/src/flamenco/features/feature_map.json
I will add a link to the top.

src/core/hash.zig
Rexicon226: I do not understand why - 99% of usages are comptime. There are only a handful (literally) runtime usages. On top of that, runtime parsing can take in a comptime string while the comptime parsing cannot take in a runtime string.

src/net/socket_utils.zig
Rexicon226: whoops, was just debugging some issue, forgot to remove.

src/replay/update_sysvar.zig
Rexicon226: I do not think it's worth it - we will want to keep the feature set in the epoch constants and the slot will not be constant.

src/core/hash.zig
yewman: I don't understand why not - why does it matter that 99% of usages are comptime? The comptime method is more specific and therefore should have the suffix imo.

conformance/src/builtins.zig
yewman: Why not?

src/replay/update_sysvar.zig
yewman: In contexts where the slot is constant I think it would improve readability and avoid needing to pass the slot everywhere we want to check a feature activation. Fundamentally you are passing the same data, it is just wrapped a little differently. What is the cost that makes it not worth it from your perspective?

conformance/src/builtins.zig
Rexicon226: It is redundant. Having a `feature_programs` namespace would be a better solution.

src/core/hash.zig
Rexicon226: Why would you increase friction for the more common behaviour? We're not designing for some future use cases; we already have hundreds of examples that prove the comptime case is more common. It only makes sense to reduce friction for it.

src/replay/update_sysvar.zig
Rexicon226: My concern is that you would be less aware of what the slot is when passing it in. I am imagining some use cases when we would need to check what the feature status is for a slot that is not the current one. Are there any cases when it's possible to plug in the *wrong* slot? If there are, then I would agree that wrapping it is better.

src/net/socket_utils.zig
Rexicon226: https://github.com/Syndica/sig/compare/e983d14ca0284afd74500742f77a249f5a1580d6..4d604b9587b9eed8fc347cbd315a079daf0eabb0

src/core/features.zon
Rexicon226: https://github.com/Syndica/sig/compare/e983d14ca0284afd74500742f77a249f5a1580d6..4d604b9587b9eed8fc347cbd315a079daf0eabb0

conformance/src/txn_execute.zig
Rexicon226: we don't see to hit any panics, so i've hardcoded it to be a panic
https://github.com/Syndica/sig/compare/e983d14ca0284afd74500742f77a249f5a1580d6..4d604b9587b9eed8fc347cbd315a079daf0eabb0

conformance/src/builtins.zig
Rexicon226: https://github.com/Syndica/sig/compare/e983d14ca0284afd74500742f77a249f5a1580d6..4d604b9587b9eed8fc347cbd315a079daf0eabb0
put them under a `feature_programs` namespace

conformance/src/bank_methods.zig
Rexicon226: https://github.com/Syndica/sig/compare/e983d14ca0284afd74500742f77a249f5a1580d6..4d604b9587b9eed8fc347cbd315a079daf0eabb0

src/core/features.zig
Rexicon226: https://github.com/Syndica/sig/compare/e983d14ca0284afd74500742f77a249f5a1580d6..4d604b9587b9eed8fc347cbd315a079daf0eabb0

conformance/src/txn_execute.zig
Rexicon226: uhh i take it back. i ran it locally and in CI but it never hit anyrhing. ill look into it in a bit

conformance/src/builtins.zig
yewman: What is the meaning behind using `feature_programs` as the name for this namespace?

I think it would be better to define these addresses where we define the program ID, and call it something like SOURCE_ID or SOURCE_ADDRESS since these are the addresses that will store the program source code once deployed as part of the core bpf migration.

conformance/src/txn_execute.zig
yewman: it hit in CI?

src/core/hash.zig
yewman: I don't see the increase in friction, but it is not a big deal so will concede

conformance/src/builtins.zig
Rexicon226: Sure, that's a good solution. My main complaint is against having naming.

src/replay/update_sysvar.zig
yewman: In the context of the SVM I don't think we would ever want to check the status of a feature for a slot other than the current slot. From the 'bank' level down, I think it would make sense to have a data structure which represents the active features at that slot but it's only a slight improvement over the current implementation so I am happy to leave it for now and reconsider down the track

conformance/src/builtins.zig
yewman: lets run with that then

InKryption (approved):

PR #833, opened 2025-08-01
ci(txn): enable more txn fixtures in CI

Rexicon226:


dnut (approved):
thank you for sorting the list. let's keep it that way

InKryption (approved):

PR #834, opened 2025-08-01
fix(runtime:program): Decode `ComputeBudgetInstruction` using BORSH codec

InKryption:
This type and relevant code interacting with it seem to have been incorrectly ported without taking into account that it doesn't actually use bincode, it uses [BORSH](https://borsh.io/).

The specific issue this causes is that the bincode decode process treats tagged union tags as `u32`s, whereas BORSH treats them as `u8`s, which was causing it to consume and interpret the 3 first bytes of the payload (or whatever else comes after in the absence of a payload) as part of the discriminant value.

In theory there's also other problems (that simply haven't and *probably* won't manifest in practice [probably, hopefully]), like the fact that where bincode permits encoding using either little or big endian, BORSH only permits little endian, and where bincode has a concept of varint encoding, BORSH always uses fixed-size int representation.

In general, while there's a good degree of overlap between bincode and BORSH, there's some small but important incompatibilities that make it very risky to mix the two.

I'm open to suggestions about the BORSH abstraction, I implemented the minimum possible amount of code whilst not entangling it too much with `ComputeBudgetInstruction`'s definition and present assumptions.

### DIFF
--- a/conformance/scripts/passing_txn_fixtures.txt
+++ b/conformance/scripts/passing_txn_fixtures.txt
@@ -97,6 +97,7 @@
 ./test-vectors/txn/fixtures/programs/0b8be053cedbc396b3f09a9d2f0c7b53037c22cc_265678.fix
 ./test-vectors/txn/fixtures/programs/0bd6fd0865fb8ae01e942d6f75a9eb633ec8b7da_265678.fix
 ./test-vectors/txn/fixtures/programs/0bebe8e65997c3e62e10bfa2f75dbcf92b6f4a51_265678.fix
+./test-vectors/txn/fixtures/programs/0befdb28dece3603f48906dce6290d685e9350a2_265678.fix
 ./test-vectors/txn/fixtures/programs/0bf290c24cfb2ee249858b800d4cc3dc233d6e1f_265678.fix
 ./test-vectors/txn/fixtures/programs/0bfb5e747dc6e661c85ba8eb29ace47860b4ff60_265678.fix
 ./test-vectors/txn/fixtures/programs/0caad5d9b537de86e48738d8770348e9f62878bc_265678.fix
@@ -249,6 +250,7 @@
 ./test-vectors/txn/fixtures/programs/1da4678abf452747827af8cf700b438698267709_265678.fix
 ./test-vectors/txn/fixtures/programs/1da4a54df18bcd644f08c98679bd16f5c765d6ad_265678.fix
 ./test-vectors/txn/fixtures/programs/1da9dd696690f1dde3a5a3a26a6cf5a05b548db1_265678.fix
+./test-vectors/txn/fixtures/programs/1dd82551ca79878cd9a1cca6e802426d40e4fc63_265678.fix
 ./test-vectors/txn/fixtures/programs/1e00c719e0e2a72cdbcdbc344d9a14c2cd431d49_265678.fix
 ./test-vectors/txn/fixtures/programs/1e03116fe66d3fc9384b1e9a772542420b0bf2b2_265678.fix
 ./test-vectors/txn/fixtures/programs/1e0b754236d4454d21ca49af12a93bb345764267_265678.fix
@@ -557,6 +559,7 @@
 ./test-vectors/txn/fixtures/programs/443d0aa061a964f35bfade6623cebe6fa1f8b98f_265678.fix
 ./test-vectors/txn/fixtures/programs/4441f0a5f48351847b1c77d9caf6406fe4834e5c_2232445.fix
 ./test-vectors/txn/fixtures/programs/44778f3b81ff0f349f29cbad376273f0b093d6e4_265678.fix
+./test-vectors/txn/fixtures/programs/449595bafdb747f068022bcfa0f6edf51af8f190_265678.fix
 ./test-vectors/txn/fixtures/programs/449e46e5ad2a675fdbcdcd23470eeb69af776fd6_265678.fix
 ./test-vectors/txn/fixtures/programs/44a4b219f71c9881c2d1936b26c6e0dd6be84984_265678.fix
 ./test-vectors/txn/fixtures/programs/44b6d4ad676c785f99cccfeffca43f73f8834c84_2213508.fix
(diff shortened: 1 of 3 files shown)


InKryption:
* **#834** <a href="https://app.graphite.dev/github/pr/Syndica/sig/834?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Syndica/sig/834?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/runtime/program/compute_budget/lib.zig
```diff
@@ -109,8 +111,169 @@ const ComputeBudgetInstruction = union(enum(u32)) {
...
+        const pl_size: borsh.EncodedSizeInt = borsh.payloadEncodedSize(tag);
+        const pl_bytes = src[1..];
+        if (pl_bytes.len < pl_size) return error.TruncatedSrc;
+        const cbi = borsh.decodeFromTagAndPayload(tag, pl_bytes[0..pl_size]) orelse
+            return error.DecodeFailed;
+        return .{ cbi, @sizeOf(u8) + pl_size };
+    }
 };
 
+// TODO: put this somewhere logical, maybe create a more general abstraction?
```
yewman: perhaps a borsh module in bincode which provides a complete implementation or borsh for compatible types?

InKryption: I would be against putting borsh in bincode, as that would give the incorrect impression that they are meaningfully related or compatible, when in fact they should never be interpreted in the same channel.

yewman: How about a separate borsh module in sig then?

InKryption: Yeah, I think that would be fine, and probably necessary if we want to be able to properly handle any other instances of BORSH encoding that may crop up (or which already exist but are being erroneously bincoded). Would rather leave that work outside of this PR though.

yewman: Hmm, in that case could you create a corresponding ticket? I want to avoid this ending up here

src/runtime/program/compute_budget/lib.zig
```diff
@@ -109,8 +111,169 @@ const ComputeBudgetInstruction = union(enum(u32)) {
...
+
+    /// Decodes the encoded ComputeBudgetInstruction, and the number of bytes it occupied in `src`.
+    pub fn borshDecodeSlice(
+        src: []const u8,
+    ) BorshDecodeError!struct { ComputeBudgetInstruction, borsh.EncodedSizeInt } {
+        if (src.len == 0) return error.TruncatedSrc;
+        const tag = borsh.tagDecode(src[0]) orelse return error.DecodeFailed;
+        const pl_size: borsh.EncodedSizeInt = borsh.payloadEncodedSize(tag);
+        const pl_bytes = src[1..];
+        if (pl_bytes.len < pl_size) return error.TruncatedSrc;
```
dnut, line 132: Couldn't you check this first and skip the 0 check?

src/runtime/program/compute_budget/lib.zig
```diff
@@ -109,8 +111,169 @@ const ComputeBudgetInstruction = union(enum(u32)) {
...
+            return error.DecodeFailed;
+        return .{ cbi, @sizeOf(u8) + pl_size };
+    }
 };
 
+// TODO: put this somewhere logical, maybe create a more general abstraction?
+fn BorshStaticTaggedUnionHelper(comptime Tagged: type) type {
+    const Tag, const has_payloads = switch (@typeInfo(Tagged)) {
+        .@"union" => |union_info| .{ union_info.tag_type.?, true },
+        .@"enum" => .{ Tagged, false },
```
dnut: Why do we need to handle this case? It's not used for enums now, and if we need to serialize enums, isn't it already trivial without this helper? It might simplify things to deal with them separately.

src/runtime/program/compute_budget/lib.zig
```diff
@@ -109,8 +111,169 @@ const ComputeBudgetInstruction = union(enum(u32)) {
...
+        else => @compileError("Unsupported type " ++ @typeName(Tagged)),
+    };
+    if (@sizeOf(Tag) > @sizeOf(u8)) @compileError("Tag type is too big");
+
+    const pl_size_min, const pl_size_max = blk: {
+        var min: usize = @sizeOf(Tagged);
+        var max: usize = 0;
```
dnut: If there are no fields, the minimum is larger than the maximum. Why don't min and max both start at the same value? It seems like maybe they should both start at 0, since the tag is given 1 byte and you account for that separately.

src/runtime/program/compute_budget/lib.zig
InKryption: 11e63083772c925a2ccb3aac5de780f08540015f & #849.

src/runtime/program/compute_budget/lib.zig
InKryption: I'm confused what you mean, need to check `src.len == 0` first in order to access `src[0]`.

src/runtime/program/compute_budget/lib.zig
InKryption: min can't start at 0, because then it would always be smaller than any of the fields.

src/runtime/program/compute_budget/lib.zig
InKryption: 1ac743d1cb44b538039e16610bdb708314d6da4b

src/runtime/program/compute_budget/lib.zig
InKryption: 1ac743d1cb44b538039e16610bdb708314d6da4b

dnut: I think `min` should start at `maxInt` when there are fields. It will be too small if the tag is smaller than all of the fields.

I was pretty confused by this code when I first looked at it. I'm looking for ways to simplify it and make it less esoteric, so anyone can quickly look at it and understand it all easily.

1. `min` is unnecessary because you can simply use 0 as the `from` value in IntFittingRange and it will return the same type.
2. EncodedSizeInt and EncodedPayloadSizeInt could be replaced with `usize` for clarity.

You could probably eliminate all the pub consts except for `ENCODED_SIZE_MAX`.

Are there reasons for all these extra pieces? I think the code would be easier to understand without them.

InKryption: `@sizeOf(Tagged)` can never be smaller than the fields, since it's the sum of the tag and the largest field.

Suppose min isn't needed or useful at present, I just usually try to use the appropriate ranges for things in preparation for the arrival of ranged integers. Having an integer type that can only represent the exact range of lengths that are actually possible is useful.
In this case, I'm fine with simplifying this some, since I don't intend for this to be anywhere near the final design for how BORSH encoding should work, and don't want to block merging this PR over disputes about the particular design specifics.

InKryption: 723dcc527a5b26eddd06cd3c56a609c85ef28802

dnut: > @sizeOf(Tagged) can never be smaller than the fields, since it's the sum of the tag and the largest field.

Oh right I forgot this was the whole union and not the tag.

OK this is a nice improvement in my eyes. I still feel that `EncodedSizeInt` is unnecessary and the only thing it does is make the code confusing. But let's just get this merged.

yewman (approved):

PR #836, opened 2025-08-04
fix(txn): writable account check

kprotty:
Fix accounts accidently presenting as writable and showing up in conformance tx output.


InKryption:
Once both this and #838 are merged, fixture 0136006d0f996e9945f0f29b51777e28fe515f0d_2190571 should also be passing.

InKryption (approved):

src/core/transaction.zig
```diff
@@ -393,47 +393,12 @@ pub const Message = struct {
...
             index < self.signature_count and
             index >= self.signature_count - self.readonly_signed_count;
         const is_readonly_unsigned = index >= self.account_keys.len - self.readonly_unsigned_count;
-        const is_writable = !(is_readonly_signed or is_readonly_unsigned);
-
-        return is_writable and for (RESERVED_ACCOUNTS) |reserved| {
-            if (reserved.equals(&self.account_keys[index])) break false;
-        } else true;
+        return !(is_readonly_signed or is_readonly_unsigned);
     }
```
dnut: Why did you remove this? This seems like the best place for this code since it's the most widely applicable function that's likely to be used in the most places for determining writability. Also this is where agave has this logic.

kprotty: Its equivalent to agave's [`Message.is_writable_index`](https://github.com/anza-xyz/solana-sdk/blob/5ff67c1a53c10e16689e377f98a92ba3afd6bb7c/message/src/versions/v0/loaded.rs#L118), but its `is_writable()` actually hits a hashset thats built with that + [`is_writable_internal`](https://github.com/anza-xyz/solana-sdk/blob/5ff67c1a53c10e16689e377f98a92ba3afd6bb7c/message/src/versions/v0/loaded.rs#L139-L150) which is what this code adds. But doing that requires the ALUT accounts in addition to `Message.account_keys` (for computing `is_upgradeable_loader_present`). Those aren't available in our `Message` type, so opted to instead do it where they are + only place where `isWritable` is called: `resolveTransaction()`

dnut: I see. It's tricky to come up with a clean general solution here.

Would you mind renaming `isWritable` to `isWritableIndex` and adding a doc comment describing its limited role?

kprotty: https://github.com/Syndica/sig/pull/836/commits/fa2d36d9579a4275bf5c328dc488504df518d1d0

dnut: nice

yewman (approved):

PR #838, opened 2025-08-05
feat(vm): implement `secp256k1_recover` syscall

Rexicon226:


src/vm/syscalls/ecc.zig
```diff
@@ -611,6 +615,55 @@ fn altBn128Operation(
...
+        registers.set(.r0, 2); // InvalidRecoveryId
+        return;
+    }
+
+    const pubkey = sig.runtime.program.precompiles.secp256k1.recoverSecp256k1Pubkey(
+        hash,
+        &signature.to(),
+        @intCast(recovery_id_val),
+    ) catch {
+        registers.set(.r0, 3);
```
yewman:
```suggestion
        registers.set(.r0, 3); // InvalidSignature
```

Rexicon226: https://github.com/Syndica/sig/compare/df5e5622b9a589e0e63e77657e2202019d7bd3af..03ae4b448df20c39b165c001040e4a768c81f53b

PR #840, opened 2025-08-05
feat(txn-harness): fixes for transaction fixture state validation prior to transaction execution

yewman:
Changes:

1. Fix precompile account insertion: do not overwrite when an account already exists and  is executable
2. Fix read lamports per signature from recent blockhashes account data: don't deserialise full sysvar from the account data, instead manually read the first entry if it exists to get the lamports per signature
3.  Fix loader v4 feature id
4. Fix conversion from blockhash queue to recent blockhashes
5. Fix hash slot to be consistent with agave and txn harness. To discuss with @dnut


Rexicon226 (approved):
looked through, nothing too controversial

src/core/rent_collector.zig
```diff
@@ -127,8 +127,10 @@ pub const RentCollector = struct {
     ) RentDue {
         if (self.rent.isExempt(lamports, data_len)) return .Exempt;
 
+        if (account_rent_epoch > self.epoch) return .{ .Paying = 0 };
+
         var slots_elapsed: u64 = 0;
-        for (account_rent_epoch..self.epoch + 1) |epoch| {
+        for (account_rent_epoch..self.epoch +| 1) |epoch| {
```
InKryption: Is the saturating arithmetic correct/needed here? I think the overflow that was being hit here was only due to `account_rent_epoch > self.epoch + 1`. If this happens to overflow, then I think that's a very different failure state which shouldn't be hidden by this saturating op.

src/runtime/program/bpf/execute.zig
```diff
@@ -133,6 +133,7 @@ pub fn execute(
         .ok => |status| if (status != 0) {
             const execution_error = sig.vm.executionErrorFromStatusCode(status);
             if (execution_error == error.Custom) ic.tc.custom_error = @intCast(status);
+            if (execution_error == error.GenericError) ic.tc.custom_error = 0;
```
InKryption: I would advise using a switch here; direct comparison against an error literal like this is dangerous, because in equality comparisons PTR takes place, and errors of any type are peer-resolved into a shared error set type. So any changes to the type of `execution_error` will not trigger any compile error here, even if we were to straight up remove or rename `GenericError`, because the comparison would simply evaluate both operands as `error{NewGenericErrorName} || error{GenericError}`, and always evaluate to false.

src/runtime/transaction_execution.zig
```diff
@@ -180,6 +180,7 @@ pub const CopiedAccount = struct {
 
 pub const ExecutedTransaction = struct {
     err: ?InstructionErrorEnum,
+    err_index: ?u32,
```
InKryption: I don't think these can be independently null or non-null, so I would suggest tying that state together: `err: ?struct { index: u32, value: InstructionErrorEnum }`.

src/runtime/transaction_execution.zig
```diff
@@ -445,6 +450,7 @@ pub fn executeTransaction(
             }
             break;
         };
+        instruction_index += 1;
```
InKryption: Suggestion:
```zig
const maybe_instruction_fail: ?struct { err: InstructionError, index: usize } =
    for (transaction.instruction_infos, 0..) |instruction_info, i| {
        executor.executeInstruction(
            allocator,
            &tc,
            instruction_info,
        ) catch |err| switch (err) {
            error.OutOfMemory => return error.OutOfMemory,
            else => |e| break .{ .err = e, .index = i },
        };
    } else null;
```
In combination with my other suggestion about tying together the nullability of `err` and `err_index`.

src/runtime/account_loader.zig
```diff
@@ -401,6 +403,7 @@ pub const BatchAccountCache = struct {
 
             const result = self.account_cache.getOrPutAssumeCapacity(key.*);
             std.debug.assert(result.found_existing);
+            allocator.free(result.value_ptr.data);
 
             result.value_ptr.* = account;
```
InKryption:
```suggestion
            const account_ptr = self.account_cache.getPtr(key.*).?;
            allocator.free(result.value_ptr.data);
            account_ptr.* = account;
```

conformance/src/txn_execute.zig
```diff
@@ -857,25 +850,19 @@ fn serializeOutput(
...
-                        0,
-                },
-                .fees_only => |fees_only| fees_only: {
-                    const err_codes = utils.convertTransactionError(fees_only.err);
-                    break :fees_only .{
-                        err_codes.err,
-                        err_codes.instruction_error,
-                    };
-                },
+            const converted: utils.Converted = switch (txn) {
```
InKryption: I think a name like `error_codes` would be better here.

InKryption (changes requested):

src/core/rent_collector.zig
yewman: 382b5b27

src/runtime/program/bpf/execute.zig
yewman: 382b5b27

src/runtime/transaction_execution.zig
yewman: 382b5b27

src/runtime/transaction_execution.zig
yewman: 382b5b27

src/runtime/account_loader.zig
yewman: 382b5b27

conformance/src/txn_execute.zig
yewman: 382b5b27

InKryption (approved):

InKryption (approved):

src/replay/resolve_lookup.zig
```diff
@@ -232,20 +229,31 @@ fn resolveLookupTableAccounts(
 fn getLookupTable(
     account_reader: SlotAccountReader,
     table_address: Pubkey,
-) !?AddressLookupTable {
+) !AddressLookupTable {
```
dnut, line 232: I think I prefer converting null into error at the call site, but it's not important.

InKryption (approved):

Rexicon226 (approved):

InKryption (approved):

dnut (approved):

src/replay/resolve_lookup.zig
dnut: feel free to resolve this

PR #842, opened 2025-08-05
fix(conformance): enable logs, filter rent_debit.rent_collected = 0

InKryption:
Once this changeset, #834, #836, and #838 are merged, fixture 0136006d0f996e9945f0f29b51777e28fe515f0d_2190571 should be passing


yewman (commented):
Could you add the newly passing fixtures

PR #845, opened 2025-08-06
fix(runtime,conformance): Check for duplicate accounts in transaction before loading and executing it

InKryption:


conformance/scripts/passing_txn_fixtures.txt
```diff
@@ -1745,3 +1745,75 @@
 ./test-vectors/txn/fixtures/programs/ffe6aa534ff1d1b32696b7e1314e26d0b84210e8_265678.fix
 ./test-vectors/txn/fixtures/programs/fff5e14663bd3ae176bd58bba58fefcd8a137e02_3004834.fix
 ./test-vectors/txn/fixtures/programs/fff87aff2cbaee9cdb9489c98db0adcc0309b97f_2202462.fix
+./test-vectors/txn/fixtures/programs/02f247a27a981000e10282625e1c489b9b6db90c_265678.fix
```
Rexicon226: Please sort the list lexigraphically.

src/runtime/transaction_execution.zig
```diff
@@ -378,6 +381,41 @@ pub fn loadAndExecuteTransaction(
...
+/// NOTE: in agave, this check is done while creating/loading the account batch:
+/// * [prepare_sanitized_batch](https://github.com/firedancer-io/agave/blob/10fe1eb29aac9c236fd72d08ae60a3ef61ee8353/runtime/src/bank.rs#L3173)
+/// * [try_lock_accounts](https://github.com/firedancer-io/agave/blob/10fe1eb29aac9c236fd72d08ae60a3ef61ee8353/runtime/src/bank.rs#L3164)
+/// * [lock_accounts](https://github.com/firedancer-io/agave/blob/10fe1eb29aac9c236fd72d08ae60a3ef61ee8353/accounts-db/src/accounts.rs#L569)
+/// * [validate_account_locks](https://github.com/firedancer-io/agave/blob/10fe1eb29aac9c236fd72d08ae60a3ef61ee8353/accounts-db/src/account_locks.rs#L122-L123)
+/// and then it is propagated to and through `load_and_execute_transactions`.
+///
+/// Our account batch creation/load process isn't designed to accommodate this, so what we do
+/// instead is do the check when we're actually trying to load and execute the transaction.
+fn hasDuplicates(account_keys: []const Pubkey) bool {
```
Rexicon226, line 398: Just a thought, but I don't really like having a O(n^2) here. There are more optimal ways to do this, although they'd require a more drastic shift in how we represent the transaction accounts. Perhaps in the future it's something to look into if this is called a lot for cases where we have a lot of accounts.

InKryption: Representing the accounts differently certainly would be a way to go about it, though one concern there is it would move the error to occur earlier than it otherwise would occur, which could make it non-conformant.

conformance/scripts/passing_txn_fixtures.txt
InKryption: 6d3edb156941430daf716d441575be7f39e1406b

src/runtime/transaction_execution.zig
Rexicon226: That could be the case. Perhaps there will be a way to make it conformant as well, but it's not worth pursuing atm.

Rexicon226 (dismissed):

yewman (approved):

yewman (approved):

Rexicon226 (approved):

PR #846, opened 2025-08-07
fix(txn): extract TransactionError values correctly

kprotty:
builds upon https://github.com/Syndica/sig/pull/836

### DIFF
--- a/conformance/scripts/fixtures.sh
+++ b/conformance/scripts/fixtures.sh
@@ -86,7 +86,7 @@ cd solana-conformance
 source test_suite_env/bin/activate
 
 export LD_PRELOAD=/lib/x86_64-linux-gnu/libasan.so.8
-export ASAN_OPTIONS=detect_leaks=0
+export ASAN_OPTIONS=detect_leaks=0 allocator_may_return_null=1
 
 echo "Generating fixtures"
 solana-test-suite create-fixtures \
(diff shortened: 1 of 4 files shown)


conformance/scripts/fixtures.sh
```diff
@@ -86,7 +86,7 @@ cd solana-conformance
 source test_suite_env/bin/activate
 
 export LD_PRELOAD=/lib/x86_64-linux-gnu/libasan.so.8
-export ASAN_OPTIONS=detect_leaks=0
+export ASAN_OPTIONS=detect_leaks=0 allocator_may_return_null=1
```
InKryption, line 89: I ran into the error that suggests this, do you have any idea/context on as to what it does?

conformance/src/txn_execute.zig
```diff
@@ -860,25 +862,32 @@ fn serializeOutput(
...
+                    const ixn_info = &sanitized.instruction_infos[converted.instruction_index];
+                    for (sig.runtime.program.precompiles.PRECOMPILES) |precompile| {
+                        if (precompile.program_id.equals(&ixn_info.program_meta.pubkey)) {
+                            converted.custom_error = 0;
+                            break;
+                        }
+                    }
+                }
+
+                break :blk converted;
```
InKryption: just a note, these changes are going to conflict with #840 which makes some similar fixes, except it appears to be missing this precompile check; may be helpful to coordinate on this cc @yewman.

conformance/scripts/fixtures.sh
kprotty: Given the amount of fixture tests running (now?), ASAN throws a hint that a large allocation may fail. I believe this silences that.

conformance/src/txn_execute.zig
kprotty: Thanks for heads up; https://github.com/Syndica/sig/pull/840 also fixes the other ASAN issue, so waiting for that to merge first.

kprotty:
Closed as resolved with https://github.com/Syndica/sig/pull/868

PR #850, opened 2025-08-08
fix(core:rent_collector): Make sure start of range is less than end of range

InKryption:
This was causing an overflow in some fixtures I was filtering through in order to come up with the list of newly passing fixtures in #842. It isn't required for the ones added in that PR, and I don't think this makes any fixtures to pass by itself, but it will at least help in getting more to pass.


InKryption:
Already done in #840.

PR #851, opened 2025-08-08
feat(replay,consensus): Implement cache tower stats

dadepo:


src/consensus/tower.zig
```diff
@@ -40,12 +40,20 @@ const ComputedBankState = struct {
...
-    },
+    failed_threshold: FailedThreshold,
+
+    /// NOTE: this is a tuple in the original rust code
+    pub const FailedThreshold = struct {
+        vote_depth: u64,
+        observed_stake: u64,
+    };
+
+    /// #[default]
```
dnut:
```suggestion
```

src/consensus/tower.zig
```diff
@@ -40,12 +40,20 @@ const ComputedBankState = struct {
...
+    pub const DEFAULT: ThresholdDecision = .passed_threshold;
```
dnut: Why is this the default? Can't you just use `.passed_threshold` directly when you want the "default"?

src/consensus/tower.zig
```diff
@@ -40,12 +40,20 @@ const ComputedBankState = struct {
...
     passed_threshold,
-    failed_threshold: struct {
-        // vote depth
-        u64,
-        // Observed stake
-        u64,
-    },
+    failed_threshold: FailedThreshold,
+
+    /// NOTE: this is a tuple in the original rust code
```
dnut: Is this comment necessary? I think it would get a bit cluttered in sig if we noted every single struct that we organized slightly differently than agave. Is there some special implication for this struct needing parity with agave's struct?

src/replay/consensus.zig
```diff
@@ -523,10 +526,43 @@ fn computeBankStats(
...
+
+    const slice = try replay_tower.checkVoteStakeThresholds(
+        allocator,
+        slot,
+        &stats.voted_stakes,
+        stats.total_stake,
+    );
+    defer allocator.free(slice);
+    stats.vote_threshold = try VoteThreshold.initCapacity(allocator, slice.len);
+    try stats.vote_threshold.appendSlice(allocator, slice);
```
dnut: unnecessary allocation

```suggestion
    stats.vote_threshold = VoteThreshold.fromOwnedSlice(slice);
```

you'll also need to change the return type of checkVoteStakeThresholds to a mut slice

src/consensus/tower.zig
dadepo: So this is existing code. See https://github.com/Syndica/sig/blob/4a962ccea6679388d37b68894b5d579ceaf7f8c5/src/consensus/progress_map.zig#L736-L751

It only shows here as a diff as it was moved to replace the duplicated one that used a tuple.

As to the rationale, we might have to defer to @InKryption as he added this version

src/consensus/tower.zig
dadepo: Same applies https://github.com/Syndica/sig/pull/851/files#r2275331731

src/consensus/tower.zig
dadepo: > Why is this the default?

Same applies https://github.com/Syndica/sig/pull/851/files#r2275331731

I can allude to the why, but @InKryption will be best to ask for choosing this style. This is only showing here as a new diff as it was moved to replace the duplicated version.

src/replay/consensus.zig
dadepo: [e4e6510](https://github.com/Syndica/sig/pull/851/commits/e4e651048f4b970664260248c2a98a78eca74ad2)

src/consensus/tower.zig
dnut: The presence of commented out rust code makes it clear that this has only undergone a first pass of "clone agave" and not gone through a second pass of careful thought about how it should actually be designed.

You're already making the change here to consolidate the type. I think this is the best opportunity we're going to get to clean it up.

My opinion is to remove both comments. I would also remove DEFAULT unless both of these are true:
- DEFAULT is currently in use in the code
- there's a clear reason why it's preferable to access this value through an obscure decl called `.DEFAULT` instead of just directly using the much clearer `.passed_threshold` value.

dadepo: > this has only undergone a first pass of "clone agave" 

I cross-checked again and indeed, the only reason I see for this style was it being a direct port of Rust's #[default]  

[792d00d](https://github.com/Syndica/sig/pull/851/commits/792d00dc3c8837344fc7fe7550983d66ba15273b)

src/consensus/tower.zig
dadepo: [792d00d](https://github.com/Syndica/sig/pull/851/commits/792d00dc3c8837344fc7fe7550983d66ba15273b)

src/consensus/tower.zig
dadepo: [792d00d](https://github.com/Syndica/sig/pull/851/commits/792d00dc3c8837344fc7fe7550983d66ba15273b)

src/replay/consensus.zig
```diff
@@ -32,6 +32,7 @@ const SlotTracker = sig.replay.trackers.SlotTracker;
 const EpochTracker = sig.replay.trackers.EpochTracker;
 
 pub const isSlotDuplicateConfirmed = sig.consensus.tower.isSlotDuplicateConfirmed;
+const VoteThreshold = sig.consensus.progress_map.ForkStats.VoteThreshold;
```
InKryption: Kind of odd to throw this import in between two exports.

src/replay/consensus.zig
```diff
@@ -523,10 +526,41 @@ fn computeBankStats(
...
 }
 
+fn cacheTowerStats(
+    allocator: std.mem.Allocator,
+    progress: *ProgressMap,
+    replay_tower: *const ReplayTower,
+    slot: Slot,
+    ancestors: *const std.AutoArrayHashMapUnmanaged(Slot, SortedSet(Slot)),
+) !void {
+    var stats = progress.getForkStats(slot) orelse return error.MissingSlot;
```
```suggestion
    const stats = progress.getForkStats(slot) orelse return error.MissingSlot;
```

src/replay/consensus.zig
```diff
@@ -523,10 +526,41 @@ fn computeBankStats(
...
+
+    const slice = try replay_tower.checkVoteStakeThresholds(
+        allocator,
+        slot,
+        &stats.voted_stakes,
+        stats.total_stake,
+    );
+    stats.vote_threshold = VoteThreshold.fromOwnedSlice(slice);
```
InKryption: Soft suggestion:
```suggestion
    stats.vote_threshold = .fromOwnedSlice(slice);
```

InKryption (changes requested):
Just a few comments, otherwise lgtm.

src/consensus/tower.zig
InKryption: Originally just ported over from agave, which specified `passed_threshold` as the default value. I have no attachment to it, it was simply convenient while reasoning about code that I was in the process of porting over.

src/replay/consensus.zig
dadepo: [78704ae](https://github.com/Syndica/sig/pull/851/commits/78704ae8dc992eec5ba9c09b844d04d6591026c7)

src/replay/consensus.zig
dadepo: [78704ae](https://github.com/Syndica/sig/pull/851/commits/78704ae8dc992eec5ba9c09b844d04d6591026c7)

src/replay/consensus.zig
dadepo: [78704ae](https://github.com/Syndica/sig/pull/851/commits/78704ae8dc992eec5ba9c09b844d04d6591026c7)

dnut (approved):

dnut (approved):

PR #853, opened 2025-08-11
feat(svm): add svm precompile entrypoints

yewman:
Passed: 3373, Failed: 542 (86%)

Adds entrypoint for precompile invocation within the svm. Requires instruction datas for whole transaction to be accessible from the transaction context. This PR takes a simple approach of adding a nullable instruction_datas slice to the transaction context which populated if a precompile is present and the move_precompiles_to_svm feature is active.

### DIFF
--- a/conformance/scripts/passing_txn_fixtures.txt
+++ b/conformance/scripts/passing_txn_fixtures.txt
@@ -16,6 +16,8 @@
 ./test-vectors/txn/fixtures/programs/019cd344312d03b3abd6414315b0ccf5dd722cad_265678.fix
 ./test-vectors/txn/fixtures/programs/01c457becc9df90ae0e14a274b1d803b343a9e86_265678.fix
 ./test-vectors/txn/fixtures/programs/01d7494b4d7dd492421e8267de2a5a70ab86ab22_553257.fix
+./test-vectors/txn/fixtures/programs/01eb407ac78a5ac2bd2a2d159f8aaa2476c23c71_2731315.fix
+./test-vectors/txn/fixtures/programs/02218de7ea8e0b2b392b334deea75d7920d284dd_1589325.fix
 ./test-vectors/txn/fixtures/programs/0233ae72e8c1120c3476503c8c698561e6bcda61_265678.fix
 ./test-vectors/txn/fixtures/programs/024d737de4d4972b4a1591e82ace975979dc6718_2230394.fix
 ./test-vectors/txn/fixtures/programs/02544cbab77086c1a3fd0d92db373b359b62c086_265678.fix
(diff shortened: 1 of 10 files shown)


src/runtime/program/lib.zig
```diff
@@ -42,5 +42,9 @@ fn initProgramEntrypoints() std.StaticStringMap(EntrypointFn) {
 }
 
 fn initPrecompileEntrypoints() std.StaticStringMap(EntrypointFn) {
-    return std.StaticStringMap(EntrypointFn).initComptime(&.{});
+    @setEvalBranchQuota(10_000);
+    return std.StaticStringMap(EntrypointFn).initComptime(&.{
+        .{ precompiles.ed25519.ID.base58String().slice(), precompiles.ed25519.execute },
+        .{ precompiles.secp256k1.ID.base58String().slice(), precompiles.secp256k1.execute },
```
kprotty, line 48: should this add TODO for `secp256r1` as well?

src/runtime/executor.zig
```diff
@@ -157,13 +157,16 @@ fn processNextInstruction(
     // Lookup native program function
     // [agave] https://github.com/anza-xyz/agave/blob/a705c76e5a4768cfc5d06284d4f6a77779b24c96/svm/src/message_processor.rs#L72-L75
     // [fd] https://github.com/firedancer-io/firedancer/blob/dfadb7d33683aa8711dfe837282ad0983d3173a0/src/flamenco/runtime/fd_executor.c#L1150-L1159
-    // TODO:
-    // - precompile feature gate move to svm otherwise noop
-    // - precompile entrypoints
+    const move_verify_precompiles_to_svm = ic.tc.feature_set.active(
```
kprotty, line 160: Is it from a newer validator version?

src/runtime/program/lib.zig
yewman: David is working on secp256r1 at the moment so I don't think we need a todo but happy to add one if you would prefer

src/runtime/executor.zig
yewman: Not sure what you mean? It is just a feature dependent branch that was not implemented yet

src/runtime/program/lib.zig
kprotty: Its fine - can be added in his PR

kprotty (dismissed):

conformance/scripts/passing_txn_fixtures.txt
```diff
@@ -725,7 +790,6 @@
 ./test-vectors/txn/fixtures/programs/399a3d6633306162cc72f3250b61271cfd988a46_2997779.fix
 ./test-vectors/txn/fixtures/programs/39a3c45f88ad951ed540132f929e020dce060f80_265678.fix
 ./test-vectors/txn/fixtures/programs/39b6d60bac630daf320a0801d99f8977a6688506_265678.fix
-./test-vectors/txn/fixtures/programs/39b8f771be5c314fa218177d6d0ceafe5c71667d_3203571.fix
```
dnut, line 728: Several fixtures are being removed from this file. Are they regressions?

src/runtime/transaction_execution.zig
```diff
@@ -509,6 +519,113 @@ pub fn executeTransaction(
     };
 }
 
+// TODO: RuntimeTransaction already contains this information which we should use in the future
+// instead of allocating a new array here.
+fn getInstructionDatasSliceForPrecompiles(
```
dnut, line 553: Why did you take this approach in the first place if the returned data is already present in the context that calls this function?

conformance/scripts/passing_txn_fixtures.txt
yewman: This is the result of a slightly incorrect implementation of reserved accounts. They are added back in the next PR #855 which fixes the reserved accounts implementation.

src/runtime/transaction_execution.zig
yewman: Execution of precompiles within the SVM requires the data of all instructions within the transaction, so we need to put this info in the transaction context if a precompile will be executed. I went with the above approach because to use the data in the form it currently exists in the calling context would require rewriting the precompile verify methods to use a slice of InstructionInfo's. A better approach might be to use a multi array list for the instruction infos in the runtime transaction which would allow us to directly use the slice of instruction datas, however, I felt that such a change fell outside the scope of this PR.

kprotty (approved):

PR #855, opened 2025-08-12
feat(runtime): implement reserved account keys with feature activations

yewman:
Passed: 3377, Failed: 538 (~86.2%)

### DIFF
--- a/src/core/bank.zig
+++ b/src/core/bank.zig
@@ -18,6 +18,7 @@ const std = @import("std");
 const sig = @import("../sig.zig");
 
 const core = sig.core;
+const reserved_accounts = sig.core.reserved_accounts;
 
 const Allocator = std.mem.Allocator;
 const Atomic = std.atomic.Value;
@@ -32,6 +33,7 @@ const HardForks = core.HardForks;
 const LtHash = core.hash.LtHash;
 const Pubkey = core.pubkey.Pubkey;
 const RentCollector = sig.core.rent_collector.RentCollector;
+const ReservedAccounts = sig.core.ReservedAccounts;
 
 const Epoch = core.time.Epoch;
 const Slot = core.time.Slot;
@@ -91,7 +93,13 @@ pub const SlotConstants = struct {
     ancestors: Ancestors,
 
     /// A map of activated features to the slot when they were activated.
+    /// NOTE: Feature activations are only applied at epoch boundaries, so should be constant
+    /// during an epoch, not just a slot. We should evaluate moving this and `reserved_accounts`
+    /// to `EpochConstants`.
     feature_set: FeatureSet,
+
+    /// A map of reserved accounts
+    reserved_accounts: ReservedAccounts,
 
     pub fn fromBankFields(
         allocator: Allocator,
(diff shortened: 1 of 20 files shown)


src/core/bank.zig
```diff
@@ -109,6 +117,11 @@ pub const SlotConstants = struct {
             .epoch_reward_status = .inactive,
             .ancestors = try bank_fields.ancestors.clone(allocator),
             .feature_set = feature_set,
+            .reserved_accounts = try reserved_accounts.initForSlot(
+                allocator,
+                &feature_set,
+                bank_fields.parent_slot + 1,
```
dnut: this is incorrect if the prior slot was skipped

```suggestion
                bank_fields.slot,
```

src/core/bank.zig
```diff
@@ -91,8 +93,14 @@ pub const SlotConstants = struct {
     ancestors: Ancestors,
 
     /// A map of activated features to the slot when they were activated.
+    /// NOTE: Feature activations are only applied at epoch boundaries, so should be constant
+    /// during an epoch, not just a slot. We should evaluate moving this and `reserved_accounts`
+    /// to `EpochConstants`.
     feature_set: FeatureSet,
 
+    /// A map of reserved accounts
```
dnut:
```suggestion
    /// A map of reserved accounts that are not allowed to acquire write locks
    /// in the current slot.
```

src/replay/service.zig
```diff
@@ -303,7 +303,6 @@ fn newSlotFromParent(
     try ancestors.ancestors.put(allocator, slot, {});
 
     var feature_set = try getActiveFeatures(allocator, account_reader.forSlot(&ancestors), slot);
-    errdefer feature_set.deinit(allocator);
```
dnut, line 306: Why did you remove this? Pretty sure it's necessary to avoid leaks on error.

src/replay/service.zig
```diff
@@ -321,6 +320,14 @@ fn newSlotFromParent(
         .epoch_reward_status = epoch_reward_status,
         .ancestors = ancestors,
         .feature_set = feature_set,
+        // This is inefficient, reserved accounts could live in epoch constants along with
+        // the feature set since feature activations are only applied at epoch boundaries.
+        // Then we only need to clone the map and update the reserved accounts once per epoch.
+        .reserved_accounts = try sig.core.reserved_accounts.initForSlot(
```
dnut: Typically I prefer to do these `try` operations outside the struct initialization, and include an errdefer for them (regardless of anything else fallible existing). When the next person comes and adds another fallible statement after this, you can't trust them to check every line of code in the function to make sure there are errdefers for everything.

src/replay/service.zig
yewman: Feature set is no longer allocated :)

src/core/bank.zig
yewman: I am genuinely baffled as to why I did this 😂 
91c634f0

src/core/bank.zig
yewman: 91c634f0

src/replay/service.zig
yewman: That is a good point!
91c634f0

dnut (approved):

PR #856, opened 2025-08-12
fix(harness): correct generation of feature list

Rexicon226:


src/core/features.zig
```diff
@@ -27,11 +29,26 @@ pub const Feature = @Type(.{ .@"enum" = .{
     .is_exhaustive = true,
 } });
 
-pub const map: std.EnumArray(Feature, Pubkey) = map: {
+const Info = struct {
+    key: Pubkey,
+    activated_on_all_clusters: bool,
```
dnut, line 34: Why did you add this back?

Rexicon226: In order to generate the feature list. We, following firedancer's lead, place the "always activated" features into the hardcoded set instead of the supported set.

dnut (approved):

PR #837, opened 2025-08-05
fix(replay): misc memory/threading issues

Sobeston:
Known issues:
- [X] accountsdb index write/free ordering 
- [ ] blockhashqueue (I suspect use after free?)
- [ ] account_locks <-> rent collector race

(probably won't fix in this PR, likely not harmful)
- [ ] sendRepairRequestBatch/Xoshiro256 shared between threads

misc fixes:
- parallelUntarToFileSystem will now use joinForDeinit (i.e. try to wait for the threads to finish), before deiniting the threadpool

fixes likely superceded by replay/commit fixes:
- ReplayState.deinit double-free

also had a fix in the accountsdb index, which was superceded by the index write/free ordering fix (i.e. no longer seems necessary)

PR #858, opened 2025-08-12
fix(accountsdb) putAccount: improve thread safety

Sobeston:
Cannibalised from #837

Now we only free old slot reference buffers *after* the reference to them has been replaced. 

This greatly reduces the likelihood of  thread safety problems when running replay

### DIFF
--- a/src/accountsdb/db.zig
+++ b/src/accountsdb/db.zig
@@ -2147,17 +2147,15 @@ pub const AccountsDB = struct {
 
         const slot_entry = try slot_ref_map.getOrPut(slot);
 
-        if (!slot_entry.found_existing) {
-            // no entry means realloc always needed, value set in realloc_needed block
-            slot_entry.value_ptr.* = .{
-                .refs = .empty,
-                .global_index = std.math.maxInt(u64), // u64-max == invalid value / replaced soon
-            };
-        }
-        const realloc_needed = slot_entry.value_ptr.refs.unusedCapacitySlice().len < pubkeys.len;
-        if (!slot_entry.found_existing) std.debug.assert(realloc_needed);
+        // no entry => realloc always needed
+        const realloc_needed = !slot_entry.found_existing or
+            slot_entry.value_ptr.refs.unusedCapacitySlice().len < pubkeys.len;
 
-        const old_refs = slot_entry.value_ptr.refs.items;
+        const old_refs = if (!slot_entry.found_existing)
+            &.{}
+        else
+            slot_entry.value_ptr.refs.items;
+
         const new_len = old_refs.len + pubkeys.len;
 
         if (realloc_needed) {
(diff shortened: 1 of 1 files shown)


src/accountsdb/db.zig
```diff
@@ -2147,17 +2147,15 @@ pub const AccountsDB = struct {
 
         const slot_entry = try slot_ref_map.getOrPut(slot);
 
-        if (!slot_entry.found_existing) {
-            // no entry means realloc always needed, value set in realloc_needed block
-            slot_entry.value_ptr.* = .{
-                .refs = .empty,
-                .global_index = std.math.maxInt(u64), // u64-max == invalid value / replaced soon
-            };
```
dnut, line 2155: Is it safe to remove this? what if this function returns with an error before setting the value?

src/accountsdb/db.zig
```diff
@@ -2147,17 +2147,15 @@ pub const AccountsDB = struct {
...
-        }
-        const realloc_needed = slot_entry.value_ptr.refs.unusedCapacitySlice().len < pubkeys.len;
-        if (!slot_entry.found_existing) std.debug.assert(realloc_needed);
+        // no entry => realloc always needed
+        const realloc_needed = !slot_entry.found_existing or
+            slot_entry.value_ptr.refs.unusedCapacitySlice().len < pubkeys.len;
```
dnut, line 2164: Is it possible for the second to be true and the first to be false? Why was it an assertion before, and why change it now to `or`?

src/accountsdb/db.zig
Sobeston: If the function errors before setting the value, the old refs will continue to be accessible+valid - it's safer to not have this

src/accountsdb/db.zig
Sobeston: If found_existing == false, the length would actually be undefined memory (as we've just used getOrPut). 

In other words if found_existing is false, or more capacity is required, we must (re)alloc. We only don't alloc when found_existing == true and we have enough capacity.

This change was a side-effect of getting rid of the premature mutation of value_ptr.

src/accountsdb/db.zig
dnut: What do you mean "the old refs"? The code I'm talking about is only run when there were no old refs.

In this case a new entry was just added to the index for this slot, which previously had no data for the slot. Now that index entry has undefined data. If there's an error, I think it would be safer to leave the data defined as `.empty` rather than leaving it undefined. Or you could have an errdefer to remove the entry from the index, but it's a bit harder to prove that this is correct.

Sobeston: Oh I misunderstood, I see the problem.

Opted to implement both guards - 01a49c82d487160020064041d0f4bc0c68a61da5. I think removing from the index in an errdefer makes the default-initialize redundant? But it's much easier to just have both than to think about it right now.

dnut: I think the default initialize is easier to reason about because what you're doing is making sure that the state is always correct at every step of execution.

I find the errdefer a little harder to reason about because it leaves the state incorrect for longer and then tries to roll it back if necessary. I think it would have the correct effect, but maybe it's less resistant to future refactoring bugs being introduced.

dnut (dismissed):

InKryption (dismissed):

dnut (approved):

PR #859, opened 2025-08-12
refactor(ledger): rename blockstore to ledger

dnut:
When I originally implemented the ledger, I left some confusing naming inconsistencies in there, where "blockstore" is used sometimes and "ledger" is used other times. I think we should consolidate on one name, and my preference is "ledger."


InKryption (approved):
long-time coming

yewman (approved):

yewman (approved):

PR #860, opened 2025-08-12
fix(tar) parallel untar: attempt to join before deinit

Sobeston:
Cannibalised from https://github.com/Syndica/sig/pull/837

Had this code fail badly a few times, this should allow for a cleaner deinit.

### DIFF
--- a/src/utils/tar.zig
+++ b/src/utils/tar.zig
@@ -79,9 +79,7 @@ pub fn parallelUntarToFileSystem(
     var pool =
         try HomogeneousThreadPool(UnTarTask).init(allocator, @intCast(n_threads), n_threads);
     defer {
-        if (!pool.joinForDeinit(.fromSecs(1))) {
-            logger.warn().log("failed to join for deinit");
-        }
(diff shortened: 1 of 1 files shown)


src/utils/tar.zig
```diff
@@ -78,8 +78,12 @@ pub fn parallelUntarToFileSystem(
 
     var pool =
         try HomogeneousThreadPool(UnTarTask).init(allocator, @intCast(n_threads), n_threads);
-    defer pool.deinit(allocator);
-
+    defer {
+        if (!pool.joinForDeinit(.{ .ns = std.time.ns_per_s })) {
```
dnut: There's a cleaner api for Duration

```suggestion
        if (!pool.joinForDeinit(.fromSecs(1))) {
```

dnut (approved):
I think this is one of those rare cases where the only reasonable option is to bypass the codecov check.

dnut (approved):

dnut:
"style" nice

Issue #791, opened 2025-06-21
chore(debugging): upgrade to tracy 0.12.1


Sobeston:
We're still on 0.11.1; 0.12 released a few weeks ago and has some great things in its [change log](https://github.com/wolfpld/tracy/releases), including "thread wakeup vizualization" which looks super useful.

This should be upgraded when we get the time, or the next time we're profiling

PR #861, opened 2025-08-12
feat(sig): tracy instrumentation (+ update to 0.12.2)

Sobeston:
It's been very useful for debugging replay. Closes https://github.com/Syndica/sig/issues/791.

- Instruments important parts of replay with tracy zones
- Upgrades tracy dependency from 0.11.1 to 0.12.2
- Adds a few extra useful ones (in e.g. accountsdb, threadpool)

A few replay slots:
image

Zoom in on some batches:
image

Replay dispatching a thread pool task:
image

A memory leak:
image

commands used in screenshots:
`zig build sig -Dno-run -Denable-tracy=true -Doptimize=ReleaseSafe`

`sudo bash -c "ulimit -n 262140 && ./zig-out/bin/sig validator -c testnet --use-disk-index -a 512 --skip-snapshot-validation --n-threads-snapshot-unpack 6 --n-threads-snapshot-load 6"`

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -50,8 +50,8 @@
             .hash = "secp256k1-0.5.0-2oLmGRAHAACjbp4BBLAoCU6Sotb_gqChbYOIIvBejN0V",
         },
         .tracy = .{
-            .url = "git+https://github.com/Rexicon226/zig-tracy/#a22e842419dd241423524f66ea371b935aac4893",
-            .hash = "zig_tracy-0.0.4-4TLLRx5xAADOcx4p3qeqpRn7WoMM6dgtwuHCvhmHDzyx",
+            .url = "git+https://github.com/Sobeston/zig-tracy#097d8496ab7ddaa5f554841a0e70ad67d2bd7e78",
+            .hash = "zig_tracy-0.12.2-4TLLRwhjAACLxeZ1j2agzhWoAZQTus3-REDfLxe9eI9I",
         },
     },
 }
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -50,8 +50,8 @@
             .hash = "secp256k1-0.5.0-2oLmGRAHAACjbp4BBLAoCU6Sotb_gqChbYOIIvBejN0V",
         },
         .tracy = .{
-            .url = "git+https://github.com/Sobeston/zig-tracy#6dc171d492307e133dab3ea25a51f49914cd43ec",
-            .hash = "zig_tracy-0.12.2-4TLLRwhjAACZ5oQoOU26JhHKbLLZoUmd-73M2UfDniTZ",
+            .url = "git+https://github.com/Syndica/tracy-zig#8fcc7521aa47e6dc8bf75de888f4ed4ed503c142",
+            .hash = "zig_tracy-0.12.2-4TLLRxpkAABhUEzSm2wXjioLd_-qruRviTnehDTbhZYR",
         },
     },
 }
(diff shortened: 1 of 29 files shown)


build.zig.zon
```diff
@@ -50,8 +50,8 @@
             .hash = "secp256k1-0.5.0-2oLmGRAHAACjbp4BBLAoCU6Sotb_gqChbYOIIvBejN0V",
         },
         .tracy = .{
-            .url = "git+https://github.com/Rexicon226/zig-tracy/#a22e842419dd241423524f66ea371b935aac4893",
-            .hash = "zig_tracy-0.0.4-4TLLRx5xAADOcx4p3qeqpRn7WoMM6dgtwuHCvhmHDzyx",
+            .url = "git+https://github.com/Sobeston/zig-tracy#6dc171d492307e133dab3ea25a51f49914cd43ec",
```
dnut: can we move this to Syndica? why is this forked anyway?

src/replay/trackers.zig
```diff
@@ -131,15 +138,16 @@ pub const SlotTracker = struct {
...
         try frozen_slots.ensureTotalCapacity(allocator, self.slots.count());
+
         for (self.slots.keys(), self.slots.values()) |slot, value| {
-            if (value.state.isFrozen()) {
-                frozen_slots.putAssumeCapacity(slot, .{
-                    .constants = &value.constants,
-                    .state = &value.state,
-                });
-            }
+            if (!value.state.isFrozen()) continue;
```
dnut, line 145: personally I prefer the old way but I'm fine with this too

build.zig.zon
Sobeston: I don't see why it couldn't be a syndica repo, it just hasn't been done.

It's forked because the original isn't very complete, and doesn't do quite what I needed. Seems like a better use of time to keep going with it than to merge it back in.

src/replay/trackers.zig
Sobeston: I prefer this way but obviously it doesn't matter, this diff just snuck in when I was making a new branch. Maybe we flip a coin

dnut: I say just leave it as is

src/replay/confirm_slot.zig
```diff
@@ -52,7 +53,13 @@ pub fn confirmSlot(
     verify_ticks_params: VerifyTicksParams,
     ancestors: *const Ancestors,
 ) !*ConfirmSlotFuture {
+    var zone = tracy.Zone.init(@src(), .{ .name = "confirmSlot" });
```
Rexicon226, line 56: Sexy.

Rexicon226 (dismissed):
overall looks good

src/runtime/check_transactions.zig
```diff
@@ -98,6 +99,9 @@ pub fn checkFeePayer(
     FeeDetails,
     TransactionRollbacks,
 }) {
+    var zone = tracy.Zone.init(@src(), .{ .name = "checkFeePayer" });
```
InKryption, line 102: This is kind of a general question: why is it that we have to specify `name` everywhere, couldn't it get that information from the `@src()`?

Sobeston: That's true, I suppose we could. I went with this way because it allowed me to choose when to include more context versus not, e.g. "verifyPoh" is fine as is (we know which function this is), but "clone" isn't (we need to append some context from the type or file).

InKryption (dismissed):

build.zig.zon
Sobeston: It's in https://github.com/Syndica/zig-tracy
Made it as private, and don't have perms to make it public. Oops. Waiting for someone to make it public

Sobeston: It's now https://github.com/Syndica/tracy-zig

dnut (approved):

InKryption (approved):

PR #863, opened 2025-08-13
fix(svm): fix regressions and activate more CI conformance checks

yewman:


src/core/features.zig
```diff
@@ -57,23 +57,21 @@ pub const map: std.EnumArray(Feature, Info) = map: {
 ///
 /// [agave] https://github.com/anza-xyz/agave/blob/8db563d3bba4d03edf0eb2737fba87f394c32b64/sdk/feature-set/src/lib.rs#L1188
 pub const Set = struct {
-    array: std.EnumArray(Feature, Slot),
+    array: std.EnumArray(Feature, ?Slot),
```
dnut, line 60: I was thinking about doing the same thing. It uses more memory but I think it's worth it for clarity/safety.

dnut (approved):

src/core/features.zig
yewman: I agree, it is also necessitated by the fact that some test fixtures does check feature activation for slot = max u64

dnut: wow interesting. you could pick a random number instead as the terminator. but still I prefer nullable

dnut (approved):

kprotty (approved):

PR #869, opened 2025-08-14
fix(programs): fix account borrow in zk elgamal

yewman:
+15 fixtures


Rexicon226 (approved):
lgtm. i did this for the close context state instruction, but forgot for the verification ones.

PR #870, opened 2025-08-15
refactor(runtime): Change ExecutedTransaction to store TransactionError

yewman:


src/runtime/transaction_execution.zig
```diff
@@ -188,7 +189,7 @@ pub const CopiedAccount = struct {
 };
 
 pub const ExecutedTransaction = struct {
-    instr_err: ?struct { u8, InstructionErrorEnum },
+    err: ?TransactionError,
```
dnut, line 193: This is a lot broader than InstructionError. Is this just to allow returning the one rent error?

I think it would be nice to tighten up some of these error sets to get more clarity about what could be returned by different areas of the code. I'm not sure of the cleanest way to do that though.

src/runtime/transaction_execution.zig
```diff
@@ -548,6 +573,60 @@ pub fn executeTransaction(
...
+        const rent_state = if (txn.accounts.items(.is_writable)[i]) blk: {
+            const account = tc.borrowAccountAtIndex(@intCast(i), .{
+                .program_id = Pubkey.ZEROES,
+                .remove_accounts_executable_flag_checks = false,
+            }) catch @panic("Account must exist in transaction context");
+            defer account.release();
+
+            if (sig.runtime.ids.NATIVE_LOADER_ID.equals(&account.account.owner)) {
+                // TODO: Native programs should not be writable. Returning null here is correct
+                // with respect to this function. However, we need to fix the is writable bug
```
dnut, line 622: do we still have a bug with account writability checks?

yewman: Yeah I seems that way. I didn't go down the road of trying to fix it in this PR because it makes sense to implement that fix in a follow up. A bunch of the remaining failing fixtures are failing simply because we include an account in the result state which we should not (i.e. we think its writable when it is not)

src/runtime/transaction_execution.zig
yewman: I had a similar thought but decided that would be a nice to have once we have conformance sorted. There is a possibility that other transaction errors may need to be returned to pass currently failing fixtures (or after upggrade). Even though I think this is unlikely I would prefer to leave it as is for now and tighten it up once we are conforming with the most recent release.

PR #875, opened 2025-08-18
feat(replay, consensus): Add replay vote sender

dadepo:
This PR incorporates the replay vote sender, which is used to send vote transaction in replay to vote listener. 

It essentially those the following:

- Set up and start the vote listener in cmd.zig
- Set up `*sig.sync.Channel(ParsedVote)` in cmd.zig and passes to both the vote_listeners (as receiver) and replay (as the sender)
- Add ability to parse ResolvedTransaction to ParsedVote
- In replay, after `processBatch` finds the vote transaction and sends it through `*sig.sync.Channel(ParsedVote)` to the vote listener

The code to processed the parsed votes already exist in the `listenAndConfirmVotes` method in the vote listener.


src/replay/service.zig
```diff
@@ -120,6 +129,9 @@ pub const Receivers = struct {
     ///       - [relevant interface invokation](https://github.com/anza-xyz/agave/blob/0315eb6adc87229654159448344972cbe484d0c7/gossip/src/duplicate_shred_listener.rs#L31)
     duplicate_slots: *sig.sync.Channel(Slot),
 
+    // Borrowed. Managed by Senders.
+    replay_votes: *sig.sync.Channel(ParsedVote),
```
dnut: I think the purpose of this struct is to define the API that replay has with other components. If this channel is used totally internally to replay, and I don't think it makes sense require it as an input to replay by placing it in this struct. Replay should instantiate it and manage it internally.

src/replay/scheduler.zig
```diff
@@ -262,6 +279,26 @@ const ProcessBatchTask = struct {
             self.exit,
         );
 
+        // On success, find and send vote transactions immediately
+        if (result == .success) {
```
dnut: Can you move all the code you've added here into the processBatch function?

ProcessBatchTask is only suppoed to be a thin wrapper around processBatch. The sole purpose of this struct is to manage the invocation of that function in a thread pool. All of the existing code here is just to satisfy the constraints of running in a thread pool. The logic for processing a batch should not be coupled to the requirements for running in a thread pool. It should be easy to use outside the constraints of the thread pool.

In the future I think we can probably generify structs like this and eliminate them, and then we'll only need to define simple functions like processBatch and easily run them in whatever multithreaded, asynchronous, or simple single threaded context that we like, without needing any extra baggage. 


-----

Actually, why didn't you put this in the committer? Doesn't agave send the votes over as soon as each transaction result is verified, and not necessarily after the entire batch is a success?

src/replay/service.zig
```diff
@@ -404,6 +419,38 @@ pub fn run(deps: ReplayDependencies) !void {
     var state = try ReplayState.init(deps);
     defer state.deinit();
 
+    // Start vote listener inside replay so it can reference replay state directly.
+    var vote_tracker: sig.consensus.VoteTracker = .EMPTY;
+    defer vote_tracker.deinit(deps.allocator);
+
+    var vote_listener_bank_forks_rw = sig.sync.RwMux(sig.consensus.vote_listener.BankForksStub)
```
dnut: this RwMux is not accomplishing its intended goal because it's only used in one thread. This struct should not be wrapped, rather each of the slot tracker and epoch tracker need to be wrapped in a RwMux because they are now used in multiple threads.

Actually the lifetimes of the pointers to the data inside those structs is going to be complicated to manage properly. I think you'll need to add reference counting in a bunch of places. It might actually prove easier to flatten out the vote listener threads here into replay, which I think would also be a cleaner design.

The way I would do that is by changing VoteListener to contain the state used by the vote listener, instead of passing the state into separate threads and containing thread handles. Then I would define a new method `VoteListener.advance` that's analogous to `advanceReplay`. It can call the `recvAndSendOnce` followed by `processVotesOnce`. Any extra state needed by those functions that is currently instantiated in `recvLoop` or `processVotesLoop` can be added to the vote listener struct.

Regardless of the approach you take, it's going to require significant changes to make the vote listener thread-safe alongside replay.

src/replay/confirm_slot.zig
```diff
@@ -200,6 +200,7 @@ pub const ConfirmSlotFuture = struct {
             thread_pool,
             svm_params,
             &future.exit,
+            null,
```
dnut: shouldn't this be the actual channel?

src/replay/service.zig
dadepo: [fe6f07d](https://github.com/Syndica/sig/pull/875/commits/fe6f07dc139a710290249a2625c46f2ee80574cd)

src/replay/confirm_slot.zig
dadepo: [adbb13e](https://github.com/Syndica/sig/pull/875/commits/adbb13e73bddd2b48596ee1ad7a5a206195f48ba)

src/replay/scheduler.zig
dadepo: [adbb13e](https://github.com/Syndica/sig/pull/875/commits/adbb13e73bddd2b48596ee1ad7a5a206195f48ba)

> Actually, why didn't you put this in the committer?

This is related to a perceived improvement: ie do not send the vote if the batch processing fails. But that might be unnecessary. Update to match Agave's behavior.

src/replay/service.zig
dadepo: adbb13e

src/replay/service.zig
```diff
@@ -183,19 +194,29 @@ const ReplayState = struct {
...
+        {
+            const ptr, var lg = self.slot_tracker_rw.writeWithLock();
+            defer lg.unlock();
+            ptr.deinit(self.allocator);
+        }
+        {
+            const ptr, var lg = self.epoch_tracker_rw.writeWithLock();
+            defer lg.unlock();
+            ptr.deinit(self.allocator);
+        }
```
dnut:
```suggestion
        var slots = self.slot_tracker_rw.tryRead() orelse @panic("slot tracker deinit while in use");
        var epochs = self.epochs_tracker_rw.tryRead() orelse @panic("epoch tracker deinit while in use");
        slots.get().deinit(self.allocator);
        epochs.get().deinit(self.allocator);
```

src/replay/service.zig
```diff
@@ -183,19 +194,29 @@ const ReplayState = struct {
 
     execution_log_helper: replay.execution.LogHelper,
 
+    replay_votes_ch: *sig.sync.Channel(ParsedVote),
```
dnut:
```suggestion
    replay_votes_channel: *sig.sync.Channel(ParsedVote),
```

src/replay/service.zig
```diff
@@ -208,6 +229,8 @@ const ReplayState = struct {
         });
         errdefer slot_tracker.deinit(deps.allocator);
 
+        const replay_votes_ch = try sig.sync.Channel(ParsedVote).create(deps.allocator);
```
dnut:
```suggestion
        const replay_votes_ch = try sig.sync.Channel(ParsedVote).create(deps.allocator);
        errdefer replay_votes.destroy();
```

src/replay/service.zig
```diff
@@ -409,8 +435,49 @@ pub fn run(deps: ReplayDependencies) !void {
...
+    // This defer is defined first, so it runs last. This ensures that any threads
+    // borrowing `state.slot_tracker_rw`/`state.epoch_tracker_rw` are joined before the RwMuxes are deinitialized.
     defer state.deinit();
 
+    // Start vote listener inside replay so it can reference replay state directly.
+    var vote_tracker: sig.consensus.VoteTracker = .EMPTY;
+    defer vote_tracker.deinit(deps.allocator);
+
+    // TODO: Should `slot_data_provider` itself be wrapped in RwMux to ensure consistency
+    // of the struct itself across multiple threads?
```
dnut: I don't see why that would be necessary if the fields are protected already. These pointers themselves should never be written again after this.

src/replay/execution.zig
```diff
@@ -194,8 +200,14 @@ fn replaySlot(state: ReplayExecutionState, slot: Slot) !ReplaySlotStatus {
         return .dead;
     }
 
-    const epoch_info = state.epochs.getForSlot(slot) orelse return error.MissingEpoch;
-    const slot_info = state.slot_tracker.get(slot) orelse return error.MissingSlot;
+    const epoch_info_ptr: *const sig.core.EpochConstants = blk: {
+        const ep, var ep_lg = state.epochs.readWithLock();
+        defer ep_lg.unlock();
+        break :blk ep.getPtrForSlot(slot) orelse return error.MissingEpoch;
+    };
```
dnut: The pointer lifetime exceeds the duration of the lock which violates the intended protection of the lock.

src/replay/edge_cases.zig
```diff
@@ -51,6 +51,11 @@ pub fn processEdgeCases(
     // have the wrong version. Our version was dead or pruned.
     // Signalled by ancestor_hashes_service.
     timer.reset();
+    // TODO: should we be holding the lock here? or pass the RwMux to the individual functions
+    // and have the locks be aquired at the site where they are accessed?
```
dnut: Personally I would have just acquired the locks in the replay service's main orchestration function, and passed down const pointers to execution, consensus, and edge cases code. That keeps things less encumbered by synchronization code. You don't need to be judicious with read locks at all when you're in the same thread that is also the only thread that acquires write locks.

src/replay/scheduler.zig
dnut: > Update to match Agave's behavior.

what do you mean here? agave's behavior is more like putting it in `commitTransactions`, no?

src/replay/service.zig
dadepo: [e91dbf1](https://github.com/Syndica/sig/pull/875/commits/e91dbf1bf38f77afee48f9d1014c7324bc3919c5)

src/replay/service.zig
dadepo: [e91dbf1](https://github.com/Syndica/sig/pull/875/commits/e91dbf1bf38f77afee48f9d1014c7324bc3919c5)

src/replay/service.zig
dadepo: [e91dbf1](https://github.com/Syndica/sig/pull/875/commits/e91dbf1bf38f77afee48f9d1014c7324bc3919c5)

src/replay/service.zig
dadepo: [9a59447](https://github.com/Syndica/sig/pull/875/commits/9a5944765f8f9e86c37c24f740bae7b387603ff3)

src/replay/execution.zig
dadepo: Updated to use the lock to create a clone of the epoch constants [618d05f](https://github.com/Syndica/sig/pull/875/commits/618d05f4f3c5091b4bc4dbfdff485d3dd8a82ff2)

Given the epoch constants should not change for a given slot/epoch I think this is semantically correct. Although I am not sure if there might be a better way to do this (one that avoids a clone for example)

dnut: I would have just held the lock for more time, but this works too.

src/replay/scheduler.zig
dadepo: > what do you mean here?

Was referring to Agave's behavior of sending the vote transactions regardless if all transactions in the batch is successful. 

> agave's behavior is more like putting it in commitTransactions, no?

Correct. To further be in parity with Agave, I have moved the sending to `commitTransactions` [94bce38](https://github.com/Syndica/sig/pull/875/commits/94bce38e96d76328773ed1aa491491390e44de32)

The testing revealed a leak whose fix was updated here [8f275fe](https://github.com/Syndica/sig/pull/875/commits/8f275fe2270d298a25e4f771c15155969748d5b7)

src/replay/edge_cases.zig
dadepo: [8c66e61](https://github.com/Syndica/sig/pull/875/commits/8c66e61293c912f1cd449b28a8c2b523173f5a19)

src/replay/execution.zig
dadepo: 8c66e61

src/replay/commit.zig
```diff
@@ -89,6 +102,38 @@ pub const Committer = struct {
             );
             try self.account_store.put(slot, pubkey, account);
         }
+
+        for (transactions, tx_results) |transaction, result| {
+            _, const tx_result = result;
+
+            switch (tx_result) {
+                .executed => |exec| {
```
dnut: the current function already has this code. see line 79. we don't need another loop and switch

src/consensus/vote_listener.zig
```diff
@@ -1738,8 +1738,9 @@ test "simple usage" {
         gossip_table.deinit();
     }
 
-    const replay_votes_channel: *sig.sync.Channel(vote_parser.ParsedVote) = try .create(allocator);
-    defer replay_votes_channel.destroy();
+    const replay_votes_channelannel: *sig.sync.Channel(vote_parser.ParsedVote) =
```
dnut: channelannel

src/consensus/vote_listener.zig
```diff
@@ -1879,8 +1880,8 @@ test "check trackers" {
     var vote_tracker: VoteTracker = .EMPTY;
     defer vote_tracker.deinit(allocator);
 
-    const replay_votes_channel = try sig.sync.Channel(vote_parser.ParsedVote).create(allocator);
-    defer replay_votes_channel.destroy();
+    const replay_votes_channelannel = try sig.sync.Channel(vote_parser.ParsedVote).create(allocator);
```
dnut: channelannel

dadepo: [ec96cb0](https://github.com/Syndica/sig/pull/875/commits/ec96cb03833affd297b98c0530686ce6747ac3d1)

src/consensus/vote_listener.zig
dadepo: [ec96cb0](https://github.com/Syndica/sig/pull/875/commits/ec96cb03833affd297b98c0530686ce6747ac3d1)

src/replay/commit.zig
dadepo: I actually saw that. But the rationale I had for having it much lower was to place the sending of the replay votes after all the updates to things like `account_store`, `stakes_cache` etc as I interpreted that to mean,  committing can only be considered successful after these updates. - but I may be wrong.

Updated [a9270fb](https://github.com/Syndica/sig/pull/875/commits/a9270fb1e45d67ea8581bf37366b304d6d4a0242)

dnut: The `continue` statement you added breaks the semantics of the loop. The loop has other code that needs to execute (inserting into the status cache), regardless of whether this is a vote transaction.

I would negate the condition and nest the sending logic in the if-statement, unless you can think of a better way to retain the correct semantics.

dadepo: Should be fixed now [b5d3d6f](https://github.com/Syndica/sig/pull/875/commits/b5d3d6fb91780e6d856d34edaaa8f446152a3d60)

src/cmd.zig
```diff
@@ -1196,14 +1196,20 @@ fn validator(
     const replay_senders: sig.replay.service.Senders = try .create(allocator);
     defer replay_senders.destroy();
 
-    const replay_receivers: sig.replay.service.Receivers = try .create(allocator);
+    const replay_receivers: sig.replay.service.Receivers = try .create(
+        allocator,
+    );
```
InKryption: Soft suggestion, feel free to ignore:
```suggestion
    const replay_receivers: sig.replay.service.Receivers =
        try .create(allocator);
```

src/cmd.zig
```diff
@@ -1196,14 +1196,20 @@ fn validator(
...
     const replay_thread = replay: {
         const epoch_stakes_map = &collapsed_manifest.bank_extra.versioned_epoch_stakes;
         const epoch_stakes = epoch_stakes_map.get(epoch) orelse
             return error.EpochStakesMissingFromSnapshot;
-
+        const current_epoch_constants = try sig.core.EpochConstants.fromBankFields(
+            bank_fields,
+            try epoch_stakes.current.convert(allocator, .delegation),
+        );
+        errdefer current_epoch_constants.deinit(allocator);
```
InKryption: I'm assuming this `convert` allocates? If yes, the conversion should be separated out into its own statement with an appropriate `errdefer deinit` so that the resources are freed in the event that `try fromBankFields` fails.

src/consensus/vote_listener.zig
```diff
@@ -19,90 +19,91 @@ const EpochTracker = sig.replay.trackers.EpochTracker;
...
+    epoch_tracker_rw: *sig.sync.RwMux(EpochTracker),
 
     fn rootSlot(self: *const SlotDataProvider) Slot {
-        return self.slot_tracker.root;
+        const root = blk: {
+            const slot_tracker, var lg = self.slot_tracker_rw.readWithLock();
+            defer lg.unlock();
+            break :blk slot_tracker.root;
+        };
+        return root;
```
InKryption: Simplification I'd suggest for all the other cases as well:
```suggestion
        const slot_tracker, var lg = self.slot_tracker_rw.readWithLock();
        defer lg.unlock();
        return slot_tracker.root;
```

src/consensus/vote_listener.zig
```diff
@@ -1264,6 +1299,163 @@ const vote_parser = struct {
...
+            var seen = std.bit_set.ArrayBitSet(usize, 256).initEmpty();
+            for (first_ix.account_indexes, 0..) |acct_index_u8, i| {
+                const acct_index: usize = acct_index_u8;
+                const index_in_callee: usize = if (seen.isSet(acct_index)) blk: {
+                    var prior_idx: usize = 0;
+                    while (prior_idx < i) : (prior_idx += 1) {
+                        if (first_ix.account_indexes[prior_idx] == acct_index_u8)
+                            break :blk prior_idx;
+                    }
+                    break :blk i;
```
InKryption: Correct me if I'm missing something, but I think this would be simpler:
```suggestion
                    break :blk std.mem.indexOfScalar(u8, first_ix.account_indexes[0..i], acct_index_u8) orelse i;
```

src/consensus/vote_listener.zig
```diff
@@ -1264,6 +1299,163 @@ const vote_parser = struct {
...
+
+                const pubkey = message.account_keys[acct_index];
+                (account_metas.addOne() catch break).* = .{
+                    .pubkey = pubkey,
+                    .index_in_transaction = @intCast(acct_index),
+                    .index_in_caller = @intCast(acct_index),
+                    .index_in_callee = @intCast(index_in_callee),
+                    .is_signer = message.isSigner(acct_index),
+                    .is_writable = false,
+                };
```
InKryption: Is ignoring overflow of account keys correct here? Not a test failure? A comment would be good.
Also, this looks a bit awkward, I would suggest:
```suggestion
                account_metas.append(.{
                    .pubkey = pubkey,
                    .index_in_transaction = @intCast(acct_index),
                    .index_in_caller = @intCast(acct_index),
                    .index_in_callee = @intCast(index_in_callee),
                    .is_signer = message.isSigner(acct_index),
                    .is_writable = false,
                }) catch break;
```

src/consensus/vote_listener.zig
```diff
@@ -1264,6 +1299,163 @@ const vote_parser = struct {
...
+                .account_metas = account_metas,
+                .instruction_data = first_ix.data,
+            };
+
+            var resolved: sig.replay.resolve_lookup.ResolvedTransaction = .{
+                .transaction = vote_tx,
+                .accounts = .{},
+                .instructions = instructions,
+            };
+            defer resolved.deinit(allocator);
```
InKryption: Pretty sure this would double-free `vote_tx`, surprised if it's not causing a failure in some way. Suggestion:
```suggestion
            const: sig.replay.resolve_lookup.ResolvedTransaction = .{
                .transaction = vote_tx,
                .accounts = .{},
                .instructions = &.{.{
                    .program_meta = .{
                        .pubkey = message.account_keys[first_ix.program_index],
                        .index_in_transaction = first_ix.program_index,
                    },
                    .account_metas = account_metas,
                    .instruction_data = first_ix.data,
                }},
            };
```

src/consensus/vote_listener.zig
```diff
@@ -1264,6 +1299,163 @@ const vote_parser = struct {
...
+                    .hash = bank_hash,
+                    .timestamp = null,
+                } },
+                .switch_proof_hash = input_hash,
+                .signature = vote_tx.signatures[0],
+            }, maybe_parsed_tx);
+        }
+
+        // Test bad program id fails
+        var vote_ix = try vote_instruction.createVote(
```
InKryption:
```suggestion
        const vote_ix = try vote_instruction.createVote(
```

src/consensus/vote_listener.zig
```diff
@@ -1264,6 +1299,163 @@ const vote_parser = struct {
...
+        if (first_ix.account_indexes.len != 0) {
+            const acct_index: usize = first_ix.account_indexes[0];
+            (try account_metas.addOne()).* = .{
+                .pubkey = message.account_keys[acct_index],
+                .index_in_transaction = @intCast(acct_index),
+                .index_in_caller = @intCast(acct_index),
+                .index_in_callee = 0,
+                .is_signer = message.isSigner(acct_index),
+                .is_writable = false,
+            };
```
InKryption: Same suggestion about just using `append` instead.

src/consensus/vote_listener.zig
```diff
@@ -1264,6 +1299,163 @@ const vote_parser = struct {
...
+            .account_metas = account_metas,
+            .instruction_data = first_ix.data,
+        };
+
+        var resolved_bad: sig.replay.resolve_lookup.ResolvedTransaction = .{
+            .transaction = vote_tx,
+            .accounts = .{},
+            .instructions = instructions,
+        };
+        defer resolved_bad.deinit(allocator);
```
InKryption: Same suggestion as on the other similar case.

src/replay/commit.zig
```diff
@@ -63,7 +76,24 @@ pub const Committer = struct {
...
+                    rent_collected += exec.loaded_accounts.rent_collected;
+                    // Skip non successful or non vote transactions.
+                    if (exec.executed_transaction.err == null and
+                        isSimpleVoteTransaction(transaction.transaction))
+                    {
+                        if (vote_listener.vote_parser.parseSanitizedVoteTransaction(
+                            allocator,
+                            transaction,
+                        ) catch null) |parsed| {
```
InKryption: Why are we treating this OOM as null? It should just propagate with `try` like all other instances (like the calls to `putAccount`).

src/replay/confirm_slot.zig
```diff
@@ -2,6 +2,7 @@ const std = @import("std");
 const sig = @import("../sig.zig");
 const replay = @import("lib.zig");
 const tracy = @import("tracy");
+const vote_listener = @import("../consensus/vote_listener.zig");
```
InKryption: I don't personally have a strong opinion on this matter, but I think the current consensus on imports like this is to alias them relative to `sig`, in this case `sig.consensus.vote_listener`.

src/replay/execution.zig
```diff
@@ -2,6 +2,7 @@ const std = @import("std");
 const sig = @import("../sig.zig");
 const replay = @import("lib.zig");
 const tracy = @import("tracy");
+const vote_listener = @import("../consensus/vote_listener.zig");
```
InKryption: Same comment as on other vote listener import.

src/replay/scheduler.zig
```diff
@@ -2,6 +2,7 @@ const std = @import("std");
 const sig = @import("../sig.zig");
 const replay = @import("lib.zig");
 const tracy = @import("tracy");
+const vote_listener = @import("../consensus/vote_listener.zig");
```
InKryption: Same comment as on other vote listener import.

InKryption (changes requested):

src/replay/scheduler.zig
dadepo: [8055677](https://github.com/Syndica/sig/pull/875/commits/8055677251a5c40de3fb83c40ff31149c696195e)

src/replay/execution.zig
dadepo: [8055677](https://github.com/Syndica/sig/pull/875/commits/8055677251a5c40de3fb83c40ff31149c696195e)

src/replay/confirm_slot.zig
dadepo: [8055677](https://github.com/Syndica/sig/pull/875/commits/8055677251a5c40de3fb83c40ff31149c696195e)

src/cmd.zig
dadepo: [e4adfad](https://github.com/Syndica/sig/pull/875/commits/e4adfad1fb2c0736ee8be01182928c7e5a1079e8)

src/consensus/vote_listener.zig
dadepo: [e4adfad](https://github.com/Syndica/sig/pull/875/commits/e4adfad1fb2c0736ee8be01182928c7e5a1079e8)

src/cmd.zig
dadepo: [435d652](https://github.com/Syndica/sig/pull/875/commits/435d652a3601180d1d97d3a98358e45c49483bfe)

src/consensus/vote_listener.zig
dadepo: [e12f0e6](https://github.com/Syndica/sig/pull/875/commits/e12f0e63c6bac30a1fa134f8d00d1b6720f7d099)

src/consensus/vote_listener.zig
dadepo: [e12f0e6](https://github.com/Syndica/sig/pull/875/commits/e12f0e63c6bac30a1fa134f8d00d1b6720f7d099)

src/consensus/vote_listener.zig
dadepo: [e12f0e6](https://github.com/Syndica/sig/pull/875/commits/e12f0e63c6bac30a1fa134f8d00d1b6720f7d099)

src/consensus/vote_listener.zig
dadepo: [e12f0e6](https://github.com/Syndica/sig/pull/875/commits/e12f0e63c6bac30a1fa134f8d00d1b6720f7d099)

src/consensus/vote_listener.zig
dadepo: [e12f0e6](https://github.com/Syndica/sig/pull/875/commits/e12f0e63c6bac30a1fa134f8d00d1b6720f7d099)

src/consensus/vote_listener.zig
dadepo: [e12f0e6](https://github.com/Syndica/sig/pull/875/commits/e12f0e63c6bac30a1fa134f8d00d1b6720f7d099)

src/consensus/vote_listener.zig
InKryption: I think a panic might be more suitable than unreachable there, so that we hit that even if testing in release mode.

src/consensus/vote_listener.zig
InKryption: The `errdefer` should be removed, that will just make it double free on encountering an error; all the resources this instruction is initialized with already have their deinitialization set up separately beforehand.

src/replay/commit.zig
dadepo: I think it boils down to the question: if an attempt to send out a vote tx down to the vote listener fails, should the remaining operations in the commit path be aborted? The current stance is, the vote tx should just be skipped, allowing the rest of the commit processing to continue.

src/consensus/vote_listener.zig
InKryption: If you remove the `dupe`, you can also remove the `defer deinit` (same rationale as in the other case).

src/replay/commit.zig
InKryption: That isn't reflected in the other parts of this function which simply propagate `try` as a normal error, which seems like it should be the default stance, unless we're going to be handling OOM in some specific way. Parsing errors we can skip, but if there's an OOM, we want to fail, otherwise we're just deferring the error to a future allocation, which will be further removed from the initial actual OOM.

src/consensus/vote_listener.zig
dadepo: [b0890fa](https://github.com/Syndica/sig/pull/875/commits/b0890fa5c32d51f31eb62dd8f280439f636a0d30)

src/consensus/vote_listener.zig
dadepo: [b0890fa](https://github.com/Syndica/sig/pull/875/commits/b0890fa5c32d51f31eb62dd8f280439f636a0d30)

src/consensus/vote_listener.zig
dadepo: [b0890fa](https://github.com/Syndica/sig/pull/875/commits/b0890fa5c32d51f31eb62dd8f280439f636a0d30)

src/replay/commit.zig
dadepo: > That isn't reflected in the other parts of this function which simply propagate try as a normal error, which seems like it should be the default stance

So the way I currently view the "parse vote, send it to vote listener" is that it is sort of auxiliary. 

On the other hand, the main function of  `commitTransaction` is to commit, and yes any error in that code path is propagated as it should. The sending votes through the channel is sort of supplementary that is now being inserted into that commit flow, and not sure if an error in that path should affect the committing. I initially had the sending of the votes as the last thing in the function mainly due to this.

But the argument can be made that an OOM is a signal that there is something fatally wrong, and even if this happens in the "parse vote, send it to vote listener" path, it should just be propagated instead of swallowed as nothing can be expected to work properly hence forth - and I guess this is the crux of your argument. 

Updated [ada7906](https://github.com/Syndica/sig/pull/875/commits/ada7906002e0e04350af48cdbc9ca61ce5086e4e)

An improvement that can be made to cleanly separate sending the votes tx from the committing is to move it out but immediately after `commitTransaction` - but I have moved this a couple of times already in this PR, I am fine leaving it as is.

src/replay/commit.zig
```diff
@@ -31,6 +45,7 @@ pub const Committer = struct {
         slot: Slot,
         transactions: []const ResolvedTransaction,
         tx_results: []const struct { Hash, ProcessedTransaction },
+        replay_votes_sender: *Channel(ParsedVote),
```
dnut: I think this makes more sense as a struct field.

src/replay/confirm_slot.zig
```diff
@@ -68,7 +73,16 @@ pub fn confirmSlot(
             allocator.free(entries);
         }
         break :fut try ConfirmSlotFuture
-            .create(allocator, logger, thread_pool, committer, entries, svm_params, slot_resolver);
+            .create(
```
dnut:
```suggestion
        break :fut try ConfirmSlotFuture.create(
```

InKryption (approved):

src/replay/confirm_slot.zig
dadepo: [80cb537](https://github.com/Syndica/sig/pull/875/commits/80cb537c153b1719bf8ceb5df8148f10c124b9c9)

src/replay/commit.zig
dadepo: [80cb537](https://github.com/Syndica/sig/pull/875/commits/80cb537c153b1719bf8ceb5df8148f10c124b9c9)

src/replay/execution.zig
```diff
@@ -484,7 +490,10 @@ pub fn processReplayResults(
                 try replay.freeze.freezeSlot(state.allocator, .init(
                     .from(state.logger),
                     state.account_store,
-                    &(state.epochs.getForSlot(slot) orelse return error.MissingEpoch),
+                    &(blk: {
+                        break :blk state.epoch_tracker.getForSlot(slot) orelse
+                            return error.MissingEpoch;
+                    }),
```
InKryption: Why not just separate this out into its own declaration?

InKryption (commented):
Just one very small nit left, otherwise lgtm.

src/replay/execution.zig
dadepo: Addressed [717d5db](https://github.com/Syndica/sig/pull/875/commits/717d5db9ec2cbf382d8fb2eaaf8e1956ada92ed8)

dnut (approved):

PR #876, opened 2025-08-18
refactor(trace): eliminate unscoped loggers

dnut:
### Rationale

All loggers should have a scope. There's no good reason to use a null scope, since there is *always* a scope, and logging that scope always gives better context to the observer of the logs.

Previously, unscoped loggers were used as an intermediary to pass loggers from one scope to another. This required the caller *and* callee to each execute a conversion from/to their respective loggers. It's more ergonomic to simply convert directly from one scope to the other. Only the caller needs to do anything with this approach.

Every context now only needs to concern itself with its own logger scope, and you simply call `.from` if passing a logger into another context.

### Overview of changes

Nearly all of the files changed here are to eliminate usages of unscoped loggers throughout sig, and change each file so that it defines its own logger with its own scope, and simply uses that throughout the entire file.

Additionally, the following changes were made to the Logger type to enforce this practice going forward:

- Eliminate the alias `Logger = ScopedLogger(null)`.
- Rename `ScopedLogger` to `Logger`.
- Make the logger type's `scope` parameter non-nullable.
- Remove `unscoped` function on logger that removes the scope from the logger.

### Codecov
99a9a16e702646f5e944a862d1bfd759332b9f81

I was also forced to add a unit test to get code coverage up, so I opted to do this for `shred_network.start` which currently has no code coverage and accounted for half of uncovered code in the report. This function is difficult to meaningfully test since all it does is allocate resources and start up other threads. So the test just makes sure that the function does not leak anything and that it shuts down quickly on exit. We've had problems with other high-level code in sig shutting down cleanly like this, so I think it is a meaningful test.

**Important:** There is one change to production code I made for this test. I changed the wait timeouts in `Channel` and `SocketThread` from 1 second to 10 ms. This makes threads shut down much more quickly, which is helpful for tests and generally makes sig's shutdown process snappier. It means these threads are woken up 100 times more frequently. But I think 10 ms would still have a negligible CPU impact, and is worth it.

### DIFF
--- a/src/accountsdb/index.zig
+++ b/src/accountsdb/index.zig
@@ -13,8 +13,7 @@ const SwissMap = sig.accounts_db.swiss_map.SwissMap;
 
 const createAndMmapFile = sig.utils.allocators.createAndMmapFile;
 
-const LOG_SCOPE = "accounts_db.index";
-const Logger = sig.trace.Logger(LOG_SCOPE);
+const Logger = sig.trace.Logger("accounts_db.index");
 
 /// reference to an account (either in a file or in the unrooted_map)
 pub const AccountRef = struct {


src/accountsdb/snapshots.zig
dadepo, line 1: There is still one `unscoped_logger` variable in this file that managed to escape being renamed. https://github.com/Syndica/sig/pull/876/files#diff-bb06cb3fac15909a4c1bc5dc82d702e2850b56f81d837901160faa7038cc422cR1252

src/accountsdb/index.zig
```diff
@@ -14,7 +14,7 @@ const SwissMap = sig.accounts_db.swiss_map.SwissMap;
 const createAndMmapFile = sig.utils.allocators.createAndMmapFile;
 
 const LOG_SCOPE = "accounts_db.index";
```
dadepo:
```suggestion
const Logger = sig.trace.Logger("accounts_db.index");
```

src/accountsdb/snapshots.zig
dnut: d0a839e13c693b43e8df8d74bdfd9e68f2d682fb

src/accountsdb/index.zig
dnut: e74f7786fdbfa6503f4c9a8e06de996d3c5161e7

PR #882, opened 2025-08-19
feat(txn): implement secp256r1 precompile program

Rexicon226:
Also deleted the ed25519 precompile fixtures in the passing list, which were just a workaround before all of them were working.

+3 passing fixtures


src/runtime/program/precompiles/secp256r1.zig
```diff
@@ -1,25 +1,509 @@
...
+
+        // check for low s in order to avoid malleable signatures
+        const s: u256 = @bitCast(signature.s);
+        if (@byteSwap(s) > HALF_ORDER) return error.InvalidSignature;
+
+        signature.verify(msg, pubkey) catch return error.InvalidSignature;
+    }
+}
+
+fn getInstructionData(
```
yewman: Looks like we have three implementations of this function, one for each precompile. Is there actually a material difference or can we consolidate?

Rexicon226: The secp256k1 program has different behaviour in its function. We can unify the ones for ed25519 and secp256r1 though.

Rexicon226: 54bc6431f3c16b2882398fc19b0d7aa012f5f282

PR #888, opened 2025-08-20
feat(replay): improve logging

dnut:
This reduces logging noise by moving redundant messages from info to debug logs, while adding some more meaningful logs to replay for novel information. This involves the introduction of a stateful `LogHelper` in replay so it avoids rapidly repeating the same message over and over when replay has no updates.

This also changes the default log level for all of sig. Instead of always being debug, now it's debug for debug builds and info for release builds.


src/replay/execution.zig
```diff
@@ -77,8 +79,8 @@ pub fn replayActiveSlots(state: ReplayExecutionState) !bool {
     defer zone.deinit();
 
     const active_slots = try state.slot_tracker.activeSlots(state.allocator);
-    defer state.allocator.free(active_slots);
```
InKryption, line 82: Is this removal intended?

src/replay/execution.zig
```diff
@@ -89,14 +91,53 @@ pub fn replayActiveSlots(state: ReplayExecutionState) !bool {
...
         const result = try replaySlot(state, slot);
         errdefer result.deinit(state.allocator);
         try slot_statuses.append(state.allocator, .{ slot, result });
     }
     return try processReplayResults(state, slot_statuses.items);
 }
 
+pub const LogHelper = struct {
+    logger: Logger,
+    last_active_slots: ?[]const u64,
```
InKryption: Why `u64` instead of `Slot`?

src/replay/execution.zig
```diff
@@ -89,14 +91,53 @@ pub fn replayActiveSlots(state: ReplayExecutionState) !bool {
...
-        state.logger.info().logf("replaying slot: {}", .{slot});
+        state.logger.debug().logf("replaying slot: {}", .{slot});
         const result = try replaySlot(state, slot);
         errdefer result.deinit(state.allocator);
         try slot_statuses.append(state.allocator, .{ slot, result });
     }
     return try processReplayResults(state, slot_statuses.items);
 }
 
+pub const LogHelper = struct {
```
dadepo, line 104: Maybe no need to make it public?

```suggestion
const LogHelper = struct {
```

Looks very specific to replay execution, it may not be applicable anywhere else?

dadepo (approved):

src/replay/execution.zig
```diff
@@ -399,33 +441,30 @@ fn updateConsensusForFrozenSlot(
...
-            status,
-        );
+        const maybe_entries = try awaitConfirmedEntriesForSlot(state, slot, status);
+
         // If entries is null, it means the slot failed or was skipped, so continue to next slot
         const entries = if (maybe_entries) |entries| entries else continue;
 
-        var slot_info = replay_state.slot_tracker.get(slot) orelse
-            return error.MissingSlotInTracker;
+        var slot_info = state.slot_tracker.get(slot) orelse return error.MissingSlotInTracker;
```
```suggestion
        const slot_info = state.slot_tracker.get(slot) orelse return error.MissingSlotInTracker;
```

src/replay/execution.zig
dnut: yes, it's freed by the log helper

src/replay/execution.zig
dnut: it is in ReplayState so it needs to be public.

src/replay/execution.zig
dnut: 88f74384bd74fe0fd137d9b9ac01db98ebca689e

src/replay/execution.zig
dnut: 164cbb1da911d48ef2ebed3eabe68827a6c65e02

yewman (approved):

PR #889, opened 2025-08-20
fix(runtime): fix remaining sanitization bugs

yewman:
+10 fixtures

### DIFF
--- a/src/replay/confirm_slot.zig
+++ b/src/replay/confirm_slot.zig
@@ -24,6 +24,8 @@ const Committer = replay.commit.Committer;
 const SvmGateway = replay.svm_gateway.SvmGateway;
 const TransactionScheduler = replay.scheduler.TransactionScheduler;
 
+const SlotHashes = sig.runtime.sysvar.SlotHashes;
+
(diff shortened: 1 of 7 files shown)


src/replay/confirm_slot.zig
```diff
@@ -78,11 +80,18 @@ pub fn confirmSlot(
     }
 
     try startPohVerify(allocator, logger, &future.poh_verifier, last_entry, entries, &future.exit);
+
+    // TODO: Sighhhhh, we need the sysvar cache...
+    const slot_hashes = try SlotHashes.init(allocator);
```
Rexicon226, line 85: a bit unrelated, but i don't like that `SlotHashes` allocates `entries`, this is a very strange design. Either we need `SlotHashes.create` which returns `*SlotHashes` or it just shouldn't allocate at all. I don't see the usecase where it needs to be allocated yet.

dnut, line 85: This is not correct and will cause breakages in replay if we merge it. I opened #891 to fix this.

yewman:
Replaced by https://github.com/Syndica/sig/pull/891

PR #891, opened 2025-08-21
 fix(runtime): fix remaining sanitization bugs v2

dnut:
supersedes #889 since that pr would introduce a bug in replay where incorrect (empty) slot hashes are going to be passed down. this pr takes those changes, and adds on top of it to get the correct slot hashes in replay.

this also refactors the lookup table resolver to combine its dependencies in a struct so it can be easily collected and passed around with clarity about what it's needed for.


yewman (approved):
lgtm!

Rexicon226 (approved):

PR #893, opened 2025-08-22
test(sig): speed up unit tests

dnut:
before:
```bash
> zig build test -Dno-run && time zig build test

real    1m33.854s
user    1m12.406s
sys     0m20.096s
```

after:
```bash
> zig build test -Dno-run && time zig build test

real    0m32.424s
user    0m22.690s
sys     0m6.637s
```

This PR is just a first pass to pick some low hanging fruit. 30 seconds is still way too slow. In the future I'd like to get this down below 10 seconds.

I took a couple strategies here. In general I targeted tests that take more than 1 second.

In some cases I was able to speed up an existing test by making it more efficient, or consolidating tests that share the same slow initialization. But most of the speedups come from the usage of a new build flag I added called `long-tests`. 

I tried to identify any slow tests that seem redundant because most of their code coverage is a duplicate of another test, or because they have an unnecessarily large number of iterations that effectively just fuzz the same code path over and over. If `long-tests` is not enabled, these tests will run in a limited capacity. As soon as `long-tests` is enabled, the full number of iterations and redundant test cases will run. `long-tests` is used in CI, but it is disabled by default when running `zig build test`.

To run the full test suite, run `zig build test -Dlong-tests`

### DIFF
--- a/src/crypto/bn254/tests.zig
+++ b/src/crypto/bn254/tests.zig
@@ -1,4 +1,3 @@
-const sig = @import("../../sig.zig");
 const std = @import("std");
 const bn254 = @import("lib.zig");
 
--- a/src/crypto/bn254/tests.zig
+++ b/src/crypto/bn254/tests.zig
@@ -1,3 +1,4 @@
+const sig = @import("../../sig.zig");
 const std = @import("std");
 const bn254 = @import("lib.zig");
 
@@ -273,7 +274,8 @@ test "pairing" {
         // zig fmt: on
     };
 
-    for (cases) |case| {
+    const cases_to_run = if (!sig.build_options.long_tests) cases[0..2] else cases;
+    for (cases_to_run) |case| {
         const N = 192 * 10;
         var buffer: [N]u8 = .{0} ** N;
         const input = try std.fmt.hexToBytes(&buffer, case.input);
--- a/src/crypto/bn254/tests.zig
+++ b/src/crypto/bn254/tests.zig
@@ -274,8 +274,7 @@ test "pairing" {
         // zig fmt: on
     };
 
-    const cases_to_run = if (!sig.build_options.long_tests) cases[0..2] else cases;
-    for (cases_to_run) |case| {
+    for (cases) |case| {
         const N = 192 * 10;
         var buffer: [N]u8 = .{0} ** N;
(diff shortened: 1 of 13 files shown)


src/crypto/bn254/tests.zig
```diff
@@ -273,7 +274,8 @@ test "pairing" {
         // zig fmt: on
     };
 
-    for (cases) |case| {
+    const cases_to_run = if (!sig.build_options.long_tests) cases[0..2] else cases;
```
Rexicon226: I'm fine with the range proof one not running u256, that could take a bit of time in debug mode, but these need to all run. It only takes a second to run. There is logical importance between each test case, so it doesn't make as much sense to run only the first two.

dnut: > these need to all run.

They will all run in CI regardless. 

> It only takes a second to run. 

I think 1 second is too long for a single unit test when you're trying to iterate and get a sanity check on local changes.

> There is logical importance between each test case, so it doesn't make as much sense to run only the first two.

Are you implying that each case has zero value unless you run every single case?

Is there another way to make this test faster?

Rexicon226: > I think 1 second is too long for a single unit test when you're trying to iterate and get a sanity check on local changes.

If you're trying to iterate fast, you would be using a filter. 

> Are you implying that each case has zero value unless you run every single case?

No, but it makes a lot less sense since it ends up just not testing 90% of the behaviour. When you run all tests, I believe it should be all unique paths/cases. Things like the agave conformance test case that takes super long, that is all doing the same computation, just doing a lot to satisfy the specific test case. Here, each entry is testing something different.

> Is there another way to make this test faster?

Don't run it.

dnut: Alright I can enable it, but I may revisit this in the future. I see your point though, and I agree about unique paths. I'm trying to balance conflicting priorities here.

I often use filters but sometimes that approach is unsatisfactory

dnut: d6b28d8a01593482ee0717548dd23065750f799b

Rexicon226 (approved):

Sobeston (approved):
Nice speed up, seems harmless :+1: 
Agreed we can go further but this is great for now

PR #895, opened 2025-08-25
feats:(replay, consensus): Process newly computed slot stats

dadepo:
The main changes is to implement the remaining part of duplicate-confirmation processing in consensus. This was left as a TODO to implement `tower_duplicate_confirmed_forks` and `mark_slots_duplicate_confirmed`

It also contain some minor diff:
- Introduces a consensus scoped Logger 
- improves resource handling for `SelectVoteAndResetForkResult`
- Update some default value for `ForkStats.EMPTY_ZEROES`


src/replay/consensus.zig
```diff
@@ -1636,3 +1719,307 @@ test "computeBankStats - same weight selects lower slot" {
...
+        testing.allocator,
+        .{ .root = root, .data = trees },
+        .active,
+    );
+
+    const SlotSet = sig.utils.collections.SortedSetUnmanaged(Slot);
+    var ancestors: std.AutoArrayHashMapUnmanaged(Slot, Ancestors) = .empty;
+    defer {
+        var it = ancestors.iterator();
+        while (it.next()) |entry| entry.value_ptr.deinit(testing.allocator);
```
```suggestion
        for (ancestors.values()) |val| val.deinit(testing.allocator);
```

src/replay/consensus.zig
```diff
@@ -1636,3 +1719,307 @@ test "computeBankStats - same weight selects lower slot" {
...
+        ancestors.deinit(testing.allocator);
+    }
+    var descendants: std.AutoArrayHashMapUnmanaged(Slot, SlotSet) = .empty;
+    defer descendants.deinit(testing.allocator);
+    defer {
+        var it = descendants.iterator();
+        while (it.next()) |entry| entry.value_ptr.deinit(testing.allocator);
+    }
```
```suggestion
    defer for (descendants.values()) |val| val.deinit(testing.allocator);
```

src/replay/consensus.zig
```diff
@@ -1636,3 +1719,307 @@ test "computeBankStats - same weight selects lower slot" {
...
+        .fork_choice = &fixture.fork_choice,
+        .ledger_reader = &ledger_reader,
+        .ledger_result_writer = &ledger_writer,
+        .ancestors = &ancestors,
+        .descendants = &descendants,
+        .vote_account = fixture.vote_pubkeys.items[0],
+        .slot_history_accessor = &(blk: {
+            var accessor = SlotHistoryAccessor.init(tsm1.accountReader());
+            break :blk accessor;
+        }),
```
InKryption: I know this is a test but this is a red flag. Prefer just declaring a variable outside instead of relying on this pointer to a temporary being mutable (which it possibly shouldn't be, and is just missing a compile error due to lvalues being propagated like this being a relatively recent-ish behavior, since 0.13 I think, and not set in stone).

InKryption (commented):
Just a couple of minor nits on some things, and a concern I'd like to raise to you as more of an awareness thing, all just in test code. Everything looks fine to me otherwise.

src/replay/consensus.zig
dadepo: [ad7506f](https://github.com/Syndica/sig/pull/895/commits/ad7506f698e4dd4f43c9d5464d9b452993790bac)

src/replay/consensus.zig
dadepo: [ad7506f](https://github.com/Syndica/sig/pull/895/commits/ad7506f698e4dd4f43c9d5464d9b452993790bac)

src/replay/consensus.zig
dadepo: [ad7506f](https://github.com/Syndica/sig/pull/895/commits/ad7506f698e4dd4f43c9d5464d9b452993790bac)

dnut (approved):

PR #896, opened 2025-08-25
fix(shred_network): Put upper limit on how many shreds to wait for

InKryption:
While debugging replay, we discovered this function ends up consuming way too many shreds at a time, which causes parts of replay to wait for unacceptably extended periods of time; 64 is a best-guess estimate for a balance between throughput and latency based on what we've observed to be typical.

Also combines the two separately allocated arraylists into a single multiarraylist.


src/shred_network/shred_processor.zig
```diff
@@ -35,35 +34,44 @@ pub fn runShredProcessor(
...
+    const MAX_SHREDS_PER_ITER = 64;
+    try shreds_buffer.ensureTotalCapacity(allocator, MAX_SHREDS_PER_ITER);
+
     while (true) {
         verified_shred_receiver.waitToReceive(.{ .unordered = exit }) catch break;
 
-        shreds.clearRetainingCapacity();
-        is_repaired.clearRetainingCapacity();
-        defer for (shreds.items) |shred| shred.deinit();
+        for (shreds_buffer.items(.shred)) |shred| shred.deinit();
```
dnut: this should be a defer to avoid leaking on exit

dnut (changes requested):

src/shred_network/shred_processor.zig
InKryption: There is a defer for this, it is declared above the loop. This is only for clearing the arraylist before re-using it

dnut: hmm OK, this approach feels redundant, but it is sound

dnut (approved):

src/shred_network/shred_processor.zig
```diff
@@ -22,53 +21,38 @@ const VariantCounter = sig.prometheus.VariantCounter;
...
+    params: Params,
 ) !void {
-    var shreds: ArrayListUnmanaged(Shred) = .{};
-    var is_repaired: ArrayListUnmanaged(bool) = .{};
     const metrics = try registry.initStruct(Metrics);
 
+    var shreds_buffer: ShredsBuffer = .empty;
+    defer shreds_buffer.deinit(allocator);
+    defer for (shreds_buffer.items(.shred)) |shred| shred.deinit();
+    try shreds_buffer.ensureTotalCapacity(allocator, MAX_SHREDS_PER_ITER);
```
Sobeston, line 45: Would it make sense to avoid the alloc as we know the size?

Sobeston (dismissed):
:+1: 
One comment, but it's nothing really so feel free to ignore

src/shred_network/shred_processor.zig
InKryption: Well, it is a pretty large amount of data, already 139_264 bytes at `MAX_SHREDS_PER_ITER = 1024`. We may or may not need to make this bigger - by how much I'm not sure, so rather leave that as a future potential change.

dnut (approved):

PR #898, opened 2025-08-26
fix(txn-harness): fix sanitisation issuesa

yewman:
+19 fixtures


conformance/src/txn_execute.zig
```diff
@@ -1263,7 +1263,7 @@ fn loadTransactionMesssage(
         else
             .v0,
         .{
-            .signature_count = @truncate(@max(1, header.num_required_signatures)),
+            .signature_count = @max(1, @as(u8, @truncate(header.num_required_signatures))),
```
Rexicon226, line 1291: What difference does this make? I don't see any semantic difference.

src/core/transaction.zig
```diff
@@ -246,6 +239,30 @@ pub const Transaction = struct {
...
+    /// [agave] https://github.com/anza-xyz/solana-sdk/blob/42711325c40b314dafe3d5a41eb5b19af49cf1dc/transaction/src/versioned/mod.rs#L120
+    pub fn validate(self: Transaction) !void {
+        try self.msg.validate();
+        try self.validateSignatures();
+    }
+
+    /// Run some sanity checks on the signature counts to ensure the internal data has consistency.
+    ///
+    /// [agave] https://github.com/anza-xyz/solana-sdk/blob/42711325c40b314dafe3d5a41eb5b19af49cf1dc/transaction/src/versioned/mod.rs#L126
+    pub fn validateSignatures(self: *const Transaction) !void {
```
Rexicon226: I'd prefer this be called `validateSignatureCounts`.

conformance/src/txn_execute.zig
```diff
@@ -753,28 +759,61 @@ fn executeTxnContext(
...
+                .err = .InvalidAddressLookupTableOwner,
+            },
+            error.InvalidAddressLookupTableData => break :blk .{
+                .err = .InvalidAddressLookupTableData,
+            },
+            error.InvalidAddressLookupTableIndex => break :blk .{
+                .err = .InvalidAddressLookupTableIndex,
+            },
+            else => std.debug.panic("Unexpected error: {s}\n", .{@errorName(err)}),
+        };
```
Rexicon226:
```suggestion
        }) catch |err| break :blk switch (err) {
            error.OutOfMemory => return error.OutOfMemory,
            error.AddressLookupTableNotFound => .{ .err = .AddressLookupTableNotFound },
            error.InvalidAddressLookupTableOwner => .{ .err = .InvalidAddressLookupTableOwner },
            error.InvalidAddressLookupTableData => .{ .err = .InvalidAddressLookupTableData },
            error.InvalidAddressLookupTableIndex => .{ .err = .InvalidAddressLookupTableIndex },
            else => std.debug.panic("Unexpected error: {s}\n", .{@errorName(err)}),
        };
```

conformance/src/txn_execute.zig
```diff
@@ -866,191 +905,187 @@ fn printLogs(result: TransactionResult) void {
...
+        .custom_error = errors.custom_error,
+
+        .return_data = return_data,
+        .resulting_state = resulting_state,
+        .fee_details = .{
+            .transaction_fee = fees.transaction_fee,
+            .prioritization_fee = fees.prioritization_fee,
         },
-    }
+        // TODO: obviously hard coded number. compute_meter counts how many units left instead of how many units consumed
```
Rexicon226: We can remove this comment.

src/replay/verify_transaction.zig
```diff
@@ -0,0 +1,168 @@
...
+
+const TransactionResult = sig.runtime.transaction_execution.TransactionResult;
+const ComputeBudgetInstructionDetails = compute_budget.ComputeBudgetInstructionDetails;
+
+pub const VerifyTransactionResult = TransactionResult(struct {
+    Hash,
+    ComputeBudgetInstructionDetails,
+});
+
+pub fn verifyTransaction(txn: Transaction) VerifyTransactionResult {
```
Rexicon226: Instead of having separate functions, take an enum to the `verifyTransaction` function,

```zig
pub fn verifyTransaction(
    txn: Transaction,
    should_verify: enum { skip, verify },
) VerifyTransactionResult {
    verifyTransactionInner(txn, should_verify) catch return .{ .err = .SanitizeFailure };
}

fn verifyTransactionInner(
    txn: Transaction,
    should_verify: enum { skip, verify },
) !VerifyTransactionResult {
    // If skipping signature verification, we still need to serialize and compute the hash
    // to ensure the transaction is well-formed.
    const msg_hash: Message = switch (should_verify) {
        .skip => m: {
            const msg_bytes = try txn.msg.serializeBounded(txn.version);
            break :m .hash(msg_bytes.slice());
        },
        .verify => try txn.verifyAndHashMessage(),
    };

    try txn.validate();

    const compute_budget_instruction_details = switch (compute_budget.execute(&txn.msg)) {
        .ok => |details| details,
        .err => |err| return .{ .err = err },
    };

    return .{ .ok = .{
        msg_hash,
        compute_budget_instruction_details,
    } };
}
```

dnut: At that point, you might as well consolidate it all down to a single function. `!VerifyTransactionResult` is a strange return type when you're just going to convert all errors into `VerifyTransactionResult.err` anyway.

Also I don't really like the name `verifyTransaction`. Verification is optional, and it's doing several other (not optional) tasks here as well. I think a more general name would be more correct. The enum can indicate whether verification is happening, and the return type indicates most of the other stuff that's going on. How about something like `preprocessTransaction`?

src/replay/verify_transaction.zig
```diff
@@ -0,0 +1,168 @@
...
+}
+
+fn verifyTransactionInner(txn: Transaction, skip_sig_verify: bool) VerifyTransactionResult {
+    // If skipping signature verification, we still need to serialize and compute the hash
+    // to ensure the transaction is well-formed.
+    const msg_hash = if (skip_sig_verify) blk: {
+        const msg_bytes = txn.msg.serializeBounded(txn.version) catch
+            return .{ .err = .SanitizeFailure };
+        break :blk Message.hash(msg_bytes.slice());
+    } else txn.verifyAndHashMessage() catch return .{ .err = .SanitizeFailure };
```
dnut: This is just a minor nitpick/personal preference. If you disagree, we can leave it as-is.

I find this style a little cumbersome to read. I prefer to avoid adding too many labelled blocks with extra declarations and break statements, where practical. It's just more syntax to parse which I find distracting. Also it requires more zig-specific knowledge, which is fine, but when all else is equal, less esoteric is better.

In this case, I think a more expressiony style is easier to read. It's definitely less code:

```suggestion
    const msg_hash = if (skip_sig_verify)
        Message.hash((txn.msg.serializeBounded(txn.version) catch
            return .{ .err = .SanitizeFailure }).slice())
    else
        txn.verifyAndHashMessage() catch return .{ .err = .SanitizeFailure };
```

src/replay/scheduler.zig
```diff
@@ -53,9 +54,17 @@ pub fn processBatch(
...
             return .exit;
         }
-        const hash = transaction.transaction.verifyAndHashMessage() catch
-            return .{ .failure = .SignatureFailure };
-        const runtime_transaction = transaction.toRuntimeTransaction(hash);
+
+        const hash, const compute_budget_details =
+            switch (verifyTransaction(transaction.transaction)) {
+                .ok => |res| res,
+                .err => |_| return .{ .failure = .SignatureFailure }, // This is not strictly true as there are other verification errors
```
dnut: If we're converting the error type at the call site, then I wouldn't return TransactionResult. `verifyTransaction` should just return a normal zig error.

Then again, regarding the code comment here, what is the correct error? If there are various errors that can occur, why aren't we distinguishing them? In that case, maybe `verifyTransaction` *should* return TransactionResult, but with the correct error, so no transformation is required here.

src/replay/scheduler.zig
```diff
@@ -53,9 +54,17 @@ pub fn processBatch(
...
+            };
+
+        const runtime_transaction = transaction.toRuntimeTransaction(
+            hash,
+            compute_budget_details,
+        );
```
dnut: mild suggestion

```suggestion
        const runtime_transaction = transaction.toRuntimeTransaction(hash, compute_budget_details);
```

src/replay/verify_transaction.zig
Rexicon226: > `!VerifyTransactionResult` is a strange return type when you're just going to convert all errors into `VerifyTransactionResult.err` anyway.

I don't understand what you mean, this simplifies the logic. Instead of having the `catch return .{ .err` everywhere, we just `try` no matter what the error is, and then catch. The inner function is not `pub`, so there's no chance of accidentally using it. This makes the code safer and clearer. "If the verification (or whatever you'd like to call it) fails, return sanitize failure".

> Also I don't really like the name `verifyTransaction`.

I can't really comment on that, whatever name it fine for me.

conformance/src/txn_execute.zig
yewman: if num_required_signatures is a multiple of 256 the former will be 0 and the later will be 1

Rexicon226: ah, good point, i did not see that. 

i don't like that..., but ok.

src/core/transaction.zig
yewman: 3ba3d661

conformance/src/txn_execute.zig
yewman: 3ba3d661

conformance/src/txn_execute.zig
yewman: 3ba3d661

src/replay/scheduler.zig
yewman: Not sure what went through my head here haha. 5e4f8ca4

src/replay/scheduler.zig
yewman: 5e4f8ca4

src/replay/verify_transaction.zig
yewman: 5e4f8ca4

src/replay/verify_transaction.zig
yewman: 58cec8e4

Rexicon226 (approved):

dnut (approved):

PR #906, opened 2025-08-27
fix(txn): update Instruction account owner 

kprotty:


src/runtime/account_loader.zig
```diff
@@ -559,7 +559,7 @@ fn accumulateAndCheckLoadedAccountDataSize(
     }
 }
 
-// [agave] https://github.com/anza-xyz/agave/blob/cb32984a9b0d5c2c6f7775bed39b66d3a22e3c46/svm/src/account_loader.rs#L639
+// [agave] https://github.com/firedancer-io/agave/blob/agave-v2.3.7-patches/svm/src/account_loader.rs#L880
```
dnut: I don't think we should be including links to agave that are hosted by firedancer. If this is really part of agave, the same code should also be in anza's repo.

kprotty: https://github.com/Syndica/sig/pull/906/commits/b88a66d97ccae29fd3940728b3f849dfb863d764

PR #909, opened 2025-08-28
feat(runtime): implement simd 186

Rexicon226:


yewman (approved):

src/runtime/account_loader.zig
```diff
@@ -867,11 +1015,150 @@ test "load accounts rent paid" {
...
+    });
+
+    var tx = try emptyTxWithKeys(allocator, &.{
+        fee_payer_address,
+        instruction_address,
+        program_address,
+    });
+    defer tx.accounts.deinit(allocator);
+
+    tx.instructions = &.{
```
kprotty, line 1108: Could this use `sig.runtime.testing.createInstructionInfo()` instead? In https://github.com/Syndica/sig/pull/917, the AccountMetas `index_in_{caller/callee}` fields are removed, so its just a heads up.

Rexicon226: `createInstructionInfo` takes in a transaction context, I'd need to create that as well, which seems like overall just more work.

kprotty (approved):

PR #910, opened 2025-08-28
feat(conformance): migrate to agave v3.0.0

yewman:


Rexicon226 (approved):
lgtm

src/runtime/program/bpf_loader/execute.zig
```diff
@@ -1285,217 +1286,220 @@ fn commonExtendProgram(
...
-            @intFromEnum(AccountIndex.program_data),
-        );
-        defer programdata.release();
-
-        var resize_delta: i64 = undefined;
-        try programdata.setDataLength(allocator, &resize_delta, 0); // set_data_from_slice(&[])
-        try programdata.setOwner(system_program.ID);
-    }
+    ic.tc.custom_error = 0x30000000;
+    return error.Custom;
```
dnut, line 1293: Is this to make it easier to detect when this function is the cause of a conformance failure?

I understand you plan to revert this soon, so it's not a big deal, but it's still a little unfortunate that we need to comment out code and its test. Is it possible to keep this compiling and running the test, without sacrificing the signal of this error code, and without adding much work?

Maybe we can set some state in the transaction context *before* this function is called. This function can check the value to inform it what context it's running in (e.g. conformance vs unit test), and then use that information to guide whether it should return `error.Custom` or execute the full logic. Is there something like this that you could quickly get working?

dnut (commented):
seems fine but I just want to understand the purpose and alternatives a little better

src/runtime/program/bpf_loader/execute.zig
yewman: Yeah exactly, we were actually passing some fixtures which called into either bpf loader v3 migrate of bpf loader v4 by chance. I would prefer to keep it as is since the bpf v3 migrate instruction does not make sense without loader v4 implemented. I think setting some state in the transaction context is a more convoluted approach to solving the same issue. If you feel strongly about it I can implement a work around but I think it is more clear to leave this code commented out for now and fix it up once loader v4 is implemented

PR #912, opened 2025-08-28
fix(replay): error when feature account not owned by feature program

dnut:
This should not result in an error. the account should just not be used to activate a feature unless it's owned by the feature program. See the agave code in the permalink I added.

This is causing a problem now on testnet because the `zkiTNuzBKxrCLMKehzuQeKZyLtX2yvFcEKMML8nExU8` account was recently funded but has ownership not yet been transferred to the feature program.

### DIFF
--- a/src/replay/service.zig
+++ b/src/replay/service.zig
@@ -656,6 +656,10 @@ fn newSlotFromParent(
     return .{ constants, state };
 }
 
+/// Determines which features are active for this slot by looking up the feature
+/// accounts in accountsdb.
+///
+/// Analogous to [compute_active_feature_set](https://github.com/anza-xyz/agave/blob/785455b5a3e2d8a95f878d6c80d5361dea9256db/runtime/src/bank.rs#L5338-L5339)
 // TODO: epoch boundary - handle feature activations
 pub fn getActiveFeatures(
     allocator: Allocator,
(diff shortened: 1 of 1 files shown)


yewman (approved):

src/replay/service.zig
```diff
@@ -688,19 +692,29 @@ test "getActiveFeatures rejects wrong ownership" {
     var accounts = std.AutoArrayHashMapUnmanaged(Pubkey, sig.core.Account).empty;
     defer accounts.deinit(allocator);
 
+    var slot_0_bytes: [9]u8 = .{ 1, 0, 0, 0, 0, 0, 0, 0, 0 };
```
Sobeston:
```suggestion
    // bincode for a feature that activated at slot 0
    var slot_0_bytes: [9]u8 = .{ 1, 0, 0, 0, 0, 0, 0, 0, 0 };
```

Sobeston (approved):

src/replay/service.zig
```diff
@@ -688,19 +692,29 @@ test "getActiveFeatures rejects wrong ownership" {
     var accounts = std.AutoArrayHashMapUnmanaged(Pubkey, sig.core.Account).empty;
     defer accounts.deinit(allocator);
```
ultd, line 693: 🚀

ultd (dismissed):

PR #913, opened 2025-08-29
refactor(accountsdb): reorganize manager loop for deterministic fuzzing

dnut:
switched from this pattern:
```zig
fn runLoop() {
    const state1 = ...;
    defer state1.deinit();

    const state2 = ...;
    defer state2.deinit();

    while (true) {
        step1();
        step2();
        ...
    }
}
```

to this pattern:

```zig
pub const Manager = struct {
    state1: ...,
    state2: ...,

    pub fn init() {}

    pub fn deinit() {}

    pub fn run() {
        while (true) self.manage();
    }

    pub fn manage() {
        step1();
        step2();
        ...
    }
};
```

this means it can be used in the same thread by calling `manager.manage()` instead of needing to spawn up another thread and wait for it to do its work non-deterministically.


yewman (approved):
lgtm

PR #920, opened 2025-09-01
feat(replay): minimal and offline replay

dnut:
This adds two features
- offline replay - replay a downloaded ledger
- minimal replay - run replay single-threaded without consensus

## CLI changes

- new subcommand: `replay-offline`: run replay offline, without gossip or shred-network, replaying only pre-downloaded data.
- new option: `--disable-consensus`
- new option: `--replay-threads`

## Offline Replay

This mode runs replay without requiring any network connectivity. Gossip and the shred network are not started. You must already have a snapshot downloaded and some shreds in the ledger building off that snapshot.

## Single-threaded replay

Currently, replay is designed to be highly concurrent and asynchronous. It executes batches concurrently based on account-locking logic, and it executes slots concurrently when they fork.

This adds a variant of replay that executes entirely single-thread. The concurrent version is having some memory safety issues. We just need something that works and my hope is this can avoid the thread-safety bugs that we haven't figured out yet.

## Disable consensus

When consensus is disabled, it runs `bypassConsensus`, which freezes the slot and optimistically roots slots using SlotTree, which was added as a simplified alternative to tracking forks with consensus. The basic idea is that it roots slots after 32 slots.

## Diff Overview

Most of the diff here is refactoring existing code. This PR adds three new run modes (online/simple, offline/full, offline/simple). To avoid copy and pasting everything three times, I extracted out logic that is used in multiple run modes. This also helped organization. For example, there's a clear division between the consensus-specific state and the other replay state, which are now distinguished with different structs.

### DIFF
--- a/src/accountsdb/snapshot/load.zig
+++ b/src/accountsdb/snapshot/load.zig
@@ -120,8 +120,6 @@ pub fn loadSnapshot(
             n_threads_snapshot_load,
             load_options.validate_snapshot,
             db_config.accounts_per_file_estimate,
-            db_config.fastload,
-            db_config.save_index,
         );
     errdefer collapsed_manifest.deinit(allocator);
 
--- a/src/cmd.zig
+++ b/src/cmd.zig
@@ -127,7 +127,6 @@ pub fn main() !void {
             current_config.genesis_file_path = params.genesis_file_path;
             params.accountsdb_base.apply(&current_config);
             params.accountsdb_download.apply(&current_config);
-            params.accountsdb_index.apply(&current_config);
             params.geyser.apply(&current_config);
             current_config.replay_threads = params.replay_threads;
             current_config.disable_consensus = params.disable_consensus;
@@ -143,7 +142,6 @@ pub fn main() !void {
             current_config.genesis_file_path = params.genesis_file_path;
             params.accountsdb_base.apply(&current_config);
             params.accountsdb_download.apply(&current_config);
-            params.accountsdb_index.apply(&current_config);
             params.geyser.apply(&current_config);
             current_config.replay_threads = params.replay_threads;
             current_config.disable_consensus = params.disable_consensus;
@@ -173,7 +171,6 @@ pub fn main() !void {
             current_config.accounts_db.snapshot_dir = params.snapshot_dir;
             current_config.genesis_file_path = params.genesis_file_path;
             params.accountsdb_base.apply(&current_config);
-            params.accountsdb_index.apply(&current_config);
             current_config.gossip.cluster = params.gossip_cluster;
             params.geyser.apply(&current_config);
             try validateSnapshot(gpa, current_config);
@@ -215,7 +212,6 @@ pub fn main() !void {
             current_config.genesis_file_path = params.genesis_file_path;
             params.accountsdb_base.apply(&current_config);
             params.accountsdb_download.apply(&current_config);
-            params.accountsdb_index.apply(&current_config);
             try mockRpcServer(gpa, current_config);
         },
     }
@@ -561,34 +557,6 @@ const Cmd = struct {
             cfg.gossip.trusted_validators = args.trusted_validators;
         }
     };
-    const AccountsDbArgumentsIndex = struct {
-        fastload: bool,
-        save_index: bool,
-
-        const cmd_info: cli.ArgumentInfoGroup(@This()) = .{
-            .fastload = .{
-                .kind = .named,
-                .name_override = null,
-                .alias = .none,
-                .default_value = false,
-                .config = {},
-                .help = "fastload the accounts db",
-            },
-            .save_index = .{
-                .kind = .named,
-                .name_override = null,
-                .alias = .none,
-                .default_value = false,
-                .config = {},
-                .help = "save the account index to disk",
-            },
-        };
-
-        fn apply(args: @This(), cfg: *config.Cmd) void {
-            cfg.accounts_db.fastload = args.fastload;
-            cfg.accounts_db.save_index = args.save_index;
-        }
-    };
     const RepairArgumentsBase = struct {
         turbine_port: u16,
         repair_port: u16,
@@ -742,7 +710,6 @@ const Cmd = struct {
         accountsdb_base: AccountsDbArgumentsBase,
         accountsdb_download: AccountsDbArgumentsDownload,
         force_new_snapshot_download: bool,
-        accountsdb_index: AccountsDbArgumentsIndex,
         geyser: GeyserArgumentsBase,
         replay_threads: u16,
         disable_consensus: bool,
@@ -763,7 +730,6 @@ const Cmd = struct {
                 .accountsdb_base = AccountsDbArgumentsBase.cmd_info,
                 .accountsdb_download = AccountsDbArgumentsDownload.cmd_info,
                 .force_new_snapshot_download = force_new_snapshot_download_arg,
-                .accountsdb_index = AccountsDbArgumentsIndex.cmd_info,
                 .geyser = GeyserArgumentsBase.cmd_info,
                 .replay_threads = replay_threads_arg,
                 .disable_consensus = disable_consensus_arg,
@@ -856,7 +822,6 @@ const Cmd = struct {
         snapshot_dir: []const u8,
         genesis_file_path: ?[]const u8,
         accountsdb_base: AccountsDbArgumentsBase,
-        accountsdb_index: AccountsDbArgumentsIndex,
         gossip_cluster: ?[]const u8,
         geyser: GeyserArgumentsBase,
 
@@ -869,7 +834,6 @@ const Cmd = struct {
                 .snapshot_dir = snapshot_dir_arg,
                 .genesis_file_path = genesis_file_path_arg,
                 .accountsdb_base = AccountsDbArgumentsBase.cmd_info,
-                .accountsdb_index = AccountsDbArgumentsIndex.cmd_info,
                 .gossip_cluster = gossip_cluster_arg,
                 .geyser = GeyserArgumentsBase.cmd_info,
             },
@@ -994,7 +958,6 @@ const Cmd = struct {
         accountsdb_base: AccountsDbArgumentsBase,
         accountsdb_download: AccountsDbArgumentsDownload,
         force_new_snapshot_download: bool,
-        accountsdb_index: AccountsDbArgumentsIndex,
 
         const cmd_info: cli.CommandInfo(@This()) = .{
             .help = .{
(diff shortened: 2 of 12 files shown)


src/accountsdb/snapshot/load.zig
```diff
@@ -0,0 +1,283 @@
...
+
+pub const LoadedSnapshot = struct {
+    allocator: Allocator,
+    accounts_db: AccountsDB,
+    combined_manifest: snapshot.FullAndIncrementalManifest,
+    collapsed_manifest: snapshot.Manifest,
+    genesis_config: GenesisConfig,
+    status_cache: ?snapshot.StatusCache,
+
+    pub fn deinit(self: *@This()) void {
```
dadepo: Nit:

```suggestion
    pub fn deinit(self: *LoadedSnapshot) void {
```

src/accountsdb/snapshot/load.zig
```diff
@@ -0,0 +1,283 @@
...
+    logger: Logger,
+    options: LoadSnapshotOptions,
+) !LoadedSnapshot {
+    const zone = tracy.Zone.init(@src(), .{ .name = "loadSnapshot" });
+    defer zone.deinit();
+
+    var validator_dir = try std.fs.cwd().makeOpenPath(sig.VALIDATOR_DIR, .{});
+    defer validator_dir.close();
+
+    // const genesis_file_path = try cfg.genesisFilePath() orelse return error.GenesisPathNotProvided;
```
dadepo: Nit
```suggestion

```

src/accountsdb/snapshot/load.zig
```diff
@@ -0,0 +1,283 @@
...
+    /// whether to load only the metadata of the snapshot
+    metadata_only: bool = false,
+};
+
+pub fn loadSnapshot(
+    allocator: Allocator,
+    config: sig_config.AccountsDB,
+    genesis_file_path: []const u8,
+    logger: Logger,
+    options: LoadSnapshotOptions,
```
dadepo: both input arguments named `options` and `config` may be confusing as they could be seen as synonyms. maybe `accounts_db_config` instead of just `config`?

src/core/entry.zig
```diff
@@ -25,6 +25,20 @@ pub const Entry = struct {
         for (self.transactions) |tx| tx.deinit(allocator);
         allocator.free(self.transactions);
     }
+
+    pub fn clone(self: Entry, allocator: std.mem.Allocator) Allocator.Error!Entry {
+        const transactions = try allocator.dupe(Transaction, self.transactions);
+        errdefer allocator.free(transactions);
+        for (transactions, 0..) |*txn, i| {
+            errdefer for (0..i) |j| transactions[j].deinit(allocator);
+            txn.* = try txn.clone(allocator);
```
dadepo, line 36: is this correct? the fact that the pointer value being cloned is being assigned to the same pointer

src/ledger/lib.zig
```diff
@@ -17,3 +17,119 @@ pub const LedgerDB = db.LedgerDB;
...
+    db: *LedgerDB,
+    reader: *LedgerReader,
+    result_writer: *LedgerResultWriter,
+    cleanup_service_handle: ?std.Thread,
+    exit: *std.atomic.Value(bool),
+
+    const std = @import("std");
+    const sig = @import("../sig.zig");
+
+    /// join first to avoid UB
```
dadepo:
```suggestion
    /// Call join() before deinit() to avoid UB.
```

Was initially confused as there are no `join` calls within `deinit`

And related question, in situation like this where certain methods needs to be called in order to maintain correct behaviour, any reason why not to enforce this order - ie call the dependent directly from the other function? instead of conveying this via comments?

src/core/entry.zig
```diff
@@ -25,6 +25,20 @@ pub const Entry = struct {
         for (self.transactions) |tx| tx.deinit(allocator);
         allocator.free(self.transactions);
     }
+
+    pub fn clone(self: Entry, allocator: std.mem.Allocator) Allocator.Error!Entry {
```
yewman, line 30: It looks like this is only used for testing, what are your thoughts on adding a comptime check for test mode?

src/core/leader_schedule.zig
```diff
@@ -162,12 +162,35 @@ pub const LeaderSchedule = struct {
         };
     }
 
+    pub fn fromVoteAccounts(
+        allocator: std.mem.Allocator,
+        epoch: Epoch,
+        slots_in_epoch: Slot,
+        vote_accounts: *const sig.core.vote_accounts.StakeAndVoteAccountsMap,
+    ) ![]const Pubkey {
+        var stakes = std.AutoArrayHashMapUnmanaged(Pubkey, u64){};
```
yewman, line 174:
```suggestion
        // TODO: this implementation is naive and performs unnecessay allocations to construct and 
        // input compatable with fromStakedNodes and re-key results. 
        // It should be addressed as part of issue #945
        var stakes = std.AutoArrayHashMapUnmanaged(Pubkey, u64){};
```

src/cmd.zig
```diff
@@ -1860,6 +1944,85 @@ fn startGossip(
...
+
+    const lt_hash = if (loaded_snapshot.collapsed_manifest.bank_extra.accounts_lt_hash) |lt_hash|
+        sig.core.LtHash{ .data = lt_hash }
+    else
+        null;
+
+    var root_slot_state = try sig.core.SlotState.fromBankFields(allocator, bank_fields, lt_hash);
+    errdefer root_slot_state.deinit(allocator);
+
+    return sig.replay.service.Dependencies{
```
dadepo:
```suggestion
    return replay.Dependencies{
```

or the return type can be updated.

initially wondered if what's defined as the return type and what is returned are actually same.

src/cmd.zig
```diff
@@ -1860,6 +1944,85 @@ fn startGossip(
...
+            try epoch_stakes.current.convert(allocator, .delegation),
+        ),
+        .hard_forks = try bank_fields.hard_forks.clone(allocator),
+    };
+}
+
+fn consensusDependencies(
+    allocator: std.mem.Allocator,
+    gossip_table: ?*sig.sync.RwMux(sig.gossip.GossipTable),
+) !replay.service.ConsensusState.Dependencies {
```
dadepo, line 1983:
```suggestion
) !replay.Dependencies {
```

To be consistent with `replayDependencies`

src/cmd.zig
```diff
@@ -1860,6 +1944,85 @@ fn startGossip(
...
+    const senders: sig.replay.ConsensusState.Senders = try .create(allocator);
+    defer senders.destroy();
+
```
dadepo, line 1986: I think using `defer` here (and also with receivers) is invalid. You probably want `errdefer`? As this will destroy the channels on a successful return, leaving the returned handles invalid.

src/replay/service.zig
```diff
@@ -30,26 +32,65 @@ const LatestValidatorVotes = sig.consensus.latest_validator_votes.LatestValidato
...
+    replay: ReplayState,
+    consensus: ?ConsensusState,
+    num_threads: u32,
+
+    pub fn init(
+        deps: Dependencies,
+        enable_consensus: ?ConsensusState.Dependencies,
+        num_threads: u32,
+    ) !Service {
+        var state = try ReplayState.init(deps, num_threads);
```
dadepo, line 68: nit:

```suggestion
        var replay_state = try ReplayState.init(deps, num_threads);
        errdefer state.deinit();

        var consensus_state: ?ConsensusState = if (enable_consensus) |consensus_deps|
            try ConsensusState.init(deps, consensus_deps, &state)
```

At the risk of slight verbosity. This makes the naming consistent (as it is not clear why the state for replay is only called state, while state for consensus is called consensus)

src/core/entry.zig
dnut: I don't see how that helps. Why should we block usage in production code?

src/core/entry.zig
dnut: yes, this converts a shallow copy into a deep copy

InKryption (dismissed):

src/accountsdb/snapshot/load.zig
dnut: This code is not from the current pr, but sure I can change it. 5e2548886c13b97da48c1dc22875bc8998e2a2b9

src/accountsdb/snapshot/load.zig
dnut: This code is not from the current pr, but sure I can change it. 5e2548886c13b97da48c1dc22875bc8998e2a2b9

src/accountsdb/snapshot/load.zig
dnut: This code is not from the current pr, but sure I can change it. 5e2548886c13b97da48c1dc22875bc8998e2a2b9

src/ledger/lib.zig
dnut: I don't think deinit functions should block indefinitely waiting for other threads because it's very confusing when the code hangs indefinitely on deinit functions. It's a problem I've seen many times. It would be good to formalize some best practices for this.

c36732d0bd649ddb00f9dacbbb210b744ee5845d

src/cmd.zig
dnut: replay.Dependencies is the wrong type

src/replay/service.zig
dnut: because state is the full state, and consensus state is the subset of state that's for consensus.

src/cmd.zig
dnut: > initially wondered if what's defined as the return type and what is returned are actually same.

it wouldn't compile otherwise

bf23e534cda7763fd763a606e6cfb5896c2fa74d

src/cmd.zig
dnut: 95776ba99294985e7d4e77f5db7ef8c99cbdb0b9

src/core/leader_schedule.zig
dnut: I think we should delete TODOs once they are represented with a ticket in a ticketing system.

dnut: 41565b30677c0772a6187e4f8db7e9de0798c267

InKryption (approved):

src/core/entry.zig
yewman: I think generally we should be avoiding cloning unless there is a very good reason. I think it is preferable to use the builtin check so that a read / potential user of this function thinks twice before calling it in a production context. If we come across a scenario where this is required, then we can remove it.

dnut: 5307eb9d746b2252eb824e549e1b3607c7d99bc9

yewman (approved):

yewman (approved):

PR #921, opened 2025-09-02
fix(replay): integer overflow due to returning stack pointer

dnut:
`epoch_info` was being copied to the stack and then a pointer to it was returned. It should be a pointer to epoch tracker so it lives long enough.

This was causing an integer overflow in the svm's `validateFeePayer` because the RentCollector had undefined data.


Sobeston (dismissed):
Nice catch!

PR #930, opened 2025-09-03
refactor(accountsdb:store): Simplify and improve ThreadSafeAccountMap

InKryption:
* Use just one `RwLock`
* Add `last_rooted_slot: ?Slot` field in order to be able to correctly replicate accountsdb behaviour of falling back to returning the last rooted slot in the case where an account is present but is not in a slot in the ancestors set.

### DIFF
--- a/src/accountsdb/account_store.zig
+++ b/src/accountsdb/account_store.zig
@@ -194,7 +194,9 @@ pub const SlotAccountReader = union(enum) {
 pub const ThreadSafeAccountMap = struct {
     rwlock: std.Thread.RwLock.DefaultRwLock,
     allocator: Allocator,
+    /// Ordering of slots (the entry keys) is not guaranteed.
     slot_map: SlotMap,
+    /// Ordering of the list of versions in each entry value is sorted highest to lowest.
     pubkey_map: PubkeyMap,
     last_rooted_slot: ?Slot,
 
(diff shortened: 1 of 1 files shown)


InKryption:
* **#930** <a href="https://app.graphite.dev/github/pr/Syndica/sig/930?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Syndica/sig/930?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Sobeston (dismissed):

dnut (approved):

src/accountsdb/account_store.zig
```diff
@@ -255,26 +252,55 @@ pub const ThreadSafeAccountMap = struct {
         address: Pubkey,
         ancestors: *const Ancestors,
     ) !?Account {
```
yewman, line 256: This seems like overkill. We need to return the account at the first slot which is in ancestors or default to the first slot which is less then or equal to the last_rooted_slot. Why would the following not work if the arrays in the pubkey map are sorted in descending order?
```zig
for (slot, account) {
    if (slot <= last_rooted_slot or ancestors.contains(slot)) {
        if (account.lamports == 0) return null;
        return account;
    }
 } 
 return null;
```

InKryption: 5868deab9f71d2e7a78b8efb706140374daf0b97

yewman (approved):

PR #931, opened 2025-09-03
feat(metrics): Add some consensus metrics

dadepo:
Adds a couple of metrics to see consensus related updates in replay

```
Fork Choice Metrics 
current_root_slot: Current rooted slot for consensus progress tracking
current_heaviest_subtree_slot - Slot with most stake (consensus weight)
current_deepest_slot - Highest tree height slot (network depth)
active_fork_count - Number of active fork candidates
total_stake_in_tree - Total stake participating in consensus
fork_choice_updates - Count of fork choice updates (consensus activity)
fork_choice_update_interval - Time between updates (performance indicator)

Replay Tower Metrics
fork_switching_successes - Successful fork switches via switch proofs
fork_switching_failures - Failed fork switch attempts
stale_votes_rejected - Old votes rejected for being too stale
stray_restored_slots - Slots restored from stray vote handling

Vote Listener Metrics
gossip_votes_received - Votes received from gossip network
replay_votes_received - Votes received from replay stage
gossip_votes_processed - Gossip votes successfully processed
replay_votes_processed - Replay votes successfully processed
```

### DIFF
--- a/src/consensus/replay_tower.zig
+++ b/src/consensus/replay_tower.zig
@@ -2779,6 +2779,9 @@ test "default thresholds" {
     const trees = try std.BoundedArray(TreeNode, MAX_TEST_TREE_LEN).init(0);
     try fixture.fillFork(allocator, .{ .root = root, .data = trees }, .active);
 
+    var registry = sig.prometheus.Registry(.{}).init(allocator);
+    defer registry.deinit();
+
     var replay_tower = try ReplayTower.init(
         allocator,
         .noop,
@@ -2786,6 +2789,7 @@ test "default thresholds" {
         Pubkey.ZEROES,
         root.slot,
         .noop,
+        &registry,
     );
     defer replay_tower.deinit(allocator);
 
--- a/src/consensus/vote_listener.zig
+++ b/src/consensus/vote_listener.zig
@@ -105,6 +105,7 @@ pub const VoteListener = struct {
         allocator: std.mem.Allocator,
         exit: sig.sync.ExitCondition,
         logger: Logger,
+        registry: *sig.prometheus.Registry(.{}),
         params: struct {
             slot_data_provider: SlotDataProvider,
             gossip_table_rw: ?*sig.sync.RwMux(sig.gossip.GossipTable),
@@ -1756,7 +1757,7 @@ test "simple usage" {
     var exit: std.atomic.Value(bool) = .init(false);
     const exit_cond: sig.sync.ExitCondition = .{ .unordered = &exit };
 
-    const vote_listener: VoteListener = try .init(allocator, exit_cond, .noop, .{
+    const vote_listener: VoteListener = try .init(allocator, exit_cond, .noop, &registry, .{
         .slot_data_provider = slot_data_provider,
         .gossip_table_rw = &gossip_table_rw,
         .ledger_ref = ledger_ref,
--- a/src/replay/consensus/core.zig
+++ b/src/replay/consensus/core.zig
@@ -223,6 +223,7 @@ pub const TowerConsensus = struct {
             deps.vote_identity,
             deps.root_slot,
             deps.account_reader.forSlot(&slot_tracker.get(slot_tracker.root).?.constants.ancestors),
+            sig.prometheus.globalRegistry(),
         );
         errdefer replay_tower.deinit(allocator);
 
@@ -238,6 +239,7 @@ pub const TowerConsensus = struct {
             allocator,
             .{ .unordered = deps.exit },
             .from(deps.logger),
+            sig.prometheus.globalRegistry(),
             .{
                 .slot_data_provider = slot_data_provider,
                 .gossip_table_rw = deps.external.gossip_table,
@@ -299,10 +301,15 @@ pub const TowerConsensus = struct {
         //
         // Analogous to [`new_from_frozen_banks`](https://github.com/anza-xyz/agave/blob/0315eb6adc87229654159448344972cbe484d0c7/core/src/consensus/heaviest_subtree_fork_choice.rs#L235)
         var heaviest_subtree_fork_choice: HeaviestSubtreeForkChoice =
-            try .init(allocator, .from(logger), .{
-                .slot = root_slot,
-                .hash = root_hash,
-            });
+            try .init(
+                allocator,
+                .from(logger),
+                .{
+                    .slot = root_slot,
+                    .hash = root_hash,
+                },
+                sig.prometheus.globalRegistry(),
+            );
         errdefer heaviest_subtree_fork_choice.deinit();
 
         var prev_slot = root_slot;
--- a/src/replay/consensus/process_result.zig
+++ b/src/replay/consensus/process_result.zig
@@ -305,10 +305,15 @@ const TestReplayStateResources = struct {
         self.progress = ProgressMap.INIT;
 
         self.fork_choice = try allocator.create(HeaviestSubtreeForkChoice);
-        self.fork_choice.* = try HeaviestSubtreeForkChoice.init(allocator, .noop, .{
-            .slot = 0,
-            .hash = Hash.ZEROES,
-        });
+        self.fork_choice.* = try HeaviestSubtreeForkChoice.init(
+            allocator,
+            .noop,
+            .{
+                .slot = 0,
+                .hash = Hash.ZEROES,
+            },
+            &self.registry,
+        );
 
         self.duplicate_slots_tracker = DuplicateSlots.empty;
         self.unfrozen_gossip_verified_vote_hashes = UnfrozenGossipVerifiedVoteHashes{


src/consensus/fork_choice.zig
```diff
@@ -202,6 +204,54 @@ pub const ForkChoice = struct {
...
+                        maybe_current_heaviest_slot = fork_info.heaviest_subtree_slot.slot;
+                    }
+                }
+
+                if (maybe_current_deepest_slot) |current_deepest_slot| {
+                    if (fork_info.deepest_slot.slot > current_deepest_slot) {
+                        maybe_current_deepest_slot = fork_info.deepest_slot.slot;
+                    }
+                }
+            }
```
Sobeston: could reduce nesting a lot here

src/consensus/vote_listener.zig
```diff
@@ -240,9 +246,10 @@ const UnverifiedVoteReceptor = struct {
 
         const epoch_tracker, var epoch_lg = slot_data_provider.epoch_tracker_rw.readWithLock();
         defer epoch_lg.unlock();
-        // TODO: Question should the lock for epoch_tracker be held here?
-        // or *sig.sync.RwMux(EpochTracker) be passed into consumeTransactionsAndSendVerified and passed
-        // down and only aquired at the site where it is accessed?
```
Sobeston, line 241: Did you intend to remove this comment?

Sobeston (approved):
All seems good, just a couple nits

src/consensus/vote_listener.zig
dadepo: Yeah, the comment is kinda redundant. Although yeah I agree may not be the best pr to remove it

src/consensus/fork_choice.zig
dadepo: An attempt [87ea287](https://github.com/Syndica/sig/pull/931/commits/87ea28715ff2dbe0b8b5bb4f6a76f652c1c6f4da)

src/consensus/fork_choice.zig
```diff
@@ -5026,3 +5107,40 @@ pub fn testEpochStakes(
...
+
+    /// Number of fork choice updates - indicates consensus activity and health
+    fork_choice_updates: *sig.prometheus.Counter,
+    /// Time between fork choice updates (seconds) - performance and network health indicator
+    fork_choice_update_interval: *sig.prometheus.Histogram,
+
+    pub const prefix = "fork_choice";
+
+    pub fn init() sig.prometheus.GetMetricError!ForkChoiceMetrics {
+        return sig.prometheus.globalRegistry().initStruct(ForkChoiceMetrics);
```
dnut: I prefer to pass this down as a dependency, so it's clear what this code actually depends on, and it's easier to control how your code works and what it interacts with. The global registry is nice for convenience, but I think global state is harmful to long-term readability and maintainability, and I lean towards removing it completely.

src/consensus/fork_choice.zig
```diff
@@ -5026,3 +5107,40 @@ pub fn testEpochStakes(
...
+        const HistogramKind = enum {
+            fork_choice_update_interval,
+        };
+
+        const time_interval_buckets = &.{ 0.001, 0.01, 0.1, 1, 10, 100, 1000 }; // seconds
+
+        return switch (@field(HistogramKind, field_name)) {
+            .fork_choice_update_interval => time_interval_buckets,
+        };
+    }
```
dnut, line 5195: FYI you can get the same buckets like this too. Since there's only one field, you don't need a function. I'm also fine with the current approach if you prefer that.

```suggestion
pub const histogram_buckets = sig.prometheus.histogram.exponentialBuckets(10, -3, 3);
```

src/consensus/replay_tower.zig
```diff
@@ -4608,3 +4701,42 @@ fn genStakes(
...
+        }
+    }
+
+    pub fn trackStaleVoteRejected(self: *const ReplayTowerMetrics) void {
+        self.stale_votes_rejected.inc();
+    }
+
+    pub fn trackStrayRestoredSlot(self: *const ReplayTowerMetrics) void {
+        self.stray_restored_slots.inc();
+    }
```
dnut: Is there a benefit of this indirection? I'd lean towards directly interacting with the metrics at the callsite.

src/consensus/replay_tower.zig
```diff
@@ -4608,3 +4701,42 @@ fn genStakes(
...
+                self.fork_switching_successes.inc();
+            },
+            .same_fork => {
+                // No metrics needed for same fork
+            },
+            .failed_switch_threshold, .failed_switch_duplicate_rollback => {
+                self.fork_switching_failures.inc();
+            },
+        }
+    }
```
dnut: You could use VariantCounter for this and remove this function. This will group the metrics all with the same name and differentiate them using labels, sort of like a histogram.

```zig
switch_fork_decision: VariantCounter(@typeInfo(SwitchForkDecision).@"union".tag_type.?),
```

The `typeInfo` stuff can be removed if VariantCounter is modified to handle tagged unions, which shouldn't be too complicated. Ideally it would look more like this:

```zig
switch_fork_decision: VariantCounter(SwitchForkDecision),  
```

This would simply track each variant separately, which is semantically different than what you're doing here, since you've added filtering and aggregation logic. With VariantCounter, you would set up your grafana dashboard to filter and aggregate the data instead. I'd lean towards that approach actually. Less opinionated raw data allows for more flexible analysis later. But if it's too much hassle to fix up the dashboard, don't worry about it.

src/consensus/replay_tower.zig
```diff
@@ -4021,6 +4048,71 @@ test "unconfirmed duplicate slots and lockouts for non heaviest fork" {
...
+    // Should return VoteTooOld error
+    try std.testing.expectError(error.VoteTooOld, result);
+
+    // Check that stale vote rejected counter was incremented
+    const final_stale_votes = tower.metrics.stale_votes_rejected.get();
+    try std.testing.expectEqual(initial_stale_votes + 1, final_stale_votes);
+}
+
+test "test trackForkSwitchingMetrics - switch_proof" {
+    var metrics = try ReplayTowerMetrics.init();
```
dnut: Validating metrics in a global registry is problematic. All the tests share the same registry so they're incrementing the same counters and will interfere with each other.

The only reason your tests are passing is because they are checking the state before and after. The reason you need to do this is because they're all sharing the same state. If they didn't share the same registry, you could use a simpler assertion that the value is a specific hardcoded number.

If zig ever upgrades to a multi-threaded test runner (I hope this happens soon), all of these tests will break. Currently they only work because they are executed sequentially, not concurrently.

src/consensus/fork_choice.zig
dadepo: [cb0db31](https://github.com/Syndica/sig/pull/931/commits/cb0db31b2f2b3f5564f5a0cdd8a04134ed4a1890)

src/consensus/replay_tower.zig
dadepo: [d25bb09](https://github.com/Syndica/sig/pull/931/commits/d25bb09710e48f161eaa78a6c04bc9f4c605267b)

src/consensus/replay_tower.zig
dadepo: [a555a4b](https://github.com/Syndica/sig/pull/931/commits/a555a4b2a65b273399dd08f864b861ceeb6a5d34)

src/consensus/replay_tower.zig
dadepo: [9939a35](https://github.com/Syndica/sig/pull/931/commits/9939a359978937ce4ec5c060f875d015fbf8d166)

dnut: Why did you remove the tests?

src/prometheus/variant_counter.zig
```diff
@@ -21,10 +27,17 @@ pub fn VariantCounter(comptime EnumOrError: type) type {
...
         pub const Type = EnumOrError;
 
-        pub fn observe(self: *Self, err: EnumOrError) void {
-            _ = self.counts[indexer.index(err)].fetchAdd(1, .monotonic);
+        pub fn observe(self: *Self, value: T) void {
+            const tag_value: EnumOrError = switch (@typeInfo(T)) {
+                .@"union" => @as(EnumOrError, value),
+                else => value,
+            };
+            _ = self.counts[indexer.index(tag_value)].fetchAdd(1, .monotonic);
```
dnut: this cast is automatic
```suggestion
            _ = self.counts[indexer.index(value)].fetchAdd(1, .monotonic);
```

src/consensus/fork_choice.zig
dnut: This still uses the global registry in tests which is a problem for the reasons I described in the other comment.

src/prometheus/variant_counter.zig
dadepo: [980afc2](https://github.com/Syndica/sig/pull/931/commits/980afc25cad7745868be43b19a3f7568ab16442c)

src/consensus/replay_tower.zig
dadepo: The tests were actually not properly written - ie they are missing asserts. Added them in a bit of a haste to satisfy codecov, but with the recent changes, the test coverage is fine, decided to remove them all. I can add them back if you consider it a plus to have these sets of tests.

src/consensus/fork_choice.zig
dadepo: [01e66d0](https://github.com/Syndica/sig/pull/931/commits/01e66d0cd08a51b6cb05d4d763f2097d736c1507)

src/consensus/replay_tower.zig
dadepo: Codecov took a dip again (not sure why) added them back [4a31020](https://github.com/Syndica/sig/pull/931/commits/4a31020065c659913cca18f7b80dfc32cb5bcdf0)

dnut: These tests as they are written now are not actually asserting anything meaningful. It just sets a const directly in the test, and then asserts the const is equal to the value that it was just set to. It's not actually making an assertion about any code outside the test itself.

The previous tests were actually asserting something about the metrics being applied correctly when you called a particular function.

dadepo: This is an example of the initial test

```zig
test "test switch_fork_decision - switch_proof" {
    var metrics = try ReplayTowerMetrics.init(sig.prometheus.globalRegistry());

    const decision = SwitchForkDecision{ .switch_proof = Hash.ZEROES };
    metrics.switch_fork_decision.observe(decision);
}
```

the updated one:

```zig
test "switch_fork_decision - switch_proof" {
    var registry = sig.prometheus.Registry(.{}).init(std.testing.allocator);
    defer registry.deinit();
    var metrics = try ReplayTowerMetrics.init(&registry);

    const decision = SwitchForkDecision{ .switch_proof = Hash.ZEROES };
    metrics.switch_fork_decision.observe(decision);

    try std.testing.expectEqual(Hash.ZEROES, decision.switch_proof); // missing before
}
``` 

It is basically the same thing but now with an assertion added. 

Can you help clarify what you are referring to, because it looks to me like the same set of tests but with assertion added

dnut: If you click the link at the top of this thread you'll see the test I'm talking about. This is the code you had in the pr when I commented:

```zig
test "test trackForkSwitchingMetrics - switch_proof" {
    var metrics = try ReplayTowerMetrics.init();

    const initial_successes = metrics.fork_switching_successes.get();

    metrics.trackForkSwitchingMetrics(.{ .switch_proof = Hash.ZEROES });

    const final_successes = metrics.fork_switching_successes.get();
    try std.testing.expectEqual(initial_successes + 1, final_successes);
```

I don't think we need tests like this, it just seems better than the alternatives since at least it asserts something meaningful.

dnut (dismissed):

dnut (approved):

PR #933, opened 2025-09-04
refactor(accountsdb): AAdd more testing for public API

InKryption:
Also add a a slew of refactors and additions.


InKryption:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Syndica/sig/933?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#933** <a href="https://app.graphite.dev/github/pr/Syndica/sig/933?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Syndica/sig/933?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#930** <a href="https://app.graphite.dev/github/pr/Syndica/sig/930?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/accountsdb/db.zig
```diff
@@ -178,6 +182,31 @@ pub const AccountsDB = struct {
...
+            return .{
+                .allocator = allocator,
+                .logger = logger,
+                .snapshot_dir = snapshot_dir,
+                .index_allocation = .ram,
+                .number_of_index_shards = 1,
+                .geyser_writer = null,
+                .gossip_view = null,
+                .buffer_pool_frames = 2,
+            };
```
Sobeston, line 209: 2 frames is a bit low, you would be unable to read any larger accounts. I'd recommend going for something like 128*1024 (64MiB, rather than 1GiB). We should maybe just lower it overall? It costs init time which is annoying, we can tune it back up later for mainnet.

src/accountsdb/db.zig
```diff
@@ -2056,34 +2092,44 @@ pub const AccountsDB = struct {
...
         if (self.geyser_writer) |geyser_writer| {
-            try geyser_writer.writePayloadToPipe(.{
+            geyser_writer.writePayloadToPipe(.{
                 .AccountPayloadV1 = .{
                     .accounts = &.{duplicated},
                     .pubkeys = &.{pubkey},
                     .slot = slot,
                 },
-            });
+            }) catch return error.GeyserWriteError;
```
Sobeston: Hm, might be a good idea to log the error if we get one? I can see this making debugging a bit trickier (i.e. your error message has less info than before).

src/accountsdb/db.zig
```diff
@@ -2151,7 +2200,15 @@ pub const AccountsDB = struct {
...
-        try self.expandSlotRefsAndInsert(slot, &.{pubkey});
+        self.expandSlotRefsAndInsert(slot, &.{pubkey}) catch |err| switch (err) {
+            error.OutOfMemory,
+            error.AllocTooBig,
+            error.AllocFailed,
+            error.AttemptedZeroAlloc,
+            error.CollapseFailed,
+            error.InsertIndexFailed,
+            => return error.OutOfMemory,
+        };
```
Sobeston, line 2099: As above, would be good to log the internal error to preserve info

src/accountsdb/db.zig
InKryption: This is just modelled off the `initForTest` values function. Didn't realize it put a limitation on reads; maybe it makes sense to make it a parameter as well so it can be tuned for each test.

src/accountsdb/db.zig
InKryption: I think in practice it isn't actually an issue, since the backtrace will show what the original error was, but a log can't hurt either.

src/accountsdb/db.zig
InKryption: 79e636c28f69812bc390df94dc92e914d0dd41f7

src/accountsdb/db.zig
InKryption: fd28289b08de31705171657e3be56218531d76a2

src/accountsdb/db.zig
InKryption: fd28289b08de31705171657e3be56218531d76a2

Sobeston (dismissed):

yewman (dismissed):

Sobeston (approved):

src/accountsdb/db.zig
```diff
@@ -178,6 +182,32 @@ pub const AccountsDB = struct {
...
+
+        pub const Index = union(AccountIndex.AllocatorConfig.Tag) {
+            ram,
+            disk,
+            parent: *sig.accounts_db.index.ReferenceAllocator,
+        };
+
+        /// Minimal configuration, mainly for tests.
+        /// Uses ram for index allocation, one index shard, no gossip view, no geyser writer, and 2 buffer pool frames.
+        pub fn minimal(
```
dnut, line 194: Did you see we already have two functions with similar aims:

- loadTestAccountsDBFromSnapshot: for tests that require a snapshot
- initForTest: for tests that do not require a snapshot and just require a very lean accountsdb

I think for your usages in the tests, `initForTest` accomplishes the same goal.

src/accountsdb/db.zig
```diff
@@ -824,9 +850,13 @@ pub const AccountsDB = struct {
...
             }
 
-            self.largest_file_id = FileId.max(self.largest_file_id, file_id);
-            _ = self.largest_rooted_slot.fetchMax(slot, .release);
-            self.largest_flushed_slot.store(self.largest_rooted_slot.load(.acquire), .release);
+            self.largest_file_id = .max(self.largest_file_id, file_id);
+            {
+                const max_slots, var max_slots_lg = self.max_slots.writeWithLock();
+                defer max_slots_lg.unlock();
+                max_slots.rooted = @max(max_slots.rooted orelse 0, slot);
```
dnut, line 857: Is 0 correct here? What if you're starting from a snapshot that's not at slot 0? Maybe the starting slot should be specified when AccountsDB is initialized, and should never be allowed to be null.

src/accountsdb/db.zig
```diff
@@ -2035,7 +2084,18 @@ pub const AccountsDB = struct {
...
+        self.expandSlotRefsAndInsert(slot, &.{pubkey}) catch |err| switch (err) {
+            error.OutOfMemory,
+            error.AllocTooBig,
+            error.AllocFailed,
+            error.AttemptedZeroAlloc,
+            error.CollapseFailed,
+            error.InsertIndexFailed,
+            => |e| {
+                self.logger.err().logf("expandSlotRefsAndInsert: {s}", .{@errorName(e)});
+                return error.OutOfMemory;
```
dnut, line 2095: OutOfMemory does not seem like an accurate description of these:
```zig
error.AttemptedZeroAlloc,  
error.CollapseFailed,  
error.InsertIndexFailed,
```

src/accountsdb/db.zig
```diff
@@ -4763,3 +4827,601 @@ test "loadAndVerifyAccountsFiles ref manager expand" {
...
+        simple.largest_rooted_slot = largest_rooted_slot;
+    }
+    {
+        const max_slots, var max_slots_lg = real.max_slots.writeWithLock();
+        defer max_slots_lg.unlock();
+        max_slots.rooted = largest_rooted_slot;
+    }
+}
+
+test "insertion basic" {
```
dnut: Since these tests define the rules of the account store abstractions, what do you think about putting them in account_store.zig?

src/accountsdb/db.zig
```diff
@@ -4763,3 +4827,601 @@ test "loadAndVerifyAccountsFiles ref manager expand" {
...
+    var prng_state: std.Random.Xoshiro256 = .init(346715);
+    const prng = prng_state.random();
+
+    var tmp_dir = std.testing.tmpDir(.{});
+    defer tmp_dir.cleanup();
+
+    var simple_state: sig.accounts_db.ThreadSafeAccountMap = .init(allocator);
+    defer simple_state.deinit();
+
+    var real_state: AccountsDB = try .init(.minimal(allocator, .noop, tmp_dir.dir, null));
```
dnut: I suggest using `.FOR_TESTS` as the logger in tests, unless it's a negative test where warn or error messages are expected. If the test is producing unpexected warn or error messages, it's nice to see that visibly when you run the tests.

src/accountsdb/manager.zig
```diff
@@ -175,32 +203,69 @@ pub const Manager = struct {
...
+            // self.db.logger.debug().logf("shrink_results: {any}", .{shrink_results});
+
+            // delete any empty account files
+            defer self.delete_account_files.clearRetainingCapacity();
+            try deleteAccountFiles(self.db, self.delete_account_files.keys());
+        }
+    }
+
+    /// Returns the latest full snapshot slot by the end of the attempt.
+    pub fn tryGenerateSnapshots(self: *Manager) !void {
```
dnut: inaccurate doc comment

dnut (changes requested):
very nice addition of test coverage. there are a few minor things i might change though

src/accountsdb/db.zig
InKryption: `CollapseFailed` seems pretty accurately described by `OutOfMemory`, given that the condition under which it occurs is `!isPossibleToAllocateUnsafe(n);`.
For `AttemptedZeroAlloc`, I find this is just a silly error returned by `RecycleBuffer`, I don't see the point in distinguishing it in the error set - on that note, I think I'll actually just remove it.
For `InsertIndexFailed`, fair enough.

InKryption: 82ed53189804fc54f774e58be9b597cd19159dbc

src/accountsdb/db.zig
InKryption: I guess I don't have a super strong preference, but should be noted that while these tests are defining the contract of the account store abstraction, they're also testing aspects of the internal workings of accountsdb as it is interacted with through the interface, so it might be a bit odd to put tests containing that over there.

src/accountsdb/db.zig
InKryption: I dislike `initForTest` with how it bundles in the creation of the `tmp_dir`, and hardcodes a bunch of options, which make it impossible to change those options during the debugging process without either switching to a more manual initialization, or changing `initForTest`. I would rather delete `initForTest` and replace it with `minimal`.

src/accountsdb/db.zig
InKryption: 0 is fine because either `slot` is 0, in which case that would be the value that ends up getting set regardless, or `slot > 0`, in which case `rooted` won't get set to 0.

src/accountsdb/db.zig
InKryption: 76cb4d2212dac47c540ade7f01d57fddfa6b8bbf

src/accountsdb/manager.zig
InKryption: f73d52bb2e12cffdd1385bc5a67ebf243f49f89d

src/accountsdb/db.zig
dnut: nice, i agree about the empty alloc

src/accountsdb/db.zig
dnut: > testing aspects of the internal workings of accountsdb

Like what? Are you just talking about how it returns the root version of an account whenever an older slot is requested? Maybe that actually should be expected of any account store.

InKryption: Moreso referring to the `"put and get zero lamports before & after cleanup"`, which uses some of the same test helpers, in addition to `expectDbUnrootedPubkeysInSlot`, which directly checks the internal state of accountsdb.

Agree on the point about expecting accountstore to return the correct account based on rootedness of a slot.

dnut: I would probably put all the tests except that one in account_store.zig. I understand that means the shared helpers will need to be pub but I don't see that as a problem. If they all need to be in the same file, I think I would personally still choose account_store.zig. But it's not really a big deal. Looks like you also have some conflicts to resolve, so you can throw in this change as well if you're OK with it, and then I'll approve whatever you decide.

InKryption: Realized that actually half of them end up touching the internal state, so may as well just chuck them all in account_store if that's your view on it; maybe it would make sense to expose some concept of "rooting" through the interface, since it's so important to its behavior? Big question mark on that one.

f37466aac6df3ab11ca9de663d53a8ded36327c4

dnut: > it would make sense to expose some concept of "rooting" through the interface,

Actually I already made that suggestion to Sebastian in his PR.

dnut (dismissed):

yewman (approved):

dnut (approved):

PR #934, opened 2025-09-04
fix(consenus) Get vote accounts from stakes cache for computed bank state

dadepo:
The initial implementation was getting the vote accounts needed to calculate the compute bank state from the epoch_stakes. Going through the Agave implementation again, made the realization this should be retrieved from the stake cache.

### DIFF
--- a/src/replay/consensus.zig
+++ b/src/replay/consensus.zig
@@ -575,7 +575,7 @@ fn computeBankStats(
         if (!fork_stat.computed) {
             // TODO Self::adopt_on_chain_tower_if_behind
             // Gather voting information from all vote accounts to understand the current consensus state.
-            const slot_info_for_stakes = slot_tracker.get(slot) orelse return error.MissingSlots;
+            const slot_info_for_stakes = slot_tracker.get(slot) orelse return error.MissingSlot;
 
             const computed_bank_state = blk: {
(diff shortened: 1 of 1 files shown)


dnut (dismissed):

src/replay/consensus.zig
```diff
@@ -571,22 +571,28 @@ fn computeBankStats(
...
-                allocator,
-                .from(logger),
-                &my_vote_pubkey,
-                slot,
-                &epoch_stakes.stakes.vote_accounts.vote_accounts,
-                ancestors,
-                progress,
-                latest_validator_votes,
-            );
+            const slot_info_for_stakes = slot_tracker.get(slot) orelse return error.MissingSlots;
```
InKryption: Tiny nit: why `MissingSlots` instead of `MissingSlot`? Think it would be better to re-use the error used just above.

InKryption (dismissed):

src/replay/consensus.zig
dadepo: [93aa06a](https://github.com/Syndica/sig/pull/934/commits/93aa06a273d1f5d903061788b7712bb750af9aae)

InKryption (approved):

dnut (approved):

PR #937, opened 2025-09-05
fix(replay): accumulate transaction fees in confirm slot

yewman:
Fixes the slot hash mismatch. 

The issue was we were not accumulating the transaction fees in confirm slot. This results no correctly distributing the transaction fees to the fee collector and thus a difference in account state. 

This PR adds fee accumulation to the confirm slot method


src/replay/freeze.zig
```diff
@@ -191,38 +192,45 @@ fn distributeTransactionFees(
...
+    }
+
+    var deposit, const burn = blk: {
+        if (collected_transaction_fees == 0) {
+            break :blk .{ 0, 0 };
+        }
+        const burned = collected_transaction_fees * 50 / 100;
+        break :blk .{ collected_transaction_fees - burned, burned };
     };
+    deposit +|= collected_priority_fees;
```
dnut: I think this is much more confusing than the old logic which accomplished the same thing in three lines of code.

I don't see any reason to special case 0 in either of these places when normal arithmetic already handles 0 correctly. The condition on depositFees makes sense because that function has a meaningful performance impact, but the other checks are redundant and make the code harder to understand. Also breaking from the block in multiple places and then mutating the data later, it feels convoluted compared to the direct calculation we had already.

src/replay/freeze.zig
```diff
@@ -191,38 +192,45 @@ fn distributeTransactionFees(
...
+        // TODO: record rewards returned by depositFees
+        _ = depositFees(
+            allocator,
+            account_store,
+            account_reader,
+            rent,
+            slot,
+            collector_id,
+            deposit,
+        ) catch break :blk;
```
dnut: I prefer to be explicit about what errors are being ignored because it's less likely to cause mistakes. For example, this hides errors that occur inside accountsdb, which normally we would propagate up.

src/replay/freeze.zig
```diff
@@ -234,16 +242,23 @@ fn tryPayoutFees(
...
-    fee_collector_account.lamports = std.math.add(u64, fee_collector_account.lamports, payout) catch
+
+    const pre_state = RentCollector.getAccountRentState(rent, account.lamports, account.data.len);
+    account.lamports = std.math.add(u64, account.lamports, fees) catch {
         return error.LamportOverflow;
-    if (!rent.isExempt(fee_collector_account.lamports, fee_collector_account.data.len)) {
+    };
+
+    const post_state = RentCollector.getAccountRentState(rent, account.lamports, account.data.len);
+    if (!core.rent_collector.RentCollector.transitionAllowed(pre_state, post_state)) {
```
dnut: What's the benefit of this more complicated approach? Is there a semantic difference?

Rent has been eliminated from solana so I think we can simplify a lot of this stuff.

src/core/rent_collector.zig
```diff
@@ -157,9 +157,9 @@ pub const RentCollector = struct {
         rent_exempt,
     };
 
-    pub fn getAccountRentState(self: RentCollector, lamports: u64, data_len: usize) RentState {
+    pub fn getAccountRentState(rent: Rent, lamports: u64, data_len: usize) RentState {
```
dnut: I like this simplification but maybe it should be a method on Rent?

src/core/leader_schedule.zig
```diff
@@ -162,12 +162,35 @@ pub const LeaderSchedule = struct {
...
+
+        for (vote_accounts.keys(), vote_accounts.values()) |key, value| {
+            try stakes.put(allocator, key, value.stake);
+        }
+
+        const vote_keyed = try fromStakedNodes(allocator, epoch, slots_in_epoch, &stakes);
+
+        for (vote_keyed) |*pk| {
+            const npk = vote_accounts.get(pk.*) orelse unreachable;
+            pk.* = npk.account.state.node_pubkey;
```
dnut: Can we avoid abbreviations? I think they make the code confusing. I can't figure out what `npk` means.

src/core/rent_collector.zig
Rexicon226: I agree^

src/replay/freeze.zig
yewman: 8500da92

src/replay/freeze.zig
yewman: 8500da92

src/replay/freeze.zig
yewman: 8500da92

src/core/rent_collector.zig
yewman: 8500da92

This change is no longer required for slot hash mismatch and thus I would rather leave it for a follow up PR as it is relatively low priority. Also such a change should be accompanied with moving Rent from sysvars to rent_collector (similar to EpochSchedule)

src/core/leader_schedule.zig
yewman: 8500da92

PR #946, opened 2025-09-10
fix(replay): get the accountsdb manager working

Sobeston:
Will follow up with another PR for snapshot generation

### DIFF
--- a/src/accountsdb/manager.zig
+++ b/src/accountsdb/manager.zig
@@ -47,11 +47,15 @@ pub fn onSlotRooted(
 ) !void {
     const zone = tracy.Zone.init(@src(), .{ .name = "onSlotRooted" });
     defer zone.deinit();
-    defer tracy.frameMarkNamed("rooted slots flushed");
+    errdefer zone.color(0xFF0000);
 
     db.logger.info().logf("flushing slot {} to disk", .{newly_rooted_slot});
+
     var failed: bool = false;
-    defer if (!failed) db.logger.info().logf("successfully flushed slot {} to disk", .{newly_rooted_slot});
+    defer if (!failed) {
+        db.logger.info().logf("successfully flushed slot {} to disk", .{newly_rooted_slot});
+        tracy.frameMarkNamed("rooted slots flushed");
+    };
     errdefer |err| {
         db.logger.err().logf("failed to flushed slot {} to disk, err: {}", .{ newly_rooted_slot, err });
         failed = true;
@@ -59,13 +63,15 @@ pub fn onSlotRooted(
 
     // make a new AccountFile for our newly rooted slot, and "flush" the data to it
     const file_id = new_id: {
-        break :new_id flushSlot(db, newly_rooted_slot) catch |err| {
-            // flush fail = loss of account data on slot -- should never happen
-            db.logger.err()
-                .logf("flushSlot {d} error: {s}", .{ newly_rooted_slot, @errorName(err) });
-            return err;
-        };
+        // flush fail = loss of account data on slot -- should never happen
+        errdefer |err| db.logger.err()
+            .logf("flushSlot {d} error: {s}", .{ newly_rooted_slot, @errorName(err) });
+
+        break :new_id try flushSlot(db, newly_rooted_slot);
     };
+
+    // TODO: audit this and related fields - not sure why this is an atomic to begin with
+    db.largest_flushed_slot.store(newly_rooted_slot, .monotonic);
 
     // when we last made a full snapshot, and when we last made an incremental snapshot (relative to
     // the last full snapshot).
@@ -77,9 +83,12 @@ pub fn onSlotRooted(
         break :blk .{ snapshot_gen_info.full.slot, incremental_info.slot };
     };
 
-    const make_full_snapshot = newly_rooted_slot >=
+    // TODO: this should be configurable - leaving disabled until snapshot generation is fixed
+    const snapshot_generation_enabled = false;
+
+    const make_full_snapshot = snapshot_generation_enabled and newly_rooted_slot >=
         latest_full_snapshot_slot + config.slots_per_full_snapshot;
-    const make_inc_snapshot = newly_rooted_slot >=
+    const make_inc_snapshot = snapshot_generation_enabled and newly_rooted_slot >=
         latest_inc_snapshot + latest_full_snapshot_slot + config.slots_per_incremental_snapshot;
 
     // TODO: might be a good idea to move snapshot creation to another thread
@@ -122,7 +131,7 @@ pub fn onSlotRooted(
                     prev.full.slot + config.slots_per_full_snapshot);
             }
 
-            gen_info.full = .{
+            gen_info.* = .{
                 .full = .{
                     .capitalization = result.capitalization,
                     .hash = result.hash,
@@ -155,17 +164,17 @@ pub fn onSlotRooted(
                 db.latest_snapshot_gen_info.writeWithLock();
             defer gen_info_lg.unlock();
 
-            // we can't make an incremental snapshot without having previously made a full snapshot?
-            const gen_info = maybe_gen_info orelse unreachable;
+            const gen_info = &((maybe_gen_info.*) orelse
+                @panic("can't make an incr snapshot without previously making a full snapshot"));
 
             if (gen_info.inc) |prev_inc| {
                 std.debug.assert(newly_rooted_slot >=
-                    prev_inc + config.slots_per_incremental_snapshot);
+                    prev_inc.slot + config.slots_per_incremental_snapshot);
             }
 
             gen_info.inc = .{
-                .capitalization = result.capitalization,
-                .hash = result.hash,
+                .capitalization = result.incremental_capitalization,
+                .hash = result.incremental_hash,
                 .slot = newly_rooted_slot,
             };
         }
@@ -174,9 +183,9 @@ pub fn onSlotRooted(
     // TODO: change APIs of {clean, shrink, delete}AccountFiles to take single accounts, and stop
     // doing this.
     // TODO: could use an arena allocator for this whole function
-    const to_shrink: std.AutoArrayHashMapUnmanaged(FileId, void) = .init(allocator, .{}, .{});
+    var to_shrink: std.AutoArrayHashMapUnmanaged(FileId, void) = try .init(allocator, &.{}, &.{});
     defer to_shrink.deinit(allocator);
-    const to_delete: std.AutoArrayHashMapUnmanaged(FileId, void) = .init(allocator, .{}, .{});
+    var to_delete: std.AutoArrayHashMapUnmanaged(FileId, void) = try .init(allocator, &.{}, &.{});
     defer to_delete.deinit(allocator);
 
     const clean_result = try cleanAccountFiles(
@@ -209,6 +218,7 @@ pub fn onSlotRooted(
 fn flushSlot(db: *AccountsDB, slot: Slot) !FileId {
     const zone = tracy.Zone.init(@src(), .{ .name = "accountsdb flushSlot" });
     defer zone.deinit();
+    errdefer zone.color(0xFF0000);
 
     var timer = try sig.time.Timer.start();
 
--- a/src/accountsdb/manager.zig
+++ b/src/accountsdb/manager.zig
@@ -49,6 +49,15 @@ pub fn onSlotRooted(
     defer zone.deinit();
     errdefer zone.color(0xFF0000);
 
+    {
+        var max_slots, var max_slots_lg = db.max_slots.writeWithLock();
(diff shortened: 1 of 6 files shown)


src/accountsdb/manager.zig
```diff
@@ -20,260 +20,201 @@ const AccountFile = sig.accounts_db.accounts_file.AccountFile;
...
     }
-};
+
+    // TODO: change APIs of {clean, shrink, delete}AccountFiles to take single accounts, and stop
+    // doing this.
+    // TODO: could use an arena allocator for this whole function
+    var to_shrink: std.AutoArrayHashMapUnmanaged(FileId, void) = try .init(allocator, &.{}, &.{});
+    defer to_shrink.deinit(allocator);
+    var to_delete: std.AutoArrayHashMapUnmanaged(FileId, void) = try .init(allocator, &.{}, &.{});
+    defer to_delete.deinit(allocator);
```
InKryption:
```suggestion
    var to_shrink: std.AutoArrayHashMapUnmanaged(FileId, void) = .empty;
    defer to_shrink.deinit(allocator);
    var to_delete: std.AutoArrayHashMapUnmanaged(FileId, void) = .empty;
    defer to_delete.deinit(allocator);
```

src/accountsdb/manager.zig
```diff
@@ -20,260 +20,201 @@ const AccountFile = sig.accounts_db.accounts_file.AccountFile;
...
-    defer manager.deinit(db.allocator);
-    try manager.run(db.allocator);
-}
+pub const Config = struct {
+    /// default copied from Agave
+    slots_per_full_snapshot: u64 = 100_000,
+    /// default copied from Agave
+    slots_per_incremental_snapshot: u64 = 100,
+    /// 0 => no thread spawning
+    zstd_nb_workers: u31 = 0,
```
yewman, line 36: Is there a specific reason for u31 here?

yewman (commented):
Looks good. I'd like to take another look after rebasing on main as a lot of diff seems to be from dnut/replay/simple

src/accountsdb/manager.zig
Sobeston: It's an i32 in the library, and negative doesn't make sense, so u31 is pretty good

src/replay/service.zig
```diff
@@ -830,6 +832,18 @@ fn bypassConsensus(state: *ReplayState, results: []const ReplayResult) !bool {
         state.logger.info().logf("rooting slot with SlotTree.reRoot: {}", .{new_root});
         slot_tracker.root = new_root;
         slot_tracker.pruneNonRooted(state.allocator);
+        if (state.account_store == .accounts_db) {
+            // TODO: get this value from the right place
+            const lamports_per_signature = 5000;
+
+            try sig.accounts_db.manager.onSlotRooted(
```
dnut: Could you make this an interface method? Replay does not depend on AccountsDB, and I'd like to keep it that way.

src/replay/service.zig
```diff
@@ -830,6 +832,18 @@ fn bypassConsensus(state: *ReplayState, results: []const ReplayResult) !bool {
         state.logger.info().logf("rooting slot with SlotTree.reRoot: {}", .{new_root});
         slot_tracker.root = new_root;
         slot_tracker.pruneNonRooted(state.allocator);
+        if (state.account_store == .accounts_db) {
+            // TODO: get this value from the right place
+            const lamports_per_signature = 5000;
```
dnut: this is in slot constants

src/accountsdb/manager.zig
```diff
@@ -20,260 +20,201 @@ const AccountFile = sig.accounts_db.accounts_file.AccountFile;
...
-                if (unrooted_slot.* <= root_slot) {
-                    // NOTE: need to flush all references <= root_slot before we call clean
-                    // or things break by trying to clean unrooted references
-                    // NOTE: this might be too much in production, not sure
-                    try self.flush_slots.append(allocator, unrooted_slot.*);
-                }
-            }
-        }
+        break :new_id try flushSlot(db, newly_rooted_slot);
+    };
```
dnut: This code is kind of convoluted


```suggestion
    const file_id = flushSlot(db, newly_rooted_slot) catch |err| {
        db.logger.err().logf("flushSlot {d} error: {s}", .{ newly_rooted_slot, @errorName(err) });
        return err;
    };
```
This fails to compile, which I can't figure out. My only guess is that it's a compiler bug.


Actually isn't the log message redundant with the errdefer right above this?


```suggestion
    const file_id = try flushSlot(db, newly_rooted_slot);
```

src/accountsdb/manager.zig
```diff
@@ -20,260 +20,201 @@ const AccountFile = sig.accounts_db.accounts_file.AccountFile;
...
+        const incremental_info = snapshot_gen_info.inc orelse
+            break :blk .{ snapshot_gen_info.full.slot, 0 };
+        break :blk .{ snapshot_gen_info.full.slot, incremental_info.slot };
+    };
 
-        if (must_flush_slots) {
-            self.db.logger.debug()
-                .logf("flushing slots: min: {}...{}", std.mem.minMax(Slot, self.flush_slots.items));
+    // TODO: this should be configurable - leaving disabled until snapshot generation is fixed
+    const snapshot_generation_enabled = builtin.is_test;
```
dnut, line 117: I thought you said you got this working in the current PR?

src/accountsdb/manager.zig
```diff
@@ -20,260 +20,201 @@ const AccountFile = sig.accounts_db.accounts_file.AccountFile;
...
-            while (unrooted_slot_iter.next()) |unrooted_slot| {
-                if (unrooted_slot.* <= root_slot) {
-                    // NOTE: need to flush all references <= root_slot before we call clean
-                    // or things break by trying to clean unrooted references
-                    // NOTE: this might be too much in production, not sure
-                    try self.flush_slots.append(allocator, unrooted_slot.*);
-                }
-            }
-        }
+        break :new_id try flushSlot(db, newly_rooted_slot);
```
dnut: What happens if the caller attempts to root a slot <= the highest rooted slot? Previously this was checked because the manager didn't know when it was notified about a rooted slot. But now I think you've baked in an assumption that the caller will never do this. So what's the behavior if the caller does that? Maybe we should have a check and:
- return error if the slot is less than the last flushed slot
- either do nothing or return error if the slot is equal to the last flushed slot

src/accountsdb/manager.zig
```diff
@@ -20,260 +20,201 @@ const AccountFile = sig.accounts_db.accounts_file.AccountFile;
...
-                    self.db.logger.err()
-                        .logf("flushing slot {d} error: {s}", .{ flush_slot, @errorName(err) });
-                    continue;
-                };
-                self.unclean_account_files.appendAssumeCapacity(unclean_file_id);
-                largest_flushed_slot = @max(largest_flushed_slot, flush_slot);
-            }
-            _ = self.db.largest_flushed_slot.fetchMax(largest_flushed_slot, .seq_cst);
-        }
+    // TODO: might be a good idea to move snapshot creation to another thread
```
dnut, line 125: Yeah, otherwise you're blocking replay. This seems pretty important actually.

src/accountsdb/manager.zig
```diff
@@ -20,260 +20,201 @@ const AccountFile = sig.accounts_db.accounts_file.AccountFile;
...
-            // self.db.logger.debug().logf("shrink_results: {any}", .{shrink_results});
+            const maybe_gen_info: *?SnapshotGenerationInfo, var gen_info_lg =
+                db.latest_snapshot_gen_info.writeWithLock();
+            defer gen_info_lg.unlock();
 
-            // delete any empty account files
-            defer self.delete_account_files.clearRetainingCapacity();
-            try deleteAccountFiles(self.db, self.delete_account_files.keys());
+            const gen_info = &((maybe_gen_info.*) orelse
+                @panic("illegal state - snapshot_gen_info (previously non-null) is now null"));
```
dnut, line 202: When could this happen? If this function is called twice concurrently?

src/accountsdb/manager.zig
Sobeston: Snapshot generation *is* working but
1) I haven't confirmed the validity of the snapshots themselves
2) This isn't useful for us to do right now

src/accountsdb/manager.zig
Sobeston: Yeah definitely. How about we put it onto a thread once we enable it properly?

src/accountsdb/manager.zig
Sobeston: I'm not aware of any place this could happen - just programming defensively here (and preferring panic over unreachable).

src/replay/service.zig
Sobeston: This function handles the flushing of rooted slots to disk (and the associated cleanup) - it only really makes sense to do it on a real accountsdb, this is about AccountFiles.
On another AccountStore, I understand this would be a no-op (unless we later decide we want snapshot generation from an AccountStore).

src/accountsdb/manager.zig
Sobeston: I removed this log, it was redundant

src/accountsdb/manager.zig
Sobeston: 9e9565b37b10a79c897fcffcf2519c9a7b95671c

Decided to log + error

src/replay/service.zig
Sobeston: 7190fdc023c52cdd23b7a0d0cc0dda8ddc4279a9

src/replay/service.zig
dnut: Rooting slots is part of the contract between replay and any account store. If it's a noop then it can be implemented as a noop. But it's not a noop in either of our AccountStore implementations right now.

Any account store will need to purge irrelevant versions of accounts. It's not sustainable to track every single version of every account. This problem is not specific to our AccountsDB implementation, it's a general problem that needs to be addressed in any implementation. So I think it's reasonable for the interface to expose this functionality to keep the memory/storage usage in check. Replay or consensus should be expected to give that information to any account store.

Also, the purpose of rooting slots in an account store extends beyond simply reducing memory/storage usage. It also has semantic impact on the get methods in the interface, determining which versions of the accounts the account store is expected to return.

Sobeston: > Also, the purpose of rooting slots in an account store extends beyond simply reducing memory/storage usage. It also has semantic impact on the get methods in the interface, determining which versions of the accounts the account store is expected to return.

I think this would be as simple as setting largest_rooted_slot in ThreadSafeAccountMap

> Replay or consensus should be expected to give that information to any account store.

Makes sense

> Any account store will need to purge irrelevant versions of accounts. It's not sustainable to track every single version of every account. 

I guess it could later be useful for longer tests? Not so bad to implement

dnut: The interface is for more than just testing, it's for defining the contract of an account store in general, of which I expect to additional implementations that will be used in production

dnut: Wow, nice, you actually implemented cleanup for ThreadSafeAccountMap. Was that actually necessary to get it to satisfy the contract? I thought you could just set the rooted slot and that would be good enough. Regardless, nice work.

Regarding the implementation, I understand what you're doing and it makes sense, but since ThreadSafeAccountMap doesn't have slot-agnostic rooted storage, it's confusing what behavior it actually *should* have when rooting accounts. The approach you took is conservative, leaving things where they are until you're sure they should be deleted, and I think that's reasonable.

But what I've realized when reading this code is that we have a pre-existing bug here. I've been considering various scenarios about the effects of your rooting method, and it leads to problems. But actually this is not due to your changes. I think the `get` method is just wrong. https://github.com/orgs/Syndica/projects/2/views/10?pane=issue&itemId=131987365

dnut: Another approach to rooting might actually fix this pre-existing bug. If you delete every account between the last rooted slot and the currently rooted slot, then it would fix the bug described in that ticket, because it would delete the state from slot 2. This approach relies on the assumption that the caller to onSlotRooted will actually call that function for every single slot that needs to be rooted, and if slots are skipped, then they truly need to be discarded.

dnut: I don't think we need to block the merge of this PR because of this pre-existing bug. But it would be nice to get this fixed soon.

InKryption (approved):

src/replay/service.zig
Sobeston: > Was that actually necessary to get it to satisfy the contract? I thought you could just set the rooted slot and that would be good enough.

Nope, I think you're right, I could have just set the rooted slot.

>  The approach you took is conservative, leaving things where they are until you're sure they should be deleted, and I think that's reasonable.

Yeah this is just a simpler version of what the real accountsdb does

> But it would be nice to get this fixed soon.

Agreed

PR #948, opened 2025-09-12
fix(accountsdb): proper ordering/duration of rwlock acquisitions in putAccount

InKryption:
The original ordering of this function permitted a situation where a lock would be acquired on the pubkey_ref_map, during a period in time in which another function had acquired a lock on the slot_reference_map, before it manages to acquire the lock on the pubkey_ref_map. The exact semantics of where the edge case mutation during this race occurs escapes me, but it isn't hard to see that having such scattered lock ordering would lead to state inconsistencies, and this fixes that, eliminating the invalid fee payer account error.

### DIFF
--- a/src/accountsdb/db.zig
+++ b/src/accountsdb/db.zig
@@ -4845,3 +4845,34 @@ test "loadAndVerifyAccountsFiles ref manager expand" {
     // reference manager.
     try std.testing.expect(ref_capacity / n_slots > accounts_per_file_estimate);
 }
+
+// TODO: this test is dumb, remove it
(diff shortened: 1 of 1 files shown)


InKryption:
Small update: using 4 threads, I did hit an assertion in `processBatch` (can't tell what assertion due to inlining); continuing to track down what it is, but this is still a step forward in progress regardless

InKryption:
Only slightly more information `-Derror-tracing`:
image

dnut (approved):

Sobeston (approved):

PR #951, opened 2025-09-14
fix(tower): Set correct value for threshold_depth/threshold_size

dadepo:
The `canVoteOnCandidateSlot` check was returning `false` and hence the slot the validator should vote: `vote_slot` was always set to null. This was due to the wrong value of 0 being set for `threshold_depth` and `threshold_size`

The `threshold_depth` indicates how far back in the tower’s vote history to look when checking stake thresholds and `threshold_size` indicates the minimum fraction of total stake that must have voted on the fork to permit voting/switching.

So having the values set to 0 was wrong.

This PR fixes that by setting these values to the default just as it is done in Agave.

### DIFF
--- a/src/consensus/replay_tower.zig
+++ b/src/consensus/replay_tower.zig
@@ -214,8 +214,8 @@ pub const ReplayTower = struct {
             .logger = logger,
             .tower = tower,
             .node_pubkey = node_pubkey,
-            .threshold_depth = 0,
-            .threshold_size = 0,
+            .threshold_depth = VOTE_THRESHOLD_DEPTH,
+            .threshold_size = VOTE_THRESHOLD_SIZE,
             .last_vote = VoteTransaction.DEFAULT,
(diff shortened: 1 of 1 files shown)


src/consensus/replay_tower.zig
```diff
@@ -1345,7 +1345,7 @@ pub const ReplayTower = struct {
             } });
 
             // Ignore shallow checks for voting purposes
-            if (vote_depth >= self.threshold_depth) {
+            if (@as(usize, vote_depth) >= self.threshold_depth) {
```
Rexicon226: You don't need this?

dadepo: Nope probably not. Removed [ddf224a](https://github.com/Syndica/sig/pull/951/commits/ddf224a3460965892bb1a2f331f91ad067e8d402)

Rexicon226 (approved):

InKryption (approved):

PR #953, opened 2025-09-14
feature(hash): refactor sha256 hash api

Rexicon226:
This is a WIP PR where I focus on improving performance around sha256 related things. The first commit re-organizes out API a bit to make it easier to streamline the hashing later on. Single hashes do not need to go through a incremental API and instead can be pipelined when done in a loop, something to-be-implemented. And when doing mixins with just two inputs of a known size, we can pipeline that as well, performing much faster hashing than what we currently achieve.


conformance/src/txn_execute.zig
```diff
@@ -1397,25 +1397,22 @@ pub fn hashSlot(
...
-                &signature_count_bytes,
-                blockhash,
-            });
+    var hash = std.crypto.hash.sha2.Sha256.init(.{});
+    hash.update(&sig.core.Hash.ZEROES.data); // no parent lt hash, start with zeroes
+    if (!feature_set.active(.remove_accounts_delta_hash, 0)) {
+        const delta_hash = try freeze.deltaMerkleHash(account_reader, allocator, 0);
+        hash.update(&delta_hash.data);
+    }
+    hash.update(&blockhash.data);
```
yewman, line 1411: Looks like you are no longer hashing in the signature bytes? Also I personally find the original implementation a little clearer but won't die on that hill

yewman (commented):
Looks good to me, just one small comment on a missed signature_count being hashed in conformance and small style preference

conformance/src/txn_execute.zig
Rexicon226: Yeah you're right, my bad. Pushed the fix: be5589e753f2862c3bfe17fec96c1bade71c676b

I'm curious how you see the original one as clearer. My version clearly highlights the important difference between having the feature active and disabled.

src/core/hash.zig
```diff
@@ -36,54 +36,33 @@ pub const Hash = extern struct {
...
-        } else if (@typeInfo(T) == .@"struct") {
-            inline for (data) |val| update(hasher, val);
-        } else if (std.meta.Elem(T) == u8) switch (@typeInfo(T)) {
-            .array => hasher.update(&data),
-            else => hasher.update(data),
-        } else {
-            for (data) |val| update(hasher, val);
-        }
+    /// re-hashes the current hash with the mixed-in byte slice(s).
+    pub fn extend(self: Hash, data: []const u8) Hash {
```
dnut, line 58: I don't see how this API is an improvement. This makes it less ergonomic. What's the benefit? I would prefer to keep it as is.

dnut (dismissed):

conformance/src/txn_execute.zig
dnut: It's a tradeoff.

With your code, yes it highlights the specific difference. I agree this is a benefit. But it does so while making the entire calculation look more convoluted. I have to spend more time to pick out all the components of the hash from the other visual noise.

With the original code, it's easier to understand what the overall hash is composed of. The code is so simple that doing a mental diff is trivial, so highlighting the difference is unnecessary.

The code is pretty easy to follow in either case, so I agree with Harold about not dying on this hill, but I think I lean towards the original.

Rexicon226: I don't see a tradeoff to be honest - the new version is clearly easier to read, is more efficient, and produces less code bloat. But again, no need to continue discussions.

src/core/hash.zig
Rexicon226: The reflection is not necessary, introduces a ton of code bloat, and makes it impossible to optimize hashing in any meaningful way. I believe this PR only proves that the reflection is not necessary, with how small a diff was required to conform to exactly `[]const []const u8` when you just think a bit about the code around callsites. We need to stop abstracting every little thing in the codebase.

dnut: > The reflection is not necessary,

I agree.

> introduces a ton of code bloat

You're saying because there are separate versions of this function for every usage?

Let's say we decided to eliminate these functions and forced every call site to directly use the hasher. I don't see code bloat as a major problem with that. Do you think this function introduces more bloat than inlining the usage of the hasher?

> makes it impossible to optimize hashing in any meaningful way. 

I don't understand why, but regardless I think this only justifies the change when it coincides with optimizations.

>  think a bit about the code around callsites

This is the annoyance I'd like to avoid. I'd rather think about it only the one time that I write this function instead of every time I use it.

Rexicon226: > You're saying because there are separate versions of this function for every usage?

Almost all usages create an anonymous tuple over which the function becomes generalized. So pretty much, yes. 

> Let's say we decided to eliminate these functions and forced every call site to directly use the hasher. I don't see code bloat as a major problem with that. Do you think this function introduces more bloat than inlining the usage of the hasher?

Yes, it does. As I said above, it creates a new version of the function for nearly all our usages, which bloats the binary and makes inlining the hasher implementation itself nearly impossible (LLVM pretty much always refuses to do it, because of how many call sites there are). Using the hasher directly would also cause this problem in most cases, which is specifically why I don't do that. I would honestly prefer it, but since we are looking for maximizing performance, having a single (or a couple in our case, we can have specialized ones for PoH, which I was working on a while ago) entry point, which is then unrolled properly a single time, is optimal.

> This is the annoyance I'd like to avoid. I'd rather think about it only the one time that I write this function instead of every time I use it.

You *want* to be thinking about it each time. Hashing is an expensive operation, and there are a lot of hashes that need to be done over the course of the validator. Increasing friction at those places is a good thing. 

I believe this sort of mindset is one of the main driving factors behind our current performance problems. We are trying to hide everything away behind abstractions; cloning everything, refcounting everything, allocating everything. This leads to the death-by-a-thousand-cuts situation that we are currently in.

We need to start thinking about the performance of every little thing. The validator itself is not a particularly complex piece of software; however, it has a lot of moving parts and nuances. All of them are small and seemingly inexpensive, and they all *need* to happen. Adding them up together leads to a *lot* of performance loss.

dnut: > You want to be thinking about it each time

While I agree with some of your general thoughts about performance, I don't think the argument applies in this scenario. Forcing me to think about whether I need to pass `hash` or `hash.data` or `&hash.data` into this function is not going to help me think more carefully about the expense of calculating a hash. If anything, it's going to distract me from that. I have limited time and attention. To be as productive as possible, I would like to focus it on the most important details.

In light of that, I'm going to resolve this conversation and dismiss my review on this PR. I don't think the downsides of this change are as costly as this debate, and I do see some upsides to the change. If others prefer it this way as you do, then I won't stand in the way.

InKryption: Fwiw, I prefer this simpler API, the original `update` always felt very magical and opaque to me.

dnut (approved):

PR #955, opened 2025-09-23
feat(replay): fix multithreaded

Sobeston:


src/replay/execution.zig
```diff
@@ -185,20 +185,12 @@ pub const LogHelper = struct {
...
-        deinit_allocator: Allocator,
-    ) void {
+    pub fn logActiveSlots(self: *LogHelper, active_slots: []const u64) void {
         self.slots_are_the_same = if (self.last_active_slots) |last_slots|
             std.mem.eql(u64, active_slots, last_slots)
         else
             false;
 
-        if (self.last_active_slots) |slots| deinit_allocator.free(slots);
-        self.last_active_slots = active_slots;
```
dnut: This changes the semantics of LogHelper. My goal when I wrote this was to avoid logging the same message repeatedly. With your change, you've made it so this rule only applies to the very first log message. Any other message will be allowed to be repeated. Was that your intention?

Maybe there was a double free here that needed to be fixed, but I think the solution for that problem (without changing the semantics here) would be to remove the defer from the calling context.

build.zig
```diff
@@ -217,38 +223,47 @@ pub fn build(b: *Build) !void {
...
         .optimize = config.optimize,
+        .force_pic = config.force_pic or config.enable_tsan,
     }).module("secp256k1");
 
     const tracy_mod = b.dependency("tracy", .{
         .target = config.target,
         .optimize = config.optimize,
         .tracy_enable = config.enable_tracy,
         .tracy_no_system_tracing = false,
+        .tracy_callstack = 6,
```
dnut, line 252: what's this for?

src/replay/scheduler.zig
```diff
@@ -212,11 +212,23 @@ pub const TransactionScheduler = struct {
                 .exit => {},
             }
         }
+    }
+
+    pub fn poll(self: *TransactionScheduler) !ConfirmSlotStatus {
+        self.collectResults();
```
dnut: Your current approach seems good, but I can see another approach where we don't need to collectResults here.

I think the main benefit of doing this on every poll, instead of just in the `done` branch, is because it triggers the exit condition the first time it detects a failure, so the threads will exit asap. I think the same can be achieved by calling collectResults in every branch of the switch, and removing the condition for collecting in the `done` branch.

What do you think, do you see a benefit of one approach over the other?

build.zig
Sobeston: This makes Tracy, when enabled, collect (up to 6) stack frames for every (memory, log, etc) event. Having this on allows you to easily find memory leaks. I personally add this whenever I use tracy so it might as well be a default.

src/replay/execution.zig
Sobeston: Switched it to using a hash of active_slots f41c5458873670a9d4dafa0be85e03c3965f6c37

src/replay/scheduler.zig
Sobeston: 973f4f42f0b1321b3bc7102045cb5bfa187eacd3

I changed it to just call collectResults in every branch, it seems cleaner than checking and retrying in the done branch.

src/replay/execution.zig
dnut: This is much better than my convoluted approach to ownership.

dnut (approved):

dnut (approved):

src/replay/execution.zig
```diff
@@ -696,30 +696,26 @@ pub const BlockError = enum {
...
-    /// takes ownership of active_slots,
-    pub fn logActiveSlots(
-        self: *LogHelper,
-        active_slots: []const u64,
-        deinit_allocator: Allocator,
-    ) void {
-        self.slots_are_the_same = if (self.last_active_slots) |last_slots|
-            std.mem.eql(u64, active_slots, last_slots)
+    pub fn logActiveSlots(self: *LogHelper, active_slots: []const u64) void {
+        const active_slots_hash = Hash.generateSha256(std.mem.sliceAsBytes(active_slots));
```
dnut, line 712: I'm happy to merge this as-is but a thought just occurred to me. You could use FnvHasher here instead, which would be more efficient. It doesn't really matter though since this function is not called very often. Feel free to resolve this conversation.

```suggestion
        const active_slots_hash = FnvHasher.hash(std.mem.sliceAsBytes(active_slots));
```

dadepo (approved):

src/replay/execution.zig
Sobeston: I think you're right, yeah, sha256 is overkill. Going to resolve to get this in asap though!

PR #956, opened 2025-09-24
refactor(vm): don't allocate for syscall registry

Rexicon226:


src/ledger/database/interface.zig
```diff
@@ -742,6 +742,8 @@ pub fn testDatabase(comptime Impl: fn ([]const ColumnFamily) type) type {
         }
 
         test "WriteBatch.deleteRange" {
+            if (true) return error.SkipZigTest;
```
InKryption, line 746: What's up with this test?

Rexicon226: It's just flaky, I've had it fail on pretty much every machine I own more than once. Gotta rewrite the ledger tbh.

Rexicon226: Removed the skip.

PR #957, opened 2025-09-25
refactor(replay)

dnut:


src/replay/service.zig
```diff
@@ -551,290 +332,13 @@ pub fn initProgressMap(
     return progress;
 }
 
-const FrozenSlotsSortCtx = struct {
+pub const FrozenSlotsSortCtx = struct {
```
InKryption, line 337: Why is this pub? Where is it used? Should be scoped to wherever it is actually needed, doesn't make sense as a part of replay's public API.

src/replay/scheduler.zig
```diff
@@ -1,106 +1,233 @@
...
-    committer: Committer,
-    transactions: []const ResolvedTransaction,
+    logger: Logger,
+    pool: *HomogeneousThreadPool(PohTask),
+    initial_hash: Hash,
+    entries: []const Entry,
     exit: *Atomic(bool),
-) !BatchResult {
-    var zone = tracy.Zone.init(@src(), .{ .name = "processBatch" });
-    zone.value(transactions.len);
```
InKryption, line 48: Is the removal of the tracy zone here intentional? Or is this just a weird diff?

src/replay/consensus/core.zig
```diff
@@ -21,215 +24,621 @@ const SlotState = sig.core.SlotState;
...
+        logger: Logger,
+        my_identity: Pubkey,
+        vote_identity: Pubkey,
+        root_slot: Slot,
+        root_hash: Hash,
+
+        // Data sources
+        account_store: AccountStore,
+        ledger_reader: *LedgerReader,
+        ledger_writer: *sig.ledger.LedgerResultWriter,
```
InKryption, line 113: Should this be using the `UnifiedLedger` instead?

src/replay/execution.zig
```diff
@@ -279,7 +471,7 @@ fn prepareSlot(state: ReplayExecutionState, slot: Slot) !PreparedSlot {
             .i_am_leader = i_am_leader,
             .epoch_stakes = &epoch_constants.stakes,
             .now = sig.time.Instant.now(),
-            .validator_vote_pubkey = state.vote_account,
+            .validator_vote_pubkey = null, // voting not currently supported
```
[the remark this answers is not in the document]
dnut: no ticket for this

src/replay/consensus/process_result.zig
```diff
@@ -63,133 +64,110 @@ pub const ProcessResultState = struct {
     purge_repair_slot_counter: *PurgeRepairSlotCounters,
 };
 
-pub fn processResult(state: ProcessResultState, result: sig.replay.execution.ReplayResult) !bool {
+pub fn processResult(params: ProcessResultParams, result: sig.replay.execution.ReplayResult) !void {
```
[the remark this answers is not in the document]
dnut: this functionality was moved elsewhere. it's still being done when freezing slots

src/replay/consensus/core.zig
```diff
@@ -21,215 +24,621 @@ const SlotState = sig.core.SlotState;
...
+            const root = slot_tracker.getRoot();
+            const root_slot = slot_tracker.root;
+            const root_hash = root.state.hash.readCopy();
+            break :blk .{ root_slot, root_hash.? };
+        };
+
+        var frozen_slots = try slot_tracker.frozenSlots(allocator);
+        defer frozen_slots.deinit(allocator);
+
+        frozen_slots.sort(replay.service.FrozenSlotsSortCtx{ .slots = frozen_slots.keys() });
```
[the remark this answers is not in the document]
dnut: i don't see a clearer place for it. honestly we should just remove this whole approach, it's a hack. but i don't think it's in scope for this pr

src/replay/consensus/core.zig
```diff
@@ -1,235 +1,661 @@
...
+        // Given a root and a list of `frozen_slots` sorted smallest to greatest by slot,
+        // initialize a new HeaviestSubtreeForkChoice
+        //
+        // Analogous to [`new_from_frozen_banks`](https://github.com/anza-xyz/agave/blob/0315eb6adc87229654159448344972cbe484d0c7/core/src/consensus/heaviest_subtree_fork_choice.rs#L235)
+        var heaviest_subtree_fork_choice = fork_choice: {
+            var heaviest_subtree_fork_choice: HeaviestSubtreeForkChoice =
+                try .init(allocator, .from(logger), .{
+                    .slot = root_slot,
+                    .hash = root_hash,
+                });
```
InKryption: Missing `errdefer`?

src/replay/consensus/core.zig
```diff
@@ -1,235 +1,661 @@
...
-    fork_choice: *ForkChoice,
+    my_identity: Pubkey,
+
+    // Core consensus state
+    fork_choice: HeaviestSubtreeForkChoice,
+    replay_tower: ReplayTower,
+    latest_validator_votes: LatestValidatorVotes,
+    status_cache: sig.core.StatusCache,
+    slot_data: SlotData,
+    arena_state: std.heap.ArenaAllocator.State,
```
InKryption, line 76: Should add a comment on this field about the fact that it is used for some temporary allocations that don't outlive functions; ie, it isn't used for any persistent data (unless that has changed during this PR, but I assume that's not the case).

src/replay/consensus/core.zig
```diff
@@ -1,235 +1,661 @@
...
+            });
+
+            const SlotSet = SortedSetUnmanaged(Slot);
+
+            // arena-allocated
+            var ancestors: std.AutoArrayHashMapUnmanaged(Slot, Ancestors) = .empty;
+            var descendants: std.AutoArrayHashMapUnmanaged(Slot, SlotSet) = .empty;
+            for (slot_tracker.slots.keys(), slot_tracker.slots.values()) |slot, info| {
+                const slot_ancestors = &info.constants.ancestors.ancestors;
+                const ancestor_gop = try ancestors.getOrPutValue(arena, slot, .EMPTY);
```
InKryption, line 396: Seems a bit counter-intuitive that `ancestors` is part of `constants`, but is being mutated here. Should it maybe instead be part of SlotState? If not, an explanation turned into a comment explaining why would be appreciated.

InKryption (commented):
A few nits, comments, and questions, but otherwise great stuff.

src/replay/service.zig
dnut: It's used in both replay/service.zig and replay/consensus/core.zig.

src/replay/scheduler.zig
dnut: This is a different function. `processBatch` was moved to `execution.zig`.

src/replay/consensus/core.zig
dnut: It doesn't need all the state from `UnifiedLedger`. I don't think we need to depend on extra stuff here. I wrote `UnifiedLedger` to simplify the initialization of the ledger, not to pass around all the ledger state as a bundle.

Soon I'd like to refactor the entire thing so that there's just a `Ledger` struct to pass around containing `LedgerDB` plus two integers, and that will be the only thing we need to pass around. The writer and reader structs can be deleted at that point, and all their methods can become normal functions that take `Ledger`. `UnifiedLedger` can stick around as a way to init/deinit all the ledger state. Maybe it can be renamed to `LedgerInitializer` or something like that, for clarity of purpose.

src/replay/consensus/core.zig
dnut: This is the map containing a collection of instances of `Ancestors`. It's not the individual instance of `Ancestors` found in slot constants.

src/replay/consensus/core.zig
dnut: c791225b61f076bcf84c6cef8e4e1091211be0ad

src/replay/consensus/core.zig
dnut: c791225b61f076bcf84c6cef8e4e1091211be0ad

src/replay/service.zig
InKryption: I see, that is a bit awkward. I'm assuming we'll come up with a better solution in the future, so in the meantime how about you just chuck a comment on it explaining that it's only exposed to be used by those two things, and that it should be removed once it's no longer needed.

src/replay/scheduler.zig
InKryption: Weird diff, gotcha

src/replay/consensus/core.zig
InKryption: Ah, you're right, I misread, my brain fudged `slot_ancestors` and `ancestors`.

src/replay/service.zig
dnut: 0c4a5c12bc205879487976d28b80e2da3d57083c

PR #958, opened 2025-09-25
feat(accountsdb): add allocator param to get methods

Sobeston:
Main objective: add an allocator parameter to account gets - call sites currently rely on using the accountsdb allocator, which is pointless

### DIFF
--- a/src/replay/freeze.zig
+++ b/src/replay/freeze.zig
@@ -366,9 +366,6 @@ pub fn deltaLtHash(
 
     assert(!parent_ancestors.containsSlot(slot));
 
-    const zone = tracy.Zone.init(@src(), .{ .name = "deltaLtHash" });
-    defer zone.deinit();
-
     var arena = std.heap.ArenaAllocator.init(tmp_alloc);
     defer arena.deinit();
     const allocator = arena.allocator();


src/replay/freeze.zig
```diff
@@ -345,12 +356,20 @@ pub fn deltaMerkleHash(account_reader: AccountReader, allocator: Allocator, slot
...
     parent_ancestors: *const Ancestors,
 ) !LtHash {
     assert(!parent_ancestors.containsSlot(slot));
 
+    const zone = tracy.Zone.init(@src(), .{ .name = "deltaLtHash" });
+    defer zone.deinit();
+
+    var arena = std.heap.ArenaAllocator.init(tmp_alloc);
+    defer arena.deinit();
+    const allocator = arena.allocator();
```
dnut, line 371: what's the point of this arena if it's not being recycled?

src/replay/resolve_lookup.zig
```diff
@@ -304,14 +306,16 @@ fn getLookupTable(
     const account_bytes = buf[0..account.data.len()];
     account.data.readAll(account_bytes);
 
-    return AddressLookupTable.deserialize(account_bytes) catch {
+    var table = AddressLookupTable.deserialize(account_bytes) catch {
         return error.InvalidAddressLookupTableData;
     };
+    table.addresses = try allocator.dupe(Pubkey, table.addresses);
```
dnut: nice catch. could you add a doc comment on `AddressLookupTable.deserialize` explaining this footgun?

here's another idea, what if deserialize took an `?Allocator` param with a doc comment explaining it, which would force the caller to think about the pointer. maybe that's awkward, just an idea

src/replay/freeze.zig
Sobeston: The arena is still useful in this case to reduce allocs overall, it brings it down from potentially 100s to something much lower (<=3 when I looked at it).

src/replay/resolve_lookup.zig
Sobeston: I added a warn in bb94ea60c6f1fb5eb5315323b69b5adf8b65f89a

Thought it'd be cleaner to have two methods here: 874faa090b15bbbc2b866eb29389b91b35546599

dnut: My reasoning behind the optional parameter was to force people to think about the parameter that they are *required* to provide. It can be null, but it still at least needs to be considered by the programmer. Convenience wasn't one of my considerations.

Anyway I don't have a strong opinion here so feel free to resolve this conversation.

src/replay/freeze.zig
dnut: Interesting. How does it reduce the number of allocations?

And what are the "allocs" you are counting? Are they syscalls? Or are you counting calls to the passed in allocator? Have you considered that it may not make an actual difference in the number of syscalls, and you're just masking calls from your instrumentation by introducing another layer?

What I mean is, consider this hypothetical: 100 calls to malloc actually only result in 3 syscalls. You've introduced an arena on top of malloc, and maybe it has the same behavior as malloc. Those 100 calls to the arena only results in 3 calls to malloc, but now each of those 3 calls to malloc result in a syscall (because they are larger). Your instrumentation is counting the number of calls to malloc, and that number went down, but maybe that number is not important. It's still the same number of calls to *some* allocator, and the same number of syscalls. Reducing the calls to the intermediary doesn't seem important.

src/accountsdb/account_store.zig
```diff
@@ -291,19 +271,16 @@ pub const ThreadSafeAccountMap = struct {
...
 
-    fn toAccount(allocator: Allocator, account: AccountSharedData) !Account {
-        const data = try allocator.dupe(u8, account.data);
-        errdefer allocator.free(account.data);
-
+    fn asAccount(account: AccountSharedData) Account {
         return .{
             .lamports = account.lamports,
-            .data = .{ .owned_allocation = data },
+            .data = .{ .unowned_allocation = account.data },
```
InKryption, line 306: Why not also just duplicate this data? This seems like it would be unsound in the event that we implement "cleaning logic" on this impl which runs while a caller is still holding onto their reference to it.

InKryption (commented):
Just one comment, otherwise love this

src/replay/freeze.zig
Sobeston: The arena is effectively a segmented list of buffers, where the buffers grow super-linearly. In almost all cases it should greatly reduce the amount of syscalls performed.

If we only call malloc 3 times, that means we're doing a maximum of 3 mmap syscalls. If we leave it to malloc, we're probably doing 3 <= X <= 100 syscalls, and I doubt that X = 3, I don't think malloc would be optimised in this way.

The other advantage of using an arena in this way is that the arena is single-threaded - calls to malloc can mean contention

src/replay/resolve_lookup.zig
Sobeston: I prefer separate functions as it means better error types

src/accountsdb/account_store.zig
Sobeston: We clone accounts on put, meaning that the ThreadSafeAccountMap itself owns all the account datas by this point already. If the caller calls .deinit() on the accounts they get from .get(), that's a no-op. 

> in the event that we implement "cleaning logic" on this impl which runs while a caller is still holding onto their reference to it.

I think we'd need to refcount the accounts in this case (or something equivalent), so that's out of scope

src/replay/resolve_lookup.zig
dnut: It would be cool if zig supported layering of comptimeness. the `?` part of the parameter is comptime while the `Allocator` part of the parameter is runtime. If you had that, the return type could be conditioned on the value of the `?` part of the parameter.

I'm sure there's a way to accomplish something like this with zig as-is by breaking it out into multiple parameters, but I expect that would be pretty ugly. It would be cool if the language somehow supported this directly in the expression of a single type.

This is probably too convoluted to be worth pursuing, but it would scratch that itch.

dnut (approved):

InKryption (approved):

InKryption (approved):

dnut (approved):

PR #973, opened 2025-09-30
vm: correct `neg` wrapping behaviour

Rexicon226:


src/net/socket_utils.zig
```diff
@@ -628,6 +628,7 @@ pub const BenchmarkPacketProcessing = struct {
 };
 
 test "benchmark packet processing" {
+    if (true) return error.SkipZigTest;
```
dnut: flagging this to resolve https://github.com/Syndica/sig/pull/974#discussion_r2391102391 before merging this pr

Rexicon226: Rebased this PR onto #974, and addressed comment there.

dnut (approved):

PR #974, opened 2025-09-30
test: disable test that hangs on MacOS 26

Rexicon226:
On macOS 26 this test just hangs; the `send`s are happening with data, but the `recv`s never get anything.
why? No idea! This could be either a macOS kernel bug or a user bug, although I don't see it being the latter yet, since the code is simple and all seems correct enough.

### DIFF
--- a/src/net/socket_utils.zig
+++ b/src/net/socket_utils.zig
@@ -628,6 +628,11 @@ pub const BenchmarkPacketProcessing = struct {
 };
 
 test "benchmark packet processing" {
+    // This test hangs on MacOS 26 for unknown reasons.
+    // It seems that performing `recv`s in this manner stopped working suddently, it could be a
+    // MacOS kernel bug, or an issue with our code.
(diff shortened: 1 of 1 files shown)


src/net/socket_utils.zig
```diff
@@ -628,6 +628,7 @@ pub const BenchmarkPacketProcessing = struct {
 };
 
 test "benchmark packet processing" {
+    if (true) return error.SkipZigTest;
```
dnut: Is this test only a problem in macos? If so, this should be conditioned on the target os being macos. Also, it would be nice to have a comment explaining why this is skipped.

Rexicon226: Both great points, one second.

Rexicon226: b6e1fe3f052c47a51a45268870942f4a5e74b047

kprotty (approved):

PR #980, opened 2025-10-02
refactor(shred-network): consolidate threads

dnut:
The current structure of the shred network is copied directly from agave. Each pair of `[square brackets]` represents a thread. In agave they use different names, and some of the components are duplicated on multiple threads in agave where we use just one, but otherwise the architecture is the same. I simply copied agave without thinking about how to improve the design.

```
[turbine socket listener]--\
                            |-->[shred receiver]-->[shred verifier]-->[shred processor]-->ledger
[repair socket listener]---/
```

This design is pointlessly complicated. It makes the code hard to understand and hard to debug. This is a sequence of steps that need to happen in order, and there's no good reason why they shouldn't happen in the same thread.

I've started to reduce the number of threads here by consolidating the logic from the shred receiver, verifier, and processor into a single thread. Here's the new approach I have in the current PR:

```
[turbine socket listener]--\
                            |-->[shred receiver]-->ledger
[repair socket listener]---/
```

I'd like to also flatten down the socket threads so that this entire thing becomes a single thread, but that's going to require a more complicated rework of our networking code.

The performance appears to be unchanged by this refactor. Before and after this change, sig is able to process about 40 slots worth of shreds per second on my computer running on my home network.

## Future optimizations

If in the future, if the number of shreds skyrockets and we really need multiple threads to handle them, then it will require a complete redesign. The old design was still overly pipelined, even if you want to parallelize this to the max. If you need more parallelism, you really only need to define two single threaded tasks: 
1. receive the shred from the network, verify them, and collect metadata about them
2. write the shreds into the ledger

You can parallelize task 1 in a thread pool if necessary. Currently we have task one split into four different threads that all do different things and I don't see any reason to keep this up.


src/net/socket_utils.zig
```diff
@@ -457,7 +459,19 @@ pub const SocketThread = struct {
         incoming_channel: *Channel(Packet),
         exit: ExitCondition,
     ) !*SocketThread {
-        return spawn(allocator, logger, socket, incoming_channel, exit, .receiver);
+        return spawn(allocator, logger, socket, incoming_channel, exit, .receiver, .empty);
+    }
+
+    /// Every packet will include the specified flags
+    pub fn spawnReceiverFlagged(
```
InKryption: Any reason to have this as a separate function, and not just update `spawnReceiver` to take an additional parameter that the caller can specify `.empty` for?

src/shred_network/service.zig
```diff
@@ -133,32 +94,39 @@ pub fn start(
...
-    // processor (thread)
-    const processor = shred_network.shred_processor;
+    // channels (cant use arena as they need to alloc/free frequently &
+    // potentially from multiple sender threads)
+    const retransmit_channel = try Channel(Packet).create(deps.allocator);
+    try defers.deferCall(Channel(Packet).destroy, .{retransmit_channel});
+
+    // receiver (threads)
+    const shred_receiver = try arena.create(ShredReceiver);
+    shred_receiver.* = try ShredReceiver.init(deps.allocator, .from(deps.logger), .{
```
InKryption: Aesthetic suggestion:
```suggestion
    shred_receiver.* = try .init(deps.allocator, .from(deps.logger), .{
```

src/shred_network/shred_receiver.zig
```diff
@@ -5,149 +5,232 @@ const shred_network = @import("lib.zig");
...
+        exit: *Atomic(bool),
+
+        repair_socket: Socket,
+        turbine_socket: Socket,
+
+        /// me --> retransmit service
+        maybe_retransmit_shred_sender: ?*Channel(Packet),
+
+        shred_version: *const Atomic(u16),
+        registry: *sig.prometheus.Registry(.{}),
```
InKryption: Considering this is only used during `init`, could this be made a direct parameter of `init`? Would make for a clean separation.

src/shred_network/shred_receiver.zig
```diff
@@ -5,149 +5,232 @@ const shred_network = @import("lib.zig");
...
+                    self.logger.err().logf(
+                        "failed to deserialize verified shred {?}.{?}: {}",
+                        .{ layout.getSlot(shred_payload), layout.getIndex(shred_payload), e },
+                    );
+                    return null;
+                };
+            } else |err| {
+                self.verifier_metrics.fail.observe(err);
+                return null;
+            }
```
InKryption:
```suggestion
            shred_verifier.verifyShred(
                &packet,
                self.params.leader_schedule,
                &self.verified_merkle_roots,
                self.verifier_metrics,
            ) catch |err| {
                self.verifier_metrics.fail.observe(err);
                return null;
            };
            self.verifier_metrics.verified_count.inc();
            if (self.params.maybe_retransmit_shred_sender) |retransmit_shred_sender| {
                try retransmit_shred_sender.send(packet);
            }
            const shred_payload = layout.getShred(&packet) orelse
                return error.InvalidVerifiedShred;
            return Shred.fromPayload(allocator, shred_payload) catch |e| {
                self.logger.err().logf(
                    "failed to deserialize verified shred {?}.{?}: {}",
                    .{ layout.getSlot(shred_payload), layout.getIndex(shred_payload), e },
                );
                return null;
            };
```

src/net/socket_utils.zig
kprotty: https://github.com/Syndica/sig/pull/980/commits/18934047424d879932597c1f11c77d32237fb29f

src/shred_network/shred_receiver.zig
kprotty: https://github.com/Syndica/sig/pull/980/commits/57328d707717908a3841e7b2be7a755b868e20d7

src/shred_network/shred_receiver.zig
kprotty: https://github.com/Syndica/sig/pull/980/commits/57328d707717908a3841e7b2be7a755b868e20d7

src/shred_network/service.zig
kprotty: https://github.com/Syndica/sig/pull/980/commits/57328d707717908a3841e7b2be7a755b868e20d7

PR #985, opened 2025-10-04
chore(testing): use random seed instead of hardcoding

Rexicon226:
It's preferred to use the random seed provided by the stdlib (which changes every run, but is deterministic, and can be passed into the test runner again to reproduce identical results), instead of just hardcoding a single number. This ensures the unit doesn't have some hidden dependence on the seed of the prng used.

Had to fix a couple cases just like this, where we would hardcode expected results which depend on the seed to the prng. The change makes it clearer what inputs are used to get said output.

Almost the entire diff is just that, maybe a few little fixes in random places.

### DIFF
--- a/src/shred_network/repair_message.zig
+++ b/src/shred_network/repair_message.zig
@@ -245,12 +245,11 @@ test "signed/serialized RepairRequest is valid" {
 
 test "RepairRequestHeader serialization round trip" {
     const allocator = std.testing.allocator;
-    var prng: std.Random.DefaultPrng = .init(std.testing.random_seed);
 
     const header: RepairRequestHeader = .{
         .signature = .ZEROES,
-        .sender = .initRandom(prng.random()),
-        .recipient = .initRandom(prng.random()),
+        .sender = .parse("2Trk3PpxjMF2AYWr9pVjwzxZ9jXN4HMawCfqscupx4Nq"),
+        .recipient = .parse("C5bJrseEwRf4ffTr9dP7wsYmWkJKScZX8VqqpzvNNf4S"),
         .timestamp = 5924,
         .nonce = 123,
     };
@@ -270,7 +269,6 @@ test "RepairRequestHeader serialization round trip" {
         255, 82,  27,  175, 162, 144, 207, 151, 36,  23,  0,   0,   0,   0,   0,
         0,   123, 0,   0,   0,
     };
-
     try std.testing.expectEqualSlices(u8, &expected, serialized);
 
     const roundtripped = try bincode.readFromSlice(


src/accountsdb/db.zig
```diff
@@ -3836,7 +3836,7 @@ test "generate snapshot & update gossip snapshot hashes" {
 
     const allocator = std.testing.allocator;
 
-    var prng = std.Random.DefaultPrng.init(123); // TODO: use `std.testing.random_seed` when we update
+    var prng = std.Random.DefaultPrng.init(std.testing.random_seed); // TODO: use `std.testing.random_seed` when we update
```
dnut:
```suggestion
    var prng = std.Random.DefaultPrng.init(std.testing.random_seed);
```

src/accountsdb/swiss_map.zig
```diff
@@ -181,10 +181,10 @@ pub fn SwissMapUnmanaged(
                 memory.len / (@sizeOf([GROUP_SIZE]KeyValue) + @sizeOf([GROUP_SIZE]State));
 
             const group_size = n_groups * @sizeOf([GROUP_SIZE]KeyValue);
-            const group_ptr: [*][GROUP_SIZE]KeyValue = @alignCast(@ptrCast(memory.ptr));
+            const group_ptr: [*][GROUP_SIZE]KeyValue = @ptrCast(@alignCast(memory.ptr));
```
dnut, line 184: does the order matter?

src/replay/freeze.zig
```diff
@@ -518,7 +518,7 @@ test "delta hashes with many accounts" {
     const hash_slot = 2; // changing this may require changing ancestors in 2 places
 
     if (generate_rust_code) std.debug.print(
-        \\.
```
dnut, line 545: This was here for a reason. The test runner trims leading whitespace which obscures the first line of output.

src/replay/freeze.zig
```diff
@@ -590,12 +590,32 @@ test "delta hashes with many accounts" {
...
+    const expected_lt_hash = [1024]u16{
```
dnut: This adds a bunch of visual clutter when scrolling through the file, which I feel makes the code harder to read. What's the benefit?

Also, you can remove `// sig fmt: off` in the line above since it was only there to prevent the style script from complaining about the line length.

When it's clear that someone made a deliberate stylistic choice, please discuss your disagreement with that person before reverting their code. I could also come back with another PR and revert this change because I disagree with it, but that would just waste both of our time. Let's come to an agreement and put a note in the style guide (does not need to be in this pr).

src/replay/freeze.zig
```diff
@@ -590,12 +590,32 @@ test "delta hashes with many accounts" {
...
+        15888, 56668, 25901, 22561, 43099, 9671,  17307, 1372,  4456,  22270, 36388, 7790,  63731, 38670, 29346, 11580, 42628, 30791, 27510, 54732, 21986, 61471, 41223, 24230, 58361, 5234,  21841, 44141, 4838,  50486, 23208, 39845, 45749, 34085, 4477,  57658, 9073,  60723, 5197,  37631, 7504,  37994, 17225, 11815, 32232, 35221, 51979, 4211,  41423, 9802,  11493, 9399,
+        32785, 34409, 12252, 62755, 17328, 45301, 16573, 37229, 49948, 48111, 1782,  42920, 63976, 24308, 64581, 33675, 21055, 24960, 12926, 46324, 17052, 15666, 9357,  33232, 1920,  47884, 25427, 20881, 7294,  1718,  24777, 55421, 35855, 63261, 40672, 62018,
+    };
     // sig fmt: on
-    const expected_merkle_hash =
-        Hash.parseRuntime("5tpzYxp8ghAETjXaXnZvxZov11iNEvSbDZXNAMoJX6ov") catch unreachable;
+    const expected_merkle_hash: Hash = .parse("5tpzYxp8ghAETjXaXnZvxZov11iNEvSbDZXNAMoJX6ov");
 
-    var parent_ancestors = Ancestors{};
+    var parent_ancestors: Ancestors = .{};
```
dnut, line 621: I noticed a lot of people on the team prefer this style, while I prefer the other approach. I think it's inconsequential which approach we take, but this is another thing we should make a decision about and put it in the style guide.

src/accountsdb/snapshot/download.zig
```diff
@@ -664,7 +664,7 @@ test "accounts_db.download: test remove untrusted peers" {
     var table = try GossipTable.init(allocator, allocator);
     defer table.deinit();
 
-    var prng = std.Random.DefaultPrng.init(0);
+    var prng = std.Random.DefaultPrng.init(std.testing.random_seed);
```
dnut, line 726: Does the zig fuzzer use this to randomize inputs?

src/accountsdb/swiss_map.zig
Rexicon226: [#24199](https://github.com/ziglang/zig/pull/24199), it will matter

src/replay/freeze.zig
Rexicon226: Understood. In general, printing in tests is a bad idea; it isn't really defined what will happen. But I understand this use case.

src/replay/freeze.zig
Rexicon226: Apologies, this change slipped in accidently. I remember we talked about it before, maybe it was on the PR that added this test case / file originally, and I understand why it was done. I'll revert it to how it was before.

src/accountsdb/snapshot/download.zig
Rexicon226: No. This has nothing to do with the zig fuzzer.

src/accountsdb/db.zig
Rexicon226: 5e86bbbad55d0da72b949da30a14482dcf93b029

src/replay/freeze.zig
Rexicon226: 5e86bbbad55d0da72b949da30a14482dcf93b029

src/replay/freeze.zig
Rexicon226: 5e86bbbad55d0da72b949da30a14482dcf93b029

src/replay/freeze.zig
Rexicon226: I agree - keep in mind the latter approach `T{}` is likely to become illegal in Zig's future.

dnut: I saw some threads in github discussing this syntax, but I didn't come to the conclusion that it is likely to become illegal. How do you know that it's likely to become illegal?

src/accountsdb/swiss_map.zig
dnut: I'm confused. The compiler will enforce this with errors? Is there a technical reason that this specific sequence is advantageous, or they just want to impose conformity?

src/replay/freeze.zig
Rexicon226: Pretty much the whole core team agrees that the RLS syntax is better, and I will be very surprised to see the proposal denied.

src/accountsdb/swiss_map.zig
Rexicon226: `zig fmt` will enforce the order, not the compiler. The reason is that it is more consistent, that is all.

PR #992, opened 2025-10-09
refactor(vote_listener): simplify `processVotesOnce`, unify name of channel, add test

InKryption:
Also, log errors instead of ignoring channel failures

### DIFF
--- a/src/consensus/vote_listener.zig
+++ b/src/consensus/vote_listener.zig
@@ -1112,50 +1112,52 @@ fn trackNewVotesAndNotifyConfirmations(
     }
 }
 
+pub fn slotTrackerElementGenesis(
+    allocator: std.mem.Allocator,
+    fee_rate_governor: sig.core.genesis_config.FeeRateGovernor,
+) std.mem.Allocator.Error!SlotTracker.Element {
+    const constants: sig.core.SlotConstants = try .genesis(allocator, fee_rate_governor);
+    errdefer constants.deinit(allocator);
+
+    const state: sig.core.SlotState = try .genesis(allocator);
+    errdefer state.deinit(allocator);
+
+    return .{
+        .constants = constants,
+        .state = state,
+    };
+}
+
 test "trackNewVotesAndNotifyConfirmations filter" {
     const allocator = std.testing.allocator;
 
     var prng_state: std.Random.DefaultPrng = .init(std.testing.random_seed);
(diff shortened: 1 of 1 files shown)


dadepo (dismissed):

src/consensus/vote_listener.zig
```diff
@@ -1025,11 +1102,178 @@ fn trackNewVotesAndNotifyConfirmations(
...
+            .parent_lt_hash = .IDENTITY,
+            .block_height = 1,
+            .collector_id = .ZEROES,
+            .max_tick_height = 1,
+            .fee_rate_governor = .DEFAULT,
+            .epoch_reward_status = .inactive,
+            .ancestors = .{ .ancestors = .empty },
+            .feature_set = .ALL_DISABLED,
+            .reserved_accounts = .empty,
+        },
```
dnut: Do any of these values actually matter for this test? Why not use `genesis`? Any time SlotConstants is modified, we have to fix any usages like this. But if you just use `genesis`, it's easier to modify SlotConstants.

src/consensus/vote_listener.zig
```diff
@@ -1025,11 +1102,178 @@ fn trackNewVotesAndNotifyConfirmations(
...
+        errdefer stakes.deinit(allocator);
+        try epoch_tracker.epochs.ensureUnusedCapacity(allocator, 1);
+        epoch_tracker.epochs.putAssumeCapacity(0, .{
+            .hashes_per_tick = 1,
+            .ticks_per_slot = 1,
+            .ns_per_slot = 1,
+            .genesis_creation_time = 1,
+            .slots_per_year = 1,
+            .stakes = stakes,
+            .rent_collector = .DEFAULT,
```
dnut: same as above

dnut (commented):
looks good, just have one thought about maintainability

src/consensus/vote_listener.zig
InKryption: I just copied another test, went ahead and updated: 3f96a73c29357af79b27100b78ed8d007e248f7a

src/consensus/vote_listener.zig
InKryption: 3f96a73c29357af79b27100b78ed8d007e248f7a

dnut (approved):

PR #993, opened 2025-10-10
fix(accountsdb): get snapshot validation working again / use delta LT hash

### LINKED ISSUE #988: fix(accountsdb): snapshot validation
currently snapshot validation fails and we need to always skip it.

Sobeston:
closes #988

changes
- uses the delta_lt hash instead of the old hash
- brings us more up to date with snapshots generated from a recent agave version
- removes incremental capitalisation checking (it seems dead in agave, and pointless)

### DIFF
--- a/src/accountsdb/db.zig
+++ b/src/accountsdb/db.zig
@@ -1152,19 +1152,20 @@ pub const AccountsDB = struct {
             },
         );
 
-        var total_lamports, const accounts_hash = blk: {
+        const total_lamports, const accounts_hash = blk: {
             var lamports_sum: u64 = 0;
             var hash: LtHash = .IDENTITY;
             for (task_results) |result| {
                 lamports_sum += result.lamports;
                 hash.mixIn(result.hash);
             }
+            // do subtraction after, to avoid potential overflows when
+            // .subtract > .lamports for a shard.
+            for (task_results) |result| {
+                lamports_sum -|= result.subtract;
+            }
             break :blk .{ lamports_sum, hash };
         };
-
-        for (task_results) |result| {
-            total_lamports -|= result.subtract;
-        }
 
         self.logger.debug().logf("collecting hashes from accounts took: {s}", .{timer.read()});
         timer.reset();


Rexicon226 (approved):

src/accountsdb/db.zig
```diff
@@ -1354,179 +1327,146 @@ pub const AccountsDB = struct {
...
+                    total_lamports += account.lamports;
+
+                    if (max_slot_ref.slot == 0) continue;
+                    break :slot max_slot_ref.slot;
+                };
+
+                // "mix out" old version of account from hash
+                {
+                    // I expect this branch to not be taken for full snapshots, however it would
+                    // still be entirely legal.
```
dnut: This is not correct if you're calculating an lthash over the entire database from scratch. The lthash of the database can be calculated by mixing every account's hash into identity. Mixing out old versions from that will result in an incorect lthash.

src/accountsdb/db.zig
```diff
@@ -1354,179 +1327,146 @@ pub const AccountsDB = struct {
...
+                    const previous_slot_ref = slotListMaxWithinBounds(
+                        ref_head.ref_ptr,
+                        null,
+                        latest_slot - 1,
+                    ) orelse continue;
+
+                    const account = try self.getAccountFromRef(allocator, previous_slot_ref);
+                    defer account.deinit(allocator);
+
+                    mix_out.mixIn(account.ltHash(previous_slot_ref.pubkey));
```
dnut: Why not `lt_hash.mixOut`? the `mix_out` variable seems unnecessary.

src/accountsdb/db.zig
Sobeston: I believe the lthash should be a hash of all *accounts* at the slot we're concerned with, not all entries for each account.

Any entry for an account which has a later entry can be considered overwritten - if a snapshot contains old entries I think this is legal but basically junk data

src/accountsdb/db.zig
Sobeston: https://github.com/Syndica/sig/pull/993/commits/bf8250e6d704848a6cf5ffb04a743b6a782a809a

Think I followed agave here, removed it, we don't need it

src/accountsdb/db.zig
dnut: > I believe the lthash should be a hash of all accounts at the slot we're concerned with, not all entries for each account.

Yes, we're on the same page here.

> if a snapshot contains old entries I think this is legal but basically junk data

It's junk data that should be totally ignored, not mixed out.

Let's say all of accountsdb has only a single account called `a`. In all of history, there have been only two versions of the account: `a0` and `a1`. The lt hash for accountsdb is `hash(a1)`. But here it seems like you're calculating the lt hash as `hash(a1) - hash(a0)`.

This makes sense when you're calculating an incremental hash from an incremental snapshot because you plan to immediately sum it with the hash from the full snapshot, which is `hash(a0)`, so the sum gives you `hash(a1)`. But if you're using this function to calculate the hash for all of accountsdb in one shot, `hash(a1) - hash(a0)` will be your final value, which is not correct.

Am I missing something about the context where this function is used?

Sobeston: Addressed in a1b8342900ba77712cf2d3ac395278c7c95620e5

src/accountsdb/db.zig
```diff
@@ -1081,16 +1071,29 @@ pub const AccountsDB = struct {
...
         },
+
+        fn min_slot(self: AccountHashesConfig) ?Slot {
+            return switch (self) {
+                .FullAccountHash => null,
+                .IncrementalAccountHash => |iac| iac.min_slot,
+            };
+        }
+
+        fn max_slot(self: AccountHashesConfig) ?Slot {
```
InKryption: Naming convention here should probably still be camelCase.

src/accountsdb/db.zig
```diff
@@ -1122,57 +1125,50 @@ pub const AccountsDB = struct {
...
+            var hash: LtHash = .IDENTITY;
+            for (task_results) |result| {
+                lamports_sum += result.lamports;
+                hash.mixIn(result.hash);
+            }
+            break :blk .{ lamports_sum, hash };
+        };
+
+        for (task_results) |result| {
+            total_lamports -|= result.subtract;
```
InKryption: Any reason not to do this subtraction alongside the addition?

Sobeston: It's possible for a shard to end with result.subtract > result.lamports. Unlikely to happen with real data? But I do this to avoid the possibility of overflow

src/accountsdb/db.zig
Sobeston: True, 18be3cb9b1b8f6d1cf48dffde2cd30981040b726

src/accountsdb/db.zig
InKryption: Moreso asking why this isn't
```suggestion
            for (task_results) |result| {
                total_lamports -|= result.subtract;
                lamports_sum += result.lamports;
                hash.mixIn(result.hash);
            }
            break :blk .{ lamports_sum, hash };
        };
```

InKryption (approved):

InKryption (approved):

dnut (approved):

PR #999, opened 2025-10-14
refactor(consensus:vote_listener): Unify recv & process threads, & more

InKryption:
* Metrics struct refactor:
  No reason for this struct of pointers to be mutable.
  And as mentioned, `VoteListener.init` was returning a reference to the
  local metrics variable, as well as passing it by reference in the
  receive and process threads, which would be invalidated as soon as the
  init method returns as well.
* Greatly simplify VoteListener:
  * Unify the receive loop and process loop into a single thread.
  * As a consequence of the former, remove `verified_vote_transactions`.
  * `gossip_vote_txs` is now a simple parameter that is passed through.
  * Trim down `VoteListener` struct's exposed state to just the thread.
  * Rename `UnverifiedVoteReceptor`->`GossipVoteReceptor`, and overhaul.
  * Remove `remaining_wait_time` wait loop from `listenAndConfirmVotes`.

### DIFF
--- a/src/consensus/vote_listener.zig
+++ b/src/consensus/vote_listener.zig
@@ -224,9 +224,6 @@ pub const VoteListener = struct {
     }
 };
 
-/// equivalent to agave's solana_gossip::cluster_info::GOSSIP_SLEEP_MILLIS
-const GOSSIP_SLEEP_MILLIS = 100 * std.time.ns_per_ms;
-
 const GossipVoteReceptor = struct {
     cursor: u64,
     vote_tx_buffer: std.ArrayListUnmanaged(Transaction),


InKryption:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Syndica/sig/999?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#999** <a href="https://app.graphite.dev/github/pr/Syndica/sig/999?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Syndica/sig/999?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#992** <a href="https://app.graphite.dev/github/pr/Syndica/sig/992?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/consensus/vote_listener.zig
```diff
@@ -200,202 +200,143 @@ pub const VoteListener = struct {
...
         }
 
-        // inc_new_counter_debug!("cluster_info_vote_listener-recv_count", votes.len());
-        if (unverified_votes_buffer.items.len == 0) {
-            std.time.sleep(GOSSIP_SLEEP_MILLIS);
-            return;
-        }
+        // Update metrics for gossip votes received
+        metrics.gossip_votes_received.add(vote_tx_buffer.items.len);
+        if (vote_tx_buffer.items.len == 0) return &.{};
```
dadepo, line 283: The original implementation had a sleep out of `GOSSIP_SLEEP_MILLIS`  in this case. Was it intentional to leave that out?

If so then the constant `GOSSIP_SLEEP_MILLIS` can be removed (it is still defined in this file)

src/consensus/vote_listener.zig
```diff
@@ -200,202 +200,143 @@ pub const VoteListener = struct {
...
 
-        // Update metrics for gossip votes received
-        metrics.gossip_votes_received.add(unverified_votes_buffer.items.len);
+        var index: usize = 0;
+        while (index != vote_tx_buffer.items.len) {
+            const vote_tx = vote_tx_buffer.items[index];
+            switch (try verifyVoteTransaction(allocator, vote_tx, epoch_tracker)) {
+                .verified => index += 1,
+                .unverified => {
+                    _ = vote_tx_buffer.swapRemove(index);
```
dadepo, line 294: I would have thought discarding the removed entry at `index` would lead to a leak. I see that `vote_tx` is being `deinited` all right but I would have read that as a copy based on how it is retrieved on line 293.

Do you know why it is not a leak?

dnut: It's a copy of the same pointer, so it will free the same memory.

src/consensus/vote_listener.zig
InKryption: Well spotted: 326b6d18397fb2116d895dc7561f1b3c7ded1770

PR #1000, opened 2025-10-15
feat(consensus): Some consensus component testing

dadepo:
Add a couple of tests covering the following scenarios:

### Basic Consensus Operations
- Vote on heaviest frozen descendant with fork stats computation
- Tower state tracking (last voted slot, vote history, lockout intervals)
- Fork choice heaviest slot selection

### Validator Votes & Stakes
- Landed votes from multiple validators populate bank statistics
- Voted stakes aggregation and lockout intervals tracking

### Root Advancement
- Automatic root progression when tower reaches MAX_LOCKOUT_HISTORY
- Pruning of old slots from SlotTracker and ProgressMap after root updates

### Some edge cases (Without channels)
- Vote refresh behavior when no new votable slots available
- Duplicate-confirmed fork detection via 52% stake threshold
- Duplicate slot detection via conflicting shreds channel

### Fork Switching
- Lockout-based fork switch prevention for locked-out siblings
- Successful fork switch to heavier sibling with sufficient stake
- Failed fork switch when stake threshold not met

### DIFF
--- a/src/replay/consensus/core.zig
+++ b/src/replay/consensus/core.zig
@@ -4427,14 +4427,21 @@ test "vote on heaviest frozen descendant with no switch" {
         .senders = stubs.senders,
         .receivers = stubs.receivers,
         .gossip_table = null,
+        .slot_leaders = null,
         .run_vote_listener = false,
     };
 
     // Build consensus dependencies
     const deps = TowerConsensus.Dependencies{
         .logger = .noop,
-        .my_identity = Pubkey.initRandom(std.crypto.random),
-        .vote_identity = Pubkey.initRandom(std.crypto.random),
+        .identity = .{
+            .vote_account = Pubkey.initRandom(std.crypto.random),
+            .validator = Pubkey.initRandom(std.crypto.random),
+        },
+        .signing = .{
+            .node = null,
+            .authorized_voters = &.{},
+        },
         .root_slot = root_slot,
         .root_hash = root_state.hash.readCopy().?,
         .account_reader = stubs.accountsdb.accountReader(),
@@ -4651,13 +4658,20 @@ test "vote accounts with landed votes populate bank stats" {
         .senders = stubs.senders,
         .receivers = stubs.receivers,
         .gossip_table = null,
+        .slot_leaders = null,
         .run_vote_listener = false,
     };
 
     const deps = TowerConsensus.Dependencies{
         .logger = .noop,
-        .my_identity = Pubkey.initRandom(std.crypto.random),
-        .vote_identity = Pubkey.initRandom(std.crypto.random),
+        .identity = .{
+            .vote_account = Pubkey.initRandom(std.crypto.random),
+            .validator = Pubkey.initRandom(std.crypto.random),
+        },
+        .signing = .{
+            .node = null,
+            .authorized_voters = &.{},
+        },
         .root_slot = root_slot,
         .root_hash = Hash.ZEROES,
         .account_reader = stubs.accountsdb.accountReader(),
@@ -4825,7 +4839,6 @@ test "root advances after vote satisfies lockouts" {
     var prng = std.Random.DefaultPrng.init(12345);
     const random = prng.random();
     const validator_vote_pubkey = Pubkey.initRandom(random);
-    const validator_identity_pubkey = Pubkey.initRandom(random);
 
     var epoch_tracker: EpochTracker = .{
         .epochs = .empty,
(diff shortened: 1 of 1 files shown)


build.zig
```diff
@@ -340,12 +346,29 @@ pub fn build(b: *Build) !void {
...
+    unit_test_exe.root_module.addImport("cli", cli_mod);
     switch (config.ledger_db) {
-        .rocksdb => unit_tests_exe.root_module.addImport("rocksdb", rocksdb_mod),
+        .rocksdb => unit_test_exe.root_module.addImport("rocksdb", rocksdb_mod),
         .hashmap => {},
     }
-    addInstallAndRun(b, test_step, unit_tests_exe, config);
+    addInstallAndRun(b, unit_test_step, unit_test_exe, config);
+
+    const consensus_test_exe = b.addTest(.{
```
dnut: I don't see why we need a separate executable for this. I would put the tests in `src` and compile them into the same binary as the other tests. I just don't see a benefit of separating it out at this point. It's inconsistent and seems like a source of confusion. I spoke to dadepo about this already but I'd like to hear your thoughts @InKryption.

These tests should run in ci. This approach is excluding them from running in ci. So we'll either need to change the circleci config or combine it with the unit tests.

InKryption: We will need a structure that resembles this in the future as soon as we extract these out into the test harness; these tests don't really make sense as part of the main test suite because they're not really unit tests, they're just preliminaries for what will become the main test cases cases for the harness.

dnut: I don't see how these are different from our other tests. They call functions in our code and assert a result. I can see the argument that they're not *unit* tests, but the same can be said about many of the other tests in our code.

I also don't see why a structure like this would be necessary if we build a test harness. I can see that you might want to keep it separate from sig for organizational purposes, but I don't think it would be strictly necessary. If we build a test harness, and it becomes clear that the code is organized better with separation, then we can split things up at that time. But for now, these look like normal tests.

I'm not convinced that we need to invest our time in a flexible test harness for Tower BFT. That would make it the most well-tested area of our code outside the SVM. Maybe our efforts would be better spent building a test harness for a component that isn't in the process of being deprecated and replaced with a totally new consensus protocol.

I'm trying to identify a concrete benefit of this approach and the only thing I can see is that it anticipates a future need for separation and proactively achieves that separation. It's a hypothetical benefit that we might see later. But there are concrete downsides that affect us right now. It's no longer straightforward to run the tests since it requires running multiple binaries and the filtering logic is inconsistent, it introduces more code complexity, and it introduces an inconsistency with how our code is organized. I see this leading to confusion and mistakes.

InKryption (approved):

src/replay/trackers.zig
```diff
@@ -175,6 +175,9 @@ pub const SlotTracker = struct {
             const parent_slot = current.constants.parent_slot;
             parents_list.appendAssumeCapacity(parent_slot);
 
+            // Stop if we've reached the root.
```
dnut:
```suggestion
            // Stop if we've reached genesis.
```
The code below makes sense but the comment is misleading.

`parent_slot == current_slot` in only one scenario: genesis. This is only true of the very first slot in history. Any other slot will specify a different slot as its parent, regardless of whether that slot has been rooted.

src/replay/consensus/core.zig
```diff
@@ -509,6 +509,13 @@ pub const TowerConsensus = struct {
                 if (!slot_info.state.isFrozen()) {
                     continue;
                 }
+
+                // Skip if the slot's hash is Hash.ZEROES (e.g., root slot)
+                const slot_hash = slot_info.state.hash.readCopy() orelse return error.MissingHash;
+                if (slot_hash.eql(Hash.ZEROES)) {
+                    continue;
+                }
+
```
dnut: A slot should never have a hash of zeroes. In agave they use zeroes to represent that the hash has not been calculated yet. We use `null` for this purpose. I would remove this check.

src/replay/trackers.zig
dadepo: [1820ef7](https://github.com/Syndica/sig/pull/1000/commits/1820ef7a85e4c6c3daa4af1e1c9f353df00d4311)

src/replay/consensus/core.zig
dadepo: [3c5ad18](https://github.com/Syndica/sig/pull/1000/commits/3c5ad182dfe1e17a439066e8ce69e9c8d67e11d7)

InKryption (approved):

dnut (approved):

PR #1001, opened 2025-10-15
feat(conformance): scripts to simplify running locally

dnut:
When I was first getting up and running with conformance, it was a painful experience cloning all the right repos, checking out the right versions, using the correct paths for all files, and running all the correct commands. I spent several hours sorting out all these details. The scripts in this PR are meant to save all these headaches so anyone can bootstrap and run the entire conformance suite locally with minimal effort.

see [conformance/README.md](https://github.com/Syndica/sig/blob/dnut/conformance/run-scripts/conformance/README.md)

I didn't include any installation scripts for system dependencies because these are very OS-specific, and it didn't give me much of a problem when I was starting. On my machine running arch linux, I only needed to install python3.11.

Also this fixes an issue in our CI. We weren't running the bpf loader v4 tests. Now we are.

### DIFF
--- a/src/consensus/vote_listener.zig
+++ b/src/consensus/vote_listener.zig
@@ -930,7 +930,7 @@ fn trackNewVotesAndNotifyConfirmations(
 
             const maybe_hash: ?Hash = get_hash: {
                 if (slot == last_vote_slot) break :get_hash last_vote_hash;
-                break :get_hash slot_data_provider.getSlotHash(last_vote_slot);
+                break :get_hash slot_data_provider.getSlotHash(slot);
             };
             const hash: Hash = maybe_hash orelse {
                 // In this case the supposed ancestor of this vote is missing. This can happen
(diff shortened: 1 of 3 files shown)


conformance/README.md
```diff
@@ -0,0 +1,22 @@
...
+- rust
+- zig 0.14.1
+
+```bash
+# set up the test environment
+scripts/setup-env.sh
+source env/pyvenv/bin/activate
+
+# compile the sig binary to test
+zig build -Doptimize=ReleaseSafe
```
Rexicon226: Should specify, `zig build libsolfuzz_sig`.

Rexicon226 (approved):
LGTM overall. i would like to move all of this stuff into the build.zig one day, but that's a future goal. Perhaps we can rewrite solana-conformance in Zig, or at least the `.fix` -> running pipeline shouldn't be that hard. Unfortunately it just has a lot of random hacks, which is why I gave up the first time I tried.

conformance/README.md
dnut: d415a9797ebab503ba541ae6227ed5a30f2a8184

yewman (approved):

PR #1002, opened 2025-10-15
fix(runtime): remove print from runtime test

dnut:
Please, no debug prints in tests.

It looks like this test was only added as a hack to satisfy codecov to run lines of code that only exist for printing test outputs while debugging tests locally. I don't think this is code we should cover in our tests.

I can see how this helped get a prior PR merged, but now we should be able to undo it.

### DIFF
--- a/src/runtime/program/testing.zig
+++ b/src/runtime/program/testing.zig
@@ -118,22 +118,15 @@ pub fn expectProgramExecuteResult(
 test expectProgramExecuteError {
     const allocator = std.testing.allocator;
 
-    // Test print_logs path
-    for ([_]bool{ false, true }) |print_logs| {
-        try expectProgramExecuteError(
-            error.UnsupportedProgramId,
-            allocator,
-            Pubkey.ZEROES, // invalid program id
-            &.{}, // empty instruction,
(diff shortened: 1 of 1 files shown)


src/runtime/program/testing.zig
```diff
@@ -118,22 +118,15 @@ pub fn expectProgramExecuteResult(
...
-        );
-    }
+    try expectProgramExecuteError(
+        error.UnsupportedProgramId,
+        allocator,
+        Pubkey.ZEROES, // invalid program id
+        &.{}, // empty instruction,
+        &.{.{ .index_in_transaction = 0 }},
+        .{ .accounts = &.{.{ .pubkey = Pubkey.ZEROES }} },
+        .{ .print_logs = false },
```
kprotty, line 128: do we need the `print_logs` feature even? Was debating to remove it (over testing it) for codecov, but wasn't sure.

dnut: my assumption is that it helps when a test fails, so you can quickly enable the logs to see what's going on. but I'm not that familiar with this code

kprotty: presuming it's fine to leave it

kprotty (approved):

PR #1003, opened 2025-10-15
feat(replay): epoch transitions

yewman:
# Overview
This PR implements epoch transition functionality including:
- feature activations
- bpf core migrations
- rewards calculation and distribution

## 1. Epoch Transitions 
Core epoch boundary processing:
- processNewEpoch() - orchestrates feature activations, stakes cache updates, epoch stakes computation, and partitioned rewards
- updateEpochStakes() - computes leader schedule epoch stakes
- applyFeatureActivations() - handles feature gate activations at epoch boundaries
- Includes BPF builtin program migration support

## 2. Rewards System 
- calculation.zig  - Stake/vote reward calculations, beginPartitionedRewards()
- distribution.zig - Partitioned reward distribution across blocks
- inflation_rewards.zig - Inflation-based point/reward computation
- hasher.zig - Reward hashing utilities 
- lib.zig - Shared types: RewardType, StakeReward, PartitionedRewards<T>

## 3. Stakes Consolidation
Merged stake.zig + vote_accounts.zig into unified module:
- Stakes(comptime stakes_type) - generic over .delegation, .stake, .account
- StakesCacheGeneric - thread-safe stakes cache with checkAndStore()
- VoteAccount with new .equals() method for proper comparison
- Simplified stake delegation calculation (removed intermediate map)

## 4. Builtin Programs
BPF migration infrastructure for built-in to BPF program transitions.

### DIFF
--- a/src/core/bank.zig
+++ b/src/core/bank.zig
@@ -446,13 +446,16 @@ pub fn parseStakes(
         const db_account_data = try db_account.data.readAllAllocate(allocator);
         defer allocator.free(db_account_data);
 
-        const db_versioned_vote_state = try sig.bincode.readFromSlice(
+        var db_versioned_vote_state = try sig.bincode.readFromSlice(
             allocator,
             sig.runtime.program.vote.state.VoteStateVersions,
             db_account_data,
             .{},
         );
-        const db_vote_state = try db_versioned_vote_state.convertToCurrent(allocator);
+        defer db_versioned_vote_state.deinit(allocator);
+
+        var db_vote_state = try db_versioned_vote_state.convertToCurrent(allocator);
+        defer db_vote_state.deinit(allocator);
 
         if (!db_vote_state.equals(&cached_vote_state)) return error.InvalidVoteAccount;
     }
--- a/src/replay/epoch_transitions.zig
+++ b/src/replay/epoch_transitions.zig
@@ -575,17 +575,18 @@ fn putAndUpdateCapitalization(
 fn burnAndPurgeAccount(
     store: SlotAccountStore,
     address: Pubkey,
-    account: AccountSharedData,
+    account: *const Account,
     capitalization: *Atomic(u64),
 ) !void {
-    const account_data_len = account.data.len;
+    try store.put(address, .{
+        .lamports = 0,
+        .data = &.{},
+        .executable = false,
+        .owner = .ZEROES,
+        .rent_epoch = 0,
+    });
     _ = capitalization.fetchSub(account.lamports, .monotonic);
-    var acc = account;
-    acc.lamports = 0;
-    @memset(acc.data, 0);
-    try store.put(address, acc);
     // NOTE: update account size delta in slot state?
-    _ = account_data_len;
 }
 
 fn putPrecompile(
@@ -601,7 +602,7 @@ fn putPrecompile(
         try burnAndPurgeAccount(
             store,
             precompile.program_id,
-            try AccountSharedData.fromAccount(allocator, &account),
+            &account,
             capitalization,
         );
     } else return;
@@ -635,12 +636,10 @@ fn putBuiltinProgramAccount(
     if (try store.reader().get(allocator, builtin_program.program_id)) |account| {
         defer account.deinit(allocator);
         if (sig.runtime.ids.NATIVE_LOADER_ID.equals(&account.owner)) return;
-        const account_shared_data = try AccountSharedData.fromAccount(allocator, &account);
-        defer allocator.free(account_shared_data.data);
         try burnAndPurgeAccount(
             store,
             builtin_program.program_id,
-            account_shared_data,
+            &account,
             capitalization,
         );
     }
--- a/src/replay/rewards/calculation.zig
+++ b/src/replay/rewards/calculation.zig
@@ -140,6 +140,8 @@ fn getRewardDistributionNumBlocks(
         MAX_PARTITIONED_REWARDS_PER_BLOCK,
     );
 
+    // Ensure at least 1 block, but cap to a maximum number of blocks
+    // MINIMUM_SLOTS_PER_EPOCH > MAX_FACTOR_OF_REWARD_BLOCKS_IN_EPOCH
     const MAX_FACTOR_OF_REWARD_BLOCKS_IN_EPOCH: u64 = 10;
     const max_chunks = try std.math.divFloor(
         u64,
@@ -500,7 +502,7 @@ fn calculateStakeVoteRewards(
         }
 
         voters_reward_entry.value_ptr.rewards +|= redeemed.voters_reward;
-        total_stake_rewards += redeemed.stakers_reward;
+        total_stake_rewards +|= redeemed.stakers_reward;
 
         try partitioned_stake_rewards.append(allocator, .{
             .stake_pubkey = stake_pubkey,
@@ -516,15 +518,12 @@ fn calculateStakeVoteRewards(
     const stake_rewards_slice = try partitioned_stake_rewards.toOwnedSlice(allocator);
     errdefer allocator.free(stake_rewards_slice);
 
-    const stake_rewards = StakeRewards{
-        .stake_rewards = try .init(allocator, stake_rewards_slice),
-        .total_stake_rewards_lamports = total_stake_rewards,
-    };
-    errdefer stake_rewards.deinit(allocator);
-
     return .{
         .vote_rewards = vote_rewards,
-        .stake_rewards = stake_rewards,
+        .stake_rewards = .{
+            .stake_rewards = try .init(allocator, stake_rewards_slice),
+            .total_stake_rewards_lamports = total_stake_rewards,
+        },
         .point_value = point_value,
     };
 }
@@ -558,7 +557,7 @@ fn calculateVoteAccountsToStore(
             },
             .account = vote_reward.account,
         });
-        total_vote_rewards += vote_reward.rewards;
+        total_vote_rewards +|= vote_reward.rewards;
     }
 
     const vote_rewards_slice = try vote_rewards.toOwnedSlice(allocator);
--- a/src/replay/rewards/distribution.zig
+++ b/src/replay/rewards/distribution.zig
@@ -36,6 +36,7 @@ pub fn distributePartitionedEpochRewards(
     capitalization: *AtomicU64,
     rent: *const Rent,
     slot_store: SlotAccountStore,
+    new_rate_activation_epoch: ?Epoch,
 ) !void {
     var stake_rewards = switch (epoch_reward_status.*) {
         .active => |active| active,
@@ -90,6 +91,7 @@ pub fn distributePartitionedEpochRewards(
             capitalization,
             stakes_cache,
             slot_store,
+            new_rate_activation_epoch,
         );
     }
 
@@ -126,6 +128,7 @@ fn distributeEpochRewardsInPartition(
     capitalization: *AtomicU64,
     stakes_cache: *StakesCache,
     slot_store: SlotAccountStore,
+    new_rate_activation_epoch: ?Epoch,
 ) !void {
     const lamports_distributed, const lamports_burnt, const updated_stake_rewards =
         try storeStakeAccountsInPartition(
@@ -135,7 +138,7 @@ fn distributeEpochRewardsInPartition(
             partition_index,
             stakes_cache,
             slot_store,
-            null, // TODO: pass in new_rate_activation_epoch
+            new_rate_activation_epoch,
         );
     defer {
         for (updated_stake_rewards) |reward| reward.deinit(allocator);
@@ -258,7 +261,7 @@ fn buildUpdatedStakeReward(
         else => return error.InvalidAccountData,
     }
 
-    account.lamports += partitioned_reward.stake_reward;
+    account.lamports +|= partitioned_reward.stake_reward;
     std.debug.assert(stake_state.stake.stake.delegation.stake +| partitioned_reward.stake_reward ==
         partitioned_reward.stake.delegation.stake);
     stake_state.stake.stake = partitioned_reward.stake;
@@ -427,6 +430,7 @@ test distributePartitionedEpochRewards {
         &capitalization,
         &rent,
         slot_store,
+        null,
     );
 }
 
@@ -560,6 +564,7 @@ test distributeEpochRewardsInPartition {
         &capitalization,
         &stakes_cache,
         slot_store,
+        null,
     );
 }
 
--- a/src/replay/rewards/inflation_rewards.zig
+++ b/src/replay/rewards/inflation_rewards.zig
@@ -214,7 +214,7 @@ fn getInflationStartSlot(slot: Slot, feature_set: *const FeatureSet) Slot {
     const devnet_and_testnet = feature_set.get(.full_inflation_devnet_and_testnet) orelse
         std.math.maxInt(Slot);
 
-    return if (slot >= mainnet_vote and slot >= mainnet_vote or slot >= devnet_and_testnet)
+    return if (slot >= mainnet and slot >= mainnet_vote or slot >= devnet_and_testnet)
         @min(mainnet, devnet_and_testnet)
     else if (feature_set.active(.pico_inflation, slot))
         feature_set.get(.pico_inflation).?
--- a/src/runtime/sysvar/stake_history.zig
+++ b/src/runtime/sysvar/stake_history.zig
@@ -91,6 +91,10 @@ pub const StakeHistory = struct {
             _ = self.entries.orderedRemove(MAX_ENTRIES - 1);
         }
 
+        if (index < self.entries.len and self.entries.buffer[index].epoch == epoch) {
+            return error.DuplicateEpoch;
+        }
+
         try self.entries.insert(index, .{ .epoch = epoch, .stake = entry });
     }
 
--- a/src/replay/service.zig
+++ b/src/replay/service.zig
@@ -427,6 +427,9 @@ pub fn trackNewSlots(
                 &state.capitalization,
                 &epoch_info.rent_collector.rent,
                 store,
+                constants.feature_set.newWarmupCooldownRateEpoch(
+                    &epoch_tracker.schedule,
+                ),
             );
 
             try updateSysvarsForNewSlot(
--- a/src/runtime/program/stake/state.zig
+++ b/src/runtime/program/stake/state.zig
@@ -63,15 +63,6 @@ pub const StakeStateV2 = union(enum) {
         delegation: Delegation,
         credits_observed: u64,
 
-        fn stake(
-            self: *const Stake,
-            epoch: Epoch,
-            history: *const sig.runtime.sysvar.StakeHistory,
-            new_rate_activation_epoch: ?Epoch,
-        ) void {
-            return self.delegation.getEffectiveStake(epoch, history, new_rate_activation_epoch);
-        }
-
         pub fn split(
             self: *Stake,
             ic: *sig.runtime.InstructionContext,


src/replay/rewards/EpochRewardsHasher.zig
```diff
@@ -0,0 +1,147 @@
...
+pub const EpochRewardsHasher = @This();
+
+hasher: SipHasher13,
+partitions: usize,
+
+pub fn init(partitions: usize, seed: *const Hash) !EpochRewardsHasher {
+    const key: [SipHasher13.key_length]u8 = @splat(0);
+    var hasher = SipHasher13.init(&key);
+    _ = try hasher.writer().write(&seed.data);
+    return EpochRewardsHasher{
```
Rexicon226:
```suggestion
    return .{
```

src/core/bank.zig
```diff
@@ -116,14 +125,14 @@ pub const SlotConstants = struct {
             .collector_id = bank_fields.collector_id,
             .max_tick_height = bank_fields.max_tick_height,
             .fee_rate_governor = bank_fields.fee_rate_governor,
-            .epoch_reward_status = .inactive,
             .ancestors = try bank_fields.ancestors.clone(allocator),
             .feature_set = feature_set,
```
Sobeston: missing errdefer?

src/core/genesis_config.zig
```diff
@@ -295,6 +338,11 @@ fn yearsAsSlots(years: f64, tick_duration: RustDuration, ticks_per_slot: u64) f6
     return years * SLOTS_PER_YEAR;
 }
 
+/// Zig pow implemetation is incorrect
+/// Failure (zig):  0.85, 4.019250798563942 -> 7.805634650110366e-2
+/// Expected:       0.85, 4.019250798563942 -> 7.805634650110367e-2
+extern fn pow(f64, f64) f64;
+
```
Sobeston: A bit misleading to say Zig's impl is incorrect.

(quite silly that this is necessary, but that's not our fault)

Might be best to quickly port over whatever glibc (or rust std) is doing here?

src/replay/trackers.zig
```diff
@@ -446,6 +450,21 @@ pub const EpochTracker = struct {
...
+    }
+
+    pub fn put(
+        self: *EpochTracker,
+        allocator: Allocator,
+        epoch: Epoch,
+        epoch_constants: EpochConstants,
+    ) Allocator.Error!void {
+        try self.epochs.ensureUnusedCapacity(allocator, 1);
+        self.epochs.putAssumeCapacity(epoch, epoch_constants);
```
Sobeston, line 466:
```suggestion
        try self.epochs.put(epoch, epoch_constants);
```

yewman: Removed entirely in follow up PR https://github.com/Syndica/sig/pull/1113

src/core/bank.zig
yewman: df77a47a

src/core/genesis_config.zig
yewman: df77a47a

src/core/bank.zig
```diff
@@ -295,8 +313,146 @@ pub const SlotState = struct {
...
+                        .lamports = vote_account.lamports,
+                        .data = data,
+                        .owner = vote_account.owner,
+                        .executable = vote_account.executable,
+                        .rent_epoch = vote_account.rent_epoch,
+                    });
+                    if (!std.meta.isError(deserialize_result)) {
+                        var deserialized = deserialize_result catch unreachable;
+                        defer deserialized.deinit(allocator);
+                        return error.VoteAccountNotCached;
```
dnut, line 398: This only returns error when there is no error? This is unintuitive. I can't see the reasoning behind it. Could you add a comment explaining it?

This branch is reached if the following conditions are met:
- a stake account was found in both the map and accountsdb
- the stake account's vote account was found in both the vote account's map and accountsdb
- the vote account is valid

How does this imply a vote account is not cached? And why is that a problem?

Also, is there a specific error you're looking for here? `fromAccountSharedData` could return a variety of errors, including OutOfMemory and InvalidOwner. Are you sure you want to consume all of these errors as if nothing went wrong?

src/core/epoch_stakes.zig
```diff
@@ -220,12 +226,12 @@ pub const NodeVoteAccounts = struct {
...
-            Pubkey,
-            random.uintLessThan(usize, max_list_entries),
-        );
-        errdefer allocator.free(vote_accounts);
-        for (vote_accounts) |*vote_account| vote_account.* = Pubkey.initRandom(random);
+        var vote_accounts = try std.ArrayListUnmanaged(Pubkey)
+            .initCapacity(allocator, max_list_entries);
+        errdefer vote_accounts.deinit(allocator);
+        for (0..random.uintAtMost(usize, max_list_entries)) |_| {
+            try vote_accounts.append(allocator, Pubkey.initRandom(random));
```
dnut:
```suggestion
            vote_accounts.appendAssumeCapacity(Pubkey.initRandom(random));
```

src/core/features.zig
```diff
@@ -127,6 +128,16 @@ pub const Set = struct {
                 set.active(.full_inflation_devnet_and_testnet, slot)) return true;
             return false;
         }
+
+        pub fn mainnetSlot(self: FullInflationFeatures, set: *const Set) ?Slot {
+            if (self.mainnet) return set.get(.full_inflation_mainnet_enable);
+            return null;
```
dnut: Why can't this just be a method on the feature set that checks the features?

Side note, something about these features feels off. Cluster-specific features seems like the wrong approach. I suppose I lack the context about why this approach was taken. I understand we have to use them to be compatible. It just seems unfortunate.

src/replay/rewards/inflation_rewards.zig
```diff
@@ -0,0 +1,854 @@
...
+        .new_credits_observed = new_credits_observed,
+        .force_credits_update_with_skipped_rewards = false,
+    };
+}
+
+/// Returns the slot at which inflation starts, based on the feature set.
+/// If full inflation is enabled for either mainnet or devnet/testnet, returns the
+/// earlier of the two feature activation slots. If neither full inflation feature is enabled,
+/// returns the pico inflation feature activation slot, or 0 if pico inflation is not enabled.
+pub fn getInflationStartSlot(slot: Slot, feature_set: *const FeatureSet) Slot {
```
dnut:
```suggestion
fn getInflationStartSlot(slot: Slot, feature_set: *const FeatureSet) Slot {
```

src/replay/rewards/inflation_rewards.zig
```diff
@@ -0,0 +1,854 @@
...
+        );
+    }
+
+    return if (feature_set.active(.pico_inflation, slot))
+        feature_set.get(.pico_inflation).?
+    else
+        0;
+}
+
+pub fn getInflationNumSlots(
```
dnut:
```suggestion
fn getInflationNumSlots(
```

src/replay/rewards/inflation_rewards.zig
```diff
@@ -0,0 +1,854 @@
...
+        return @min(
+            mainnet_slot orelse std.math.maxInt(Slot),
+            devnet_and_testnet_slot orelse std.math.maxInt(Slot),
+        );
+    }
+
+    return if (feature_set.active(.pico_inflation, slot))
+        feature_set.get(.pico_inflation).?
+    else
+        0;
```
dnut: This could be implemented simply without FullInflationFeatures. I'm wondering whether we need that struct. Food for thought:

```suggestion
    const mainnet_vote = feature_set.get(.full_inflation_mainnet_vote) orelse std.math.maxInt(Slot);
    const mainnet = feature_set.get(.full_inflation_mainnet_enable) orelse std.math.maxInt(Slot);
    const devnet_and_testnet = feature_set.get(.full_inflation_devnet_and_testnet) orelse std.math.maxInt(Slot);

    return if (slot >= mainnet_vote and slot >= mainnet_vote or slot >= devnet_and_testnet)
        @min(mainnet, devnet_and_testnet)
    else if (feature_set.active(.pico_inflation, slot))
        feature_set.get(.pico_inflation).?
    else
        0;
```

src/consensus/tower_state.zig
```diff
@@ -0,0 +1,95 @@
+const std = @import("std");
+const sig = @import("../sig.zig");
+
+const Slot = sig.core.Slot;
+const Lockout = sig.runtime.program.vote.state.Lockout;
+const MAX_LOCKOUT_HISTORY = sig.runtime.program.vote.state.MAX_LOCKOUT_HISTORY;
+const VoteAccount = sig.core.stakes.VoteAccount;
+
+pub const TowerVoteState = struct {
```
dnut: looks like you added this file back by accident

src/core/ReservedAccounts.zig
```diff
@@ -8,39 +8,48 @@ const Feature = sig.core.features.Feature;
 const FeatureSet = sig.core.FeatureSet;
 const Slot = sig.core.Slot;
 
-pub const ReservedAccounts = sig.utils.collections.PubkeyMap(void);
+pub const ReservedAccounts = @This();
+
+map: sig.utils.collections.PubkeyMap(void),
+
+pub const EMPTY: ReservedAccounts = .{ .map = .empty };
```
dnut: Doing `EMPTY` as uppercase feels so wrong due to the contrast with std.

src/accountsdb/buffer_pool.zig
```diff
@@ -897,6 +897,20 @@ pub const AccountDataHandle = union(enum) {
...
+    ///
+    /// Returns the number of bytes written into buf, which should be equal to end - start
+    pub fn readRange(
+        self: *const AccountDataHandle,
+        start: FileOffset,
+        end: FileOffset,
+        buf: []u8,
+    ) u32 {
+        std.debug.assert(start <= self.len());
+        std.debug.assert(end <= self.len());
```
dnut: readRange is redundant. Pass a slice to read with a length equal to `end - start` and it will have the same result as calling readRange.

src/cmd.zig
```diff
@@ -1463,11 +1466,39 @@ fn replayOffline(
...
-        errdefer allocator.free(leader_schedule);
+        errdefer allocator.free(leader_schedule_0);
 
         try epoch_context_manager.put(epoch, .{
             .staked_nodes = staked_nodes_cloned,
-            .leader_schedule = leader_schedule,
+            .leader_schedule = leader_schedule_0,
+        });
+    }
+    { // TODO: This was a quick ugly fix, needs to be handled correctly before merging
```
dnut: how would you like to address this?

src/core/bank.zig
```diff
@@ -285,6 +302,7 @@ pub const SlotState = struct {
             .stakes_cache = .{ .stakes = .init(stakes) },
             .collected_transaction_fees = .init(0),
             .collected_priority_fees = .init(0),
+            .reward_status = parent.reward_status,
```
dnut: I think this needs to be cloned.

src/replay/rewards/calculation.zig
```diff
@@ -0,0 +1,1265 @@
...
+    slots_per_year: f64,
+    previous_epoch: Epoch,
+    previous_epoch_capitalization: u64,
+    epoch_schedule: *const EpochSchedule,
+    feature_set: *const FeatureSet,
+    inflation: *const Inflation,
+    stakes_cache: *StakesCache,
+    epoch_vote_accounts: *const VoteAccounts,
+    new_warmup_and_cooldown_rate_epoch: ?Epoch,
+) !RewardsForPartitioning {
```
dnut, line 304: Since this type is only used in this file, maybe it should be defined here?

src/replay/rewards/calculation.zig
```diff
@@ -0,0 +1,1265 @@
...
+fn calculateValidatorRewards(
+    allocator: Allocator,
+    slot: Slot,
+    feature_set: *const FeatureSet,
+    rewarded_epoch: Epoch,
+    rewards: u64,
+    stakes_cache: *StakesCache,
+    epoch_vote_accounts: *const VoteAccounts,
+    new_warmup_and_cooldown_rate_epoch: ?Epoch,
+) !?ValidatorRewards {
```
dnut, line 371: Since this type is only used in this file, maybe it should be defined here?

src/replay/rewards/lib.zig
```diff
@@ -0,0 +1,299 @@
...
+        return switch (self) {
+            .calculating => |s| .{ .calculating = s.clone() },
+            .distributing => |s| .{ .distributing = s.clone() },
+        };
+    }
+};
+
+pub const EpochRewardStatus = union(enum) {
+    active: EpochRewardPhase,
+    inactive,
```
dnut, line 194: The layering of types here seems excessive. You can eliminate EpochRewardPhase and maybe consolidate StartBlockHeightAndRewards and StartBlockHeightAndPartitionedRewards.


```suggestion
pub const EpochRewardStatus = union(enum) {
    active: struct {
        distribution_start_block_height: u64,
        all_stake_rewards: PartitionedStakeRewards,
        partitioned_indices: ?PartitionedIndices,
    },
    inactive,
```

```suggestion
pub const EpochRewardStatus = union(enum) {
    calculating: StartBlockHeightAndRewards,
    distributing: StartBlockHeightAndPartitionedRewards,
    inactive,
```

src/replay/epoch_transitions.zig
```diff
@@ -0,0 +1,1224 @@
...
+                allocator,
+                slot,
+                slot_store,
+                &slot_state.capitalization,
+                epoch_constants,
+                feature_set,
+                builtin_program.program_id,
+                core_bpf_config,
+                .stateless_builtin,
+            ) catch {
```
dnut: Is it correct to ignore all errors? What about OOM?

src/core/stakes.zig
```diff
@@ -0,0 +1,1787 @@
+const builtin = @import("builtin");
```
dnut, line 1: I think you meant to delete stake.zig when you added this file, but it's still there.

src/core/genesis_config.zig
```diff
@@ -295,6 +338,14 @@ fn yearsAsSlots(years: f64, tick_duration: RustDuration, ticks_per_slot: u64) f6
     return years * SLOTS_PER_YEAR;
 }
 
+/// Floats are not ideal for consensus relevant computations.
+/// Zig and Rust implementations of pow are both complient yet can return different results.
+/// The C implementation is consistent with Rust, so we use it here for inflation rate calculations.
+/// Zig:        0.85, 4.019250798563942 -> 7.805634650110366e-2
+/// Rust:       0.85, 4.019250798563942 -> 7.805634650110367e-2
+/// C:          0.85, 4.019250798563942 -> 7.805634650110367e-2
+extern fn pow(f64, f64) f64;
```
dnut: Is this true for both glibc and musl libc? Maybe we should provide our own implementation to guarantee consistency.

src/replay/epoch_transitions.zig
```diff
@@ -0,0 +1,1224 @@
...
+                    &slot_state.capitalization,
+                );
+            }
+        }
+    }
+
+    // [agave] https://github.com/anza-xyz/agave/blob/b6c96e84b10396b92912d4574dae7d03f606da26/runtime/src/bank.rs#L5526-L5540
+    for (builtin_programs.STATELESS_BUILTINS) |builtin_program| {
+        const core_bpf_config = builtin_program.core_bpf_migration_config orelse continue;
+        if (new_feature_activations.active(core_bpf_config.enable_feature_id, 0)) {
```
dnut: 0 doesn't seem right here. Shouldn't we use the current slot, or a slot in the relevant epoch? I understand there are soeme oddities with the way activation slots are used, but I can't remember it because it's not intuitive. Could you add a comment somewhere explaining why 0 makes sense here? Either here, or if it's a more general pattern, maybe on the `active` function.

I'd like to put some more thought into feature activation slots to see if we can come up with a pattern that's more intuitive while still properly interpreting the on-chain state.

src/replay/rewards/distribution.zig
```diff
@@ -0,0 +1,937 @@
...
+        try updateSysvarAccount(
+            EpochRewards,
+            allocator,
+            epoch_rewards_sysvar,
+            .{
+                .slot = slot,
+                .slot_store = slot_store,
+                .rent = rent,
+                .capitalization = capitalization,
+            },
```
dnut: On the last slot of distribution, this performs the final distribution, and then it updates the sysvar to say rewards are inactive in the same slot (since the conditions overlap for one slot). Then every slot after that, this repeatedly updates the sysvar. Why is it updated repeatedly, and why only after distributing rewards?

src/replay/epoch_transitions.zig
```diff
@@ -0,0 +1,1224 @@
...
+    migration_config: builtin_programs.CoreBpfMigrationConfig,
+    migration_type: enum { builtin, stateless_builtin },
+) !void {
+    const loader_v3 = program.bpf_loader.v3;
+
+    const target = .{
+        .program_account = switch (migration_type) {
+            .builtin => blk: {
+                const account = (slot_store.reader()
+                    .get(allocator, builtin_program_id) catch null) orelse
```
dnut: Why is there `catch null` here? If `get` returns an error, that indicates OOM, database corruption, or some other issue that's difficult to recover from. I think these errors should be propagated. There are several other `catch null`s in this function, and the same idea applies.

src/replay/rewards/distribution.zig
```diff
@@ -0,0 +1,937 @@
...
+    allocator: Allocator,
+    stake_accounts: std.AutoArrayHashMapUnmanaged(Pubkey, Stake),
+    partitioned_reward: sig.replay.rewards.PartitionedStakeReward,
+    slot_store: SlotAccountStore,
+) !StakeReward {
+    _ = stake_accounts.get(partitioned_reward.stake_pubkey) orelse
+        return error.MissingStakeAccountInCache;
+
+    var account = try slot_store.reader().get(allocator, partitioned_reward.stake_pubkey) orelse
+        return error.MissingStakeAccountInDb;
```
dnut, line 253:
```suggestion
    var account = try slot_store.reader().get(allocator, partitioned_reward.stake_pubkey) orelse
        return error.MissingStakeAccountInDb;
    defer account.deinit(allocator);
```

src/replay/rewards/calculation.zig
```diff
@@ -0,0 +1,1265 @@
...
+    stakes_cache: *StakesCache,
+    vote_rewards: []const PartitionedVoteReward,
+    new_warmup_and_cooldown_rate_epoch: ?Epoch,
+) !void {
+    for (vote_rewards) |vote_reward| {
+        const account = (try slot_store.reader().get(allocator, vote_reward.vote_pubkey)) orelse
+            return error.MissingVoteAccount;
+        defer account.deinit(allocator);
+
+        // TODO: NO NEED TO CLONE, CHANGE CHECK AND STORE TO USE ACCOUNT
```
dnut: will this take long to address?

src/replay/rewards/calculation.zig
```diff
@@ -0,0 +1,1265 @@
...
+    slot: Slot,
+    slot_constants: SlotConstants,
+    slot_state: SlotState,
+    epoch_tracker: EpochTracker,
+
+    pub fn deinit(self: *TestEnvironment, allocator: Allocator) void {
+        self.slot_constants.deinit(allocator);
+        self.slot_state.deinit(allocator);
+        self.epoch_tracker.deinit(allocator);
+        self.account_map.deinit();
```
dnut: not a field in the struct. i guess this function is unused?

src/replay/epoch_transitions.zig
```diff
@@ -0,0 +1,1224 @@
...
+                    allocator,
+                    slot,
+                    slot_store,
+                    &slot_state.capitalization,
+                    epoch_constants,
+                    feature_set,
+                    builtin_program.program_id,
+                    core_bpf_config,
+                    .builtin,
+                ) catch {
```
dnut: In general I don't think we should ever swallow errors without being explicit about which errors are OK to swallow. I would use `catch |e| switch (e)` and list out the specific errors that indicate that it's not core bpf, and otherwise propagate any other errors with an else branch.

src/replay/rewards/calculation.zig
```diff
@@ -0,0 +1,1265 @@
...
+        const vote_account = cached_vote_accounts.getAccount(vote_pubkey) orelse continue;
+
+        const redeemed = redeemRewards(
+            rewarded_epoch,
+            stake,
+            &vote_account.state,
+            &point_value,
+            stake_history,
+            new_warmup_and_cooldown_rate_epoch,
+        ) catch continue;
```
dnut: `continue` should probably only happen for `error.NoCreditsToRedeem` right?

src/core/bank.zig
```diff
@@ -204,8 +214,12 @@ pub const SlotState = struct {
     /// 100% paid to leader
     collected_priority_fees: Atomic(u64),
 
+    /// Reward status, use to track reward distributions for N slots after an epoch boundary
+    reward_status: EpochRewardStatus,
```
dnut, line 222: SlotState now contains reference counted data, so there are important lifetime considerations. Deiniting a reference from one thread while trying to clone it into another thread could cause a race. I don't think we have this problem with the current design but I think it's worth documenting the reference counting pattern on SlotState because it's not obvious that this is happening under the hood.

src/replay/epoch_transitions.zig
```diff
@@ -0,0 +1,1224 @@
...
+        slot_state,
+        slot_store,
+        epoch_constants,
+        feature_set,
+        &new_feature_activations,
+        allow_new_activations,
+    );
+
+    if (new_feature_activations.active(.raise_block_limits_to_100m, slot)) {
+        // TODO: Implement
```
dnut: when should this be addressed?

src/core/bank.zig
```diff
@@ -295,8 +313,146 @@ pub const SlotState = struct {
...
+        if (!stake.delegation.eql(&value.getDelegation()))
+            return error.InvalidDelegation;
+
+        try stake_accounts.put(allocator, key, StakeStateV2.Stake{
+            .delegation = value,
+            .credits_observed = stake.credits_observed,
+        });
+    }
+
+    // TODO: Validate vote accounts. We should consider changing the type of VoteAccount used
```
dnut: when should this be addressed?

yewman: At the moment our stakes cache stores a minimal representation of vote accounts do lower the overhead of copying full account datas. Since this version of the stakes cache is used to deserialise, the information required to validate the vote accounts is not present. We should change to have a stakes cache which is used purely for serialisation / deserialisation and a stakes cache which is used in the runtime and convert between the two after validation. I am part way through a stakes cache refactor which will allow us to take this approach.

src/core/bank.zig
yewman: I have added comments for clarity, and explicitly handle the OutOfMemory error. If the vote account for a stake delegation is not in the cached vote accounts, then it must not be in a valid initialised state in accounts db. Hence the reason we are returning an error here if the vote account deserialisation works when it exists in accounts db but does not exist in the vote accounts cache. I am not looking for a specific error, the failure case here is if the vote account deserialisation succeeds (I have handled the OutOfMemory explicitly).
9dbd25a0

src/core/epoch_stakes.zig
yewman: 9dbd25a0

src/core/ReservedAccounts.zig
yewman: I agree, changed to lowercase. 4e89b1da

src/consensus/tower_state.zig
yewman: 4e89b1da

src/accountsdb/buffer_pool.zig
yewman: 9dbd25a

src/core/features.zig
yewman: 7c5754fa

src/replay/rewards/inflation_rewards.zig
yewman: 7c5754fa

src/replay/rewards/inflation_rewards.zig
yewman: 7c5754fa

src/cmd.zig
yewman: 7c5754fa

src/core/bank.zig
yewman: 7c5754fa

src/core/stakes.zig
yewman: 7c5754fa

src/replay/rewards/inflation_rewards.zig
yewman: 7c5754fa

src/replay/rewards/calculation.zig
yewman: Yeah, the only other error which could be returned is a RewardsOverflow error. I would probably rather that panics or something similar because it should never happen so I don't see a benefit to propagating that error.

src/replay/rewards/calculation.zig
yewman: 02c5021

src/replay/rewards/calculation.zig
yewman: 02c5021

src/replay/rewards/lib.zig
yewman: 02c5021

src/replay/rewards/distribution.zig
yewman: 02c5021

src/replay/rewards/calculation.zig
yewman: 02c5021

src/replay/epoch_transitions.zig
yewman: Fair enough, is there really a purpose to propagating OOM's. Can we actually do anything meaningfull in that case?

src/replay/rewards/calculation.zig
yewman: Alternatively I can just return the RewardsOverflow error here and continue only on the NoCreditsToRedeem error

src/core/genesis_config.zig
yewman: I believe so, @Rexicon226 can you weigh in on this?

src/replay/epoch_transitions.zig
yewman: Yeah we may as well use the slot here as it is more correct, practically speaking there is no effect since the new feature activations contain only features which where just activated. I agree there are some oddities, we could improve documentation on the feature set to detail exactly how feature activations work in Agave so that effects are clear. Maybe there is a better way to encapsulate this in the actual feature set methods but I need to look at it more closely to determine if we can make any improvements there. I think add comprehensive docs first, and see if that reveals a way we could make the methods more intuitive. I have created an issue to address this https://github.com/orgs/Syndica/projects/2?pane=issue&itemId=146011217&issue=Syndica%7Csig%7C1155

src/replay/rewards/calculation.zig
yewman: I think it would be best addressed as part of a stakes cache refactor to remove duplicate work. I can create an issue if you'd like. Lmk

src/replay/rewards/distribution.zig
yewman: At the top of the function if the epoch_rewards_status is inactive we return immediately, so it will not be updated repeatedly

src/replay/rewards/calculation.zig
yewman: f03f5269

src/replay/epoch_transitions.zig
yewman: f03f5269

src/replay/epoch_transitions.zig
yewman: f03f5269

src/replay/epoch_transitions.zig
yewman: f03f5269

src/replay/epoch_transitions.zig
yewman: f03f5269

src/core/bank.zig
yewman: 0b8d803c

src/replay/epoch_transitions.zig
yewman: I don't think we are actually rejecting blocks which violate cu limits atm. Once that is implemented it would make sense to implement this. Unless I am mistaken, in which case let me know and I can implement it before merging

src/core/bank.zig
dnut: I get the idea now, but it would be nice if the comment was more explicit about the reasoning behind these checks. The comment explains how the code works but not why it works that way.

> The stakes cache is required to contain all valid vote accounts. If there are any valid vote accounts in accounts db that are missing from the stakes cache, this means the stakes cache is invalid.

I also see that the validation is limited. It's not as comprehensive as it could be:
1. It's not checking the inverse (if the account is in the cache, it should also be in accountsdb)
2. It's not checking for *all* vote accounts, only those whose keys are listed in the stakes cache. I assume this is to keep the performance characteristics reasonable, so we don't need to iterate over every account in accountsdb. It just means the validation is more of a "best effort" check and not a full validation.

src/replay/epoch_transitions.zig
```diff
@@ -0,0 +1,1224 @@
...
+
+            if ((try slot_store.reader().get(allocator, program_data_address)) != null) {
+                return error.ProgramHasDataAccount;
+            }
+            break :blk program_data_address;
+        },
+    };
+
+    const source = blk: {
+        const buffer_account = (try slot_store.reader().get(
```
dnut, line 387: This account leaks. There are three other calls to `get` above that will also leak when they return an account

src/replay/epoch_transitions.zig
```diff
@@ -0,0 +1,1224 @@
...
+            .{},
+        ) catch return error.InvalidBufferAccount;
+
+        break :blk switch (state) {
+            .buffer => |args| .{
+                .account = buffer_account,
+                .authority = args.authority_address,
+                .data = buffer_data,
+            },
+            else => return error.InvalidBufferAccount,
```
dnut, line 416: What's the purpose of having this block that needs a break and extra errdefers instead of just directly using these variables in the outer scope? I think it would be simpler without that.

src/replay/epoch_transitions.zig
```diff
@@ -0,0 +1,1224 @@
...
+
+    // calculate_and_update_accounts_data_size_delta_off_chain
+    {
+        // These are assert()!s
+        const old_size = std.math.cast(i64, old_data_size).?;
+        const new_size = std.math.cast(i64, new_data_size).?;
+        const delta_size = new_size -| old_size;
+        if (delta_size > 0) {
+            // TODO: update accounts_data_size_delta_off_chain atomically with saturading add delta.
+        }
```
dnut: Is there a way to address these TODOs now? I think any of these options would be preferable to leaving a TODO in the code.
- Implement it.
- Add a draft to the github project explaining more context. I'm confused by the todo messages. What does it mean? Why is it needed? 
- If we don't need it, remove the word "TODO" and change the comment to something like "agave does X here but we don't need it because..."

Removing from the list of builtins seems like it might be necessary if that list is used to determine whether to execute the native implementation of a program.

Also what is this comment about asserts? Can you clarify that?

src/replay/epoch_transitions.zig
```diff
@@ -0,0 +1,1224 @@
...
+            .lt => {
+                const delta = std.math.sub(
+                    u64,
+                    lamports_to_fund,
+                    lamports_to_burn,
+                ) catch return error.ArithmeticOverflow;
+                _ = capitalization.fetchAdd(delta, .monotonic);
+            },
+            .eq => {},
+        }
```
dnut: The checked arithmetic is redundant with switch(order).

You could simplify this to:

```suggestion
        switch (std.math.order(lamports_to_burn, lamports_to_fund)) {
            .gt => _ = capitalization.fetchSub(lamports_to_burn - lamports_to_fund, .monotonic),
            .lt => _ = capitalization.fetchAdd(lamports_to_fund - lamports_to_burn, .monotonic),
            .eq => {},
        }
```

Alternatively this is even simpler. It's more atomic operations but that probably doesn't matter.

```suggestion
        _ = capitalization.fetchAdd(lamports_to_fund, .monotonic);
        _ = capitalization.fetchSub(lamports_to_burn, .monotonic);
```

src/replay/epoch_transitions.zig
```diff
@@ -0,0 +1,1224 @@
...
+        const lamports_to_burn = std.math.add(
+            u64,
+            if (target.program_account) |account| account.lamports else 0,
+            source.account.lamports,
+        ) catch return error.ArithmeticOverflow;
+        const lamports_to_fund = std.math.add(
+            u64,
+            new_target_account.lamports,
+            new_target_pda.lamports,
+        ) catch return error.ArithmeticOverflow;
```
dnut: For this checked arithmetic to fail, that means there needs to be at least 2^64 lamports in the cluster, which is not allowed. I wonder if this is the kind of thing we should be returning errors for, or just assuming it's never true (using assertions where it helps clarity or safety).

src/core/stakes.zig
```diff
@@ -0,0 +1,1787 @@
...
+            errdefer new_vote_accounts.deinit(allocator);
+            const keys = stakes.vote_accounts.vote_accounts.keys();
+            const values = stakes.vote_accounts.vote_accounts.values();
+            for (keys, values) |vote_pubkey, stake_and_vote_account| {
+                const delegated_stake = delegated_stakes.get(vote_pubkey) orelse 0;
+                try new_vote_accounts.vote_accounts.put(allocator, vote_pubkey, .{
+                    .stake = delegated_stake,
+                    .account = stake_and_vote_account.account.getAcquire(),
+                });
+            }
```
dnut: `delegated_stakes` is not needed
```suggestion
            var new_vote_accounts = VoteAccounts{};
            errdefer new_vote_accounts.deinit(allocator);
            const keys = stakes.vote_accounts.vote_accounts.keys();
            const values = stakes.vote_accounts.vote_accounts.values();
            for (keys, values) |vote_pubkey, stake_and_vote_account| {
                try new_vote_accounts.vote_accounts.put(allocator, vote_pubkey, .{
                    .stake = 0,
                    .account = stake_and_vote_account.account.getAcquire(),
                });
            }

            for (stakes.stake_accounts.values()) |stake_delegation| {
                const delegation = stake_delegation.getDelegation();
                const vote_account = new_vote_accounts.vote_accounts.getPtr(delegation.voter_pubkey).?;
                vote_account.stake += delegation.getEffectiveStake(
                    epoch,
                    &stakes.stake_history,
                    new_activation_rate_epoch,
                );
            }
```

src/replay/epoch_transitions.zig
dnut: https://github.com/orgs/Syndica/projects/2/views/10?pane=issue&itemId=149549898

src/core/bank.zig
dnut: https://github.com/orgs/Syndica/projects/2/views/10?pane=issue&itemId=149550576

src/core/genesis_config.zig
Rexicon226: It is true for both glibc and musl. I had considered just copying in our own function for it, and it's probably a good idea. The comment here is also not really good, since "Rust" and "C" mean nothing to where the actual implementation comes from. 
I can push a commit that implements our own pow().

Rexicon226: Thought about it some more, and implementing pow() ourselves likely isn't a good idea. The implementation, at least as far as I've gotten, is around 300-400 lines of code, and they are hard to verify as correct. It makes more sense to use the libc definition, as we know that any Agave compilation will also be using it.

As for the comment, it should say something like:
```
/// Zig's `std.math.pow` may return a result that is off by up to one ULP, when comparing to glibc or musl's `pow()`.
/// As these calculations affect consensus, that is an unacceptable difference for us, so we import libc's pow and 
/// use that. For reference:
/// - `std.math.pow`: pow(0.85, 4.019250798563942) -> 7.805634650110366e-2
/// - glibc/musl: pow(0.85, 4.019250798563942) -> 7.805634650110367e-2
```

conformance/src/txn_execute.zig
Sobeston, line 1327: Are we deleting this or?
edit: diff not in this branch

src/accountsdb/two/Two.zig
Sobeston, line 81: rebase needed

src/core/stakes.zig
```diff
@@ -0,0 +1,1787 @@
...
+        };
+    }
+
+    pub fn initRandom(
+        allocator: Allocator,
+        random: std.Random,
+        max_list_entries: usize,
+    ) Allocator.Error!VoteAccounts {
+        // TODO: Uncomment once not required by bank init random
+        // if (!builtin.is_test) @compileError("only for testing");
```
Sobeston: Are we fixing before merging or after

src/core/stakes.zig
```diff
@@ -0,0 +1,1787 @@
...
+    }
+
+    pub fn getStake(self: StakeAccount) Stake {
+        return self.stake;
+    }
+};
+
+pub const VoteAccount = struct {
+    // account: AccountSharedData,
+    account: MinimalAccount,
```
Sobeston, line 744: commented line

src/core/stakes.zig
```diff
@@ -0,0 +1,1787 @@
...
+            allocator,
+            pubkey,
+            account_0,
+            .init(.delegation, &stakes, null),
+        );
+        defer maybe_old.?.deinit(allocator);
+
+        try std.testing.expectEqual(42, maybe_old.?.stake);
+        // try std.testing.expect(account_0.equals(&maybe_old.?.account));
+        // try std.testing.expect(account_0.equals(&vote_accounts.getAccount(pubkey).?));
```
Sobeston: Lots of commented out expects in this file

src/replay/rewards/calculation.zig
```diff
@@ -0,0 +1,1264 @@
...
+    );
+
+    const MAX_FACTOR_OF_REWARD_BLOCKS_IN_EPOCH: u64 = 10;
+    const max_chunks = try std.math.divFloor(
+        u64,
+        epoch_schedule.slots_per_epoch,
+        MAX_FACTOR_OF_REWARD_BLOCKS_IN_EPOCH,
+    );
+
+    return if (num_chunks < 1) 1 else if (num_chunks > max_chunks) max_chunks else num_chunks;
```
Sobeston:
```zig
return @min(@max(num_chunks, 1), max_chunks);
```

src/replay/rewards/calculation.zig
```diff
@@ -0,0 +1,1264 @@
...
+    feature_set: *const FeatureSet,
+    stakes: *const Stakes(.stake),
+) !FilteredStakesDelegations {
+    var result = FilteredStakesDelegations{};
+    if (feature_set.active(.stake_minimum_delegation_for_rewards, slot)) {
+        const min_delegation = @max(sig.runtime.program.stake.getMinimumDelegation(
+            slot,
+            feature_set,
+        ), 1_000_000_000); // LAMPORTS_PER_SOL
+
```
Sobeston, line 413: (feel free to ignore nit): should probably define lamports_per_sol somewhere, we use `1_000_000_000` >50 times in the codebase

src/replay/rewards/calculation.zig
```diff
@@ -0,0 +1,1264 @@
...
+        for (stakes.stake_accounts.keys(), stakes.stake_accounts.values()) |key, value| {
+            if (value.delegation.stake >= min_delegation) {
+                try result.append(allocator, .{ .pubkey = key, .stake = value });
+            }
+        }
```
Sobeston, line 417: (feel free to ignore nit)
```zig
        for (stakes.stake_accounts.keys(), stakes.stake_accounts.values()) |key, value| {
            if (value.delegation.stake < min_delegation) continue;
            try result.append(allocator, .{ .pubkey = key, .stake = value });
        }
```

src/replay/rewards/calculation.zig
```diff
@@ -0,0 +1,1264 @@
...
+) !ValidatorRewards {
+    var vote_account_rewards_map = std.AutoArrayHashMapUnmanaged(Pubkey, VoteReward).empty;
+    defer vote_account_rewards_map.deinit(allocator);
+    try vote_account_rewards_map.ensureTotalCapacity(
+        allocator,
+        cached_vote_accounts.vote_accounts.count(),
+    );
+
+    var partitioned_stake_rewards = std.ArrayListUnmanaged(PartitionedStakeReward){};
+    errdefer partitioned_stake_rewards.deinit(allocator);
```
Sobeston: (feel free to ignore nit): This can be a regular defer, it's only persisted if .toOwnedSlice is used.

src/replay/rewards/inflation_rewards.zig
```diff
@@ -0,0 +1,849 @@
...
+        // If is_split is true, there should be lamports allocated to both staker and voter.
+        return (self.staker_rewards == 0 or self.voter_rewards == 0) and self.is_split;
+    }
+};
+
+pub fn commissionSplit(commission: u8, rewards: u64) CommissionSplit {
+    return switch (@min(commission, 100)) {
+        0 => .{ .voter_rewards = 0, .staker_rewards = rewards, .is_split = false },
+        100 => .{ .voter_rewards = rewards, .staker_rewards = 0, .is_split = false },
+        else => |split| .{
```
Sobeston, line 132: Was this meant to be a `@max`? The 0 branch is unreachable

src/replay/epoch_transitions.zig
```diff
@@ -0,0 +1,1224 @@
...
+/// This ensures that the epoch constants are available at the point of transition.
+pub fn updateEpochStakes(
+    allocator: Allocator,
+    slot: Slot,
+    stakes_cache: *StakesCache,
+    epoch_tracker: *EpochTracker,
+) !void {
+    const leader_schedule_epoch = epoch_tracker.schedule.getLeaderScheduleEpoch(slot);
+    if (!epoch_tracker.epochs.contains(leader_schedule_epoch)) {
+        const current = epoch_tracker.getForSlot(slot) orelse {
```
Sobeston: (feel free to ignore nit): would prefer to get rid of the function-wide nest here
```zig
    const leader_schedule_epoch = epoch_tracker.schedule.getLeaderScheduleEpoch(slot);
    if (epoch_tracker.epochs.contains(leader_schedule_epoch)) return;
    const current = epoch_tracker.getForSlot(slot) orelse return error.EpochConstantsNotFound;
```

src/replay/epoch_transitions.zig
```diff
@@ -0,0 +1,1224 @@
...
+
+        var ancestors = Ancestors.EMPTY;
+        errdefer ancestors.deinit(allocator);
+
+        var epoch_tracker: EpochTracker = .{ .schedule = .INIT };
+        errdefer epoch_tracker.deinit(allocator);
+
+        var epoch_constants = EpochConstants.genesis(genesis_config);
+        errdefer epoch_constants.deinit(allocator);
+        try epoch_tracker.put(allocator, 0, epoch_constants);
```
Sobeston: Slightly misleading grouping here - the `epoch_tracker` put is grouped with the `epoch_constants` init code

src/replay/freeze.zig
Sobeston, line 393: rebase needed

Sobeston (changes requested):

src/replay/epoch_transitions.zig
yewman: @kprotty did you have a reason behind this?

src/replay/epoch_transitions.zig
yewman: I can see where you are coming from. there are probably quite a few places where we are returning errors for addition overflows which you may be able to make a similar argument. I don't have a strong preference in this case. It was implemented to conform to the error returns of Agave, it is used in the conformance harnesses, however, I am not certain whether this error can actually be hit within those harnesses.

src/replay/epoch_transitions.zig
yewman: I prefer the block since it groups a set of related logic, I am happy to flatten if you feel strongly.

src/replay/epoch_transitions.zig
yewman: We don't track it anywhere because it was deemed non 'essential' and not added to slot state. I have created an issue to add it and implement the deltas everywhere it is required. I think it makes sense to add it in an isolated PR and comprehensively reviewing everywhere it needs to be incremented. 

I have added a better comment regarding asserts.

fb902d7

src/replay/epoch_transitions.zig
yewman: fb902d7

src/replay/epoch_transitions.zig
yewman: Good suggestion.
fb902d7

src/core/stakes.zig
yewman: 0fcead80

src/core/stakes.zig
yewman: 0fcead80

src/core/stakes.zig
yewman: 0fcead80

src/replay/rewards/calculation.zig
yewman: I agree, I have added it as an issue to address since it will be applied in many places outside the scope of this PR.
https://github.com/orgs/Syndica/projects/2?pane=issue&itemId=149667910

src/replay/rewards/inflation_rewards.zig
yewman: What if commission is 0?

src/replay/epoch_transitions.zig
yewman: fb902d7a

src/replay/epoch_transitions.zig
yewman: fb902d7a

src/replay/rewards/calculation.zig
yewman: c9b6ffa0

src/replay/rewards/inflation_rewards.zig
Sobeston: Oops I misread, all is fine

src/replay/rewards/calculation.zig
yewman: c9b6ffa0

src/replay/rewards/calculation.zig
yewman: 3e5300c

src/core/stakes.zig
yewman: bc8fcc5d

src/core/genesis_config.zig
yewman: 1e2e9fa9

src/replay/epoch_transitions.zig
dnut: I'll just quickly share what I remember of my initial impression. I would need to reread the code and think more deeply to understand how it groups related code, but I don't think it's worth investing much more time into this. I trust that you see a real advantage, so feel free to resolve this conversation.

It seemed like an arbitrary grouping to me. Usually when I see blocks like this, it's for assembling some state where the block cleanly groups together some necessary errdefers in a way that simplifies the code. But in this case I think it makes the code more complex. You need more errdefers, not less, and it spends code constructing an anonymous type to `break` that simply repackages most of the state from inside the block. I didn't sense a clean encapsulation or abstraction here.

src/core/bank.zig
yewman: Yeah I understand where you are coming from. I have made some more changes which hopefully provide better clarity 5b720430
1. The inverse is now checked under vote account validation 
2. This is because the check pertains only to the validity of vote accounts referenced within stake accounts. Vote accounts are validated below (previously they were not, it was just a comment. Now I have added partial validation and updated comments)

Sobeston (approved):

src/replay/epoch_transitions.zig
```diff
@@ -0,0 +1,1208 @@
...
+) !void {
+    const account_data_len = account.data.len;
+    _ = capitalization.fetchSub(account.lamports, .monotonic);
+    var acc = account;
+    acc.lamports = 0;
+    @memset(acc.data, 0);
+    try store.put(address, acc);
+    // NOTE: update account size delta in slot state?
+    _ = account_data_len;
+}
```
ajw221, line 590: what is the point of this function and where it's being called?

src/core/bank.zig
```diff
@@ -295,8 +317,183 @@ pub const SlotState = struct {
...
+
+        const db_versioned_vote_state = try sig.bincode.readFromSlice(
+            allocator,
+            sig.runtime.program.vote.state.VoteStateVersions,
+            db_account_data,
+            .{},
+        );
+        const db_vote_state = try db_versioned_vote_state.convertToCurrent(allocator);
+
+        if (!db_vote_state.equals(&cached_vote_state)) return error.InvalidVoteAccount;
```
ajw221: Should either `db_versioned_vote_state` or `db_vote_state` be cleaned up here? Also is there a non allocating way to do this comparison?

src/replay/rewards/calculation.zig
```diff
@@ -0,0 +1,1265 @@
...
+    };
+}
+
+fn storeVoteAccountsPartitioned(
+    allocator: Allocator,
+    slot_store: SlotAccountStore,
+    stakes_cache: *StakesCache,
+    vote_rewards: []const PartitionedVoteReward,
+    new_warmup_and_cooldown_rate_epoch: ?Epoch,
+) !void {
```
ajw221, line 252: we should potentially add a comment on this function noting that `vote_rewards` should all exist in `slot_store`, because if this function fails anywhere after we've proceeded past the first iteration, we have no way of recovering or undoing this process

src/runtime/sysvar/stake_history.zig
```diff
@@ -66,6 +78,22 @@ pub const StakeHistory = struct {
...
+            epoch,
+            Entry.searchCmp,
+        );
+
+        if (self.entries.len == MAX_ENTRIES) {
+            if (epoch < self.entries.buffer[MAX_ENTRIES - 1].epoch) return;
+            _ = self.entries.orderedRemove(MAX_ENTRIES - 1);
+        }
+
+        try self.entries.insert(index, .{ .epoch = epoch, .stake = entry });
```
ajw221, line 98: should this return an error for duplicate entry? if not, should it mutate the current state of the stake's entry to avoid needing to call `.insert`?

src/replay/rewards/calculation.zig
yewman: If this fails the runtime should fail because we will end up in a corrupt state. This goes to some of the discussion we've had about error handling and having a consistent approach with respect to how we propagate unrecoverable errors. We need to make a decision regarding the approach we take here. I am in favour of failing hard and fast at the relevant call site unless we will actually handle the associated errors. With process isolation this should still allow other processes to shutdown normally in the event a single process encounters an irrecoverable error.

src/runtime/sysvar/stake_history.zig
yewman: 8b04fc37
Yes good catch

src/core/bank.zig
yewman: 8b04fc37
Nice catch

src/replay/epoch_transitions.zig
yewman: 8b04fc37
We need to write an entry with zero account lamports to effectively 'delete' an account. It is called during activations when updating builtin programs and precompiles. If the existing accounts are in an invalid state, they need to be 'deleted' and their lamports burnt

dnut (approved):

PR #1005, opened 2025-10-16
refactor(consensus:vote_listener): Rename VoteListener to VoteCollector, make single-threaded

InKryption:


src/consensus/vote_listener.zig
dadepo, line 909: From what I can see, the two point of synchronization between vote collector and replay/consensus are:

- Gossip vote hashes sent and consumed by replay via `process_gossip_verified_vote_hashes`
- Duplicate-confirmed slot notifications sent and consumed by replay via `process_duplicate_confirmed_slots`.

And this improvement to make vote collector single threaded hugely simplifies that point of synchronization.

The only observation I have, due to my recent look into how replay/consensus ties into rpc, is the fact that the part of the vote collector that now notifies rpc is now tied to be in lockstep with replay's progression.

One of such places is here sending through `senders.bank_notification` and the other is on line 954 below, sending through `senders.subscriptions.notifyVote`.

Perhaps, with the hook design it might even be easier to properly move out those notification? 

But in the meantime, if the above observation is a valid one, perhaps a TODO would be helpful.

src/consensus/vote_listener.zig
```diff
@@ -399,120 +332,106 @@ pub const VoteListenerMetrics = struct {
...
+            .latest_vote_slot_per_validator = .empty,
+            .last_process_root = now,
+            .vote_processing_time = .ZEROES,
+            .metrics = try .init(registry),
+        };
+    }
 
-    var latest_vote_slot_per_validator: std.AutoArrayHashMapUnmanaged(Pubkey, Slot) = .empty;
-    defer latest_vote_slot_per_validator.deinit(allocator);
+    pub fn collectAndProcessVotes(
```
dadepo, line 372: I might be missing something, but this looks like the entry point to exercise the vote collector functionality but I do not see it being called in replay/consensus code? I only see it being called in tests.

InKryption: You are correct actually, sort of slipped my mind to make it part of this PR, but it is re-introduced in #1029 (on top of this one).

src/consensus/vote_listener.zig
InKryption: Added TODO comments on the relevant fields of `Senders`.

src/consensus/vote_listener.zig
dadepo: Gotcha. But I think we should include that in this PR? if it is being merged to main right now. If not then the vote collector functionality is essentially disabled until #1029 lands on main.

InKryption: That would be kind of hard to re-structure at this point, would be more feasible to first review #1029 and merge it into this one.

PR #1007, opened 2025-10-16
feat(replay): epoch transition - migrate core builtins to bpf  

kprotty:


src/replay/apply_feature_activations.zig
```diff
@@ -167,13 +183,248 @@ fn applyBuiltinProgramFeatureTransitions(
...
+            .builtin => blk: {
+                const account = (slot_store.get(allocator, builtin_program_id) catch null) orelse
+                    return error.AccountNotFound;
+                if (!account.owner.equals(&sig.runtime.ids.NATIVE_LOADER_ID))
+                    return error.IncorrectOwner;
+                break :blk account;
+            },
+            .stateless_builtin => blk: {
+                if ((slot_store.get(allocator, builtin_program_id) catch null) != null)
+                    return error.AcocuntExists;
```
yewman:
```suggestion
                    return error.AccountExists;
```

src/replay/apply_feature_activations.zig
```diff
@@ -167,13 +183,248 @@ fn applyBuiltinProgramFeatureTransitions(
...
+        source.account.data.len(),
+    ) catch return error.ArithmeticOverflow;
+    const new_data_size = std.math.add(
+        u64,
+        new_target_account.data.len,
+        new_target_pda.data.len,
+    ) catch return error.ArithmeticOverflow;
+
+    // NOTE: duplicated from src/runtime/program/bpf_loader/execute.zig:deployProgram
+    {
```
yewman, line 333: This functionality probably needs to stay in sync, do you think we could abstract it into a function?

kprotty: https://github.com/Syndica/sig/pull/1007/commits/fa6303c6f5d53884680465096a240ff5e317a4d6

src/replay/apply_feature_activations.zig
kprotty: https://github.com/Syndica/sig/pull/1007/commits/fa6303c6f5d53884680465096a240ff5e317a4d6

PR #1010, opened 2025-10-17
feat(consensus): Support Voting

dadepo:


src/cmd.zig
```diff
@@ -1119,27 +1121,36 @@ fn validator(
...
+                        epoch_schedule.slots_per_epoch,
+                        try collapsed_manifest.epochVoteAccounts(epoch),
+                    )
+                else
+                    try LeaderSchedule.fromStakedNodes(
+                        allocator,
+                        epoch,
+                        epoch_schedule.slots_per_epoch,
+                        staked_nodes,
+                    );
```
InKryption:
```suggestion
                const ls_slots: LeaderSchedule = try .fromVoteAccounts(
                    allocator,
                    epoch,
                    epoch_schedule.slots_per_epoch,
                    if (true) try collapsed_manifest.epochVoteAccounts(epoch) else staked_nodes,
                );
```

src/cmd.zig
```diff
@@ -1119,27 +1121,36 @@ fn validator(
...
+                const ls_struct = LeaderSchedule{
+                    .allocator = allocator,
+                    .slot_leaders = ls_slots,
+                };
+                try leader_schedule_cache.put(epoch, ls_struct);
```
InKryption:
```suggestion
                try leader_schedule_cache.put(epoch, .{
                    .allocator = allocator,
                    .slot_leaders = ls_slots,
                });
```

src/replay/consensus/core.zig
```diff
@@ -878,6 +974,407 @@ fn pushVote(
...
+        .V4 => network.Socket.create(.ipv4, .udp) catch |err| {
+            logger.err().logf("Failed to create socket for vote send: {}", .{err});
+            return err;
+        },
+        .V6 => network.Socket.create(.ipv6, .udp) catch |err| {
+            logger.err().logf("Failed to create socket for vote send: {}", .{err});
+            return err;
+        },
+    };
+    defer socket.close();
```
InKryption:
```suggestion
    const addr_family: network.AddressFamily = switch (tpu_address) {
        .V4 => .ipv4,
        .V6 => .ipv6,
    };
    const socket = network.Socket.create(addr_family, .udp) catch |err| {
        logger.err().logf("Failed to create socket for vote send: {}", .{err});
        return err;
    };
    defer socket.close();
```
Though I think this is the wrong place to do this; I think a pair of ipv4 and ipv6 UDP sockets should be created outside of the loop and passed as parameters to this function, with the one appropriate for the tpu_address used.

src/replay/consensus/core.zig
```diff
@@ -878,6 +974,407 @@ fn pushVote(
...
+    }
+
+    const gossip_table, var gossip_table_lg = gossip_table_rw.readWithLock();
+    defer gossip_table_lg.unlock();
+
+    var seen_sockets: std.AutoHashMapUnmanaged(SocketAddr, void) = .empty;
+    defer seen_sockets.deinit(allocator);
+
+    var leader_sockets: std.ArrayListUnmanaged(SocketAddr) = .empty;
+    defer leader_sockets.deinit(allocator);
```
InKryption: Small suggestion: remove `leader_sockets`, and make the return be `return try allocator.dupe(SocketAddr, seen_sockets.keys());`

src/replay/consensus/core.zig
```diff
@@ -878,6 +974,407 @@ fn pushVote(
...
+
+fn sendVoteToGossip(
+    vote_op: VoteOp,
+    gossip_table_rw: *sig.sync.RwMux(sig.gossip.GossipTable),
+    my_keypair: sig.identity.KeyPair,
+) !void {
+    const gossip_table, var gossip_table_lock = gossip_table_rw.writeWithLock();
+    defer gossip_table_lock.unlock();
+
+    const now = sig.time.getWallclockMs();
```
InKryption: Would prefer to make this a parameter that can be sent down from `sendVote`, so tests can supply this as a deterministic value.

src/replay/consensus/core.zig
```diff
@@ -2205,4 +2702,1018 @@ test "computeBankStats - same weight selects lower slot" {
...
+        allocator,
+        vote_account_pubkey,
+        empty_keypairs,
+        node_kp,
+        .same_fork,
+        &replay_tower,
+        account_reader,
+        &fixture.slot_tracker,
+        &epoch_tracker,
+    );
```
InKryption, line 2853: Should still `errdefer result.deinit(allocator)` to avoid a leak in the event that the expectation below is incorrect and, for whatever reason, `result == .tx` - same comment applies to all other instances of this in the other tests.

src/replay/consensus/core.zig
```diff
@@ -2205,4 +2702,1018 @@ test "computeBankStats - same weight selects lower slot" {
...
+    var fixture = try TestFixture.init(allocator, root);
+    defer fixture.deinit(allocator);
+
+    fixture.slot_tracker.get(0).?.state.hash.set(root.hash);
+
+    var replay_tower = try createTestReplayTower(1, 0.67);
+    defer replay_tower.deinit(allocator);
+
+    replay_tower.last_vote = sig.consensus.vote_transaction.VoteTransaction{
+        .tower_sync = sig.runtime.program.vote.state.TowerSync{
```
InKryption: Any reason for these type specifications? These could both be written as `.{`. Same comment in general for all instances of this.

src/replay/consensus/core.zig
dadepo: Unfortunately `keys` is not marked 'pub' on `std.AutoHashMapUnmanaged`.

src/cmd.zig
dadepo: Processed this comment and all the other comments in the commit [9458e92](https://github.com/Syndica/sig/pull/1010/commits/9458e92c93a264e304b7575329cac70b79f1ab8e)

src/replay/consensus/core.zig
dadepo: 79a8f6d36456c0a79c9429fd4c1a6772231f46c2

src/replay/consensus/core.zig
InKryption: I would suggest just using `AutoArrayHashMapUnmanaged` then

src/replay/consensus/core.zig
InKryption: I see you replaced `VoteTransaction` but not `TowerSync`; any reason for that? I was kind of referring to both of them. I would in general avoid the `T{` syntax, because there is an extremely high likelihood it will be removed from the language: https://github.com/ziglang/zig/issues/5038 - despite the lack of an accepted label, Andrew and others in the core team have shown strong support for it, and are writing code and refactoring the stdlib in a way that seems to imply it is likely it will be implemented.

src/replay/consensus/core.zig
dadepo: [779bf4b](https://github.com/Syndica/sig/pull/1010/commits/779bf4b2bfe0d94aedf40d94767be30302a38d69)

src/replay/consensus/core.zig
dadepo: That was me just being in a haste. Updated [779bf4b](https://github.com/Syndica/sig/pull/1010/commits/779bf4b2bfe0d94aedf40d94767be30302a38d69)

src/replay/consensus/core.zig
```diff
@@ -82,6 +103,10 @@ pub const TowerConsensus = struct {
     receivers: Receivers,
     verified_vote_channel: *Channel(VerifiedVote),
 
+    // Sending votes
+    gossip_table: ?*sig.sync.RwMux(sig.gossip.GossipTable),
+    leader_schedule: ?*sig.core.leader_schedule.LeaderScheduleCache,
```
dnut: Depend on SlotLeaders, not *LeaderScheduleCache

src/replay/consensus/core.zig
```diff
@@ -844,30 +891,80 @@ fn handleVotableBank(
...
         replay_tower,
+        slot_tracker,
+        epoch_tracker,
+        account_reader,
+        node_keypair,
+        authorized_voter_keypairs,
+        vote_account_pubkey,
+        switch_fork_decision,
+        gossip_table_rw,
+        leader_schedule,
```
dnut: The sheer number of parameters for this chain of function calls feels excessive. Inlining handleVotableBank and pushVote into the sole call site would clean up many lines of code. Is there a benefit to keeping these as separate functions?

src/replay/consensus/core.zig
```diff
@@ -878,6 +975,414 @@ fn pushVote(
...
+
+    var leader_iter = seen_leaders.keyIterator();
+    while (leader_iter.next()) |leader_pubkey_ptr| {
+        const leader_pubkey = leader_pubkey_ptr.*;
+        const contact_info = gossip_table.getThreadSafeContactInfo(leader_pubkey) orelse continue;
+
+        const socket_addr = contact_info.tpu_addr orelse continue;
+
+        try seen_sockets.put(allocator, socket_addr, {});
+    }
```
dnut: let's try to minimize contention on the gossip table

```suggestion
    var seen_sockets: std.AutoArrayHashMapUnmanaged(SocketAddr, void) = .empty;
    try seen_sockets.ensureTotalCapacity(allocator, seen_leaders.count());
    defer seen_sockets.deinit(allocator);

    {
        const gossip_table, var gossip_table_lg = gossip_table_rw.readWithLock();
        defer gossip_table_lg.unlock();

        var leader_iter = seen_leaders.keyIterator();
        while (leader_iter.next()) |leader_pubkey_ptr| {
            const leader_pubkey = leader_pubkey_ptr.*;
            const contact_info = gossip_table.getThreadSafeContactInfo(leader_pubkey) orelse continue;

            const socket_addr = contact_info.tpu_addr orelse continue;

            seen_sockets.putAssumeCapacity(socket_addr, {});
        }
    }
```

src/replay/consensus/core.zig
```diff
@@ -57,11 +66,23 @@ const check_slot_agrees_with_cluster =
 
 const MAX_VOTE_REFRESH_INTERVAL_MILLIS: usize = 5000;
 
+pub const VoteOp = union(enum) {
+    push_vote: struct {
+        tx: Transaction,
+        tower_slots: []Slot,
```
dnut: This field is populated but does not appear to be used anywhere, and I believe it also leaks memory because I don't see it being freed anywhere except an errdefer.

src/replay/consensus/core.zig
```diff
@@ -878,6 +975,414 @@ fn pushVote(
...
+        logger.err().logf("Failed to create IPv4 socket for vote send: {}", .{err});
+        return err;
+    };
+    defer ipv4_socket.close();
+
+    const ipv6_socket = network.Socket.create(.ipv6, .udp) catch |err| {
+        logger.err().logf("Failed to create IPv6 socket for vote send: {}", .{err});
+        return err;
+    };
+    defer ipv6_socket.close();
```
dnut: it seems wasteful to create these for each vote. can we init them once and reuse them?

src/replay/consensus/core.zig
dadepo: [31e2a35](https://github.com/Syndica/sig/pull/1010/commits/31e2a351563e99f9700bcf691d2441232befc247)

src/replay/consensus/core.zig
dadepo: Inlined [49bc11b](https://github.com/Syndica/sig/pull/1010/commits/49bc11b5ed574f2753936be8ad4bd7a1e43dad3b)

src/replay/consensus/core.zig
dadepo: [289b29a](https://github.com/Syndica/sig/pull/1010/commits/289b29a4b310a238e714e43dcc2aae00ee697ed1)

src/cmd.zig
```diff
@@ -1119,27 +1133,27 @@ fn validator(
...
-                    staked_nodes,
+                    if (true) try collapsed_manifest.epochVoteAccounts(epoch) else staked_nodes,
                 );
-        };
+                try leader_schedule_cache.put(epoch, .{
+                    .allocator = allocator,
+                    .slot_leaders = ls_slots,
+                });
+                break :ls ls_slots;
+            };
```
dnut: The epoch context manager is a leader schedule provider. You didn't need to add an instance of LeaderScheduleCache or make any of these changes.

src/cmd.zig
```diff
@@ -1196,13 +1210,24 @@ fn validator(
         &app_base,
         &ledger,
         &epoch_context_manager,
+        cfg.voting_enabled,
     );
     defer replay_deps.deinit(allocator);
 
+    const maybe_vote_sockets: ?replay.consensus.core.VoteSockets = if (cfg.voting_enabled)
+        replay.consensus.core.VoteSockets.init() catch null
```
dnut: Why are you hiding this error? Shouldn't the validator fail if it's supposed to be voting but it can't even open the socket for voting?

src/cmd.zig
```diff
@@ -1329,13 +1354,24 @@ fn replayOffline(
         &app_base,
         &ledger,
         &epoch_context_manager,
+        cfg.voting_enabled,
     );
     defer replay_deps.deinit(allocator);
 
+    const maybe_vote_sockets: ?replay.consensus.core.VoteSockets = if (cfg.voting_enabled)
```
dnut: Replay offline should not have networking.

dadepo: [266c672](https://github.com/Syndica/sig/pull/1010/commits/266c67208ef639b4b5c64770660955061fca3b21)

src/cmd.zig
dadepo: [266c672](https://github.com/Syndica/sig/pull/1010/commits/266c67208ef639b4b5c64770660955061fca3b21)

src/cmd.zig
dadepo: [266c672](https://github.com/Syndica/sig/pull/1010/commits/266c67208ef639b4b5c64770660955061fca3b21)

dnut: You still have it using leader_schedule_cache, it's just not being populated.

src/replay/consensus/core.zig
dadepo: This has been processed but forgot to reply with the commit 33d1bf74b10337203140ef699c1eaf2aa074a1a6

src/replay/consensus/core.zig
dadepo: This has been processed but forgot to reply with the commit [ee05539](https://github.com/Syndica/sig/pull/1010/commits/ee05539483c669e2669236d18fdbfe7470df0b41) - had the idea of passing the last slot instead of the slice, but can do that or leave as is

src/cmd.zig
dadepo: I reverted the changes. Thought that addressed the original comment. Will look at it again.

Edit: Are you suggesting to pass  down `epoch_context_manager` as the dependencies and use that to get the slot leaders instead of `leader_schedule_cache.slotLeaders()`

dnut: You didn't revert your change completely. On main we don't use a leader schedule cache.

I'm saying the epoch context manager is a slot leader provider. Call `epoch_context_manager.slotLeaders()` and pass that down.

src/replay/consensus/core.zig
dnut: good idea. let's avoid allocations whenever possible. something like this seems right (not sure about the name):
```zig
last_tower_slot: ?Slot,
```

src/cmd.zig
dadepo: [1fc8a63](https://github.com/Syndica/sig/pull/1010/commits/1fc8a63595fd7f9105ac58faffdb38178c9f3996)

src/replay/consensus/core.zig
dadepo: [d41de60](https://github.com/Syndica/sig/pull/1010/commits/d41de60aeadcea844abfbf0f65c9052a43a77560)

InKryption (approved):

PR #1027, opened 2025-10-22
feat(consensus): Send duplicated slot from shred network to consensus

dadepo:


src/ledger/Reader.zig
```diff
@@ -2681,3 +2691,34 @@ test getConfirmedSignaturesForAddress {
...
+    defer state.deinit();
+
+    const reader = state.reader();
+    const result_writer = state.resultWriter();
+
+    // Test case: Slot with no duplicate proof returns false
+    {
+        const slot: Slot = 42;
+        const is_duplicate = try reader.isDuplicateSlot(slot);
+        try std.testing.expectEqual(false, is_duplicate);
```
InKryption: Can give a more informative error message that reflects the fact that we "expect" this to be false, and not an error. Not always applicable, but nice to do when possible.
```suggestion
        try std.testing.expectEqual(false, reader.isDuplicateSlot(slot));
```

src/ledger/ResultWriter.zig
```diff
@@ -90,6 +90,22 @@ pub fn setDeadSlot(
...
+    self: *const ResultWriter,
+    slot: Slot,
+    shred1: []const u8,
+    shred2: []const u8,
+) !void {
+    const duplicate_slot_proof = ledger_mod.meta.DuplicateSlotProof{
+        .shred1 = shred1,
+        .shred2 = shred2,
+    };
+    try self.ledger.db.put(schema.duplicate_slots, slot, duplicate_slot_proof);
```
InKryption:
```suggestion
    try self.ledger.db.put(schema.duplicate_slots, slot, .{
        .shred1 = shred1,
        .shred2 = shred2,
    });
```

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,794 @@
...
+const DuplicateShredData = sig.gossip.data.DuplicateShred;
+
+const ResultWriter = sig.ledger.ResultWriter;
+const LedgerReader = sig.ledger.Reader;
+const DuplicateSlotProof = sig.ledger.meta.DuplicateSlotProof;
+
+pub const GOSSIP_SLEEP_MILLIS: u64 = 100;
+pub const MAX_NUM_CHUNKS: usize = 3;
+pub const MAX_NUM_ENTRIES_PER_PUBKEY: usize = 128;
+pub const BUFFER_CAPACITY: usize = 512 * MAX_NUM_ENTRIES_PER_PUBKEY;
```
InKryption: Any reason these are `pub`?

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,794 @@
...
+const Key = struct { slot: Slot, from: Pubkey };
+
+const BufferEntry = struct {
+    chunks: [MAX_NUM_CHUNKS]?[]u8,
+};
+
+pub fn recvLoop(
+    allocator: Allocator,
+    logger: Logger,
+    params: Params,
```
InKryption: This all seems super similar to the vote listener, do you think it would make sense to unroll it into a single-threaded construct that can then be manually placed in another thread running synchronously with other processes? Specifically the fact that both the vote listener and this shred listener have a cursor they're using to go over the gossip table makes it seem like to me they could operate on the same thread and share a cursor, which could potentially reduce contention on the gossip table (since it would reduce the number of possible readers any given writer is waiting for by one).

lmk your thoughts.

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,794 @@
...
+        const new_root = self.params.ledger_reader.maxRoot();
+        if (new_root == self.last_root) return;
+        self.last_root = new_root;
+        const epoch = self.params.epoch_schedule.getEpoch(self.last_root);
+        self.cached_slots_in_epoch = self.params.epoch_schedule.getSlotsInEpoch(epoch);
+    }
+
+    fn shouldConsumeSlot(self: *DuplicateShredHandler, slot: Slot) !bool {
+        const max_slot = self.last_root +| self.cached_slots_in_epoch;
+        if (!(slot > self.last_root and slot < max_slot)) return false;
```
InKryption: For just a little added clarity
```suggestion
        const slot_in_range = slot > self.last_root and slot < max_slot;
        if (!slot_in_range) return false;
```

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,794 @@
...
+    params: Params,
+    // Because we use UDP for packet transfer, we can normally only send ~1500 bytes
+    // in each packet. We send both shreds and meta data in duplicate shred proof, and
+    // each shred is normally 1 packet(1500 bytes), so the whole proof is larger than
+    // 1 packet and it needs to be cut down as chunks for transfer. So we need to piece
+    // together the chunks into the original proof before anything useful is done.
+    dup_buffer: std.AutoHashMapUnmanaged(Key, BufferEntry),
+    consumed: std.AutoHashMapUnmanaged(Slot, bool),
+    last_root: Slot,
+    cached_slots_in_epoch: u64,
```
InKryption, line 115: iiuc `consumed` and `last_root` are also "cached" state, yeah? If so, would be nice to annotate that, preferably by reflecting that in their names - at the very least a minimal high level description of how they are updated would be appreciated.

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,794 @@
...
+        }
+        return !gop.value_ptr.*;
+    }
+
+    fn maybePruneBuffer(self: *DuplicateShredHandler) void {
+        if (self.dup_buffer.count() < BUFFER_CAPACITY * 2) return;
+
+        var counts: std.AutoHashMapUnmanaged(Pubkey, usize) = .empty;
+        defer counts.deinit(self.allocator);
+        var keys_to_remove = std.ArrayListUnmanaged(Key){};
```
InKryption: remember, the default field init will be removed from arraylist, use the `.empty` decl.

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,794 @@
...
+        defer keys_to_remove.deinit(self.allocator);
+
+        var it = self.dup_buffer.keyIterator();
+        while (it.next()) |key_ptr| {
+            const key = key_ptr.*;
```
InKryption:
```suggestion
        for (self.dup_buffer.keys()) |key| {
```
in addition to this, I suggest making `dup_buffer` an ArrayHashMap; any time you need to iterate over a hash map with any amount of frequency, it's basically always preferable to use ArrayHashMap to save on iteration performance.

Apart from that, it will also be beneficial to use an ArrayHashMap here because of the pattern of removals, which the current-day stdlib implementation of normal HashMap handles pretty badly without manual calls to `rehash`, whereas ArrayHashMap has consistent performance regardless of removal patterns. See the doc comment on `std.HashMap(...).rehash` for some context on that.

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,794 @@
...
+                keys_to_remove.append(self.allocator, key) catch {};
+            }
+        }
+        for (keys_to_remove.items) |k| {
+            _ = self.dup_buffer.remove(k);
+        }
+
+        if (self.dup_buffer.count() < BUFFER_CAPACITY) return;
+
+        var tmp = std.ArrayListUnmanaged(struct { u64, Key }){};
```
InKryption: Same comment about using the `.empty` decl.

Also, is there maybe a better name than `tmp`?

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,794 @@
...
+        while (it2.next()) |key_ptr| {
+            const key = key_ptr.*;
+            const epoch = self.params.epoch_schedule.getEpoch(key.slot);
+            const ctx = self.params.epoch_ctx_mgr.get(epoch);
+            var stake: u64 = 0;
+            if (ctx) |c| {
+                if (c.staked_nodes.get(key.from)) |s| stake = s;
+                self.params.epoch_ctx_mgr.release(c);
+            }
+            tmp.append(self.allocator, .{ stake, key }) catch {};
```
InKryption: Why do we ignore this OOM error?

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,794 @@
...
+            .consumed = .empty,
+            .last_root = 0,
+            .cached_slots_in_epoch = params.epoch_schedule.slots_per_epoch,
+        };
+    }
+
+    pub fn deinit(self: *DuplicateShredHandler) void {
+        var it = self.dup_buffer.valueIterator();
+        while (it.next()) |entry| {
+            for (entry.chunks) |maybe_chunk| if (maybe_chunk) |chunk| self.allocator.free(chunk);
```
InKryption: Same comment as on other similar instances.
```suggestion
        for (self.dup_buffer.values()) |entry| {
            for (entry.chunks) |maybe_chunk| if (maybe_chunk) |chunk| self.allocator.free(chunk);
```

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,794 @@
...
+            _ = self.dup_buffer.remove(k);
+        }
+
+        if (self.dup_buffer.count() < BUFFER_CAPACITY) return;
+
+        var tmp = std.ArrayListUnmanaged(struct { u64, Key }){};
+        defer tmp.deinit(self.allocator);
+        var it2 = self.dup_buffer.keyIterator();
+        while (it2.next()) |key_ptr| {
+            const key = key_ptr.*;
```
InKryption: Same comment about using `keys()` instead of the iterator (and everything else).
```suggestion
        for (self.dup_buffer.keys()) |key| {
```

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,794 @@
...
+        const key = Key{ .slot = dup_shred_data.slot, .from = dup_shred_data.from };
+
+        if (try self.params.ledger_reader.isDuplicateSlot(key.slot)) {
+            self.cleanupEntry(key);
+            return;
+        }
+
+        const gop = try self.dup_buffer.getOrPut(self.allocator, key);
+        if (!gop.found_existing) {
+            gop.value_ptr.* = .{ .chunks = [_]?[]u8{null} ** MAX_NUM_CHUNKS };
```
InKryption:
```suggestion
            gop.value_ptr.* = .{ .chunks = @splat(null) };
```

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,794 @@
...
+        var total_len: usize = 0;
+        for (0..dup_shred_data.num_chunks) |i| {
+            if (entry.chunks[i]) |chunk| {
+                filled += 1;
+                total_len += chunk.len;
+            }
+        }
+        if (filled != dup_shred_data.num_chunks) return;
+
+        var data = try self.allocator.alloc(u8, total_len);
```
InKryption: I suggest always writing variables as `const` by default, and only changing them to `var` when the compiler yells at you, or if you explicitly know you need it to be `var`, so that way you don't miss things like this due to the "unnecessary `var`" error being suppressed by derefs and indexing operations.
```suggestion
        const data = try self.allocator.alloc(u8, total_len);
```

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,794 @@
...
+            return error.InvalidSignature;
+        };
+
+        const same_fec =
+            shred1.commonHeader().erasure_set_index == shred2.commonHeader().erasure_set_index;
+        const mr1 = shred1.merkleRoot() catch null;
+        const mr2 = shred2.merkleRoot() catch null;
+        var conflict_ok = false;
+        if (same_fec) {
+            if ((mr1 == null and mr2 != null) or (mr1 != null and mr2 == null)) {
```
InKryption:
```suggestion
            if ((mr1 == null) != (mr2 == null)) {
```

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,794 @@
...
+                conflict_ok = true;
+            } else if (mr1) |h1| {
+                if (mr2) |h2| {
+                    if (!std.mem.eql(u8, &h1.data, &h2.data)) {
+                        conflict_ok = true;
+                    }
+                }
+            }
```
InKryption, line 417: Since their non-nullness is confirmed to be in the same state at this point:
```suggestion
            } else if (mr1) |h1| {
                const h2 = mr2.?;
                if (!std.mem.eql(u8, &h1.data, &h2.data)) {
                    conflict_ok = true;
                }
            }
```

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,794 @@
...
+            if (shred1.commonHeader().index == shred2.commonHeader().index) {
+                if (std.mem.eql(u8, shred1.payload(), shred2.payload())) {
+                    return error.InvalidDuplicateShreds;
+                }
+                conflict_ok = true;
+            } else {
+                const is_data = switch (shred1) {
+                    .data => true,
+                    else => false,
+                };
```
InKryption, line 431: Any reason you're using `else` here?
```suggestion
                const is_data = switch (shred1) {
                    .data => true,
                    .code => false,
                };
```

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,794 @@
...
+        errdefer thread.join();
+
+        thread.setName("sigCiEntryLstnr") catch {};
+        return .{ .thread = thread };
+    }
+
+    pub fn join(self: DuplicateShredListener) void {
+        self.thread.join();
+    }
+};
```
InKryption: Is there any reason for this to exist? It doesn't appear to be used anywhere, everything just uses `recvLoop`.

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,794 @@
...
+const Atomic = std.atomic.Value;
+const Channel = sig.sync.Channel;
+const GossipTable = sig.gossip.GossipTable;
+const Logger = sig.trace.Logger("duplicate_shred_listener");
+const Pubkey = sig.core.Pubkey;
+const RwMux = sig.sync.RwMux;
+const Slot = sig.core.Slot;
+
+const GossipVersionedData = sig.gossip.data.GossipVersionedData;
+const DuplicateShredData = sig.gossip.data.DuplicateShred;
```
InKryption: Why the alias? Was confused for a minute because I couldn't find `DuplicateShredData` in the codebase, since it is declared as `DuplicateShred`.

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,794 @@
...
+    var chunk_bytes = [_]u8{ 1, 2, 3 };
+    const dup: DuplicateShredData = .{
+        .from = Pubkey.initRandom(std.crypto.random),
+        .wallclock = sig.time.getWallclockMs(),
+        .slot = slot,
+        .shred_index = 0,
+        .shred_type = .Data,
+        .num_chunks = 2,
+        .chunk_index = 2, // invalid (>= num_chunks)
+        .chunk = &chunk_bytes,
```
InKryption, line 487: This isn't really directly related to this change, but this change is affected by this: is there any reason this `chunk` field is a mutable slice? If not, could take the opportunity in this PR to make it `[]const u8`, to allow this and other instances of it to directly inline the literal value.

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,794 @@
...
+        .duplicate_slots_sender = &dup_slots_channel,
+        .leader_schedule = epoch_ctx.slotLeaders(),
+        .shred_version = &shred_version,
+        .epoch_schedule = epoch_ctx.schedule,
+        .epoch_ctx_mgr = &epoch_ctx,
+    });
+    defer handler.deinit();
+
+    const slot: Slot = 11;
+    const from = Pubkey.initRandom(std.crypto.random);
```
InKryption: Do not introduce non-deteriminism into tests like this. If you need random numbers, use `std.testing.random_seed`, e.g. `var prng_state: std.DefaultPrng = .init(std.testing.random_seed); const prng = prng_state.random();` at the top of your test.

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,794 @@
...
+        .duplicate_slots_sender = &dup_slots_channel,
+        .leader_schedule = epoch_ctx.slotLeaders(),
+        .shred_version = &shred_version,
+        .epoch_schedule = epoch_ctx.schedule,
+        .epoch_ctx_mgr = &epoch_ctx,
+    });
+    defer handler.deinit();
+
+    const slot: Slot = 12;
+    const from = Pubkey.initRandom(std.crypto.random);
```
InKryption: Same comment about non-determinism.

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,794 @@
...
+    defer handler.deinit();
+
+    // Mark slot as duplicate in ledger
+    const shred: sig.ledger.shred.Shred =
+        try .fromPayload(allocator, &sig.ledger.shred.test_data_shred);
+    defer shred.deinit();
+    const slot: Slot = shred.commonHeader().slot;
+    try ledger.resultWriter().storeDuplicateSlot(slot, shred.payload(), shred.payload());
+
+    const from = Pubkey.initRandom(std.crypto.random);
```
InKryption: Same comment about non-determinism.

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,794 @@
...
+        .shred_version = &shred_version,
+        .epoch_schedule = epoch_ctx.schedule,
+        .epoch_ctx_mgr = &epoch_ctx,
+    });
+    defer handler.deinit();
+
+    const slot: Slot = 10;
+    var chunk_bytes = [_]u8{ 1, 2, 3 };
+    const dup: DuplicateShredData = .{
+        .from = Pubkey.initRandom(std.crypto.random),
```
InKryption: Same comment about non-determinism.

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,794 @@
...
+    defer dup_slots_channel.deinit();
+
+    var base = try sig.ledger.shred.Shred.fromPayload(allocator, &sig.ledger.shred.test_data_shred);
+    defer base.deinit();
+
+    const slot: Slot = base.commonHeader().slot;
+    const version: u16 = base.commonHeader().version;
+
+    const KeyPair = std.crypto.sign.Ed25519.KeyPair;
+    const keypair = try KeyPair.generateDeterministic(.{1} ** 32);
```
InKryption:
```suggestion
    const keypair = try KeyPair.generateDeterministic(@splat(1));
```

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,794 @@
...
+    const leader = Pubkey.fromPublicKey(&keypair.public_key);
+
+    var one_slot_leader_buf = try allocator.alloc(Pubkey, 1);
```
InKryption:
```suggestion
    const one_slot_leader_buf = try allocator.alloc(Pubkey, 1);
```

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,794 @@
...
+        var setter = try handler.params.result_writer.setRootsIncremental();
+        defer setter.deinit();
+        try setter.addRoot(1);
+        try setter.commit();
+
+        handler.cacheRootInfo();
+        const expected_epoch = handler.params.epoch_schedule.getEpoch(1);
+        const expected_slots_in_epoch =
+            handler.params.epoch_schedule.getSlotsInEpoch(expected_epoch);
+        try std.testing.expectEqual(@as(Slot, 1), handler.last_root);
```
InKryption: I don't think this `@as` is necessary, is it?

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,794 @@
...
+    defer ledger.deinit();
+
+    var dup_slots_channel = try Channel(Slot).init(allocator);
+    defer dup_slots_channel.deinit();
+
+    var shred_version = std.atomic.Value(u16).init(0);
+
+    var handler = DuplicateShredHandler.init(allocator, .noop, .{
+        .exit = undefined,
+        .gossip_table_rw = undefined,
```
InKryption: Would it be possible to split out the relevant subset of functionality from `DuplicateShredHandler` into its own thing that can be easily tested without needing to initialize these fields?

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,794 @@
...
+
+    const from = Pubkey.ZEROES;
+    for (0..total_entries) |i| {
+        const key = Key{
+            .slot = handler.last_root + handler.cached_slots_in_epoch + 100 + i,
+            .from = from,
+        };
+        const gop = try handler.dup_buffer.getOrPut(allocator, key);
+        if (!gop.found_existing) {
+            gop.value_ptr.* = .{ .chunks = [_]?[]u8{null} ** MAX_NUM_CHUNKS };
```
InKryption:
```suggestion
            gop.value_ptr.* = .{ .chunks = @splat(null) };
```

src/shred_network/collector/shred_receiver.zig
```diff
@@ -186,13 +201,338 @@ pub const ShredReceiver = struct {
...
+        slot: Slot,
+        shred_payload: []const u8,
+        duplicate_payload: []const u8,
+    ) !void {
+        // Store in ledger
+        self.params.result_writer.storeDuplicateSlot(
+            slot,
+            shred_payload,
+            duplicate_payload,
+        ) catch |err| {
```
InKryption: Would be nice to have a general comment somewhere about these caught-and-logged-but-not-returned ledger errors, describing why they don't need to be caught.

src/shred_network/collector/shred_receiver.zig
```diff
@@ -32,6 +33,9 @@ const VariantCounter = sig.prometheus.VariantCounter;
 const Logger = sig.trace.Logger("shred_receiver");
 const VerifiedMerkleRoots = sig.utils.lru.LruCache(.non_locking, sig.core.Hash, void);
 
+const DUPLICATE_SHRED_HEADER_SIZE: u64 = 63;
+const DUPLICATE_SHRED_MAX_PAYLOAD_SIZE: u16 = 512;
+
 /// Analogous to [ShredFetchStage](https://github.com/anza-xyz/agave/blob/aa2f078836434965e1a5a03af7f95c6640fe6e1e/core/src/shred_fetch_stage.rs#L34)
 pub const ShredReceiver = struct {
     params: Params,
```
InKryption, line 39: Does `Params` need to be a field of `ShredReceiver`? Seems like it would be pretty easy to make this a parameter that's passed to `run`, and then in there to whatever else needs access to these things.

src/shred_network/shred_receiver.zig
```diff
@@ -186,13 +201,338 @@ pub const ShredReceiver = struct {
...
+                    const shred_slot = shred.commonHeader().slot;
+                    // Unlike the other cases we have to wait until here to decide to handle the duplicate and store
+                    // in ledger. This is because the duplicate could have been part of the same insert batch,
+                    // so we wait until the batch has been written.
+                    if (try self.params.ledger_reader.isDuplicateSlot(shred_slot)) {
+                        continue; // A duplicate is already recorded, skip
+                    }
+
+                    const existing_shred_payload =
+                        (try self.params.ledger_reader.isShredDuplicate(allocator, shred)) orelse
```
InKryption: You don't actually need these parens, the precedence ought to work out to the equivalent of `try foo orelse continue` = `foo catch |e| return e orelse continue`.

src/shred_network/shred_receiver.zig
```diff
@@ -186,13 +201,338 @@ pub const ShredReceiver = struct {
...
+                        conflict.conflict.data,
+                    );
+                },
+            }
+        }
+    }
+
+    fn handleDuplicateSlot(
+        self: *ShredReceiver,
+        sender: anytype,
```
InKryption: Why is this `anytype`? Shouldn't it just be `*Channel(Slot)`?

src/shred_network/shred_receiver.zig
```diff
@@ -186,13 +201,338 @@ pub const ShredReceiver = struct {
...
+        const proof_bytes = try serializeDuplicateProof(allocator, shred_payload, other_payload);
+        defer allocator.free(proof_bytes);
+
+        // Build chunks that will be converted to CRDS.
+        const chunks = try self.buildDuplicateShredChunks(
+            gossip,
+            slot,
+            shred_payload,
+            proof_bytes,
+            @as(usize, DUPLICATE_SHRED_MAX_PAYLOAD_SIZE),
```
InKryption: Why the `@as`?

src/shred_network/collector/shred_receiver.zig
```diff
@@ -186,13 +201,338 @@ pub const ShredReceiver = struct {
...
+        );
+    }
+
+    fn hasDuplicateForSlot(
+        gossip: *sig.gossip.GossipService,
+        my_pubkey: sig.core.Pubkey,
+        slot: Slot,
+    ) bool {
+        var gossip_table, var lock = gossip.gossip_table_rw.readWithLock();
+        defer lock.unlock();
```
InKryption: Why not just accept `gossip_table: *RwMux(sig.gossip.GossipTable)` here instead of passing in the entire gossip service, which otherwise goes unused besides the table?

src/shred_network/collector/shred_receiver.zig
```diff
@@ -186,13 +201,338 @@ pub const ShredReceiver = struct {
...
+        gossip: *sig.gossip.GossipService,
+        my_pubkey: sig.core.Pubkey,
+    ) u16 {
+        const MAX_DUPLICATE_SHREDS = sig.gossip.data.MAX_DUPLICATE_SHREDS;
+        var num_dup_shreds: u16 = 0;
+        var oldest_index: u16 = 0;
+        var maybe_oldest_wallclock: ?u64 = null;
+
+        var gossip_table, var lock = gossip.gossip_table_rw.readWithLock();
+        defer lock.unlock();
```
InKryption: Same comment here about just accepting the table.

src/shred_network/collector/shred_receiver.zig
```diff
@@ -186,13 +201,338 @@ pub const ShredReceiver = struct {
...
+        var proof_data: std.ArrayListUnmanaged(u8) = .empty;
+        errdefer proof_data.deinit(allocator);
+        try bincode.write(proof_data.writer(allocator), proof, bincode.Params.standard);
+        const bytes = try proof_data.toOwnedSlice(allocator);
+        return bytes;
+    }
+
+    fn buildDuplicateShredChunks(
+        self: *ShredReceiver,
+        gossip: *sig.gossip.GossipService,
```
InKryption: The only thing this appears to be used for is its `allocator` field; replace it with `allocator: std.mem.Allocator`.

src/shred_network/shred_receiver.zig
```diff
@@ -186,13 +201,338 @@ pub const ShredReceiver = struct {
...
+                .from = sig.core.Pubkey.fromPublicKey(&self.params.keypair.public_key),
+                .wallclock = wallclock,
+                .slot = slot,
+                .shred_index = shred_index,
+                .shred_type = shred_type,
+                .num_chunks = num_chunks,
+                .chunk_index = chunk_index,
+                .chunk = try gossip.allocator.dupe(u8, chunk_data),
+            };
+            chunks.appendAssumeCapacity(duplicate_shred);
```
InKryption: Aesthetic suggestion
```suggestion
            chunks.appendAssumeCapacity(.{
                .from = .fromPublicKey(&self.params.keypair.public_key),
                .wallclock = wallclock,
                .slot = slot,
                .shred_index = shred_index,
                .shred_type = shred_type,
                .num_chunks = num_chunks,
                .chunk_index = chunk_index,
                .chunk = try gossip.allocator.dupe(u8, chunk_data),
            });
```

src/shred_network/shred_receiver.zig
```diff
@@ -487,3 +831,782 @@ pub const ShredValidationError = error{
...
+    // Create a result with a duplicate shred
+    const shred: Shred = try .fromPayload(allocator, &sig.ledger.shred.test_data_shred);
+
+    var duplicate_shreds = std.ArrayList(PossibleDuplicateShred).init(allocator);
+    try duplicate_shreds.append(.{ .Exists = shred });
+
+    const result = ShredInserter.Result{
+        .completed_data_set_infos = std.ArrayList(
+            ShredInserter.CompletedDataSetInfo,
+        ).init(allocator),
```
InKryption:
```suggestion
        .completed_data_set_infos = .init(allocator),
```

src/shred_network/shred_receiver.zig
```diff
@@ -487,3 +831,782 @@ pub const ShredValidationError = error{
...
+    try shred_receiver.handleDuplicateSlots(allocator, &result);
+}
+
+test "handleDuplicateSlots: no duplicate shreds" {
+    const allocator = std.testing.allocator;
+    var registry = sig.prometheus.Registry(.{}).init(allocator);
+    defer registry.deinit();
+
+    const keypair = try sig.identity.KeyPair.generateDeterministic(.{1} ** 32);
+    var exit: Atomic(bool) = Atomic(bool).init(false);
```
InKryption:
```suggestion
    var exit: Atomic(bool) = .init(false);
```

src/shred_network/shred_receiver.zig
```diff
@@ -487,3 +831,782 @@ pub const ShredValidationError = error{
...
+        null,
+        .noop,
+    );
+    defer {
+        gossip_service.shutdown();
+        gossip_service.deinit();
+        allocator.destroy(gossip_service);
+    }
+
+    const insertDup = struct {
```
InKryption: Should probably be snake case `insert_dup`, keeping with conventions.

InKryption (changes requested):
An assortment of suggestions, questions, concerns, and other stuff. Feel free to reach out to me if you want to expedite working through all of this, though I hope nothing posed here is too difficult to parse through.

src/ledger/Reader.zig
dadepo: [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/ledger/ResultWriter.zig
dadepo: [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/duplicate_shred_listener.zig
dadepo: [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/duplicate_shred_listener.zig
dadepo: [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/duplicate_shred_listener.zig
dadepo: [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/duplicate_shred_listener.zig
dadepo: [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/duplicate_shred_listener.zig
dadepo: [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/duplicate_shred_listener.zig
dadepo: [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/duplicate_shred_listener.zig
dadepo: [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/duplicate_shred_listener.zig
dadepo: [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/duplicate_shred_listener.zig
dadepo: [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/duplicate_shred_listener.zig
dadepo: [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/duplicate_shred_listener.zig
dadepo: [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/duplicate_shred_listener.zig
dadepo: [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/duplicate_shred_listener.zig
dadepo: [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/duplicate_shred_listener.zig
dadepo: It was a remnant of a previous implemented. Forgot to clean it up after refactor. Removed now [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/duplicate_shred_listener.zig
dadepo: [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/duplicate_shred_listener.zig
dadepo: [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/duplicate_shred_listener.zig
dadepo: [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/duplicate_shred_listener.zig
dadepo: [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/duplicate_shred_listener.zig
dadepo: [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/duplicate_shred_listener.zig
dadepo: [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/duplicate_shred_listener.zig
dadepo: [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/shred_receiver.zig
dadepo: [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/shred_receiver.zig
dadepo: Left in after a refactor [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/shred_receiver.zig
dadepo: [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/shred_receiver.zig
dadepo: [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/shred_receiver.zig
dadepo: [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/shred_receiver.zig
dadepo: [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/shred_receiver.zig
dadepo: [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/collector/shred_receiver.zig
dadepo: I am sure there is a possibility it can be moved, but this was already here and was not added in this PR. I looked into updating it as suggested, but the diff would  be a distraction to the core changes being added in the PR

src/shred_network/duplicate_shred_listener.zig
dadepo: [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/duplicate_shred_listener.zig
dadepo: [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/duplicate_shred_listener.zig
dadepo: [eb9ca4a](https://github.com/Syndica/sig/pull/1027/commits/eb9ca4a660a2e0b3a3bbd007ea705a26d86d3b64)

src/shred_network/duplicate_shred_listener.zig
dadepo: > This all seems super similar to the vote listener, do you think it would make sense to unroll it into a single-threaded construct

Indeed it is similar. And the preference to go with single-threaded construct dictated where the sending side was placed ie inlined in an existing thread (instead of being spurn in a thread as it is done in Agave) 

On the receiving side though, I could not see a clear/cut place, where this can be inlined. The vote listener came to mind, but that does something semantically different. It process votes from gossip, while this process proof of equivocation. If placed in vote listener then that would have to be renamed.

InKryption: Yeah that's fair enough, if you don't see a clear way forward to pursue this, we can leave it as a future possible improvement, probably intertwined with the change to make gossip itself notify the vote listener and duplicate shred listener, rather than having each of them constantly iterating through the gossip table.

src/shred_network/collector/shred_receiver.zig
dadepo: [652525e](https://github.com/Syndica/sig/pull/1027/commits/652525e8d2404b563d04eabc4d267dce04cdba98)

src/shred_network/duplicate_shred_listener.zig
dadepo: [00e7fd9](https://github.com/Syndica/sig/pull/1027/commits/00e7fd98080e976c2812092f902f638efed2cc9a) alongside some minor changes to be in parity with Agave

src/shred_network/duplicate_shred_listener.zig
dadepo: I have a slight preference for the way it is namely because the struct exposes it's functionality via just one main function: `handle` - and that is what all the tests test.

src/shred_network/collector/shred_receiver.zig
dadepo: [5771599](https://github.com/Syndica/sig/pull/1027/commits/57715997cefe5656672d58ce99589a1757adc979)

src/shred_network/collector/shred_receiver.zig
dadepo: [5771599](https://github.com/Syndica/sig/pull/1027/commits/57715997cefe5656672d58ce99589a1757adc979)

src/shred_network/collector/shred_receiver.zig
dadepo: [5771599](https://github.com/Syndica/sig/pull/1027/commits/57715997cefe5656672d58ce99589a1757adc979)

src/shred_network/duplicate_shred_listener.zig
InKryption: Well, clearly this test doesn't need these two pieces of information, and they aren't exactly a fundamental piece of "shred handler". My main gripe here is just the fact that `DuplicateShredHandler.init` takes `exit` and `gossip_table` as parameters, despite clearly not actually needing them.

dadepo: [fcc4de3](https://github.com/Syndica/sig/pull/1027/commits/fcc4de35749d4da000446d2292d51aa925a6d828)

InKryption (dismissed):

src/shred_network/collector/shred_receiver.zig
```diff
@@ -186,13 +201,344 @@ pub const ShredReceiver = struct {
...
     }
 
     const MAX_SHREDS_PER_ITER = 1024;
 
+    /// Handles detected duplicate slots by:
+    /// - Send duplicate slot notifications to be handled in consensus part of replay
+    /// - Store the duplicate proof in the ledger
+    /// - Broadcast the duplicate proof via gossip
+    /// Analogous to [check_duplicate](https://github.com/anza-xyz/agave/blob/aa2f078836434965e1a5a03af7f95c6640fe6e1e/core/src/window_service.rs#L155)
+    fn handleDuplicateSlots(
```
dnut: this code isn't about receiving shreds, so i find it unintuitive/unexpected to find this code within the shred receiver. can you move this implementation to another file? maybe create a struct called DuplicateShredHandler that contains all the necessary dependencies, and ShredReceiver can accept it as a param just like ShredInserter.

taking this approach should significantly reduce boilerplate in the tests.

dadepo: [991a0c6](https://github.com/Syndica/sig/pull/1027/commits/991a0c69c69cf029251cfd2bb61a658825f6db5b)

dnut: now you can simplify the tests. they shred receiver is not relevant to the tests.

also i'm confused, there are two structs with the same name now?

src/shred_network/collector/duplicate_shred_handler.zig
```diff
@@ -0,0 +1,354 @@
...
+    };
+    var proof_data: std.ArrayListUnmanaged(u8) = .empty;
+    errdefer proof_data.deinit(allocator);
+    try bincode.write(proof_data.writer(allocator), proof, bincode.Params.standard);
+    const bytes = try proof_data.toOwnedSlice(allocator);
+    return bytes;
+}
+
+fn enqueueDuplicateShredCrdsValues(
+    gossip: *sig.gossip.GossipService,
```
dnut: depending on the entire gossip service struct? do we do that anywhere else in the codebase?

src/shred_network/collector/shred_receiver.zig
dadepo: > also i'm confused, there are two structs with the same name now?

Ha yes. There is a duplicate shred handling for what is received via gossip - logic is different as the data is in a different form. 

And there is the one that process the ones detected during shred insertion. 

Not sure the better way to handle this:
- Move both into the same struct even though it would contain two seperate data handling
- Rename to show the difference: `GossipDuplicateShredHandler` and `ShredDuplicateHandler`

Update: Renamed the one in gossip to GossipDuplicateShredHandler. Left the other struct as is. [6d950b9](https://github.com/Syndica/sig/pull/1027/commits/6d950b9259c900be668cf7821492ce494631c59d)

src/shred_network/collector/duplicate_shred_handler.zig
```diff
@@ -0,0 +1,354 @@
...
+const Logger = sig.trace.Logger("duplicate_shred_handler");
+
+pub const DUPLICATE_SHRED_HEADER_SIZE: u64 = 63;
+pub const DUPLICATE_SHRED_MAX_PAYLOAD_SIZE: u16 = 512;
+
+pub const DuplicateShredHandler = struct {
+    ledger_reader: sig.ledger.Reader,
+    result_writer: sig.ledger.ResultWriter,
+    duplicate_slots_sender: ?*Channel(Slot),
+    gossip_service: ?*sig.gossip.GossipService,
```
dnut: Please do not add dependencies on the entire gossip service. There must be a better way

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,955 @@
...
+            defer lock.unlock();
+            break :blk gossip_table.getClonedEntriesWithCursor(
+                allocator,
+                &buf,
+                &cursor,
+            ) catch |e| {
+                logger.err().logf("duplicate_shred_listener: get entries failed: {}", .{e});
+                break :blk &.{};
+            };
+        };
```
dnut, line 70: Locking the gossip table to clone every single value out of it is very unnecessary. Let's filter the duplicate shreds when they are inserted into the gossip table, and send them over a channel.

src/shred_network/collector/duplicate_shred_handler.zig
```diff
@@ -0,0 +1,354 @@
...
+        self: *DuplicateShredHandler,
+        gossip: *sig.gossip.GossipService,
+        slot: Slot,
+        shred_payload: []const u8,
+        other_payload: []const u8,
+    ) !void {
+        const my_pubkey = sig.core.Pubkey.fromPublicKey(&self.keypair.public_key);
+
+        // Early return if we already have a duplicate for this slot.
+        if (hasDuplicateForSlot(&gossip.gossip_table_rw, my_pubkey, slot)) {
```
dnut: We shouldn't need to look in the gossip table to determine this.

src/shred_network/collector/duplicate_shred_handler.zig
```diff
@@ -0,0 +1,354 @@
...
+            return;
+        }
+
+        // Compute ring offset where new entries should be placed/overwritten.
+        const ring_offset = computeRingOffset(&gossip.gossip_table_rw, my_pubkey);
+
```
dnut: This function iterates over all the validator's values in the gossip table just to pull out duplicate shreds, and so does the last one. Why the redundancy?

src/shred_network/collector/duplicate_shred_handler.zig
```diff
@@ -0,0 +1,354 @@
...
+    var gossip_table, var lock = gossip_table_rw.readWithLock();
+    defer lock.unlock();
+
+    if (gossip_table.pubkey_to_values.get(my_pubkey)) |records| {
+        for (records.keys()) |record_ix| {
+            const versioned_data = gossip_table.store.getByIndex(record_ix);
+            switch (versioned_data.data) {
+                .DuplicateShred => |dup| {
+                    _, const dup_shred = dup;
+                    if (dup_shred.slot == slot) return true;
```
dnut: DuplicateShredHandler should keep track of the slots it has reported so we don't need to look in the gossip table.

src/shred_network/collector/duplicate_shred_handler.zig
```diff
@@ -0,0 +1,354 @@
...
+                        maybe_oldest_wallclock = wc;
+                        oldest_index = index;
+                    }
+                },
+                else => {},
+            }
+        }
+    }
+
+    return if (num_dup_shreds < MAX_DUPLICATE_SHREDS) num_dup_shreds else oldest_index;
```
dnut: what is the meaning/purpose of this ring offset and can we calculate it without reading from the gossip table?

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,955 @@
...
+    handler: HandlerParams,
+};
+
+const Key = struct { slot: Slot, from: Pubkey };
+
+const BufferEntry = struct {
+    chunks: [MAX_NUM_CHUNKS]?[]u8,
+};
+
+pub fn recvLoop(
```
dnut, line 47: I think we should remove this function, call `handle` from inside consensus, and `handle` should return the duplicate shreds instead of sending them over a channel.

dnut (commented):
This implementation increases contention on the gossip table, and generally is very coupled to the inner workings of gossip. I think we should do some architectural rework to avoid this now, rather than introducing more tech debt.

I already started to implement sending gossip data out on receipt, instead of having every thread need to iterate over all the values in the gossip table. I'll open a PR soon.

src/shred_network/collector/shred_receiver.zig
dadepo: > now you can simplify the tests. they shred receiver is not relevant to the tests.

5a214b4

src/shred_network/duplicate_shred_listener.zig
dadepo: That can work. Similar to the vote collector bit. Only slight concern is if these operations won't add more latency to the replay/consensus loop. But I guess that is something, if it happens can then justify splitting it off to a separate channel.

src/shred_network/collector/duplicate_shred_handler.zig
dadepo: [8baf17a](https://github.com/Syndica/sig/pull/1027/commits/8baf17a029df9d5b64f8a5fe297192bed42f88ea)

src/shred_network/duplicate_shred_listener.zig
dadepo: > Only slight concern is if these operations won't add more latency to the replay/consensus loop

also the duplicate handling from gossip does more than return the slot to consensus. It also stores the proof in the ledger. An argument can be made that the pace of doing that should not be tied to the cadence of replay/consensus loop.

src/shred_network/collector/duplicate_shred_handler.zig
dadepo: [f8fbb78](https://github.com/Syndica/sig/pull/1027/commits/f8fbb781c89a368dac4e66495d06905e6a8abe06)

src/shred_network/collector/duplicate_shred_handler.zig
dadepo: [f8fbb78](https://github.com/Syndica/sig/pull/1027/commits/f8fbb781c89a368dac4e66495d06905e6a8abe06)

src/shred_network/collector/duplicate_shred_handler.zig
dadepo: [f13123a](https://github.com/Syndica/sig/pull/1027/commits/f13123a96125e49ce71ba28425e8eb6bda974185)

src/shred_network/collector/duplicate_shred_handler.zig
dadepo: [f13123a](https://github.com/Syndica/sig/pull/1027/commits/f13123a96125e49ce71ba28425e8eb6bda974185)

dadepo: Had another thought I'll like to confirm regarding the interplay on how this is inserted. Since the `push_queue` only buffers the gossip data but actually writing to the gossip table is handled by `drainPushQueueToGossipTable`

src/shred_network/collector/duplicate_shred_handler.zig
dadepo: https://github.com/Syndica/sig/pull/1027#discussion_r2585532556

src/shred_network/collector/duplicate_shred_handler.zig
dadepo: ordering in entry in the push_queue is preserved when draining push queue to gossip table in `drainPushQueueToGossipTable`

PR #1029, opened 2025-10-23
test(replay,consensus): cluster sync tests + refactor state management

InKryption:
This adds more component level tests for consensus, specifically oriented for testing the cluster sync logic.

refactoring:
* Rename VoteListener to VoteCollector, make single-threaded
* In cmd.zig, group all replay and consensus state into a single struct `ReplayConsensusState`.
* Simplify lifetimes of fields in replay `Dependencies`, making it no longer be an owning data structure, but a simple parameter struct, and make functions that once returned it instead return ReplayState.
* Rectify ownership contract for `SlotTracker.init`.
* Turn channels directed from vote collector to replay into array lists.
* Pull out the `Senders` and `Receivers` structs as fields from the `TowerConsensus` struct, make them parameters for relevant functions instead, also deleting `External`.
* Delete the `replay.Service` struct, replace it with a simple function and struct parameter pair `advanceReplay` & `AdvanceReplayParams`, comprised of the directly relevant inputs required for a single tick of replay.

### DIFF
--- a/src/replay/consensus/core.zig
+++ b/src/replay/consensus/core.zig
@@ -5,7 +5,6 @@ const tracy = @import("tracy");
 const network = @import("zig-network");
 
 const cluster_sync = replay.consensus.cluster_sync;
-
 const Allocator = std.mem.Allocator;
 const AtomicBool = std.atomic.Value(bool);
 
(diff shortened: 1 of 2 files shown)


InKryption:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Syndica/sig/1029?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1029** <a href="https://app.graphite.dev/github/pr/Syndica/sig/1029?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Syndica/sig/1029?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1005** <a href="https://app.graphite.dev/github/pr/Syndica/sig/1005?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#999** <a href="https://app.graphite.dev/github/pr/Syndica/sig/999?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#992** <a href="https://app.graphite.dev/github/pr/Syndica/sig/992?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/replay/service.zig
```diff
@@ -28,100 +28,146 @@ const EpochTracker = replay.trackers.EpochTracker;
...
+pub const VoteCollectionParams = struct {
+    collector: *sig.consensus.VoteCollector,
+    /// Scanned by the vote collector if provided.
+    gossip_table: ?*sig.sync.RwMux(sig.gossip.GossipTable),
+    /// The vote collector sends verified votes through this channel.
+    verified_vote_channel_sender: *sig.sync.Channel(sig.consensus.vote_listener.VerifiedVote),
 };
 
-/// Create one instance and always pass by pointer.
+pub fn runConsensus(
```
dnut: What is the point of this function? TowerConsensus.process is supposed to be the entrypoint into consensus. I would just absorb this code into that function.

src/replay/service.zig
```diff
@@ -28,100 +28,146 @@ const EpochTracker = replay.trackers.EpochTracker;
...
+
 const updateSysvarsForNewSlot = replay.update_sysvar.updateSysvarsForNewSlot;
 
 pub const Logger = sig.trace.Logger("replay");
 
-pub const Service = struct {
-    replay: ReplayState,
-    consensus: ?TowerConsensus,
-    num_threads: u32,
+pub const AdvanceReplayParams = struct {
```
dnut: Personally I think it's clearer to name this Service or ReplayService and have an advance function on it. That makes it clear that it's the high-level runner for all of replay. Now this just looks like some param struct for some arbitrary function. It's not as clear that this is the sole all-encompassing replay struct.

src/replay/service.zig
```diff
@@ -28,100 +28,146 @@ const EpochTracker = replay.trackers.EpochTracker;
...
+    want_multi_threaded_replay: bool,
+    replay_state: *ReplayState,
+    consensus: ?struct {
+        tower: *TowerConsensus,
```
dnut: Why did you make ReplayState and TowerConsensus pointers? It's more state to juggle in whatever context needs to create this struct. I liked how this was all owned by one entity.

src/replay/service.zig
```diff
@@ -28,100 +28,146 @@ const EpochTracker = replay.trackers.EpochTracker;
...
+    // run consensus
+    if (params.consensus) |consensus| {
+        var gossip_verified_vote_hashes: std.ArrayListUnmanaged(GossipVerifiedVoteHash) = .empty;
+        defer gossip_verified_vote_hashes.deinit(allocator);
 
-        // replay slots
-        const slot_results = try replay.execution.replayActiveSlots(&self.replay, self.num_threads);
-        defer allocator.free(slot_results);
+        var duplicate_confirmed_slots: std.ArrayListUnmanaged(ThresholdConfirmedSlot) = .empty;
+        defer duplicate_confirmed_slots.deinit(allocator);
```
dnut, line 101: What's the point of creating these here? These seem like implementation details that belong deeper inside runConsensus.

src/replay/service.zig
```diff
@@ -201,38 +217,49 @@ pub const ReplayState = struct {
...
 
         var slot_tracker: SlotTracker = try .init(deps.allocator, deps.root.slot, .{
-            .constants = deps.root.constants.take(),
-            .state = deps.root.state.take(),
+            .constants = deps.root.constants,
+            .state = deps.root.state,
         });
         errdefer slot_tracker.deinit(deps.allocator);
+        errdefer {
+            // do not free the root slot data parameter, we don't own it unless the function returns successfully
```
dnut, line 202: I don't like this ambiguous ownership model because it's difficult to handle correctly in the calling scope. Actually in cmd.zig, it is not handled correctly because the items will be double-freed if an error occurs after this function returns without error.

Managing this ownership correctly means the calling code needs to carefully wrap its calls in blocks and put errdefers in the right places to enforce the correct ownership rules of each function it calls, which is easy to screw up. If there are multiple functions using the same pattern that are called in the same scope, it may not even be possible to use block-based approach. The alternative is to track some state needs to be maintained to decide whether an item is owned or not, like an `xxx_is_owned` variable that is mutated when ownership is transferred. This approach is always viable but it's a pain to manage the separate variable, and it's easy to forget that it's necessary. Either of these aproaches can be difficult/ugly to do correctly.

The `Owned` wrapper that you removed is one potential state-based approach I was referring to. I like it because it bundles the state with the owned item, forcing you to use it correctly. I think it is cleaner and less mistake-prone than other state-based approaches I've experimented with. It makes it clear to both scopes whether they own it. It's something you're forced to think about when you interact with it, but it brings clarity instead of letting you make subtle mistakes.

src/replay/service.zig
InKryption: Well, it is just a param struct, although it is convenient to have it named so it can be re-used by anything else calling `advanceReplay`. It's not an all-encompassing struct, it just acts as a means to pass in all the relevant state, which is easier for setting up in tests, without compromising the ability to set up the state in a sane way in the actual validator.

src/replay/service.zig
InKryption: Having it all owned by one entity was significantly more complicated due to the layered-ness of the optionals, both in terms of testing the relevant components, and the fact that it obligated the use of much more layered and complicated ownership rules, which ties in to the removal of the `Owned` wrappers; without all these layers of ownership, how data is supposed to be handled becomes much simpler to reason about.

src/replay/service.zig
InKryption: They're passed in as input-output parameters, which makes it trivial to push in values during a test.

src/replay/service.zig
dnut: But you didn't change the ownership rules. All you did was remove the wrapper around the owned fields. You still have the ownership passing from caller to callee in the same way. The whole idea of the wrapper was to ensure the ownership is being passed down correctly. By removing that, you've lost those guardrails. Likewise you also introduced bugs, which is understandable because it's difficult to manage the ownership correctly without some guardrails.

I don't see what you mean about it being complicated due to layerdness of optionals or testing components. Your change just means that you now need to have two values that you pass by pointer instead of one. You actually added another struct in cmd.zig called `ReplayAndConsensusServiceState` to wrap all the state, which essentially serves the same role as the prior `Service` struct, and then you decompose it into multiple pointers to wrap again in `AdvanceReplayParams`. I don't understand why you don't just pass down `*ReplayAndConsensusServiceState` (aka `*Service`) instead of wrapping pointers to data within it in another struct. These are unnecessary layers that make the code more convoluted.

src/replay/service.zig
dnut: I don't see anywhere that `runConsensus` is called in a test. This use case you're describing is not actually being used.

Also, why would you push in values during a test? Something feels off about that. If you're trying to test this function, shouldn't you be testing it as it will be used in production -- with empty lists being passed in? It's a strange use case.

InKryption: It isn't used in this PR, it's made use of in the PR on top.

src/replay/service.zig
InKryption: > Actually in cmd.zig, it is not handled correctly because the items will be double-freed if an error occurs after this function returns without error.

I don't see the bug; the lifetimes are all managed correctly, as described by the relevant comment: `/// ownership transferred to replay; won't be freed if `ReplayState.init` returns an error.`; if the `try .init` fails, all the `errdefer` run, and this prevents this scope from double-freeing this resource. If it returns successfully, those external errdefers don't run, and this takes ownership of the values, so only when `errdefer replay_state.deinit()` runs is it freed.

src/replay/service.zig
InKryption: > Your change just means that you now need to have two values that you pass by pointer instead of one.

The change means that I can choose to initialize them separately, without an additional `init` layer that initializes both of them in a way that isn't necessarily conducive to what I'm trying to accomplish. They can accomplish the same exact tasks, but this has fewer embedded abstractions around how you initialize the state, which were making it very difficult to actually understand where any given part of the state was intertwined or independent.

src/replay/service.zig
InKryption: Went ahead and did exactly that f552649a

src/replay/service.zig
dnut: I see that it's more flexible but it adds more layers that make the code more confusing to read. I don't see how the flexibility is actually being taken advantage of in this scenario. If it's just for testing, I'd rather not clutter the production code with extra layers just to make testing a little more convenient.

I'm not sure I see how there are any fewer embedded abstractions. Are you talking about removing `Owned`? The abstraction and semantics that it represented are still present. They're just explained with a comment instead of being enforced by a type, which I think actually harms clarity by giving the illusion that the semantics are simpler than they really are.

src/replay/service.zig
dnut: You're right, I don't see a bug in cmd.zig now. I could have sworn I looked in cmd.zig and saw that replay_state was *not* initialized in its own block. I'm looking again now and I see that it is.

Regardless I still prefer the `Owned` approach because it automatically manages the ownership correctly, rather than hoping the author of calling code reads the comments on the Dependencies struct and uses that information to correctly manage the values whose ownership transfers conditionally depending on whether the function returns an error.

I actually can think of other places in the code where `Owned` would make it easier to correctly transfer ownership between different contexts, and I'd like to experiment with that. The whole idea here is to eliminate footguns.

src/consensus/vote_listener.zig
```diff
@@ -72,12 +72,12 @@ pub const SlotDataProvider = struct {
 
 pub const Senders = struct {
     verified_vote: *sig.sync.Channel(VerifiedVote),
-    gossip_verified_vote_hash: *sig.sync.Channel(GossipVerifiedVoteHash),
+    gossip_verified_vote_hashes: *std.ArrayListUnmanaged(GossipVerifiedVoteHash),
```
dadepo, line 75: Given these are no longer channels, it could be a bit misleading having them still grouped under `Senders` with rest of the channels.

src/consensus/vote_listener.zig
```diff
@@ -881,24 +892,13 @@ fn trackNewVotesAndNotifyConfirmations(
...
 
             if (is_gossip_vote and is_new and stake > 0) {
-                senders.gossip_verified_vote_hash.send(.{ vote_pubkey, slot, hash }) catch |err| {
-                    logger.err().logf(
-                        "{s}: gossip verified vote hash couldn't send: " ++
-                            ".{{ .vote_pubkey = {f}, .slot = {}, .hash = {f} }}",
-                        .{ @errorName(err), vote_pubkey, slot, hash },
-                    );
-                };
+                const gossip_verified_vote_hash = senders.gossip_verified_vote_hashes;
```
dadepo: Any reason for the variable when it seems appending directly works?

```suggestion
                try senders.gossip_verified_vote_hashes.append(
                    allocator,
                    .{ vote_pubkey, slot, hash },
                );
```

src/consensus/vote_listener.zig
```diff
@@ -881,24 +892,13 @@ fn trackNewVotesAndNotifyConfirmations(
...
-                if (senders.duplicate_confirmed_slots) |sender| {
-                    sender.send(.{ .slot = slot, .hash = hash }) catch |err| {
-                        logger.err().logf(
-                            "{s}: duplicate confirmed slot couldn't send: {} (hash: {f})",
-                            .{ @errorName(err), slot, hash },
-                        );
-                    };
-                }
+            if (reached_threshold_results.isSet(0)) blk: {
+                const dupe_confirmed_slots = senders.duplicate_confirmed_slots orelse break :blk;
```
dadepo: Same question as https://github.com/Syndica/sig/pull/1029/files#r2500348666

dadepo (approved):
Some couple of comments but overall looks good.

src/consensus/vote_listener.zig
InKryption: In this case it's because of the need to unwrap the optional; not sure of a better way to express this.

src/consensus/vote_listener.zig
InKryption: Was mostly to avoid hitting the column length limit while not splaying the parameters onto multiple lines, but I can change it to that.

src/consensus/vote_listener.zig
InKryption: I tried to refactor these out, but it makes it a fair bit more difficult to just set up the inputs/outputs. Actually in the future I would like to see if it would be possible/beneficial to turn all of these channels into buffers, to make the vote collector fully "pure", where if these still need to be sent over channels, they can be sent over channels after the outputs are collected into these buffers.
With that in mind, would you be satisfied with a compromise where these are grouped together for now, despite not all being of uniform types?

src/consensus/vote_listener.zig
InKryption: 9a1e4731c32e820193e26501690c8099b2301dca

src/consensus/vote_listener.zig
InKryption: 9a1e4731c32e820193e26501690c8099b2301dca

src/consensus/vote_listener.zig
dadepo: Might be instructive to leave a comment that the current grouping is a bit odd? if doing it now would take more effort to be done in this PR? I'll resolve this for now

dadepo (approved):

src/replay/consensus/core.zig
```diff
@@ -413,37 +360,60 @@ pub const TowerConsensus = struct {
...
+                .subscriptions = .{},
+            },
+            .receivers = .{ .replay_votes = params.receivers.replay_votes },
+            .ledger = params.ledger,
+            .gossip_table = params.gossip_table,
+        });
+
+        // Process replay results
+        for (params.results) |r| {
+            try self.processResult(
```
dadepo, line 382: Looking at this again, it seems there is an ordering issue as follows:

`self.vote_collector.collectAndProcessVotes` processes new votes from the network and populates `gossip_verified_vote_hashes` and `duplicate_confirmed_slots`

But, `gossip_verified_vote_hashes` and `duplicate_confirmed_slots` are not passed into `self.processResult` so in the code path where slot is marked as dead, or slot got frozen, the process is not based on the newly accumulated votes from the network.

I think the newly fetched data from the network should also be passed into `self.processResult` as it depends on that too

dnut: I don't fully understand what you're suggesting we should do. It seems that you're implying that processResult is not properly implemented because it's missing the logic to use these inputs. How exactly should it make use of those inputs? And how is this relevant to the current PR? I don't think the semantics of processResult have changed in this PR.

dadepo: > It seems that you're implying that processResult is not properly implemented because it's missing the logic to use these inputs. 

Nope. The logic is still there. My question is regarding how the inputs is being supplied to it. 

In processResult if it is a dead slots, there is a procedure that reads from `duplicate_confirmed_slots`

https://github.com/Syndica/sig/blob/d81dd201fa6793c3f8d555bdbb3cfde14bf9f83a/src/replay/consensus/process_result.zig#L111


Currently that does not include the newly retrieved `duplicate_confirmed_slots` from the `self.vote_collector.collectAndProcessVotes` call that happened just before it. Instead it only includes the data from the previous run https://github.com/Syndica/sig/blob/d81dd201fa6793c3f8d555bdbb3cfde14bf9f83a/src/replay/consensus/core.zig#L469

Same also applies when it is a frozen slot https://github.com/Syndica/sig/blob/d81dd201fa6793c3f8d555bdbb3cfde14bf9f83a/src/replay/consensus/process_result.zig#L190

And in case of frozen slots, the `unfrozen_gossip_verified_vote_hashes` that is moved to `latest_validator_votes` here https://github.com/Syndica/sig/blob/d81dd201fa6793c3f8d555bdbb3cfde14bf9f83a/src/replay/consensus/process_result.zig#L230-L244 does not include any vote that could have been newly fetched by the `self.vote_collector.collectAndProcessVotes` call. Because it got it from what was saved in the previous run https://github.com/Syndica/sig/blob/d81dd201fa6793c3f8d555bdbb3cfde14bf9f83a/src/replay/consensus/process_result.zig#L328


So the semantic of `processResult` is unchanged and is still correct. My question was more about if they should be fed the newly retrieved inputs from the network  when `self.vote_collector.collectAndProcessVotes` runs

dadepo: Going through the flow in Agave, I think what we have is identical. ie the data being pulled from the network in the vote listener thread is sent down the channel but only updates both `unfrozen_gossip_verified_vote_hashes` and `duplicate_confirmed_slots` later down in the replay channel after process_replay_result - so even if there is new data, it is not available until later.

dadepo (approved):

PR #1030, opened 2025-10-23
feat(rpc): RPC hooks

kprotty:
Simple interface for implementing RPC methods across services.


src/rpc/hooks.zig
```diff
@@ -0,0 +1,128 @@
...
+        const cref = try ContextRef.init(allocator, context);
+        defer cref.ctx_ref.dec(allocator);
+
+        const Context = @TypeOf(context);
+        inline for (comptime std.meta.declarations(Context)) |decl| {
+            const method = comptime for (std.meta.fieldNames(Method)) |method_name| {
+                if (std.mem.eql(u8, method_name, decl.name)) {
+                    break @field(Method, method_name);
+                }
+            } else @compileError("No RPC method named: " ++ decl.name);
```
InKryption:
```suggestion
            if (!@hasField(Method, decl.name)) @compileError(
                "No RPC method named: " ++ decl.name,
            );
            const method = @field(Method, decl.name);
```

src/rpc/hooks.zig
```diff
@@ -0,0 +1,128 @@
...
+        while (it.next()) |i| {
+            self.map.values[i].ctx_ref.dec(allocator);
+        }
+    }
+
+    pub fn set(
+        self: *Hooks,
+        allocator: std.mem.Allocator,
+        context: anytype,
+    ) !void {
```
InKryption: Would be good to add some documentation on this about the expected "shape" of `context`, including commentary about the fact that it is expected to be a value of a type with a namespace, wherein the list of all `pub` declarations must be methods with names all matching a tag of the `Method` enum, with appropriate parameters and return type and all that.

Also would be nice to have an explicit error set here to communicate that all it returns is `OutOfMemory` and `MethodAlreadyImplemented`.

src/rpc/hooks.zig
```diff
@@ -0,0 +1,128 @@
...
+            const method = comptime for (std.meta.fieldNames(Method)) |method_name| {
+                if (std.mem.eql(u8, method_name, decl.name)) {
+                    break @field(Method, method_name);
+                }
+            } else @compileError("No RPC method named: " ++ decl.name);
+
+            const callback = @field(Context, decl.name);
+            if (self.map.contains(method)) {
+                return error.MethodAlreadyImplemented;
+            }
```
InKryption, line 94: I do wonder, do we want this to be an error? There's nothing we can really do to "handle" it, so maybe it should just be a panic, and there should be a comment on the function that the programmer has to make sure not to make overlapping calls to `set` (because if they do, that is explicitly just programmer error).

src/rpc/hooks.zig
```diff
@@ -0,0 +1,128 @@
...
+                        _allocator: std.mem.Allocator,
+                        ctx_ref: *ContextRef,
+                        args: sig.rpc.methods.Request(method),
+                    ) sig.rpc.methods.Result(method) {
+                        const _cref: *CRef = @alignCast(@fieldParentPtr("ctx_ref", ctx_ref));
+                        if (@call(.auto, callback, .{ _cref.value, _allocator, args })) |response| {
+                            return .{ .ok = response };
+                        } else |err| {
+                            return .{ .err = .{
+                                .code = @enumFromInt(@intFromError(err)),
```
InKryption, line 117: This conversion doesn't seem sound? Just the idea of returning random based-on-compilation-unit error codes as an RPC error makes me a bit uneasy, but if we do there should at least be some minimal conversion layer to make sure it doesn't happen to coincide with any of the pre-defined values in `ErrorCode`.

src/rpc/methods.zig
```diff
@@ -19,6 +19,22 @@ const Pubkey = sig.core.Pubkey;
...
+}
+
+pub fn Request(comptime method: MethodAndParams.Tag) type {
+    for (@typeInfo(MethodAndParams).@"union".fields) |field| {
+        if (std.mem.eql(u8, field.name, @tagName(method))) {
+            if (field.type == noreturn) @compileError("TODO: impl " ++ @tagName(method));
+            return field.type;
+        }
+    } else unreachable;
+}
```
InKryption, line 34:
```suggestion
pub fn Request(comptime method: MethodAndParams.Tag) type {
    const FieldType = @FieldType(MethodAndParams, @tagName(method));
    if (FieldType == noreturn) @compileError("TODO: impl " ++ @tagName(method));
    return FieldType;
}
```

src/rpc/methods.zig
```diff
@@ -123,6 +145,34 @@ pub const MethodAndParams = union(enum) {
     };
 };
 
+pub const GetSnapshot = struct {
+    path: []const u8,
+    get: enum { file, size },
```
InKryption: Perhaps bind this to a name and then also use in `pub const Response = union(GetKind) {` to signify the relationship.

src/rpc/methods.zig
```diff
@@ -123,6 +145,34 @@ pub const MethodAndParams = union(enum) {
...
+        size: u64,
+
+        pub fn jsonStringify(
+            _: Response,
+            /// `*std.json.WriteStream(...)`
+            jw: anytype,
+        ) @TypeOf(jw.*).Error!void {
+            @panic("GetSnapshot is not a real JSON-RPC method" ++
+                "It is meant only for RPC server. Do not serialize");
+        }
```
InKryption, line 172: Does any runtime path end up forcing this to get analyzed? If not, it could be a `@compileError` instead of a `@panic`.

src/rpc/server/basic.zig
```diff
@@ -387,97 +387,72 @@ fn handleRpcRequest(
...
+                // Avoid eager-eval of any inline-else method to call()'s rcp.methods.Request().
+                @setEvalBranchQuota(10_000);
+                inline for (@typeInfo(rpc.methods.MethodAndParams).@"union".fields) |field| {
+                    if (comptime std.mem.eql(u8, field.name, @tagName(method))) {
+                        if (comptime field.type == noreturn) {
+                            try sendFinalMethodNotFound(request, logger, method, rpc_request.id);
+                            return;
+                        }
+                    }
+                }
```
InKryption:
```suggestion
                if (@FieldType(rpc.methods.MethodAndParams, @tagName(method)) == noreturn) {
                    try sendFinalMethodNotFound(request, logger, method, rpc_request.id);
                    return;
                }
```

src/rpc/server/basic.zig
```diff
@@ -387,97 +387,72 @@ fn handleRpcRequest(
...
-            const config: rpc.methods.GetAccountInfo.Config = params.config orelse .{};
-            const encoding = config.encoding orelse .base64;
-            if (config.commitment) |commitment| {
-                std.debug.panic("TODO: handle commitment={s}", .{@tagName(commitment)});
-            }
+    switch (@as(rpc.methods.MethodAndParams.Tag, rpc_request.method)) {
+        inline else => |method| {
+            const allocator = json_arena;
+            const result = blk: {
+                // Avoid eager-eval of any inline-else method to call()'s rcp.methods.Request().
```
InKryption: Could make that `Request` function return a `?type`, or a variant of it that `Request` calls internally, in order to avoid the noreturn compile error; would make it easier to update in the future when we've finally implemented all the methods there and remove the optional-ness to find all these sites.

Edit: does calling it actually cause a compile error? I would have expected the fact that the compile-erroring tags are `noreturn` to make the `inline else => |method|` here to never return eval to any of those anyway, since they're unreachable branches.

InKryption (commented):
I like the new design, just have a number of suggestions and concerns I'd like to address.

src/rpc/hooks.zig
kprotty: JSON-RPC 2.0 [spec](https://www.jsonrpc.org/specification#error_object) only holds `-32_000`+ values noted in `sig.rpc.response.ErrorCode`, which is outside the range of numbers returned by `@intFromError()` which is a u16 or some unsigned/positive integer.

The error conditions for Solana RPC calls are marked as "unstable" (see [getHealth](https://solana.com/docs/rpc/http/gethealth)'s "Unhealthy" tab docs), so currently implementation-defined. The goal of `intFromError` casts was to make it easier for registered hook functions to return "some error or Response" instead of fiddling with what exactly to return on RPC error.

InKryption: Would be good to have a comment explaining exactly that on this bit of code.

src/rpc/server/basic.zig
kprotty: Tried out `?type`, but it added a lot of boilerplate. Instead opted to simplify the inline check.

And yea, the `@compileError` is triggered by a call to `Request(method)` in the function signature of `Hooks.call`

InKryption (dismissed):

yewman (dismissed):
Looks good to me 👍

PR #1033, opened 2025-10-27
fix(replay:consensus): Don't `fetchSwapRemove` on copy of map

InKryption:
This would be reflected as just a swap in the actual map, not a removal. I can't imagine this is intended.

### DIFF
--- a/src/replay/consensus/process_result.zig
+++ b/src/replay/consensus/process_result.zig
@@ -226,11 +226,7 @@ fn updateConsensusForFrozenSlot(params: ProcessResultParams, slot: Slot) !void {
     }
 
     // Move unfrozen_gossip_verified_vote_hashes entries to latest_validator_votes
-    if (params
-        .unfrozen_gossip_verified_vote_hashes.votes_per_slot
-        .get(slot)) |slot_hashes_const|
-    {
-        var slot_hashes = slot_hashes_const;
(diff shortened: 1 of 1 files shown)


src/replay/consensus/process_result.zig
```diff
@@ -226,11 +226,7 @@ fn updateConsensusForFrozenSlot(params: ProcessResultParams, slot: Slot) !void {
     }
 
     // Move unfrozen_gossip_verified_vote_hashes entries to latest_validator_votes
-    if (params
-        .unfrozen_gossip_verified_vote_hashes.votes_per_slot
-        .get(slot)) |slot_hashes_const|
-    {
-        var slot_hashes = slot_hashes_const;
+    if (params.unfrozen_gossip_verified_vote_hashes.votes_per_slot.getPtr(slot)) |slot_hashes| {
```
dadepo, line 229: Actually based on my recent going though the code, I think it should be:

```suggestion
    if (params.unfrozen_gossip_verified_vote_hashes.votes_per_slot.fetchRemove(slot)) |slot_hashes| {
```

the entry should actually be removed from the `votes_per_slot` map

dadepo: I'm wrong with the `fetchRemove` suggestion, and the changes if fine

PR #1039, opened 2025-10-28
fix(consensus) Correctly pass gossip votes to `latest_validator_votes`

dadepo:
The votes being moved from `unfrozen_gossip_verified_vote_hashes` to `latest_validator_votes` after freezing a slot should be of kind `.gossip`


src/replay/consensus/process_result.zig
```diff
@@ -236,7 +237,7 @@ fn updateConsensusForFrozenSlot(params: ProcessResultParams, slot: Slot) !void {
                     pubkey,
                     slot,
                     hash,
-                    .replay,
+                    .gossip,
```
dadepo, line 240: Corresponding code in Agave

https://github.com/anza-xyz/agave/blob/fccef82a12366ecfb102356c9167d06f308e32f9/core/src/replay_stage.rs#L3425

the parameter is set to `false` which is `is_replay_vote` hence in sig it should be passed in as `.gossip` vote which is what it is.

PR #1041, opened 2025-10-29
fix(vote_listener, consensus) Fix race when accessing the root ancestors

### LINKED ISSUE #983: Panic when accessing root slot's ancestors.
### Description

This line leads to a panic

https://github.com/Syndica/sig/blob/a3250c5546c59e950517ba7ecba74f584617740c/src/consensus/vote_listener.zig#L465

```
0x2c6c876 in processVotesOnce (sig)
                .ancestors = slot_data_provider.getSlotAncestorsPtr(root_slot).?, // must exist for the root slot
                                                                              ^
/home/ubuntu/sig/src/consensus/vote_listener.zig:412:37: 0x2c6e411 in processVotesLoop (sig)
        switch (try processVotesOnce(
                                    ^
/home/ubuntu/.zvm/0.14.1/lib/std/Thread.zig:507:21: 0x2a3b3c2 in callFn__anon_194242 (sig)
                    @call(.auto, f, args) catch |err| {
                    ^
/home/ubuntu/.zvm/0.14.1/lib/std/Thread.zig:757:30: 0x29703c4 in entryFn (sig)
                return callFn(f, args_ptr.*);
                             ^
???:?:?: 0x730fde29caa3 in ??? (libc.so.6)
Unwind information for `libc.so.6:0x730fde29caa3` was not available, trace may be incomplete
```

### How to Reproduce the Bug

Run offline replay with consensus enabled on test net

`sig replay-offline -c testnet --replay-threads 1 --disable-consensus=false  --use-disk-index --skip-snapshot-validation`

After a while, the panic should pop up.

### Additional Context

_No response_

dadepo:
The panic:

```
/home/ubuntu/sig/src/consensus/vote_listener.zig:487:79: 0x2efbffa in processVotesOnce (sig)
                .ancestors = slot_data_provider.getSlotAncestorsPtr(root_slot).?, // must exist for the root slot
                                                                              ^
/home/ubuntu/sig/src/consensus/vote_listener.zig:443:29: 0x2efd185 in processVotesLoop (sig)
        try processVotesOnce(
                            ^
/home/ubuntu/.zvm/0.14.1/lib/std/Thread.zig:507:21: 0x2cc72db in callFn__anon_195406 (sig)
                    @call(.auto, f, args) catch |err| {
                    ^
/home/ubuntu/.zvm/0.14.1/lib/std/Thread.zig:757:30: 0x2bfcd24 in entryFn (sig)
                return callFn(f, args_ptr.*);
                             ^
```

This is due to the rooted slot being advance between the call  `const root_slot = slot_data_provider.rootSlot();` which retrieves the `root_slot` and `slot_data_provider.getSlotAncestorsPtr(root_slot).?` where the ancestors of the retrieved `root_slot` is fetched.

Fixes #983

### DIFF
--- a/src/consensus/vote_listener.zig
+++ b/src/consensus/vote_listener.zig
@@ -474,23 +474,43 @@ fn processVotesOnce(
     vote_processing_time: *VoteProcessingTiming,
     metrics: VoteListenerMetrics,
 ) !void {
-    const root_slot = slot_data_provider.rootSlot();
-    const root_hash = slot_data_provider.getSlotHash(root_slot);
+    const RootSnapshot = struct {
+        slot: sig.core.Slot,
+        hash: ?sig.core.Hash,
+        ancestors: sig.core.Ancestors,
+    };
+
+    const maybe_root_snapshot: ?RootSnapshot = blk: {
+        const slot_tracker, var lg = slot_data_provider.slot_tracker_rw.readWithLock();
+        defer lg.unlock();
+
+        const slot = slot_tracker.root;
+        if (slot_tracker.get(slot)) |ref| {
+            const hash = ref.state.hash.readCopy();
+            const ancestors = try ref.constants.ancestors.clone(allocator);
+            break :blk RootSnapshot{ .slot = slot, .hash = hash, .ancestors = ancestors };
+        }
+        break :blk null;
+    };
+    defer if (maybe_root_snapshot) |*root_snapshot| root_snapshot.ancestors.deinit(allocator);
 
     if (last_process_root.elapsed().asMillis() > DEFAULT_MS_PER_SLOT) {
-        const unrooted_optimistic_slots = try confirmation_verifier.verifyForUnrootedOptimisticSlots(
-            allocator,
-            ledger,
-            .{
-                .slot = root_slot,
-                .hash = root_hash,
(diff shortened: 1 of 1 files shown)


src/consensus/vote_listener.zig
```diff
@@ -474,23 +474,43 @@ fn processVotesOnce(
...
-        defer allocator.free(unrooted_optimistic_slots);
+        if (maybe_root_snapshot) |root_snapshot| {
+            const unrooted_optimistic_slots =
+                try confirmation_verifier.verifyForUnrootedOptimisticSlots(
+                    allocator,
+                    ledger,
+                    .{
+                        .slot = root_snapshot.slot,
+                        .hash = root_snapshot.hash,
+                        .ancestors = &root_snapshot.ancestors,
```
dadepo, line 506: The other solution I considered is to expose a method on `slot_tracker` something like `slot_tracker.getRootsAncestors()` that returns a copy of `Ancestors`.

Instead of the step of first getting the `root_slot`, and then retrieve the ancestors for that, which left a gap for a race.

Did not go with that as I am not sure if that fits into future design consideration for the slot tracker to have less responsibilities.

dadepo:
Closing in favour of https://github.com/Syndica/sig/pull/1005

PR #1042, opened 2025-10-29
refactor(gossip, table): update data overwrite method with more precise return values TODO

usebeforefree:
The logic for inserting new data into the `GossipTable` is very convoluted and unclear. Though it is directly taken from the Agave implementation.

The diff in this PR is not the prettiest, but it does make perfect sense.

Anyways, in the Agave implementation this is the control flow:

```zig
if (!result.found_existing) {
...
// We are checking if an overwrite is supposed to happen
} else if (versioned_value.overwrites(&result.entry.getVersionedData())) {
        // New value time is more recent
        // or new value hash is bigger
} else {
    if (current_entry.value_hash.order(&metadata.value_hash) != .eq) {
        // Old value time is more recent
        // or old value hash is bigger
        return .IgnoredOldValue;
     } else {
         // The hashes are equal
         return .IgnoredDuplicateValue;
     }
}

```
I propose this implementation:

```zig
if (!result.found_existing) {
...
// We are checking if an overwrite is supposed to happen
} else switch (versioned_value.overwrites(&result.entry.getVersionedData())) {
    // overwrite existing entry
    .new => {
        // New value time is more recent
        // or new value hash is bigger
    },
    .old => {
        // Old value time is more recent
        // or old value hash is bigger
        return .IgnoredOldValue;
    },
    // The hashes are equal
    .eq => return .IgnoredDuplicateValue,
}

```

### DIFF
--- a/src/gossip/data.zig
+++ b/src/gossip/data.zig
@@ -49,7 +49,7 @@ pub const GossipVersionedData = struct {
     }
 
     /// Returns which value should overwrite the other, or if they're equal
-    pub fn overwrites(new_value: *const @This(), old_value: *const @This()) Overwrites {
+    pub fn overwrites(new_value: *const @This(), old_value: *const @This()) OverwritesResult {
         // labels must match
         std.debug.assert(@intFromEnum(new_value.data.label()) == @intFromEnum(old_value.data.label()));
 
@@ -70,7 +70,7 @@ pub const GossipVersionedData = struct {
         }
     }
 
-    pub const Overwrites = enum {
+    pub const OverwritesResult = enum {
         new,
         old,
         eq,


InKryption (approved):
Seems like a solid improvement to me

Rexicon226:
Good improvement, just keep in mind that our CI setup is kinda broken for PRs from forks. I really want to fix this, just haven't thinked up a good solution yet. Perhaps if we move back to Github Actions, it'll be easier, they support requiring approval on workflows unlike circleci.

The solana-conformance CI needs to download some artifacts, and even public repo downloads through the API requires giving it a PAT, which we can't risk leaking to malicious fork PRs :).

For now, someone on the team just needs to re-make this PR. That said, all the work is appreciated @usebeforefree  ❤️ .

PR #1045, opened 2025-10-30
perf(runtime): remove account cache & init SvmGateway per-slot

dnut:
I removed the BatchAccountCache struct and all account pre-loading logic, and I replaced it with direct usages of the account store in the runtime.

Now that there's no batch-specific state, I changed SvmGateway to only load once per slot, instead of once per batch.

I consolidated CopiedAccount and CachedAccount into something called LoadedAccount which represents any loaded accounts. For now, all accounts are copied. This means frequently used accounts are copied extra times. In the future, we should optimize this by eliminating AccountSharedData and using Account throughout the runtime to avoid copies by directly pointing to memory in the database, but this will be a significant change.

I also simplified the return type from the transaction executor, which enabled some cascading simplifications to other code in the committer and the transaction harness. it returned a bunch of unnecessary information that just made life harder for callers. this also revealed a bug in solfuzz_agave, which i also needed to implement in our transaction harness to ensure we have the same buggy behavior for conformance.

I removed TransactionRollbacks and just replaced it with a list of writes. Rollback logic has been inlined into check fee payer.

I ran a simple benchmark by running replay-offline on testnet for 60 seconds:
- master: 121 slots
- this branch: 125 slots

### DIFF
--- a/src/runtime/account_loader.zig
+++ b/src/runtime/account_loader.zig
@@ -78,17 +78,13 @@ pub const LoadedTransactionAccounts = struct {
     }
 };
 
-// An account that was loaded to execute a transaction. It may be either a
-// pointer to the account map, or an owned copy, as indicated by the is_owned
-// field. All writable accounts are expected to be owned, but there are some
-// cases where an owned account is not writable.
+// An account that was loaded to execute a transaction. The data slice is owned.
 pub const LoadedAccount = struct {
     pubkey: Pubkey,
     account: AccountSharedData,
-    is_owned: bool,
 
     pub fn deinit(self: LoadedAccount, allocator: Allocator) void {
-        if (self.is_owned) self.account.deinit(allocator);
+        self.account.deinit(allocator);
     }
 };
 
@@ -189,7 +185,7 @@ fn loadTransactionAccountsSimd186(
             slot,
             &account_key,
         );
-        errdefer if (loaded_account.is_owned) loaded_account.account.deinit(allocator);
+        errdefer loaded_account.account.deinit(allocator);
 
         try loaded.increase(
             loaded_account.loaded_size,
@@ -240,7 +236,6 @@ fn loadTransactionAccountsSimd186(
         loaded.accounts.appendAssumeCapacity(.{
             .account = loaded_account.account,
             .pubkey = account_key,
-            .is_owned = loaded_account.is_owned,
         });
     }
 
@@ -304,7 +299,7 @@ fn loadTransactionAccountsOld(
             slot,
             &account_key,
         );
-        errdefer if (loaded_account.is_owned) loaded_account.account.deinit(allocator);
+        errdefer loaded_account.account.deinit(allocator);
 
         try loaded.increase(
             loaded_account.loaded_size,
@@ -328,7 +323,6 @@ fn loadTransactionAccountsOld(
         loaded.accounts.appendAssumeCapacity(.{
             .account = loaded_account.account,
             .pubkey = account_key,
-            .is_owned = loaded_account.is_owned,
         });
     }
 
@@ -389,7 +383,6 @@ fn loadTransactionAccount(
     account: AccountSharedData,
     loaded_size: usize,
     rent_collected: u64,
-    is_owned: bool,
 } {
     if (key.equals(&runtime.sysvar.instruction.ID)) {
         @branchHint(.unlikely);
@@ -404,7 +397,6 @@ fn loadTransactionAccount(
             },
             .loaded_size = 0,
             .rent_collected = 0,
-            .is_owned = true,
         };
     }
 
@@ -423,7 +415,6 @@ fn loadTransactionAccount(
             .account = account,
             .loaded_size = 0,
             .rent_collected = 0,
-            .is_owned = true,
         };
     };
     errdefer account.account.deinit(allocator);
@@ -448,7 +439,6 @@ fn loadTransactionAccount(
         .account = account_shared_data,
         .loaded_size = account.loaded_size,
         .rent_collected = rent_collected.rent_amount,
-        .is_owned = true,
     };
 }
 
--- a/src/accountsdb/buffer_pool.zig
+++ b/src/accountsdb/buffer_pool.zig
@@ -3,7 +3,6 @@ const sig = @import("../sig.zig");
 const builtin = @import("builtin");
 const tracy = @import("tracy");
 
-const Allocator = std.mem.Allocator;
 const Atomic = std.atomic.Value;
 const IoUring = std.os.linux.IoUring;
 
--- a/src/runtime/check_transactions.zig
+++ b/src/runtime/check_transactions.zig
@@ -167,7 +167,6 @@ pub fn checkFeePayer(
                 .executable = maybe_nonce.?.account.executable,
                 .rent_epoch = payer_shared.rent_epoch,
             },
-            .is_owned = true,
         }) catch unreachable;
     } else {
         var rollback_payer = try payer_shared.clone(allocator);
@@ -179,7 +178,6 @@ pub fn checkFeePayer(
         rollbacks.append(.{
             .pubkey = fee_payer_key,
             .account = rollback_payer,
-            .is_owned = true,
         }) catch unreachable;
     }
 
@@ -394,7 +392,6 @@ fn checkLoadAndAdvanceMessageNonceAccount(
             .executable = nonce_account.executable,
             .rent_epoch = nonce_account.rent_epoch,
         },
-        .is_owned = true,
     };
 
     return .{
@@ -819,7 +816,6 @@ test "checkFeePayer: happy path with same nonce and fee payer" {
         .{
             .pubkey = transaction.fee_payer,
             .account = nonce_account,
-            .is_owned = true,
         },
         &sig.core.rent_collector.defaultCollector(10),
         &sig.core.FeatureSet.ALL_DISABLED,
@@ -888,7 +884,6 @@ test "checkFeePayer: happy path with separate nonce and fee payer" {
         .{
             .pubkey = Pubkey.initRandom(prng.random()),
             .account = nonce_account,
-            .is_owned = true,
         },
         &sig.core.rent_collector.defaultCollector(10),
         &sig.core.FeatureSet.ALL_DISABLED,


src/replay/svm_gateway.zig
```diff
@@ -92,21 +91,26 @@ pub const SvmGateway = struct {
...
+        for (batches) |batch| for (batch.transactions) |transaction| {
+            try account_preload.load(
+                &accounts,
                 allocator,
                 params.account_reader,
                 &transaction.accounts,
                 transaction.instructions,
             );
-        }
+        };
```
Sobeston: We talked about this before, but I'm not a fan of this because it reduces the pipelining we can do.

Let's pull out the AccountMap soon

src/runtime/account_loader.zig
```diff
@@ -70,613 +77,422 @@ pub const LoadedTransactionAccounts = struct {
...
+        0;
+
+    var is_owned = false;
+    const maybe_account: ?AccountSharedData =
+        if (key.equals(&runtime.sysvar.instruction.ID)) account: {
+            @branchHint(.unlikely);
+            is_owned = true;
+            break :account try constructInstructionsAccount(allocator, transaction);
+        } else map.get(key.*);
+
```
Sobeston: missing errdefer?

src/runtime/account_preload.zig
```diff
@@ -0,0 +1,210 @@
...
+pub fn initSufficientCapacity(allocator: Allocator, transactions: anytype) !AccountMap {
+    var map: AccountMap = .{};
+    try map.ensureUnusedCapacity(allocator, maxAccounts(transactions));
+
+    return map;
+}
+
+/// Counts the maximum number of accounts that may be needed to process these transactions.
+/// This is a naive upper bound that doesn't check for duplicates.
+pub fn maxAccounts(transactions: anytype) usize {
```
Sobeston: Do these need to be anytype - what types can this take?

src/runtime/account_preload.zig
```diff
@@ -0,0 +1,210 @@
...
+
+/// Initializes a new instance with all the accounts needed for all the
+/// provided transactions.
+pub fn initFromAccountsDb(
+    allocator: Allocator,
+    account_reader: SlotAccountReader,
+    transactions: []const RuntimeTransaction,
+) !std.AutoArrayHashMapUnmanaged(Pubkey, AccountSharedData) {
+    var map: AccountMap = .{};
+    try map.ensureUnusedCapacity(allocator, maxAccounts(transactions));
```
Sobeston: Could call initSufficientCapacity here

src/runtime/account_preload.zig
```diff
@@ -0,0 +1,210 @@
...
+fn getAccountSharedData(
+    allocator: Allocator,
+    reader: SlotAccountReader,
+    pubkey: Pubkey,
+) error{ OutOfMemory, GetAccountFailedUnexpectedly }!?AccountSharedData {
+    const account: sig.core.Account = reader.get(allocator, pubkey) catch |err| switch (err) {
+        error.OutOfMemory => return error.OutOfMemory,
+        error.FileIdNotFound,
+        error.InvalidOffset,
+        error.SlotNotFound,
```
Sobeston: An aside, we should get rid of these errors with accountsdb 2

src/runtime/testing.zig
```diff
@@ -158,10 +158,10 @@ pub fn createTransactionContext(
     }
 
     for (account_keys.items) |key| {
-        accounts.appendAssumeCapacity(TransactionContextAccount.init(
-            key,
-            account_cache.account_cache.getPtr(key) orelse unreachable,
-        ));
+        const cached_account = account_map.getPtr(key) orelse unreachable;
+        const account = try allocator.create(sig.runtime.AccountSharedData);
+        account.* = try cached_account.clone(allocator);
```
Sobeston, line 191: missing errdefer

Sobeston (changes requested):

conformance/src/txn_execute.zig
```diff
@@ -916,160 +910,78 @@ fn serializeOutput(
...
-    };
+    } else if (result.ok.outputs == null and result.ok.err != null) {
+        // This block exists solely to replicate a bug in solfuzz_agave.
+        // solfuzz_agave zips the rollback accounts with the list of all
+        // accounts, but they don't line up correctly. The rollback accounts are
+        // actually all writeable, but solfuzz_agave filters them out because it
+        // mistakenly compares them against the is_writable field from a totally
+        // different account. This is nonsense, but we need to replicate the bug
+        // in order to return a conformant set of accounts.
+        const writes = result.ok.writes.slice();
```
InKryption, line 949: Is there a tracking issue for this, either on our tracker or agave's? If yes, a link to either or both here would be good - if (somehow) not, we should open one maybe?

src/runtime/account_preload.zig
```diff
@@ -0,0 +1,210 @@
...
+    accounts: *const std.MultiArrayList(sig.core.instruction.InstructionAccount),
+    instructions: []const sig.runtime.InstructionInfo,
+) !void {
+    // we assume the largest is allowed
+    const max_data_len = sig.runtime.program.compute_budget.MAX_LOADED_ACCOUNTS_DATA_SIZE_BYTES;
+
+    var tx_data_loaded: LoadedTransactionAccounts = .DEFAULT;
+
+    { // load txes account_keys
+        for (accounts.items(.pubkey)) |account_key| {
```
InKryption: Curious, what's with the surrounding block?

src/runtime/transaction_execution.zig
```diff
@@ -467,7 +362,7 @@ test hasDuplicates {
 pub fn executeTransaction(
     allocator: std.mem.Allocator,
     transaction: *const RuntimeTransaction,
-    loaded_accounts: []const CachedAccount,
+    loaded_accounts: []LoadedAccount,
```
InKryption, line 348: Just a clarification on this change, is this now an output parameter, or does this just need to be mutable for some other reason? If the latter, a comment would be good.

src/runtime/account_loader.zig
```diff
@@ -70,613 +77,422 @@ pub const LoadedTransactionAccounts = struct {
...
-        // TODO: we should probably do this at init time
-        _ = is_writable;
+    var account = (try loadAccount(
+        map,
+        allocator,
+        transaction,
+        key,
+        is_writable,
+        feature_set.active(.formalize_loaded_transaction_data_size, slot),
+    )) orelse {
```
InKryption: I believe you can drop the parens here?
```suggestion
    var account = try loadAccount(
        map,
        allocator,
        transaction,
        key,
        is_writable,
        feature_set.active(.formalize_loaded_transaction_data_size, slot),
    ) orelse {
```

src/replay/svm_gateway.zig
dnut: > Let's pull out the AccountMap soon

done

src/runtime/account_loader.zig
dnut: 6a14047b106822320926032901b1d9ebbdada499

src/runtime/account_loader.zig
dnut: c851949860d09c78c5b0cb1a06a4c60a67af97fd

src/runtime/account_preload.zig
dnut: I have since removed all the code in this file

src/runtime/account_preload.zig
dnut: I have since removed all the code in this file

src/runtime/testing.zig
dnut: not my code but nice catch

0fc1cc4c9c6ea9aa86f260e29151634853eb3c70

src/runtime/account_preload.zig
dnut: not sure, I didn't write this code. Anyway I've now deleted this entire file, so it no longer exists.

src/runtime/transaction_execution.zig
dnut: The data here was always mutable. Transactions modify accounts.

Previously CachedAccount contained a mutable pointer, so the slice of CachedAccount was const and the mutations were performed through the nested pointer. Now LoadedAccount just directly contains the account without another layer of indirection.

Do you think every mutable pointer should be documented?

e8378241576d38652afb1892c44f4698369bc7a7

src/accountsdb/account_store.zig
```diff
@@ -34,6 +34,17 @@ pub const AccountStore = union(enum) {
         };
     }
 
+    pub fn forSlot(self: AccountStore, slot: Slot, ancestors: *const Ancestors) SlotAccountStore {
+        const zone = tracy.Zone.init(@src(), .{ .name = "AccountStore.forSlot" });
+        defer zone.deinit();
```
Sobeston: Curious if this tracy zone was useful, it should return instantly

src/accountsdb/account_store.zig
```diff
@@ -173,17 +184,62 @@ pub const SlotModifiedIterator = union(enum) {
...
+pub const SlotAccountStore = union(enum) {
+    accounts_db: struct { *AccountsDB, Slot, *const Ancestors },
+    /// Contains many versions of accounts and becomes fork-aware using
+    /// ancestors, like accountsdb.
+    thread_safe_map: struct { *ThreadSafeAccountMap, Slot, *const Ancestors },
+    account_shared_data_map: struct {
+        Allocator,
+        *std.AutoArrayHashMapUnmanaged(Pubkey, AccountSharedData),
+    },
+    noop,
```
Sobeston, line 193: Not sure about this struct, I mean we don't ever need the ancestors to put. Can't we just use the AccountStore, knowing the slot already?

src/runtime/account_loader.zig
```diff
@@ -70,613 +78,429 @@ pub const LoadedTransactionAccounts = struct {
...
-    // one isn't intended to be meaningful.
-    // NOTE: we could take this field out, and have the caller deinit it from inside LoadedTransactionAccounts
-    sysvar_instruction_account_datas: std.ArrayListUnmanaged(AccountSharedData) = .{},
+/// Wraps calls to AccountsDB and convert all errors except OutOfMemory into AccountsDBError.
+pub fn wrapDB(item: anytype) AccountLoadError!@typeInfo(@TypeOf(item)).error_union.payload {
+    return item catch |err| switch (err) {
+        error.OutOfMemory => error.OutOfMemory,
+        else => error.AccountsDBError,
+    };
+}
```
Sobeston, line 99: I agree that accountsDB should only return OutOfMemory, or some "internal error" value like this. But this function feels a bit odd? Should maybe exist elsewhere. Not a big deal, can resolve this.

src/runtime/account_loader.zig
```diff
@@ -70,613 +78,429 @@ pub const LoadedTransactionAccounts = struct {
...
 
-    // NOTE: we might want to later add another field that keeps a copy of all writable accounts.
+/// Assumes `transaction` was in initFromAccountsDb's `transactions` parameter.
```
Sobeston: Isn't initFromAccountsDb gone?

src/runtime/account_loader.zig
```diff
@@ -70,613 +78,429 @@ pub const LoadedTransactionAccounts = struct {
...
+/// Reports account loading errors.
+/// By the time this is called, we have no dependency on accountsdb.
+pub fn loadTransactionAccounts(
+    // note: think we make this a *const by moving sysvar instruction account construction into init
```
Sobeston: comment seems outdated

dnut: fef41bef73d56b8125d71264d97444fbabe24ec1

src/runtime/account_loader.zig
dnut: fef41bef73d56b8125d71264d97444fbabe24ec1

conformance/src/txn_execute.zig
dnut: I'll open an issue in solfuzz_agave, but it will take some time to properly explain and construct an example to reproduce the bug. I added a ticket to our backlog so I don't forget about it

fd2cd1af4d28d3faf21fff15c93e07d8c92ba237

https://github.com/orgs/Syndica/projects/2/views/10?pane=issue&itemId=139114012

src/accountsdb/account_store.zig
dnut: I agree AccountStore is satisfactory when all you need to do is write. But this is used in contexts where both writes *and* reads need to happen. We need to pass the ancestors into those contexts somehow.

There is some awkwardness inherent to the asymmetry of the inputs for writes and reads, and the fact that reads need more information than writes, which seems worse than the opposite asymmetry. I'm not sure how to avoid that awkwardness without encapsulating full fork-awareness into AccountsDB.

src/accountsdb/account_store.zig
dnut: I probably wrote this function by copy and pasting another one without thinking.

d36a1cdfd24508fbd70c1b05915114401fa37bd2

src/runtime/account_loader.zig
dnut: I thought about adding this to the account store but I'm torn about it. The moment you consolidate to a single error, you lose information about what went wrong (unless error tracing is enabled, which won't be used in production). I added this here because I liked how there was a straightforward error set previously and I didn't want to clutter it. But actually I'm doubting the utility of this, and I think I'd rather just define more comprehensive error sets and propagate errors as-is without converting them. I think error conversions should only occur if the original error is recorded in logs or metrics.

src/accountsdb/account_store.zig
Sobeston: Yeah I see, not sure of a better away to do this right now. Not that important anyway, feel free to resolve

src/runtime/account_loader.zig
Sobeston: > I think error conversions should only occur if the original error is recorded in logs or metrics.

Yeah I think that's maybe what we should do.

Feel free to resolve this for now, it's a bit out of scope

Sobeston (approved):

PR #1049, opened 2025-10-31
feat(version): add `Clients` and set our version to `Sig`

Rexicon226:
let's wait for https://github.com/anza-xyz/agave/pull/8735 to merge, since they still might change the IDs

### DIFF
--- a/src/version/version.zig
+++ b/src/version/version.zig
@@ -10,7 +10,7 @@ pub const ClientVersion = struct {
     patch: u16,
     commit: u32, // first 4 bytes of the sha1 commit hash
     feature_set: u32, // first 4 bytes of the FeatureSet identifier
-    client: Clients,
+    client: ClientId,
 
     pub const CURRENT: ClientVersion = .{
         .major = version.major,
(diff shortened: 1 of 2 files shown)


src/version/version.zig
```diff
@@ -19,11 +19,38 @@ pub const ClientVersion = struct {
         .commit = std.fmt.parseInt(u32, version.build orelse "0", 16) catch
             @compileError("failed to parse build"),
         .feature_set = 0,
-        .client = 1,
+        .client = .sig,
+    };
+
+    const Clients = enum(u16) {
```
dnut: An instance of this type represents a singular client, not plural. Also it's a client's "id" so we can specific about that as well.
```suggestion
    const ClientId = enum(u16) {
```

src/version/version.zig
```diff
@@ -19,11 +19,38 @@ pub const ClientVersion = struct {
...
+        solana_labs = 0,
+        jito_labs = 1,
+        frankendancer = 2,
+        agave = 3,
+        agave_paladin = 4,
+        agave_bam = 5,
+        sig = 6,
+        firedancer = 7,
```
dnut: How about you change it to be consistent with this file for now.

https://github.com/solana-foundation/solana-validator-client-ids/blob/main/client-ids.csv

We can merge it like this and open additional PRs whenever that repo is updated.

We can make an arbitrary best-guess for sig, since we need to announce *something* to the network. No harm in merging a guess now since we're currently using 1 which is already a wrong guess.

Rexicon226: cd286a6fdb0919cf8b4231039f290baa23c03220

src/version/version.zig
Rexicon226: cd286a6fdb0919cf8b4231039f290baa23c03220

dadepo (approved):

PR #1069, opened 2025-11-07
feat(accounts_db): rework the accounts database from scratch

Rexicon226:


src/accountsdb/two/Unrooted.zig
```diff
@@ -0,0 +1,175 @@
...
+
+        if (ancestors.containsSlot(index.slot)) {
+            if (index.slot < best_slot) continue;
+            const data = index.entries.get(address) orelse continue;
+            result = data.asAccount();
+            best_slot = index.slot;
+        }
+    }
+
+    return result;
```
dnut, line 133: This returns a pointer to the data in the database, which is good because it avoids copies, but we may run into lifetime issues in the future. When the slot is rooted, the pointer will be invalidated, but it may still be in use by callers of this function. With the current implementation of Sig, I don't think this will happen anywhere, but once we implement RPC or move rooting out of the replay thread, we'll need a solution to ensure the lifetime is sufficient. One solution that comes to mind is reference counting.

src/accountsdb/two/Unrooted.zig
```diff
@@ -0,0 +1,175 @@
...
+    var result: ?Account = null;
+
+    for (self.slots) |*index| {
+        if (index.is_empty.load(.acquire)) continue;
+
+        index.lock.lockShared();
+        defer index.lock.unlockShared();
+
+        if (ancestors.containsSlot(index.slot)) {
+            if (index.slot < best_slot) continue;
```
dnut: `index.slot > best_slot` is cheaper and more likely to be false than `ancestors.containsSlot(index.slot)`
```suggestion
        if (index.slot > best_slot and ancestors.containsSlot(index.slot)) {
```

Actually, can't we just iterate from the highest slot to the lowest slot? The first match you find: return it. No need to keep iterating if you already found the best slot. I understand the slots may not be in sequential numeric order, but they should be topologically sorted, which is good enough for this approach to work as long as the ancestors input is well-formed.

src/accountsdb/two/Unrooted.zig
```diff
@@ -0,0 +1,175 @@
...
+    data: AccountSharedData,
+) !void {
+    const index = slot % MAX_SLOTS;
+    const entry = &self.slots[index];
+
+    entry.lock.lock();
+    defer entry.lock.unlock();
+
+    if (entry.slot != slot) {
+        std.debug.assert(entry.is_empty.load(.acquire));
```
dnut, line 102: The validator will crash here if consensus fails to root a slot after executing 4096 slots. Could you add this as a comment?

src/accountsdb/two/Unrooted.zig
```diff
@@ -0,0 +1,175 @@
...
+    // TODO: it's possible for us to pre-populate a SlotIndex with pubkeys (and null values) in
+    // advance of replaying the slot, which would let us drop this mutex entirely.
+    lock: std.Thread.RwLock,
+    slot: Slot,
+    is_empty: Atomic(bool),
+    entries: std.ArrayHashMapUnmanaged(Pubkey, AccountSharedData, Context, true),
+
+    const Context = struct {
+        pub fn hash(_: Context, pubkey: Pubkey) u32 {
+            return @bitCast(pubkey.data[0..4].*);
```
dnut: It's common for people to grind vanity keys to target the first few bytes, so they are not the most random.

Also there are some pre-defined vanity keys in the validator that are not on the curve. Sysvars come to mind. All of them start in base58 with "Sysvar". But then all the validator vanity keys typically end in "...11111" after whatever words they wanted to say, so taking the end or middle would not be ideal either.

We could take bytes [4..8] to avoid most of these collisions, but it still leaves one: the system and vote programs.

src/accountsdb/two/Two.zig
```diff
@@ -0,0 +1,174 @@
...
+
+    // Get write lock on unrooted slot, remove from unrooted
+    {
+        rooted_index.lock.lock();
+        defer rooted_index.lock.unlock();
+
+        std.debug.assert(rooted_index.slot == newly_rooted_slot);
+        std.debug.assert(!rooted_index.is_empty.load(.acquire));
+
+        rooted_index.entries.clearRetainingCapacity();
```
dnut: Doesn't each account's data need to be freed?

src/accountsdb/two/Unrooted.zig
Rexicon226: If by reference counting you mean we'd track references to accounts stored in the unrooted storage, and the RPC server would acquire these references, this is not a good idea. We cannot allow the RPC server to in any way block or increase the memory pressure of the database. Instead I imagine we'll be fine just copying out the data on the rpc side (`getCopy` type interface), or perhaps having a streaming approach where we'd write account data into the json object directly.

As for rooting, reference counting there is still a terrible idea. All we need is a bit of synchronization between the `get` of a replay thread and the point at which we free the data in the unrooted storage. In 99% of cases the two events won't race and the replay thread will see the data in the rooted storage first, but just to be safe, we can have some sort of synchronization on the `is_empty` *after* we get the account data, a sort of CAS approach.
I'll think about this some more, but it shouldn't block merging this PR, since we haven't moved rooting off of the replay thread.

src/accountsdb/two/Unrooted.zig
Rexicon226: The first suggestion makes sense.

Regarding the second one, maybe I'm missing something, but I don't understand how that can work. Consider the case where the latest known slot for ancestors is at index `10`, and the previously known one is at index `4090`. We iterate through it backwards, and see the one in `4090` first and just say that's the latest one? We need to search the entire ring, because otherwise how can we know at which point in time that specific fork's ancestors wrapped around. Did I not correctly understand your suggestion? The whole point is that it's a ring buffer.

I guess this could be solved with some more advanced tracking, but our `get`s are already quite cheap, and if we want to improve their performance that should be done by reducing the number of indices through a linked list rather than tracking wrapping points.

src/accountsdb/two/Two.zig
Rexicon226: Hmm yeah, I think so. 

Aside: We really need to take a better look at the memory management of the account data it kinda sucks right now. I think moving to per-slot and per-epoch arenas will happen first, then we can revisit removing duplication here. Also writing our own rooted db might be somewhere there, in case mainnet perf issues come from sqlite fundamentally.

src/accountsdb/two/Unrooted.zig
Rexicon226: I see your point. 

That said, why would sysvars and system programs ever be in the unrooted storage? This is just impossible, we write directly to the rooted storage when initializing it. Also keep in mind that this map exists *per-slot*, you'd need many accounts in the same slot to have excessive amounts of vanity, all of the same type.

I think the best approach here is to do a sort of murmur-like construct to bucket the bytes. Would be pretty much free and should be good enough. 

In the future, I have great plans for creating a specialized map type that's built for being keyed with public keys. Since it's a fixed array input, many interesting optimizations are possible with radix-trees and left/right sums.

Rexicon226: fcaf023fda6f0f6e218ab270659fd4594b759f86

src/accountsdb/two/Two.zig
Rexicon226: fcaf023fda6f0f6e218ab270659fd4594b759f86

src/accountsdb/two/Unrooted.zig
Rexicon226: fcaf023fda6f0f6e218ab270659fd4594b759f86

src/accountsdb/two/Unrooted.zig
Rexicon226: Applies first suggestion: fcaf023fda6f0f6e218ab270659fd4594b759f86

dnut: The problem is that I wrote this comment without actually thinking about how you're assigning indices to slots. I assumed slots are contiguously appended to the buffer by tracking and incrementing a head index, but obviously that's not how this works.

The current approach seems fine for now. It does seem like a lot of unnecessary iteration, but I don't have a good sense of the actual performance cost.

src/accountsdb/two/Unrooted.zig
dnut: The mention of reference counting was meant to be illustrative. I don't think we need to solve the problem right now, and your points make sense. I just want to keep this on our radar so we address it before introducing problems. Feel free to resolve this.

https://github.com/orgs/Syndica/projects/2/views/10?pane=issue&itemId=139769269

src/accountsdb/two/Unrooted.zig
Rexicon226: Ah wait, you confused me there for a minute!
Your first suggestion isn't valid either, I've removed it from the commit. I had a good feeling it wasn't valid when I first read it, but didn't spend the time to consider it. I knew I had done that on purpose when i was writing it originally.

dnut: What's wrong with my first suggestion? It's logically equivalent to the original code, except that I used `>` instead of `>=`. You could use `>=`, but it shouldn't matter since you won't encounter a slot multiple times.

src/accountsdb/two/Unrooted.zig
dnut: Sysvars and native programs have accounts. Transactions may specify them as ordinary accounts that need to be loaded just like any other account. We have code to inject sysvars into accountsdb for every slot, since they change from slot to slot.

https://github.com/Syndica/sig/blob/85f9172b89bbb109a17a0586512ae58dbcad5c08/src/replay/update_sysvar.zig#L346

dnut: leaving open for visibility. feel free to resolve this comment

src/accountsdb/two/Two.zig
```diff
@@ -0,0 +1,175 @@
...
+    self.unrooted.deinit(allocator);
+}
+
+/// Clones the account shared data.
+pub fn put(self: *Db, slot: Slot, address: Pubkey, data: AccountSharedData) !void {
+    // used for tests, tracking what the largest slot is and
+    // making sure we don't write to something before that.
+    if (builtin.is_test)
+        if (self.rooted.largest_rooted_slot) |lrs|
+            if (lrs >= slot) return error.CannotWriteRootedSlot;
```
dnut: Why is this only run during tests? Tests expect this error for a reason. The whole point is to reflect the way we expect the code to behave in production. Doesn't code like this defeat the purpose of the test? At the least, it makes the makes the test code misleading.

src/accountsdb/two/Two.zig
```diff
@@ -0,0 +1,175 @@
...
+
+    const cloned = try data.clone(self.allocator);
+    errdefer cloned.deinit(self.allocator);
+    try self.unrooted.put(self.allocator, slot, address, cloned);
+}
+
+// TODO: this should be trivial to put on another thread; this currently blocks the main replay
+// thread for 30-40ms per slot on testnet.
+pub fn onSlotRooted(self: *Db, newly_rooted_slot: Slot) error{FailedToRoot}!void {
+    const rooted_index: *Unrooted.SlotIndex = for (self.unrooted.slots) |*index| {
```
dnut:
```suggestion
    const unrooted_index: *Unrooted.SlotIndex = for (self.unrooted.slots) |*index| {
```

src/accountsdb/two/Two.zig
```diff
@@ -0,0 +1,175 @@
...
+        if (index.is_empty.load(.acquire)) continue;
+        if (index.slot != newly_rooted_slot) continue;
```
dnut: If rooting moves to another thread, this could race with a write in `Unrooted.put`. What do you think about unifying the `slot` and `is_empty` fields to a single atomic u64?

src/accountsdb/two/Rooted.zig
```diff
@@ -0,0 +1,441 @@
...
+
+/// Only used in tests.
+pub fn getLargestRootedSlot(self: *const Rooted) ?Slot {
+    if (!builtin.is_test) @compileError("only used in tests");
+    return self.largest_rooted_slot;
+}
+
+pub fn err(self: *Rooted, code: c_int) !void {
+    if (code == OK) return;
+    std.debug.print("err ({}): {s}\n", .{ code, sql.sqlite3_errmsg(self.handle) });
```
dnut: Can we use the logger here?

src/accountsdb/two/Two.zig
```diff
@@ -0,0 +1,175 @@
...
+        std.debug.assert(rooted_index.slot == newly_rooted_slot);
+        std.debug.assert(!rooted_index.is_empty.load(.acquire));
+
+        self.rooted.beginTransation() catch return error.FailedToRoot;
+
+        var entries = rooted_index.entries.iterator();
+        while (entries.next()) |entry| {
+            self.rooted.put(entry.key_ptr.*, newly_rooted_slot, entry.value_ptr.*) catch
+                return error.FailedToRoot;
+        }
```
dnut: shouldn't there be an errdefer to rollback the transaction?

src/accountsdb/two/Rooted.zig
```diff
@@ -0,0 +1,441 @@
...
+
+    return self;
+}
+
+pub fn deinit(self: *Rooted) void {
+    _ = sql.sqlite3_close(self.handle);
+}
+
+/// Call this before a thread that accesses Rooted closes. Safe to call multiple times.
+pub fn deinitThreadLocals() void {
```
dnut, line 114: I don't see how we're actually going ensure this is called. I think in most cases these handles are going to leak.

src/accountsdb/two/Rooted.zig
```diff
@@ -0,0 +1,441 @@
...
+        get_stmt = null;
+    }
+}
+
+fn accountsHash(self: *Rooted) !sig.core.LtHash {
+    const zone = tracy.Zone.init(@src(), .{ .name = "Rooted.accountsHash" });
+    defer zone.deinit();
+
+    const query =
+        \\SELECT address, owner, data, lamports, executable, rent_epoch, last_modified_slot FROM entries;
```
dnut, line 130: how is this not failing the line length limit lint?
```suggestion
        \\SELECT address, owner, data, lamports, executable, rent_epoch, last_modified_slot
        \\FROM entries;
```

src/accountsdb/two/Rooted.zig
```diff
@@ -0,0 +1,441 @@
...
+
+/// Returns `null` if no such account exists.
+///
+/// The `data` field in the returned `AccountSharedData` is owned by the caller and is allocated
+/// by the provided allocator.
+///
+/// TODO: we really don't want to be doing these clones, so some other solution would be good.
+pub fn get(self: *Rooted, allocator: std.mem.Allocator, address: Pubkey) !?AccountSharedData {
+    return self.getInner(allocator, address) catch |e| switch (e) {
+        error.SqliteError => @panic(""),
```
dnut: why is this a panic?

src/accountsdb/two/Two.zig
Rexicon226: I was planning to do the same thing, I already did something similar in my unrooted 2.0 design. 
I'll wait for offloaded rooting, since there is a number of things we'll need to change, and I want to spend the time thinking and verifying all the threading stuff at the same time.

src/accountsdb/two/Rooted.zig
Rexicon226: I mean we don't ensure that file descriptors are closed either, but we make it work. This is mainly needed for tests, since otherwise sqlite retain the state of the previously executed tests. Usually if you're writing a test and forget to close the statements, you'll have some sort of failure from the next test to use accounts db.

src/accountsdb/two/Rooted.zig
Rexicon226: We just shouldn't be logging here.

src/accountsdb/two/Rooted.zig
Rexicon226: Probably should have a `std.debug.panic` instead. This error shouldn't be returned unless something goes very wrong, at which point we should just abort the program.

src/accountsdb/two/Two.zig
Rexicon226: Good point, I think I originally made this only in tests because I believed it would need some extra synchronization, but I see now that's not the case :). Again, I'm going to keep assuming that flushing slots happens on the same thread, but I'll make it always be checked.

src/accountsdb/two/Two.zig
Rexicon226: This should never fail unless something is super wrong, but I guess maybe? cc @Sobeston wdyt?

src/accountsdb/two/Unrooted.zig
Rexicon226: Nah, you're right, I just confused myself again lol.

src/accountsdb/two/Two.zig
Sobeston: Yeah this shouldn't ever fail. But I guess it wouldn't hurt?

src/accountsdb/two/Rooted.zig
dnut: Care to explain your reasoning? I strongly believe that we should be using the logger for any messages like this because they won't be as easily observable otherwise.

src/accountsdb/two/Two.zig
dnut: I think this should either be `unreachable`, or if we return an error then we should assume it might happen and handle it properly.

src/accountsdb/two/Rooted.zig
dnut: so you're saying we should just leak these in production and it doesn't matter?

src/accountsdb/two/Rooted.zig
Rexicon226: Nothing should be failing here, unless something goes really wrong. We'll panic if the case is tripped, with the sqlite3 error message (that `@panic("") to be replaced, still working through changes locally).
Passing a logger down all the way here is pointless.

src/accountsdb/two/Two.zig
Rexicon226: Up to @Sobeston, this was all his logic. iirc his idea was the still be able to execute even if we fail to root for a few thousand more slots, although idk why we'd want to do that.

src/accountsdb/two/Rooted.zig
Rexicon226: What do you mean leak, we just never close the db because it's running. We have no mechanism for shutting down the validator currently (like agave's shutdown web hook).

src/accountsdb/two/Rooted.zig
Rexicon226: We currently ignore these cases,
https://github.com/Syndica/sig/blob/c17779dd873c85497e904eded0aaed1f13f0e1cd/scripts/style.py#L199
and I think that's a good thing.

src/accountsdb/two/Rooted.zig
Rexicon226: db51e427bf2c4d860a02296e9cfad0700999c5f4

src/accountsdb/two/Two.zig
Rexicon226: db51e427bf2c4d860a02296e9cfad0700999c5f4 ended up needing it anyways to fix that slot hash mismatch

src/accountsdb/two/Unrooted.zig
Rexicon226: db51e427bf2c4d860a02296e9cfad0700999c5f4

src/accountsdb/two/Two.zig
Sobeston: Thinking we should just have this panic. It is possible for this to fail (unreachable would be inappropriate), but we can't really recover from it.

src/accountsdb/two/Rooted.zig
dnut: i find sql queries more readable when they're on multiple lines like i suggested, but it's not important

i agree it's reasonable to allow lines starting with `\\` to exceed the limit, but i'd like to remove the ignore for `//`. if anything, i'd like comments to use *fewer* columns than code.

Rexicon226: I personally find the single line clearer.

I also strongly disagree with putting any constraint on the contents of comments, there are many cases when they need to express arbitrary information. Their formatting should be up to the discretion of the author, presuming they are reasonably formatted for the information being conveyed.

I would prefer we not have any column limit, however I fear the code will only get sloppy, and many people likely don't want to spend the time thinking about how to format their code (even though they should!). There is a underlying feeling of suggesting mass formatting changes on reviews as "too pedantic".

dnut: Regarding the SQL, I'm surprised you find it more readable that way, but it's not consequential enough to debate so I guess just leave it.

Regarding comments, my take is that comments exist to be readable, and if you can't express your comments within 100 characters of width, then you're failing at making a readable comment. I don't see any examples in our codebase that have a good reason for exceeding that limit other than URLs.

src/accountsdb/two/Rooted.zig
Sobeston: Leaking these is also fine for now, we have an upper bound of the amount of threads that use this. (X replay threads + one or two extra).

src/accountsdb/two/Two.zig
Sobeston: This is the SlotIndex for the newly rooted slot. So maybe newly_rooted_index? `unrooted` feels a bit confusing

Sobeston: Actually this variable was removed / this is outdated

src/accountsdb/two/Two.zig
Sobeston: Currently when rooting is happening, no replay worker threads will be running. So the rooted store can be seen as single writer *or* multiple reader. 

For the time being this is safe, and I'm not sure that having rooting and replay threads running at the same time is a good idea for a bit. In testing I found that the options required to keep sqlite safe under those conditions caused major slowdowns. So I'd suggest solving this at another time

src/accountsdb/two/Rooted.zig
Sobeston: 4078e518038050462ec2b1e0978a41eec7d443e0

src/accountsdb/two/Two.zig
Sobeston: Removed this error. e1fb5f30bc6036d8cba9bdcc000625f18842c230

src/accountsdb/two/Rooted.zig
```diff
@@ -0,0 +1,441 @@
...
+            const len: usize = @intCast(sql.sqlite3_column_bytes(stmt, 2));
+            if (len == 0) break :blk &.{}; // sqlite returns null pointers for 0-len data
+            const ptr: [*]const u8 = @ptrCast(sql.sqlite3_column_blob(stmt, 2));
+            break :blk ptr[0..len];
+        };
+
+        const lamports: u64 = @bitCast(sql.sqlite3_column_int64(stmt, 3));
+        const executable: bool = sql.sqlite3_column_int(stmt, 4) != 0;
+        const rent_epoch: u64 = @bitCast(sql.sqlite3_column_int64(stmt, 5));
+        const last_modified_slot: Slot = @bitCast(sql.sqlite3_column_int64(stmt, 6));
```
yewman: Do we need to be reading this?

src/accountsdb/two/Rooted.zig
```diff
@@ -0,0 +1,441 @@
...
+
+fn err(self: *Rooted, code: c_int) void {
+    if (code == OK) return;
+    std.debug.panic(
+        "internal accountsdb sqlite error ({}): {s}\n",
+        .{ code, sql.sqlite3_errmsg(self.handle) },
+    );
+}
+
+pub fn beginTransation(self: *Rooted) void {
```
yewman:
```suggestion
pub fn beginTransaction(self: *Rooted) void {
```

src/accountsdb/two/Rooted.zig
```diff
@@ -0,0 +1,441 @@
...
+    self.err(sql.sqlite3_exec(self.handle, "BEGIN TRANSACTION;", null, null, null));
+}
+
+pub fn commitTransation(self: *Rooted) void {
```
yewman:
```suggestion
pub fn commitTransaction(self: *Rooted) void {
```

yewman (changes requested):

src/accountsdb/two/Rooted.zig
Rexicon226: This function is pretty much for debugging, we don't actually call it anywhere. It was useful before, do we still want to keep it @Sobeston. It isn't really practical as it runs extremely slowly, and we want to end up creating our own sqlite dbs in the future, and that will be able to lthash much better.

src/identity.zig
```diff
@@ -25,105 +23,108 @@ pub const SigningKeys = struct {
     authorized_voters: []const KeyPair,
 };
 
-/// Returns the keypair from {app data directory}/{IDENTITY_KEYPAIR_PATH} or creates a new one
+/// Returns the keypair from $APP_DATA/IDENTITY_KEYPAIR_PATH or creates a new one
 /// if the file does not exist. If the file is invalid, an error is returned.
 pub fn getOrInit(allocator: std.mem.Allocator, logger: Logger) !KeyPair {
-    const app_data_dir_path = try std.fs.getAppDataDir(allocator, "sig");
-    defer allocator.free(app_data_dir_path);
+    const data_directory_path = try std.fs.getAppDataDir(allocator, "sig");
```
dnut, line 29: I don't see how the changes in this file are relevant in any way to accountsdb. In the future please make changes like this in separate PRs.

dnut: feel free to resolve this conversation

src/accountsdb/two/Rooted.zig
Sobeston: I vote keep it, it does successfully verify the snapshot if you do choose to use it. But yes 1) it isn't exposed by the cli, and 2) it is quite slow. I think we should improve it in a follow-up.

dnut (approved):

PR #1070, opened 2025-11-08
perf(thread): simplify `HomogeneousThreadPool`

kprotty:
The HomogeneousThreadPool was inefficient with the use of mutexes/conditions per task and spinning to join. It also leaked task memory through `MemoryPool` if the caller never joins what was spawned. This PR replaced it with a simpler API under the name `ScopedThreadPool`, taking the function (instead of a task struct) for convenience and solving the task leak with an internal free list.

There aren't many use cases of it. Eventually, we should remove `ScopedThreadPool` and use `ThreadPool.schedule` directly as needed (an example is the new `TransactionScheduler` merged in #1074 )

### DIFF
--- a/src/utils/thread.zig
+++ b/src/utils/thread.zig
@@ -1,4 +1,5 @@
 const std = @import("std");
+const tracy = @import("tracy");
 const sig = @import("../sig.zig");
 
 const Allocator = std.mem.Allocator;
@@ -139,6 +140,9 @@ pub fn ScopedThreadPool(comptime func: anytype) type {
             args: Args,
 
             fn run(pool_task: *ThreadPool.Task) void {
+                const zone = tracy.Zone.init(@src(), .{ .name = "ScopedTask.run" });
+                defer zone.deinit();
+
                 const self: *ScopedTask = @alignCast(@fieldParentPtr("pool_task", pool_task));
                 switch (@typeInfo(ReturnType)) {
                     .error_union => @call(.auto, func, self.args) catch |e| {


src/utils/thread.zig
```diff
@@ -100,251 +92,119 @@ pub fn spawnThreadTasks(
...
-        }
+        const Self = @This();
+        const ScopedTask = struct {
+            pool_task: ThreadPool.Task = .{ .callback = run },
+            next: ?*ScopedTask = null,
+            pool: *Self,
+            args: Args,
+
+            fn run(pool_task: *ThreadPool.Task) void {
+                const self: *ScopedTask = @alignCast(@fieldParentPtr("pool_task", pool_task));
```
dnut, line 146: previously we had a tracy zone here. is it not useful? cc: @Sobeston

kprotty: I think it's better to have a zone on the scoped task function itself. In this case, the direct users already have one:
* `RepairRequester.sendRepairRequestBatch`
* `tar.writeToFile`

along with the indirect users via `spawnThreadTasks` in accountsdb:
* `loadAndVerifyAccountsFilesMultiThread`
* `mergeThreadIndexesMultiThread`
* `getHashesFromIndex`

Sobeston: The zone:

```zig
        fn run(pool_task: *ThreadPool.Task) void {
            const zone = tracy.Zone.init(
                @src(),
                .{ .name = "HomogeneousThreadPool.run: " ++ @typeName(TaskType) },
            );
            defer zone.deinit();

            var self: *Self = @fieldParentPtr("pool_task", pool_task);

            self.result = self.typed_task.run();
```

I found this to be useful in my branch, but now the zone isn't showing up when I build on main. Do we now call it differently? :thinking:

Sobeston: If we can get a tracy "run" zone on any task run (ideally with typing, if available) in this PR, that'd be great.

If it's too much work then we can go without it

kprotty: https://github.com/Syndica/sig/pull/1070/commits/a9dbcfdbac3088169dc3e437ee0cb9b2cb0b9d55

InKryption (approved):

dnut (approved):

PR #1071, opened 2025-11-08
chore(consensus): Add more logging in consensus

dadepo:


src/consensus/replay_tower.zig
```diff
@@ -317,9 +321,57 @@ pub const ReplayTower = struct {
...
+            .log("┌──────────────┬────────────────┬──────────────────┬──────────────┐");
+        self
+            .logger
+            .info()
+            .log("│     Slot     │ Confirmations  │  Lockout Expires │  Lockout Len │");
+        self
+            .logger
+            .info()
+            .log("├──────────────┼────────────────┼──────────────────┼──────────────┤");
+
```
InKryption, line 332: I don't think this is the right way to do this; the logger interface isn't really designed for this kind of stylized, structured, tabular output; isn't it going to get interspersed with various other logs?

PR #1072, opened 2025-11-08
feat(replay,db): performance improvements

Rexicon226:


src/accountsdb/two/Rooted.zig
```diff
@@ -0,0 +1,439 @@
...
+fn accountsHash(self: *Rooted) !sig.core.LtHash {
+    const zone = tracy.Zone.init(@src(), .{ .name = "Rooted.accountsHash" });
+    defer zone.deinit();
+
+    const query =
+        \\SELECT address, owner, data, lamports, executable, rent_epoch, last_modified_slot FROM entries;
+    ;
+
+    var stmt: ?*sql.sqlite3_stmt = undefined;
+    if (sql.sqlite3_prepare_v2(self.handle, query, -1, &stmt, null) != OK)
```
InKryption: `undefined` is probably fine here but makes me uneasy, would prefer erring on the side of distrusting whatever's on the other side of the C ABI boundary and passing initializing to `null`.

InKryption (approved):

src/cmd.zig
```diff
@@ -2391,3 +2391,138 @@ fn loggingPanic(message: []const u8, first_trace_addr: ?usize) noreturn {
     sig.trace.logfmt.writeLog(writer, "panic", .err, .{}, "{s}", .{message}) catch {};
     std.debug.defaultPanic(message, first_trace_addr);
 }
+
+// https://github.com/facebook/folly/blob/1c8bc50e88804e2a7361a57cd9b551dd10f6c5fd/folly/memcpy.S
+export fn memcpy(maybe_dest: ?[*]u8, maybe_src: ?[*]const u8, len: usize) callconv(.C) ?[*]u8 {
```
Sobeston: We can move this to another file to be neater. Not important, feel free to resolve.

src/replay/service.zig
```diff
@@ -428,6 +428,11 @@ pub fn newSlotFromParent(
 
     var ancestors = try parent_constants.ancestors.clone(allocator);
     errdefer ancestors.deinit(allocator);
+
+    // TODO: Ponder on this number some more. We need to move to using a bitset
+    // for ancestors, which will solve the issue where it just keeps growing forever.
+    // The unrooted db also requires us to cap the length of ancestors at below MAX_SLOTS.
+    if (ancestors.ancestors.count() > 512) ancestors.ancestors.orderedRemoveAt(0);
```
Sobeston: It would give me some confidence if this came along with some debug asserts and/or warn logs.

src/accountsdb/account_store.zig
```diff
@@ -908,7 +900,15 @@ test "insertion out of order" {
         break :slot slots[random.uintLessThan(usize, slots.len)];
     };
     const pk_of_ones: Pubkey = .{ .data = @splat(1) };
-    setRootedLargestSlotForTest(&simple_state, &real_state, slot_to_try_write_while_rooted);
+    // setRootedLargestSlotForTest(&simple_state, real_state, slot_to_try_write_while_rooted);
```
Sobeston: forgot this comment?

src/accountsdb/two/Unrooted.zig
```diff
@@ -96,15 +96,24 @@ pub fn put(
...
+        // of this in any way, so we must assume that any entries that remain after
+        // MAX_SLOTS are from bad/dropped forks.
+        // current solution is to just comment out the assert that we've managed the entry before,
+        // and clear it "on-demand" when we wrap back around to it.
+
         // If consensus fails to root a slot after executing MAX_SLOTS - 1 other slots
         // this will panic. It should be reasonable to assume that doesn't happen in
         // a real-world situation.
-        std.debug.assert(entry.is_empty.load(.acquire));
+        // std.debug.assert(entry.is_empty.load(.acquire));
```
Sobeston: can we make it a panic then?

src/accountsdb/two/Two.zig
```diff
@@ -164,64 +164,3 @@ pub fn slotModifiedIterator(self: *Db, slot: Slot) ?SlotModifiedIterator {
         .cursor = 0,
     };
 }
```
yewman: Why are we removing these?

src/cmd.zig
Rexicon226: Yeah, we'd want to use this in all our binaries in theory (although the sig one is the only one that really matters). Was thinking of putting it into a separate file (memcpy.zig) and then just compiling and linking it. Wdyt?

src/replay/service.zig
Rexicon226: Reworking it all into a bitset approach now anyways, this is going to exist for like 3 hours until I implement that.
Although what logs are you looking for here? This is a completely natural state to be have the ancestors set grow larger as we append more stuff to it. Endgame would would have an ancestors set *per-fork*.

src/accountsdb/two/Unrooted.zig
Rexicon226: Confused what you mean. The commit you've commented on comments it out, because with the approach taken there this assert could be tripped and it would be valid for `is_empty` to be true. So what would making it a panic have done?

Note that *now*, with the latest version of this branch, the assert is back in place, and rooting is done differently which ensures that this assert *should* never be tripped (although still todo more longer running testing for that).

src/accountsdb/account_store.zig
Rexicon226: Removed: [5425ee7](https://github.com/Syndica/sig/pull/1072/commits/5425ee73a2345a3d913263b46fa21a7c3ccdd19e)

src/accountsdb/two/Two.zig
```diff
@@ -164,64 +173,3 @@ pub fn slotModifiedIterator(self: *Db, slot: Slot) ?SlotModifiedIterator {
...
```
yewman, line 173: Why are we removing these?

src/cmd.zig
yewman: Yeah I think that is a good idea 👍

yewman (commented):
Looking good, just a couple questions

src/replay/service.zig
Sobeston: Why not `>= MAX_SLOTS` instead of `> 512`?

src/accountsdb/two/Two.zig
```diff
@@ -29,10 +28,19 @@ pub fn init(allocator: std.mem.Allocator, rooted: Rooted) !Db {
...
 
-/// Initializes a temporary empty rooted storage. Call tmp.cleanup() when done with it.
-pub fn initTest(allocator: std.mem.Allocator) !struct { Db, std.testing.TmpDir } {
-    if (!builtin.is_test) @compileError("only used in tests");
```
dnut, line 34: is it used outside tests now?

src/replay/resolve_lookup.zig
```diff
@@ -536,47 +551,28 @@ test resolveBatch {
     }
 }
 
-fn put(
```
dnut, line 553: why duplicate this code in multiple places?

src/replay/freeze.zig
```diff
@@ -367,49 +380,118 @@ pub fn deltaMerkleHash(account_reader: AccountReader, allocator: Allocator, slot
...
-    var hash = LtHash.IDENTITY;
-    var i: usize = 0;
-    while (try iterator.next(allocator)) |pubkey_account| : (i += 1) {
-        const pubkey, const account = pubkey_account;
-        defer account.deinit(allocator);
+    const Task = struct {
+        const Result = union(enum) {
+            new: LtHash,
+            existed: struct { LtHash, LtHash }, // old, new
+        };
```
dnut: my main suggestion is in https://github.com/Syndica/sig/pull/1072#discussion_r2641423450, which supersedes this comment.

if you don't go with that suggestion, then i suggest this:
```suggestion
        const Result = struct {
            old: ?LtHash,
            new: LtHash,
        };
```

src/replay/freeze.zig
```diff
@@ -367,49 +380,118 @@ pub fn deltaMerkleHash(account_reader: AccountReader, allocator: Allocator, slot
...
-            defer old_acct.deinit(allocator);
+        fn hashAccounts(
+            gpa: std.mem.Allocator,
+            slot_index: *sig.accounts_db.Two.Unrooted.SlotIndex,
+            db: *sig.accounts_db.Two,
+            ancestors: *const Ancestors,
+            hashes: []std.ArrayListUnmanaged(Result),
+            task: sig.utils.thread.TaskParams,
+        ) !void {
+            const hash = &hashes[task.thread_id];
```
dnut: my main suggestion is in https://github.com/Syndica/sig/pull/1072#discussion_r2641423450, which supersedes this comment.

if you don't go with that suggestion, then why don't you preallocate this arraylist?

src/replay/freeze.zig
```diff
@@ -367,49 +380,118 @@ pub fn deltaMerkleHash(account_reader: AccountReader, allocator: Allocator, slot
...
+            }
+        }
+    };
 
-            if (!old_acct.equals(&account)) {
-                hash.mixOut(old_acct.ltHash(pubkey));
-                hash.mixIn(account.ltHash(pubkey));
+    const NUM_THREADS = 8;
+
+    var hashes: [NUM_THREADS]std.ArrayListUnmanaged(Task.Result) = @splat(.empty);
```
dnut: why don't you sum the hashes in each thread, and then just do a final sum of each thread's hash at the end? no need to allocate and copy all this data.
```suggestion
    var hashes: [NUM_THREADS]LtHash = @splat(.IDENTITY);
```

src/replay/resolve_lookup.zig
Rexicon226: i dont see the benefit of having a put function, it is only used a couple (two) times

dnut: the benefit is that you avoid duplicating identical code which in principle can reduce maintenance costs. i lean towards using a function for this, but perhaps the benefit is negligible, so i'm not concerned about it. choosing to remove the pre-existing function was an opinionated move, so it made me curious

src/accountsdb/two/Two.zig
dnut: @Rexicon226, you mentioned that you addressed this comment, but @yewman said he doesn't see how it's addressed. can you explain please?

src/accountsdb/two/Two.zig
Rexicon226: Yeah it's used in the conformance harness. I've been meaning to make a build options which is only enabled for the test builds, and the conformance harness builds, but disallows calling a function for the regular sig exe build. Would be useful here, but I want to make that change a more general one.

src/replay/service.zig
Rexicon226: No reason, it's using `>= MAX_SLOTS` now. https://github.com/Syndica/sig/pull/1072/changes/053c75543c1d2f58129c0980580f9c6517472a2a

src/accountsdb/two/Two.zig
Rexicon226: Guess I hadn't added the changeset when I pushed before, it's here: 053c75543c1d2f58129c0980580f9c6517472a2a

src/replay/freeze.zig
Rexicon226: A great idea. I've switched over to using this strategy, although in the end we should just move the delta lt-hash computation into the generation execution pool. We're wasting time by not doing it, since on pretty much any testnet slot and especially on mainnet slot, we have enough gaps where we are waiting for transactions to finish.

src/accountsdb/two/Two.zig
```diff
@@ -65,44 +73,42 @@ pub fn put(self: *Db, slot: Slot, address: Pubkey, data: AccountSharedData) !voi
...
+    // Ancestors are kept in a way where the "head" slot is stored.
+    std.debug.assert(ancestors.containsSlot(newly_rooted_slot));
+    // This assert is satifisied via the implementation of ancestors, however
+    // we want to show that we'll never need to iterate over more than MAX_SLOTs cases.
+    std.debug.assert(ancestors.ancestors.count() < Unrooted.MAX_SLOTS);
+
+    // TODO: We can probably avoid running through the full range of slots
+    // by computing something like:
+    // range = new_slot - old_slot;
+    // for (0..range) |i| (old_slot + i) % MAX_SLOTS;
```
dnut: You may want to implement this now to address a bug in this function.

Let's say we modified an account in every slot from 0-4096, and there are no forks. We've rooted up until slot 4094. The modification from slot 4095 will be at the end of the unrooted buffer, and the modification from 4096 will wrap around and be at the beginning. 

onSlotRooted is not called for slot 4095. When onSlotRooted is called for slot 4096, this loop is going to first persist the version from 4096 in the rooted store because it's at position 0, and then when it reaches the end of the unrooted slots, it will overwrite that account with the version from 4095.

Here's a test that demonstrates the issue:

```zig
test "rooting must handle wraparound and non-consecutive roots" {

    var test_state = try Db.initTest(allocator);
    defer test_state.deinit();
    const db = &test_state.db;

    var ancestors = Ancestors.EMPTY;
    defer ancestors.deinit(allocator);

    for (0..Unrooted.MAX_SLOTS * 2) |slot| {
        try ancestors.addSlot(allocator, slot);
        ancestors.cleanup();

        try db.put(slot, .ZEROES, .{
            .lamports = slot * 10,
            .data = &.{},
            .owner = .ZEROES,
            .executable = false,
            .rent_epoch = 0,
        });

        {
            const account = try db.get(allocator, .ZEROES, &ancestors);
            defer account.?.deinit(allocator);
            try std.testing.expectEqual(slot * 10, account.?.lamports);
        }

        if (slot % 2 == 0) {
            db.onSlotRooted(slot, &ancestors);
            const account = try db.get(allocator, .ZEROES, &ancestors);
            defer account.?.deinit(allocator);
            try std.testing.expectEqual(slot * 10, account.?.lamports);
        }
    }
}
```

I see two potential solutions:
1. Change the loop to persist the slots in order.
2. Change rooted put so it doesn't replace versions from newer slots with versions from older slots.

Rexicon226: A fair point, I understand the issue you're showing here. I've added the optimization/fix here:
88f503e294ea48ae8b98c542d8be6c5c6f7ce880
along with the test you described (which passes now!)

Issue #1051, opened 2025-11-02
Suboptimal transaction scheduling?


Sobeston:
### Description

Most of the time, are main replay thread can near-fully utilise its threadpool with replayBatch like this:

<img width="1850" height="989" alt="Image" src="https://github.com/user-attachments/assets/680dbe97-cc10-4f97-b972-6b81a7391ea8" />

Sometimes we get blocks which are scheduled in a very sparse way, like this: 

<img width="1839" height="1111" alt="Image" src="https://github.com/user-attachments/assets/b8d36e8e-c6f5-4731-a543-c4e1d4ee922b" />

This represents the majority of our slowest replayed slots. In the profile captured on my computer, our mean block time is ~200ms, but over the 3000 slots I executed about 5 of them took 700-900ms, all of which look like this.

It seems unlikely to me that this is (correctly) caused by account locking

### How to Reproduce the Bug

- make a release build with tracy enabled
- run the validator a populated accountsdb and ledger
- query tracy for the advanceReplay calls with the highest times and look closely

PR #1074, opened 2025-11-10
perf(replay): redo `replaySlotAsync`

kprotty:
* Removes the old `poll()` + `awaitBlocking` architecture from `ReplaySlotFuture`
* `replayActiveSlotsAsync` now queues pool tasks in batches & waits on a single `WaitGroup` at the end
* general cleanup of `PoHVerifier` & `TransactionScheduler` scheduling code

Additionally, this moves to graph-based, per-transaction scheduling instead of per-Entry batch scheduling. Allowing the most amount of parallelism available into `replayBatch`. But as a side-effect, it means the currently unnecessary overheads inside `replayBatch` are more exaggerated as it's no longer amortized over a batch of transactions (TODO: address this).


Sobeston:
Do you think this resolves https://github.com/Syndica/sig/issues/1051?

src/replay/execution.zig
```diff
@@ -208,47 +197,24 @@ pub fn replaySlotAsync(
...
-    };
-    errdefer future.destroy(allocator);
-
-    if (verifyTicks(logger, params.entries, params.verify_ticks_params)) |block_error| {
-        future.status = .{ .done = .{ .invalid_block = block_error } };
-        return future;
-    }
-
-    try replay.exec_async.startPohVerify(
+    ReplaySlotFuture.startAsync(
```
dnut: replaySlotAsync seems redundant with ReplaySlotFuture.startAsync

dnut (commented):
let's do some slot hash testing against agave before merging this

src/replay/exec_async.zig
```diff
@@ -11,916 +8,426 @@ const core = sig.core;
...
+                self.future.logger.err().log("poh verification failed");
+                self.future.setError(.{ .invalid_block = .InvalidEntryHash });
+            }
         }
-        try self.results.send(.{ .batch_index = self.batch_index, .result = result });
-    }
+    };
 };
 
-test "TransactionScheduler: happy path" {
```
dnut, line 440: Why did you remove all these tests? Can't they be adapted to work with the new approach?

dnut (commented):
.

src/replay/exec_async.zig
```diff
@@ -11,916 +8,426 @@ const core = sig.core;
...
+    svm_gateway: SvmGateway,
+    committer: Committer,
+    workers: std.ArrayListUnmanaged(Worker) = .{},
+
+    // zig fmt: off
+    const Error =
+        Allocator.Error ||
+        error { ConflictingBatchAccountLock } ||
+        @typeInfo(@typeInfo(@TypeOf(resolveTransaction)).@"fn".return_type.?).error_union.error_set ||
+        @typeInfo(@typeInfo(@TypeOf(replayBatch)).@"fn".return_type.?).error_union.error_set;
```
InKryption: Do these values return any error values that have to be checked for to conduct specific control flow? If not, I would change this to:
```zig
    const Error = Allocator.Error || error {
        ConflictingBatchAccountLock,
        ResolveTransactionFail,
        ReplayBatchFail,
    };
```
And then `catch return error.` at each of the corresponding function calls.

src/replay/execution.zig
```diff
@@ -1178,6 +1131,44 @@ pub const TestState = struct {
...
+        const entries_copy = try allocator.dupe(Entry, entries);
+        errdefer allocator.free(entries_copy);
+
+        var i: usize = 0;
+        errdefer for (entries_copy[0..i]) |e| e.deinit(allocator);
+        for (entries_copy) |*entry| {
+            entry.* = try entries[i].clone(allocator);
+            i += 1;
+            try self.makeTransactionsPassable(allocator, entry.transactions);
+        }
```
```suggestion
        for (entries_copy, 0..) |*entry, i| {
            errdefer for (entries_copy[0..i]) |e| e.deinit(allocator);
            entry.* = try entry.clone(allocator);
            errdefer entry.deinit(allocator);
            try self.makeTransactionsPassable(allocator, entry.transactions);
        }
```

InKryption (dismissed):
Just a couple of fairly minor suggestions, otherwise LGTM

src/replay/exec_async.zig
kprotty: the error is what's reported up to `replayActiveSlots`. The specific error is already being logged, so yea I can generalize it like that.

kprotty: https://github.com/Syndica/sig/pull/1074/commits/5bb14ee96599004931c86030b744bf597f229294

src/replay/execution.zig
kprotty: https://github.com/Syndica/sig/pull/1074/commits/5fd527a684393c147f6297f5af31b2e79afdd32e

src/replay/exec_async.zig
kprotty: https://github.com/Syndica/sig/pull/1074/commits/281e2a60a3b8f93248950ccfdeba93d740eaf058

src/replay/execution.zig
kprotty: https://github.com/Syndica/sig/pull/1074/commits/1a82de688865dc67665ae980437bc6ec77f7044a

PR #1083, opened 2025-11-10
feat(db): hook up accountsdb to cmd, stop creating old index

Sobeston:


Rexicon226:
moved commit into base branch

PR #1085, opened 2025-11-12
feat(conformance): script to vendor agave in solfuzz_agave for debugging

dnut:
solfuzz_agave normally depends on agave through git links in Cargo.toml, but it can be useful to debug agave with a locally cloned copy. This is a huge pain to set up manually because you need to go through the Cargo.toml and edit a bunch of dependencies to point to all different non-obvious paths on your filesystem. This script `vendor-agave.py` I added here automates that process.

### DIFF
--- a/conformance/README.md
+++ b/conformance/README.md
@@ -20,4 +20,7 @@ zig build -Doptimize=ReleaseSafe solfuzz_sig
 
 # for more options
 ./run.py --help
+
+# optional: edit and debug agave's code:
+scripts/vendor-agave.py
 ```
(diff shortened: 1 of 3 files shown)


conformance/README.md
```diff
@@ -20,4 +20,7 @@ zig build -Doptimize=ReleaseSafe solfuzz_sig
 
 # for more options
 ./run.py --help
+
+# optional: edit and debug agave's code:
+scripts/vendor-agave.py
```
kprotty, line 25: solfuzz-agave includes a script to patch Cargo links to a local or specific agave commit: https://github.com/firedancer-io/solfuzz-agave/tree/agave-v3.0.3?tab=readme-ov-file#note-for-local-copies

what about making a bash script that (clones agave &) calls into this instead

dnut: interesting, wish i knew about that earlier.

PR #1088, opened 2025-11-12
fix(consensus, voting): Various fixes to vote handling

dadepo:
A couple of fixes and improvements:

- Added ability to pass the vote account
- Fixed parsing of the vote state from the vote account
- Fix fallback logic when pushing votes out and no upcoming_leader_sockets
- Add some logging when pushing votes out to the network


src/consensus/tower.zig
```diff
@@ -303,19 +302,18 @@ pub fn lastVotedSlotInBank(
...
-    vote_account_pubkey: *const Pubkey,
 ) (error{BincodeError} || std.mem.Allocator.Error)!VoteState {
-    const buf = try allocator.alloc(u8, vote_account.data.len());
-    // TODO Not sure if this is the way to get the data from the vote account. Review.
-    _ = vote_account.serialize(vote_account_pubkey, buf);
+    const data_len = vote_account.data.len();
+    const buf = try allocator.alloc(u8, data_len);
+    defer allocator.free(buf);
+    vote_account.data.readAll(buf);
     const versioned_state = sig.bincode.readFromSlice(
```
InKryption: How about we just use `bincode.read` with `vote_account.iterator().reader()`? Don't see a reason to allocate here.

src/cmd.zig
```diff
@@ -1382,25 +1438,22 @@ fn replayOffline(
...
     var replay_service_state: ReplayAndConsensusServiceState = try .init(allocator, .{
         .app_base = &app_base,
         .loaded_snapshot = &loaded_snapshot,
         .ledger = &ledger,
         .epoch_context_manager = &epoch_context_manager,
         .replay_threads = cfg.replay_threads,
         .disable_consensus = cfg.disable_consensus,
-        .voting_enabled = cfg.voting_enabled,
+        .voting_enabled = false,
+        .vote_account_pubkey = Pubkey.ZEROES,
```
InKryption: Do you think it would make sense to remove `voting_enabled`, make `vote_account_pubkey` an optional, and instead declare voting to be enabled on the basis of whether the vote account is null or not?

src/cmd.zig
```diff
@@ -2049,6 +2102,7 @@ const ReplayAndConsensusServiceState = struct {
             replay_threads: u32,
             disable_consensus: bool,
             voting_enabled: bool,
+            vote_account_pubkey: ?Pubkey,
```
InKryption: the _pubkey suffix seems somewhat redundant in context, no?

src/cmd.zig
```diff
@@ -1015,6 +1027,25 @@ fn identity(allocator: std.mem.Allocator, cfg: config.Cmd) !void {
...
+    if (end_pos != buf.len) return error.InvalidKeypairFile;
+
+    const file_len = try file.readAll(&buf);
+    if (file_len != buf.len) return error.InvalidKeypairFile;
+
+    const secret_key = try std.crypto.sign.Ed25519.SecretKey.fromBytes(buf);
+    const keypair = try std.crypto.sign.Ed25519.KeyPair.fromSecretKey(secret_key);
+    return Pubkey.fromPublicKey(&keypair.public_key);
+}
+
```
InKryption: I think a better place to put this would be in `src/identity.zig`.

src/consensus/tower.zig
dadepo: [0869be0](https://github.com/Syndica/sig/pull/1088/commits/0869be006ae7d4c7692ad8ea78affe266c3631e8)

src/cmd.zig
dadepo: I think it does make sense. I considered it but did not want to include it in this change set. I think it is a refactor that can be done on it own.

src/cmd.zig
dadepo: No strong feelings, but there exist the usage of this suffix across the codebase

src/cmd.zig
dadepo: [8fe7ba9](https://github.com/Syndica/sig/pull/1088/commits/8fe7ba99e2df1c3b636585444859ffcade57ac4a)

InKryption (dismissed):

src/cmd.zig
dnut: I don't think we should be adding code if we don't have a reason for that code to exist. I don't think the existence of a pattern is a reason to propagate the pattern--what if it's a bad pattern?

In general, I think we overuse the word "pubkey" in sig, where we should be using the word "address", nothing, or something more specific.

I would change this to `vote_account` or `vote_account_address`. Personally, I lean toward the concise name but either one is OK.

src/cmd.zig
```diff
@@ -1214,7 +1226,31 @@ fn validator(
...
+            defer allocator.free(app_data_dir_path);
+            break :default_path try std.fs.path.join(
+                allocator,
+                &.{ app_data_dir_path, "vote-account.key" },
+            );
+        };
+        defer if (cfg.vote_account_path == null) allocator.free(vote_keypair_path);
+
+        break :blk sig.identity.readBinaryKeypairPubkey(vote_keypair_path) catch |err| {
+            app_base.logger.warn().logf(
```
dnut: i think error makes more sense instead of warn here.

src/consensus/tower.zig
```diff
@@ -89,28 +92,48 @@ pub const Tower = struct {
...
-            ) catch |err| switch (err) {
-                error.OutOfMemory => |e| return e,
-                error.InvalidOffset,
-                error.FileIdNotFound,
-                error.SlotNotFound,
-                => null,
-            };
+            );
             break :blk maybe_vote_account orelse {
+                self.logger.warn().logf(
```
dnut: isn't this a normal situation for a new validator? why would it be a warning message?

src/replay/consensus/core.zig
```diff
@@ -1261,35 +1263,63 @@ fn generateVoteTx(
...
 
-    const node_kp = node_keypair orelse return .non_voting;
+    const node_kp = node_keypair orelse {
+        logger.debug().log("No node keypair");
+        return .non_voting;
+    };
 
-    const last_voted_slot = replay_tower.lastVotedSlot() orelse return .failed;
+    const last_voted_slot = replay_tower.lastVotedSlot() orelse {
+        logger.warn().log("No last voted slot");
```
dnut: info? seems expected on startup

src/replay/consensus/core.zig
```diff
@@ -1261,35 +1263,63 @@ fn generateVoteTx(
...
+        return .failed;
+    };
 
-    const slot_info = slot_tracker.get(last_voted_slot) orelse return .failed;
+    const slot_info = slot_tracker.get(last_voted_slot) orelse {
+        logger.warn().logf(
```
dnut, line 1283: does this imply corrupted state? maybe should be error?

src/replay/consensus/core.zig
```diff
@@ -1261,35 +1263,63 @@ fn generateVoteTx(
...
+            "Slot info not found in slot_tracker for last_voted_slot {}",
+            .{last_voted_slot},
+        );
+        return .failed;
+    };
 
     const vote_account_result = account_reader.forSlot(&slot_info.constants.ancestors)
-        .get(allocator, vote_account_pubkey) catch return .failed;
+        .get(allocator, vote_account_pubkey) catch |err| {
+        logger.warn().logf("Failed to read vote account: {}", .{err});
```
dnut: accountsdb errors should probably be propagated, not caught.

src/replay/consensus/core.zig
```diff
@@ -1261,35 +1263,63 @@ fn generateVoteTx(
...
+        return .failed;
+    };
+
     const vote_account = vote_account_result orelse {
+        logger.warn().log("Vote account not found");
```
dnut: does this imply corrupted state? maybe should be error?

src/replay/consensus/core.zig
```diff
@@ -1261,35 +1263,63 @@ fn generateVoteTx(
...
         return .failed;
     };
     defer vote_account.deinit(allocator);
 
-    const vote_account_data = vote_account.data.readAllAllocate(allocator) catch return .failed;
+    const vote_account_data = vote_account.data.readAllAllocate(allocator) catch |err| {
+        logger.warn().logf("Failed to read vote account data: {}", .{err});
```
dnut: does this imply corrupted state? maybe should be error?

src/replay/consensus/core.zig
```diff
@@ -1261,35 +1263,63 @@ fn generateVoteTx(
...
     defer allocator.free(vote_account_data);
 
     var vote_state_versions = sig.bincode.readFromSlice(
         allocator,
         sig.runtime.program.vote.state.VoteStateVersions,
         vote_account_data,
         .{},
-    ) catch return .failed;
+    ) catch |err| {
+        logger.warn().logf("Failed to deserialize vote state versions: {}", .{err});
```
dnut: does this imply corrupted state? maybe should be error?

src/replay/consensus/core.zig
```diff
@@ -1261,35 +1263,63 @@ fn generateVoteTx(
...
+        return .failed;
+    };
     defer vote_state_versions.deinit(allocator);
 
-    var vote_state = vote_state_versions.convertToCurrent(allocator) catch return .failed;
+    var vote_state = vote_state_versions.convertToCurrent(allocator) catch |err| {
+        logger.warn().logf("Failed to convert vote state to current: {}", .{err});
```
dnut: does this imply corrupted state? maybe should be error?

src/replay/consensus/core.zig
```diff
@@ -1300,6 +1330,7 @@ fn generateVoteTx(
     const current_epoch = epoch_tracker.schedule.getEpoch(last_voted_slot);
 
     const authorized_voter_pubkey = vote_state.voters.getAuthorizedVoter(current_epoch) orelse {
+        logger.warn().logf("No authorized voter for epoch {}", .{current_epoch});
```
dnut: does this imply corrupted state? maybe should be error?

src/cmd.zig
dadepo: Went with `vote_account_address` [f032fa7](https://github.com/Syndica/sig/pull/1088/commits/f032fa7870ff1d196182670deb3f4fe9caadf1d0)

src/replay/consensus/core.zig
dadepo: [fd45281](https://github.com/Syndica/sig/pull/1088/commits/fd452811a85c8959243f0ebf03eb9180734cadf1)

src/replay/consensus/core.zig
dadepo: [fd45281](https://github.com/Syndica/sig/pull/1088/commits/fd452811a85c8959243f0ebf03eb9180734cadf1)

src/replay/consensus/core.zig
dadepo: [fd45281](https://github.com/Syndica/sig/pull/1088/commits/fd452811a85c8959243f0ebf03eb9180734cadf1)

src/replay/consensus/core.zig
dadepo: [fd45281](https://github.com/Syndica/sig/pull/1088/commits/fd452811a85c8959243f0ebf03eb9180734cadf1)

src/replay/consensus/core.zig
dadepo: [fd45281](https://github.com/Syndica/sig/pull/1088/commits/fd452811a85c8959243f0ebf03eb9180734cadf1)

src/replay/consensus/core.zig
dadepo: [fd45281](https://github.com/Syndica/sig/pull/1088/commits/fd452811a85c8959243f0ebf03eb9180734cadf1)

src/replay/consensus/core.zig
dadepo: [fd45281](https://github.com/Syndica/sig/pull/1088/commits/fd452811a85c8959243f0ebf03eb9180734cadf1)

src/replay/consensus/core.zig
dadepo: [fd45281](https://github.com/Syndica/sig/pull/1088/commits/fd452811a85c8959243f0ebf03eb9180734cadf1)

src/consensus/tower.zig
dadepo: My understanding is that this should not happen except something went wrong. The vote account should be already setup and on chain, and the address/pubkey passed to the voting validator. So if it is not found it should be a sign of something wrong ie a misconfiguration.

When adding this log, I debated even propagating the error instead of falling back `self.initializeRoot(fork_root);` but left it as as I think that is a separate change.

dadepo: > isn't this a normal situation for a new validator?

Thinking about it again, this may indeed be the scenario if the validator starts with an older state that does not have a record of the vote account. Switched to info [cb8441c](https://github.com/Syndica/sig/pull/1088/commits/cb8441c014ed588540291a4ca18ea6c0b97bc158)

PR #1091, opened 2025-11-13
fix(replay, consensus): Fix unbounded channel issues

dadepo:
- Use same channel to send replay votes from replay to consensus
- Comment out sending data to `sig.sync.Channel(VerifiedVote)` as it is not used yet in repair service.

### DIFF
--- a/src/replay/Committer.zig
+++ b/src/replay/Committer.zig
@@ -89,7 +89,6 @@ pub fn commitTransactions(
                     }
                 }
             }
-
         }
 
         const recent_blockhash = &transaction.transaction.msg.recent_blockhash;


kprotty (approved):

kprotty (approved):

src/replay/consensus/core.zig
```diff
@@ -294,16 +294,20 @@ pub const TowerConsensus = struct {
         /// Sent by WindowService and DuplicateShred handlers
         duplicate_slots: *Channel(Slot),
         /// Sent by `replayActiveSlots`, received by the vote collector.
-        replay_votes: *Channel(ParsedVote),
+        /// Optional - null when consensus is disabled.
+        replay_votes: ?*Channel(ParsedVote),
```
dnut: If consensus is disabled, why does this struct exist at all?

dnut (approved):

src/replay/service.zig
```diff
@@ -203,8 +210,11 @@ pub const ReplayState = struct {
             deps.allocator.destroy(slot_tracker.slots.fetchSwapRemove(deps.root.slot).?.value);
         }
 
-        const replay_votes_channel = try Channel(ParsedVote).create(deps.allocator);
-        errdefer replay_votes_channel.destroy();
+        const replay_votes_channel: ?*Channel(ParsedVote) = if (consensus_status == .enabled)
+            try Channel(ParsedVote).create(deps.allocator)
+        else
+            null;
```
kprotty, line 216: could be a `switch`, but it's fine

PR #1094, opened 2025-11-13
fix(consensus): notify `AccountStore` when rooting slot.

kprotty:
When consensus is disabled, the `bypassConsensus` function calls `AccountStore.onSlotRooted` after updating the slot_stracker which frees up `commitTransaction`-inserted unrooted accounts. However when consensus is enabled, it forgot to call this; resulting in unrooted accounts leaking memory.

### DIFF
--- a/src/replay/consensus/core.zig
+++ b/src/replay/consensus/core.zig
@@ -2410,6 +2410,7 @@ test "checkAndHandleNewRoot - missing slot" {
         &slot_tracker,
         &fixture.progress,
         &fixture.fork_choice,
+        .noop,
         123, // Non-existent slot
     );
 
@@ -2466,6 +2467,7 @@ test "checkAndHandleNewRoot - missing hash" {
         slot_tracker2_ptr,
         &fixture.progress,
         &fixture.fork_choice,
+        .noop,
         root.slot, // Non-existent hash
     );
 
@@ -2507,6 +2509,7 @@ test "checkAndHandleNewRoot - empty slot tracker" {
         slot_tracker3_ptr,
         &fixture.progress,
         &fixture.fork_choice,
+        .noop,
         root.slot,
     );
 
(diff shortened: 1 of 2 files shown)


src/replay/consensus/core.zig
```diff
@@ -1479,6 +1482,12 @@ fn checkAndHandleNewRoot(
     slot_tracker.root = new_root;
     // Prune non rooted slots
     slot_tracker.pruneNonRooted(allocator);
+    // Tell the account_store about it for its unrooted accounts
+    try account_store.onSlotRooted(
```
Rexicon226, line 1486: Noobie question, can you call this before `pruneNonRooted`? I've seen the `pruneNonRooted` call take quite a while, and our plan is to make `onSlotRooted` async, since there is no reason to wait for it to finish right here.

kprotty: `pruneNonRooted` is rather simple: scans map of (usually ~30-70) tracked slot entries. If key < new_root, remove & deinit entry. Wonder if the cost is in the deinit() process..

`pruneNonRooted` doesnt access the account_store, so `onSlotRooted` can be called before. But I assume the concurrency question also includes if anything relies on onSlotRooted finishing right there; `generateVoteTx` is called after this & reads `last_vote_slot`, but that may not overlap with unrooted slots flushed during onSlotRooted.

Rexicon226: The way we designed the new db is that onSlotRooted can finish whenever, nothing "relies on it finishing", since we first move to rooted and then remove from unrooted. There is plenty of buffer space in unrooted to have some extra slots in it. 

I guess that probably isn't the case with the old db, so doesn't apply to this PR (since it's still on it), but it is an idea to consider later.

dnut: I don't see what difference it makes by reordering these calls. Seems fine either way.

PR #1096, opened 2025-11-17
fix(consensus): switch fork

dadepo:
This PR contains a couple of fixes:

- Fixed lockout intervals iteration
- Fixed ancestor/descendants double-counting
- Fixed threshold check: In checkVoteStakeThreshold, changed to return fork_stake instead of 0 for observed_stake when threshold fails
- Fixed isRecent logic
- Fixed frozen slots processing order: Added sorting of frozen slots by slot number before processing
- Fixed lastVoteAbleToLand
- Misc: logging, refactors etc

### DIFF
--- a/src/consensus/replay_tower.zig
+++ b/src/consensus/replay_tower.zig
@@ -4566,7 +4566,7 @@ test "switch threshold" {
         }
 
         // Set root to 43
-        tower.tower.vote_state.root_slot = 43;
+        tower.tower.root = 43;
 
         const decision = try tower.makeCheckSwitchThresholdDecision(
             allocator,
@@ -4863,7 +4863,7 @@ test "switch threshold use gossip votes" {
     // then the switch proof will now fail since that validator's vote can no longer be included
     {
         // Set root to 44 - this makes slot 112 (descendant of 43) unreachable
-        tower.tower.vote_state.root_slot = 44;
+        tower.tower.root = 44;
 
         // Update ancestors and descendants to reflect new root
         // Rebuild ancestors from root 44
--- a/src/consensus/tower.zig
+++ b/src/consensus/tower.zig
@@ -285,11 +285,8 @@ pub const Tower = struct {
 };
 
 test "isRecent with no votes checks root_slot" {
-    var tower: Tower = .init(.noop);
-
     // Set up tower with no votes but with a root
-    tower.vote_state.root_slot = 100;
-    tower.vote_state.votes = .{};
+    var tower: Tower = .{ .root = 100 };
 
     // Slots at or before root should NOT be recent
     try std.testing.expect(!tower.isRecent(99));
@@ -303,9 +300,7 @@ test "isRecent with no votes checks root_slot" {
 test "isLockedOut with no votes and root" {
     const allocator = std.testing.allocator;
 
-    var tower: Tower = .init(.noop);
-    tower.vote_state.root_slot = 100;
-    tower.vote_state.votes = .{};
+    var tower: Tower = .{ .root = 100 };
 
     // Create ancestors for a slot after root
     var ancestors_map: std.AutoArrayHashMapUnmanaged(Slot, void) = .empty;


src/replay/consensus/core.zig
```diff
@@ -1615,9 +1619,17 @@ fn computeConsensusInputs(
...
-    // If not, then we can avoid sorting here which may be verbose given frozen_slots is a map.
+
+    var sorted_frozen_slots =
+        try std.ArrayListUnmanaged(Slot).initCapacity(allocator, frozen_slots.count());
+    defer sorted_frozen_slots.deinit(allocator);
+
     for (frozen_slots.keys()) |slot| {
+        sorted_frozen_slots.appendAssumeCapacity(slot);
+    }
+    std.mem.sort(Slot, sorted_frozen_slots.items, {}, std.sort.asc(Slot));
```
dnut: why are you copying the list before sorting? can't you just sort the original? we even have FrozenSlotsSortCtx aready which makes this very easy. let's try to avoid unnecessary allocations and copies when possible.

src/replay/consensus/core.zig
```diff
@@ -418,9 +418,13 @@ pub const TowerConsensus = struct {
             for (params.slot_tracker.slots.keys(), params.slot_tracker.slots.values()) |slot, info| {
                 const slot_ancestors = &info.constants.ancestors.ancestors;
                 const ancestor_gop = try ancestors.getOrPutValue(arena, slot, .EMPTY);
+                // Ensure every slot has a descendants entry (even if empty)
+                _ = try descendants.getOrPutValue(arena, slot, .empty);
                 try ancestor_gop.value_ptr.ancestors
                     .ensureUnusedCapacity(arena, slot_ancestors.count());
                 for (slot_ancestors.keys()) |ancestor_slot| {
+                    // Exclude the slot itself from ancestors.
+                    if (ancestor_slot == slot) continue;
```
dnut, line 439: could you please link to the relevant area of agave? i'm curious how they do this, since it's different from the ancestors they store in the bank

src/consensus/replay_tower.zig
```diff
@@ -601,6 +603,14 @@ pub const ReplayTower = struct {
                 // stray slots (which should always be empty_ancestors).
                 //
                 // This is covered by test_future_tower_* in local_cluster
+                self
+                    .logger
+                    .info()
+                    .logf(
+                    "failed_switch_threshold. switch_proof_stake: {}, total_stake: {}",
+                    .{ 0, total_stake },
```
dnut, line 579: this should at least be warn since it represents a bad state. if it requires action, then it should be error

src/consensus/replay_tower.zig
```diff
@@ -620,56 +630,44 @@ pub const ReplayTower = struct {
...
-            const is_candidate_eq_last_voted_slot = if (!is_descendant_computed)
-                (candidate_slot == last_voted_slot)
-            else
-                is_descendant_computed;
+            const is_descendant_computed = for (candidate_descendants.items()) |d| {
+                if (progress.getForkStats(d)) |stats| {
+                    if (stats.computed) break true;
+                }
+            } else false;
+            if (is_descendant_computed) continue;
```
dnut, line 616: more concise version:
```suggestion
            for (candidate_descendants.items()) |d| {
                if (progress.getForkStats(d)) |stats| {
                    if (stats.computed) break;
                }
            } else continue;
```

src/consensus/replay_tower.zig
```diff
@@ -620,56 +630,44 @@ pub const ReplayTower = struct {
...
-            // 3) Don't consider lockouts on the `last_vote` itself
-            // 4) Don't consider lockouts on any descendants of
-            //    `last_vote`
-            // 5) Don't consider any banks before the root because
-            //    all lockouts must be ancestors of `last_vote`
             const is_progress_computed = if (progress.getForkStats(candidate_slot)) |stats|
                 stats.computed
             else
                 false;
+            if (!is_progress_computed) continue;
```
dnut: more concise version:
```suggestion
            const stats = progress.getForkStats(candidate_slot) orelse continue;
            if (!stats.computed) continue;
```

src/consensus/replay_tower.zig
```diff
@@ -1191,8 +1202,9 @@ pub const ReplayTower = struct {
         (my_latest_landed_vote_slot >= last_voted_slot) or
             // 2. Already voting at the tip
             (last_voted_slot >= heaviest_bank_on_same_voted_fork) or
-            // 3. Last vote is within slot hashes, regular refresh is enough
-            slot_history.check(last_voted_slot) == .found;
+            // 3. Last vote is within this bank's slot hashes history, regular refresh is enough
+            (last_voted_slot < heaviest_bank_on_same_voted_fork and
+                slot_history.check(last_voted_slot) == .found);
```
dnut: this is logically equivalent to the original, it's just more computation (not sure if the compiler will optimize it).

the check you added is implied by the prior being false

src/replay/consensus/core.zig
dadepo: The analogous part in Agave where the ancestors are retreived

https://github.com/anza-xyz/agave/blob/dfe2a34904f072df75c1adbcb1c037c3c6084880/core/src/replay_stage.rs#L778-L780

it uses `.ancestors()` method on BankForks which is here https://github.com/anza-xyz/agave/blob/dfe2a34904f072df75c1adbcb1c037c3c6084880/runtime/src/bank_forks.rs#L182-L191 

and its implementation depends on `proper_ancestors` on Bank here https://github.com/anza-xyz/agave/blob/dfe2a34904f072df75c1adbcb1c037c3c6084880/runtime/src/bank.rs#L1724-L1730 and that filters out the slot itself from the ancestors set.

src/consensus/replay_tower.zig
```diff
@@ -689,25 +687,35 @@ pub const ReplayTower = struct {
...
+                lockout_intervals.map.values(),
+            ) |lockout_interval_end, *intervals_keyed_by_end| {
+                // Only consider intervals that expire at or after last_voted_slot
+                if (lockout_interval_end < last_voted_slot) {
+                    continue;
+                }
+
+                for (intervals_keyed_by_end.items) |vote_account| {
+                    const voted_slot = vote_account[0];
+                    const voted_pubkey = vote_account[1];
```
```suggestion
                    const voted_slot, const voted_pubkey = vote_account;
```

src/replay/consensus/core.zig
dadepo: [aebe9d0](https://github.com/Syndica/sig/pull/1096/commits/aebe9d093ef5a1da742d920d75bb650e75007265)

src/consensus/replay_tower.zig
dadepo: [5442cc4](https://github.com/Syndica/sig/pull/1096/commits/5442cc4bad72b8db246ecc227491dfebbad59a50)

src/consensus/replay_tower.zig
dadepo: [5442cc4](https://github.com/Syndica/sig/pull/1096/commits/5442cc4bad72b8db246ecc227491dfebbad59a50)

src/consensus/replay_tower.zig
dadepo: The suggestion does not work as this only breaks from in the inner loop and then continues the outer loop

src/consensus/replay_tower.zig
dadepo: 5442cc4

src/consensus/replay_tower.zig
dnut: Oh I had the condition reversed. The same basic idea applies but the outer loop needs a label.

```suggestion
            for (candidate_descendants.items()) |d| {
                if (progress.getForkStats(d)) |stats| {
                    if (stats.computed) continue :outer;
                }
            }
```

It's not that important. I'll leave the suggestion open for you to decide how to resolve it.

dnut (approved):

InKryption (approved):

PR #1103, opened 2025-11-20
feat(time): monotonic clock

dnut:
this was a misguided change

### DIFF
--- a/src/time/lib.zig
+++ b/src/time/lib.zig
@@ -1,5 +1,3 @@
-const std = @import("std");
-
 pub const estimate = @import("estimate.zig");
 pub const stake_weighted_timestamp = @import("stake_weighted_timestamp.zig");
 pub const time = @import("time.zig");


src/time/time.zig
```diff
@@ -25,6 +25,37 @@ const std = @import("std");
...
+/// monotonic timestamp analogous to std.time.microTimestamp
+pub fn microTimestamp() i64 {
+    return @as(i64, @intCast(@divFloor(nanoTimestamp(), std.time.ns_per_us)));
+}
+
+/// global state to ensure nanoTimestamp() is monotonic
+var last_clock: std.atomic.Value(i64) = .init(0);
+
+fn nanoTimestamp() i64 {
+    const current: i64 = @intCast(std.time.nanoTimestamp());
```
kprotty, line 47: nanoTimestamp() uses `CLOCK_REALTIME` instead of `CLOCK_MONOTONIC`, which sometimes only has micro-second precision. Does this matter anywhere?

PR #1107, opened 2025-11-22
fix(consensus): handle unspecified vote account

dnut:
Vote accounts are optional on the CLI. If unspecified, the vote account is set to Pubkey.ZEROES which is the system program's address. This causes an error initializing the Tower because system program is owned by the native loader, which is the wrong owner -- it expects the vote account to be owned by the vote program.

```
time=2025-11-21T23:57:09.858Z level=info scope=service_manager message="starting Shred Receiver"
time=2025-11-21T23:57:09.858Z level=info scope=service_manager message="starting Repair Service"
time=2025-11-21T23:57:09.992Z level=info scope=consensus.tower.Tower message="initializeLockoutsFromBank: fork_root=371633688, vote_pubkey=11111111111111111111111111111111"
time=2025-11-21T23:57:09.992Z level=error scope=consensus.tower.Tower message="Invalid vote account owner. Expected: Vote111111111111111111111111111111111111111, Got: NativeLoader1111111111111111111111111111111"
```

This PR fixes the problem by expressing the optionality of the vote account through the type system as a `?Pubkey` set to `null` when unspecified, and then properly handling the null value by skipping account lookup and voting.

I also refactored Tower and its initialization process to simplify the code. I consolidated Tower with TowerVoteState and moved the logic that loads the tower from accountsdb into a single function loadTower.

### DIFF
--- a/src/consensus/replay_tower.zig
+++ b/src/consensus/replay_tower.zig
@@ -1108,11 +1108,11 @@ pub const ReplayTower = struct {
             i += 1;
         }
 
-        var retained = std.BoundedArray(Lockout, MAX_LOCKOUT_HISTORY){};
-        for (self.tower.votes.constSlice(), 0..) |vote, index| {
-            if (flags.isSet(index)) retained.append(vote) catch unreachable;
+        const stale_votes = self.tower.votes.constSlice();
+        self.tower.votes.clear();
+        for (stale_votes, 0..) |vote, index| {
+            if (flags.isSet(index)) self.tower.votes.appendAssumeCapacity(vote);
         }
-        self.tower.votes = retained;
 
         if (self.tower.votes.len == 0) {
             // we might not have banks for those votes so just reset.
--- a/src/consensus/tower.zig
+++ b/src/consensus/tower.zig
@@ -86,9 +86,9 @@ pub const Tower = struct {
 
         // this slice is overwritten in place, but the loop capture stays ahead of those writes
         const stale_votes = self.votes.constSlice();
-        self.votes.len = 0;
+        self.votes.clear();
         for (stale_votes) |vote| {
-            if (vote.slot > new_root) self.votes.append(vote) catch unreachable;
+            if (vote.slot > new_root) self.votes.appendAssumeCapacity(vote);
         }
     }
 


kprotty (approved):

src/consensus/tower.zig
```diff
@@ -88,18 +88,18 @@ pub const Tower = struct {
...
         self.logger.info().logf(
-            "initializeLockoutsFromBank: fork_root={}, vote_pubkey={}",
+            "initializeLockoutsFromBank: fork_root={}, vote_pubkey={?}",
             .{ fork_root, vote_account_pubkey },
         );
         const vote_account = blk: {
             const maybe_vote_account = try slot_account_reader.get(
                 allocator,
-                vote_account_pubkey.*,
+                vote_account_pubkey orelse return,
```
Rexicon226: We should be checking that the pubkey exists before calling the function, there isn't much point in it being optional here. If you specifically want that log message, it should be raised up higher and renamed to something else.

dnut: I agree, and actually I think all this initialization code is confusing. I made some more simplifications.

3e7350f2d5cc7b55a224c09f67175a45ae505e2d

src/consensus/replay_tower.zig
```diff
@@ -1140,9 +1108,13 @@ pub const ReplayTower = struct {
             i += 1;
         }
 
-        try self.tower.initializeLockouts(flags);
+        var retained = std.BoundedArray(Lockout, MAX_LOCKOUT_HISTORY){};
+        for (self.tower.votes.constSlice(), 0..) |vote, index| {
+            if (flags.isSet(index)) retained.append(vote) catch unreachable;
+        }
+        self.tower.votes = retained;
```
Rexicon226: This is what I'm talking about :), yet another BoundedArray copy. Please just clear `self.tower.votes` and then append into it. Also bounded array has a `appendAssumeCapacity` function, should be used instead of the unreachable.

dnut: This is out of scope. I've already spent a few hours making improvements here and the rabbit hole keeps going deeper and deeper. At a certain point I need to stop.

Rexicon226: Ok, I can overlook the extra copy (can be pretty trivially removed by iterating over the bitset and swap removing), but at least use appendAssumeCapacity please.

Rexicon226: Thanks.

Sobeston (approved):

PR #1108, opened 2025-11-22
feat(identity): json private keys and base58 option for pubkeys

dnut:
1. Save private keys to disk as json instead of binary. This improves compatibility with all the existing solana tooling that is built around json key files and generally makes the data easier to work with since it's plain text.
2. Allow configuration of vote account with either the private key file path or the vote account address's base58 pubkey string. The private key file is useful in the event that the operator still has the file after generating it. But there's really no need for that file. You only need the address. The operator may have deleted the private key because it's no longer needed. In that case you can pass in the base58 encoded pubkey on the cli and it still works. This is the same way agave works.

### DIFF
--- a/src/identity.zig
+++ b/src/identity.zig
@@ -62,20 +62,26 @@ pub fn getOrInit(allocator: std.mem.Allocator, logger: Logger) !KeyPair {
 }
 
 /// Reads a pubkey either as a base58 string or from a json keypair file
-pub fn readPubkeyFlexible(logger: Logger, base58_or_kp_path: []const u8) !Pubkey {
+pub fn readPubkeyFlexible(
+    logger: Logger,
+    base58_or_kp_path: []const u8,
+) error{InvalidPubkeySource}!Pubkey {
     const pk_err = if (Pubkey.parseRuntime(base58_or_kp_path)) |p| return p else |e| e;
 
-    errdefer |e| logger.err().logf(
+    const kp_err = err: {
+        const file = std.fs.cwd().openFile(base58_or_kp_path, .{}) catch |e| break :err e;
+        defer file.close();
+
+        const keypair = parseKeypairJson(file.reader()) catch |e| break :err e;
+
+        return Pubkey.fromPublicKey(&keypair.public_key);
+    };
+
+    logger.err().logf(
         "Could not interpret '{s}' as a base58 pubkey due to {} or a json keypair file due to {}",
-        .{ base58_or_kp_path, pk_err, e },
+        .{ base58_or_kp_path, pk_err, kp_err },
     );
-
-    const file = try std.fs.cwd().openFile(base58_or_kp_path, .{});
-    defer file.close();
-
-    const keypair = try parseKeypairJson(file.reader());
-
-    return Pubkey.fromPublicKey(&keypair.public_key);
+    return error.InvalidPubkeySource;
 }
 
 fn parseKeypairJson(reader: anytype) !KeyPair {
@@ -117,7 +123,7 @@ test "readPubkeyFlexible for keypair filename" {
 }
 
 test "readPubkeyFlexible for nonsense string" {
-    try std.testing.expectError(error.FileNotFound, readPubkeyFlexible(.noop, "nonsense"));
+    try std.testing.expectError(error.InvalidPubkeySource, readPubkeyFlexible(.noop, "nonsense"));
 }
 
 const test_pubkey = "4gE1hGfMN781mVSpoRJwHdsbWLqZDWLNqTtrtimfhuWR";


src/identity.zig
```diff
@@ -77,27 +54,73 @@ pub fn getOrInit(allocator: std.mem.Allocator, logger: Logger) !KeyPair {
...
+/// Reads a pubkey either as a base58 string or from a json keypair file
+pub fn readPubkeyFlexible(logger: Logger, base58_or_kp_path: []const u8) !Pubkey {
+    const pk_err = if (Pubkey.parseRuntime(base58_or_kp_path)) |p| return p else |e| e;
+
+    errdefer |e| logger.err().logf(
+        "Could not interpret '{s}' as a base58 pubkey due to {} or a json keypair file due to {}",
+        .{ base58_or_kp_path, pk_err, e },
+    );
+
+    var file = try std.fs.cwd().openFile(base58_or_kp_path, .{});
```
InKryption:
```suggestion
    const file = try std.fs.cwd().openFile(base58_or_kp_path, .{});
```

src/identity.zig
```diff
@@ -77,27 +54,73 @@ pub fn getOrInit(allocator: std.mem.Allocator, logger: Logger) !KeyPair {
...
-pub fn readBinaryKeypairPubkey(path: []const u8) !sig.core.Pubkey {
-    var file = try std.fs.cwd().openFile(path, .{});
+/// Reads a pubkey either as a base58 string or from a json keypair file
+pub fn readPubkeyFlexible(logger: Logger, base58_or_kp_path: []const u8) !Pubkey {
+    const pk_err = if (Pubkey.parseRuntime(base58_or_kp_path)) |p| return p else |e| e;
+
+    errdefer |e| logger.err().logf(
+        "Could not interpret '{s}' as a base58 pubkey due to {} or a json keypair file due to {}",
+        .{ base58_or_kp_path, pk_err, e },
+    );
```
InKryption: `errdefer |e|` is due to be removed as a feature, so I would avoid using it here.

src/identity.zig
```diff
@@ -77,27 +54,73 @@ pub fn getOrInit(allocator: std.mem.Allocator, logger: Logger) !KeyPair {
...
         },
     }
 }
 
-/// Reads a binary keypair file and returns its public key as `Pubkey`.
-pub fn readBinaryKeypairPubkey(path: []const u8) !sig.core.Pubkey {
-    var file = try std.fs.cwd().openFile(path, .{});
+/// Reads a pubkey either as a base58 string or from a json keypair file
+pub fn readPubkeyFlexible(logger: Logger, base58_or_kp_path: []const u8) !Pubkey {
+    const pk_err = if (Pubkey.parseRuntime(base58_or_kp_path)) |p| return p else |e| e;
```
InKryption: Suggestion: why not just check if `std.mem.endsWith(u8, base58_or_kp_path, ".json")`? base58 doesn't contain `.` so should be unambiguous.

InKryption: Wouldn't we continue to want this to be a keypair, and not just the pubkey?

src/identity.zig
dnut: is it removed from 0.15? it's hard to imagine another way to represent this without making the code a lot uglier

src/identity.zig
dnut: The purpose of this function is to return a pubkey, not a keypair, so no.

This function is currently only used for the vote account. For the vote account, the private key is useless--we have no need for it. You only need to specify the vote account's pubkey in vote transactions. The signature comes from the validator's identity key, which is a separate key.

A vote account must to be owned by the vote program to be usable as a vote account. The private key is only ever used to sign a single transaction: the one transferring ownership to the vote program (this would happen in the solana cli before running sig). The vote program takes full control over the account. After that, the private key no longer has any use. The pubkey becomes nothing more than an address used to identify and lookup the account.

It's not unusual for people to dispose of private keys once an account is program-owned. A validator operator may not even be able to provide the private key.

src/identity.zig
dnut: 318ed6c94d1c0b1a9b1299a321244e7ae4696c21

src/identity.zig
InKryption: It's not removed yet, but it will certainly be removed. Just think it would be best to structure the code without it now, rather than later when you don't have the context/intention in mind.

I would probably base this on a branch on the condition of `std.mem.endsWith(u8, base58_or_kp_path, ".json")`, which would be unambiguous since the pubkey can't contain a period.

dnut: All I found is a proposal to have it removed: https://github.com/ziglang/zig/issues/23734

I don't see any clear conclusion indicating that the proposal will be accepted. I also don't find the justification for the proposal to be very convincing either.

I say we take full advantage of the tools at our disposal unless it's clear that we're going to need to stop.

> I would probably base this on a branch on the condition of std.mem.endsWith(u8, base58_or_kp_path, ".json"), which would be unambiguous since the pubkey can't contain a period.

I don't think this solves the problem. The solution is to add duplicated catches or split this into multiple functions.

dnut: Alright I see the accepted tag. Here's my best attempt at a clean implementation without errdefer capturing: 60a0409a43bb14ebd8d1afe9f94d5400fcdd276f

src/cmd.zig
```diff
@@ -1228,17 +1229,20 @@ fn validator(
 
     // Vote account. if not provided, disable voting
     const maybe_vote_pubkey: ?Pubkey = if (cfg.voting_enabled) blk: {
-        const vote_keypair_path = cfg.vote_account_path orelse default_path: {
+        const vote_keypair_path = cfg.vote_account orelse default_path: {
```
dadepo, line 1232:
```suggestion
        const vote_keypair = cfg.vote_account orelse default_path: {
```

dadepo (approved):

src/cmd.zig
dnut: actually i don't think that name is applicable either because this is either a keypair path or a pubkey string. i think the entire logic here should probably be reworked a bit which we can do later. right now this needs to merge to simplify vote account setup for the demo

PR #1113, opened 2025-11-24
feat(validator): Validator aware state tracking for epoch stakes and leader schedules

yewman:
# Summary
Introduces `EpochTracker` as a **temporary workaround** for fork-aware epoch stakes and leader schedule tracking. This satisfies immediate requirements for epoch transitions on testnet but should be replaced by fork-aware tracking of block contexts outside of replay. 

## EpochTracker (src/core/epoch_tracker.zig)
Cluster staked nodes are saved off in the first slot of a new epoch. These stakes are used during subsequent epochs for consensus, networking, and leader schedule computations. When forks occur across an epoch boundary, it is possible for different versions of this data to exist, until the first slot of the new epoch is 'rooted'. At this point, any competing forks are dropped and the staked nodes saved by the first rooted slot become absolute. 

The epoch tracker makes this information available for future epochs, and temporarily allows multiple 'unrooted' versions to exist until the first slot of a new epoch is rooted. When a slot is executed as the first slot in a new epoch, it creates a new entry in the 'UnrootedEpochBuffer' which stores the staked nodes and leader schedule computed from the current stakes cache (final state from previous epoch). If there are three forks across the boundary, three unique entries will be created and store in the 'UnrootedEpochBuffer'. When the first slot is rooted by consensus, we move the associated entry from the 'UnrootedEpochBuffer' to the 'RootedEpochBuffer' where it is persisted for future epochs. 

## LeaderSchedules (src/core/leader_schedule.zig)
The leader schedule is computed using the staked nodes. It is determined in advance of the epoch in which it applies. The leader schedule for epoch E is fully determined when the first slot of epoch E-1 is rooted, and is computed using the staked nodes as they exist at the end of epoch E-2. 

During transition periods, we need to be able to access leader schedules for both epochs. To support this 'LeaderSchedules' provides access to the current and next leader schedule (if it has been determined). 
**Note**: The current implementation does not provide access to the previous epoch leader schedule (E-1) through the `LeaderSchedules` once the first slot of E has been rooted.

### DIFF
--- a/src/core/ancestors.zig
+++ b/src/core/ancestors.zig
@@ -72,12 +72,9 @@ pub const Ancestors = struct {
         return .{ .ancestors = try self.ancestors.clone(allocator) };
     }
 
-    pub fn maxSlot(self: *const Ancestors) Slot {
-        var max_slot: Slot = 0;
-        for (self.ancestors.keys()) |slot| {
-            if (slot > max_slot) max_slot = slot;
-        }
-        return max_slot;
+    pub fn last(self: *const Ancestors) Slot {
+        const keys = self.ancestors.keys();
+        return keys[keys.len - 1];
     }
 
     pub fn subsetInto(
--- a/src/core/epoch_tracker.zig
+++ b/src/core/epoch_tracker.zig
@@ -784,7 +784,7 @@ test EpochTracker {
 
         _ = try epoch_tracker.insertUnrootedEpochInfo(
             allocator,
-            branch.maxSlot(),
+            branch.last(),
             &branch,
             try sig.core.stakes.randomEpochStakes(
                 allocator,
@@ -794,7 +794,7 @@ test EpochTracker {
             &.ALL_DISABLED,
         );
 
-        try epoch_tracker.onSlotRooted(allocator, branch.maxSlot(), &branch);
+        try epoch_tracker.onSlotRooted(allocator, branch.last(), &branch);
     }
 
     // Check that the root slot is 9 * 32 and epochs 6, 7, 8, 9 are available
@@ -815,10 +815,10 @@ test EpochTracker {
     // Check that trying to insert epoch info for epoch 9 fails because it is already rooted
     const branch = try Ancestors.initWithSlots(allocator, &.{319});
     defer branch.deinit(allocator);
-    fail_stakes.stakes.epoch = epoch_schedule.getEpoch(branch.maxSlot());
+    fail_stakes.stakes.epoch = epoch_schedule.getEpoch(branch.last());
     try std.testing.expectError(error.InvalidInsert, epoch_tracker.insertUnrootedEpochInfo(
         allocator,
-        branch.maxSlot(),
+        branch.last(),
         &branch,
         fail_stakes,
         &.ALL_DISABLED,
@@ -845,12 +845,12 @@ test EpochTracker {
     defer allocator.free(insert_ptrs);
     for (0..4) |i| insert_ptrs[i] = try epoch_tracker.insertUnrootedEpochInfo(
         allocator,
-        branches[i].maxSlot(),
+        branches[i].last(),
         &branches[i],
         try sig.core.stakes.randomEpochStakes(
             allocator,
             random,
-            .{ .epoch = epoch_schedule.getEpoch(branches[i].maxSlot()) },
+            .{ .epoch = epoch_schedule.getEpoch(branches[i].last()) },
         ),
         &.ALL_DISABLED,
     );
@@ -862,17 +862,17 @@ test EpochTracker {
     );
 
     // Check that another insert hits max forks
-    fail_stakes.stakes.epoch = epoch_schedule.getEpoch(branches[4].maxSlot());
+    fail_stakes.stakes.epoch = epoch_schedule.getEpoch(branches[4].last());
     try std.testing.expectError(error.MaxForksExceeded, epoch_tracker.insertUnrootedEpochInfo(
         allocator,
-        branches[4].maxSlot(),
+        branches[4].last(),
         &branches[4],
         fail_stakes,
         &.ALL_DISABLED,
     ));
 
     // Root the first slot from branch 2
-    try epoch_tracker.onSlotRooted(allocator, branches[2].maxSlot(), &branches[2]);
+    try epoch_tracker.onSlotRooted(allocator, branches[2].last(), &branches[2]);
 
     // Check that the pointers returned from insert match the pointers returned from get
     for (0..4) |i| try std.testing.expectEqual(
@@ -880,13 +880,13 @@ test EpochTracker {
         if (i != 2)
             try epoch_tracker.unrooted_epochs.get(&branches[i])
         else
-            try epoch_tracker.rooted_epochs.get(epoch_schedule.getEpoch(branches[i].maxSlot())),
+            try epoch_tracker.rooted_epochs.get(epoch_schedule.getEpoch(branches[i].last())),
     );
 
     // Check we can't insert unrooted if the epoch is already rooted
     try std.testing.expectError(error.InvalidInsert, epoch_tracker.insertUnrootedEpochInfo(
         allocator,
-        branches[4].maxSlot(),
+        branches[4].last(),
         &branches[4],
         fail_stakes,
         &.ALL_DISABLED,


src/core/leader_schedule.zig
```diff
@@ -279,41 +175,156 @@ pub const LeaderSchedule = struct {
...
-    pub fn slotLeaders(self: *SingleEpochSlotLeaders) SlotLeaders {
-        return SlotLeaders.init(self, get);
+    return slot_leaders;
+}
+
+pub fn computeFromMap(
+    allocator: Allocator,
+    slot_leaders_map: *const sig.utils.collections.PubkeyMap([]const u64),
+) !LeaderSchedule {
+    var slot_leaders = std.ArrayListUnmanaged(struct { Slot, Pubkey }).empty;
```
dnut: leak (see next comment for an alternate approach)
```suggestion
    var slot_leaders = std.ArrayListUnmanaged(struct { Slot, Pubkey }).empty;
    defer slot_leaders.deinit(allocator);
```

src/core/leader_schedule.zig
```diff
@@ -279,41 +175,156 @@ pub const LeaderSchedule = struct {
...
+    for (slot_leaders.items, 0..) |entry, i| {
+        leaders[i] = entry[1];
+    }
+
+    return .{
+        .leaders = leaders,
+        .start = slot_leaders.items[0][0],
+        .end = slot_leaders.items[slot_leaders.items.len -| 1][0],
+    };
+}
```
dnut, line 327: alternate approach without temp allocation

```suggestion
pub fn computeFromMap(
    allocator: Allocator,
    slot_leaders_map: *const sig.utils.collections.PubkeyMap([]const u64),
) !LeaderSchedule {
    var start: u64 = std.math.maxInt(u64);
    var end: u64 = 0;
    for (slot_leaders_map.values()) |slots| {
        for (slots) |i| {
            start = @min(i, start);
            end = @max(i, end);
        }
    }
    if (start > end) return error.NoSlotLeaders;

    const leaders = try allocator.alloc(Pubkey, 1 + end - start);
    for (slot_leaders_map.keys(), slot_leaders_map.values()) |leader, slots| {
        for (slots) |slot| leaders[slot - start] = leader;
    }

    return .{
        .leaders = leaders,
        .start = start,
        .end = end,
    };
}
```

src/core/stakes.zig
```diff
@@ -989,6 +990,194 @@ fn createStakeAccount(
...
+                options.commission_max,
+            ),
+            options.epoch + 1,
+        );
+        errdefer vote_state.deinit(allocator);
+        var vote_account = try VoteAccount.init(allocator, .{
+            .lamports = 1_000_000_000,
+            .owner = vote_program.ID,
+        }, vote_state);
+        errdefer vote_account.deinit(allocator);
```
dnut, line 1083: this could double free with the above since they both free the vote state

src/core/leader_schedule.zig
yewman: e24fb1336

src/core/leader_schedule.zig
yewman: e24fb1336

src/core/stakes.zig
yewman: e24fb1336

src/core/epoch_tracker.zig
```diff
@@ -0,0 +1,891 @@
...
+};
+
+/// Epoch Ring Buffer which holds 4 EpochInfo entries.
+///
+/// Inserts must increase monotonically by exactly 1 epoch.
+/// Gets must be within range of the last 4 inserted epochs.
+///
+/// No process should ever need to access E-3 or older EpochInfos and it is thus safe to
+/// deinitialize the corresponding data when overwriting entries with new epoch data.
+pub const RootedEpochBuffer = struct {
```
dnut:
```suggestion
const RootedEpochBuffer = struct {
```

I would also make all the methods private unless they're actually used in other files, which I assume they are not.

src/core/epoch_tracker.zig
```diff
@@ -0,0 +1,891 @@
...
+        if (root == 0 and (epoch == 1 or std.mem.allEqual(
+            ?*const EpochInfo,
+            &self.buf,
+            null,
+        ))) return true;
+        return epoch == root + 1;
+    }
+};
+
+pub const UnrootedEpochBuffer = struct {
```
dnut:
```suggestion
const UnrootedEpochBuffer = struct {
```
+ methods

src/core/epoch_tracker.zig
```diff
@@ -0,0 +1,891 @@
...
+
+/// Epoch Ring Buffer which holds 4 EpochInfo entries.
+///
+/// Inserts must increase monotonically by exactly 1 epoch.
+/// Gets must be within range of the last 4 inserted epochs.
+///
+/// No process should ever need to access E-3 or older EpochInfos and it is thus safe to
+/// deinitialize the corresponding data when overwriting entries with new epoch data.
+pub const RootedEpochBuffer = struct {
+    buf: [4]?*const EpochInfo = @splat(null),
```
dnut, line 398: Why is this `*const EpochInfo`? Why not directly store the EpochInfo? These allocations make it more complicated and risky to manage the data and I think it would be cleaner to do this without any allocations. You can allocate EpochTracker as a whole on the heap if it gets too big.

src/core/epoch_tracker.zig
```diff
@@ -0,0 +1,891 @@
...
+
+    pub fn deinit(self: *RootedEpochBuffer, allocator: Allocator) void {
+        for (&self.buf) |*maybe_entry| {
+            if (maybe_entry.*) |entry| {
+                entry.deinit(allocator);
+                allocator.destroy(entry);
+            }
+            maybe_entry.* = null;
+        }
+        self.* = undefined;
```
dnut: I thought as a team we decided against this pattern, and we actively oppose it with the `var mut_thing = thing; mut_thing.deinit()` approach. Would you mind changing this `deinit` function to accept a const pointer?

src/core/epoch_tracker.zig
```diff
@@ -0,0 +1,891 @@
...
+
+    pub fn deinit(self: *UnrootedEpochBuffer, allocator: Allocator) void {
+        for (&self.buf) |*maybe_entry| {
+            if (maybe_entry.*) |entry| {
+                entry.info.deinit(allocator);
+                allocator.destroy(entry.info);
+            }
+            maybe_entry.* = null;
+        }
+        self.* = undefined;
```
dnut: same as above

src/core/epoch_tracker.zig
```diff
@@ -0,0 +1,891 @@
...
+        const entry = try self.unrooted_epochs.take(ancestors);
+        errdefer {
+            entry.deinit(allocator);
+            allocator.destroy(entry);
+        }
+        try self.rooted_epochs.insert(
+            allocator,
+            entry,
+        );
+    }
```
dnut, line 230: This is going to trigger UB if there are any concurrent reads. I think the data in this struct needs to be protected with a RwLock.

It's kind of a pain because EpochInfo contains allocated data that you don't want to clone on every read, so you need some way for the caller to unlock it.
```zig
pub const EpochTracker = struct {
    lock: RwLock,
    ...

    pub fn getEpochInfo(
        self: *EpochTracker, // mut
        slot: Slot,
    ) !EpochInfoGuard {
        self.lock.lockShared();
        // ...
    }
        
    pub const EpochInfoGuard = struct {
        epoch_info: *const EpochInfo,
        lock: *RwLock,

        pub fn unlock(self: *const EpochInfoGuard) void {
            self.lock.unlockShared();
        }
    }
}
```

yewman: Why will it trigger UB? The entry is an allocation behind a const pointer. This operation just moves that pointer from being stored in the unrooted buffer to being stored in the rooted buffer. The underlying memory it points to does not change so I am not sure why it would cause UB for concurrent readers. Maybe I am missing something

src/core/epoch_tracker.zig
yewman: It was just to support moving entries from unrooted to rooted without needing to worry about refs to the unrooted entries. Once an EpochInfo is created, it will never be written, using an allocation means we can move from the unrooted to the rooted buffer without affecting readers who have a pointer to the unrooted epoch info

src/core/epoch_tracker.zig
yewman: 0f26db86

src/core/epoch_tracker.zig
yewman: 0f26db86

src/core/epoch_tracker.zig
yewman: 0f26db86

src/core/epoch_tracker.zig
yewman: 0f26db86

src/core/epoch_tracker.zig
dnut: Actually I was talking about these lines in `insert` that are going to destroy the old value:
```
        if (self.buf[index]) |old_value| {
            old_value.deinit(allocator);
            allocator.destroy(old_value);
        }
```
But I see how the data is assumed not to be accessed any more because it's too old, and you have that documented already. I just didn't read the doc.

src/core/epoch_tracker.zig
```diff
@@ -0,0 +1,895 @@
...
+const Epoch = sig.core.Epoch;
+const EpochSchedule = sig.core.epoch_schedule.EpochSchedule;
+const EpochStakes = sig.core.EpochStakes;
+const FeatureSet = sig.core.features.Set;
+const Pubkey = sig.core.Pubkey;
+const Slot = sig.core.Slot;
+const LeaderSchedules = sig.core.leader_schedule.LeaderSchedules;
+const LeaderSchedule = sig.core.leader_schedule.LeaderSchedule;
+
+pub const Cluster = struct {
```
dnut: A type called `Cluster` should directly represent a cluster itself. For example this could be data like the cluster's name, or its unique set of entrypoints. This type seems to be the subset of the genesis config that is immutable. I can see how it's cluster-scoped but I don't think it represents a cluster itself. I would prefer a name that indicates what the data actually is. How about `ClusterConstants`? Some other ideas:
- `ClusterConfig`
- `ClusterParams`
- `Chain*` (any of above suffixes work here. if anything i think this data is actually more chain-scoped than cluster-scoped since a cluster in principle can be rebooted to)
- `GenesisPermanentConfig`
- `ImmutableGenesisConfig`
- `ImmutableClusterParams`

src/core/epoch_tracker.zig
```diff
@@ -0,0 +1,895 @@
...
+            @as(f64, @floatFromInt(self.ticks_per_second));
+        return @as(u64, @intFromFloat(std.time.ns_per_s * seconds_per_slot));
+    }
+};
+
+/// This is a WORK AROUND for our current lack of fork awareness outside of replay.
+/// It ATTEMPTS to satisfy immediate requirements to run on testnet.
+/// It MUST be replaced by a better implementation as soon as possible.
+pub const EpochTracker = struct {
+    /// Cluster constants
```
dnut: this doc would be superfluous if the type name were more descriptive

dnut (changes requested):

src/core/epoch_tracker.zig
yewman: 7bc70ddf

src/core/epoch_tracker.zig
yewman: 7bc70ddf

dnut (approved):

src/core/ancestors.zig
```diff
@@ -72,6 +72,14 @@ pub const Ancestors = struct {
         return .{ .ancestors = try self.ancestors.clone(allocator) };
     }
 
+    pub fn maxSlot(self: *const Ancestors) Slot {
```
Rexicon226: Are our ancestors not currently stored in a way where the last entry will always be the latest one?

src/transaction_sender/leader_info.zig
```diff
@@ -5,17 +5,86 @@ const Allocator = std.mem.Allocator;
...
+/// NOTE: This struct is not really a 'cache', we should consider renaming it
+/// to a SlotLeaders and maybe even moving it outside of the core module.
+/// This more accurately describes the purpose of this struct as caching is a means
+/// to an end, not the end itself. It may then follow that we could remove the
+/// above pointer closure in favor of passing the SlotLeaders directly.
+pub const LeaderScheduleCache = struct {
+    epoch_schedule: EpochSchedule,
+    leader_schedules: RwMux(std.AutoArrayHashMap(Epoch, LeaderSchedule)),
+
+    const Self = @This();
```
Rexicon226: Don't use `const Self = @This()`.

src/core/ancestors.zig
yewman: Technically yes, however, it is not actually enforced so I went with the 'safer' impl. Happy to add insertion checks and then just return the last element here if you'd prefer, lmk. This will require changing a couple of tests which insert out of order, otherwise this method is only used for tests so could be behind an is_test comptime check so that the inefficient operation doesn't raise eyebrows. Ultimately I would still like to move away from ancestors as a hashmap, didn't you work on a bitset approach or something similar at one point? Did intergration prove painful for some reason?

src/transaction_sender/leader_info.zig
yewman: e250f1fb

dnut (dismissed):

src/core/ancestors.zig
dnut: > didn't you work on a bitset approach or something similar at one point? Did intergration prove painful for some reason?

I implemented this and got all the code to compile and all the tests to pass, but then I had conformance failures and decided not to invest any more time in it. I wonder if David ran into a similar issue?

Rexicon226: The behaviour of either implementation (hashmap or bitset based ancestors) will be that we know the highest slot without needing to search for it. I think it makes sense to just design the code around that.

Regarding the conformance failures, I haven't had any issues with that. I needed to work on something else, and this wasn't a priority, so I haven't finished the work.

PR #1115, opened 2025-11-25
fix(consensus): update `StatusCache` when slot is rooted

kprotty:
`StatusCache` is inserted to during `commitTransaction`, fetched from during` loadAndExecuteTransaction` to check if already committed, but is never pruned with a call to `addRoot/purgeRoot` so it leaks. This addresses that & fixes the `purgeRoot` logic to properly clean the cache & slot_deltas. It also merges all `RwMux` into one as the overhead isnt that bad.

### DIFF
--- a/src/core/status_cache.zig
+++ b/src/core/status_cache.zig
@@ -132,7 +132,17 @@ pub const StatusCache = struct {
         const forks = try hash_map.getOrPutValue(allocator, lookup_key, ForkStatus{});
         try forks.value_ptr.append(allocator, .{ .slot = slot });
 
-        try addToSlotDelta(allocator, state.mut(), blockhash, slot, key_index, &lookup_key);
+        // Add this key slice to the list of key slices for this slot and blockhash combo.
+        const fork_entry = try state.mut().slot_deltas.getOrPutValue(allocator, slot, .empty);
+        const fork_map: *StatusKv = fork_entry.value_ptr;
+
+        const hash_entry = try fork_map.getOrPutValue(
+            allocator,
+            blockhash.*,
+            .{ .status = .{}, .key_index = key_index },
+        );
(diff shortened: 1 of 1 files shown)


src/core/status_cache.zig
```diff
@@ -138,72 +132,79 @@ pub const StatusCache = struct {
...
+        const min_root = blk: {
+            var slot: ?Slot = null;
+            for (state.roots.keys()) |rooted| slot = @min(rooted, slot orelse rooted);
+            break :blk slot orelse return;
+        };
 
-            if (highest_fork.slot <= min_root) {
-                var purged_fork: HighestFork = (cache.mut().fetchOrderedRemove(key) orelse
-                    unreachable).value; // we just found this key!
+        std.debug.assert(state.roots.swapRemove(min_root));
```
dnut: This function works as long as it's called every time we add a new root. Which it is, so the code is fine. A part of me is paranoid though that someone will change StatusCache to violate that assumption without realizing.

You could add an assertion about the assumption:

```zig
std.debug.assert(state.roots.count() - MAX_CACHE_ENTRIES == 1);
```

Alternatively you could change this to ensure that it always deletes `state.roots.count() - MAX_CACHE_ENTRIES`. I see a few ways of doing this. The easiest would be wrapping it in a loop, which isn't optimal, but it's simple and straightforward to implement.

What do you think?

kprotty: I dont think purgeRoots as a separate function actually need to exist. I merged it with addRoot in a previous iteration, but decided to separate it out to keep with the theme of the "functions matching agave" in StatusCache. Can merge it back

kprotty: https://github.com/Syndica/sig/pull/1115/commits/ee4346ee993898461274950f3a8ce13c000a343d

Issue #1056, opened 2025-11-04
Transaction Error - InsufficientFundsForRent


dnut:
```
testnet: after 2870 slots in like 30m,
replayed slot 367794996 with error: replay.execution.ReplaySlotError{ .invalid_transaction = ledger.transaction_status.TransactionError{ .InsufficientFundsForRent = ledger.transaction_status.TransactionError__struct_146554{ .account_index = 0 } } }

```

dnut:
reproduced it again. 
```
validator -c testnet --disable-consensus --use-disk-index --skip-snapshot-validation --max-shreds 1100100100
```
- accounts and ledger: `2024-11-06-InsufficientFundsForRent`
- [sig.log](https://github.com/user-attachments/files/23417206/sig.log)

PR #1126, opened 2025-12-01
fix(replay): Match agave fee payer behaviour / hardcode lamports per signature

Sobeston:
closes https://github.com/Syndica/sig/issues/1056

See comment for details.

### DIFF
--- a/src/runtime/check_transactions.zig
+++ b/src/runtime/check_transactions.zig
@@ -136,7 +136,7 @@ pub fn checkFeePayer(
 
     // NOTE: FeeDetails (transaction fee, prioritization fee) in Agave is set
     // at the same time that compute budget limits are calculated. This value
-    // does not actually come from the feerate governor, but rather a field
+    // does not actually come from the fee rate governor, but rather a field
     // in the bank (fee_structure.lamports_per_signature).
     // The bank initialises this field using impl default for FeeStructure,
     // and never actually mutates it, meaning it remains the default value.


src/runtime/check_transactions.zig
```diff
@@ -132,16 +134,31 @@ pub fn checkFeePayer(
         rent_collector,
     ).rent_amount;
 
+    // NOTE: FeeDetails (transaction fee, prioritization fee) in Agave is set
+    // at the same time that compute budget limits are calculated. This value
+    // does not actually come from the feerate governor, but rather a field
```
yewman:
```suggestion
    // does not actually come from the fee rate governor, but rather a field
```

src/runtime/check_transactions.zig
```diff
@@ -132,16 +134,31 @@ pub fn checkFeePayer(
...
+    // in the bank (fee_structure.lamports_per_signature).
+    // The bank initialises this field using impl default for FeeStructure,
+    // and never actually mutates it, meaning it remains the default value.
+    // This means that lamports_per_signature is in effect *always* 5000, even
+    // when the fee rate governor disagrees.
+    //
+    // The other fields of FeeStructure, lamports_per_write_lock and
+    // compute_fee_bins, are also effectively unused.
+    //
+    // TODO: Stop hardcoding this value.
```
yewman, line 149: What do you think about removing the fee rate governor in Sig. From what I can tell that is not used either

Sobeston: Yeah I'm not sure. If Agave fixes this, it'd be better to keep more 1:1 with them. But they might not?

yewman: I would lean toward removing anything that is currently unused unless there it will be activated / used imminently. Not really within scope for this bug fix though so I have created an issue https://github.com/Syndica/sig/issues/1133

yewman (approved):

dnut (approved):

PR #1127, opened 2025-12-02
fix(time): use uptime clock & prevent overflow

kprotty:


src/shred_network/collector/shred_tracker.zig
```diff
@@ -410,8 +410,8 @@ const MonitoredSlot = struct {
     shreds: ShredSet = ShredSet.initEmpty(),
     max_seen: ?u32 = null,
     last_shred: ?u32 = null,
-    first_received_timestamp: Instant = Instant.UNIX_EPOCH,
-    last_unique_received_timestamp: Instant = Instant.UNIX_EPOCH,
+    first_received_timestamp: Instant = Instant.EPOCH_ZERO,
```
Rexicon226: Could you make these  `= .EPOCH_ZERO`

PR #1130, opened 2025-12-02
feat(sig): replace stdlib `assert` with custom assert.

kprotty:
Replace all instances of `std.debug.assert` with `sig.trace.assert`.
Should allow for better (or build-configured) error reporting.


dnut:
Another benefit is that you can distinguish assertion failures within sig from other unreachables

PR #1132, opened 2025-12-02
fix: reduce logging noise

dnut:
We keep seeing .TooFewTicks getting error logged in the alerts which is unnecessary because it happens ordinarily on forks.

We also see error logs about a missing vote account when starting the validator with consensus disabled, which makes no sense. Actually voting should not even be enabled even when consensus is enabled, unless explicitly enabled by the user. I turned it off by default, which removes the error logs unless the user includes the option, at which point the error would be justified.

To satisfy codecov I needed to add a way to test the logger so I added `TestLogger` which is a logger implementation just for asserting what gets logged.

### DIFF
--- a/src/trace/log.zig
+++ b/src/trace/log.zig
@@ -292,9 +292,8 @@ pub const TestLogger = struct {
         args: anytype,
     ) void {
         _ = fields; // we haven't needed to validate this in any tests yet.
-        const string = self.allocator.alloc(u8, std.fmt.count(fmt, args)) catch
+        const string = std.fmt.allocPrint(self.allocator, fmt, args) catch
             @panic("allocation failed in test logger");
-        _ = std.fmt.bufPrint(string, fmt, args) catch unreachable;
         self.messages.append(self.allocator, .{
             .level = level,
             .scope = scope orelse "",


src/trace/log.zig
```diff
@@ -255,6 +258,57 @@ pub const DirectPrintLogger = struct {
...
+    allocator: std.mem.Allocator,
+    messages: std.ArrayListUnmanaged(Message),
+
+    pub const Message = struct {
+        level: Level,
+        scope: []const u8,
+        content: []const u8,
+    };
+
+    const Self = @This();
```
Rexicon226: Do not use `Self`.

src/trace/log.zig
```diff
@@ -255,6 +258,57 @@ pub const DirectPrintLogger = struct {
...
+        comptime scope: ?[]const u8,
+        level: Level,
+        fields: anytype,
+        comptime fmt: []const u8,
+        args: anytype,
+    ) void {
+        _ = fields; // we haven't needed to validate this in any tests yet.
+        const string = self.allocator.alloc(u8, std.fmt.count(fmt, args)) catch
+            @panic("allocation failed in test logger");
+        _ = std.fmt.bufPrint(string, fmt, args) catch unreachable;
```
Rexicon226: `std.fmt.allocPrint`

src/trace/log.zig
dnut: 3050c837c8869e2fedf8a9296551b8aeb46877a0

src/trace/log.zig
dnut: d9ba9a2115170ef7dc94e7fd81dda0532a4477d3

Rexicon226 (approved):

PR #1135, opened 2025-12-03
feat(trace): filter logs by scope

dnut:
Previously, we could only specify a single log level for the entire application. With this change, you can specify different log levels for different scopes. For example:

`-l info,gossip=debug`

`info` will be used as default log level throughout the application, except the scope called `gossip` or any scope starting with `gossip.` will use `debug` as its log filter.

### DIFF
--- a/src/trace/level.zig
+++ b/src/trace/level.zig
@@ -3,7 +3,7 @@ pub const std = @import("std");
 pub const Level = enum {
     /// Error: something has gone wrong. This might be recoverable or might
     /// be followed by the program exiting.
-    err,
+    @"error",
     /// Warning: it is uncertain if something has gone wrong or not, but the
     /// circumstances would be worth investigating.
     warn,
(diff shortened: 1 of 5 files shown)


yewman (dismissed):
Makes sense to me

src/trace/level.zig
```diff
@@ -22,4 +24,228 @@ pub const Level = enum {
             .trace => "trace",
         };
     }
+
+    pub fn parse(string: []const u8) !Level {
+        // zig fmt: off
+        return
```
Rexicon226: `std.meta.stringToEnum`

src/trace/level.zig
```diff
@@ -22,4 +24,228 @@ pub const Level = enum {
...
+                try scopes.append(allocator, .{
+                    .name = scope,
+                    .level = try .parse(filter[eq_index + 1 ..]),
+                });
+            } else if (main_filter == null) {
+                main_filter = try .parse(filter);
+            } else return error.DuplicateDefaultFilter;
+        }
+
+        return Filters{
```
Rexicon226:
```suggestion
        return .{
```

src/trace/level.zig
```diff
@@ -22,4 +24,228 @@ pub const Level = enum {
...
+            .default = if (main_filter) |f| f else (Filters{}).default,
```
Rexicon226: Having a `default` field is strange, make it a decl and then just reference it.

dnut: The default is determined at runtime with command line argument.

Rexicon226: I mean you initialize a `Filters` with defualt values `{}`, and access the `default` field from it. That will always be `info`.

src/trace/level.zig
dnut: That's how I originally had this implemented but there is a mismatch with "err" and "error"

Rexicon226: Then name the field `@"error"`, I think that would be better anyways.

src/trace/level.zig
dnut: That's only if `main_filter` is null. If the default is specified on the cli, it uses that.

For example, let's say this is passed in the cli: `debug,gossip=info`, the default will be debug, as in:
```zig
.{
    .default = debug,
    .scopes = &.{.{ .name = "gossip", .level = .info }}
}
```

src/trace/level.zig
dnut: e14bc8d1afdf5d177582ad1b7a2331b4e2eb6b29

src/trace/level.zig
Rexicon226: I understand that fine, I'm just saying that the way you're writing the else branch here is bad, and should be a decl + `else default` instead.

dnut: 9782cb644c7b30c2e35232f40df55cc6332b6897

src/trace/level.zig
dnut: 400da8fe83cc85f885ae8bc492599af8b5ac3871

PR #1136, opened 2025-12-03
fix(gossip): Match Agave's gossip vote txn sanitization behaviour

yewman:
Agave uses two different sanitisation methods for transactions in the runtime and gossip. The gossip sanitisation allows transactions with more signatures than the transaction message specifies, the runtime sanitisation does not allow this. It looks like the gossip vote transactions which are failing sanitisation all have duplicated transaction signatures (two identical signatures). This change ommits error logging when the gossip verification fails with TooManySignatures

### DIFF
--- a/src/gossip/data.zig
+++ b/src/gossip/data.zig
@@ -624,7 +624,16 @@ pub const Vote = struct {
 
     pub fn sanitize(self: *const Vote) !void {
         try sanitizeWallclock(self.wallclock);
-        try self.transaction.validate();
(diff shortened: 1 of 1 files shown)


InKryption (dismissed):

insert_data_shred.err
dnut, line 1: did you commit this by accident?

yewman: yes

InKryption (approved):

PR #1137, opened 2025-12-03
refactor(gossip): improve startup time & misc

InKryption:
Measured improvements:
* Increase the buffer size used in `buildPullRequests` which is passed to `getThreadSafeGossipNodes`, aka `MAX_NUM_PULL_REQUESTS`. elements.
* Shorten `PULL_REQUEST_RATE` to 1 second, closer to agave's 500ms.

Other misc refactors:
* Store `entrypoints` as slice instead of arraylist.
* Remove some instances of `const Self = @This();`
* Make some parameters `*const`, and some unnecessary `var`s `const`.
* Add `Duration.mul`.
* Use some more decl literals.
* Unroll `VerifyMessageTask.run` and simplify it.
* Better static type safety for `getThreadSafeGossipNodes`.

---

Notes from some further investigation:

Originally the `PULL_REQUEST_RATE` was brought down to 500ms, to directly match agave, but that seemingly lead seemingly somewhat more volatile behavior, with the channels growing to rather huge sizes (indicating a struggle to keep up with the increased push and pull of gossip messages). An experiment in multi-threading the verifyMessage tasks by @kprotty seemed to lighten the load quite a bit, so that's probably worth looking into in the future.

Agave encodes imposes a limit on the number of push messages it will send out ([`MAX_NUM_PUSHES = 1 << 12`](https://github.com/anza-xyz/agave/blob/7fa607cf5a9a34b7ccbc9d881b4f0011d0edbdfe/gossip/src/crds_gossip_push.rs#L180)), but does not appear to impose a limit on the number of pull requests it sends out in the gossip logic itself, instead appearing to be bounded only to [1/16th](https://github.com/anza-xyz/agave/blob/7fa607cf5a9a34b7ccbc9d881b4f0011d0edbdfe/gossip/src/crds_gossip.rs#L342) of the nodes it can send pull requests to, which is distinct from our approach of having an actual strict cutoff, within which we send a maximal number of pull requests.
May be worth taking inspiration from that approach, or experimenting with other heuristics.

### DIFF
--- a/src/gossip/service.zig
+++ b/src/gossip/service.zig
@@ -233,7 +233,7 @@ pub const GossipService = struct {
 
         const verified_incoming_channel: *Channel(GossipMessageWithEndpoint) = try .create(allocator);
         errdefer verified_incoming_channel.destroy();
-        packet_outgoing_channel.name = "gossip verified incoming channel";
+        verified_incoming_channel.name = "gossip verified incoming channel";
 
         // setup the socket (bind with read-timeout)
         const gossip_address = my_contact_info.getSocket(.gossip) orelse return error.GossipAddrUnspecified;
@@ -525,10 +525,12 @@ pub const GossipService = struct {
             while (self.packet_incoming_channel.tryReceive()) |packet| {
                 defer self.metrics.gossip_packets_received_total.inc();
 
-                const message = verifyMessage(self.gossip_data_allocator, self.logger, &packet) catch |err| switch (err) {
-                    error.DeserializeFail => continue,
-                    error.SanitizeFail => continue,
-                    error.VerifyFail => continue,
+                const message = verifyMessage(
+                    self.gossip_data_allocator,
+                    self.logger,
+                    &packet,
+                ) catch |err| switch (err) {
+                    error.DeserializeFail, error.SanitizeFail, error.VerifyFail => continue,
                 };
                 errdefer bincode.free(self.gossip_data_allocator, message);
                 try self.verified_incoming_channel.send(.{


src/gossip/pull_request.zig
```diff
@@ -16,7 +16,7 @@ const RwMux = sig.sync.RwMux;
 const exp = std.math.exp;
 
 pub const MAX_BLOOM_SIZE: usize = 928;
-pub const MAX_NUM_PULL_REQUESTS: usize = 20; // labs - 1024;
+pub const MAX_NUM_PULL_REQUESTS: usize = 256;
```
dadepo, line 19: this looks like the corresponding constant in Agave https://github.com/anza-xyz/agave/blob/37f6e9ee9ebaff2251d7d205092c508521345c28/gossip/src/crds_gossip_pull.rs#L172 and it is set much higher. Any reason not to match the same value?

src/gossip/service.zig
```diff
@@ -223,15 +223,15 @@ pub const GossipService = struct {
...
 
-        var packet_outgoing_channel = try Channel(Packet).create(allocator);
+        const packet_outgoing_channel: *Channel(Packet) = try .create(allocator);
         errdefer packet_outgoing_channel.destroy();
         packet_outgoing_channel.name = "gossip packet outgoing channel";
 
-        var verified_incoming_channel = try Channel(GossipMessageWithEndpoint).create(allocator);
+        const verified_incoming_channel: *Channel(GossipMessageWithEndpoint) = try .create(allocator);
         errdefer verified_incoming_channel.destroy();
         packet_outgoing_channel.name = "gossip verified incoming channel";
```
dadepo: Not part of the diff but noticed it seems the name assigned here is wrong? variable is `packet_outgoing_channel` but the name is "gossip verified incoming channel"

src/gossip/service.zig
```diff
@@ -543,8 +525,16 @@ pub const GossipService = struct {
             while (self.packet_incoming_channel.tryReceive()) |packet| {
                 defer self.metrics.gossip_packets_received_total.inc();
 
-                task.packet = packet;
-                try task.run();
+                const message = verifyMessage(self.gossip_data_allocator, self.logger, &packet) catch |err| switch (err) {
+                    error.DeserializeFail => continue,
+                    error.SanitizeFail => continue,
+                    error.VerifyFail => continue,
+                };
```
dadepo:
```suggestion
                    error.DeserializeFail, error.SanitizeFail, error.VerifyFail => continue,
                };
```

src/gossip/pull_request.zig
InKryption: Nice find, didn't think to look at the filter. To answer: this value was mainly selected because using it yielded stable enough behavior. I had tested with a much higher value of 4096, which was a bit too volatile and caused startup to take longer, so I settled on 256. Once testnet is back up I'll try it out with 1024 and see if it meshes well with our code.

src/gossip/service.zig
InKryption: oop, nice catch, what a silly copy&paste mistake.

src/gossip/service.zig
InKryption: 75abc28954066b0021b80d831015529a61b32fa6

src/gossip/service.zig
InKryption: 75abc28954066b0021b80d831015529a61b32fa6

dadepo (approved):

src/gossip/service.zig
```diff
@@ -243,30 +243,26 @@ pub const GossipService = struct {
...
 
-        // setup the active set for push messages
-        var active_set = ActiveSet.init(allocator);
-        errdefer active_set.deinit();
-
         // setup entrypoints
-        var entrypoints = ArrayList(Entrypoint).init(allocator);
-        errdefer entrypoints.deinit();
+        const entrypoints: []Entrypoint = entrypoints: {
+            const entrypoint_addrs = maybe_entrypoints orelse break :entrypoints &.{};
```
kprotty, line 254: could this block be an `allocator.dupe`?

InKryption: I had the same initial gut reaction, but it's a transformation from `[]SocketAddr` to `[]Entrypoint`.

PR #1139, opened 2025-12-04
perf(replay): misc performance improvements

Sobeston:
Adds a bunch of useful tracy zones, plus some optimisations:
- Use a faster hash context for ArrayHashMaps of Pubkeys (PubkeyMap)
- Pointer params for Features.Set (this type is ~4KB, we shouldn't copy it on every function call)
- Optimised Stakes.convert (copies a hashmap more directly instead of re-constructing, saves 20-30ms/slot (!))

### DIFF
--- a/src/accountsdb/two/Unrooted.zig
+++ b/src/accountsdb/two/Unrooted.zig
@@ -28,28 +28,7 @@ pub const SlotIndex = struct {
     lock: std.Thread.RwLock,
     slot: Slot,
     is_empty: Atomic(bool),
-    entries: std.ArrayHashMapUnmanaged(Pubkey, AccountSharedData, Context, true),
-
-    const Context = struct {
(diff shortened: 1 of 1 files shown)


src/replay/consensus/core.zig
```diff
@@ -1580,6 +1592,11 @@ fn checkAndHandleNewRoot(
     status_cache: ?*sig.core.StatusCache,
     new_root: Slot,
 ) !void {
+    const zone = tracy.Zone.init(@src(), .{ .name = "checkAndHandleNewRoot" });
+    defer zone.deinit();
+
+    std.debug.print("checkAndHandleNewRoot\n", .{});
```
dnut: please remove this

Sobeston: removed in previous force push

dnut (approved):

InKryption (dismissed):

PR #1144, opened 2025-12-04
perf(replay): distribute PoH verify tasks by `entry.num_hashes`

kprotty:
PoH verifying currently distributes tasks based on entry count. This ends up in some calls to `verifyPoh` taking so long that they delay the entire of `replayActiveSlotsAsync`. Turns out that, inside `verifyPoh`, most time is spent on the iterative hash:

```zig
for (1..entry.num_hashes) |_| {
    current_hash = Hash.init(&current_hash.data);
}
```

This changes its task distribution to be even based on `num_hashes` instead of entry counts.

### DIFF
--- a/src/replay/exec_async.zig
+++ b/src/replay/exec_async.zig
@@ -172,7 +172,7 @@ pub const ReplaySlotFuture = struct {
 
 const PohVerifier = struct {
     future: *ReplaySlotFuture,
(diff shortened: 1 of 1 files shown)


Sobeston:
The size of the verifyPoh tasks are now very even - before we'd typically see one or two huge ones that would stall the whole slot sometimes.

Before
image

After
image

Sobeston (dismissed):

PR #1146, opened 2025-12-11
feat(gossip): publish gossip data to other services using channels

dnut:
Several components in the validator need to receive certain messages, such as votes, duplicate shreds, and epoch slots, as soon as they are received by gossip.

Previously, those components would depend on a pointer to the gossip table and periodically acquire the lock and scan over all the values in table to find the data they need. Now, gossip just sends the data they need over channels as soon as it is received.

I've implemented this approach only for votes so far. Getting duplicate shreds from gossip is not yet merged into main. I'll stack that on top of this PR.

There's also data that the validator will need to send *out* over gossip, but I'm not addressing this here.

I decided to apply an upper bound on the number of votes to process in a single iteration of the vote collector. This should be as large as the maximum number of expected validators, so we can process one vote per validator per slot. I explicitly defined this in MAX_VALIDATORS in sig.zig so the assumption is prominent and easily adjustable. I set it to 20,000. Currently there are 800 validators on mainnet.


InKryption (dismissed):
Great stuff all around

src/gossip/table.zig
```diff
@@ -131,26 +131,19 @@ pub const GossipTable = struct {
...
-        /// to properly de-allocate any returned data (like a previous, now overwritten entry).
-        pub fn deinit(self: InsertResult, allocator: std.mem.Allocator) void {
-            switch (self) {
-                .OverwroteExistingEntry => |existing| existing.deinit(allocator),
-                else => {},
-            }
-        }
+        pub const SuccessDetails = struct {
+            replaced: bool = false,
+            timeout: bool = false,
```
yewman: It seems like the only place the timeout result is used is in handleBatchPullResponses where we iterate over the insert_results and only call updateRecordTimestamp if timeout is false. If timeout is true, it is still possible that the entry was added or replaced right? It seems unintuitive to update the data regardless of the timeout and only update the timestamp if the timeout is false, could you provide some clarity with regards to this behaviour? Perhaps I am just misunderstanding something...

dnut: I agree this is weird. It's how gossip already worked, and it's how agave works as well. What I did was change the way the information is communicated between the gossip table and the gossip service while retaining the same logic.

Since your comment, I have refactored this further to inline the timeout check directly in handleBatchPullMessage instead of communicating it through the result type. This is because the timeout logic is specific to the context and doesn't really make sense within a general purpose insertion function in the gossip table. Actually the timeout needs to be handled differently for push messages than pull messages, so I fixed a bug (that I introduced) in this commit as well by handling each case separately. ea54e81e3225ca7a7cfe367f5686c687054b29e0

Now the question is *why* does it work this way? I spent some time looking at the code and here's my best guess...

We want to track which nodes have recently updated their gossip data. A simple way to do this would be to keep track of the latest wallclock we've seen from each node. But our code and agave don't take this code. We only accepts wallclocks within a specific range, and then save `now` instead of `wallclock` as the node's recent timestamp.

What if the wallclock is fabricated? We should put the least amount of trust we can in that number, and only extract information from it that would be difficult to fabricate. One strategy is to require the wallclock to be very close to right now. It would be difficult for an attacker to produce messages with inaccurate wallclocks and get them to propagate and influence other nodes, given this tight window filtering is in place in every node. Seeing a current wallclock is taken as a signal that the message was actually produced recently. After verifying it's a recent message, we can record a timestamp indicating the node was alive around this time. We save `now` instead of `wallclock` as the time the node was recently active. Perhaps this is because we still don't fully trust whatever time they specified, and we'd rather store a value that isn't in their control.

yewman: Nice, thanks for clarifying that. I think your justification makes sense 👍

PR #1147, opened 2025-12-16
refactor(consensus,replay): Unmanage ForkChoice & misc

InKryption:
Ordered map fixups:
* Add some capacity-related methods.
* Make it actually respect the `orderFn` config field.
* Use stricter type for `orderFn` field.
* Re-structure the default `order` function to be properly typed.

ForkChoice refactors:
* Use unmanaged sorted map variants.
* Use unmanaged array hash maps instead of managed hash maps.
* Privatize a couple things that don't need to be public
* Remove allocator field from `ForkChoice`, add allocator parameter to every function and usage site that requires it.
* Stop using silly hashmap iterators, take advantage of array hash map direct key/value slice iteration.
* Take opportunity to RLS-ify a bunch of code (decl literals and misc).


Sobeston (dismissed):
Looking good as a cleanup, didn't spot any regressions / issues

PR #1149, opened 2025-12-18
feat(runtime): execution performance

Sobeston:
- more tracy zones (particularly in runtime code)

- Optimisations (in rough order of importance)
  - moves AccountMetas from a BoundedArray to an ArrayList (really big perf improvement, these structs are huge)
  - puts most replayBatch allocations into an arena
  - optimises PROGRAM maps (no more b58 string hashmap conversion + lookup, now a plain pubkey array lookup)
  - removes some enumarray.get calls (this takes self as a copy, which can be large)
  - use a const pointer param for BorrowedAccount (reduce copy)
  - other misc

- temporarily disables fuzz=true on libsolfuzz_sig (slows down fixture execution, not currently used)
- fixed account.data memory leak in program loading

### DIFF
--- a/src/consensus/vote_listener.zig
+++ b/src/consensus/vote_listener.zig
@@ -1396,12 +1396,14 @@ pub const vote_parser = struct {
             const first_ix = message.instructions[0];
 
             var dedupe_map: [sig.runtime.InstructionInfo.MAX_ACCOUNT_METAS]u8 = @splat(0xff);
-            var account_metas: sig.runtime.InstructionInfo.AccountMetas =
-                try .initCapacity(allocator, 4);
-            defer account_metas.deinit(allocator);
 
             // Vote instructions have 4 accounts, below MAX_ACCOUNT_METAS (256)
-            std.debug.assert(first_ix.account_indexes.len == 4);
+            const n_instr_accounts = 4;
+            var account_metas: sig.runtime.InstructionInfo.AccountMetas =
+                try .initCapacity(allocator, n_instr_accounts);
+            defer account_metas.deinit(allocator);
+
+            std.debug.assert(first_ix.account_indexes.len == n_instr_accounts);
             for (first_ix.account_indexes, 0..) |acct_index_u8, i| {
                 const acct_index: usize = acct_index_u8;
                 if (dedupe_map[i] == 0xff)


src/runtime/transaction_execution.zig
```diff
@@ -148,14 +148,15 @@ pub fn TransactionResult(comptime T: type) type {
 
 /// [agave] https://github.com/firedancer-io/agave/blob/403d23b809fc513e2c4b433125c127cf172281a2/svm/src/transaction_processor.rs#L323-L324
 pub fn loadAndExecuteTransaction(
+    gpa: std.mem.Allocator,
     allocator: std.mem.Allocator,
```
dnut: Why does this take two allocators? Can you choose more descriptive names for each of them, to clarify the distinction?

In general I'd like to avoid using the name "gpa." It's no more descriptive than "allocator," while being more cryptic.

src/runtime/program_loader.zig
```diff
@@ -76,7 +77,7 @@ pub const LoadedProgram = union(enum(u8)) {
 };
 
 pub fn loadIfProgram(
-    allocator: std.mem.Allocator,
+    gpa: std.mem.Allocator,
```
dnut: Why did you change this to gpa? Is it just to correlate it with the gpa param in loadAndExecuteTransaction?

src/runtime/instruction_info.zig
```diff
@@ -25,7 +28,8 @@ pub const InstructionInfo = struct {
     /// transaction execution. We construct the account metas before transaction execution, so using an
     /// array of size MAX_ACCOUNTS_METAS + 1 allows us to check the account metas length during transaction
     /// execution and return the appropriate error.
-    pub const AccountMetas = std.BoundedArray(AccountMeta, MAX_ACCOUNT_METAS + 1);
+    // pub const AccountMetas = std.BoundedArray(AccountMeta, MAX_ACCOUNT_METAS + 1);
```
dnut:
```suggestion
```

src/vm/syscalls/lib.zig
```diff
@@ -110,7 +110,7 @@ pub const Syscall = enum {
                 if (value == bytes) break @enumFromInt(i); // assumes that the EnumArray will be in "array" mode, which it should be.
             } else return null; // no such hash
             // TODO: consider just making a build flag for harness builds that removes this check entirely for perf
-            return if (self.is_stubbed) &stubbed else map.get(syscall);
+            return if (self.is_stubbed) &stubbed else map.getPtrConst(syscall).*;
```
dnut, line 113: what difference does this make?

src/vm/interpreter.zig
```diff
@@ -112,13 +119,13 @@ pub const Vm = struct {
         }
 
         self.instruction_count += 1;
-        if (self.registers.get(.pc) >= self.executable.instructions.len) {
+        if (self.registers.getPtrConst(.pc).* >= self.executable.instructions.len) {
```
dnut, line 122: same question as above. just noting here to remind myself that this appears many times in this file.

no need to reply here unless there's something unique about this

src/vm/syscalls/lib.zig
Sobeston: The .get method on EnumArray unfortunately copies self, this avoids the copy

src/runtime/transaction_execution.zig
Sobeston: The reason for the two allocators is to separate the non-persistent arena allocations, and the allocations which must be persisted. Any naming suggestions?

src/runtime/program_loader.zig
Sobeston: Yes as it's a persistent allocator

src/runtime/transaction_execution.zig
dnut: I think `persistent_allocator` and `ephemeral_allocator` would be an improvement. Or maybe `program_map_allocator` is a better name since that's the only place where it's used, unless you think we should go with the more general mental model.

It might also be useful to add a comment to clarify that one allocator is for allocations that will not exceed the duration of the function call, while the other is for allocations that will be stored in the program map parameter.

src/runtime/instruction_info.zig
Sobeston: removed in d7c1c8e8603ec78ed5d861eecfa1d7e060ee9ca4

src/runtime/transaction_execution.zig
Sobeston: Renamed them to be more specific d7c1c8e8603ec78ed5d861eecfa1d7e060ee9ca4

src/runtime/program_loader.zig
Sobeston: Renamed them to be more specific d7c1c8e8603ec78ed5d861eecfa1d7e060ee9ca4

src/consensus/vote_listener.zig
```diff
@@ -1396,19 +1396,24 @@ pub const vote_parser = struct {
             const first_ix = message.instructions[0];
 
             var dedupe_map: [sig.runtime.InstructionInfo.MAX_ACCOUNT_METAS]u8 = @splat(0xff);
-            var account_metas = sig.runtime.InstructionInfo.AccountMetas{};
+            var account_metas: sig.runtime.InstructionInfo.AccountMetas =
+                try .initCapacity(allocator, 4);
```
kprotty: worth putting 4 in a constant?

Sobeston: c1fc175b494c93b0dc3fc59f4dc12992382491ac

PR #1154, opened 2025-12-19
feat(vm): redesign interpreter for jump table

Rexicon226:


src/vm/interpreter.zig
```diff
@@ -75,23 +89,47 @@ pub const Vm = struct {
...
+                self.result = .{ .err = error.ExecutionOverrun };
                 break;
+            }
+
+            self.step(undefined, pc) catch |err| switch (err) {
+                error.Stop => break,
+                else => |e| {
+                    self.result = .{ .err = e };
+                    break;
+                },
```
Sobeston, line 126: Confusing

Rexicon226: What exactly is confusing?

Sobeston: Passing in undefined without explanation

Rexicon226: Ah. That happens because I need a tail-call, which means the functino prototype needs to match with the table functions, while at the same time the actual instruction parameter is ignored. I can add that all as a comment when I get back.

Rexicon226: I've actually just simplified it further to not need to undefined, rather it takes the instruction we're going to call into and simply checks its opcode to index into the jump table.

src/vm/executable.zig
```diff
@@ -213,18 +214,20 @@ pub const Executable = struct {
 
                 .div64_imm,
                 .mod64_imm,
-                => if (!version.enablePqr()) {
-                    if (inst.imm == 0) return error.DivisionByZero;
-                } else {
+                => if (version.enablePqr()) {
                     store = true;
+                } else {
```
yewman: Just a thought regarding style, same applies below if agreed
```suggestion
                } else if (inst.imm == 0) {
                    return error.DivisionByZero;
                }
```

src/vm/interpreter.zig
```diff
@@ -60,6 +64,16 @@ pub const Vm = struct {
...
             .transaction_context = ctx,
             .result = .{ .ok = 0 },
+            .jump_table = switch (executable.version) {
+                // The only difference between v0 and v1 is how function calls work with
+                // dynamic frame pointers, which we can just eat the cost of checking the
+                // version in `pushCallFrame`, it doesn't seem to matter much.
+                .v0, .v1 => &v0.table,
+                .v2 => &v2.table,
+                .v3 => &v3.table,
+                .reserved => @panic("TODO"), // NOTE: cannot actually be hit
```
yewman: Can we replace the TODO with an appropriate message then?

src/vm/interpreter.zig
```diff
@@ -75,23 +89,52 @@ pub const Vm = struct {
...
+            if (self.executable.config.enable_instruction_meter and
+                self.instruction_count >= current_meter)
+            {
+                @branchHint(.unlikely);
+                self.result = .{ .err = error.ExceededMaxInstructions };
+                break;
+            }
+            self.instruction_count += 1;
+
+            const pc = self.registers.getPtrConst(.pc).*;
```
yewman, line 110: Out of curiosity why get the const ptr and deref instead of just using get here?

yewman: Ahh never mind I see that get takes self by value

src/vm/executable.zig
Rexicon226: I can agree with this yeah, I kinda made this change on auto-pilot.

src/vm/interpreter.zig
Rexicon226: This `reserved` tag will be removed while upgrading test-vectors, it was just a hack to match some silly behaviour Agave used to have. I'll just make it an unreachable with a TODO comment.

Rexicon226: d5f1d065a93f30064db63be29362b3bef257da12

src/vm/executable.zig
Rexicon226: d5f1d065a93f30064db63be29362b3bef257da12

PR #1156, opened 2025-12-19
feat(poh): faster hashing

Rexicon226:


src/benchmarks.zig
```diff
@@ -65,24 +65,36 @@ const benchmarks: std.EnumMap(Filter, []const Benchmark) = .init(.{
...
-        .resolution = .micros,
-    }},
-    .ledger = &.{ .{
-        .type = @import("ledger/benchmarks.zig").BenchmarkLedger,
-        .resolution = .nanos,
-    }, .{
-        .type = @import("ledger/benchmarks.zig").BenchmarkLedgerSlow,
-        .resolution = .millis,
-    } },
+    .gossip = &.{
```
kprotty, line 68: formatting oopsie?

src/core/entry.zig
```diff
@@ -142,58 +122,56 @@ pub fn verifyPoh(
...
+    if (num_signatures == 0) return .ZEROES;
+
+    // NOTE: The Firedancer code is currently designed around never having more
+    // than 30 transactions in a microblock, and pretty much always puts one
+    // transaction per entry. The Agave code used to batch, however has also
+    // switched to one transaction per entry recently.
+    //
+    // 1_000 txns per entry is 32kb, which is a fine amount of stack space to use.
+    // TODO: we should investigate what the true limit here is, perhaps switching
+    // to a slower but less space intensive algorithm if we're hit with strange
```
kprotty: Could try the merkle-tree stack algorithm from blake3: https://github.com/BLAKE3-team/BLAKE3/blob/master/reference_impl/reference_impl.rs#L320-L354

It requires only `node_stack: [log2(max_sigs)]Hash` memory, so given max sigs per txn is `1232 / 64 = 19`, `[25]Hash` would be enough for 1 million txns over all entries.

src/benchmarks.zig
Rexicon226: No, intentional, it was annoying to comment out individual benchmarks.

src/core/entry.zig
Rexicon226: Interesting idea, will try today.

src/core/entry.zig
```diff
@@ -142,58 +120,56 @@ pub fn verifyPoh(
...
```
yewman: I am not aware of a protocol enforced limit here so we should probably have some backup mechanism. Unless there is an implicitly defined limit which does ensure that there will never be more than 1_000 txns in an entry. 

Perhaps an assert would be better than nothing for now anyway

Rexicon226: Yeah the main thing stopping this PR is that I want to implement this stack based approach, since it lets us compute the merkle root with a reasonable number of hashes on the stack. The real limits are ~95k due to CUs and ~280k due to shred size limits. Not reasonable numbers to have on the stack.

yewman: Yeah I understand, at the moment if the assumptions are violated it will hit an index error right? Perhaps putting an explicit assert or check would make it easier to know what went wrong if the assumption does breakdown at some point. Then we can consider a fallback solution to handle such cases.

Rexicon226: We just won't merge this until I have a log stack impl working well, need to find that code again and I'll work on it more.

PR #1158, opened 2025-12-25
refactor(gossip,net,shred_network): Remove zig-network dependency

InKryption:
Just plucking all of our ties to zig-network.
Required implementing a very small `UdpSocket` type to replace the `Socket` abstraction from zig-network. For the rest, it was mostly replacing zig-network's `Endpoint` with our `SocketAddr`, and `std.net.Address`.

### DIFF
--- a/src/gossip/service.zig
+++ b/src/gossip/service.zig
@@ -1,5 +1,4 @@
 const std = @import("std");
-const network = @import("zig-network");
 const sig = @import("../sig.zig");
 const tracy = @import("tracy");
 
@@ -12,8 +11,6 @@ const ArrayList = std.ArrayList;
 const Thread = std.Thread;
 const Atomic = std.atomic.Value;
 const KeyPair = std.crypto.sign.Ed25519.KeyPair;
-const EndPoint = network.EndPoint;
-const UdpSocket = network.Socket;
 
 const Bloom = sig.bloom.Bloom;
 const Pubkey = sig.core.Pubkey;
@@ -53,7 +50,6 @@ const Duration = sig.time.Duration;
 const ExitCondition = sig.sync.ExitCondition;
 const SocketThread = sig.net.SocketThread;
 
-const endpointToString = sig.net.endpointToString;
 const globalRegistry = sig.prometheus.globalRegistry;
 const getWallclockMs = sig.time.getWallclockMs;
 const deinitMux = sig.sync.mux.deinitMux;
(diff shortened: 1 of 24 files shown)


src/gossip/service.zig
```diff
@@ -1176,23 +1177,23 @@ pub const GossipService = struct {
...
                 }
             }
         }
 
-        var push_iter = push_messages.iterator();
-        while (push_iter.next()) |push_entry| {
-            const gossip_values: *const ArrayList(SignedGossipData) = push_entry.value_ptr;
-            const to_endpoint: *const EndPoint = push_entry.key_ptr;
-
+        // var push_iter = push_messages.iterator();
```
yewman: May as well delete these

src/net/net.zig
```diff
@@ -1,7 +1,144 @@
...
+    }
+
+    fn toNativeAddressFamily(family: AddressFamily) u32 {
+        return @intFromEnum(family);
+    }
+};
+
+pub const UdpSocket = struct {
+    family: AddressFamily,
+    handle: posix.socket_t,
```
yewman, line 21: NIT: 'fd' instead of 'handle'?

src/net/net.zig
```diff
@@ -1,7 +1,144 @@
...
+    endpoint: ?std.net.Address,
```
yewman: Do we need/use this?

src/net/net.zig
Rexicon226: I would prefer handle. I don't see its connection to files other than it sharing the same underlying descriptor format.

src/gossip/service.zig
InKryption: Nice catch e42c2583f9a1a6f92890dcd46b1ac3eb2bf17f37

src/net/net.zig
InKryption: Nope, good observation 99dfac582e64614f5fdbaef18b4da67d4c718110

src/net/net.zig
InKryption: Yeah I don't think `fd` makes sense here. That name would imply that it's a generic file descriptor, but it's not, that's just the interface through which we use the socket - a more specific name would be "socket handle" or "socket file descriptor", but I think the fact that it's the field of a type called `UdpSocket` already provides that context.

Rexicon226 (dismissed):
lgtm

We also need to refactor the whole ContactInfo thing, but that'll be done in a later gossip refactor for v2.

yewman (approved):
Just a nit and a q

PR #1159, opened 2025-12-25
refactor(consensus,replay): Remove intermediate allocations from `UpdateOperations`

InKryption:
Still retains the data for tests, but makes the allocations entirely unnecessary for the normal operation of `ForkChoice`.

### DIFF
--- a/src/consensus/fork_choice.zig
+++ b/src/consensus/fork_choice.zig
@@ -178,8 +178,8 @@ pub const ForkChoice = struct {
     fn updateMetrics(self: *const ForkChoice) void {
         const now = Instant.now();
         const update_interval = now.elapsedSince(self.last_root_time);
-        self.metrics.fork_choice_update_interval.observe(update_interval.asMicros());
-        self.metrics.fork_choice_updates.inc();
+        self.metrics.update_interval.observe(update_interval.asMicros());
+        self.metrics.updates.inc();
 
         // Calculate basic consensus metrics
         var total_stake: u64 = 0;
@@ -467,6 +467,7 @@ pub const ForkChoice = struct {
                 }
             }
         }
+        self.metrics.pubkey_vote_batch_size.set(pubkey_votes.len);
 
         try self.latest_votes.ensureUnusedCapacity(allocator, pubkey_votes.len);
         for (pubkey_votes) |pubkey_vote| {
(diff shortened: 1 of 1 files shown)


src/consensus/fork_choice.zig
```diff
@@ -85,6 +53,8 @@ const ForkInfo = struct {
     /// and all competing forks for the same slot are invalid.
     is_duplicate_confirmed: bool,
 
+    const ForkWeight = u64;
```
yewman: Thoughts on removing this type alias? Personally I am not a fan of aliasing primitive types, I'm sure there are different views on this so perhaps it would be good to discuss in a an upcoming style chat.

src/consensus/fork_choice.zig
```diff
@@ -442,18 +409,148 @@ pub const ForkChoice = struct {
...
+        epoch_stakes: *const EpochStakesMap,
+        epoch_schedule: *const EpochSchedule,
+        /// Expects a value with methods:
+        /// * `fn subtractSlotStake(ctx, slot_hash_key: SlotAndHash, stake: u64) !void`
+        /// * `fn aggregateSlot(ctx, slot_hash_key: SlotAndHash) !void`
+        /// * `fn addSlotStake(slot_hash_key: SlotAndHash, stake: u64) !void`
+        ctx: anytype,
+    ) !SlotAndHash {
+        // Check for duplicate pubkeys in the same batch.
+        for (pubkey_votes, 0..) |current, i| {
```
yewman, line 463: Do you know how big these batches are in practice?

src/consensus/fork_choice.zig
InKryption: I don't have strong feelings about it 43888fe9e0ea56434466ff35d9a2edd93fb77a7f

src/consensus/fork_choice.zig
InKryption: I assume they aren't huge, since the main place this is called is [here](https://github.com/Syndica/sig/tree/43888fe9e0ea56434466ff35d9a2edd93fb77a7f/src/consensus/fork_choice.zig#L1443), where we can see these pubkey votes come from the dirty vote set of `latest_validator_votes`, which generally should just scale with the number of voting validators iiuc.

InKryption: I can throw in a metric for it and we can re-visit this if need-be.

InKryption: 37638c9ff3855ede4ae45d25e2b99cf586bb920a

yewman (approved):

PR #1160, opened 2026-01-05
chore(version): update ClientId

dnut:
Update to latest version of [client-ids.csv](https://github.com/solana-foundation/solana-validator-client-ids/blob/main/client-ids.csv) which is currently [0fff9f8e016972ff55680d011a4f81922c452f72](https://github.com/solana-foundation/solana-validator-client-ids/blob/0fff9f8e016972ff55680d011a4f81922c452f72/client-ids.csv)


src/version/version.zig
```diff
@@ -23,14 +23,16 @@ pub const ClientVersion = struct {
     };
 
     /// Keep up to date with:
-    /// https://github.com/solana-foundation/solana-validator-client-ids/blob/9d7afce390114e20fff0a0287ba7ba009a80805f/client-ids.csv
+    /// https://github.com/solana-foundation/solana-validator-client-ids/blob/main/client-ids.csv
```
InKryption: Minor nit: maybe should still be referring to a specific commit here instead of `main`, since what it points to may change?

Rexicon226: I agree, this should be a specific commit not main.

dnut: We need to keep up with main, not some arbitrary commit. I don't see any value in putting a permalink here. If that permalink differs in any way from main, it's simply wrong and we need to update this to whatever it has on main. People are going to look at that permalink and what is it going to tell them? This code was correct at some point in history? Why does that matter? What matters is that it's consistent with main. I'd rather just be direct and link to the actual relevant commit that this should be compared with.

Rexicon226: > People are going to look at that permalink and what is it going to tell them? 

It tells them why the code is written in that way. An outdated perma link is much more useful than a broken one. It also tells us when the last time it was updated was. Are you going to be checking the link every single day to make sure it hasn't changed? 
The contents of the link should not be "compared", but it explains the source of the code. 
If you wish it to be consistent with main, then actually do something that verifies that instead of simply making the link less reliable. Write a script that checks it, it would be only a few lines of code.

dnut: I'll add the permalink since you both would like to see it. I don't see the point, but I don't see the harm either.

I don't think the permalink provides any useful information. If someone wants to know when it was updated, they can look at the git log (which will include the permalink btw). If they want to understand why the code was written that way, it's safe to assume that there was previously a csv file with the same names and integers. There's no need to look at the permalink.

I see value in providing information to verify the correctness of the code. The only way to verify the correctness of the code is by comparing it to the current code on main. It doesn't matter whether it was correct in the past, it only matters if it's correct now.

yewman: > I agree, this should be a specific commit not main.

Agreed

dnut: I still don't see why a permalink is useful, but anyway I added it: cb744c775661d2745214359e459f2096baa8b104

src/version/version.zig
```diff
@@ -22,15 +22,17 @@ pub const ClientVersion = struct {
         .client = .sig,
     };
 
-    /// Keep up to date with:
-    /// https://github.com/solana-foundation/solana-validator-client-ids/blob/9d7afce390114e20fff0a0287ba7ba009a80805f/client-ids.csv
+    /// Keep up to date with: https://github.com/solana-foundation/solana-validator-client-ids/blob/main/client-ids.csv
```
Rexicon226, line 25: Instead of having this comment, how about actually ensuring that this happens! Here's a script I whipped up in 30 seconds,
```py
# Helper script that runs in CI and checks if our "ClientId" enum matches
# what the Solana Foundation has written down in their official IDs table. 

import urllib.request
import re
import sys

csv_text = urllib.request.urlopen(
    "https://raw.githubusercontent.com/"
    "solana-foundation/solana-validator-client-ids/"
    "main/client-ids.csv"
).read().decode()

csv = []
for line in csv_text.splitlines()[1:]:
    if not line:
        continue
    i, name = line.split(",", 1)
    csv.append((int(i), name.strip().lower().replace(" ", "_")))

text = open("src/version/version.zig").read()
m = re.search(r"ClientId\s*=\s*enum\s*\(\s*u16\s*\)\s*\{([^}]*)\}", text, re.S)
if not m:
    print("ClientId enum not found")
    sys.exit(1)

sig = []
for line in m.group(1).splitlines():
    line = line.strip()
    if not line:
        continue
    m2 = re.match(r"(\w+)\s*=\s*(\d+)", line)
    if not m2:
        continue
    sig.append((int(m2.group(2)), m2.group(1)))

if csv != sig:
    print("ClientId is outdated:")
    print("Solana:", csv)
    print("Sig:", sig)
    sys.exit(1)
```

Just run this in CI.

Rexicon226 (dismissed):

src/version/version.zig
dnut: Sure a script would be nice but I don't think it's in scope for this PR. We will also need to decide when this runs and how to get notified about the results.

PR #1167, opened 2026-01-05
fix(runtime): some out of bounds / overflows

Sobeston:


src/runtime/executor.zig
```diff
@@ -245,21 +245,27 @@ pub fn prepareCpiInstructionInfo(
...
 
         const index_in_callee_ptr = &dedupe_map[index_in_transaction];
         if (index_in_callee_ptr.* < deduped_account_metas.items.len) {
             const prev = &deduped_account_metas.items[index_in_callee_ptr.*];
             prev.is_signer = prev.is_signer or account.is_signer;
             prev.is_writable = prev.is_writable or account.is_writable;
-            try deduped_account_metas.append(tc.allocator, prev.*);
+
+            std.debug.assert(prev.index_in_transaction < InstructionInfo.MAX_ACCOUNT_METAS);
+            const new = prev.*; // this seems to dodge a miscompilation?
```
Rexicon226: Can you please change the comment to say that this is PRO bug, not a miscompilation. For all intensive purposes, this failing is valid Zig semantics (even if those semantics are flawed and will be changed at some point).

src/replay/resolve_lookup.zig
```diff
@@ -257,6 +257,7 @@ pub const LookupTableAccounts = struct {
     readonly: []const Pubkey,
 };
 
+// [agave] https://github.com/anza-xyz/agave/blob/b7547496510defa0a5b7541d99386ce1526bed06/accounts-db/src/accounts.rs#L105
```
Rexicon226: Can these links be head links for v.3.1.4? Or any later version. That's what I'm updating to everything in my conformance PR.

Rexicon226 (changes requested):

src/replay/resolve_lookup.zig
Sobeston: Updated to 3.1.4 5b35aa9a55c371d7cd2f14647f3ee633d14de9ff

src/runtime/executor.zig
Sobeston: 5b35aa9a55c371d7cd2f14647f3ee633d14de9ff

dnut (approved):

PR #1169, opened 2026-01-07
perf(snapshot): load snapshot tar directly into Rooted DB

kprotty:
We can skip `untar`'ing it to disk as folders & files, which is especially slow on certain file systems.

### DIFF
--- a/src/accountsdb/snapshot/load.zig
+++ b/src/accountsdb/snapshot/load.zig
@@ -159,7 +159,7 @@ pub fn loadSnapshot(
 
         const info = snapshot_files.incremental() orelse break :blk .{ null, null };
         const incremental_path = info.snapshotArchiveName();
-        const manifest, const status_cache = try insertFromSnapshotArchive(
+        break :blk try insertFromSnapshotArchive(
             allocator,
             logger,
             snapshot_dir,
@@ -167,7 +167,6 @@ pub fn loadSnapshot(
             maybe_rooted_db, // incremental doesnt have too many accounts to overwrite
             info.slotAndHash(),
         );
-        break :blk .{ manifest, status_cache };
     };
     errdefer if (maybe_incremental_manifest) |manifest| manifest.deinit(allocator);
     defer if (maybe_incremental_status_cache) |status_cache| status_cache.deinit(allocator);
@@ -234,11 +233,11 @@ pub fn loadSnapshot(
             return error.PubkeyNotInIndex;
         defer account.deinit(allocator);
 
-        var fba = std.io.fixedBufferStream(account.data);
+        var fbs = std.io.fixedBufferStream(account.data);
         break :blk try sig.bincode.read(
             allocator,
             sig.runtime.sysvar.SlotHistory,
-            fba.reader(),
+            fbs.reader(),
             .{},
         );
     };


src/accountsdb/snapshot/load.zig
```diff
@@ -22,6 +33,7 @@ pub const LoadedSnapshot = struct {
     collapsed_manifest: snapshot.Manifest,
     genesis_config: GenesisConfig,
     status_cache: ?snapshot.StatusCache,
+    tmp_dir: ?std.testing.TmpDir = null,
```
InKryption: Why does this need to exist? I think it would be preferable if the output directory was specified as a parameter to `loadSnapshot2`, that way the caller can place it wherever it is desired.

src/accountsdb/snapshot/load.zig
```diff
@@ -45,6 +58,489 @@ const LoadSnapshotOptions = struct {
     metadata_only: bool = false,
 };
 
+/// Similar to `loadSnapshot`, but avoids untar-ing the snapshot into files on disk.
+/// If provided, populates a Rooted DB instance with the accounts if DB is empty.
+pub fn loadSnapshot2(
+    allocator: Allocator,
+    rooted_db: *Rooted,
+    db_config: sig_config.AccountsDB,
```
InKryption: Let's take this opportunity to stop depending on sig_config here, and just throw the options required from this into `LoadSnapshotOptions`:
* `snapshot_dir: []const u8`
* `force_new_snapshot_download: bool`
* `max_number_of_snapshot_download_attempts: u64`

Although I do have some other thoughts on this about separating these responsibilities, so maybe take this comment in tandem with those.

src/accountsdb/snapshot/load.zig
```diff
@@ -45,6 +58,489 @@ const LoadSnapshotOptions = struct {
...
+        .max_threads = @intCast(@max(1, try std.Thread.getCpuCount())),
+    });
+    defer {
+        pool.shutdown();
+        pool.deinit();
+    }
+
+    const snapshot_dir_str = db_config.snapshot_dir;
+    var snapshot_dir = try std.fs.cwd().makeOpenPath(snapshot_dir_str, .{ .iterate = true });
+    defer snapshot_dir.close();
```
InKryption: Ideally, `snapshot_dir` would be a parameter, so that the caller can feed an existing directory or specify an output directory wherever they want, without needing to fiddle with paths. Would just need to specify in a doc comment that it's expected to be iterable.

src/accountsdb/snapshot/load.zig
```diff
@@ -45,6 +58,489 @@ const LoadSnapshotOptions = struct {
...
+            @intCast(min_mb_per_sec),
+            db_config.max_number_of_snapshot_download_attempts,
+            download_timeout,
+        );
+        defer full.close();
+        defer if (maybe_inc) |inc| inc.close();
+    }
+
+    const snapshot_files: SnapshotFiles = maybe_snapshot_files orelse
+        (try SnapshotFiles.find(allocator, snapshot_dir));
```
InKryption: What would you think of extracting this into its own function that can be consumed by this one? with a signature something like:
```zig
pub fn getSnapshotsFromOrDownloadIntoDir(
    params: struct {
        /// Get the snapshot file names from this directory,
        /// or download them from gossip into this directory.
        dir: std.fs.Dir,
        /// Setting this to null skips any attempt to download new snapshots.
        download: ?struct {
            gossip_service: ?*GossipService,
            max_number_of_snapshot_download_attempts: u64,
        },
    },
) !SnapshotFiles {
    // -- similar to current code -- //
}
```
Would allow tests for this to avoid invoking the download path by simply not calling this function (would instead just supply hard-coded `SnapshotFiles` parameter with a pre-populated dir), and all you'd need to do to obtain original behaviour is call this potentially-downloading function before `loadSnapshot2`.

src/accountsdb/snapshot/load.zig
```diff
@@ -45,6 +58,489 @@ const LoadSnapshotOptions = struct {
     metadata_only: bool = false,
 };
 
+/// Similar to `loadSnapshot`, but avoids untar-ing the snapshot into files on disk.
+/// If provided, populates a Rooted DB instance with the accounts if DB is empty.
+pub fn loadSnapshot2(
+    allocator: Allocator,
+    rooted_db: *Rooted,
+    db_config: sig_config.AccountsDB,
+    genesis_file_path: []const u8,
```
InKryption: How about we also make this just directly be `genesis_config: *const GenesisConfig`? The procedure for loading it isn't actually entangled with or related to loading the snapshot.

src/accountsdb/snapshot/load.zig
```diff
@@ -45,6 +58,489 @@ const LoadSnapshotOptions = struct {
...
+    logger: Logger,
```
InKryption, line 41: Nit: I think logger should go towards the start of the parameter list, right after allocator.

src/accountsdb/snapshot/load.zig
```diff
@@ -45,6 +58,489 @@ const LoadSnapshotOptions = struct {
...
+            snapshot_dir,
+            path.constSlice(),
+            .{ .slot = snapshot_files.full.slot, .hash = snapshot_files.full.hash },
+            .{ .snapshot_type = .full, .put_accounts = db_populate },
+        );
+    };
+    errdefer full_manifest.deinit(allocator);
+    errdefer full_status_cache.deinit(allocator);
+
+    const verify_hashes = !load_options.metadata_only and load_options.validate_snapshot;
```
InKryption: Maybe would be good to unify the `metadata_only` and `validate_snapshot` flags into one enum, like `behavior: enum { load_and_validate, load_no_validate, metadata_only }`. The current conditions can be derived from that, and then we'll no longer have a 4th redundant state for the purposes of snapshot validation.

src/accountsdb/snapshot/load.zig
```diff
@@ -45,6 +58,489 @@ const LoadSnapshotOptions = struct {
...
+        const manifest, const status_cache = try insertFromSnapshotArchive(
+            allocator,
+            logger,
+            rooted_db,
+            snapshot_dir,
+            incremental_path.constSlice(),
+            info.slotAndHash(),
+            .{ .snapshot_type = .incremental, .put_accounts = db_populate },
+        );
+        break :blk .{ manifest, status_cache };
```
InKryption: May as well do the same in both instances:
```suggestion
        break :blk try insertFromSnapshotArchive(
            allocator,
            logger,
            rooted_db,
            snapshot_dir,
            incremental_path.constSlice(),
            info.slotAndHash(),
            .{ .snapshot_type = .incremental, .put_accounts = db_populate },
        );
```

src/accountsdb/snapshot/load.zig
```diff
@@ -45,6 +58,489 @@ const LoadSnapshotOptions = struct {
...
+
+        // Incremental takes precedent.
+        if (maybe_incremental_status_cache) |inc_status_cache| {
+            full_status_cache.deinit(allocator);
+            break :blk inc_status_cache;
+        }
+
+        break :blk full_status_cache;
+    };
+    errdefer status_cache.deinit(allocator);
```
InKryption: Could just turn the `errdefer`s for deiniting the previous status cache `var`s into `defer`s, seeing as both will be made into noops once the `defer`s for emptying those `var`s take effect here.

src/accountsdb/snapshot/load.zig
```diff
@@ -45,6 +58,489 @@ const LoadSnapshotOptions = struct {
...
+
+    const slot_history = blk: {
+        const account = (try rooted_db.get(allocator, sig.runtime.sysvar.SlotHistory.ID)) orelse
+            return error.PubkeyNotInIndex;
+        defer account.deinit(allocator);
+
+        var fba = std.io.fixedBufferStream(account.data);
```
InKryption:
```
        var fbs = std.io.fixedBufferStream(account.data);
```

InKryption (changes requested):
I think this is good, but since we're here and about to start moving away from the old cruft, I have some suggestions I would like to see applied, so that we can start fresh with how we deal with snapshots.

InKryption (approved):

PR #1172, opened 2026-01-09
feat(consensus): Send duplicate slots from shred network to consensus (revival of #1027)

InKryption:


src/shred_network/collector/duplicate_shred_handler.zig
```diff
@@ -0,0 +1,810 @@
...
+
+        const num_chunks_usize = (proof_bytes.len + chunk_size - 1) / chunk_size;
+        if (num_chunks_usize > std.math.maxInt(u8)) return error.TooManyChunks;
+        const num_chunks: u8 = @intCast(num_chunks_usize);
+
+        const wallclock = sig.time.getWallclockMs();
+
+        const shred_index = layout.getIndex(shred_payload) orelse return error.InvalidShred;
+
+        const shred_type: sig.gossip.data.ShredType = .Code;
```
dnut: This doesn't seem right. I think it should be based on the actual shred's type.

```suggestion
        const shred_type = (layout.getShredVariant(shred_payload) orelse return error.InvalidShred).shred_type;
```

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,932 @@
...
+                    keys_to_remove.append(self.allocator, key) catch {};
+                    continue;
+                };
+                if (!g.found_existing) g.value_ptr.* = 0;
+                g.value_ptr.* +%= 1;
+                if (g.value_ptr.* > MAX_NUM_ENTRIES_PER_PUBKEY) {
+                    keys_to_remove.append(self.allocator, key) catch {};
+                }
+            } else {
+                keys_to_remove.append(self.allocator, key) catch {};
```
dnut: These catches don't make a lot of sense to me. Aren't they all OOM or db errors? I think these should all be `try`s.

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,932 @@
...
+            // Refresh cached staked nodes from epoch context
+            if (self.params.epoch_ctx_mgr.get(epoch)) |ctx| {
+                defer self.params.epoch_ctx_mgr.release(ctx);
+
+                // Clear and repopulate cached staked nodes
+                self.cached_staked_nodes.clearRetainingCapacity();
+                for (ctx.staked_nodes.keys(), ctx.staked_nodes.values()) |pubkey, stake| {
+                    self.cached_staked_nodes.put(self.allocator, pubkey, stake) catch {
+                        break;
+                    };
```
dnut:
```suggestion
                    try self.cached_staked_nodes.put(self.allocator, pubkey, stake);
```

src/shred_network/duplicate_shred_listener.zig
```diff
@@ -0,0 +1,932 @@
...
+const Pubkey = sig.core.Pubkey;
+const Slot = sig.core.Slot;
+
+const DuplicateShred = sig.gossip.data.DuplicateShred;
+
+const ResultWriter = sig.ledger.ResultWriter;
+const LedgerReader = sig.ledger.Reader;
+const DuplicateSlotProof = sig.ledger.meta.DuplicateSlotProof;
+
+const GOSSIP_SLEEP_MILLIS: u64 = 100;
```
dnut: unused

src/shred_network/collector/duplicate_shred_handler.zig
```diff
@@ -0,0 +1,810 @@
...
+const Allocator = std.mem.Allocator;
+const KeyPair = sig.identity.KeyPair;
+
+const Channel = sig.sync.Channel;
+const ShredInserter = sig.ledger.ShredInserter;
+const Slot = sig.core.Slot;
+
+const Logger = sig.trace.Logger("duplicate_shred_handler");
+
+pub const DUPLICATE_SHRED_HEADER_SIZE: u64 = 63;
```
dnut:
```suggestion
pub const DUPLICATE_SHRED_HEADER_SIZE: usize = 63;
```

src/shred_network/collector/duplicate_shred_handler.zig
```diff
@@ -0,0 +1,810 @@
...
+                .ChainedMerkleRootConflict => |conflict| {
+                    const shred_slot = conflict.original.commonHeader().slot;
+
+                    // TODO: Check feature flag for chained_merkle_conflict_duplicate_proofs
+                    // For now, we'll store it unconditionally. When feature checking is implemented,
+                    // this should check: if (!chained_merkle_conflict_duplicate_proofs) continue;
+
+                    // Although this proof can be immediately stored on detection, we wait until
+                    // here in order to check the feature flag, as storage in ledger can
+                    // preclude the detection of other duplicate proofs in this slot
```
yewman: Can we not pass the feature set and check it now?

src/shred_network/collector/duplicate_shred_handler.zig
```diff
@@ -0,0 +1,810 @@
...
+        try self.slots_pushed_to_gossip.append(slot);
+    }
+};
+
+pub fn serializeDuplicateProof(
+    allocator: std.mem.Allocator,
+    shred_payload: []const u8,
+    other_payload: []const u8,
+) ![]const u8 {
+    // TODO validate the shreds. Implement check_shreds in Agave
```
yewman, line 298: Are we relying on prior validations here? Should we implement that validation now?

src/shred_network/collector/duplicate_shred_handler.zig
```diff
@@ -0,0 +1,810 @@
...
+        shred1.payload(),
+    );
+    try ledger.db.commit(&write_batch);
+
+    const result: ShredInserter.Result = result: {
+        // Now create the duplicate shreds with the modified payloads
+        const duplicate_shred1: Shred = try .fromPayload(allocator, modified_payload1);
+        errdefer duplicate_shred1.deinit();
+        const duplicate_shred2: Shred = try .fromPayload(allocator, modified_payload2);
+        errdefer duplicate_shred1.deinit();
```
yewman:
```suggestion
        errdefer duplicate_shred2.deinit();
```

src/shred_network/collector/duplicate_shred_handler.zig
InKryption: Good catch 4894a44e3d1f0396159718dbb23b57691d154c09

src/shred_network/duplicate_shred_listener.zig
InKryption: Verily 25be6b44b27caaec9c292e3b1c4f4a41f4ce5b6a

src/shred_network/duplicate_shred_listener.zig
InKryption: 25be6b44b27caaec9c292e3b1c4f4a41f4ce5b6a

src/shred_network/duplicate_shred_listener.zig
InKryption: e1ffafe7bac0acbbfaea5277ce802110b855b8a4

src/shred_network/collector/duplicate_shred_handler.zig
InKryption: 7e4ebdc4dc761ab17b7969e82024536a150b2b5f

src/shred_network/collector/duplicate_shred_handler.zig
InKryption: e2a2b4de3785c08ca1ac7de8932cdbdd20a121a1

src/shred_network/collector/duplicate_shred_handler.zig
InKryption: Interesting note: it looks like in agave, the shred_type field of the duplicate shred type is marked as "semantically unused":
https://github.com/anza-xyz/agave/blob/ef5d037af9edd9273a2ac60538c7ec6460fbf7d3/gossip/src/duplicate_shred.rs#L28-L42

src/shred_network/collector/duplicate_shred_handler.zig
InKryption: I've looked through it, and it's actually not entirely clear to me what the best way would be to pass the feature set through.

yewman: What issue do you see with passing the feature set in? We could potentially add it to EpochInfo in the EpochTracker unless we move the slot tracker outside replay. While the feature set is constant for most of the duration of an epoch, it can be different across forks in an epoch until the first slot in said epoch is rooted. This could be addressed on top of epoch-state-tracking if required. Another thing to check is whether this feature is activated on all clusters. If it is, just add a note here and we can remove the check.

InKryption: 1e2810adf60a6f294e7551f190a1545ce6be24c4

src/shred_network/collector/duplicate_shred_handler.zig
InKryption: https://github.com/Syndica/sig/issues/1231

src/shred_network/collector/duplicate_shred_handler.zig
```diff
@@ -0,0 +1,830 @@
...
+                        conflict.original.payload(),
+                        conflict.conflict.data,
+                    );
+                },
+                .ChainedMerkleRootConflict => |conflict| {
+                    const shred_slot = conflict.original.commonHeader().slot;
+
+                    const chained_merkle_conflict_duplicate_proofs: bool = feature: {
+                        const epoch_tracker = self.epoch_tracker orelse break :feature false;
+                        const epoch_info = epoch_tracker.getEpochInfo(shred_slot) catch
```
yewman: The naming getEpochInfo is probably a bit confusing. This returns the epoch info containing the leader schedule and stakes which are applicable in the epoch of the provided slot. For features, we need the last epoch info which was rooted. 
```suggestion
                        const root_slot = epoch_tracker.root_slot.load(.monotonic);
                        const root_epoch = epoch_tracker.epoch_schedule.getEpoch(root_slot);
                        const epoch_info = epoch_tracker.rooted_epochs.get(root_epoch) catch
```
Note the use of the root bank here: https://github.com/anza-xyz/agave/blob/2717084afeeb7baad4342468c27f528ef617a3cf/turbine/src/cluster_nodes.rs#L771

PR #1196, opened 2026-02-02
fix(consensus): get votes to actually send and get retained

InKryption:
This PR does a few things:
* It actually passes through the `tpu_vote_addr` in contact info and other relevant places, so that consensus/replay can actually send votes through to the correct place.
* Actually passes the gossip table into consensus, so it can get contact info for leaders, and send votes through it.
* Use just one actual underlying allocator so as to avoid an invalid free between them, caused by gossip's local message broker (changed in this way, because it makes no sense for gossip to use two allocators anyway).
* Simplifies the logic for sending packets for votes so as to not accidentally give `sendmmsg` empty iovecs with empty buffers.
* Send our contact info to gossip every 500ms instead of every 15s, so that it actually survives long enough & often enough.
* Use just one `latest_validator_votes`; one was leftover before the overarching `SlotData` struct was introduced, meaning there were two inconsistent states tracking things differently.
* Fixed an integer overflow in the `RpcContextService` that occurs when on epoch 0 (which can arise when starting from genesis/testing on a local cluster).
* Replace incorrect usage of `std.posix.errno` with `std.os.linux.E.init`.
* A few other small fixes, like making a `var` into a `const`, removing an empty IES,

### DIFF
--- a/src/cmd.zig
+++ b/src/cmd.zig
@@ -2717,7 +2717,7 @@ const ReplayAndConsensusServiceState = struct {
                         // For now, default to using the node keypair as the authorized voter
                         // (same as Agave's default behavior when no --authorized-voter is specified)
                         // ref https://github.com/anza-xyz/agave/blob/67a1cc9ef4222187820818d95325a0c8e700312f/validator/src/commands/run/execute.rs#L136-L138
-                        &.{params.app_base.my_keypair}
+                        (&params.app_base.my_keypair)[0..1]
                     else
                         &.{},
                 },


src/adapter.zig
```diff
@@ -124,50 +122,48 @@ pub const RpcEpochContextService = struct {
...
+        if (self.state.contains(this_epoch)) return;
+
+        if (self.getLeaderSchedule(gpa, old_slot)) |ls2| {
+            const ctx2: sig.core.EpochContext = .{
+                .staked_nodes = .empty,
+                .leader_schedule = ls2,
+            };
+            errdefer gpa.free(ls2);
+            try self.state.put(this_epoch, ctx2);
+        } else |e| if (this_epoch == this_epoch) {
```
dnut: always true

src/gossip/service.zig
```diff
@@ -384,6 +384,8 @@ pub const GossipService = struct {
             for (push_msg_queue.queue.items) |*v| v.deinit(push_msg_queue.data_allocator);
             push_msg_queue.queue.deinit();
         }
+
+        // self.broker.deinit();
```
dnut: is this intentional?

src/core/transaction.zig
```diff
@@ -144,12 +144,21 @@ pub const Transaction = struct {
...
+        // TODO: remove these allocations, we have well-known bounds on all the numbers
+        const signatures = try allocator.alloc(Signature, message.signature_count);
         errdefer allocator.free(signatures);
 
-        for (signatures, keypairs) |*signature, keypair| {
-            const msg_signature = keypair.sign(msg_bytes, null) catch return error.SigningError;
-            signature.* = .fromSignature(msg_signature);
+        // TODO: This method can result in us signing the same message for the same keypair more than once.
+        // Should investigate whether a hashmap to amortize the signing is faster. The current only usecase
+        // is when we send votes, and that *does* sign the same message twice.
```
dnut: It's rare to have more than 2-3 signers in a transaction so I doubt a hashmap would be favorable over a linear search.

Is "This method can result in us signing the same message for the same keypair more than once." supposed to be part of the todo, or is it a separate note?

InKryption: Fair point - I left this in as part of @Rexicon226's original suggestion (pinged in case he has any specific views or intel on this).

The redundant signage should probably be its own comment.

src/gossip/service.zig
InKryption: No, nice catch. At some point early on I was debugging gossip, and wanted to fix some leaks for some local testing code, but the specific way in which this works caused issues in the actual validator on exits I believe, so I commented it out. Probably will just delete it, unless you think it's worth handling deallocation here for whatever reason.

src/core/transaction.zig
InKryption: f247f21f55053cba86e8c524e57d984194f79fb2

src/gossip/service.zig
InKryption: 0e64d2fe9d01b85045af4403a8015baca7a45d85

src/adapter.zig
InKryption: src/adapter.zig now deleted on main.

src/gossip/service.zig
```diff
@@ -66,7 +66,7 @@ const GossipMessageWithEndpoint = struct {
 
 pub const PULL_REQUEST_RATE: Duration = .fromSecs(1);
 pub const PULL_RESPONSE_TIMEOUT: Duration = .fromSecs(5);
-pub const ACTIVE_SET_REFRESH_RATE: Duration = .fromSecs(15);
+pub const ACTIVE_SET_REFRESH_RATE: Duration = .fromMillis(500);
```
hamza-syndica, line 69: I don't have enough context about this refresh rate change, but just pointing out that, I hope it's intentional to change the refresh rate from `15s` to `500ms`. Because, it feels it's a huge change.

InKryption: Yeah, this is 100% intentional. During a debugging session from a while ago, we figured out that 15 seconds is way too long of a period to wait to our active set, since it results in having very low retention, especially in terms of keeping our contact info in the synchronized table.

dnut (approved):

PR #1197, opened 2026-02-02
feat(rpc): implement getSlot RPC method

ajw221:
### Intent
- Implement getSlot RPC method
### Implementation
- Added ForkChoiceProcessedSlot and OptimisticallyConfirmedSlot atomic trackers to src/replay/trackers.zig
- Added SlotHookContext to src/rpc/methods.zig which reads slot state from SlotTracker based on commitment level
- Updated SlotTracker.root to be atomic for thread-safe access from RPC
- Wired latest_processed_slot updates in consensus mode (during vote handling) and bypass mode (from SlotTree.tip())
- Wired latest_confirmed_slot updates in vote listener when optimistic confirmation threshold is reached
### Ramifications
- ForkChoiceProcessedSlot uses store() rather than fetchMax() because the processed slot can decrease when fork choice switches to a different fork
- In bypass-consensus mode, processed slot is set to the highest fork tip, which differs from Agave's semantics where it only updates when voting
### Tests
- Updated existing unit tests to include the new slot tracker dependencies
- Added assertions verifying slot trackers remain at initial state after expected failures

### DIFF
--- a/src/replay/consensus/core.zig
+++ b/src/replay/consensus/core.zig
@@ -136,7 +136,7 @@ pub const TowerConsensus = struct {
             signing: sig.identity.SigningKeys,
             account_reader: AccountReader,
             ledger: *sig.ledger.Ledger,
-            slot_tracker: *SlotTracker,
+            slot_tracker: *const SlotTracker,
             /// Usually `.now()`.
             now: sig.time.Instant,
             registry: *sig.prometheus.Registry(.{}),


src/rpc/methods.zig
```diff
@@ -686,3 +686,58 @@ pub const common = struct {
...
+        return slot;
+    }
+
+    fn getLatestFinalizedSlot(self: @This()) !Slot {
+        const slot = self.slot_tracker.root.load(.monotonic);
+        if (slot == 0) {
+            return error.RpcNoFinalizedSlot;
+        }
+        return slot;
+    }
```
prestonsn: check if we need these broken up so redundantly like this

prestonsn: Could be useful for other methods, but I'd prefer to pass in a commitment enum arg.

prestonsn: Fixed in https://github.com/Syndica/sig/commit/40fae043fa0a03a0d9101fecd414913ea92c67ed

src/replay/consensus/process_result.zig
```diff
@@ -305,9 +305,15 @@ const TestReplayStateResources = struct {
...
+        var latest_confirmed_slot: sig.replay.trackers.OptimisticallyConfirmedSlot = .{};
+        const root: std.atomic.Value(Slot) = .init(0);
+
         self.slot_tracker = SlotTracker{
             .slots = .empty,
-            .root = 0,
+            .latest_processed_slot = &latest_processed_slot,
+            .latest_confirmed_slot = &latest_confirmed_slot,
+            .root = root,
         };
```
dnut: the stack pointers will be invalid after this function returns

src/replay/service.zig
```diff
@@ -1254,6 +1282,8 @@ pub const DependencyStubs = struct {
 
             .replay_threads = 1,
             .stop_at_slot = null,
+            .latest_processed_slot = &latest_processed_slot,
+            .latest_confirmed_slot = &latest_confirmed_slot,
```
dnut: returning stack pointers

src/replay/service.zig
```diff
@@ -1327,6 +1360,8 @@ pub const DependencyStubs = struct {
 
             .replay_threads = num_threads,
             .stop_at_slot = null,
+            .latest_processed_slot = &latest_processed_slot,
+            .latest_confirmed_slot = &latest_confirmed_slot,
```
dnut: returning stack pointers

src/consensus/vote_listener.zig
```diff
@@ -1995,6 +2027,11 @@ test "check trackers" {
         expected_trackers.items,
         actual_trackers.items,
     );
+
+    // Votes were processed but no stake was delegated to validators, so
+    // optimisitic confirmation was not reached.
+    try std.testing.expectEqual(0, confirmed_slot.get());
+    try std.testing.expectEqual(0, confirmed_slot.get());
```
dnut:
```suggestion
    try std.testing.expectEqual(0, confirmed_slot.get());
```

src/replay/consensus/core.zig
```diff
@@ -4930,12 +4954,11 @@ test "edge cases - duplicate confirmed slot" {
...
     const progress_map = &replay_state.progress_map;
-    std.debug.assert(slot_tracker.root == 0);
+    const root_slot0 = slot_tracker.root.load(.monotonic);
+    std.debug.assert(root_slot0 == 0);
 
-    const root_slot0 = slot_tracker.root;
     const root_slot0_hash = slot_tracker.getRoot().state.hash.readCopy().?;
-
-    std.debug.assert(root_slot0 == 0); // assert initial root value
+    std.debug.assert(root_slot0_hash.eql(.ZEROES)); // assert initial root hash
```
dnut, line 4938: redundant with line below

src/consensus/replay_tower.zig
```diff
@@ -5453,10 +5453,16 @@ pub const TestFixture = struct {
...
-                .constants = constants,
-                .state = state,
-            });
+            var latest_processed_slot: sig.replay.trackers.ForkChoiceProcessedSlot = .{};
+            var latest_confirmed_slot: sig.replay.trackers.OptimisticallyConfirmedSlot = .{};
+
+            break :blk try .init(
+                allocator,
+                &latest_processed_slot,
+                &latest_confirmed_slot,
```
dnut: returning stack pointers

src/rpc/methods.zig
```diff
@@ -686,3 +686,26 @@ pub const common = struct {
...
+        const slot = self.slot_tracker.getSlotForCommitment(commitment);
+        if (config.minContextSlot) |min_slot| {
+            if (slot < min_slot) {
+                return error.RpcMinContextSlotNotMet;
+            }
+        }
+        return slot;
+    }
+
+    pub fn getSlot(self: @This(), _: std.mem.Allocator, params: GetSlot) !GetSlot.Response {
```
dnut: in general you should use the struct name if it exists, not `@This()`. if the name does not exist (e.g. generics), you can define an alias
```suggestion
    pub fn getSlot(self: SlotHookContext, _: std.mem.Allocator, params: GetSlot) !GetSlot.Response {
```

src/rpc/methods.zig
```diff
@@ -686,3 +686,26 @@ pub const common = struct {
         shredVersion: ?u16 = null,
     };
 };
+
+pub const SlotHookContext = struct {
+    slot_tracker: *const sig.replay.trackers.SlotTracker,
+
+    fn getSlotImpl(
```
dnut: this function feels redundant. would you mind inlining it into getSlot?

src/replay/trackers.zig
```diff
@@ -10,6 +10,34 @@ const EpochSchedule = sig.core.EpochSchedule;
...
+        return self.slot.load(.monotonic);
+    }
+};
+
+pub const OptimisticallyConfirmedSlot = struct {
+    slot: std.atomic.Value(Slot) = .init(0),
+
+    pub fn update(self: *@This(), new_slot: Slot) void {
+        _ = self.slot.fetchMax(new_slot, .monotonic);
+    }
```
dnut: In theory, this number can go down. It's extremely unlikely to go down, but there's no guarantee that it increases monotonically.

src/cmd.zig
```diff
@@ -1460,6 +1460,9 @@ fn validator(
     else
         null;
 
+    var latest_processed_slot: sig.replay.trackers.ForkChoiceProcessedSlot = .{};
+    var latest_confirmed_slot: sig.replay.trackers.OptimisticallyConfirmedSlot = .{};
```
dnut: I don't think these need to be allocated all the way up here. I would push the ownership into one of the structs that is currently holding a pointer to it. I think it makes the most sense for SlotTracker to own the data if it can be done cleanly.

src/replay/trackers.zig
```diff
@@ -10,6 +10,34 @@ const EpochSchedule = sig.core.EpochSchedule;
...
 const SlotState = sig.core.SlotState;
+const Commitment = sig.rpc.methods.common.Commitment;
+
+pub const ForkChoiceProcessedSlot = struct {
+    slot: std.atomic.Value(Slot) = .init(0),
+
+    /// Set the current processed slot (heaviest fork tip).
+    /// Uses store() because this can decrease when the fork choice
+    /// switches to a different fork with a lower slot.
+    pub fn set(self: *@This(), new_slot: Slot) void {
```
dnut:
```suggestion
    pub fn set(self: *ForkChoiceProcessedSlot, new_slot: Slot) void {
```

similar pattern applies throughout the pr

https://github.com/Syndica/sig/blob/main/docs/docusaurus/docs/contributing/style-guide.mdx#method-parameters

dnut (changes requested):
my main overall thought is to move all the slot integers to be owned by the slot tracker which would solve several of my comments including the stack pointer uafs

src/replay/trackers.zig
prestonsn: i should have started here, thanks :)

src/consensus/vote_listener.zig
prestonsn: Oops, first one should be `processed`

src/replay/consensus/process_result.zig
prestonsn: https://github.com/Syndica/sig/pull/1197/commits/ccb5d8b65be0611196677f834c76f56039c81f57

src/replay/service.zig
prestonsn: https://github.com/Syndica/sig/pull/1197/commits/ccb5d8b65be0611196677f834c76f56039c81f57

src/replay/service.zig
prestonsn: https://github.com/Syndica/sig/pull/1197/commits/ccb5d8b65be0611196677f834c76f56039c81f57

src/consensus/vote_listener.zig
prestonsn: https://github.com/Syndica/sig/pull/1197/commits/ccb5d8b65be0611196677f834c76f56039c81f57

src/consensus/replay_tower.zig
prestonsn: https://github.com/Syndica/sig/commit/ccb5d8b65be0611196677f834c76f56039c81f57

src/replay/consensus/core.zig
prestonsn: https://github.com/Syndica/sig/pull/1197/commits/e95e1dbfc63837c5bc8eea8c2d566ac054de707c

src/rpc/methods.zig
prestonsn: https://github.com/Syndica/sig/pull/1197/commits/e792f674c7847fe8fe09fb84e0c5ce4e7ecd3b3c

src/cmd.zig
prestonsn: Agreed, this also worked out cleaner: https://github.com/Syndica/sig/pull/1197/commits/ccb5d8b65be0611196677f834c76f56039c81f57

src/replay/trackers.zig
prestonsn: https://github.com/Syndica/sig/pull/1197/commits/2135f2ffc805cd898b3cd1f114af3e681fd75899

src/rpc/methods.zig
prestonsn: https://github.com/Syndica/sig/pull/1197/commits/2135f2ffc805cd898b3cd1f114af3e681fd75899

src/replay/trackers.zig
prestonsn: I was mostly looking at this from the perspective of how Agave handles updating their optimistically confirmed bank. All updates of this bank are guarded with `if new_slot > optimistically_confirmed_bank.bank.slot()`: 

* https://github.com/anza-xyz/agave/blob/v3.1.8/rpc/src/optimistically_confirmed_bank_tracker.rs#L299-L301
* https://github.com/anza-xyz/agave/blob/v3.1.8/rpc/src/optimistically_confirmed_bank_tracker.rs#L376-L377
* https://github.com/anza-xyz/agave/blob/v3.1.8/rpc/src/optimistically_confirmed_bank_tracker.rs#L386-L387

So i'm not seeing a codepath in the Agave code where it can decrease (at least not from the perspective of how they are tracking their optimistically confirmed bank). 

I can see a scenario where the current confirmed slot could abruptly decrease after a higher slot already crossed confirmation threshold (network lag on some high stake votes for a lower slot perhaps?)

src/rpc/methods.zig
```diff
@@ -686,3 +686,19 @@ pub const common = struct {
...
+    pub fn getSlot(self: SlotHookContext, _: std.mem.Allocator, params: GetSlot) !GetSlot.Response {
+        const config = params.config orelse common.CommitmentSlotConfig{};
+        const commitment = config.commitment orelse .finalized;
+        const slot = self.slot_tracker.getSlotForCommitment(commitment);
+        if (config.minContextSlot) |min_slot| {
+            if (slot < min_slot) {
+                return error.RpcMinContextSlotNotMet;
+            }
+        }
+        return slot;
```
Sobeston: could early return here, e.g. 
```zig
const min_slot = config.minContextSlot orelse return slot;
return if (slot >= min_slot) slot else error.RpcMinContextSlotNotMet;
```

prestonsn: clean! https://github.com/Syndica/sig/pull/1197/commits/cf75dc5b2f7ac0c07934e1a74bb63328ca0d6f95

src/replay/consensus/core.zig
```diff
@@ -136,7 +136,7 @@ pub const TowerConsensus = struct {
             signing: sig.identity.SigningKeys,
             account_reader: AccountReader,
             ledger: *sig.ledger.Ledger,
-            slot_tracker: *const SlotTracker,
+            slot_tracker: *SlotTracker,
```
dnut: this doesn't need to be mutable

src/replay/trackers.zig
dnut: this means that agave is incorrect but it's not trivial to handle this correctly, and realistically it would probably never actually need to go down unless something very crazy happened to the cluster. it's fine to just target conformance with agave for now

src/replay/consensus/core.zig
prestonsn: https://github.com/Syndica/sig/pull/1197/commits/7d843701a40202a801fab5e619ae2a87ba6bbca4

PR #1203, opened 2026-02-03
chore(cmd): utils to support slot hash debug tooling

yewman:
# Summary
- Add option to specify validator directory 
- Support genesis downloads instead of relying on local copies
- Add ledger tool for checking ledger bounds and trimming to range
- Add flag to log the highest completed slot in the shred tracker

### DIFF
--- a/src/cmd.zig
+++ b/src/cmd.zig
@@ -1341,10 +1341,19 @@ fn ensureGenesis(
         &.{ cfg.validator_dir, "genesis.bin" },
     );
     errdefer allocator.free(existing_path);
-    if (!std.meta.isError(std.fs.cwd().access(existing_path, .{}))) {
+    const maybe_genesis_file: ?std.fs.File = std.fs.cwd().openFile(
+        existing_path,
+        .{},
+    ) catch |err| switch (err) {
+        error.FileNotFound => null,
+        else => return err,
+    };
+    if (maybe_genesis_file != null) {
         logger.info().logf("Using existing genesis file: {s}", .{existing_path});
+        maybe_genesis_file.?.close();
         return existing_path;
     }
+    allocator.free(existing_path);
 
     // Determine cluster for genesis
     const cluster = try cfg.gossip.getCluster() orelse {
--- a/src/config.zig
+++ b/src/config.zig
@@ -43,15 +43,21 @@ pub const Cmd = struct {
             .localnet => return error.MustProvideGenesisFileForLocalHost,
         } else return null;
 
-        std.fs.cwd().access(local_path, .{ .read = true }) catch {
+        const genesis_file = std.fs.cwd().openFile(local_path, .{}) catch {
             return null;
         };
+        genesis_file.close();
+
+        return local_path;
     }
 
     /// Derives a path relative to validator_dir if the param equals the default value.
     /// This is used to allow paths like snapshot_dir and geyser.pipe_path to be relative
     /// to validator_dir when using their default values, while still allowing explicit
     /// overrides.
+    /// TODO:
+    ///     - remove alloc by using fixed-size buffers for paths
+    ///     - https://github.com/orgs/Syndica/projects/2/views/10?pane=issue&itemId=156092227)
     pub fn derivePathFromValidatorDir(
         self: Cmd,
         allocator: std.mem.Allocator,
--- a/src/core/genesis_download.zig
+++ b/src/core/genesis_download.zig
@@ -36,12 +36,12 @@ pub fn downloadAndExtractGenesis(
     output_dir: []const u8,
     logger: Logger,
 ) DownloadError![]const u8 {
-    const genesis_url = try std.fmt.allocPrint(
-        allocator,
+    var genesis_url_buf: [std.fs.max_path_bytes]u8 = @splat(0);
+    const genesis_url = try std.fmt.bufPrint(
+        &genesis_url_buf,
         "{s}/{s}",
         .{ cluster_url, DEFAULT_GENESIS_ARCHIVE },
     );
-    defer allocator.free(genesis_url);
 
     logger.info().logf("Downloading genesis from {s}...", .{genesis_url});
     const archive_data = try downloadGenesisArchive(allocator, genesis_url, logger);
@@ -91,7 +91,7 @@ fn downloadGenesisArchive(
 /// Decompresses bz2 data using libbz2
 fn decompressBz2(allocator: Allocator, compressed_data: []const u8) DownloadError![]u8 {
     // Start with an estimate of 10x compression ratio
-    var dest_len: u32 = @intCast(@min(compressed_data.len * 10, 500 * 1024 * 1024));
+    var dest_len: u32 = @intCast(compressed_data.len * 10);
     var decompressed = allocator.alloc(u8, dest_len) catch return error.OutOfMemory;
     errdefer allocator.free(decompressed);
 
@@ -135,13 +135,8 @@ fn extractGenesisFromTar(
     output_dir: []const u8,
     logger: Logger,
 ) DownloadError![]const u8 {
-    std.fs.cwd().makePath(output_dir) catch |err| {
-        logger.err().logf("Failed to create output directory: {}", .{err});
-        return error.TarExtractError;
-    };
-
-    var dir = std.fs.cwd().openDir(output_dir, .{}) catch |err| {
-        logger.err().logf("Failed to open output directory: {}", .{err});
+    var dir = std.fs.cwd().makeOpenPath(output_dir, .{}) catch |err| {
+        logger.err().logf("Failed to open/create output directory: {}", .{err});
         return error.TarExtractError;
     };
     defer dir.close();


src/core/genesis_download.zig
```diff
@@ -0,0 +1,192 @@
+//! Genesis file download and unpacking from the network.
+//!
+//! This module handles downloading genesis.tar.bz2 archives from RPC nodes
+//! and unpacking them to the validator directory, similar to how Agave handles
+//! genesis bootstrap.
+
+const std = @import("std");
+const sig = @import("../sig.zig");
+const bzlib_c = @cImport({
```
Rexicon226: Do not use `cImport`, it is buggy and soon to be removed. Instead use `addTranslateC` in the build.zig, following existing libraries as an example.

src/core/genesis_download.zig
```diff
@@ -0,0 +1,192 @@
...
+        return error.HttpRequestFailed;
+    }
+
+    return response_body.toOwnedSlice() catch return error.OutOfMemory;
+}
+
+/// Decompresses bz2 data using libbz2
+fn decompressBz2(allocator: Allocator, compressed_data: []const u8) DownloadError![]u8 {
+    // Start with an estimate of 10x compression ratio
+    var dest_len: c_uint = @intCast(@min(compressed_data.len * 10, 500 * 1024 * 1024));
```
Rexicon226: Please use `u32` instead of `c_uint` here.

src/core/genesis_download.zig
```diff
@@ -0,0 +1,192 @@
...
+            @intCast(compressed_data.len),
+            0, // small: 0 = use normal algorithm
+            0, // verbosity: 0 = quiet
+        );
+
+        switch (result) {
+            bzlib_c.BZ_OK => {
+                // Shrink to actual size
+                if (dest_len < decompressed.len) {
+                    decompressed = allocator.realloc(decompressed, dest_len) catch decompressed;
```
Rexicon226: This makes little sense, `realloc` fails you if run out of memory (remap returns false + new alloc ooms), in which case you'll just be leaking the tail of the decompress region. Instead actually return the OOM if it happens, as there is no other way to handle this without leaking memory.

This is a pretty dubious API, but looking through the bzip2 code I guess this is effectively the only one.

src/core/genesis_download.zig
```diff
@@ -0,0 +1,192 @@
...
+                }
+                return decompressed[0..dest_len];
+            },
+            bzlib_c.BZ_OUTBUFF_FULL => {
```
Rexicon226: Why no realloc here? Perfect usecase for it.

src/cmd.zig
```diff
@@ -1029,8 +1109,152 @@ const Cmd = struct {
...
+        logger.err().log(
+            \\No genesis file path provided and no cluster specified. 
+            \\Use --genesis-file-path or --cluster"
+        );
+        return error.GenesisPathNotProvided;
+    };
+
+    // Otherwise, download genesis from network
+    logger.info().logf("Downloading genesis from {s} cluster...", .{@tagName(cluster)});
+    const cluster_url = switch (cluster) {
```
Rexicon226, line 1369: Use `getClusterEntrypoints` instead here.

src/config.zig
```diff
@@ -32,14 +33,35 @@ pub const Cmd = struct {
...
+        std.fs.cwd().access(local_path, .{ .read = true }) catch {
+            return null;
+        };
+    }
+
+    /// Derives a path relative to validator_dir if the param equals the default value.
+    /// This is used to allow paths like snapshot_dir and geyser.pipe_path to be relative
+    /// to validator_dir when using their default values, while still allowing explicit
+    /// overrides.
+    pub fn derivePathFromValidatorDir(
```
Rexicon226, line 61: It likely doesn't matter, but I want to point out that this API makes it impossible to free the memory without knowing afterwards whether it was equal or not.

yewman: Would duping param_value be preferable in your opinion? Or could return a bool to indicate whether to free or not? (We never actually free the config, but may as well make the api compatible with freeing the memory)

src/cmd.zig
yewman: Aren't they gossip entrypoints not RPC entrypoints?

Rexicon226: Er, you're right, I meant `rpcUrl`. I know this requires a bit of refactor, but I highly recommend you finally combine these silly `ClusterType` and `Cluster` types.

Basically I just want to share the code here:
https://github.com/Syndica/sig/blob/1d3c9bce60b0f40f29df2c51ae4127449e131068/src/cmd.zig#L1883

yewman: Haha yeah this needs a refactor for sure. I wanted to prevent a blow out in terms of PR size. I am addressing it in its own PR now.

yewman: https://github.com/Syndica/sig/pull/1206

src/core/genesis_download.zig
yewman: 2021d651

src/core/genesis_download.zig
yewman: 2021d651

src/core/genesis_download.zig
yewman: b1f83bb47c40

src/core/genesis_download.zig
yewman: b1f83bb47c4

src/cmd.zig
```diff
@@ -1140,6 +1292,91 @@ const AllocationMetrics = struct {
...
+        return try allocator.dupe(u8, provided_path);
+    }
+
+    // If genesis already exists in validator dir, use it
+    const existing_path = try std.fs.path.join(
+        allocator,
+        &.{ cfg.validator_dir, "genesis.bin" },
+    );
+    errdefer allocator.free(existing_path);
+    if (!std.meta.isError(std.fs.cwd().access(existing_path, .{}))) {
```
Rexicon226: Using access() like this is always a race. Instead you want to try and open the file, and catch a "doesn't exist" error.

src/config.zig
Rexicon226: In the best case, we'd take in a buffer of size `std.fs.max_path_bytes`, and just always copy out to it. Do not take ownership of the input memory.

src/core/genesis_download.zig
```diff
@@ -0,0 +1,191 @@
...
+}
+
+/// Extracts genesis.bin from a tar archive
+fn extractGenesisFromTar(
+    allocator: Allocator,
+    tar_data: []const u8,
+    output_dir: []const u8,
+    logger: Logger,
+) DownloadError![]const u8 {
+    std.fs.cwd().makePath(output_dir) catch |err| {
```
Rexicon226: Use `makeOpenPath` here instead. There's no way to make+open on Linux in a single atomic operation, but it's still just cleaner imo.

src/core/genesis_download.zig
```diff
@@ -0,0 +1,191 @@
...
+        return error.HttpRequestFailed;
+    }
+
+    return response_body.toOwnedSlice() catch return error.OutOfMemory;
+}
+
+/// Decompresses bz2 data using libbz2
+fn decompressBz2(allocator: Allocator, compressed_data: []const u8) DownloadError![]u8 {
+    // Start with an estimate of 10x compression ratio
+    var dest_len: u32 = @intCast(@min(compressed_data.len * 10, 500 * 1024 * 1024));
```
Rexicon226: This `@min` doesn't make much sense, we already ensure that the download won't be larger than 100mb above, so why limit it to 500mb now?

src/core/genesis_download.zig
```diff
@@ -0,0 +1,191 @@
...
+/// and extracts genesis.bin to the specified output directory.
+///
+/// Returns the path to the extracted genesis.bin file.
+pub fn downloadAndExtractGenesis(
+    allocator: Allocator,
+    cluster_url: []const u8,
+    output_dir: []const u8,
+    logger: Logger,
+) DownloadError![]const u8 {
+    const genesis_url = try std.fmt.allocPrint(
```
Rexicon226: Please use `bufPrint` + a `std.fs.max_path_bytes` buffer.

Rexicon226 (changes requested):

src/config.zig
yewman: I agree, in this instance that would require a slightly more significant refactor of cmd and config. I would prefer to address in a cmd / config refactor which is out of scope for this PR imo https://github.com/orgs/Syndica/projects/2/views/10?pane=issue&itemId=156092227.

src/core/genesis_download.zig
yewman: 62899f5f821

src/core/genesis_download.zig
yewman: Agreed 62899f5f821

src/core/genesis_download.zig
yewman: 62899f5f821

src/cmd.zig
yewman: Interesting, What makes this always a race? d80ce7817df6

Rexicon226: Google TOCTOU. You decide something based on whether the file path exists, however it may be deleted after the access() and before the open.

Rexicon226: You do not need this `blk` also.

yewman: ae300191cd09

Rexicon226 (approved):

dnut (approved):

PR #1204, opened 2026-02-03
feat(v2): init code + basic shred receiver

Sobeston:
Working
- Child process initialisation
- Child process exit handling
  - fmt printing error return traces
  - fmt printing panics traces
- memory sharing
- high level of security
  - regions are shared only as-needed, with write perms as-needed
  - mseal to stop our shared regions from being later modified by anyone
  - closing out all FDs, except for an optional stderr
  - seccomp to ban almost all syscalls
    - write syscalls only allowed on single provided stderr FD
    - intending that these will be per-service once we have some
- segfault/signal handling
- a basic net service 
- a basic shred receiver service

Example output:
```
$ zig build run -- config/testnet.zig.zon 
config: .{ .cluster = .testnet, .leader_schedule_file = { 115, 99, 104, 101, 100, 117, 108, 101, 46, 116, 120, 116 }, .gossip = .{ .port = 8001 }, .shred_network = .{ .recv_port = 8002 } }
Initialising: .net_pair
Initialised: Region `net_pair` shared with [ shred_receiver_0 (rw), net_0 (rw), ]
Initialising: .leader_schedule
Initialised: Region `leader_schedule` shared with [ shred_receiver_0 (ro), ]
Starting Service `shred_receiver_0`, pid: 975397
Starting Service `net_0`, pid: 975398
(net)binding 0.0.0.0:8002
Waiting for shreds on port 8002
slot: 442890464
erasure_set_index: 160
index: 162
shred_type: .code

slot: 442890464
erasure_set_index: 160
index: 169
shred_type: .data

slot: 442890464
erasure_set_index: 160
index: 173
shred_type: .data
 ```
    
What a minimal service looks like:
    
```zig
const std = @import("std");
const start = @import("start");

comptime {
    _ = start;
}

pub const name = "prng";
pub const panic = start.panic;

pub const ReadWrite = struct {
    prng_state: *std.Random.Xoroshiro128,
};

pub fn main(writer: *std.io.Writer, rw: ReadWrite) !noreturn {
    _ = writer;

    rw.prng_state.seed(123);
    while (true) rw.prng_state.seed(rw.prng_state.next());
}
```

### DIFF
--- a/v2/src/common/ring.zig
+++ b/v2/src/common/ring.zig
@@ -42,7 +42,7 @@ pub fn Ring(N: comptime_int, T: type) type {
 
             return .{
                 .ring = self,
-                .pos = tail % N,
+                .idx = tail % N,
                 .len = N - size,
             };
         }
--- a/v2/src/services/net.zig
+++ b/v2/src/services/net.zig
@@ -74,7 +74,7 @@ fn mainInner(pairs: []const *Pair) !noreturn {
         // recv
         for (pairs, sockets[0..sockets_len]) |pair, sock| {
             var slice = pair.recv.getWritable() catch continue;
-            const ptr = slice.one();
+            const ptr = slice.get(0);
             ptr.size = @intCast(std.posix.recvfrom(
                 sock,
                 &ptr.data,


v2/src/services/logger.zig
```diff
@@ -0,0 +1,41 @@
...
+
+pub const name: []const u8 = "logger";
+pub const _start = {};
+pub const panic = start.panic;
+
+pub const ReadOnly = struct {
+    prng_state: []const u8 = &.{},
+};
+
+pub fn main(writer: *std.io.Writer, ro: ReadOnly) !noreturn {
```
dnut: Please choose a different name. `main` has a very specific meaning and this does not conform to the pattern. My preference would be `run` but there are many other acceptable options.

some ideas:
- run
- entrypoint
- run_service
- service_entry
- service_main

v2/src/services/net.zig
```diff
@@ -0,0 +1,89 @@
...
+pub fn main(_: *std.io.Writer, rw: ReadWrite) !noreturn {
+    rw.ping.init(8000);
+
+    try tile(&.{rw.ping});
+}
+
+const MAX_SOCKETS = 10;
+
+/// `ports` is the list of ports it'll listen on.
+fn tile(pairs: []const *Pair) !noreturn {
```
dnut: let's avoid this name too. also why does this function need to be separate from the function above?

v2/src/common/net.zig
```diff
@@ -0,0 +1,23 @@
...
+pub const Packet = extern struct {
+    data: [1232]u8,
+    size: u16,
+    addr: std.net.Address,
+};
+
+pub const Pair = extern struct {
+    recv: Ring(MAX_PACKETS, Packet),
+    send: Ring(MAX_PACKETS, Packet),
+    port: u16,
```
dnut, line 23: Is there a benefit to requiring both send and recv for every socket? I get that we'll almost always need both, but if there's even a single exception to that rule, this abstraction falls apart. Why don't we just pass them in separately to be more flexible?

v2/.gitignore
```diff
@@ -0,0 +1,2 @@
+.zig-cache/
+zig-out/
```
dnut: I think this belongs in the root `.gitignore`

v2/build.zig.zon
```diff
@@ -0,0 +1,12 @@
+.{
+    .name = .sig,
+    .version = "0.2.0",
+    .paths = .{
+        "build.zig.zon",
+        "build.zig",
+        "src",
+    },
+    .minimum_zig_version = "0.15.2",
```
dnut, line 9: v1 and v2 should use the same version of zig to facilitate code reuse.

dnut (changes requested):

v2/.gitignore
Sobeston: it's fine to have multiple, conceptually this is a different project, this follows the same pattern as the conformance folder

v2/src/services/logger.zig
Sobeston: > main has a very specific meaning and this does not conform to the pattern. 

I don't really agree with this - every programming environment has its own idea of what `main` means; even zig has multiple kinds of main functions.

We're basically opting out of Zig's start.zig and using our own startup code instead. So I think the name reflects well?

v2/src/common/net.zig
Sobeston: > Is there a benefit to requiring both send and recv for every socket

I think this will be the norm for usage

> if there's even a single exception to that rule, this abstraction falls apart

Not really? If one of the two goes unused that's fine

v2/.gitignore
dnut: I understand it's fine to have multiple, but if the root project had the same lines in .gitignore, then adding these lines here would have no effect. It's in the same monorepo so it's not realistic to try to create a .gitignore as if this is an isolated repo. I don't see the benefit to this redundancy.

v2/src/services/logger.zig
dnut: Sure we could reinterpret the word main because we have a different startup, but I just don't see the point. I only see potential for confusion. I don't see any compelling reason why the word main is better than any of the other names I've suggested.

v2/.gitignore
dnut: I don't have a strong feeling that we need to remove this file. My main concern is that we have inconsistency in our approach and it feels slightly chaotic. In here, and in conformance, we ignore these names at any nested level, but in the root we only ignore them if they exist at the root level. 

My thought is that the approach here is actually good: we shouldn't be committing folders called `.zig-cache` or `zig-out` regardless of where they are. The problem is that it's only being applied locally in a few places. I'd rather use this as the general approach. So I would change the repo root's .gitignore to follow this same pattern. Once we do that, there's not really a need for this file. But deleting this file isn't necessary my goal here, so I'm open to leaving it if you see some good reasoning for it to exist.

v2/src/services/logger.zig
yewman: I prefer 'main'. What specific meaning are you referring to that does not conform to this pattern? In this context it is the entry point to a service which is intended to run as isolated process. Personally I don't see what is confusing about it or how any of the suggested names are an improvement

dnut: The problem is simply that it's a source of confusion. The word "main" is almost always used make it so the file can be compiled to a standalone binary executable using the standard build process. That's not what's going on here. I don't see why you would use the word "main" to mean something different than what it's normally used for, when you have the freedom to select any name, and there are actually more descriptive names you could choose. This already caused confusion when this was demoed to the team with multiple people making wrong assumptions about how this code works because of this name. I don't want to use a name that I know is going to confuse people when there are equally valid alternatives that are less confusing.

v2/src/common/net.zig
dnut: fair enough

v2/build.zig.zon
dnut: Never mind, I have a solution. Stick with 0.15

v2/src/services/logger.zig
jbuckmccready: I'll pitch in my opinion:

`main` as a symbol, by default, in most (overwhelming majority?) programming languages is the the executable entry point when your code is compiled and linked. If `main` isn't the entry point for the executable then I don't think we should use `main`, it can only add confusion.

Using `service_main`, `service_entry`, `{service_name}_entry` etc. will communicate the same thing, and eliminate any possible confusion to those coming from other projects or languages.

yewman: I don't mind. Anything works

v2/src/start_service.zig
```diff
@@ -0,0 +1,311 @@
...
+
+    const stderr: std.fs.File = .{ .handle = params.stderr };
+    var writer_buf: [0]u8 = undefined;
+    var writer = stderr.writer(&writer_buf);
+
+    const ret_val =
+        if (!@hasDecl(root, "ReadWrite")) err: {
+            var read_only: root.ReadOnly = .{};
+            inline for (@typeInfo(root.ReadOnly).@"struct".fields, 0..) |field, i| {
+                @field(read_only, field.name) = @ptrCast(params.ro[i].?[i..params.ro_len[i]]);
```
yewman:
```suggestion
                @field(read_only, field.name) = @ptrCast(params.ro[i].?[0..params.ro_len[i]]);
```
And same for the rest

v2/src/services/logger.zig
dnut: I feel strongly that this should not be named `main`. I'm fine with a variety of other options.

kprotty: My 2 cents: `main` is the entrypoint of a compiled program. The distinction here is "program" instead of "executable". For example, DLLs on windows are compiled programs with entrypoints (`DllMain`) but they arent executable (`WinMain`). The services here are separately compiled ~~dynamic~~ static libraries; question is whether to call both entrypoints the same thing.


Personally had no confusion; The `pub fn main()` in zig doesnt always allow it to be compiled as an executable with `zig build-exe` (e.g. when os target is freestanding, wasm, or BPF). The part which allows that is done through the stdlib's `start.zig`, which is being overwritten in v2 due to the stdlib's version allocating memory & doing other syscalls in an uncontrolled manner. Given that, `pub fn main()` (to me at least) only indicates it's an entrypoint to _something_. I usually look to `build.zig` for whether files containing `main` will be used as an executable (e.g. the benchmark or fuzzer files).

Sobeston: We will need to switch to execve at some point for security, so these will become even more "real" entry points in future

v2/.gitignore
Sobeston: 1a935bf101980a3d23f3202e2586613e00dabfa5

v2/src/services/net.zig
Sobeston: This has been removed

v2/src/services/logger.zig
Sobeston: changed in https://github.com/Syndica/sig/pull/1204/changes/8dd42f054aada25fc9ad0055dcc973a95c05758d#diff-93014e408c6c7e5059d0ace10a7f1cca42d41ebdcae2d71ebbbf8eb6d292c63a

v2/src/common/crypto/ed25519/pippenger.zig
```diff
@@ -0,0 +1,151 @@
...
+        const bit_offset = i * w;
+        const u64_idx = bit_offset / 64;
+        const bit_idx: u6 = @truncate(bit_offset);
+        const element = scalars[u64_idx] >> bit_idx;
+
+        const below = bit_idx < @as(u64, 64) - w or u64_idx == 3;
+        const bit_buf: u64 = switch (below) {
+            true => element,
+            else => element | (scalars[1 + u64_idx] << @intCast(@as(u64, 64) - bit_idx)),
+        };
```
InKryption: I don't have any specific feedback, but this is blursed. Why `true` and `else`?

v2/src/common/crypto/ed25519/straus.zig
```diff
@@ -0,0 +1,161 @@
...
+    scalars: []const CompressedScalar,
+) ed25519.ReturnType(encoded, ristretto) {
+    std.debug.assert(points.len <= max);
+    std.debug.assert(scalars.len <= max);
+    std.debug.assert(points.len == scalars.len);
+
+    var scalars_in_radix: std.BoundedArray([64]i8, max) = .{};
+    for (scalars) |scalar| {
+        scalars_in_radix.appendAssumeCapacity(asRadix16(scalar));
+    }
```
InKryption: Probably should avoid using BoundedArray:
```suggestion
    var scalars_in_radix_buf: [max][64]i8 = undefined;
    var scalars_in_radix: std.ArrayListUnmanaged([64]i8) = .initBuffer(&scalars_in_radix_buf);
    for (scalars) |scalar| {
        scalars_in_radix.appendAssumeCapacity(asRadix16(scalar));
    }
```

v2/src/common/crypto/ed25519/straus.zig
```diff
@@ -0,0 +1,161 @@
...
+
+    var lookup_tables: std.BoundedArray(LookupTable, max) = .{};
```
InKryption: Same here

v2/src/common/crypto/ed25519.zig
```diff
@@ -0,0 +1,700 @@
...
+    message: []const u8,
+) !void {
+    std.debug.assert(signatures.len <= max);
+    std.debug.assert(public_keys.len <= max);
+    std.debug.assert(signatures.len == public_keys.len);
+
+    var s_batch: std.BoundedArray(CompressedScalar, max) = .{};
+    var a_batch: std.BoundedArray(Edwards25519, max) = .{};
+    var hram_batch: std.BoundedArray(CompressedScalar, max) = .{};
+    var expected_r_batch: std.BoundedArray(Edwards25519, max) = .{};
```
InKryption: Same comment about not using BoundedArray

v2/src/common/crypto/ed25519.zig
```diff
@@ -0,0 +1,700 @@
...
+    } };
+
+    if (a.x.isZero() // first pair
+    or a.y.isZero() // second pair
+    or a.y.equivalent(y0) // third pair
+    or a.y.equivalent(y1) // fourth pair
+    ) return error.WeakPublicKey;
+}
+
+pub fn ReturnType(encoded: bool, ristretto: bool) type {
```
InKryption: Could use a better name, ie `MulMultiReturnType`.

v2/src/common/solana/hash.zig
```diff
@@ -0,0 +1,203 @@
...
+    pub fn initMany(data: []const []const u8) Hash {
+        var new = Sha256.init(.{});
+        for (data) |d| new.update(d);
+        return .{ .data = new.finalResult() };
+    }
+
+    /// re-hashes the current hash with the mixed-in byte slice(s).
+    pub fn extend(self: Hash, data: []const u8) Hash {
+        return .initMany(&.{ &self.data, data });
+    }
```
InKryption: This comment isn't quite accurate. Should probably be more along the lines of "hashes the current hash with the given bytes to produce a new one".

v2/src/common/solana/leader_schedule.zig
```diff
@@ -0,0 +1,68 @@
...
+                exp_slot.* += 1;
+            } else {
+                expect = slot + 1;
+                start_slot = slot;
+            }
+            const node_str = nextNonEmpty(&word_iter) orelse return error.MissingPubkey;
+            const node_pk = try Pubkey.parseRuntime(node_str);
+            schedule.leaders[i] = node_pk;
+            i += 1;
+        }
```
InKryption: What do you think of an approach like this:
```zig
const slot_max_len = comptime std.fmt.count("{d}", .{std.math.maxInt(Slot)});
const hash_max_len = 44;
std.debug.assert(reader.buffer.len >= @max(slot_max_len, hash_max_len));

while (true) {
    try skipSpaces(reader);
    
    const slot_str = reader.takeDelimiterExclusive(' ') catch |err| switch (err) {
        error.ReadFailed => |e| return e,
        error.EndOfStream => break,
        error.StreamTooLong => return error.InvalidSlot,
    };
    if (std.mem.findScalar(u8, slot_str, '\n') != null) return error.InvalidLine;
    const slot = std.fmt.parseInt(Slot, slot_str, 10) catch return error.InvalidSlot;
    
    try skipSpaces(reader);
    
    const node_str = reader.takeDelimiterExclusive('\n') catch |err| switch (err) {
        error.ReadFailed => |e| return e,
        error.EndOfStream => break,
        error.StreamTooLong => return error.InvalidPubkey,
    };
    const node_pk = Pubkey.parseRuntime(node_str) catch return error.InvalidPubkey;

    // -- snip -- //

    if (reader.buffer.len == 0) break; // no '\n' delimiter means end of stream.
    reader.toss(1);
}
```
```zig
fn skipSpaces(r: *std.Io.Reader) error{ReadFailed}!void {
    while (true) {
        const peeked = r.peekGreedy(1) catch |err| switch (err) {
            error.EndOfStream => return,
            error.ReadFailed => |e| return e,
        };
        const non_space_idx = std.mem.findScalar(u8, peeked, ' ');
        r.toss(non_space_idx orelse peeked.len);
        if (non_space_idx != null) return;
    }
}
```

v2/src/common/solana/pubkey.zig
```diff
@@ -0,0 +1,125 @@
...
+
+    pub fn fromPublicKey(public_key: *const std.crypto.sign.Ed25519.PublicKey) Pubkey {
+        return .{ .data = public_key.bytes };
+    }
+
+    pub fn initRandom(random: std.Random) Pubkey {
+        var bytes: [SIZE]u8 = undefined;
+        random.bytes(&bytes);
+        return .{ .data = bytes };
+    }
```
InKryption, line 32: Note: this doesn't produce a valid pubkey on the curve. Maybe it should?

v2/src/common/solana/pubkey.zig
```diff
@@ -0,0 +1,125 @@
...
+        if (decoded_len != SIZE) return error.InvalidLength;
+        return .{ .data = decoded_buf[0..SIZE].* };
+    }
+
+    pub const BASE58_MAX_SIZE = base58.encodedMaxSize(SIZE);
+    pub const Base58String = std.BoundedArray(u8, BASE58_MAX_SIZE);
+
+    pub fn base58String(self: Pubkey) Base58String {
+        return BASE58_ENDEC.encodeArray(SIZE, self.data);
+    }
```
InKryption, line 81: Should maybe avoid porting over the bounded string stuff, with BoundedArray going away and all that.

v2/src/common/solana/signature.zig
```diff
@@ -0,0 +1,98 @@
...
+        if (decoded.len != SIZE) return error.InvalidSignature;
+        return .fromBytes(decoded.constSlice()[0..SIZE].*);
+    }
+
+    pub const BASE58_MAX_SIZE = base58.encodedMaxSize(SIZE);
+    pub const Base58String = std.BoundedArray(u8, BASE58_MAX_SIZE);
+
+    pub fn base58String(self: Signature) Base58String {
+        return BASE58_ENDEC.encodeArray(SIZE, self.toBytes());
+    }
```
InKryption, line 81: Ditto

v2/src/common/ring.zig
```diff
@@ -0,0 +1,125 @@
...
+
+/// The ring that holds the packet buffers.
+pub fn Ring(N: comptime_int, T: type) type {
+    return extern struct {
+        magic: u32, // might want one at the end also?
+        head: Pos align(std.atomic.cache_line),
+        tail: Pos align(std.atomic.cache_line),
+        array: [N]T,
+
+        const MAGIC = 0xAABBCCDD;
```
InKryption: What if
```zig
        magic: enum(u32) { valid = 0xAABBCCDD, _ },
```
so that checks would just be `std.debug.assert(ring.magic == .valid)`

v2/src/common/net.zig
```diff
@@ -0,0 +1,30 @@
...
+    size: u16,
+    addr: std.net.Address,
+
+    pub const len = 1232;
+};
+
+pub const Pair = extern struct {
+    recv: Ring(MAX_PACKETS, Packet),
+    send: Ring(MAX_PACKETS, Packet),
+    port: u16,
```
kprotty, line 23: does port need to live here? IIRC its only used for the configuration of the net service rather than at runtime for the channel

v2/src/common/shred.zig
```diff
@@ -0,0 +1,2227 @@
...
+        return Generic.chainedMerkleRoot(self);
+    }
+
+    pub fn retransmitterSignature(self: Self) !Signature {
+        return Generic.retransmitterSignature(self);
+    }
+};
+
+/// Analogous to [Shred trait](https://github.com/anza-xyz/agave/blob/8c5a33a81a0504fd25d0465bed35d153ff84819f/ledger/src/shred/traits.rs#L6)
+fn GenericShred(shred_type: ShredType) type {
```
InKryption: Maybe should be called something like `ShredMethods`?

v2/src/common/shred/shred_type.zig
```diff
@@ -0,0 +1,4 @@
+pub const ShredType = enum(u8) {
```
kprotty: does this and ShredConstants.zig need to be their own files? Wondering if they could be coalesced & moved into their own folders:

In general, removing the services folder in favor of the {service-name} folder setup from before (e.g. src/shred/whatever.zig, src/crypto/ed25519.zig). It makes more sense to group service-related files together than group services themselves together.

This would also implicitly remove the common/ namespace in favor of e.g. src/gossip doing @import("src/shred/types.zig").ShredType (Gossip needs this type for DuplicateShred messages). I think this should work as long as types.zig doesnt contain any tests (which would defeat the "separate compilation" setup).

The common types could move to `src/solana` & `src/linux` as needed. Maybe even have a `src/init.zig` (and keep `src/start.zig` - dunno if it's needed to group it into the same folder as the init stuff)

v2/src/common/shred.zig
```diff
@@ -0,0 +1,2227 @@
...
+    /// aka "fec_set_index"
+    erasure_set_index: u32,
+
+    pub const bk_config: binkode.Codec(CommonHeader) = .standard(.tuple(.{
+        .leader_signature = .from(Signature.bk_config),
+        .variant = ShredVariant.bk_codec,
+        .slot = .fixint,
+        .index = .fixint,
+        .version = .fixint,
+        .erasure_set_index = .fixint,
```
InKryption, line 853: Never realized these were hard-coded to be fixints (non-actionable comment).

v2/src/common/shred.zig
```diff
@@ -0,0 +1,2227 @@
...
+};
+
+pub const ShredVariant = struct {
+    shred_type: ShredType,
+    proof_size: u8,
+    chained: bool,
+    resigned: bool,
+
+    pub const bk_codec: binkode.StdCodec(ShredVariant) = .from(.implement(void, void, struct {
+        const bk = binkode;
```
InKryption: Any reason not to just make this the main import name?

v2/src/common/ring.zig
```diff
@@ -0,0 +1,125 @@
...
+
+                pub fn into(self: *@This(), slice: []const T) void {
+                    defer self.markUsed(@intCast(slice.len));
+                    std.debug.assert(self.count == slice.len);
+
+                    const back = self.first();
+                    const front = self.second();
+                    @memcpy(back, slice[0..back.len]);
+                    @memcpy(front, slice[back.len..][0..front.len]);
+                }
```
kprotty: this sorta defeats the idea behind Slice's being zero copy. Would re-evaluate the caller.

v2/src/common/shred.zig
```diff
@@ -0,0 +1,2227 @@
...
+    };
+};
+
+pub const ShredVariant = struct {
+    shred_type: ShredType,
+    proof_size: u8,
+    chained: bool,
+    resigned: bool,
+
+    pub const bk_codec: binkode.StdCodec(ShredVariant) = .from(.implement(void, void, struct {
```
InKryption: I think it would be better to use `Codec` instead of `StdCodec`; the latter isn't really intended to be used this way. Can be composed with other StdCodecs using `.from()`.

Also: I think it would be nice to have a doc comment on this explaining that it encodes and decodes it as its single byte representation, with references to `toByte` and `fromByte`.

v2/src/common/shred.zig
```diff
@@ -0,0 +1,2227 @@
...
+            return error.InvalidShredVariant;
+        }
+        if (shard.len != try capacity(constants, common_header.variant)) {
+            return error.InvalidShardSize;
+        }
+        if (shard.len + constants.headers_size > constants.payload_size) {
+            return error.InvalidShardSize;
+        }
+
+        const payload = try allocator.alloc(u8, constants.payload_size);
```
kprotty, line 168: Now in v2, this could instead be passed into this function as a buffer to fill `payload: *[constants.payload_size]u8` instead of dynamically allocated. Removes the need for an allocator here + in the returned `Shred`

v2/src/common/shred.zig
```diff
@@ -0,0 +1,2227 @@
...
+    erasure_set_index: u32,
+
+    pub const bk_config: binkode.Codec(CommonHeader) = .standard(.tuple(.{
+        .leader_signature = .from(Signature.bk_config),
+        .variant = ShredVariant.bk_codec,
+        .slot = .fixint,
+        .index = .fixint,
+        .version = .fixint,
+        .erasure_set_index = .fixint,
+    }));
```
kprotty: could be remove `binkode` lib in favor of a simpler local solution that wuo`? Something [like this](https://gist.github.com/kprotty/1789cd5dfbf66ec151cdf57b704a259c#file-g2-zig-L309-L368) (from v2 gossip prototyping). 

In this case, the replacement would be having the `Signature` and `ShredVariant` struct contain their own `bincodeRead/bincodeWrite` functions allowing serialization to be explicit.

v2/src/common/shred.zig
```diff
@@ -0,0 +1,2227 @@
...
+            _: void,
+        ) bk.DecodeSkipError!void {
+            try reader.discardAll(value_count);
+            decoded_count.* = value_count;
+        }
+
+        pub const free = null;
+    }));
+
+    const Self = @This();
```
InKryption: I'd replace this with `ShredVariant` now

v2/src/services/net.zig
```diff
@@ -0,0 +1,91 @@
...
+};
+
+pub fn serviceMain(writer: *std.io.Writer, rw: ReadWrite) !noreturn {
+    try mainInner(writer, &.{rw.pair});
+}
+
+const MAX_SOCKETS = 10;
+
+/// `ports` is the list of ports it'll listen on.
+fn mainInner(writer: *std.io.Writer, pairs: []const *Pair) !noreturn {
```
kprotty: Just realized, this is more like the "udp" tile rather than the "net" tile. Theres some random TCP functionality across the validator that is technically "networking" but doesnt flow through this atm. Some examples:
* Echo server to get our public ip (needed by Gossip)
* Snapshot downloading over http (will be needed by snapshot tile)
* maybes: (RPC? tracing/prometheus? geyser?)

v2/src/services/net.zig
```diff
@@ -0,0 +1,91 @@
...
+    for (pairs, 0..) |pair, i| {
+        errdefer for (sockets[0..i]) |socket| std.posix.close(socket);
+        const socket = try std.posix.socket(
+            std.posix.AF.INET,
+            std.posix.SOCK.DGRAM | std.posix.SOCK.NONBLOCK,
+            0,
+        );
+        errdefer std.posix.close(socket);
+
+        try writer.print("(net)binding 0.0.0.0:{}\n", .{pair.port});
```
kprotty: Thoughts on a custom logger type that namespaces + newlines this for us (giving its for printing)? e.g `logger: *sig.log(.net)`

Mostly a nice-to-have

v2/src/services/net.zig
```diff
@@ -0,0 +1,91 @@
...
+
+    var sockets: [MAX_SOCKETS]std.posix.fd_t = undefined;
+    var sockets_len: usize = 0;
+
+    for (pairs, 0..) |pair, i| {
+        errdefer for (sockets[0..i]) |socket| std.posix.close(socket);
+        const socket = try std.posix.socket(
+            std.posix.AF.INET,
+            std.posix.SOCK.DGRAM | std.posix.SOCK.NONBLOCK,
+            0,
```
kprotty: can do `std.posix.IPPROTO.UDP` (its functionally the same though)

v2/src/services/net.zig
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    while (true) {
+        // send
+        for (pairs, sockets[0..sockets_len]) |pair, sock| {
+            var slice = pair.send.getReadable() catch continue;
+            const flags = std.posix.MSG.NOSIGNAL;
+            var counter: u32 = 0;
+            inline for (.{ slice.first(), slice.second() }) |packets| {
+                for (packets) |p| {
```
kprotty, line 59: multiple packets here could use `std.os.linux.sendmmsg` (send multiple UDP packets in same syscall). Just as a small opt (feel free to ignore)

v2/src/services/net.zig
```diff
@@ -0,0 +1,91 @@
...
+            }
+            slice.markUsed(counter);
+        }
+
+        // recv
+        for (pairs, sockets[0..sockets_len]) |pair, sock| {
+            var slice = pair.recv.getWritable() catch continue;
+            const ptr = slice.one();
+            var dummy: u32 = 0;
+            ptr.size = @intCast(std.posix.recvfrom(
```
kprotty, line 78: Since the socket is non-blocking, it should try and receive as many packets into the pair's `getWritable()` before moving on to the next pair/sock. (Maybe also use `std.os.linux.recvmmsg` here too).

v2/src/services/shred_receiver.zig
```diff
@@ -0,0 +1,234 @@
...
+    }
+};
+
+fn validateShred(
+    packet: *const Packet,
+    root: Slot,
+    shred_version: *const Atomic(u16),
+    max_slot: Slot,
+) ShredValidationError!void {
+    const packet_shred = layout.getShred(packet, false) orelse return error.insufficient_shred_size;
```
kprotty: y the errors snake case. Who snaked you

v2/src/common.zig
```diff
@@ -0,0 +1,110 @@
...
+    ro: [max_regions]?[*]align(page_size_min) const u8,
+    ro_len: [max_regions]usize,
+
+    pub const max_regions = 4; // chosen arbitrarily
+    const page_size_min = std.heap.page_size_min;
+};
+
+pub const ServiceFn = *const fn (ResolvedArgs) callconv(.c) void;
+
+/// This value should be written to before a service exits. Both fields may be active at once.
```
InKryption: What does "both fields" refer to here? Is this comment outdated?

v2/src/main.zig
```diff
@@ -0,0 +1,92 @@
...
+
+    const ShredNetwork = struct {
+        recv_port: u16,
+    };
+};
+
+pub fn main() !void {
+    var gpa = std.heap.GeneralPurposeAllocator(.{}){};
+    defer _ = gpa.deinit();
+    const allocator = gpa.allocator();
```
InKryption:
```suggestion
    var dba_state: std.heap.DebugAllocator(.{}) = .init;
    defer _ = dba_state.deinit();
    const gpa = dba_state.allocator();
```

v2/src/common/shred/shred_type.zig
Sobeston: > does this and ShredConstants.zig need to be their own files? 

Not really, they could be moved into shred.zig. Initially wanted to split up shred.zig, but might not make the most sense

> It makes more sense to group service-related files together 

I've got it in common as there's probably multiple services that need to use this code

v2/src/common/shred.zig
Sobeston: If we could do it with a packed struct that might be ideal?

For porting from v1 to v2, I'm just preferring binkode over our older sig bincode

v2/src/main.zig
Sobeston: https://github.com/Syndica/sig/pull/1204/commits/ee0bbe368bd9b212c5ea962464d1f79941ba95ce

v2/src/common/shred.zig
Sobeston: https://github.com/Syndica/sig/pull/1204/commits/9aba909a4242aff69f9ef025604bc400b4f3762c

v2/src/common/shred.zig
Sobeston: https://github.com/Syndica/sig/pull/1204/commits/9aba909a4242aff69f9ef025604bc400b4f3762c

v2/src/common/shred.zig
Sobeston: https://github.com/Syndica/sig/pull/1204/commits/9aba909a4242aff69f9ef025604bc400b4f3762c

v2/src/common/shred.zig
Sobeston: https://github.com/Syndica/sig/pull/1204/commits/9aba909a4242aff69f9ef025604bc400b4f3762c

v2/src/common.zig
Sobeston: https://github.com/Syndica/sig/pull/1204/commits/fe81528715465a3b6f5071e1dc5706f9b5a4236c

v2/src/services/net.zig
Sobeston: https://github.com/Syndica/sig/pull/1204/commits/da114cab1ea2fac1532e08534cc08d0a153aee04

v2/src/common/net.zig
Sobeston: Not really? I guess the side that isn't the net service doesn't necessarily need to know what the port is.

This is mostly just a consequence of how we pass values into services

thought: in theory a malicious service could edit the port value before the net service reads it, which would compromise things a little

v2/src/common/shred.zig
Sobeston: https://github.com/Syndica/sig/commit/9aba909a4242aff69f9ef025604bc400b4f3762c

v2/src/services/shred_receiver.zig
Sobeston: unsnaked https://github.com/Syndica/sig/pull/1204/commits/8ef789df29abd724d94c2f3f20cc8cb9c7c953d0

v2/src/common/solana/hash.zig
Sobeston: https://github.com/Syndica/sig/pull/1204/commits/c89d740a970ae6dc5ad4772a240b571743da2f0e

v2/src/common/solana/pubkey.zig
Sobeston: I think this is a valid fix https://github.com/Syndica/sig/pull/1204/commits/73e22aada212dd56c029f1e2766280da9cb7821b

v2/src/common/ring.zig
Sobeston: https://github.com/Syndica/sig/pull/1204/commits/260fa445901a86dbc7fe9eb7423a391e568c8123

v2/src/services/net.zig
Sobeston: Yeah that's fair, it might end up doing UDP only.

v2/src/services/net.zig
Sobeston: Preferring to ignore net optimisations for now

v2/src/services/net.zig
Sobeston: Preferring to ignore net optimisations for now

v2/src/common/crypto/ed25519/pippenger.zig
Sobeston: https://github.com/Syndica/sig/pull/1204/changes/c8d262cd826625dfff4bc6a68cc68ddf3de18f3f

v2/src/common/crypto/ed25519/straus.zig
Sobeston: https://github.com/Syndica/sig/pull/1204/changes/c8d262cd826625dfff4bc6a68cc68ddf3de18f3f

v2/src/common/crypto/ed25519/straus.zig
Sobeston: https://github.com/Syndica/sig/pull/1204/changes/c8d262cd826625dfff4bc6a68cc68ddf3de18f3f

v2/src/common/crypto/ed25519.zig
Sobeston: https://github.com/Syndica/sig/pull/1204/changes/c8d262cd826625dfff4bc6a68cc68ddf3de18f3f

v2/src/common/crypto/ed25519.zig
Sobeston: https://github.com/Syndica/sig/pull/1204/changes/c8d262cd826625dfff4bc6a68cc68ddf3de18f3f

v2/src/common/solana/leader_schedule.zig
Sobeston: https://github.com/Syndica/sig/pull/1204/commits/650133962c15ac5deade62b79b70b6944b6e3497

v2/src/common/solana/pubkey.zig
Sobeston: https://github.com/Syndica/sig/pull/1204/changes/c8d262cd826625dfff4bc6a68cc68ddf3de18f3f

v2/src/common/solana/signature.zig
Sobeston: https://github.com/Syndica/sig/pull/1204/changes/c8d262cd826625dfff4bc6a68cc68ddf3de18f3f

v2/src/common/shred.zig
```diff
@@ -0,0 +1,2215 @@
...
+    };
+    const CustomHeader = switch (shred_type) {
+        .data => DataHeader,
+        .code => CodeHeader,
+    };
+    const constants = Self.constants;
+
+    return struct {
+        fn fromPayload(payload: []const u8) !Self {
+            // NOTE(x19): is it ok if payload.len > constants.payload_size? the test_data_shred is 1207 bytes
```
yewman, line 405: Yes, truncation is fine. See https://github.com/anza-xyz/agave/blob/2717084afeeb7baad4342468c27f528ef617a3cf/ledger/src/shred/merkle.rs#L540-L545

Lets remove this note

v2/src/common/shred/shred_type.zig
yewman: Shred also seems to me like a solana specific type which could be placed in src/common/solana/shred.zig and could include the shred constants and shred type. Just a thought

dnut: > Shred also seems to me like a solana specific type which could be placed in src/common/solana/shred.zig and could include the shred constants and shred type. Just a thought

I would only do that if there are other services that need to use it.

v2/src/services/net.zig
Sobeston: https://github.com/Syndica/sig/pull/1204/commits/8f3936963a04b5880ab7ca309dad3ca0ac5fcf28

v2/src/services/net.zig
Sobeston: https://github.com/Syndica/sig/pull/1204/commits/8f3936963a04b5880ab7ca309dad3ca0ac5fcf28

v2/src/common/shred.zig
Sobeston: Going to make this an extern struct as part of a follow-up

v2/src/common/shred.zig
Sobeston: Going to replace this code in a follow-up

v2/src/common/shred/shred_type.zig
Sobeston: I think generally we should opt for increasing the number of modules we have in future.

Splitting away most of common makes sense to me, e.g. making ed25519 its own module makes sense. Shrinking common also has compile time benefits.

I think a Shred module would make sense, which would then be imported by any services which handle shreds (in this case shred receiver, maybe a future shred sender, or something like that). 

For now I know that this code will be shared between services so I've put it in common. I've removed the extra files https://github.com/Syndica/sig/pull/1204/commits/14ce530b3b6b5f4c4b1c30d4fcb5cd18c926a8c6.

fwiw, most of this shred code is being rewritten as part of fec set resolution.

v2/src/common/ring.zig
Sobeston: removed https://github.com/Syndica/sig/pull/1204/commits/7479d3bd69bf0f9640deea5f31de7a431b38a75f

v2/build.zig
```diff
@@ -0,0 +1,145 @@
...
+                    .name = "services",
+                    .module = b.createModule(.{ .root_source_file = b.path("src/services.zon") }),
+                },
+            },
+        }),
+    });
+    const validate_services_list_run = b.addRunArtifact(validate_services_list_exe);
+    validate_services_list_run.addDirectoryArg(b.path("src/services"));
+    b.getInstallStep().dependOn(&validate_services_list_run.step);
+}
```
kprotty, line 145: Could inline the service list here instead of the zon file. Would remove the need for the zon file + the validate_service_list script.

v2/src/common/shred.zig
```diff
@@ -0,0 +1,2223 @@
...
+
+pub const MAX_SHREDS_PER_SLOT: usize =
+    CodeShred.constants.max_per_slot +
+    DataShred.constants.max_per_slot;
+
+pub const DATA_SHREDS_PER_FEC_BLOCK: usize = 32;
+const SIZE_OF_MERKLE_ROOT: usize = Hash.SIZE;
+
+/// Analogous to [Shred](https://github.com/anza-xyz/agave/blob/8c5a33a81a0504fd25d0465bed35d153ff84819f/ledger/src/shred.rs#L245)
+pub const Shred = union(ShredType) {
```
dnut, line 27: Does this need to be in common? There's a good chance this won't be used outside the shred service. I'd rather keep things scoped in services and only move them into common when it's clearly needed by multiple services.

v2/src/common/solana.zig
```diff
@@ -0,0 +1,15 @@
+const std = @import("std");
+
+test {
+    _ = std.testing.refAllDecls(@This());
+}
```
dnut: Is this going to be needed for every namespace that contains tests? I get that refAllDeclsRecursive isn't perfect, but I also don't like needing to add this to every file.

Also, I don't think you need to put this in a `test {}` block. This makes the test counts confusing because it adds another unit test for every instance of `test {}`. It's not a big problem but you could just put it in a comptime block instead to achieve the same thing. `refAllDecls` is a noop if it's not a test build.

```suggestion
comptime {
    std.testing.refAllDecls(@This());
}
```

v2/src/common/solana/leader_schedule.zig
```diff
@@ -0,0 +1,84 @@
+//! Must be replaced - cannot handle epoch transitions at all
+
+const common = @import("../../common.zig");
+
+const std = @import("std");
+
+const Slot = common.solana.Slot;
+const Pubkey = common.solana.Pubkey;
```
dnut:
```suggestion
const std = @import("std");

const solana = @import("../solana.zig");

const Slot = solana.Slot;
const Pubkey = solana.Pubkey;
```

The code is more modular if it only reaches out into the nearest parent scope that's necessary to get its dependencies. We shouldn't make every file depend on the overall structure of common if it's not necessary. It'll be easier to refactor the code if things are more tightly scoped, and I don't see a downside to it.

v2/src/main.zig
```diff
@@ -0,0 +1,99 @@
...
+        defer cfg_file.close();
+
+        const cfg_str = try cfg_file.readToEndAllocOptions(allocator, 1024 * 1024, null, .@"1", 0);
+        defer allocator.free(cfg_str);
+
+        var diag: std.zon.parse.Diagnostics = .{};
+        defer diag.deinit(allocator);
+
+        break :cfg std.zon.parse.fromSlice(Config, allocator, cfg_str, &diag, .{}) catch |err| {
+            std.debug.print("{f}\n", .{diag});
```
dnut: should this be std.log?

v2/src/main.zig
```diff
@@ -0,0 +1,99 @@
...
+                .{ .instance = .{ .service = .net }, .rw = true },
+            },
+        },
+        .{
+            .region = .{ .leader_schedule = .{ .schedule_string = &reader.interface } },
+            .shares = &.{
+                .{ .instance = .{ .service = .shred_receiver } },
+            },
+        },
+    };
```
dnut, line 85: It's not in scope right now but I'm curious on your thoughts. It seems like the types and numbers of expected shares per service could actually be partly inferred from each serviceMain's parameters and validated at comptime against these shares.

v2/src/services.zig
```diff
@@ -0,0 +1,648 @@
...
+
+const linux = std.os.linux;
+const sigaction_fn = linux.Sigaction.sigaction_fn;
+const memfd = common.linux.memfd;
+const E = linux.E;
+const e = E.init;
+
+pub const Service = enum {
+    net,
+    shred_receiver,
```
dnut, line 29: Why don't we just call this `shred` since it'll be doing all shred operations, not just receiving.

v2/src/common/ring.zig
```diff
@@ -0,0 +1,115 @@
+const std = @import("std");
+
+test {
+    _ = std.testing.refAllDecls(@This());
+}
+
+const Atomic = std.atomic.Value;
+
+/// The ring that holds the packet buffers.
+pub fn Ring(N: comptime_int, T: type) type {
```
dnut, line 10: We need some more documentation on this struct, particularly the Slice methods, but docs all around would be helpful. @kprotty

v2/src/common/ring.zig
```diff
@@ -0,0 +1,115 @@
...
+    return extern struct {
+        magic: Magic, // might want one at the end also?
```
dnut: is this supposed to be a test to see if you have memory corruption?

v2/src/common/shred.zig
```diff
@@ -0,0 +1,2223 @@
...
+    const SIZE_OF_VERSION: usize = 2;
+
+    const OFFSET_OF_SHRED_VARIANT: usize = SIZE_OF_SIGNATURE; // 64
+    const OFFSET_OF_SHRED_SLOT: usize = SIZE_OF_SIGNATURE + SIZE_OF_SHRED_VARIANT; // 64 + 1 = 65
+    const OFFSET_OF_SHRED_INDEX: usize = OFFSET_OF_SHRED_SLOT + SIZE_OF_SHRED_SLOT; // 65 + 8 = 73
+    const OFFSET_OF_ERASURE_SET_INDEX: usize =
+        OFFSET_OF_SHRED_INDEX + SIZE_OF_INDEX + SIZE_OF_VERSION;
+
+    pub fn getShredId(packet: *const Packet) !ShredId {
+        const shred = getShred(packet) orelse return error.InvalidShred;
```
dnut: missing `is_repair` param

v2/src/common/solana/pubkey.zig
```diff
@@ -0,0 +1,142 @@
...
+
+    pub fn format(self: *const Pubkey, writer: *std.Io.Writer) std.Io.Writer.Error!void {
+        var buf: [BASE58_MAX_SIZE]u8 = undefined;
+        const str = self.base58String(&buf);
+        return writer.writeAll(str);
+    }
+
+    pub fn jsonStringify(self: Pubkey, write_stream: anytype) !void {
+        try write_stream.write(self.base58String().slice());
```
dnut: missing `buffer` param

Maybe we shouldn't be adding dead code into v2.

v2/src/services.zig
```diff
@@ -0,0 +1,648 @@
...
+    // mapped in children
+    for (exit_meta.items(.exit), exit_memfds) |*exit, exit_fd| {
+        exit.* = @ptrCast(try exit_fd.mmap(null));
+    }
+
+    // Wait for the first child to exit
+    var status: u32 = 0;
+    const exited_pid: i32 = pid: {
+        const ret: usize = linux.waitpid(-1, &status, 0);
+        std.debug.assert(ret != -1);
```
dnut: this is a tautology for usize.

is this more meaningful?
```suggestion
        std.debug.assert(ret != std.math.maxInt(usize));
```

or should we be checking `e(ret) != .SUCCESS`?

v2/src/start_service.zig
```diff
@@ -0,0 +1,364 @@
...
+
+pub const panic = std.debug.FullPanic(servicePanic);
+
+pub const options: std.Options = .{
+    .logFn = serviceLog,
+};
+
+pub const panic_state = struct {
+    pub var stderr: std.os.linux.fd_t = undefined;
+    pub var exit: *common.Exit = undefined;
```
dnut, line 55: is it safe to set these to undefined? what if something tries to use one of these before they are set?

v2/src/common/shred.zig
```diff
@@ -0,0 +1,2223 @@
...
+    }
+
+    pub fn sanitize(self: MerkleProofEntryList) !void {
+        if (self.len * merkle_proof_entry_size != self.bytes.len) {
+            return error.InvalidMerkleProof;
+        }
+    }
+
+    pub fn get(self: *const MerkleProofEntryList, index: usize) ?MerkleProofEntry {
+        if (index > self.len) return null;
```
dnut: pre-existing bug

```suggestion
        if (index >= self.len) return null;
```

v2/src/common/linux/bpf.zig
```diff
@@ -0,0 +1,120 @@
+const std = @import("std");
+
+test {
+    _ = std.testing.refAllDecls(@This());
+}
+
+const SECCOMP = std.os.linux.SECCOMP;
+const syscalls = std.os.linux.syscalls.X64;
```
dnut:
```suggestion
const syscalls = std.os.linux.SYS;
```

v2/src/services.zig
```diff
@@ -0,0 +1,648 @@
...
+
+fn closeAllFdsExceptStderr(maybe_stderr: ?linux.fd_t) void {
+    const max_fd = std.math.maxInt(linux.fd_t);
+
+    if (maybe_stderr) |stderr| {
+        // close (0..stderr, stderr+1..=max)
+        if (std.os.linux.syscall3(.close_range, 0, @intCast(stderr - 1), 0) != 0)
+            std.debug.panic("close_range failed\n", .{});
+        if (std.os.linux.syscall3(.close_range, @intCast(stderr + 1), max_fd, 0) != 0)
+            std.debug.panic("close_range failed\n", .{});
```
dnut: not that i would expect stderr to actually reach numbers this high or low, but saturating is equally correct and theoretically safer.
```suggestion
        if (std.os.linux.syscall3(.close_range, 0, @intCast(stderr -| 1), 0) != 0)
            std.debug.panic("close_range failed\n", .{});
        if (std.os.linux.syscall3(.close_range, @intCast(stderr +| 1), max_fd, 0) != 0)
            std.debug.panic("close_range failed\n", .{});
```

v2/src/common/shred.zig
yewman: Keeping services lean and focused on the data structures and algorithms which facilitate their higher level logic is preferable. A specific data structure being used by only a single service is not a good justification for its implementation being defined within that service. For example we would not implement the zksdk inside the zk_elgamal program simply because that is the only place it is used at the moment. Instead the zk_elgamal program defines only its instruction type and execution function which together define its 'higher level logic' (i.e. what is it doing).

dnut: Let's consider a hypothetical data structure that defines a domain-specific concept which is only meaningful in the one very narrow context that is fully implemented by a single service. The concept has no relevance in any other context outside that service, and it never well. It's not general purpose code that will likely to be useful in any other context. In that case, it makes no sense for this data structure to exist in a common library. Can we agree on this much? My claim is that `Shred` follows this pattern.

v2/build.zig
InKryption: The point of the script & the list is to verify that everything in the `services/` folder is being build script'd. This used to be a `Dir.iterate`, but not allowed to do that in build.zig.

v2/src/common/shred.zig
InKryption: I would tend to agree with Drew's original comment here. I think the `common` folder should be reserved for "header declarations", ie, things that are needed, either for re-use, or to expose the types that represent the memory regions of the associated service.

v2/src/services/net.zig
```diff
@@ -0,0 +1,92 @@
...
+                }
+            }
+            slice.markUsed(counter);
+        }
+
+        // recv
+        for (pairs, sockets[0..sockets_len]) |pair, sock| {
+            var slice = pair.recv.getWritable() catch continue;
+            const ptr = slice.one();
+            var dummy: u32 = 0;
```
yewman: Since `dummy` (addrlen) is zero, `recvfrom` will not write to ptr.addr.any. If this is intended, why not just pass null for the src_addr and addrlen args?

v2/src/services.zig
```diff
@@ -0,0 +1,648 @@
...
+    }
+
+    var map = try serviceMap(allocator, services, regions, region_memfds);
+    defer {
+        for (map.values()) |*value| value.deinit(allocator);
+        map.deinit(allocator);
+    }
+
+    var reset_event: std.Thread.ResetEvent = .{};
+    var finished_service_idx: ?u16 = null;
```
yewman: Data race on setting this in the event of simultaneous service exits?

v2/src/start_service.zig
Sobeston: This is immediately set on startup

v2/src/main.zig
Sobeston: I do think that we can do better at defining the topology, but we can do this in a follow up

v2/src/services.zig
Sobeston: We can rename it to `shred` when it does more than receiving

v2/src/common/shred.zig
Sobeston: Leaving this until we decide upon a dir layout

v2/src/common/ring.zig
Sobeston: Yes + that it was initialised
cc @kprotty

v2/src/services.zig
Sobeston: Hm, I guess so.

(The no-sandbox code isn't intended for production use)

However if two threads exited at the same time, either way we would only dump the info of one of them; whether we dump out the "1st" or "2nd"  doesn't matter. Unless I'm misunderstanding?

v2/src/services/net.zig
Sobeston: Passing null now https://github.com/Syndica/sig/pull/1204/commits/b0cc617812eaa8faad1889ea60e22ee6cf55e007

v2/src/common/solana/pubkey.zig
Sobeston: removed https://github.com/Syndica/sig/pull/1204/commits/bba94c1d5eb06aca5226003c64bcfd975bab617b

v2/src/common/shred.zig
Sobeston: removed function https://github.com/Syndica/sig/pull/1204/commits/bba94c1d5eb06aca5226003c64bcfd975bab617b

v2/src/services.zig
Sobeston: https://github.com/Syndica/sig/pull/1204/changes/7a3a20a16e742f7cb5cbc0ee23033c738b06df36

v2/src/common/linux/bpf.zig
Sobeston: https://github.com/Syndica/sig/pull/1204/changes/7a3a20a16e742f7cb5cbc0ee23033c738b06df36

v2/src/services.zig
Sobeston: https://github.com/Syndica/sig/pull/1204/changes/7a3a20a16e742f7cb5cbc0ee23033c738b06df36

v2/src/common/shred.zig
Sobeston: https://github.com/Syndica/sig/pull/1204/changes/7a3a20a16e742f7cb5cbc0ee23033c738b06df36

v2/src/main.zig
Sobeston: https://github.com/Syndica/sig/pull/1204/commits/b9cad3cf62ecab7dfdc5fad32578b129ce2aa6fc

v2/src/common/solana/leader_schedule.zig
Sobeston: I don't think this really matters, I couldn't find other instances of this

but fixed in https://github.com/Syndica/sig/pull/1204/commits/9faf77bc9c21bb6f15c0d944d9590fbd9bd06437

v2/src/common/solana.zig
Sobeston: https://github.com/Syndica/sig/pull/1204/commits/7a3a20a16e742f7cb5cbc0ee23033c738b06df36

> I get that refAllDeclsRecursive isn't perfect

I would much rather not use refAllDecls at all than to reach for Recursive, this stuff is not good

dnut: how do you run all the tests without refAllDecls?

v2/src/services.zig
yewman: Yeah I see your point, is there a cost to making it atomic since it is technically UB? Happy for you to make the call here and resolve this comment if you see fit. Just wanted to raise as a potential issue

v2/src/common/solana.zig
Sobeston: https://github.com/ghostty-org/ghostty/blob/c3febabd286dfaa47c3d7d251ede22e8e382b6f3/src/main_ghostty.zig#L181

we could do it similar to this, where we just import all the files

v2/src/services.zig
Sobeston: True, might as well https://github.com/Syndica/sig/pull/1204/commits/77505393344bca551b01d02d83cdc8c049c802f8

v2/src/common/solana.zig
dnut: is there an advantage of this approach? seems like yet another annoying thing to maintain that is likely to get mishandled at some point.

dnut: feel free to resolve this conversation

v2/src/common/ring.zig
kprotty: I dont think `magic` here is needed. For corruption, the best thing to do would be detect it on a per-element basis e.g. with checksumming. `Pos` corruption would primarily end up either hitting the asserts in `getReadable/getWritable`, or reading invalid items (resolved by checksum if done)

v2/src/start_service.zig
dnut: my main concern was with a panic occurring right at the very beginning of service startup, even before this is set. in general there are a lot of things that can implicitly panic. i don't see anything specific that concerns me, but it's more about theoretical safety and protecting against future changes that might violate our current assumptions

i was thinking that stderr and exit could be set to default values other than undefined.

stderr could start out as `2` as a defensive value, and then be changed to the desired value on startup. or a more offensive approach would be to start with a predefined sentinel value that indicates "unset" and panic if you attempt to use that value.

exit could be an optional value that's null by default and the servicePanic implementation falls back to the default panic implementation if it's null.

it's not important, so if there are downsides to this approach, i'm fine keeping it as is. i just thought we could make this code a bit more theoretically sound without any drawbacks.

v2/src/common/ring.zig
kprotty: https://github.com/Syndica/sig/pull/1204/commits/d0200c3b20c37f5c8b2a22bb808c983e25cc7473

v2/src/common/ring.zig
kprotty: https://github.com/Syndica/sig/pull/1204/commits/36c759ae3aa875e364d6b9ca0a2968923e9d4f7f

v2/src/start_service.zig
Sobeston: > it's more about theoretical safety and protecting against future changes that might violate our current assumptions

If someone writes a service incorrectly, it will not trigger a panic in this code - only a mistake in startup could.

If this goes wrong it doesn't cause a *safety* issue - it will just result in a process exit without helpful information. 

>  and panic if you attempt to use that value.

If a service tries to panic before this is set, it cannot meaningfully panic, it will just crash

PR #1205, opened 2026-02-03
feat(rpc): implement getGenesisHash RPC method

ajw221:
### Intent

- Implement `getGenesisHash` RPC method

### Implementation

- Updated `GenesisConfig` to store its precomputed hash
- Updated `Hooks.set` to be able to take a pointer to a value
- Created `StaticHookContext` for implementing any future RPC methods containing static data

### Ramifications

- The approach to storing `genesis_hash` on `StaticHookContext` currently only works if a pointer to the rpc hook is given, otherwise if an owned value for the rpc hook is given, an integer overflow will occur on method call due to the passed in value becoming a dangling slice. Because the lifetime of the value lives for the entire runtime of the RPC server this shouldn't be an issue.

### Tests

- Added asserts to `GenesisConfig` unit tests around ensuring each precomputed hash matches agave's values.

### DIFF
--- a/src/rpc/methods.zig
+++ b/src/rpc/methods.zig
@@ -385,25 +385,6 @@ pub const GetGenesisHash = struct {
         ) !void {
             try jw.write(self.hash.base58String().slice());
         }
-
-        test "jsonStringify outputs base58 string" {
-            const allocator = std.testing.allocator;
-
-            const slice = "4sGjMW1sUnHzSxGspuhpqLDx6wiyjNtZAMdL4VZHirAn";
-            const expected_slice = try std.fmt.allocPrint(allocator, "\"{s}\"", .{slice});
-            defer allocator.free(expected_slice);
-
-            // Use a known hash value for testing
-            const hash = sig.core.Hash.parse(slice);
-            const response: GetGenesisHash.Response = .{ .hash = hash };
-
-            // Serialize the response to JSON
-            const json = try std.json.stringifyAlloc(allocator, response, .{});
-            defer allocator.free(json);
-
-            // The jsonStringify should output the hash as a base58 string (without wrapping object)
-            try std.testing.expectEqualStrings(expected_slice, json);
-        }
     };
 };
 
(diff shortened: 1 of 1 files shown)


src/core/genesis_config.zig
```diff
@@ -290,12 +290,28 @@ pub const GenesisConfig = struct {
...
+        const file_bytes = try file.readToEndAlloc(allocator, std.math.maxInt(usize));
+        defer allocator.free(file_bytes);
+
+        // Compute hash from original file bytes (this matches Agave's behavior)
+        var hash_bytes: [32]u8 = undefined;
+        std.crypto.hash.sha2.Sha256.hash(file_bytes, &hash_bytes, .{});
+
+        // Parse the genesis config from the bytes
+        var fbs = std.io.fixedBufferStream(file_bytes);
+        var config = try bincode.read(allocator, GenesisConfig, fbs.reader(), .{});
```
Rexicon226: use `readFromSlice`

src/cmd.zig
```diff
@@ -1174,6 +1174,12 @@ fn validator(
     );
     defer loaded_snapshot.deinit();
 
+    const static_rpc_ctx: sig.rpc.methods.StaticHookContext = .{
+        .genesis_hash = loaded_snapshot.genesis_config.hash.base58String(),
```
Rexicon226: Don't copy around the bounded array, just store the hash and do `base58String()` in `getGensisHash`

src/core/genesis_config.zig
ajw221: [40fb3ca](https://github.com/Syndica/sig/pull/1205/commits/40fb3ca953dc0dd391b30e1a701c255f2ef7d26b)

src/cmd.zig
ajw221: [9fcae97](https://github.com/Syndica/sig/pull/1205/commits/9fcae97e9e1307feaff84be872294d75bc17c4e4)

yewman (dismissed):

src/rpc/methods.zig
```diff
@@ -374,14 +374,39 @@ pub const GetEpochSchedule = struct {
...
+            const hash = sig.core.Hash.parse(slice);
+            const response: GetGenesisHash.Response = .{ .hash = hash };
+
+            // Serialize the response to JSON
+            const json = try std.json.stringifyAlloc(allocator, response, .{});
+            defer allocator.free(json);
+
+            // The jsonStringify should output the hash as a base58 string (without wrapping object)
+            try std.testing.expectEqualStrings(expected_slice, json);
+        }
```
InKryption: Would prefer if we put this test outside the struct, and at the root namespace of the file, like all other tests. Rationale:
1. Nesting tests like this may become unsupported in the future, depending on how the zig core team decide to re-design test discovery.
2. This will become an undiscoverable test without `refAllDeclsRecursive`, or without manually referencing the Response struct in a comptime or other test block - the former is going to be removed, and the latter just isn't very nice.

I personally don't have any preference for where the test is situated wrt other code, but the convention seems to usually be somewhere near the bottom of the file.

ajw221: [072f762](https://github.com/Syndica/sig/pull/1205/commits/072f762a26db7989f51bf9f18691ce19592d3689)

InKryption (approved):

yewman (approved):

PR #1206, opened 2026-02-03
refactor(cluster): remove duplicate cluster types

yewman:
# Summary 
- Remove duplicate cluster types -> Consolidate to genesis_config.ClusterType
- Panics when getting entrypoints or RPC urls for 'development' cluster type. 
    -  When running with 'development' type we should provide entrypoints and rpc urls through cli or config that make sense instead of relying on inline defaults 
- Remove LocalHost, and Custom cluster types since they are unused. These should be replaced by using 'development' type with appropriate cli or config.

### DIFF
--- a/src/cmd.zig
+++ b/src/cmd.zig
@@ -1366,7 +1366,7 @@ fn ensureGenesis(
 
     // Otherwise, download genesis from network
     logger.info().logf("Downloading genesis from {s} cluster...", .{@tagName(cluster)});
-    const cluster_url = cluster.getRpcUrl();
+    const cluster_url = cluster.getRpcUrl() orelse @panic("No RPC Url!");
     const genesis_path = downloadAndExtractGenesis(
         allocator,
         cluster_url,
@@ -1612,7 +1612,8 @@ fn validator(
     defer epoch_tracker.deinit(allocator);
 
     const rpc_cluster_type = loaded_snapshot.genesis_config.cluster_type;
-    var rpc_client = try sig.rpc.Client.init(allocator, rpc_cluster_type, .{});
+    const rpc_url = rpc_cluster_type.getRpcUrl() orelse @panic("No RPC Url for cluster type!");
+    var rpc_client = try sig.rpc.Client.init(allocator, rpc_url, .{});
     defer rpc_client.deinit();
 
     const turbine_config = cfg.turbine;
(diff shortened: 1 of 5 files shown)


src/core/genesis_config.zig
```diff
@@ -252,15 +252,48 @@ pub const Inflation = struct {
 
 /// Analogous to [ClusterType](https://github.com/anza-xyz/agave/blob/cadba689cb44db93e9c625770cafd2fc0ae89e33/sdk/src/genesis_config.rs#L46)
```
Rexicon226: Could you update this link to the new place
https://github.com/anza-xyz/solana-sdk/blob/a467058aabc453c7d749a4993c56df293d1d75c3/cluster-type/src/lib.rs#L19

yewman: 66c0af428cb91

src/cmd.zig
```diff
@@ -2147,18 +2122,12 @@ fn printLeaderSchedule(allocator: std.mem.Allocator, cfg: config.Cmd) !void {
...
-        const cluster_type: ClusterType = switch (cluster) {
-            .mainnet => .MainnetBeta,
-            .devnet => .Devnet,
-            .testnet => .Testnet,
-            .localnet => .LocalHost,
-        };
-
         var rpc_client = try sig.rpc.Client.init(allocator, cluster_type, .{});
         defer rpc_client.deinit();
+        defer rpc_client.deinit();
```
kprotty, line 2159: double free?

yewman: lol, it doesn't get much more blatant than that haha cffcdb3d7

src/core/genesis_config.zig
```diff
@@ -250,17 +250,50 @@ pub const Inflation = struct {
...
-        url: []const u8,
-    },
+pub const ClusterType = enum(u8) {
+    testnet = 0,
+    mainnet = 1,
+    devnet = 2,
+    development = 3,
+
+    /// Returns entrypoints for public clusters, null for development.
+    /// For development, the caller must provide entrypoints manually.
```
InKryption:
```suggestion
    /// For development this returns an empty list, because the caller
    /// must provide entrypoints manually.
```

src/core/genesis_config.zig
```diff
@@ -250,17 +250,50 @@ pub const Inflation = struct {
...
+
+    /// Returns the RPC URL for this cluster.
+    pub fn getRpcUrl(self: ClusterType) []const u8 {
+        return switch (self) {
+            .mainnet => "https://api.mainnet-beta.solana.com",
+            .testnet => "https://api.testnet.solana.com",
+            .devnet => "https://api.devnet.solana.com",
+            .development => @panic("Not available for development cluster"),
+        };
+    }
```
InKryption, line 297: I don't have a strong opinion on this, but maybe this should return an optional? Otherwise a caller may just have to switch on the cluster type themselves before calling this function, which ends up being a bit redundant. Can move the panic to the callsite, maybe even with a more specific message.

yewman: Yeah I agree. I have changed it to optional and also changed the RpcClient to be initialised with an rpc url instead of a cluster type. I have left plumbing in a custom rpc url to the rpc client for development clusters as remaining work but at least it will fail clearly and can be addressed when required. 
790946d6

src/core/genesis_config.zig
yewman: 790946d6

InKryption (dismissed):

kprotty (approved):

hamza-syndica (approved):

PR #1208, opened 2026-02-04
feat(rpc): implement getVoteAccounts RPC

prestonsn:
### Intent
- Implement`getVoteAccounts` RPC method
### Implementation
- Added `getVoteAccounts` handler to `RpcHookContext` with full parameter support (`commitment` `votePubkey` filter, `keepUnstakedDelinquents`, `delinquentSlotDistance`)
- Wired epoch_tracker to RPC hooks for epoch vote account lookups
- Partitions validators into current/delinquent based on last vote slot relative to delinquentSlotDistance (default 128)
- Limits epoch credits to last 5 entries per Agave's `MAX_RPC_VOTE_ACCOUNT_INFO_EPOCH_CREDITS_HISTORY`
- Renamed `SlotHookContext` to `RpcHookContext` for broader RPC method support
### Ramifications
- Reads from stakes_cache with a read lock to access vote accounts and their stake
- Uses epoch_tracker.getPtrForSlot() to determine if a vote account is active in the current epoch
### Tests
- Compared response from our impl with Agave's, they seem to match at each commitment level
- TODO: unit tests

### DIFF
--- a/src/rpc/methods.zig
+++ b/src/rpc/methods.zig
@@ -18,6 +18,7 @@ const ParseOptions = std.json.ParseOptions;
 const Pubkey = sig.core.Pubkey;
 const Signature = sig.core.Signature;
 const Slot = sig.core.Slot;
+const Commitment = common.Commitment;
 
 pub fn Result(comptime method: MethodAndParams.Tag) type {
     return union(enum) {
@@ -590,7 +591,7 @@ pub const GetVoteAccounts = struct {
         commitment: ?common.Commitment = null,
         votePubkey: ?Pubkey = null,
         keepUnstakedDelinquents: ?bool = null,
-        delinquintSlotDistance: ?u64 = null,
+        delinquentSlotDistance: ?u64 = null,
     };
 
     pub const Response = struct {
@@ -653,7 +654,7 @@ pub const common = struct {
 
     /// Used to configure several RPC method requests
     pub const CommitmentSlotConfig = struct {
-        commitment: ?Commitment = null,
+        commitment: ?common.Commitment = null,
         minContextSlot: ?sig.core.Slot = null,
     };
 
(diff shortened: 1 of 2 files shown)


src/rpc/methods.zig
```diff
@@ -695,14 +696,123 @@ pub const common = struct {
...
+        const root_ref = self.slot_tracker.getRoot();
+
+        // Setup config consts for the request.
+        const delinquent_distance = config.delinquentSlotDistance orelse
+            DELINQUENT_VALIDATOR_SLOT_DISTANCE;
+        const keep_unstaked = config.keepUnstakedDelinquents orelse false;
+        const filter_pk = config.votePubkey;
+
+        // Get epoch info for epochVoteAccounts check
+        // const epoch = self.epoch_tracker.schedule.getEpoch(slot);
```
yewman:
```suggestion
```

src/rpc/methods.zig
```diff
@@ -695,14 +696,123 @@ pub const common = struct {
...
+        const epoch_constants = try self.epoch_tracker.getEpochInfo(slot);
+        const epoch_stakes = epoch_constants.stakes.stakes;
+        const epoch_vote_accounts = &epoch_stakes.vote_accounts.vote_accounts;
+
+        var current_list = std.ArrayList(GetVoteAccounts.VoteAccount).init(allocator);
```
yewman: Use ArrayListUnmanaged

src/rpc/methods.zig
```diff
@@ -695,14 +696,123 @@ pub const common = struct {
...
+        errdefer current_list.deinit();
```
yewman: These errdefers should also iteratively deinit and votes accounts epoch credits 
```suggestion
        errdefer {
            for (current_list.items) |v| allocator.free(v.epochCredits);
            current_list.deinit();
        };
```

src/rpc/methods.zig
```diff
@@ -695,14 +696,123 @@ pub const common = struct {
...
+        allocator: std.mem.Allocator,
+        params: GetVoteAccounts,
+    ) !GetVoteAccounts.Response {
+        const config: GetVoteAccounts.Config = params.config orelse .{};
+
+        // get slot for requested commitment. Agave uses finalized as default.
+        const slot = self.slot_tracker.getSlotForCommitment(config.commitment orelse .finalized);
+
+        // Get the root slot's state which contains stakes_cache.
+        const root_ref = self.slot_tracker.getRoot();
```
yewman: How come we use the root ref not the slot for commitment?

prestonsn: whoops, good catch. Fixed to use `slot_tracker.get(slot)` in https://github.com/Syndica/sig/pull/1208/commits/c397c101a94bb5d313109774603a5599cb633aa2

Re-tested against testnet Agave and all commitments work as expected

src/rpc/methods.zig
prestonsn: https://github.com/Syndica/sig/commit/c397c101a94bb5d313109774603a5599cb633aa2

src/rpc/methods.zig
prestonsn: https://github.com/Syndica/sig/commit/c397c101a94bb5d313109774603a5599cb633aa2

src/rpc/methods.zig
prestonsn: https://github.com/Syndica/sig/commit/c397c101a94bb5d313109774603a5599cb633aa2

src/rpc/methods.zig
```diff
@@ -695,14 +696,130 @@ pub const common = struct {
...
+                .rootSlot = vote_state.root_slot orelse 0,
+            };
+
+            if (is_current) {
+                try current_list.append(allocator, info);
+            } else {
+                try delinqt_list.append(allocator, info);
+            }
+        }
+        stakes_guard.unlock();
```
InKryption: If anything errors/returns before this, it'll leave the rwmux locked forever. This should probably go into a defer right after it's locked.

src/rpc/methods.zig
```diff
@@ -695,14 +696,130 @@ pub const common = struct {
...
+
+        return .{
+            .current = try current_list.toOwnedSlice(allocator),
+            .delinquent = try delinqt_list.toOwnedSlice(allocator),
+        };
```
InKryption, line 833: The `errdefer`s on `current_list` and `delinqt_list` should be turned into `defer`s, and this code should become:

```suggestion
        const current = try current_list.toOwnedSlice(allocator);
        errdefer {
            for (current) |va| allocator.free(va.epochCredits);
            allocator.free(current);
        }
        const delinquent = try delinqt_list.toOwnedSlice(allocator);
        errdefer {
            for (delinquent) |va| allocator.free(va.epochCredits);
            allocator.free(delinquent);
        }
        return .{
            .current = current,
            .delinquent = delinquent,
        };
```

Rationale: `toOwnedSlice` will empty the arraylist if it succeeds, such that the `defer list.deinit()` on the `list` becomes a safe noop. If it fails, the list will remain untouched, so that the `deinit` can proceed.
Currently, if the first `toOwnedSlice` succeeds, the but the second one fails, the first one will leak, since the original `errdefer`/`defer` will be a noop, as explained.

There is some discussion about adding some functions to improve this workflow, which would make this code look like this in the future:
```zig
try current_list.shrinkToLen();
try delinquent_list.shrinkToLen();
return .{
    .current = current_list.stealSlice(), // asserts capacity == len
    .delinquent = delinquent_list.stealSlice(),
};
```

prestonsn: Ahh that makes sense, thanks for the explanation + context. I have a couple other PRs where this feedback is totally valid :sob: . 

https://github.com/Syndica/sig/pull/1208/commits/caf1f2c4640dec5e781b28490ea88178348b65c5

src/rpc/methods.zig
prestonsn: Good catch! thank you: https://github.com/Syndica/sig/pull/1208/commits/caf1f2c4640dec5e781b28490ea88178348b65c5

InKryption (approved):

PR #1209, opened 2026-02-05
fix(sync): Channel length calculation bug

ajw221:
### Intent

- Fix `Channel.len` block transition adjustment logic, where wrong modulo divisor caused the condition to never trigger

### Implementation

- Replaced `% (LAP - 1)` with `% LAP` in both head and tail if statement conditions

### Ramifications

- The bug only manifests during concurrent access when `len()` reads head/tail at the exact moment of a block transition and just so happened to surface while testing another area of the code that uses `Channel`.

### Tests

N/A


Rexicon226 (approved):
LGTM. The actual typo here is using `%` instead of `&`, but this fix is identical in behaviour.

PR #1212, opened 2026-02-06
feat(rpc): implement getBalance RPC method

prestonsn:
Implements [getBalance](https://solana.com/docs/rpc/http/getbalance) RPC method for accounts db two

### DIFF
--- a/src/rpc/methods.zig
+++ b/src/rpc/methods.zig
@@ -19,6 +19,7 @@ const Pubkey = sig.core.Pubkey;
 const Signature = sig.core.Signature;
 const Slot = sig.core.Slot;
 const Commitment = common.Commitment;
+const ClientVersion = sig.version.ClientVersion;
 
 pub fn Result(comptime method: MethodAndParams.Tag) type {
     return union(enum) {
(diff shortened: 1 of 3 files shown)


src/rpc/methods.zig
```diff
@@ -838,6 +840,42 @@ pub const RpcHookContext = struct {
...
+        // https://github.com/anza-xyz/agave/blob/v3.1.8/rpc/src/rpc.rs#L348
+        const commitment = config.commitment orelse .finalized;
+
+        const slot = self.slot_tracker.getSlotForCommitment(commitment);
+        if (config.minContextSlot) |min_slot| {
+            if (slot < min_slot) return error.RpcMinContextSlotNotMet;
+        }
+
+        // Get slot reference to access ancestors
+        const ref = self.slot_tracker.get(slot) orelse return error.SlotNotFound;
```
yewman: NIT: Error SlotNotFound is inconsistent with other errors returned for the same step in other methods. We should use a consistent error at this step

prestonsn: good catch, made this consistent with getVoteAccounts:
```zig
const ref = self.slot_tracker.get(slot) orelse return error.SlotNotAvailable;
```

yewman (approved):

PR #1217, opened 2026-02-07
feat(rpc): impl getAccountInfo RPC for accounts_db v2

prestonsn:
Implements `getAccountInfo` RPC for accounts db `two`. The PR is a bit large mostly due to porting over agave's [account-decoder](https://github.com/anza-xyz/agave/tree/v3.1.8/account-decoder) crate to zig. This also includes porting over all related unit tests.

There's some quirkiness (specific to SPL token 22 metadata accounts) in the code due to us not having a borsh lib available. Instead of implementing one that we may not need all the functionality of, this PR hand-rolls deserialization for things like an array of strings. The number of cases we need to handle was pretty minimal, so it didn't make sense implementing something more complete/robust.

Testing done (against Agave testnet RPC @ v3.1.8): 
| Category | What Was Tested |
|----------|-----------------|
| Encodings | jsonParsed, base64, base64+zstd, base58, fallback format |
| Commitment Levels | finalized, confirmed, processed |
| Parameters | dataSlice (offset/length) |
| SPL Token Mint | Fields: mintAuthority, supply, decimals, freezeAuthority, isInitialized |
| SPL Token Account | Fields: owner, mint, tokenAmount, state, isNative, delegate, closeAuthority |
| Token-2022 | Discriminator offset, extension parsing, program name, MintCloseAuthority, TransferFee, InterestBearingConfig |
| Vote Account | Base fields + v4 SIMD-0185 fields (inflationRewardsCommissionBps, inflationRewardsCollector, blockRevenueCollector, blockRevenueCommissionBps, pendingDelegatorRewards, blsPubkeyCompressed) |
| Stake Account | Initialized state, stake: null handling |
| Address Lookup Table | Parsed data, kebab-case program name |
| BPF Upgradeable Loader | ProgramData, authority: null handling |
| Sysvars | Clock, EpochSchedule, EpochRewards, Rent, SlotHashes, StakeHistory, SlotHistory, LastRestartSlot |
| Response Structure | context.slot, context.apiVersion |

### DIFF
--- a/src/cmd.zig
+++ b/src/cmd.zig
@@ -1694,19 +1694,14 @@ fn validator(
     });
     defer replay_service_state.deinit(allocator);
 
-    const account_store = sig.accounts_db.AccountStore{
-        .accounts_db = &new_db,
-    };
-
     try app_base.rpc_hooks.set(allocator, sig.rpc.methods.RpcHookContext{
         .slot_tracker = &replay_service_state.replay_state.slot_tracker,
         .epoch_tracker = &epoch_tracker,
-        .account_reader = account_store.reader(),
     });
 
     try app_base.rpc_hooks.set(
         allocator,
-        sig.rpc.methods.AccountHookContext{
+        sig.rpc.hook_contexts.AccountHookContext{
             .slot_tracker = &replay_service_state.replay_state.slot_tracker,
             .account_reader = replay_service_state.replay_state.account_store.reader(),
         },
--- a/src/rpc/account_codec/parse_account_lookup_table.zig
+++ b/src/rpc/account_codec/parse_account_lookup_table.zig
@@ -1,5 +1,5 @@
-/// Types for parsing a address lookup table accounts for RPC responses using the `jsonParsed` encoding.
-/// [agave]: https://github.com/anza-xyz/agave/blob/v3.1.8/account-decoder/src/parse_address_lookup_table.rs
+//! Types for parsing a address lookup table accounts for RPC responses using the `jsonParsed` encoding.
+//! [agave]: https://github.com/anza-xyz/agave/blob/v3.1.8/account-decoder/src/parse_address_lookup_table.rs
 const std = @import("std");
 const sig = @import("../../sig.zig");
 
--- a/src/rpc/methods.zig
+++ b/src/rpc/methods.zig
@@ -186,7 +186,7 @@ pub const GetAccountInfo = struct {
     pub const Config = struct {
         commitment: ?common.Commitment = null,
         minContextSlot: ?u64 = null,
-        encoding: ?common.AccountEncoding = null,
+        encoding: ?account_codec.AccountEncoding = null,
         dataSlice: ?common.DataSlice = null,
     };
 
@@ -1193,8 +1193,6 @@ pub const common = struct {
         apiVersion: []const u8 = ClientVersion.API_VERSION,
     };
 
-    pub const AccountEncoding = account_codec.AccountEncoding;
-
     // TODO field types
     pub const RpcContactInfo = struct {
         /// Pubkey of the node as a base-58 string
@@ -1245,7 +1243,6 @@ pub const common = struct {
 
 pub const RpcHookContext = struct {
     slot_tracker: *sig.replay.trackers.SlotTracker,
-    account_reader: sig.accounts_db.AccountReader,
     epoch_tracker: *sig.core.EpochTracker,
 
     // Limit the length of the `epoch_credits` array for each validator in a `get_vote_accounts`
@@ -1380,43 +1377,6 @@ pub const RpcHookContext = struct {
         return .{
             .current = current,
             .delinquent = dlinqt,
-        };
-    }
-
-    pub fn getBalance(
-        self: RpcHookContext,
-        allocator: std.mem.Allocator,
-        params: GetBalance,
-    ) !GetBalance.Response {
-        const config = params.config orelse common.CommitmentSlotConfig{};
-        // [agave] Default commitment is finalized:
-        // https://github.com/anza-xyz/agave/blob/v3.1.8/rpc/src/rpc.rs#L348
-        const commitment = config.commitment orelse .finalized;
-
-        const slot = self.slot_tracker.getSlotForCommitment(commitment);
-        if (config.minContextSlot) |min_slot| {
-            if (slot < min_slot) return error.RpcMinContextSlotNotMet;
-        }
-
-        // Get slot reference to access ancestors
-        const slot_ref = self.slot_tracker.get(slot) orelse return error.SlotNotAvailable;
-        defer slot_ref.release();
-        const slot_reader = self.account_reader.forSlot(&slot_ref.constants().ancestors);
-
-        // Look up account
-        const maybe_account = try slot_reader.get(allocator, params.pubkey);
-
-        const lamports: u64 = if (maybe_account) |account| blk: {
-            defer account.deinit(allocator);
-            break :blk account.lamports;
-        } else 0;
-
-        return .{
-            .context = .{
-                .slot = slot,
-                .apiVersion = ClientVersion.API_VERSION,
-            },
-            .value = lamports,
         };
     }
 };
--- a/src/rpc/account_codec/lib.zig
+++ b/src/rpc/account_codec/lib.zig
@@ -1,5 +1,5 @@
-/// This module provides encoding and decoding of Solana accounts for RPC responses.
-/// Supports `jsonParsed`, `base58`, `base64`, and `base64+zstd` encodings.
+//! This module provides encoding and decoding of Solana accounts for RPC responses.
+//! Supports `jsonParsed`, `base58`, `base64`, and `base64+zstd` encodings.
 const std = @import("std");
 const base58 = @import("base58");
 const zstd = @import("zstd");
@@ -168,8 +168,8 @@ pub fn encodeJsonParsed(
     // [agave] https://github.com/anza-xyz/agave/blob/8803776abe/rpc/src/rpc.rs#L2504-L2509
     // SPL token accounts ignore data_slice in the fallback path because
     // get_parsed_token_account hardcodes it to None.
-    const is_spl_token = sig.runtime.ids.SPL_TOKEN_PROGRAM_ID.equals(&account.owner) or
-        sig.runtime.ids.SPL_TOKEN_2022_PROGRAM_ID.equals(&account.owner);
+    const is_spl_token = sig.runtime.ids.TOKEN_PROGRAM_ID.equals(&account.owner) or
+        sig.runtime.ids.TOKEN_2022_PROGRAM_ID.equals(&account.owner);
     const fallback_slice = if (is_spl_token) null else data_slice;
     var encoded = try std.ArrayListUnmanaged(u8).initCapacity(
         allocator,
@@ -515,8 +515,8 @@ const ParsableAccount = enum {
         // [agave] https://github.com/anza-xyz/agave/blob/v3.1.8/account-decoder/src/parse_account_data.rs#L48
         if (program_id.equals(&sig.runtime.sysvar.OWNER_ID)) return .sysvar;
         if (program_id.equals(&sig.runtime.program.config.ID)) return .config;
-        if (program_id.equals(&sig.runtime.ids.SPL_TOKEN_PROGRAM_ID)) return .token;
-        if (program_id.equals(&sig.runtime.ids.SPL_TOKEN_2022_PROGRAM_ID)) return .token_2022;
+        if (program_id.equals(&sig.runtime.ids.TOKEN_PROGRAM_ID)) return .token;
+        if (program_id.equals(&sig.runtime.ids.TOKEN_2022_PROGRAM_ID)) return .token_2022;
         return null;
     }
 
@@ -655,8 +655,8 @@ pub fn buildTokenAdditionalData(
     slot_reader: sig.accounts_db.SlotAccountReader,
 ) AdditionalAccountData {
     // Check if this is a token account
-    const is_token_program = account.owner.equals(&sig.runtime.ids.SPL_TOKEN_PROGRAM_ID) or
-        account.owner.equals(&sig.runtime.ids.SPL_TOKEN_2022_PROGRAM_ID);
+    const is_token_program = account.owner.equals(&sig.runtime.ids.TOKEN_PROGRAM_ID) or
+        account.owner.equals(&sig.runtime.ids.TOKEN_2022_PROGRAM_ID);
     if (!is_token_program) return .{};
 
     // Read account data to extract mint pubkey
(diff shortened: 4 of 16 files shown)


src/rpc/account_decoder/parse_token.zig
```diff
@@ -0,0 +1,1561 @@
...
+        .extensions = parseExtensions(extension_data),
+    } };
+}
+
+fn parseAsMultisig(data: []const u8) ?TokenAccountType {
+    const multisig = Multisig.unpack(data) catch return null;
+    if (!multisig.is_initialized) return null;
+    // Collect non-zero signers up to n valid signers
+    var signers: JsonArray(Pubkey, Multisig.MAX_SIGNERS) = .{};
+    for (multisig.signers[0..multisig.n]) |signer| {
```
yewman: multisig.n could be out of bounds

src/rpc/account_decoder/parse_config.zig
```diff
@@ -0,0 +1,310 @@
...
+    if (keys_size > data.len) return ParseError.InvalidAccountData;
+    const config_data = data[keys_size..];
+    // Deserialize ValidatorInfo (length-prefixed string)
+    const validator_info = bincode.readFromSlice(allocator, ValidatorInfo, config_data, .{}) catch
+        return ParseError.InvalidAccountData;
+    defer allocator.free(validator_info.info);
+    // Build UI keys array
+    const ui_keys = allocator.alloc(UiConfigKey, config_keys.keys.len) catch
+        return ParseError.OutOfMemory;
+    for (config_keys.keys, 0..) |key, i| {
```
yewman: No need to manually catch re-return error.OutOfMemory. Also will leak if next alloc fails
```suggestion
    const ui_keys = try allocator.alloc(UiConfigKey, config_keys.keys.len);
    errdefer allocator.free(ui_keys);
    for (config_keys.keys, 0..) |key, i| {
```

src/rpc/account_decoder/parse_config.zig
```diff
@@ -0,0 +1,310 @@
...
+        ui_keys[i] = UiConfigKey{
+            .pubkey = key.pubkey,
+            .signer = key.is_signer,
+        };
+    }
+    // Copy the info string (we need to own it since we're freeing validator_info)
+    const info_copy = allocator.dupe(u8, validator_info.info) catch
+        return ParseError.OutOfMemory;
```
yewman: While not strictly necessary, errdefer's for the last allocation in blocks is a useful practice imo. It reduces the burden on a reader needing to understand if there is a failure point between the allocation and the block exit, and also lowers the risk of forgetting to add an errdefer in the future if another fallible operation is added.
```suggestion
    const info_copy = allocator.dupe(u8, validator_info.info);
    errdefer allocator.free(info_copy);
```

src/rpc/account_decoder/parse_bpf_upgradeable_loader.zig
```diff
@@ -0,0 +1,356 @@
...
+fn readRemainingBytes(
+    allocator: Allocator,
+    reader: anytype,
+    data_len: u32,
+    metadata_size: usize,
+) ![]const u8 {
+    if (data_len < metadata_size) return error.InvalidData;
+    const bytecode_len = data_len - metadata_size;
+    if (bytecode_len == 0) return &.{};
+    const bytecode = allocator.alloc(u8, bytecode_len) catch return error.OutOfMemory;
```
yewman:
```suggestion
    const bytecode = try allocator.alloc(u8, bytecode_len);
    errdefer allocator.free(bytecode);
```

src/rpc/methods.zig
```diff
@@ -239,7 +242,7 @@ pub const GetAccountInfo = struct {
                 ) std.json.ParseError(@TypeOf(source.*))!Data {
                     return switch (try source.peekNextTokenType()) {
                         .array_begin => .{ .encoded = try std.json.innerParse(
```
yewman: Is the current approach to handling `object_begin` correct given these changes?

src/rpc/methods.zig
```diff
@@ -706,3 +716,225 @@ pub const SlotHookContext = struct {
...
+        // However, `Binary` is deprecated and `Base64` is preferred for performance.
+        // We default to base64 as it's more efficient and the recommended encoding.
+        const encoding = config.encoding orelse common.AccountEncoding.base64;
+
+        const slot = self.slot_tracker.getSlotForCommitment(commitment);
+        if (config.minContextSlot) |min_slot| {
+            if (slot < min_slot) return error.RpcMinContextSlotNotMet;
+        }
+
+        // TODO: is this the best way to get the right slot to use?
```
yewman: Is this still relevant? Aren't we getting the right slot to use in the previous step?

src/rpc/methods.zig
```diff
@@ -706,3 +716,225 @@ pub const SlotHookContext = struct {
...
+                while (iter.nextFrame()) |frame_slice| {
+                    try base64_ctx.writer().writeAll(frame_slice);
+                }
+                try base64_ctx.flush();
+            },
+            .@"base64+zstd" => {
+                var stream = sig.utils.base64.EncodingStream.init(std.base64.standard.Encoder);
+                const base64_ctx = stream.writerCtx(writer);
+                // TODO: propagate more specifi errors.
+                const compressor = zstd.Compressor.init(.{}) catch return error.OutOfMemory;
```
yewman: Why not just use try?

prestonsn: :100: :100:

src/rpc/account_codec/parse_account_lookup_table.zig
```diff
@@ -0,0 +1,204 @@
...
+    const addresses = allocator.alloc(Pubkey, lookup_table.addresses.len) catch
+        return ParseError.OutOfMemory;
+    @memcpy(addresses, lookup_table.addresses);
+
+    return .{ .lookup_table = .{
+        .deactivationSlot = .{ .value = lookup_table.meta.deactivation_slot },
+        .lastExtendedSlot = .{ .value = lookup_table.meta.last_extended_slot },
+        .lastExtendedSlotStartIndex = lookup_table.meta.last_extended_slot_start_index,
+        .authority = lookup_table.meta.authority,
+        .addresses = addresses,
```
hamza-syndica, line 48: Please document that the caller must free `.addresses`. Right?

src/cmd.zig
```diff
@@ -1666,6 +1666,17 @@ fn validator(
...
+        .accounts_db_two = &new_db,
+    };
+    try app_base.rpc_hooks.set(
+        allocator,
+        sig.rpc.methods.AccountHookContext{
+            .slot_tracker = &replay_service_state.replay_state.slot_tracker,
+            .account_reader = account_store.reader(),
+        },
+    );
+
```
ajw221: we should be able to also reuse `account_store` from `replay_service_state.replay_state.account_store`
```suggestion
    try app_base.rpc_hooks.set(
        allocator,
        sig.rpc.methods.AccountHookContext{
            .slot_tracker = &replay_service_state.replay_state.slot_tracker,
            .account_reader = replay_service_state.replay_state.account_store.reader(),
        },
    );

```

src/rpc/methods.zig
```diff
@@ -706,3 +716,225 @@ pub const SlotHookContext = struct {
...
+            if (slot < min_slot) return error.RpcMinContextSlotNotMet;
+        }
+
+        // TODO: is this the best way to get the right slot to use?
+        const ref = self.slot_tracker.get(slot) orelse return error.SlotNotFound;
+        const slot_reader = self.account_reader.forSlot(&ref.constants.ancestors);
+        const maybe_account = try slot_reader.get(allocator, params.pubkey);
+
+        if (maybe_account) |account| {
+            defer account.deinit(allocator);
```
ajw221:
```suggestion
        const maybe_account = try slot_reader.get(allocator, params.pubkey);
        defer if (maybe_account) |account| account.deinit(allocator);

        if (maybe_account) |account| {
```

src/rpc/methods.zig
```diff
@@ -706,3 +716,225 @@ pub const SlotHookContext = struct {
...
+
+            const data: GetAccountInfo.Response.Value.Data = if (encoding == .jsonParsed)
+                try encodeJsonParsed(allocator, params.pubkey, account, slot_reader)
+            else
+                try encodeStandard(allocator, account, encoding, config.dataSlice);
+
+            return GetAccountInfo.Response{
+                .context = .{ .slot = slot, .apiVersion = ClientVersion.API_VERSION },
```
ajw221: thoughts on defaulting the struct field instead?
```suggestion
                .context = .{ .slot = slot },
```

src/rpc/methods.zig
```diff
@@ -11,13 +11,21 @@
...
 
 const Pubkey = sig.core.Pubkey;
 const Signature = sig.core.Signature;
 const Slot = sig.core.Slot;
+const ClientVersion = sig.version.ClientVersion;
+
+const account_decoder = sig.rpc.account_decoder;
+
+const MAX_BASE58_INPUT_LEN = 128;
+const MAX_BASE58_OUTPUT_LEN = base58.encodedMaxSize(MAX_BASE58_INPUT_LEN);
```
ajw221: if these are `getAccount` specific, we should move these where encoding for accounts will happen generally

src/rpc/account_decoder/lib.zig
```diff
@@ -0,0 +1,590 @@
...
+
+    pub fn jsonStringify(self: ParsedContent, jw: anytype) @TypeOf(jw.*).Error!void {
+        switch (self) {
+            inline else => |content| try content.jsonStringify(jw),
+        }
+    }
+};
+
+/// Enum of programs that support jsonParsed.
+const ParsableProgram = enum {
```
ajw221: should add agave's reference line here and also rename to `ParsableAccount` since there is actually an existing `ParsableProgram` that is similar but not the same

src/rpc/account_decoder/parse_token.zig
prestonsn: Checked the agave src for this (https://github.com/anza-xyz/agave/blob/2717084afeeb7baad4342468c27f528ef617a3cf/account-decoder/src/parse_token.rs#L100C1-L100C24)

We could instead just iterate over all of `multisig.signers` (`MAX_SIGNER` entries total). Avoids oob entirely

src/rpc/account_decoder/parse_config.zig
prestonsn: Fixed this and a couple other similar areas with a similar issue

src/rpc/account_decoder/parse_config.zig
prestonsn: Thanks for the context, and I agree that it's just less to keep track of/easier to trust. I'll keep this in mind :saluting_face:

src/rpc/methods.zig
prestonsn: Yup! though parsing this stuff for sake of the client unit tests is likely a bunch more code (jsonParse impl required for a bunch of the types in `account_decoder`). Likely something we can come back to if we need it in the future.

prestonsn: For now, removing that panic makes sense. Going to use `std.json.Value` to deal this unit test for now. Can come back in the future if we need for deserialization support

src/rpc/methods.zig
prestonsn: good catch, thanks

src/rpc/account_codec/parse_account_lookup_table.zig
prestonsn: Good call - Added a doc comment for this. Also noticed that caller wasn't actually freeing `.addresses` at all. Fixed by using an `ArenaAllocator` to encapsulate all allocations during serialization. Much easier to track and free all allocations at once this way

src/rpc/methods.zig
prestonsn: Thanks, done!

src/rpc/methods.zig
prestonsn: Moved into the new `account_codec` module.

src/rpc/account_decoder/lib.zig
prestonsn: Done, also added link to Agave

src/rpc/methods.zig
prestonsn: refactored, thanks!

src/cmd.zig
prestonsn: refactored, thanks!

yewman (approved):

src/rpc/account_codec/parse_account_lookup_table.zig
```diff
@@ -0,0 +1,208 @@
+/// Types for parsing a address lookup table accounts for RPC responses using the `jsonParsed` encoding.
+/// [agave]: https://github.com/anza-xyz/agave/blob/v3.1.8/account-decoder/src/parse_address_lookup_table.rs
```
prestonsn: Wrong syntax for doc comment - need to fix across files

src/rpc/methods.zig
```diff
@@ -902,3 +832,66 @@ pub const StaticHookContext = struct {
         return .{ .hash = self.genesis_hash };
     }
 };
+
+pub const AccountHookContext = struct {
```
dnut: could you move this to the `rpc/hook_contexts` folder

src/rpc/methods.zig
```diff
@@ -902,3 +832,66 @@ pub const StaticHookContext = struct {
...
+    ) !GetAccountInfo.Response {
+        const config = params.config orelse GetAccountInfo.Config{};
+        // [agave] Default commitment is finalized:
+        // https://github.com/anza-xyz/agave/blob/v3.1.8/rpc/src/rpc.rs#L348
+        const commitment = config.commitment orelse .finalized;
+        // [agave] Default encoding in AGave is `Binary` (legacy base58):
+        // https://github.com/anza-xyz/agave/blob/v3.1.8/rpc/src/rpc.rs#L545
+        // However, `Binary` is deprecated and `Base64` is preferred for performance.
+        // We default to base64 as it's more efficient and the recommended encoding.
+        const encoding = config.encoding orelse common.AccountEncoding.base64;
```
dnut: The API specification says binary is the default, so I believe this is incorrect.

Also btw you don't need to fully qualify the enum variant here.

```suggestion
        // We default to base64 as it's more efficient and the recommended encoding.
        const encoding = config.encoding orelse .binary;
```

src/rpc/methods.zig
```diff
@@ -205,70 +199,7 @@ pub const GetAccountInfo = struct {
...
-                        ) },
-                        .array_end, .object_end => error.UnexpectedToken,
-                        else => {
-                            try source.skipValue();
-                            return error.UnexpectedToken;
-                        },
-                    };
-                }
-            };
+            pub const Data = account_codec.AccountData;
```
dnut: This alias is a little confusing. Is there a reason for using inconsistent naming?

src/rpc/methods.zig
prestonsn: just spoke to @ajw221 about this, i think best thing to do here is to create `hook_contexts/Account.zig` to house all the account related RPCs (akin to what Adam's done in `hook_contexts/Ledger.zig` for getBlock)

src/rpc/methods.zig
prestonsn: Thanks! done: https://github.com/Syndica/sig/pull/1217/commits/64cf59027532c62541a4570a7608b44b020b5b5e

src/rpc/methods.zig
prestonsn: No reason, forgot i had. Fixed: https://github.com/Syndica/sig/pull/1217/commits/372d35ea5796a7e9f976eccc8e4267f977ad3e2e

src/rpc/methods.zig
prestonsn: There are some existing hook contexts (outside of the changes in this PR) like `RpcHookContext` and `StaticHookContext` that will have to be moved too for consistency.

prestonsn: Alright, moved getAccountInfo and getBalance (also an accounts based RPC method) into `hook_contexts/Account.zig` https://github.com/Syndica/sig/pull/1217/commits/55f937b961249499b7758b552b29583f850cb91c

prestonsn: There are some remaining HookContexts (static and the generic Rpc one) lingering in `methods.zig`, but i feel like moving those are outside the scope of this PR. Can move them all if you prefer tho

yewman (approved):

PR #1222, opened 2026-02-09
feat(rpc): add webzockets dependency for json rpc websockets

jbuckmccready:
## Summary

Adds `webzockets` package to path `src/rpc/webzockets`. `webzockets` implements websockets in zig, built on top of `libxev` for network IO. See `src/rpc/webzockets/README.md` for details on the package itself.

This is only adds a websocket implementation needed to implement the json RPC websocket API. The json RPC API will be in a different PR. 

## Changes

Updated the `libxev` version used by sig in `build.zig.zon` to new commit (current HEAD of zig-0.14 branch): https://github.com/Syndica/libxev/commit/11a00ee9c97e09c80a075af73599be4e1f549efa
This commit is on top of a slightly later version of `libxev` than we were previously using that is still compatible with zig 0.14.1, and the commit fixes error returns for the `kqueue` backend (associated pull request upstream: https://github.com/mitchellh/libxev/pull/207).

Updated `build.zig` and `build.zig.zon` with `webzockets` dependency, and added import to `src/rpc/lib.zig` so `webzockets` is part of the build.

Ran `docs/generate.py` so `webzockets` README.md is included in the published docs.

Updated `scripts/style.py` to skip cache dirs.

### Update 2026-02-13:
Fixed another bug in libxev where io_uring gave unexpected error on shutdown when socket already closed.
Upstream PR: https://github.com/mitchellh/libxev/pull/211
Syndica fork commit: https://github.com/Syndica/libxev/commit/c16162f0f9047f27d6230c548e1e25070b9be163
NOTE: for kqueue/epoll backends it could just issue `std.posix.shutdown` directly to avoid the event loop processing entirely, but io_uring backend actually queues the shutdown command to avoid a syscall.

### Update 2026-02-14:
Added read pause/resume (https://github.com/Syndica/sig/commit/002805389404d33bd258026144cb5153ee4e9e4b)
- Added heuristic for compaction to avoid small TCP reads (e.g., want to avoid a full TCP read cycle through IO loop if there is only 46 bytes remaining in the read buffer but compaction would free 8KB).
- Added ability for connections to pause reads to exert back pressure on the TCP connection. This is helpful for future jrpc layer to allow server handlers to not have any queue of received messages by having the control flow: receive sub/unsub request from client -> pause reads -> send sub/unsub response -> write complete -> resume reads.

Removed BufferPool (https://github.com/Syndica/sig/commit/51641cccdd79e9cca270acb27c94dd3f10c6cb59), it will not be used by the jrpc server (embedded read buffer will cover sub/unsub requests) and adds complexity.

### Update 2026-02-19:
Optimized close initiated by caller to not wait for handshake for the close frame response (not required by websockets RFC) and simplified logic a bit for closing. Commit: https://github.com/Syndica/sig/pull/1222/changes/19cf287dd894bf507a057e09ef5b9ff49639f791

### Update 2026-02-22:
Added `sendRaw` to api to allow the caller to prepare the web socket frames (header + payload), this avoids the 2 sends per message required for zero-copy use (one for header, one for payload) without having to change the backend to support vectored writes. Jrpc broadcasting benchmarks show about 50% increase in performance with this small change.

Note even with vectored writes this API is useful as it allows the caller to batch an arbitrary number of websocket messages into a single TCP write buffer, avoids creating the frame on the IO loop, and can support the caller performing per-message deflate compression in the future (allowing for shared broadcasting across web socket clients utilizing per-message deflate).

### Update 2026-02-24:
Fix missing timeout on handshake upgrade: https://github.com/Syndica/sig/commit/3a7db8f96161944c81b1616b84e0b6437e96fb5b
Needed otherwise TCP socket can stay open indefinitely and not release associated resources.

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -26,8 +26,8 @@
             .hash = "lsquic-4.4.0-oZGGDWvvEQFN7TuaRMwK0XPZyqHRhIF_URW14bo-4A8a",
         },
         .xev = .{
-            .url = "git+https://github.com/Syndica/libxev#6d5de49dde845e3045bac30b2018da32a4915c84",
-            .hash = "libxev-0.0.0-86vtczgdEwAKwvPWRNzuZMW7FuhDVdSbwgpJAJ0FmICg",
+            .url = "git+https://github.com/mitchellh/libxev#7e7d2f2ab4700544657f8ec268715c8ef320d839",
+            .hash = "libxev-0.0.0-86vtcwE9EwB942iWRnaNMXHv3n0BeLAs_tVhrs5cT8cQ",
         },
         .prettytable = .{
             .url = "git+https://github.com/dying-will-bullet/prettytable-zig#43e8fbbff4cbacbef195aa4fea281375a77c2026",
--- a/src/rpc/webzockets/build.zig.zon
+++ b/src/rpc/webzockets/build.zig.zon
@@ -9,8 +9,8 @@
 
     .dependencies = .{
         .libxev = .{
-            .url = "git+https://github.com/Syndica/libxev#6d5de49dde845e3045bac30b2018da32a4915c84",
-            .hash = "libxev-0.0.0-86vtczgdEwAKwvPWRNzuZMW7FuhDVdSbwgpJAJ0FmICg",
+            .url = "git+https://github.com/mitchellh/libxev#7e7d2f2ab4700544657f8ec268715c8ef320d839",
+            .hash = "libxev-0.0.0-86vtcwE9EwB942iWRnaNMXHv3n0BeLAs_tVhrs5cT8cQ",
         },
     },
     .paths = .{


src/rpc/webzockets/src/buffer.zig
```diff
@@ -0,0 +1,170 @@
...
+const Allocator = std.mem.Allocator;
+
+/// Growable buffer pool with runtime-configurable buffer size.
+/// Uses an intrusive free list (like std.heap.MemoryPool) for efficient reuse.
+/// Thread-safe for shared use across connections.
+///
+/// Pool grows beyond preheat size as needed. Use preheat to pre-allocate
+/// buffers for expected concurrent usage, avoiding allocations in the
+/// common case.
+pub const BufferPool = struct {
```
Sobeston: Not convinced we need something like this, we already have e.g. arena allocators

src/rpc/webzockets/src/buffer.zig
```diff
@@ -0,0 +1,170 @@
...
+    buffer_size: usize,
+
+    /// Arena for growing allocations (never shrinks until deinit).
+    arena: std.heap.ArenaAllocator,
+
+    /// Head of the intrusive free list (null if no free buffers).
+    free_list: ?*FreeNode,
+
+    /// Mutex for thread-safe access.
+    mutex: std.Thread.Mutex,
```
Sobeston: We can do without multi-threading here - if we do have a pool (or arena), per-thread is probably better.

src/rpc/webzockets/autobahn/client/client.zig
```diff
@@ -0,0 +1,460 @@
...
+const read_buf_size: usize = 4096;
+
+/// Maximum reassembled message size — Autobahn sends up to ~16MB.
+const max_message_size: usize = 20 * 1024 * 1024;
+
+const AutobahnClient = ws.Client(AutobahnClientHandler, read_buf_size);
+
+// ============================================================================
+// AutobahnRunner — orchestrates sequential test-case execution
+// ============================================================================
```
Sobeston: Bad doc string

src/rpc/webzockets/autobahn/client/client.zig
```diff
@@ -0,0 +1,460 @@
...
+
+        next_res catch |err| {
+            log.err("failed to start next step: {}", .{err});
+        };
+    }
+};
+
+// ============================================================================
+// AutobahnClientHandler — echo handler for the Autobahn testsuite
+// ============================================================================
```
Sobeston: Bad doc string

src/rpc/webzockets/autobahn/client/client.zig
```diff
@@ -0,0 +1,460 @@
...
+        self.runner.onConnectionDone();
+    }
+};
+
+// ============================================================================
+// main
+// ============================================================================
+
+pub fn main() !void {
+    const allocator = std.heap.c_allocator;
```
Sobeston: Let's avoid the c_allocator and prefer the gpa

src/rpc/webzockets/src/reader.zig
```diff
@@ -0,0 +1,1233 @@
...
+        /// Position within buf that we've read into (end of data).
+        pos: usize,
+
+        /// Position in buf where the current message starts.
+        /// Used for over-read handling when multiple messages arrive.
+        start: usize,
+
+        /// Maximum allowed message size.
+        max_message_size: usize,
+
```
Sobeston: Many redundant comments

src/rpc/webzockets/src/reader.zig
```diff
@@ -0,0 +1,1233 @@
...
+const BufferPool = buffer.BufferPool;
+
+/// WebSocket frame reader with automatic buffer tier escalation.
+///
+/// Manages read buffer state with automatic tier escalation:
+/// - Starts with an embedded buffer for small messages
+/// - Upgrades to pooled or dynamic buffer when needed
+/// - Retains larger buffer for subsequent messages (performance optimization)
+/// - Restores to embedded only on explicit reset()
```
Sobeston, line 12: Not convinced that we need any of this

src/rpc/webzockets/src/server/server.zig
```diff
@@ -0,0 +1,415 @@
...
+        /// Periodic timer that polls for drain completion during shutdown.
+        shutdown_timer: xev.Timer,
+        /// xev completion for the shutdown drain-check timer.
+        shutdown_timer_completion: xev.Completion,
+        /// xev completion for closing the listen socket during shutdown.
+        listen_close_completion: xev.Completion,
+        /// Absolute nanoTimestamp after which shutdown times out.
+        shutdown_deadline: i128,
+        /// Type-erased context pointer for the shutdown completion callback.
+        shutdown_userdata: ?*anyopaque,
```
Sobeston: redundant comments

src/rpc/webzockets/src/mask.zig
```diff
@@ -0,0 +1,145 @@
+const std = @import("std");
+
+/// Whether the current backend supports SIMD vector operations.
+const backend_supports_vectors = switch (@import("builtin").zig_backend) {
+    .stage2_llvm, .stage2_c => true,
+    else => false,
+};
+
```
Sobeston: Is this correct? If so can you link something on the issue tracker. Worth noting we only use the llvm backend

src/rpc/webzockets/src/mask.zig
```diff
@@ -0,0 +1,145 @@
...
+/// Applies XOR masking/unmasking to payload in-place using the 4-byte mask key.
+/// The same function works for both masking and unmasking (XOR is its own inverse).
+/// Uses SIMD acceleration when available, with scalar fallback.
+pub fn mask(m: []const u8, payload: []u8) void {
+    std.debug.assert(m.len == 4);
```
Sobeston: if m must be len 4, then why not e.g. `*const [4]u8`

src/rpc/webzockets/src/root.zig
```diff
@@ -0,0 +1,37 @@
...
+pub const ClientConnection = client_connection.ClientConnection;
+pub const Client = client.Client;
+pub const ClientMaskPRNG = types.ClientMaskPRNG;
+
+test {
+    // Uncomment to see logs during tests
+    // std.testing.log_level = .debug;
+    std.testing.refAllDecls(@This());
+    std.testing.refAllDecls(@import("control_queue.zig"));
+}
```
Sobeston, line 42: We should be moving away from refAllDecls

src/rpc/webzockets/autobahn/client/client.zig
jbuckmccready: https://github.com/Syndica/sig/commit/9402c82b06cd72b1543c15fafa38fbce6e2db7d1

src/rpc/webzockets/autobahn/client/client.zig
jbuckmccready: https://github.com/Syndica/sig/commit/9402c82b06cd72b1543c15fafa38fbce6e2db7d1

src/rpc/webzockets/src/server/server.zig
jbuckmccready: https://github.com/Syndica/sig/commit/aa9cb5c298c4a7578ecedf5839802affaffaaea7

src/rpc/webzockets/autobahn/client/client.zig
jbuckmccready: https://github.com/Syndica/sig/commit/744fc9a98a21381af2d464557f091e4775ece034

src/rpc/webzockets/src/mask.zig
jbuckmccready: I copied from: https://github.com/karlseguin/websocket.zig/blob/97fefafa59cc78ce177cff540b8685cd7f699276/src/proto.zig#L7-L10

I didn't drill into the compiler to see details. I put it in mostly just to note a place where disabling may be needed - for our use case it's not needed.

src/rpc/webzockets/autobahn/client/client.zig
jbuckmccready: I can update autobahn server as well. I can change to GPA but still no graceful exit (no leak check), or change to GPA and add `sigint` handler and graceful shutdown and check for leak (this adds some lines). LMK

src/rpc/webzockets/src/mask.zig
jbuckmccready: Yeah, and style wise should I pass by value or *const here? Likely compiles to same thing but style wise is there preference for small values/arrays like this?

src/rpc/webzockets/src/buffer.zig
jbuckmccready: I don't quite follow, how would I use arena allocator effectively for server side read buffer per client?

This pooling is maybe an over optimization, this approach is biased to the patterns I saw from networking around edge-gateway.

The idea is if you know the use case for the server/API then you can tune the buffer pool size according to message sizes you expect from clients, then clients connecting, sending message, and disconnecting turns into single pop and push rather than going to a more generic/global allocator and relying on it to pool efficiently.

For our very specific use case of broadcasting jrpc subscriptions I I don't think this buffer pool will be used at all since client messages will always fit within the embedded read buffer each client struct has - but this should be effective for a more general web server use case and doesn't add much complexity.

src/rpc/webzockets/src/buffer.zig
jbuckmccready: Yeah I'll change it.

My original thinking was the uncontested path is extremely fast anyways and this would allow more IO loops to share something that even under high volume will be largely uncontested, and it would avoid dead/unbalanced distributions across threads, but that's something easy to refine later. 

Current code does bad things with the lock too (holds while allocating new buffer...). EDIT: oh right it would have to if it's backed by the arena.

src/rpc/webzockets/src/mask.zig
jbuckmccready: https://github.com/Syndica/sig/commit/96a343e994c593a49b774580f2ad4acfa2499ae0

Removed conditional stuff with backend, changed to pass by value.

src/rpc/webzockets/src/reader.zig
jbuckmccready: Ties in with https://github.com/Syndica/sig/pull/1222#discussion_r2790299577

src/rpc/webzockets/src/root.zig
jbuckmccready: Should there be explicit import for each file inside test block? I'm not familiar with Zig on this.

src/rpc/webzockets/src/reader.zig
jbuckmccready: https://github.com/Syndica/sig/commit/189852a83697e871b44489f2713c5f8dff212b69

src/rpc/webzockets/src/buffer.zig
jbuckmccready: https://github.com/Syndica/sig/commit/6a75ec748a1304ebfcd03b5819ed8e9986c5e6d8

src/rpc/webzockets/src/mask.zig
jbuckmccready: https://github.com/Syndica/sig/commit/96a343e994c593a49b774580f2ad4acfa2499ae0

src/rpc/webzockets/src/buffer.zig
jbuckmccready: Removed BufferPool, commit: https://github.com/Syndica/sig/commit/51641cccdd79e9cca270acb27c94dd3f10c6cb59

src/rpc/webzockets/src/reader.zig
jbuckmccready: Removed buffer pool, commit: https://github.com/Syndica/sig/commit/51641cccdd79e9cca270acb27c94dd3f10c6cb59

src/rpc/webzockets/autobahn/client/client.zig
```diff
@@ -0,0 +1,458 @@
...
+            return .disarm;
+        };
+        const self = self_opt.?;
+        self.getCaseCount() catch |err| {
+            log.err("getCaseCount failed: {}", .{err});
+        };
+        return .disarm;
+    }
+
+    fn deinit(_: *AutobahnRunner) void {}
```
InKryption: What's the point of this?

src/rpc/webzockets/autobahn/client/client.zig
```diff
@@ -0,0 +1,458 @@
...
+            &self.csprng,
+            .{
+                .address = std.net.Address.initIp4(.{ 127, 0, 0, 1 }, 9001),
+                .path = "/getCaseCount",
+                .max_message_size = max_message_size,
+            },
+        );
+        try self.client.connect();
+        log.debug("getCaseCount: connect submitted to loop", .{});
+    }
```
InKryption: I think this should just be in the `init` function, there's no reason to leave `conn`, `client`, or `handler` undefined like this; it's effectively still uninitialized until you call this function.

Wrt the self-referential `handler` field, if this really needs to be self-referential, initialization should just look like this:
```zig
var runner: AutobahnRunner = undefined;
try runner.init(allocator, &loop);
defer runner.deinit();
```

src/rpc/webzockets/e2e_tests/client/close_tests.zig
```diff
@@ -0,0 +1,251 @@
...
+    });
+    try client.connect();
+    try env.loop.run(.until_done);
+
+    try testing.expect(handler.open_called);
+    try testing.expect(handler.close_called);
+    conn.deinit();
+}
+
+test "e2e client close: close frame still sent when ping write is in flight" {
```
InKryption: just a note: unit tests already include their leading path in their name, so this test can already be filtered for using `<module string>.test.<test name>`, which, if it's referenced in the same structure that its path is, would be something like `e2e_tests.client.close_tests.test.e2e client close: close frame still sent when ping write is in flight`.
Not sure what your intention is here, but thought I'd mention it since it's something I see commonly when folks are unaware of this fact. Feel free to resolve this comment at will.

src/rpc/webzockets/e2e_tests/client/close_tests.zig
```diff
@@ -0,0 +1,251 @@
...
+        return .{ .listener = listener, .port = port, .thread = thread };
+    }
+
+    fn stop(self: *FrameCaptureServer) void {
+        std.posix.close(self.listener);
+        self.thread.join();
+    }
+};
+
+fn acceptAndCaptureFrames(listener: std.posix.socket_t, ctx: *CaptureContext) void {
```
InKryption: Having zero stack traces or error reporting for this will make any failure it has very confusing. I would suggest making this return `!void`, and wrap it up like:
```zig
fn acceptAndCaptureFramesHandled(listener: std.posix.socket_t, ctx: *CaptureContext) void {
    acceptAndCaptureFrames(listener, ctx) catch |err| {
        const ert = if (@errorReturnTrace()) |ert| ert.* else .{
            .index = 0,
            .instruction_addresses = &.{},
        };
        std.debug.print("error: {s}{}", .{ @errorName(err), ert });
    };
}
```
And then you can replace all the `catch return`s with `try`s.

src/rpc/webzockets/e2e_tests/client/close_tests.zig
```diff
@@ -0,0 +1,251 @@
...
+        const n = stream.read(frame_buf[frame_end..]) catch return;
+        if (n == 0) {
+            return;
+        }
+        frame_end += n;
+
+        while (frame_start < frame_end) {
+            const available = frame_buf[frame_start..frame_end];
+            const header = ws.frame.parseHeader(available) catch |err| {
+                if (err == error.InsufficientData) {
```
InKryption: Fair warning: using `==` on errors can be a tad bit of a footgun, because `==` performs PTR (Peer Type Resolution), meaning if you compare two errors from different error sets, like `error.Foo == error.foo` (notice the casing mismatch), it will coerce them to their Peer Type of `error{Foo,foo}`. It's fine here, but in general I would suggest defaulting to `switch`ing on errors for exhaustive checks against mistakes like this.

src/rpc/webzockets/e2e_tests/support/client_handlers.zig
```diff
@@ -0,0 +1,587 @@
...
+        const data = self.send_data orelse return;
+
+        switch (kind) {
+            .ping => {
+                // sendPing copies the payload into an internal queue (no onWriteComplete),
+                // so no allocation or lifetime management is needed.
+                conn.sendPing(data) catch return;
+            },
+            .text => {
+                const copy = self.allocator.dupe(u8, data) catch return;
```
InKryption: Is there no way to return an error indicating critical failure at all? I think there should be. Ignoring these kinds of failures entirely seems ripe for a very confounding debugging experience. Even if only one error is provided for it, so that this would become `catch return error.ClientFailed` - it would retain the error origin in the stack trace.

src/rpc/webzockets/e2e_tests/support/fd_leak.zig
```diff
@@ -0,0 +1,132 @@
...
+/// after cleanup. FD counts are process-wide, so tests using this should be
+/// run serially to avoid false positives from concurrent tests.
+///
+/// Usage:
+/// ```zig
+/// const fd_check = FdLeakDetector.baseline();
+/// defer fd_check.assertNoLeaks();
+/// // ... test body ...
+/// ```
+pub const FdLeakDetector = struct {
```
InKryption: Could just make this be the file. i.e. rename this file to `FdLeakDetector`, and encapsulate everything inside it.

src/rpc/webzockets/e2e_tests/support/fd_leak.zig
```diff
@@ -0,0 +1,132 @@
...
+        if (current != self.baseline_count) {
+            const cur: isize = @intCast(current);
+            const base: isize = @intCast(self.baseline_count);
+            const delta: isize = cur - base;
+            std.debug.panic(
+                "FD leak detected: {d} more FDs open " ++
+                    "than at baseline (baseline={d}, current={d})",
+                .{ delta, base, current },
+            );
+        }
```
InKryption: The conventional thing to do here would be to log the leak, rather than crash the program, such that any other leak detection that needs to take place can also log its errors, so as to provide as much information as possible. See `std.heap.DebugAllocator`'s `deinit` & `detectLeaks` for reference.

src/rpc/webzockets/e2e_tests/support/fd_leak.zig
```diff
@@ -0,0 +1,132 @@
...
+    const detector = FdLeakDetector.baseline();
+    // No FDs opened — should not panic.
+    detector.assertNoLeaks();
+}
+
+test "FdLeakDetector: detects leaked FD" {
+    const detector = FdLeakDetector.baseline();
+
+    // Open a file to leak an FD.
+    const leaked_fd = std.posix.open("/dev/null", .{}, 0) catch return;
```
InKryption:
```suggestion
    const leaked_fd = try std.posix.open("/dev/null", .{}, 0);
```
No reason to ignore this error and act like the test did what it was supposed to. If it fails it's probably due to an issue with the environment, in which case we would want a failure so we can figure out what's going wrong.

src/rpc/webzockets/e2e_tests/support/raw_client.zig
```diff
@@ -0,0 +1,370 @@
...
+    stream: std.net.Stream,
+    allocator: std.mem.Allocator,
+    read_buf: []u8,
+    read_pos: usize, // start of unconsumed data
+    read_end: usize, // end of valid data in read_buf
+
+    pub const default_read_buf_size: usize = 4096;
+
+    pub const ConnectOpts = struct {
+        read_buf_size: usize = default_read_buf_size,
```
InKryption: Any reason this isn't just assigned directly to the field? What else needs to access it?

src/rpc/webzockets/e2e_tests/support/raw_client.zig
```diff
@@ -0,0 +1,370 @@
...
+    /// for tests that receive large echoed messages).
+    pub fn connectEx(allocator: std.mem.Allocator, port: u16, opts: ConnectOpts) !RawClient {
+        const address = std.net.Address.initIp4(.{ 127, 0, 0, 1 }, port);
+        const stream = try std.net.tcpConnectToAddress(address);
+        errdefer stream.close();
+
+        // Set SO_RCVTIMEO for blocking reads (defaults to 2s) to reduce scheduler-jitter flakes.
+        const timeout = std.posix.timeval{
+            .sec = @intCast(opts.read_timeout_ms / 1000),
+            .usec = @intCast((opts.read_timeout_ms % 1000) * 1000),
```
InKryption, line 46: Pretty sure these `@intCast`s are unnecessary, `u32` coerces to `i64` and `isize` (assuming 64 bit target, which we do).

src/rpc/webzockets/e2e_tests/support/raw_client.zig
```diff
@@ -0,0 +1,370 @@
...
+            stream.handle,
+            std.posix.SOL.SOCKET,
+            std.posix.SO.RCVTIMEO,
+            std.mem.asBytes(&timeout),
+        );
+
+        // Generate WebSocket key and build upgrade request
+        var key_buf: [24]u8 = undefined;
+        var raw_key: [16]u8 = undefined;
+        std.crypto.random.bytes(&raw_key);
```
InKryption: I'm not going to die on this hill, but given this is for testing, I think it would be ideal if the random data could be seeded by `std.testing.random_seed`, so that we are able to reproduce a given run if necessary.

src/rpc/webzockets/e2e_tests/support/server_runner.zig
```diff
@@ -0,0 +1,117 @@
...
+            self.thread_pool = xev.ThreadPool.init(.{});
+            errdefer {
+                self.thread_pool.shutdown();
+                self.thread_pool.deinit();
+            }
+
+            self.loop = try allocator.create(xev.Loop);
+            errdefer allocator.destroy(self.loop);
+            self.loop.* = try xev.Loop.init(.{ .thread_pool = &self.thread_pool });
+            errdefer self.loop.deinit();
```
InKryption: Considering the xev loop is already going to live in this ServerRunner, which is stored at a stable, heap-allocated pointer, you could just store the xev loop directly inside it, and elide this double indirection.

src/rpc/webzockets/e2e_tests/support/test_clients.zig
```diff
@@ -0,0 +1,146 @@
...
+
+        var seed: [ws.ClientMaskPRNG.secret_seed_length]u8 = undefined;
+        std.crypto.random.bytes(&seed);
+        self.csprng = ws.ClientMaskPRNG.init(seed);
+    }
+
+    /// Create a client wired to this env's loop, buffer pool, and CSPRNG.
+    pub fn initClient(
+        self: *TestEnv,
+        comptime ClientType: type,
```
InKryption, line 135:
```suggestion
        /// Expected to be `ws.Client(Handler, n)`
        comptime ClientType: type,
```
At first thought this was an undocumented ad-hoc interface, before realizing it was just expecting a generic instance of a Client.

src/rpc/webzockets/e2e_tests/support/test_helpers.zig
```diff
@@ -0,0 +1,47 @@
...
+
+/// Connect to the server, complete a WebSocket handshake, send a text message,
+/// verify the echo, and close. Used to confirm the server is still functional
+/// after stress or error scenarios.
+pub fn verifyServerFunctional(port: u16) !void {
+    var client = try RawClient.connect(testing.allocator, port);
+    defer client.deinit();
+
+    var msg = "hi".*;
+    try client.write(&msg);
```
InKryption, line 15: There's no reason this should have to be a mutable slice. The corresponding chain of function calls leading to `writeFrameEx` should all be accepting `[]const u8`.

src/rpc/webzockets/e2e_tests/tests.zig
```diff
@@ -0,0 +1,11 @@
+const std = @import("std");
+
+test {
+    // Uncomment to see logs during tests
+    // std.testing.log_level = .debug;
+}
```
InKryption: This isn't how this works; `std.testing.log_level` is reset for every test under the standard test runner.

src/rpc/webzockets/examples/simple_client.zig
```diff
@@ -0,0 +1,124 @@
...
+};
+
+const SimpleClient = ws.Client(ClientHandler, 4096);
+
+fn usage(exe_name: []const u8) void {
+    std.debug.print(
+        "Usage: {s} [ip] [port] [message]\n" ++
+            "Defaults: ip=127.0.0.1 port=8080 message=hello\n",
+        .{exe_name},
+    );
```
InKryption:
```suggestion
    std.debug.print((
        \\Usage: {s} [ip] [port] [message]
        \\Defaults: ip=127.0.0.1 port=8080 message=hello
        \\
    ), .{exe_name});
```

src/rpc/webzockets/examples/simple_client.zig
```diff
@@ -0,0 +1,124 @@
...
+    try loop.run(.until_done);
+
+    if (handler.received) |data| {
+        defer allocator.free(data);
+        std.debug.print("Response: {s}\n", .{data});
+    } else {
+        std.debug.print("No response received\n", .{});
+    }
+
+    conn.deinit();
```
InKryption: Any reason for this `conn.deinit()` to go at the end, and not in a `defer` right after initializing `client`?

src/rpc/webzockets/src/client/client.zig
```diff
@@ -0,0 +1,282 @@
...
+///    │                     ├───────────────────────>│──── WS frame ──────>│
+///    │                     │                        │                     │
+///    │                     │                        │<──── WS frame ──────│
+///    │                     │   onReadCallback       │                     │
+///    │  handler.onMessage()│<───────────────────────│                     │
+///    │<────────────────────│                        │                     │
+/// ```
+///
+/// On failure at any stage, the socket is closed and `Handler.onSocketClose`
+/// is called (if declared).
```
InKryption, line 81: I dislike `@hasDecl` logic like this, it gets quite complex if you start layering it at all. I would request just requiring the type to implement `onSocketClose` unconditionally.

src/rpc/webzockets/src/client/client.zig
```diff
@@ -0,0 +1,282 @@
...
+pub fn Client(comptime Handler: type, comptime read_buf_size: usize) type {
+    return struct {
+        // -- Caller-provided state --
+        conn: *Conn,
+        handler: *Handler,
```
InKryption, line 86: Doesn't `conn` already store `handler`? I would remove `handler` and just refer to `conn.handler`.

Would you be open to making these types intrusive, in order to avoid the very awkward management of self-referential fields? It would entail making `ClientConnection` refer to its handler field as `const handler: *Handler = @fieldParentPtr(field_name, conn_ptr);`, instead of storing it as a field. If access to the actual handler/context pointers are needed outside of the types themselves (which likely you will for `handler.onSocketClose()`, for example), you could add some methods which just do this `@fieldParentPtr` operation.

src/rpc/webzockets/src/client/client.zig
```diff
@@ -0,0 +1,282 @@
...
+            self.socket.connect(
+                self.loop,
+                &self.connect_completion,
+                self.config.address,
+                ClientSelf,
+                self,
+                onConnectComplete,
+            );
+            log.debug("connect: connect submitted to loop", .{});
+        }
```
InKryption, line 179: Could we move this into `init`?

src/rpc/webzockets/src/client/client.zig
```diff
@@ -0,0 +1,282 @@
...
+            handler: *Handler,
+            conn: *Conn,
+            csprng: *types.ClientMaskPRNG,
+            config: Config,
+        ) ClientSelf {
+            // The kqueue and epoll backends require a thread pool to be set on
+            // the loop, otherwise they cannot perform socket close operations.
+            if (comptime @hasField(xev.Loop, "thread_pool")) {
+                std.debug.assert(loop.thread_pool != null);
+            }
```
InKryption, line 149: This seems slightly fragile. I would maybe instead do this:
```suggestion
            // The kqueue and epoll backends require a thread pool to be set on
            // the loop, otherwise they cannot perform socket close operations.
            switch (xev.backend) {
                .io_uring, .wasi_poll, .iocp => comptime std.debug.assert(!@hasDecl(xev.Loop, "thread_pool")),
                .epoll, .kqueue => std.debug.assert(loop.thread_pool != null),
            }
```
otherwise we wouldn't detect if xev was refactored to move or rename this thread_pool field.

src/rpc/webzockets/src/client/client.zig
```diff
@@ -0,0 +1,282 @@
...
+        ///
+        /// Must be pointer-stable and outlive the `ClientConnection`. The PRNG
+        /// is mutated and is not thread-safe; only use it from the `loop.run()`
+        /// thread and do not share it across loops/threads.
+        csprng: *types.ClientMaskPRNG,
+
+        // -- Transient handshake state --
+        hs: Handshake,
+        connect_completion: xev.Completion,
+        socket: xev.TCP,
```
InKryption, line 102: I see this in both this struct, and in the `ClientConnection` struct. Couldn't we just refer to `conn.csprng` instead of storing it redundantly?
Same goes for `allocator`, `loop`, and `socket`.

src/rpc/webzockets/src/client/connection.zig
```diff
@@ -0,0 +1,926 @@
...
+        const log = std.log.scoped(.client_connection);
+
+        /// Comptime flags for optional handler capabilities.
+        const has = struct {
+            const on_open = @hasDecl(Handler, "onOpen");
+            const on_ping = @hasDecl(Handler, "onPing");
+            const on_pong = @hasDecl(Handler, "onPong");
+            const on_socket_close = @hasDecl(Handler, "onSocketClose");
+            const on_bytes_read = @hasDecl(Handler, "onBytesRead");
+        };
```
InKryption, line 112: Similar to my other comment about it, if we're going to do ducktyped interfaces like this, let's avoid conditional hasDecls like this. It leaves the door open to silly stringly-typed mistakes.

src/rpc/webzockets/src/client/client.zig
```diff
@@ -0,0 +1,282 @@
+const std = @import("std");
+const xev = @import("xev");
+
+const types = @import("../types.zig");
+const client_connection = @import("connection.zig");
```
InKryption, line 5: Broad question, now that I've looked over a good bit of the related code: is there a reason these are defined separately? It seems like a lot of state in `ClientConnection` and `Client` are just duplicates of each other, and much of the state in the former has to be directly managed by the latter. With all of my other comments on this file in mind, would you be open to merging these types?

src/rpc/webzockets/src/client/client.zig
```diff
@@ -0,0 +1,282 @@
...
+            // TCP connected — start WebSocket handshake using connection's read buffer
+            // Safe: connection isn't started yet, so read_buf is unused
+            self.hs = Handshake.init(
+                self.socket,
+                self.loop,
+                &self.conn.read_buf,
+                self,
+                self.csprng,
+            );
+            self.hs.start(self.config.address, self.config.path);
```
InKryption, line 217: I think it would be better to just pass most of the duplicated fields initialized through `Handshake.init` directly to `Handshake.start` here instead.

src/rpc/webzockets/src/server/connection.zig
```diff
@@ -0,0 +1,1033 @@
...
+    comptime read_buf_size: usize,
+) type {
+    comptime {
+        if (!@hasDecl(Handler, "onMessage"))
+            @compileError("Handler must declare an onMessage method");
+        if (!@hasDecl(Handler, "onWriteComplete"))
+            @compileError("Handler must declare an onWriteComplete method (sends are zero-copy)");
+        if (!@hasDecl(Handler, "onClose"))
+            @compileError("Handler must declare an onClose method");
+    }
```
InKryption, line 54: Some comment about `@hasDecl`

src/rpc/webzockets/src/server/server.zig
```diff
@@ -0,0 +1,394 @@
+const std = @import("std");
+const xev = @import("xev");
+
+const slot_pool = @import("slot_pool.zig");
+const server_hs = @import("handshake.zig");
+const server_conn = @import("connection.zig");
```
InKryption, line 6: Same comment as on `ClientConnection`.

src/rpc/webzockets/src/frame.zig
```diff
@@ -0,0 +1,1091 @@
...
+/// Server frames are never masked.
+///
+/// Usage:
+/// ```
+/// const framed = comptime frame.frameStatic(.text, "Hello");
+/// ```
+pub fn frameStatic(
+    comptime opcode: Opcode,
+    comptime msg: []const u8,
+) [calculateFrameLen(msg.len)]u8 {
```
InKryption: Why does this function exist? It doesn't seem that useful.

src/rpc/webzockets/src/frame.zig
```diff
@@ -0,0 +1,1091 @@
...
+
+// --- Fuzz tests ---
+
+test "fuzz parseHeader: no crash on arbitrary input" {
+    try testing.fuzz({}, struct {
+        fn run(_: void, input: []const u8) anyerror!void {
+            // parseHeader must either return a valid header or an error, never crash.
+            _ = parseHeader(input) catch return;
+        }
+    }.run, .{
```
InKryption:
```suggestion
    const helper = struct {
        fn run(_: void, input: []const u8) anyerror!void {
            // parseHeader must either return a valid header or an error, never crash.
            _ = parseHeader(input) catch return;
        }
    };
    try testing.fuzz({}, helper.run, .{
```

src/rpc/webzockets/src/http.zig
```diff
@@ -0,0 +1,666 @@
...
+    protocols: []const u8,
+    /// Sec-WebSocket-Extensions header value (empty slice if absent).
+    extensions: []const u8,
+};
+
+/// Bitmask bits for tracking required headers during parsing.
+const required_upgrade: u4 = 0b0001;
+const required_connection: u4 = 0b0010;
+const required_key: u4 = 0b0100;
+const required_version: u4 = 0b1000;
```
InKryption: I would replace this with a packed struct, e.g.
```zig
const FoundRequired = packed struct(u4) {
    upgrade: bool = false,
    connection: bool = false,
    key: bool = false,
    version: bool = false,
};
```

src/rpc/webzockets/src/http.zig
```diff
@@ -0,0 +1,666 @@
...
+    var decoded: [16]u8 = undefined;
+    try std.base64.standard.Decoder.decode(&decoded, key);
+    try testing.expectEqualSlices(u8, &raw, &decoded);
+}
+
+test "encodeKey: different inputs produce different keys" {
+    var buf1: [24]u8 = undefined;
+    var buf2: [24]u8 = undefined;
+    const raw1 = [_]u8{0x01} ** 16;
+    const raw2 = [_]u8{0x02} ** 16;
```
InKryption:
```suggestion
    const raw1: [16]u8 = @splat(0x01);
    const raw2: [16]u8 = @splat(0x02);
```

src/rpc/webzockets/src/mask.zig
```diff
@@ -0,0 +1,136 @@
...
+        payload[i] = b ^ mask_key[i & 3];
+    }
+}
+
+const testing = std.testing;
+
+test "mask: empty payload is no-op" {
+    var buf = [_]u8{};
+    mask(.{ 0xAA, 0xBB, 0xCC, 0xDD }, &buf);
+}
```
InKryption, line 34: I don't quite understand what this is testing.

src/rpc/webzockets/src/reader.zig
InKryption: I would tend to agree with the original comment here, I don't quite understand why we need this.

src/rpc/webzockets/src/root.zig
```diff
@@ -0,0 +1,35 @@
...
+pub const Role = types.Role;
+pub const Server = server.Server;
+pub const ClientHandshake = client_handshake.ClientHandshake;
+pub const ClientConnection = client_connection.ClientConnection;
+pub const Client = client.Client;
+pub const ClientMaskPRNG = types.ClientMaskPRNG;
+
+test {
+    // Uncomment to see logs during tests
+    // std.testing.log_level = .debug;
```
InKryption: Same comment as on other instance of this.

src/rpc/webzockets/src/root.zig
InKryption: Generally yes. In the future the situation will be improved by overhauling the test discovery mechanism, but for now we'll just have to deal with it. I think using refAllDecls in the meantime is kind of alright for some situations, but if we can avoid it that's better.

src/rpc/webzockets/src/reader.zig
jbuckmccready: Comments have been cleaned up. We need a read buffer of some kind, what is the suggested change here?

src/rpc/webzockets/e2e_tests/tests.zig
jbuckmccready: I used this extensively to debug failing tests, and just tested it again now to confirm my sanity, here is the behavior (on Zig 0.14.1).

- Without setting `std.testing.log_level`: test fails -> no debug logs.
- With setting `std.testing.log_level = .debug;`: test fails -> shows debug logs.
- With setting `std.testing.log_level = .debug`: test passes -> no debug logs (I was using it for failing tests).

src/rpc/webzockets/src/root.zig
jbuckmccready: https://github.com/Syndica/sig/commit/c3c253756a3091f83f6a6108506478f27bec0c9c

src/rpc/webzockets/src/client/client.zig
jbuckmccready: It probably can be simplified while still retaining the original intent, it's kind of roughly thrown together due to it being unused for actual jrpc websocket implementation in the validator. Let's do quick huddle on it so I can learn.

Roughly the idea is:
- `Client`: orchestrates TCP connect and websocket handshake, state associated with this whole process is transient and can be pooled or thrown away after creating an upgraded connection at the caller's discretion.
- `ClientConnection`: actual upgraded websocket connection, caller uses this to do things, also may be recycled and reused with `Client`.

Ideally the upgraded `ClientConnection` should not have any state dragged around that is only for the handshake, which is what it currently accomplishes, but it's pretty hacked.

src/rpc/webzockets/autobahn/client/client.zig
jbuckmccready: https://github.com/Syndica/sig/commit/76aef919092f6378482ae601b60d216ae07ca0de

src/rpc/webzockets/autobahn/client/client.zig
jbuckmccready: https://github.com/Syndica/sig/commit/76aef919092f6378482ae601b60d216ae07ca0de

src/rpc/webzockets/autobahn/client/client.zig
jbuckmccready: Updated autobahn server to avoid requirement of linking with libc, but didn't add code/logic for graceful exit to check for leaks: https://github.com/Syndica/sig/commit/2f08d64f305fc49a3856d85c80fa2a3baa1bd5f6

src/rpc/webzockets/e2e_tests/client/close_tests.zig
jbuckmccready: https://github.com/Syndica/sig/commit/78b06d6af68e3a54c31696042c29e00b7f3e3f7f

src/rpc/webzockets/e2e_tests/client/close_tests.zig
jbuckmccready: https://github.com/Syndica/sig/commit/78b06d6af68e3a54c31696042c29e00b7f3e3f7f

src/rpc/webzockets/e2e_tests/client/close_tests.zig
jbuckmccready: The tests were originally alongside other tests that were more like unit tests and file name they were in didn't categorize them but after getting reorganized I kept the naming pattern which is redundant.

I notice it doesn't print the top level directory with how the build step works now  (missing `e2e_tests` prefix):
`error: 'client.close_tests.test.e2e client close: server-initiated close disconnects cleanly' failed`

But it has the full file path which includes it so I removed the redundant prefixes on the tests: https://github.com/Syndica/sig/commit/b6c2264f8fb19949e76946107167561192b23186

src/rpc/webzockets/autobahn/client/client.zig
jbuckmccready: Left over from old code that used it, wasn't sure if I was going to need it again, and then I forgot to cleanup.

src/rpc/webzockets/e2e_tests/support/client_handlers.zig
jbuckmccready: Yeah definitely shouldn't swallow the errors. Not sure what the best decision is here if using the builtin test runner:
- Panic which will give stack trace and fail the test, but also stops other tests from running 
- `std.log.err` shows test as failing but confusingly also states that it passed in the final count summary
- Track the error state in a field on the handler then assert it within the test but that requires more code and loses the call stack information

Is there another way, or which of these do you think is preferred pattern?

src/rpc/webzockets/e2e_tests/support/fd_leak.zig
jbuckmccready: https://github.com/Syndica/sig/commit/9344840de5c506932d0b7bcaa2cae5d37611831d

src/rpc/webzockets/e2e_tests/support/fd_leak.zig
jbuckmccready: https://github.com/Syndica/sig/commit/9344840de5c506932d0b7bcaa2cae5d37611831d

src/rpc/webzockets/e2e_tests/support/fd_leak.zig
jbuckmccready: https://github.com/Syndica/sig/commit/9344840de5c506932d0b7bcaa2cae5d37611831d

src/rpc/webzockets/e2e_tests/support/fd_leak.zig
jbuckmccready: I kept the same panic semantics in tests, I'm mixed on it: with panic it stops other tests from running, without panic it shows tests as successful in final counts (but does show test failure from error log) which is confusing.

src/rpc/webzockets/e2e_tests/support/raw_client.zig
jbuckmccready: Nice! I didn't know about this, updated another location I needed it as well. https://github.com/Syndica/sig/commit/269acd9a65d56b07e7fddeda7ea80fd037459f98

src/rpc/webzockets/e2e_tests/support/raw_client.zig
jbuckmccready: https://github.com/Syndica/sig/commit/269acd9a65d56b07e7fddeda7ea80fd037459f98

src/rpc/webzockets/e2e_tests/support/raw_client.zig
jbuckmccready: https://github.com/Syndica/sig/commit/269acd9a65d56b07e7fddeda7ea80fd037459f98

src/rpc/webzockets/e2e_tests/support/raw_client.zig
jbuckmccready: Remaining tests cleaned up to use `std.testing.random_seed`: https://github.com/Syndica/sig/commit/c922088398750029960925624d523b869a26630e

src/rpc/webzockets/e2e_tests/support/raw_client.zig
jbuckmccready: Ah they are i32 from system on MacOS: https://app.circleci.com/pipelines/github/Syndica/sig/7638/workflows/45552a03-18f8-4070-9206-52e7f357e9a7/jobs/122722

Reverted: https://github.com/Syndica/sig/pull/1222/changes/f463465707a0be9baaf77b5426621d5bc8f73eb9

src/rpc/webzockets/e2e_tests/support/server_runner.zig
jbuckmccready: https://github.com/Syndica/sig/commit/874f4ac8784aa94e2ed9ddef18391eb2aad51efe

src/rpc/webzockets/e2e_tests/support/test_clients.zig
jbuckmccready: https://github.com/Syndica/sig/commit/72b911c49ec9249cebd16af4629fa2616dd08ec8

src/rpc/webzockets/e2e_tests/support/test_helpers.zig
jbuckmccready: Currently it masks the bytes in place before writing to the TCP stream. I can change to dupe allocate or leave as is.

src/rpc/webzockets/examples/simple_client.zig
jbuckmccready: It only gets initialized on the success path of `client.connect()`, which is weird, this ties in with simplifying the client API/structures. I put it where it should currently go: https://github.com/Syndica/sig/commit/e6b0d918f4092f1e8bb96f618079cd820c7f69fe

But API/structures should likely change.

src/rpc/webzockets/examples/simple_client.zig
jbuckmccready: https://github.com/Syndica/sig/commit/e6b0d918f4092f1e8bb96f618079cd820c7f69fe

src/rpc/webzockets/src/frame.zig
jbuckmccready: I agree, not useful enough to add it, can add it back if it's found needed.

jbuckmccready: https://github.com/Syndica/sig/commit/db2c03e4e8d37dc324e0740f21cadb2f5226b3a4

src/rpc/webzockets/src/frame.zig
jbuckmccready: https://github.com/Syndica/sig/commit/db2c03e4e8d37dc324e0740f21cadb2f5226b3a4

src/rpc/webzockets/src/http.zig
jbuckmccready: Much nicer.

https://github.com/Syndica/sig/commit/49f5cf5d99e786211d0cff4973c00e5103343c60

src/rpc/webzockets/src/http.zig
jbuckmccready: https://github.com/Syndica/sig/commit/49f5cf5d99e786211d0cff4973c00e5103343c60

src/rpc/webzockets/src/mask.zig
jbuckmccready: Passing in empty payload/buffer to be masked doesn't panic/cause problems.

src/rpc/webzockets/e2e_tests/support/client_handlers.zig
InKryption: Is there any possibility you could adjust the `onOpen` callback to allow for it to return an error?

src/rpc/webzockets/e2e_tests/support/fd_leak.zig
InKryption: When a file is a struct, the convention is to name it the same way as the type itself, ie in this case `FdLeakDetector.zig`

src/rpc/webzockets/e2e_tests/support/fd_leak.zig
InKryption: I think it's fine if tests "pass" with an error log, since that will still alert us to something going wrong. For tests testing the leak detector itself, or where you really want to mark failure to free all fds as a test failure, I would add a `try std.testing.expectEqual(.ok, fd_check.check());`

src/rpc/webzockets/e2e_tests/support/test_helpers.zig
InKryption: Ah, I missed that specific part of the code. Is there no way that the `mask` function could be refactored to stream the results directly to a writer? Once we move to 0.15 it would also very easily permit re-SIMDifying it again using the writer's mutable buffer API.

src/rpc/webzockets/e2e_tests/tests.zig
InKryption: That doesn't make a lot of sense to me, in `lib/compiler/test_runner.zig:207:9`, you can see that before running every test, it sets `testing.log_level = .warn;` (in zig 0.14.x). If this test block is working, then I believe it is a bug, and we should not be relying on it.

src/rpc/webzockets/src/mask.zig
InKryption: Well, just a note then: `&.{}` can be used to initialize mutable buffers.

src/rpc/webzockets/e2e_tests/support/fd_leak.zig
jbuckmccready: Brain fart when updating, fixed: https://github.com/Syndica/sig/pull/1222/commits/4fb8fa51884f0b51d08fa1d6538c9df6979c8654

src/rpc/webzockets/e2e_tests/support/fd_leak.zig
jbuckmccready: Aligned with gpa API and avoiding panics (std.log.err still causes non-zero exit when running the test): https://github.com/Syndica/sig/pull/1222/commits/d5c6d40dbbd9f288d052af1e60672f2c014d9ef9

src/rpc/webzockets/src/mask.zig
jbuckmccready: Made the minor cleanup.

src/rpc/webzockets/e2e_tests/tests.zig
jbuckmccready: Removed the comments as it likely wont work in the future when bug is fixed.

https://github.com/Syndica/sig/pull/1222/commits/a0f1a460df0e86883412f6d9b0c5bd9f79459eb0

src/rpc/webzockets/src/reader.zig
InKryption: With the merge of the 0.15 update, I think it would be worth looking at trying to use `std.Io.Reader` instead, and simplifying some of the requirements. My gut instinct is that most of what we actually need can be expressed through that interface.

docs/docusaurus/docs/code/webzockets.md
```diff
@@ -0,0 +1,271 @@
+# Webzockets
+
+A WebSocket (RFC 6455) library for Zig 0.14.1, built on `libxev`. Server and client. No hidden allocators — all memory is allocated through caller-provided allocators with caller-configured pool sizes and limits. Sends are zero-copy. Server connections are memory-pooled; client connections are caller-owned.
```
yewman: Version bump?
```suggestion
A WebSocket (RFC 6455) library for Zig 0.15.2, built on `libxev`. Server and client. No hidden allocators — all memory is allocated through caller-provided allocators with caller-configured pool sizes and limits. Sends are zero-copy. Server connections are memory-pooled; client connections are caller-owned.
```

src/rpc/webzockets/e2e_tests/support/client_handlers.zig
jbuckmccready: https://github.com/Syndica/sig/commit/3bca7af161837c5318ce469068dfd53af10cb110
Just used unreachable to simplify things, it's just a test helper and these should never fail, call stack will be logged.

src/rpc/webzockets/src/client/client.zig
jbuckmccready: https://github.com/Syndica/sig/commit/e3d90bfcc77122ef017d4a7befa0a7bd4f82f1f0

docs/docusaurus/docs/code/webzockets.md
jbuckmccready: https://github.com/Syndica/sig/commit/ffa94403783bb5aa1b3a8bada7de83e8dcee0b01

src/rpc/webzockets/src/client/client.zig
jbuckmccready: https://github.com/Syndica/sig/issues/1274

src/rpc/webzockets/src/client/client.zig
jbuckmccready: https://github.com/Syndica/sig/issues/1274

src/rpc/webzockets/src/client/client.zig
jbuckmccready: https://github.com/Syndica/sig/issues/1274

src/rpc/webzockets/e2e_tests/support/test_helpers.zig
jbuckmccready: https://github.com/Syndica/sig/issues/1274

src/rpc/webzockets/src/client/client.zig
jbuckmccready: https://github.com/Syndica/sig/issues/1274

src/rpc/webzockets/src/client/connection.zig
jbuckmccready: https://github.com/Syndica/sig/issues/1274

src/rpc/webzockets/src/client/client.zig
jbuckmccready: https://github.com/Syndica/sig/issues/1274

src/rpc/webzockets/src/client/client.zig
jbuckmccready: https://github.com/Syndica/sig/issues/1274

src/rpc/webzockets/src/server/connection.zig
jbuckmccready: https://github.com/Syndica/sig/issues/1274

src/rpc/webzockets/src/server/server.zig
jbuckmccready: https://github.com/Syndica/sig/issues/1274

src/rpc/webzockets/src/reader.zig
jbuckmccready: https://github.com/Syndica/sig/issues/1274

src/rpc/webzockets/e2e_tests/support/client_handlers.zig
InKryption: My concern was moreso with the way these callbacks are invoked by the interface that they implement, ie making that `try` the functions, but `catch unreachable` should be easy enough to grep for and fix for now - do note, however, unreachable is still unchecked illegal behavior in release even inside tests.

jbuckmccready: Updated https://github.com/Syndica/sig/issues/1274 with bullet on this.

PR #1232, opened 2026-02-17
refactor the vote state with union type

hamza-syndica:
Refactored VoteState to use a union type, extracted into a dedicated `state.zig` module. Updated all consumers across consensus, replay, rewards, and runtime to use the new type.
Changes

- Introduced `src/runtime/program/vote/state.zig` (around ~1000 lines) with union-based vote state representation
- Refactored `vote/execute.zig` to use the new state types
- Updated stake, bank, leader schedule, and reward calculation modules accordingly
- Added tests for the new vote state logic

### DIFF
--- a/src/runtime/program/vote/state.zig
+++ b/src/runtime/program/vote/state.zig
@@ -6316,6 +6316,7 @@ test "state.VoteState.delegating methods" {
             allocator,
             Pubkey.ZEROES,
             Pubkey.ZEROES,
+            Pubkey.ZEROES,
             0,
             0,
         ) };


dnut (dismissed):

dnut (approved):

src/runtime/program/vote/state.zig
```diff
@@ -792,6 +881,509 @@ pub const VoteStateVersions = union(enum(u32)) {
...
+            .v4 => |*s| s.inflation_rewards_commission_bps = @as(u16, commission_pct) * 100,
+        }
+    }
+
+    pub fn votes(self: *const VoteState) *const std.ArrayListUnmanaged(LandedVote) {
+        return switch (self.*) {
+            .v3 => |*s| &s.votes,
+            .v4 => |*s| &s.votes,
+        };
+    }
```
InKryption: There's not much sense in returning a constant referenco the arraylist.
```suggestion
    pub fn votes(self: *const VoteState) []const LandedVote {
        return switch (self.*) {
            inline .v3, .v4 => |s| s.votes.items,
        };
    }
```

src/runtime/program/vote/state.zig
```diff
@@ -792,6 +881,509 @@ pub const VoteStateVersions = union(enum(u32)) {
...
+            .v4 => |*s| &s.authorized_voters,
+        };
+    }
+
+    pub fn epochCreditsList(self: *const VoteState) *const std.ArrayListUnmanaged(EpochCredit) {
+        return switch (self.*) {
+            .v3 => |*s| &s.epoch_credits,
+            .v4 => |*s| &s.epoch_credits,
+        };
+    }
```
InKryption: Ditto
```suggestion
    pub fn epochCreditsList(self: *const VoteState) []const EpochCredit {
        return switch (self.*) {
            inline .v3, .v4 => |s| s.epoch_credits.items,
        };
    }
```

src/runtime/program/vote/state.zig
```diff
@@ -792,6 +881,509 @@ pub const VoteStateVersions = union(enum(u32)) {
...
+            .v3 => |s| s.isUninitialized(),
+            .v4 => |s| s.isUninitialized(),
+        };
+    }
+
+    pub fn epochCredits(self: *const VoteState) u64 {
+        return switch (self.*) {
+            inline else => |*s| s.epochCredits(),
+        };
+    }
```
InKryption, line 1071: I would use `inline .v3, .v4` instead of `inline else`. Expresses intent more clearly.

src/runtime/program/vote/state.zig
```diff
@@ -792,6 +881,509 @@ pub const VoteStateVersions = union(enum(u32)) {
...
+
+    // ── Version queries ──────────────────────────────────────────────
+
+    pub fn isV4(self: VoteState) bool {
+        return self == .v4;
+    }
+
+    pub fn isV3(self: VoteState) bool {
+        return self == .v3;
+    }
```
InKryption: These seem somewhat pointless compared to the caller just writing `vote_state == .<tag>`.

src/runtime/program/vote/state.zig
```diff
@@ -792,6 +881,509 @@ pub const VoteStateVersions = union(enum(u32)) {
...
+            .v4 => null,
+        };
+    }
+
+    pub fn asV4Const(self: *const VoteState) ?*const VoteStateV4 {
+        return switch (self.*) {
+            .v3 => null,
+            .v4 => |*s| s,
+        };
+    }
```
InKryption: I don't see the benefit of these methods over just directly `switch`ing over the vote state.

src/runtime/program/vote/state.zig
```diff
@@ -792,6 +881,509 @@ pub const VoteStateVersions = union(enum(u32)) {
...
+        return .{ .v4 = try VoteStateV4.init(
+            allocator,
+            node_pubkey,
+            authorized_voter,
+            withdrawer,
+            commission_pct,
+            voter_epoch,
+            vote_pubkey,
+        ) };
+    }
```
InKryption: These initialization methods don't seem very purposeful. The caller can just as easily write `.{ .v4 = try .init(all, the, same, parameters) }`. You even bypass these just a hundred or so lines prior, so that you can initialize the payload directly with managed memory.

src/runtime/program/stake/lib.zig
```diff
@@ -681,6 +682,8 @@ fn delegate(
             vote_account.deserializeFromAccountData(allocator, VoteStateVersions),
         };
     };
+    var vote_state_to_deinit: ?*const anyerror!VoteStateVersions = &vote_state;
+    if (vote_state_to_deinit) |fallible| if (fallible.*) |*vs| vs.deinit(allocator) else |_| {};
```
InKryption: Is this missing a `defer`? Or otherwise, what's going on here.

src/runtime/program/vote/execute.zig
```diff
@@ -1179,60 +1036,58 @@ fn validateIsSigner(
...
+        .v4 => |v4_state| {
+            // [SIMD-0185] v4: resize to 3762 if smaller, then check rent exempt, then serialize v4.
+            if (account.constAccountData().len < VoteStateV4.MAX_VOTE_STATE_SIZE) {
+                if (std.meta.isError(account.setDataLength(
+                    allocator,
+                    resize_delta,
+                    VoteStateV4.MAX_VOTE_STATE_SIZE,
+                ))) {
+                    return InstructionError.AccountNotRentExempt;
+                }
```
InKryption: I highly advise advise against using `std.meta` in general unless there is no better alternative:

```suggestion
                try account.setDataLength(
                    allocator,
                    resize_delta,
                    VoteStateV4.MAX_VOTE_STATE_SIZE,
                );
```
I would have suggested `catch return error.AccountNotRentExempt;`, but I noticed that this function returns `Allocator.Error || InstructionError`, which is identical to the enclosing function's error set. Is there any reason not to propagate it directly?

src/runtime/program/vote/execute.zig
```diff
@@ -1179,60 +1036,58 @@ fn validateIsSigner(
...
+        },
+        .v3 => |v3_state| {
+            if (account.constAccountData().len < VoteStateV3.MAX_VOTE_STATE_SIZE and
+                (!rent.isExempt(account.account.lamports, VoteStateV3.MAX_VOTE_STATE_SIZE) or
+                    std.meta.isError(account.setDataLength(
+                        allocator,
+                        resize_delta,
+                        VoteStateV3.MAX_VOTE_STATE_SIZE,
+                    ))))
+            {
```
InKryption: This is a bad idea, the `isError` will treat `error.OutOfMemory` equally to an `InstructionError`.

src/runtime/program/vote/state.zig
dnut: when would you use `inline else`?

i think the intent is clear either way. there's an expectation for all the types to have this method. if not, it will fail to compile. what's the problem?

InKryption: I think there's a threshold of complexity. Right now there are exactly two members possibilities which are identical. If we add a new one, I think it's desirable for the compile error to say "missing case in switch", rather than it complaining about a missing declaration. If there came a point at which there are so many members that listing out every case in every site becomes mental overhead unto itself, I think `inline else` becomes justifiable.

Just generally speaking, the use of `inline else` implies a ducktyping-defined contract, i.e. some sort of interface, and I think it is preferable to avoid writing contracts incidentally like that.

PR #1236, opened 2026-02-18
refactor(vm): align memory allocations to 16 bytes

prestonsn:
- Use alignedAlloc for ro_section, heap, stack, and regions
- Simplify sysvar syscall by using direct assignment after memset

### DIFF
--- a/src/vm/lib.zig
+++ b/src/vm/lib.zig
@@ -73,8 +73,8 @@ pub const EbpfError = error{
 
 pub const State = struct {
     vm: Vm,
-    stack: []u8,
-    heap: []u8,
+    stack: []align(16) u8,
+    heap: []align(16) u8,
     regions: []memory.Region,
 
     pub fn deinit(self: *State, allocator: std.mem.Allocator) void {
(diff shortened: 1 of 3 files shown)


src/vm/lib.zig
```diff
@@ -108,16 +108,16 @@ pub fn init(
...
     errdefer allocator.free(heap);
 
-    const stack = try allocator.alloc(u8, stack_size);
+    const stack = try allocator.alignedAlloc(u8, 16, stack_size);
     @memset(stack, 0);
     errdefer allocator.free(stack);
 
     // 3 regions for the input, stack, and heap.
-    const regions = try allocator.alloc(memory.Region, 3 + trailing_regions.len);
+    const regions = try allocator.alignedAlloc(memory.Region, 16, 3 + trailing_regions.len);
```
kprotty: The region slice itself likely doesnt need to be aligned (it holds pointer to host memory e.g. heap/stack, which do need to be aligned)

Sobeston (commented):
I guess this can't hurt, but do we know for sure that this fixes something?

src/vm/lib.zig
yewman: Yeah that makes sense 20bc00f7

PR #1242, opened 2026-02-20
review stricter_abi_and_runtime_constraints

hamza-syndica:
- added: stricter_abi_and_runtime_constraints check in new places

### DIFF
--- a/src/vm/syscalls/sysvar.zig
+++ b/src/vm/syscalls/sysvar.zig
@@ -16,8 +16,6 @@ const RegisterMap = sig.vm.interpreter.RegisterMap;
 const SyscallError = sig.vm.SyscallError;
 const TransactionContext = sig.runtime.TransactionContext;
 
-const MM_INPUT_START = memory.INPUT_START;
-
 // https://github.com/anza-xyz/agave/blob/master/programs/bpf_loader/src/syscalls/sysvar.rs#L164
 const SYSVAR_NOT_FOUND = 2;
 // https://github.com/anza-xyz/agave/blob/master/programs/bpf_loader/src/syscalls/sysvar.rs#L165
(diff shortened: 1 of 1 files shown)


src/vm/syscalls/sysvar.zig
```diff
@@ -13,8 +13,11 @@ const Pubkey = sig.core.Pubkey;
 const MemoryMap = memory.MemoryMap;
 const InstructionError = sig.core.instruction.InstructionError;
 const RegisterMap = sig.vm.interpreter.RegisterMap;
+const SyscallError = sig.vm.SyscallError;
 const TransactionContext = sig.runtime.TransactionContext;
 
+const MM_INPUT_START = memory.INPUT_START;
+
```
prestonsn: nit: I think this alias is a little unnecessary (only other place it's used is in the unit tests, where it's referred to via `memory.INPUT_START` anyways). 

I think other files in the codebase also use `memory.INPUT_START` too

src/vm/syscalls/sysvar.zig
```diff
@@ -31,6 +34,16 @@ fn getSyscall(comptime T: type) fn (*TransactionContext, *MemoryMap, *RegisterMa
...
             const value_addr = registers.get(.r1);
+
+            // SIMD-0219: The destination address of all sysvar related syscalls
+            // must be on the stack or heap, meaning their virtual address is
+            // inside `0x200000000..0x400000000`.
+            if (value_addr >= MM_INPUT_START and
+                tc.feature_set.active(.stricter_abi_and_runtime_constraints, tc.slot))
+            {
+                return SyscallError.InvalidPointer;
+            }
```
prestonsn, line 45: I like this code comment, but could also mention this for clarity: 

> (bytecode/rodata regions below 0x200000000 are always readonly, so mutable translation to addresses below that will result in an `AccessViolation` anyways).

Agave: https://github.com/anza-xyz/agave/blob/64aa73ba52a9eb9ceb55b4b9adf654ae8ef52b9a/syscalls/src/sysvar.rs#L21-L27

src/vm/syscalls/sysvar.zig
```diff
@@ -72,6 +85,15 @@ pub fn getSysvar(
     const mem_cost = @max(tc.compute_budget.mem_op_base_cost, buf_cost);
     try tc.consumeCompute(tc.compute_budget.sysvar_base_cost +| id_cost +| mem_cost);
 
+    // SIMD-0219: The destination address of all sysvar related syscalls
+    // must be on the stack or heap, meaning their virtual address is
+    // inside `0x200000000..0x400000000`.
```
prestonsn, line 90: Same here, minor nit about the comment clarity

PR #1245, opened 2026-02-23
feat(rpc): implement getIdentity 

hamza-syndica:


hamza-syndica:
@dnut the `getIdentity` is already implemented on `main`

PR #1247, opened 2026-02-23
feat(rpc): implement getHealth

hamza-syndica:
Implement `getHealth` RPC method

- Implements the getHealth JSON-RPC method that checks if the node is healthy by comparing the node's latest optimistically confirmed slot against the cluster's slot from gossip (analogous to RpcHealth::check (https://github.com/anza-xyz/agave/blob/master/rpc/src/rpc_health.rs#L44-L107))

- Adds proper JSON-RPC error responses (code -32005) for unhealthy states, matching Agave's NodeUnhealthy error (https://github.com/anza-xyz/agave/blob/master/rpc-client-api/src/custom_error.rs#L149-L159)

- Updates HTTP GET /health endpoint to return actual health status (ok, behind, unknown) instead of hardcoded unknown (matching rpc_service.rs (https://github.com/anza-xyz/agave/blob/master/rpc/src/rpc_service.rs#L332-L340))

### DIFF
--- a/src/rpc/methods.zig
+++ b/src/rpc/methods.zig
@@ -45,14 +45,14 @@ pub const MethodAndParams = union(enum) {
     getBlock: GetBlock,
     getBlockCommitment: GetBlockCommitment,
     getBlockHeight: GetBlockHeight,
-    getBlockProduction: noreturn,
+    getBlockProduction: GetBlockProduction,
     getBlocks: GetBlocks,
     getBlocksWithLimit: GetBlocksWithLimit,
     getBlockTime: noreturn,
     getClusterNodes: GetClusterNodes,
     getEpochInfo: GetEpochInfo,
     getEpochSchedule: GetEpochSchedule,
-    getFeeForMessage: noreturn,
+    getFeeForMessage: GetFeeForMessage,
     getFirstAvailableBlock: noreturn,
 
     /// https://github.com/Syndica/sig/issues/557
@@ -72,7 +72,7 @@ pub const MethodAndParams = union(enum) {
     getLeaderSchedule: GetLeaderSchedule,
     getMaxRetransmitSlot: noreturn,
     getMaxShredInsertSlot: noreturn,
-    getMinimumBalanceForRentExemption: noreturn,
+    getMinimumBalanceForRentExemption: GetMinimumBalanceForRentExemption,
     getMultipleAccounts: GetMultipleAccounts,
     getProgramAccounts: noreturn,
     getRecentPerformanceSamples: GetRecentPerformanceSamples,
@@ -82,7 +82,7 @@ pub const MethodAndParams = union(enum) {
     getSlot: GetSlot,
     getSlotLeader: noreturn,
     getSlotLeaders: noreturn,
-    getStakeMinimumDelegation: noreturn,
+    getStakeMinimumDelegation: GetStakeMinimumDelegation,
     getSupply: noreturn,
     getTokenAccountBalance: GetTokenAccountBalance,
     getTokenAccountsByDelegate: noreturn,
@@ -848,7 +848,52 @@ pub const GetBlockHeight = struct {
     pub const Response = u64;
 };
 
-// TODO: getBlockProduction
+pub const GetBlockProduction = struct {
+    config: ?Config = null,
+
+    pub const Config = struct {
+        commitment: ?common.Commitment = null,
+        /// Filter results to a single validator identity (base-58 encoded pubkey)
+        identity: ?[]const u8 = null,
+        range: ?Range = null,
+    };
+
+    pub const Range = struct {
+        firstSlot: Slot,
+        lastSlot: ?Slot = null,
+    };
+
+    pub const Response = struct {
+        context: common.Context,
+        value: Value,
+    };
+
+    pub const Value = struct {
+        byIdentity: ByIdentity,
+        range: ResponseRange,
+    };
+
+    pub const ResponseRange = struct {
+        firstSlot: Slot,
+        lastSlot: Slot,
+    };
+
+    /// Map of base58 pubkey string -> [leader_slots, blocks_produced]
+    pub const ByIdentity = struct {
+        map: sig.utils.collections.PubkeyMap(struct { u64, u64 }),
+
+        pub fn jsonStringify(self: ByIdentity, jw: anytype) !void {
+            try jw.beginObject();
+            for (self.map.keys(), self.map.values()) |key, value| {
+                const base58string = key.base58String();
+                try jw.objectField(base58string.constSlice());
+                try jw.write(value);
+            }
+            try jw.endObject();
+        }
+    };
+};
+
 // TODO: getBlockTime
 
 pub const GetBlocks = struct {
@@ -998,6 +1043,24 @@ pub const GetGenesisHash = struct {
 
 // TODO: getHealth
 
+/// [agave] https://github.com/anza-xyz/agave/blob/d70b1714b1153674c16e2b15b68790d274dfe953/rpc/src/rpc.rs#L3580-L3586
+pub const GetFeeForMessage = struct {
+    /// Base64-encoded serialized VersionedMessage
+    message: []const u8,
+    config: ?Config = null,
+
+    pub const Config = struct {
+        commitment: ?common.Commitment = null,
+        minContextSlot: ?Slot = null,
+    };
+
+    pub const Response = struct {
+        context: common.Context,
+        /// Fee in lamports, or null if the blockhash has expired.
+        value: ?u64,
+    };
+};
+
 pub const GetHighestSnapshotSlot = struct {
     pub const Response = ?SnapshotSlotInfo;
 
@@ -1107,7 +1170,17 @@ pub const GetLeaderSchedule = struct {
 
 // TODO: getMaxRetransmitSlot
 // TODO: getMaxShredInsertSlot
-// TODO: getMinimumBalanceForRentExemption
+
+/// Returns minimum balance required to make account rent exempt.
+/// https://solana.com/docs/rpc/http/getminimumbalanceforrentexemption
+pub const GetMinimumBalanceForRentExemption = struct {
+    /// The Account's data length
+    data_len: usize,
+    config: ?common.CommitmentSlotConfig = null,
+
+    /// Returns minimum lamports required in account to remain rent free.
+    pub const Response = u64;
+};
 
 pub const GetMultipleAccounts = struct {
     pubkeys: []const Pubkey,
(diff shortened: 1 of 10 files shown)


src/rpc/methods.zig
```diff
@@ -889,6 +897,73 @@ pub const RpcHookContext = struct {
...
+    ///
+    /// A node is considered healthy if:
+    /// - `override_health_check` is set to true, OR
+    /// - The node's latest optimistically confirmed slot is within `health_check_slot_distance`
+    ///   of the cluster's latest optimistically confirmed slot.
+    ///
+    /// Returns `RpcHealthStatus.ok`, `RpcHealthStatus.unknown`, or `RpcHealthStatus.behind`.
+    ///
+    /// Analogous to [RpcHealth::check](https://github.com/anza-xyz/agave/blob/8803776d/rpc/src/rpc_health.rs#L44-L107)
+    fn checkHealth(self: RpcHookContext, allocator: std.mem.Allocator) RpcHealthStatus {
```
dnut: why is this a separate function from getHealth? can you inline it?

hamza-syndica: Tried to keep the same agave style, you are right. It could've been inline.
Updated: https://github.com/Syndica/sig/pull/1247/commits/f8f785d1f648d03577e19130d9f1cc4bd3806c62

dnut: the agave codebase is a mess. i wouldn't make any stylistic choices just because agave did it

src/rpc/server/server.zig
```diff
@@ -645,3 +645,356 @@ fn testHttpFetchSelf(
...
+        pub fn getHealth(
+            _: anytype,
+            _: std.mem.Allocator,
+            _: sig.rpc.methods.GetHealth,
+        ) !sig.rpc.methods.GetHealth.Response {
+            return .ok;
+        }
+    }{});
+
+    const sock_addr = std.net.Address.initIp4(.{ 0, 0, 0, 0 }, 0);
```
kprotty: nitpick: could be `127.0.0.1` instead of `0.0.0.0` since it doesn't need to accept connections from outside the host machine

src/rpc/server/server.zig
```diff
@@ -645,3 +645,356 @@ fn testHttpFetchSelf(
 
     return response_content;
 }
+
+test "HTTP GET /health returns ok when hook returns ok status" {
```
kprotty, line 589: we should probably be adding `if (sig.build_options.no_network_tests) return error.SkipZigTest;` on top of these

hamza-syndica: https://github.com/Syndica/sig/pull/1247/commits/da9e80de6c84f4c9126f7886aac94e6db82e82fd

src/rpc/server/server.zig
hamza-syndica: https://github.com/Syndica/sig/pull/1247/commits/da9e80de6c84f4c9126f7886aac94e6db82e82fd

src/rpc/methods.zig
```diff
@@ -1514,3 +1581,193 @@ fn JsonSkippable(comptime T: type) type {
...
+    // Analogous to: https://github.com/anza-xyz/agave/blob/8803776d/rpc/src/rpc_health.rs#L163-L164
+    var override = std.atomic.Value(bool).init(true);
+    var slot_tracker = try sig.replay.trackers.SlotTracker.initEmpty(std.testing.allocator, 0);
+    defer slot_tracker.deinit(std.testing.allocator);
+
+    const ctx = testHealthContext(&slot_tracker, null, &override, 10);
+    const status = try ctx.getHealth(std.testing.allocator, .{});
+    try std.testing.expect(status.eql(.ok));
+}
+
```
dnut: This isn't a very meaningful test. It verifies a trivial implementation of getHealth that only exists to be tested in this test. How about we just remove the `override_health_check` parameter and any tests that use it.

There's another test that's functionally identical to this test, and I'm not sure why it exists at all.

btw your permalinks in these tests are dead

src/rpc/methods.zig
```diff
@@ -1485,6 +1493,65 @@ pub const RpcHookContext = struct {
...
+        // Need ledger to check cluster's optimistic slot
+        const ledger = self.ledger orelse return .unknown;
+
+        // Get the node's latest optimistically confirmed slot from replay
+        // This is the slot the node has replayed and validated
+        const my_latest_optimistically_confirmed_slot =
+            self.slot_tracker.latest_confirmed_slot.get();
+
+        // Get the cluster's latest optimistically confirmed slot from blockstore
```
dnut: we don't have anything called a blockstore
```suggestion
        // Get the cluster's latest optimistically confirmed slot from ledger
```

src/rpc/methods.zig
```diff
@@ -1485,6 +1493,65 @@ pub const RpcHookContext = struct {
...
+        // This comes from gossip observations
+        var optimistic_slots = ledger.reader().getLatestOptimisticSlots(allocator, 1) catch {
+            return .unknown;
+        };
+        defer optimistic_slots.deinit();
+
+        if (optimistic_slots.items.len == 0) {
+            // No optimistic slots in blockstore yet
```
dnut:
```suggestion
            // No optimistic slots in ledger yet
```

src/rpc/server/basic.zig
```diff
@@ -134,8 +134,33 @@ fn handleGetOrHead(
         const path = target[1..];
 
         if (std.mem.eql(u8, path, "health")) {
-            // TODO: https://github.com/Syndica/sig/issues/558
-            request.respond("unknown", .{
+            // HTTP GET /health always returns 200 OK with a plain-text status string.
+            // This matches agave's behavior:
+            // See: https://github.com/anza-xyz/agave/blob/master/rpc/src/rpc_service.rs#L332-L340
```
dnut: please use a permalink
```suggestion
            // See: https://github.com/anza-xyz/agave/blob/feb813a810442e9f7fc82dfcbd08118dec12efac/rpc/src/rpc_service.rs#L332-L340
```

actually this link may be more pertinent
```suggestion
            // See: https://github.com/anza-xyz/agave/blob/feb813a810442e9f7fc82dfcbd08118dec12efac/rpc/src/rpc_service.rs#L396-L397
```

dnut (changes requested):

src/rpc/methods.zig
```diff
@@ -1350,6 +1350,14 @@ pub const common = struct {
...
     slot_tracker: *sig.replay.trackers.SlotTracker,
     epoch_tracker: *sig.core.EpochTracker,
+    /// Ledger access for reading cluster's optimistic slots from gossip (for health check)
+    ledger: ?*sig.ledger.Ledger = null,
+    /// When true, always report healthy (for testing or maintenance)
+    override_health_check: ?*const std.atomic.Value(bool) = null,
+    /// Maximum allowed slot distance before node is considered unhealthy.
+    /// Default is 128 slots, matching agave's DELINQUENT_VALIDATOR_SLOT_DISTANCE.
+    /// See: https://github.com/anza-xyz/agave/blob/master/rpc-client-types/src/request.rs#L158
+    health_check_slot_distance: u64 = 128,
```
dnut:
```suggestion
    health_check_slot_distance: u64 = DELINQUENT_VALIDATOR_SLOT_DISTANCE,
```

src/rpc/server/basic.zig
```diff
@@ -134,8 +134,33 @@ fn handleGetOrHead(
...
+            // The hooks system wraps the response in Result union
+            const status_str = switch (health_result) {
+                .ok => |response| response.httpStatusString(),
+                .err => "unknown", // fallback for hook-level errors
+            };
+
+            request.respond(status_str, .{
                 .status = .ok,
                 .keep_alive = false,
             }) catch return error.SystemIoError;
```
dnut: should this match the error handling above?
```suggestion
            }) catch |err| switch (err) {
                error.WriteFailed => return,
                error.HttpExpectationFailed => return error.SystemIoError,
            };
```

src/rpc/methods.zig
hamza-syndica: Was blindly following everything from `agave` initially, I guess we don't need `override_health_check` at all. I have removed that field and related changes.

https://github.com/Syndica/sig/pull/1247/commits/a07ab1500ef7a357a4b6cff7a231f53d26510347

src/rpc/methods.zig
hamza-syndica: https://github.com/Syndica/sig/pull/1247/commits/a07ab1500ef7a357a4b6cff7a231f53d26510347

src/rpc/methods.zig
hamza-syndica: https://github.com/Syndica/sig/pull/1247/commits/a07ab1500ef7a357a4b6cff7a231f53d26510347

src/rpc/server/basic.zig
hamza-syndica: https://github.com/Syndica/sig/pull/1247/commits/a07ab1500ef7a357a4b6cff7a231f53d26510347

Also, asked a question on Slack regarding the pattern, please assist on this decision.

src/rpc/server/basic.zig
hamza-syndica: https://github.com/Syndica/sig/pull/1247/commits/641ec7ac2873763f0ebda63083b3712d96aba454

src/rpc/methods.zig
hamza-syndica: https://github.com/Syndica/sig/pull/1247/commits/641ec7ac2873763f0ebda63083b3712d96aba454

src/rpc/server/basic.zig
InKryption: Why are we ignoring `error.WriteFailed`?

dnut: Yeah that's a good question. I don't see a good reason for it. My suggestion here was mainly to address the inconsistency.

hamza-syndica: @InKryption should I simply return the `SystemIoError` for all errors case like I was doing here? Or any other specific error-set that I should create?

InKryption: Yeah I would return `error.SystemIoError` like that in general.

hamza-syndica: updated: https://github.com/Syndica/sig/pull/1247/commits/60f4060e9d43bac158ea99d4329408d3f95528f3

InKryption: that looks better to me

dnut (dismissed):

kprotty (dismissed):

dnut (approved):

PR #1248, opened 2026-02-23
feat(replay): improve feature activation handling and rent deprecation (simd-194)

yewman:
- Refactor feature state detection to return pending/activated/invalid states
- Update activated feature accounts with serialized activation slot
- Handle deprecate_rent_exemption_threshold feature activation
- Update rent sysvar when exemption threshold is deprecated

### DIFF
--- a/src/replay/epoch_transitions.zig
+++ b/src/replay/epoch_transitions.zig
@@ -191,13 +191,12 @@ pub fn computeFeatureSet(
     var new_activations = FeatureSet.ALL_DISABLED;
     var inactive_iterator = feature_set.iterator(slot, .inactive);
     while (inactive_iterator.next()) |feature| {
-        defer _ = arena.reset(.retain_capacity);
-
         const feature_id = features.map.get(feature).key;
         const feature_account = try slot_store.reader().get(
             arena.allocator(),
             feature_id,
         ) orelse continue;
+        defer feature_account.deinit(arena.allocator());
 
         switch (try features.activationStateFromAccount(feature_account)) {
             .activated => |activation_slot| if (slot >= activation_slot) feature_set.setSlot(


src/replay/epoch_transitions.zig
```diff
@@ -164,6 +168,67 @@ pub fn getEpochStakes(
...
+    feature_set: *FeatureSet,
+    logger: Logger,
+) !if (allow_new_activations) FeatureSet else void {
+    // Iterate through the inactive features and:
+    // 1. Try and load the feature account from the accounts db.
+    // 2. If the account exists, check if it has been activated already.
+    // 3. If it has been activated, add it to the active set.
+    // 4. If it has not been activated, and new activations are allowed,
+    //    add it to the active set and activate it by setting the slot
+    //    and writing it back to the accounts db.
```
kprotty: want the extra comment?

yewman: absolutely not 😂  0f21761ee

kprotty (dismissed):

src/replay/epoch_transitions.zig
```diff
@@ -164,6 +168,60 @@ pub fn getEpochStakes(
...
+                new_activations.setSlot(feature, slot);
+
+                const new_data = try allocator.alloc(u8, 9);
+                defer allocator.free(new_data);
+                const maybe_slot: ?Slot = slot;
+                _ = try sig.bincode.writeToSlice(new_data, maybe_slot, .{});
+
+                try slot_store.put(feature_id, .{
+                    .lamports = feature_account.lamports,
+                    .data = new_data,
```
dnut:
```suggestion
                var new_data: [9]u8 = undefined;
                const maybe_slot: ?Slot = slot;
                _ = try sig.bincode.writeToSlice(&new_data, maybe_slot, .{});

                try slot_store.put(feature_id, .{
                    .lamports = feature_account.lamports,
                    .data = &new_data,
```

yewman: Nice catch ef97321e

src/core/features.zig
```diff
@@ -181,11 +181,28 @@ pub const Set = struct {
...
+/// [solana-sdk] https://github.com/anza-xyz/solana-sdk/blob/54449336c03ae8a99bc37745ac97ab90a77eb24b/feature-gate-interface/src/state.rs#L37
+pub fn activationStateFromAccount(account: sig.core.Account) !FeatureActivationState {
+    if (account.data.len() < 9 or
+        !account.owner.equals(&sig.runtime.ids.FEATURE_PROGRAM_ID)) return .invalid;
+
     var feature_bytes = [_]u8{0} ** 9;
     account.data.readAll(&feature_bytes);
-    return sig.bincode.readFromSlice(failing_allocator, ?u64, &feature_bytes, .{});
+    const maybe_slot = try sig.bincode.readFromSlice(failing_allocator, ?u64, &feature_bytes, .{});
+
```
Sobeston: Might be nice to catch this and return a meaningful error, or even panic
this would be illegal behaviour, right?

src/replay/epoch_transitions.zig
```diff
@@ -166,6 +170,59 @@ pub fn getEpochStakes(
...
+    feature_set: *FeatureSet,
+    logger: Logger,
+) !if (allow_new_activations) FeatureSet else void {
+    var new_activations = FeatureSet.ALL_DISABLED;
+    var inactive_iterator = feature_set.iterator(slot, .inactive);
+    while (inactive_iterator.next()) |feature| {
+        const feature_id = features.map.get(feature).key;
+
+        const feature_account = try slot_store.reader().get(allocator, feature_id) orelse continue;
+        defer feature_account.deinit(allocator);
```
Sobeston: Might be nice to use an arena for these, resetting on each loop iter.
Probably not important, feel free to ignore

dnut (approved):

src/core/features.zig
yewman: Yeah given the assertion that the account data len is at least 9 I am not aware of any way this could fail. As such I agree a panic is a better way to go 7f806c524e

src/replay/epoch_transitions.zig
yewman: 7f806c524e

dnut (approved):

src/replay/epoch_transitions.zig
```diff
@@ -185,13 +185,19 @@ pub fn computeFeatureSet(
...
+    defer arena.deinit();
+
     var new_activations = FeatureSet.ALL_DISABLED;
     var inactive_iterator = feature_set.iterator(slot, .inactive);
     while (inactive_iterator.next()) |feature| {
-        const feature_id = features.map.get(feature).key;
+        defer _ = arena.reset(.retain_capacity);
 
-        const feature_account = try slot_store.reader().get(allocator, feature_id) orelse continue;
-        defer feature_account.deinit(allocator);
```
dnut: Accounts should be deinitialized regardless of whether you passed an arena when getting it. You shouldn't assume that the only thing that needs to happen is a free with the passed allocator. It's not how the buffer pool worked, and it's not how an optimal use of sqlite would work.

yewman: ea3b2f18c4

PR #1252, opened 2026-02-24
fix(ci): extract SHA for kcov upload in CI

yewman:
**Summary**
- [codecov-status-stuck-at-wating](https://community.codecov.com/t/codecov-status-stuck-at-waiting-for-status-to-be-reported-on-github/341/40)
- add verbose logging to codecov action
- log commits for debugging if issue persists

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,8 +1,5 @@
 version: 2.1
 
-orbs:
-  codecov: codecov/codecov@5.4.3
-
 commands:
   docker_as_user:
     description: "Configure Docker to run as the job user to prevent root-owned files in workspace"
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -266,8 +263,26 @@ jobs:
...
+            # and what GitHub watches.
+            echo "=== Codecov SHA debug ==="
+            echo "CIRCLE_SHA1:  $CIRCLE_SHA1"
+            echo "CIRCLE_BRANCH: ${CIRCLE_BRANCH:-unset}"
+            echo "HEAD:         $(git rev-parse HEAD)"
+            echo "HEAD^1:       $(git rev-parse HEAD^1 2>/dev/null || echo 'N/A')"
+            echo "HEAD^2:       $(git rev-parse HEAD^2 2>/dev/null || echo 'N/A (not a merge commit)')"
+            echo "========================="
+
+            curl -fsSLo codecov https://cli.codecov.io/latest/linux/codecov
```
dnut: pinning the version will protect against breaking changes

```suggestion
            curl -fsSLo codecov https://cli.codecov.io/v11.2.6/linux/codecov
```

ajw221 (approved):

.circleci/config.yml
```diff
@@ -266,8 +263,26 @@ jobs:
...
+            chmod +x codecov
+            ./codecov upload-coverage \
+              --dir kcov-merged/kcov-merged \
+              --commit-sha "$CIRCLE_SHA1"
```
dnut: This is the same behavior that the circleci orb has by default. The `_get_commit_sha` function reads the `CIRCLE_SHA1` variable.

https://github.com/getsentry/prevent-cli/blob/48c818e0955a4ce7509db98023010dea981f76fd/codecov-cli/codecov_cli/helpers/ci_adapters/circleci.py#L11-L12

yewman: Yeah, I would like to use the verbose flag and log the hashes to see if there is any issue there. Some basic research led me to believe if could be the source of an issue, so I would like to rule it out if possible

PR #1253, opened 2026-02-24
feat(v2): gossip

kprotty:
Decently compliant gossip implementation

### DIFF
--- a/v2/src/common/gossip.zig
+++ b/v2/src/common/gossip.zig
@@ -87,7 +87,8 @@ pub const ClusterInfo = extern struct {
 
                 // set timeout of 1s for connect, read, write.
                 const tv = comptime std.mem.asBytes(&std.posix.timeval{ .sec = 1, .usec = 0 });
-                std.posix.setsockopt(socket, std.posix.SOL.SOCKET, std.posix.SO.RCVTIMEO, tv) catch continue;
+                std.posix.setsockopt(socket, std.posix.SOL.SOCKET, std.posix.SO.RCVTIMEO, tv) catch
+                    continue;
                 std.posix.connect(socket, &entry_addr.any, entry_addr.getOsSockLen()) catch {
                     continue;
                 };
--- a/v2/src/services/gossip.zig
+++ b/v2/src/services/gossip.zig
@@ -811,17 +811,22 @@ const Gossip = struct {
         value: GossipValue,
     ) !?struct { Key, u8 } {
         // Extract key information from the data.
+        var deprecated = false;
         const from: Pubkey, const wallclock: u64, const index: u16 = switch (value.data) {
             inline .vote, .lowest_slot, .epoch_slots, .duplicate_shred => |v| blk: {
                 break :blk .{ v.from, v.wallclock, v.index };
             },
-            .snapshot_hashes => |v| blk: {
-                break :blk .{ v.from, v.wallclock, 0 };
-            },
             .contact_info => |ci| blk: {
                 break :blk .{ ci.from, ci.wallclock.value, 0 };
             },
-            else => unreachable,
+            inline .snapshot_hashes, .restart_heaviest_fork, .restart_last_voted_fork => |v| blk: {
+                deprecated = value.data != .snapshot_hashes;
+                break :blk .{ v.from, v.wallclock, 0 };
+            },
+            inline else => |v| {
+                comptime std.debug.assert(@TypeOf(v) == bincode.Deprecated);
+                unreachable;
+            },
         };
 
         // Serialize the value & validate its signature.
@@ -836,13 +841,17 @@ const Gossip = struct {
 
         // Check wallclock in general
         const key: Key = .{ .from = from, .tag = std.meta.activeTag(value.data), .index = index };
-        const update_contact = switch (caller) {
-            .us => true,
+        const update_contact = (switch (caller) {
+            .us => blk: {
+                assert(!deprecated); // we should not be inserting our own deprecated data
+                break :blk true; // update our own ContactInfo's last_updated
+            },
             .push => blk: {
-                if (wallclock <= (now -| STALE_PUSH_THRESHOLD_MS)) return null;
-                if (wallclock >= (now +| STALE_PUSH_THRESHOLD_MS)) return null;
-                if (from.equals(&self.identity())) return null;
-                break :blk false;
+                if (wallclock <= (now -| STALE_PUSH_THRESHOLD_MS)) return null; // out of range
+                if (wallclock >= (now +| STALE_PUSH_THRESHOLD_MS)) return null; // out of range
+                if (from.equals(&self.identity())) return null; // push sent our own thing to us
+                if (deprecated) break :blk null; // discover, but dont insert
+                break :blk false; // push msgs dont update ContactInfo's last_updated
             },
             .pull => blk: {
                 // TODO: currently assumes all nodes are staked.
@@ -855,15 +864,19 @@ const Gossip = struct {
                     threshold = 432_000 * 400; // slots_in_epoch * slot_ms
                 }
 
-                if (now <= wallclock +| threshold) break :blk true;
-                if (value.data == .contact_info) break :blk false;
+                if (!deprecated) {
+                    if (now <= wallclock +| threshold) break :blk true; // within threshold
+                    if (value.data == .contact_info) break :blk false; // Contact outside threshold
+                }
 
-                // Value is too old while not being a ContactInfo.
-                // Record that we've seen it, but dont insert it.
-                try self.onDiscoveredValue(now, key, value);
-                self.addExpired(now, hash);
-                return null;
+                // deprecated, or old non-Contact outside threshold
+                break :blk null;
             },
+        }) orelse {
+            // Record that we've seen it, but don't insert it.
+            try self.onDiscoveredValue(now, key, value);
+            self.addExpired(now, hash);
+            return null;
         };
 
         const exists, const v = blk: {
@@ -974,7 +987,11 @@ const Gossip = struct {
                     // TODO: if map.get(.tpu_vote): send to consensus service
                     // TODO: if map.get(.rpc): send to snapshot service
                 },
-                else => unreachable,
+                .restart_heaviest_fork, .restart_last_voted_fork => {}, // deprecated
+                inline else => |v| {
+                    comptime std.debug.assert(@TypeOf(v) == bincode.Deprecated);
+                    unreachable;
+                },
             }
         }
     }
@@ -1415,8 +1432,25 @@ const GossipData = union(enum(u32)) {
             bytes: ShortVec(u8),
         }),
     },
-    restart_last_voted_fork: bincode.Deprecated,
-    restart_heaviest_fork: bincode.Deprecated,
+    /// Deprecated and unused. But nodes may still send them for us to addExpired.
+    restart_last_voted_fork: struct {
+        from: Pubkey,
+        wallclock: u64,
+        offsets: Vec(union(enum(u32)) {
+            rle: Vec(VarInt(u16)),
+            raw: BitVec(u8),
+        }),
+        last_voted: SlotAndHash,
+        shred_version: u16,
+    },
+    /// Deprecated and unused. But nodes may still send them for us to addExpired.
+    restart_heaviest_fork: struct {
+        from: Pubkey,
+        wallclock: u64,
+        last_slot: SlotAndHash,
+        observed_stake: u64,
+        shred_version: u16,
+    },
 
     const SocketKey = enum(u8) {
         gossip,


v2/gossip_prototype.zig
```diff
@@ -0,0 +1,1545 @@
...
+                };
+
+                std.log.debug("PullRequest({}, {?}) bits={}", .{
+                    from, getGossipAddr(pr.contact_info.data) catch null,
+                    pr.mask_bits,
+                });
+
+                // Try to insert the contact_info
+                verifyValue(pr.contact_info, now) catch continue;
+                try onNewValue(
```
Rexicon226: If any of these trys fails, the whole loop exits. An invalid packet would bring down gossip.
Recommend you place handling into inner function and catch switch, raising OOM but continue for other.

v2/gossip_prototype.zig
```diff
@@ -0,0 +1,1545 @@
...
+
+const Bloom = struct {
+    const max_bloom_bits = 928 * 8;
+    const false_rate = 0.1;
+
+    fn add(keys: []align(1) u64, words: []align(1) u64, bytes: []const u8) void {
+        for (keys) |k| {
+            var h = std.hash.Fnv1a_64{ .value = k };
+            h.update(bytes);
+            const bit = h.final() % (words.len * 8);
```
Rexicon226: You insert attacker-controlled words into the bloom, but they could be len==0 causing this to crash.

v2/gossip_prototype.zig
kprotty: The gossip prototype was a first draft. Will delete.

The real one is in `v2/src/services/gossip.zig`

Rexicon226: Sure, just don't make same mistake :P

v2/src/common/gossip.zig
```diff
@@ -0,0 +1,133 @@
+const std = @import("std");
+
+test {
+    _ = std.testing.refAllDecls(@This());
+}
```
InKryption: Will just point out, this doesn't actually do anything in this file except reference `KeyPair`, `Config`, and `ClusterInfo`; unless this is going to export other files, could just remove this.

v2/src/common/gossip.zig
```diff
@@ -0,0 +1,133 @@
...
+                defer std.posix.close(socket);
+
+                // set timeout of 1s for connect, read, write.
+                const tv = comptime std.mem.asBytes(&std.posix.timeval{ .sec = 1, .usec = 0 });
+                try std.posix.setsockopt(socket, std.posix.SOL.SOCKET, std.posix.SO.RCVTIMEO, tv);
+                std.posix.connect(socket, &entry_addr.any, entry_addr.getOsSockLen()) catch {
+                    continue;
+                };
+
+                var stream_writer = (std.net.Stream{ .handle = socket }).writer(&io_buf);
```
InKryption:
```suggestion
                var stream_writer = std.net.Stream.writer(.{ .handle = socket }, &io_buf);
```

v2/src/common/gossip.zig
```diff
@@ -0,0 +1,133 @@
...
+                const writer = &stream_writer.interface;
+                _ = try writer.splatByte(0, 4 + (4 * 2) + (4 * 2)); // hdr + tcp ports + udp ports
```
InKryption:
```suggestion
                try writer.splatByteAll(0, 4 + (4 * 2) + (4 * 2)); // hdr + tcp ports + udp ports
```
or something like this, if your concern is with avoiding the loop in `*All`:
```suggestion
                const hdr_tcp_udp_ports_len = 4 + (4 * 2) + (4 * 2);
                std.debug.assert(hdr_tcp_udp_ports_len == try writer.splatByte(
                    0,
                    hdr_tcp_udp_ports_len,
                )); // hdr + tcp ports + udp ports
```

v2/src/common/gossip.zig
```diff
@@ -0,0 +1,133 @@
...
+
+        pub fn toNetAddress(self: *const Address) std.net.Address {
+            if (self.is_v6) return .initIp6(self.ip, self.port, 0, 0);
+            return .initIp4(self.ip[0..4].*, self.port);
+        }
+    };
+
+    pub fn getFromEcho(gossip_port: u16, cluster: common.solana.ClusterType) !ClusterInfo {
+        var io_buf: [4096]u8 = undefined;
+        var addr_buf: [4096]u8 = undefined;
```
InKryption: Any reason to have pre-declare here instead of next to where they're actually used?

v2/src/common/gossip.zig
```diff
@@ -0,0 +1,133 @@
...
+
+                var stream_writer = (std.net.Stream{ .handle = socket }).writer(&io_buf);
+                const writer = &stream_writer.interface;
+                _ = try writer.splatByte(0, 4 + (4 * 2) + (4 * 2)); // hdr + tcp ports + udp ports
+                try writer.writeByte('\n'); // trailer
+                writer.flush() catch continue;
+
+                var stream_reader = (std.net.Stream{ .handle = socket }).reader(&io_buf);
+                const reader: *std.Io.Reader = stream_reader.interface();
+                _ = reader.takeInt(u32, .little) catch continue;
```
InKryption: Any reason not to use `discard` here?
```suggestion
                reader.discardAll(@sizeOf(u32)) catch continue;
```

v2/src/services/gossip.zig
```diff
@@ -0,0 +1,1477 @@
...
+    };
+
+    pub const Config = struct {
+        netpair: *Pair,
+        keypair: *const common.gossip.KeyPair,
+        entry_addr: std.net.Address,
+        contact_info: GossipData,
+    };
+
+    pub fn init(fba: *std.heap.FixedBufferAllocator, config: Config) !Gossip {
```
InKryption, line 247: Why not just accept a generic allocator? If the reason is to avoid having to exhaustively clean up memory, could just have `arena: std.mem.Allocator`.

v2/src/services/gossip.zig
```diff
@@ -0,0 +1,1477 @@
...
+    _ = start;
+}
+
+pub const name = .gossip;
+pub const panic = start.panic;
+pub const std_options = start.options;
+
+pub const ReadWrite = struct {
+    pair: *Pair,
+    scratch_mem: *[common.gossip.scratch_memory_size]u8,
```
InKryption: Are we supposed to be avoiding allocation in services entirely by relying on enough shared memory being supplied as space?
If not, any reason to not just allocate normally? Or even to just use a global variable, so that it doesn't have to be in shared memory.

v2/src/services/gossip.zig
```diff
@@ -0,0 +1,1477 @@
...
+                const from, const shred_version = switch (pr.contact_info.data) {
+                    inline .legacy_contact_info, .contact_info => |v| .{ v.from, v.shred_version },
+                    else => return error.InvalidPullRequestContactInfo,
+                };
+
+                std.log.debug("Received PullRequest(from:{f})", .{from});
+
+                // Unverified peers must respond to a ping first.
+                const peer = (try self.getOrTrackPeer(now, shred_version, addr, from)) orelse
+                    return error.PullRequestFromUnverifiedPeer;
```
InKryption: Just fyi the parentheses here are unnecessary, `try` binds stronger than `orelse`:
```suggestion
                const peer = try self.getOrTrackPeer(now, shred_version, addr, from) orelse
                    return error.PullRequestFromUnverifiedPeer;
```

v2/src/common/gossip.zig
dnut: Even if there are tests to reference, I don't like using test blocks for this because it pollutes the test output with meaningless tests that don't do anything. Why not use comptime blocks instead? refAllDecls is a noop if it's not a test build.

v2/src/services/gossip.zig
kprotty: `fba` serves same purpose as `arena`, but more accurate to memory needed as it avoids the internal details of arena allocating large amortized chunks from its parent allocator (which can overflow the `scratch_mem` region size)

v2/src/services/gossip.zig
kprotty: a global makes a lot more sense. Will remove scratch_mem region.

v2/src/common/gossip.zig
kprotty: small trick to avoid `memset(stack, 0xaa)` inside the loops. Ultimately doesnt matter though.

v2/src/common/gossip.zig
kprotty: https://github.com/Syndica/sig/pull/1253/commits/216ecaa805d1058d6c971bf6f187ce012901e9bf

v2/src/common/gossip.zig
kprotty: https://github.com/Syndica/sig/pull/1253/commits/216ecaa805d1058d6c971bf6f187ce012901e9bf

v2/src/common/gossip.zig
kprotty: https://github.com/Syndica/sig/pull/1253/commits/216ecaa805d1058d6c971bf6f187ce012901e9bf

v2/src/common/gossip.zig
kprotty: https://github.com/Syndica/sig/pull/1253/commits/216ecaa805d1058d6c971bf6f187ce012901e9bf

v2/src/services/gossip.zig
kprotty: https://github.com/Syndica/sig/pull/1253/commits/36bfc0cbeaa4dfd9a5e6aa99339ba03f4365b7a6

v2/src/services/gossip.zig
kprotty: https://github.com/Syndica/sig/pull/1253/commits/36bfc0cbeaa4dfd9a5e6aa99339ba03f4365b7a6

v2/src/common/gossip.zig
InKryption: Interesting. Yeah that actually makes sense - I think it's worth adding a comment on here about that, that it's to improve performance in safe modes by avoiding that.

v2/src/services/gossip.zig
InKryption: I'm not proposing using an actual arena allocator instead, just suggesting the name as a description of the intention. I would also accept `fba: std.mem.Allocator` if you think it's a very important distinction, mostly I just don't see a reason to force this code to use the fba implementation.

kprotty: Given v2 will have no heap alloc at runtime, every arena-like pattern would be a bump allocator on fixed memory, hence fba. Using the concrete type over Allocator is a small type-safety thing to avoid passing in an incompatible allocator. Should I change the arg name from `fba` to `arena`?

dnut: Is there a specific reason why this function *needs* the parameter to be an fba or arena? If it could work with any allocator, then it should just be called `allocator`.

InKryption: It can't work with any allocator, because the code doesn't clean up memory in the error path, and the return value assumes it doesn't have to clean up memory, so it doesn't expose anything like a `deinit` method.

v2/src/common/gossip.zig
kprotty: https://github.com/Syndica/sig/pull/1253/commits/dc65f393a4fb29bc19f20cb9f7aac1633fdc1012

v2/gossip_prototype.zig
kprotty: https://github.com/Syndica/sig/pull/1253/commits/36bfc0cbeaa4dfd9a5e6aa99339ba03f4365b7a6

v2/src/services/gossip.zig
```diff
@@ -0,0 +1,1489 @@
...
+        keys: [MAX_BLOOM_KEYS]u64,
+        words: [MAX_BLOOM_BYTES / 8]u64,
+
+        fn init(prng: std.Random) @This() {
+            var keys: [MAX_BLOOM_KEYS]u64 = undefined;
+            for (&keys) |*key| key.* = prng.int(u64);
+            return .{ .keys = keys, .words = @splat(0) };
+        }
+
+        fn asBloomFilter(self: *@This(), num_keys: ?usize, num_bits: ?usize) BloomFilter {
```
InKryption: I think we should stick to the usual convention of using the actual name of the type instead of `@This()` where feasible.

v2/src/services/gossip.zig
```diff
@@ -0,0 +1,1497 @@
...
+                            peer.addr = addr;
+                            peer.shred_version = ci.shred_version;
+                        }
+                    }
+
+                    // TODO: if map.get(.serve_repair): send to repair service
+                    // TODO: if map.get(.tpu_vote): send to consensus service
+                    // TODO: if map.get(.rpc): send to snapshot service
+                },
+                .restart_last_voted_fork => {}, // TODO: implement wen-restart protocol (SIMD-0046).
```
Rexicon226: Will never be implemented btw, Anza has removed wen-restart.

kprotty: very based

v2/src/common/solana/cluster.zig
```diff
@@ -0,0 +1,46 @@
+/// Analogous to [ClusterType](https://github.com/anza-xyz/solana-sdk/blob/a467058aabc453c7d749a4993c56df293d1d75c3/cluster-type/src/lib.rs#L19)
+/// Explicit numbers are added to ensure we don't mess up the order of the fields and break bincode reading.
+pub const ClusterType = enum(u8) {
+    testnet = 0,
+    mainnet = 1,
+    devnet = 2,
+    development = 3,
```
dnut: Do we need this variant? How would this actually be used? The other three options all clearly identify a specific cluster and provide some useful data about it. `development` is just a useless variant that tells us nothing except that we don't know what cluster it is. Maybe in contexts where .development might be used, ClusterType should just be optional. And in that case, if all ClusterType is used for is identifying actual clusters, maybe we should call it Cluster.

v2/src/common/gossip.zig
```diff
@@ -0,0 +1,128 @@
...
+
+pub const Config = extern struct {
+    keypair: KeyPair,
+    cluster_info: ClusterInfo,
+    turbine_recv_port: u16,
+};
+
+pub const ClusterInfo = extern struct {
+    public_ip: Address,
+    entry_addr: Address,
```
dnut: Why only one? What if it's unreliable? I think we should use use all the entrypoints.

v2/src/services/gossip.zig
```diff
@@ -0,0 +1,1497 @@
...
+        last_voted: SlotAndHash,
+        shred_version: u16,
+    },
+    restart_heaviest_fork: struct {
+        from: Pubkey,
+        wallclock: u64,
+        last_slot: SlotAndHash,
+        observed_stake: u64,
+        shred_version: u16,
+    },
```
dnut: technically these are still allowed but they are deprecated and will probably be removed at some point. we should keep them for now but could leave a comment that it's deprecated
```suggestion
    /// deprecated and unused
    restart_last_voted_fork: struct {
        from: Pubkey,
        wallclock: u64,
        offsets: Vec(union(enum(u32)) {
            rle: Vec(VarInt(u16)),
            raw: BitVec(u8),
        }),
        last_voted: SlotAndHash,
        shred_version: u16,
    },
    /// deprecated and unused
    restart_heaviest_fork: struct {
        from: Pubkey,
        wallclock: u64,
        last_slot: SlotAndHash,
        observed_stake: u64,
        shred_version: u16,
    },
```

v2/src/services/gossip.zig
```diff
@@ -0,0 +1,1497 @@
...
+        serve_repair_quic: SocketAddr,
+        tpu: SocketAddr,
+        tpu_forwards: SocketAddr,
+        tpu_vote: SocketAddr,
+        rpc: SocketAddr,
+        rpc_pubsub: SocketAddr,
+        serve_repair: SocketAddr,
+        wallclock: u64,
+        shred_version: u16,
+    },
```
dnut: This is no longer permitted to be replicated in the cluster. Agave discards these before deserializing them, and never inserts them into the table. So we don't need to support it.

You could remove the union field entirely and manually specify tags for the others, but that feels awkward for a few reasons. I would specify the type as `noreturn` for all the rejected gossip data variants and leave a doc comment explaining that this variant is no longer permitted in the gossip table. What do you think?
```suggestion
    /// no longer permitted
    legacy_contact_info: noreturn,
```

The same applies to:
- legacy_snapshot_hashes
- account_hashes
- legacy_version
- version
- node_instance

v2/src/services/gossip.zig
```diff
@@ -0,0 +1,1497 @@
...
+        var msg_reader: std.Io.Reader = .fixed(packet.data[0..packet.size]);
+        const msg = bincode.read(&msg_fba, &msg_reader, GossipMessage) catch |e| {
+            std.log.err("invalid msg from ({f}, size={}): {}", .{ packet.addr, packet.size, e });
+            return;
+        };
+
+        self.processMessage(now, packet.addr, msg) catch |e| {
+            std.log.err("failed to process msg ({f}, {s}) {}", .{ packet.addr, @tagName(msg), e });
+            return;
+        };
```
dnut: This only reads one message per iteration? Do we really need to send out pushes and pulls for every individual message that we receive? I think we should drain the recv queue before sending any pull requestts.

v2/src/services/gossip.zig
```diff
@@ -0,0 +1,1497 @@
...
+            i += 1;
+            if (peer.last_ping.wallclock <= now -| ACTIVE_PONG_THRESHOLD_MS) {
+                try self.sendPing(peer.addr);
+                peer.last_ping.wallclock = now;
+                peer.last_ping.count_since_pong += 1;
+            }
+        }
+    }
+
+    fn processPushMessages(self: *Gossip, now: u64) !void {
```
dnut: This name confused me. I thought it was for processing existing push messages that were received from other nodes. But it's actually for producing and sending them. I see you already have a method called `sendPushMessages`. It's hard to rename this function because its role is so specific. I think a more accurate name would be `updateActiveSetAndSendPushMessages` or  `runSendPushMessagesStep`, but the verbosity is a mild code smell.

What about this idea. You could remove these lines from this function and instead call them right after calling this function:
```zig
        // Add a new instance of our contact info.
        try self.insertOurOwnData(now, self.config.contact_info);

        // Send out push messages
        try self.sendPushMessages();
```
Then you could rename this function `refreshPushActiveSet`.

v2/src/services/gossip.zig
```diff
@@ -0,0 +1,1497 @@
...
+                    return;
+                };
+
+                peer.last_pong = now;
+                peer.last_ping.count_since_pong = 0;
+            },
+        }
+    }
+
+    fn processPings(self: *Gossip, now: u64) !void {
```
dnut: same thought as processPushMessages. Can we call this sendPings?

v2/src/services/gossip.zig
```diff
@@ -0,0 +1,1497 @@
...
+
+                // TODO: enable this? matching shred_version
+                // if (peer.shred_version != self.config.contact_info.contact_info.shred_version)
+                //     continue;
+
+                // TODO: enable this? active-enough ContactInfo
+                // const ci_key: Key = .{ .from = from, .tag = .contact_info, .index = 0 };
+                // const ci = self.table.getPtr(ci_key) orelse continue;
+                // if (ci.last_updated <= now -| ACTIVE_VALUE_THRESHOLD_MS)
+                //     continue;
```
dnut: Why is this commented out? Seems like you could just uncomment it and it would work. I assume this is logic ported over from agave. Are you thinking its actually not a good idea to do this?

v2/src/services/gossip.zig
```diff
@@ -0,0 +1,1497 @@
...
+        }
+
+        if (self.ping_timeout <= now) {
+            self.ping_timeout = now + PING_INTERVAL_MS;
+            try self.processPings(now);
+        }
+
+        var slice = self.config.netpair.recv.getReadable() catch return;
+        defer slice.markUsed(1);
+        const packet: *const Packet = slice.get(0);
```
dnut, line 300: This only reads one message per iteration? Do we really need to send out pushes and pulls for every individual message that we receive? I think we should drain the recv queue before sending pull requestts.

v2/src/services/gossip.zig
```diff
@@ -0,0 +1,1497 @@
...
+
+        var writer: std.Io.Writer = .fixed(&packet.data);
+        try bincode.write(&writer, msg);
+        packet.size = @intCast(writer.buffered().len);
+
+        slice.markUsed(1);
+    }
+};
+
+const bincode = struct {
```
dnut, line 1087: I love how simple this is.

v2/src/services/gossip.zig
kprotty: It does a bit more than sending pings (updating ping token window, removes dead peers)

v2/src/services/gossip.zig
kprotty: PushMessages and PullRequests are only sent out when their timeouts fire, rather than when a message is received.

v2/src/services/gossip.zig
kprotty: Its a bit tricky; The current bootstrap flow is `PullReq -> entrypoint` which responds with `-> Ping (tracking entrypoint) -> Pong`, then now a good `PullReq -> entrypoint -> PullResp`. There's no guarantee the PullResp will contain the entrypoint's ContactInfo to bring it into our active_set under those checks, which is what lets us send a PushMessage with our ContactInfo and get us into the cluster. 

In practice, the initial PullResp from the entrypoint only sends back older values (epoch_slots, duplicate_shreds) until we send our first PushMessage. Im trying some stuff out to keep these checks but special-case that first PushMessage elsewhere.

v2/src/common/gossip.zig
kprotty: Each one could have a different shred_version returned by EchoResponse. Unclear which one should be used for our ContactInfo. Started trying to impl it, but not sure if its worth it.

v2/src/services/gossip.zig
kprotty: https://github.com/Syndica/sig/pull/1253/commits/049c6b090fa1540764cd808ca6144f502a4fe391

v2/src/services/gossip.zig
kprotty: Opted to bring back the changes + send to entrypoints anyways if active_set is empty on the next push-contact-info timer expiry

https://github.com/Syndica/sig/pull/1253/commits/f4adf482015b337cc92046d07e2ea4e63fc69a49

v2/src/common/solana/cluster.zig
kprotty: https://github.com/Syndica/sig/pull/1253/commits/765090f2890632768298fb2ca1fb44af668bac3f

v2/src/services/gossip.zig
kprotty: https://github.com/Syndica/sig/commit/049c6b090fa1540764cd808ca6144f502a4fe391

v2/src/services/gossip.zig
kprotty: https://github.com/Syndica/sig/pull/1253/commits/af43e8d651bb7e819c92d98da9af8981345d921c

v2/src/common/gossip.zig
kprotty: https://github.com/Syndica/sig/pull/1253/commits/af43e8d651bb7e819c92d98da9af8981345d921c

v2/src/services/gossip.zig
kprotty: Changed it to `updatePeers`: https://github.com/Syndica/sig/commit/af43e8d651bb7e819c92d98da9af8981345d921c

v2/src/services/gossip.zig
dnut: These types are actually still allowed and are replicated by agave's gossip. These are different than the other ones (like legacy contact info) which are strictly blocked now by agave. I don't think these are actually produced any more by agave, so maybe it's fine to deprecate them like this. But if anyone in the cluster is sending them out, then our gossip table will diverge from agave's, which can cause some noise as we send out pull requests and other nodes in the cluster think they need to send these to us values.

kprotty: https://github.com/Syndica/sig/pull/1253/commits/cdaffb8a8fce8c2ed8ab48ef69888a010f10dc0c

PR #1254, opened 2026-02-24
fix(replay, rpc): make SlotTracker & EpochTracker thread safe 

prestonsn:
Adds `RwMux` and reference counting to `SlotTracker` and `EpochTracker` to allow for safe concurrent access from the RPC thread. This fixes data races currently present due to the addition of the RPC server thread now accessing data. One should note that the data races are very rare. SlotTracker prunes unrooted slots (which the RPC thread never has an interest in), and the EpochTracker only deinits on epoch boundaries (likelihood of this causing a data race with an RPC request is hence very low).   

Reference counting was necessary for correctness with `SlotTracker` since methods on the tracker would hand out references into `Element`(s) by visiting the top-level `slots` hashmap. There is a code-path where replay calls `pruneNonRooted()` while the RPC thread could be accessing that pruned `Element`, invalidating its pointers. This should be quite rare (only pruning non-rooted combined with RPC not being interested in these slots, but it's still a potential foot gun).  

This PR adds an `RwMux` around that map access, and adds reference counting to the `Reference` wrapper type that is handed out. Consumers call `.release()` when they're done with their borrow. This does mean that `Element` now holds onto the `Allocator` so that it can be destroyed on either the replay or RPC thread safely. 

For `EpochTracker`,, the same pattern applies to `rooted_epochs`. The replay thread writes new `EpochInfo` entries when epochs are rooted (once every ~2 days on mainnet), while the RPC thread reads from it via `getVoteAccounts` via `getEpochInfo()`. There is a race where replay's `insert()` overwrites a buffer slot (calling `deinit` + `destroy` on the old `EpochInfo`) while the RPC thread holds a pointer into that `EpochInfo`'s stakes data. 

This PR wraps `rooted_epochs` in an `RwMux` and adds a `ReferenceCounter` plus a stored Allocator to `EpochInfo`. `get()` increments the RC under the read lock, callers use the pointer freely without holding any lock, and call `release()` when done-if the RC hits zero, the `EpochInfo` deinits and frees itself. `insert()` overwriting an old entry calls release() instead of directly destroying it, so the entry stays alive if another thread still holds a reference. `getLeaderSchedules()` returns a new `LeaderSchedulesWithEpochInfos` wrapper that bundles the schedule with its RC'd `EpochInfo` references, requiring callers to call `release()` when done.

### Minor potential leak site fix 

In `SlotTracker.put()`, there's a potential leak due to not properly de allocating clobbered `Element`s if `put()` is called with a slot that already has an `Element` in the `slots` map. This isn't possible given the access pattern, but is still a potential leak site if things change in the future. Fixed in this PR.

### Testing

Ran sig on testnet a few times, re-checked available RPCs. Did not see sig fall behind the network. slot replay timings seem around the same as before from the logs (likely since most of these accesses through the `RwMux` are uncontended 99.99% of the time).

### DIFF
--- a/src/core/epoch_tracker.zig
+++ b/src/core/epoch_tracker.zig
@@ -503,7 +503,7 @@ pub const EpochInfo = struct {
 /// deinitialize the corresponding data when overwriting entries with new epoch data.
 const RootedEpochBuffer = struct {
     buf: [4]?*const EpochInfo = @splat(null),
-    root: Atomic(Epoch) = .init(0),
+    root: Epoch = 0,
 
     fn deinit(self: *RootedEpochBuffer) void {
         for (&self.buf) |*maybe_entry| {
@@ -527,7 +527,7 @@ const RootedEpochBuffer = struct {
         }
 
         self.buf[index] = value;
-        self.root.store(epoch, .monotonic);
+        self.root = epoch;
     }
 
     /// Returns a ref-counted reference to the EpochInfo for the given epoch.
@@ -536,7 +536,7 @@ const RootedEpochBuffer = struct {
         self: *const RootedEpochBuffer,
         epoch: Epoch,
     ) !*const EpochInfo {
-        const root_epoch = self.root.load(.monotonic);
+        const root_epoch = self.root;
 
         if (root_epoch == 0 and std.mem.allEqual(
             ?*const EpochInfo,
@@ -558,13 +558,12 @@ const RootedEpochBuffer = struct {
     }
 
     pub fn isNext(self: *const RootedEpochBuffer, epoch: Epoch) bool {
-        const root = self.root.load(.monotonic);
-        if (root == 0 and (epoch == 1 or std.mem.allEqual(
+        if (self.root == 0 and (epoch == 1 or std.mem.allEqual(
             ?*const EpochInfo,
             &self.buf,
             null,
         ))) return true;
-        return epoch == root + 1;
+        return epoch == self.root + 1;
     }
 };
 
--- a/src/shred_network/duplicate_shred_listener.zig
+++ b/src/shred_network/duplicate_shred_listener.zig
@@ -1007,7 +1007,7 @@ test "GossipDuplicateShredHandler: reconstructShredsFromData returns shreds on v
     {
         var lock = epoch_tracker.rooted_epochs.write();
         defer lock.unlock();
-        lock.mut().root.store(base_epoch, .monotonic);
+        lock.mut().root = base_epoch;
     }
 
     var ledger = try sig.ledger.tests.initTestLedger(gpa, @src(), .FOR_TESTS);


src/replay/trackers.zig
```diff
@@ -82,22 +86,32 @@ pub const SlotTracker = struct {
...
+            return &self.element.constants;
+        }
+
+        pub fn state(self: Reference) *SlotState {
+            return &self.element.state;
+        }
+
+        pub fn release(self: Reference) void {
+            self.element.destroy();
+        }
```
prestonsn, line 109: little confusing that we call destroy always?

src/replay/trackers.zig
```diff
@@ -134,16 +154,22 @@ pub const SlotTracker = struct {
...
+        const slots = slots_lg.mut();
+
+        defer tracy.plot(u32, "slots tracked", @intCast(slots.count()));
 
-        try self.slots.ensureUnusedCapacity(allocator, 1);
+        try slots.ensureUnusedCapacity(allocator, 1);
         const elem = try allocator.create(Element);
         elem.* = slot_init;
-        self.slots.putAssumeCapacity(slot, elem);
+        slots.putAssumeCapacity(slot, elem);
```
prestonsn: potential leak here if new replaces an existing Element

src/replay/trackers.zig
prestonsn: https://github.com/Syndica/sig/pull/1254/commits/e045b515131a79e260898c2a5372755b9d3599d5

src/replay/trackers.zig
prestonsn: https://github.com/Syndica/sig/pull/1254/commits/c85583916628fb0fa241d1f7c7b6a4dc4c43fd55

src/consensus/progress_map.zig
```diff
@@ -1249,16 +1249,19 @@ test "ForkProgress.init" {
     const slot = bank_data.slot;
     const slot_consts: sig.core.SlotConstants =
         try .fromBankFields(allocator, &bank_data, .ALL_DISABLED);
-    defer slot_consts.deinit(allocator);
 
     var slot_state: sig.core.SlotState =
         try .fromBankFieldsForTest(allocator, &bank_data, null);
-    defer slot_state.deinit(allocator);
```
kprotty: Should have `errdefer slot_consts.deinit` in case `slot_state = try` fails. But then once `slot_info` is created, the errdefer would be invalid due to its `defer slot_info.deinit`. Could maybe built it in a block:
```zig
const slot_info = blk: {
  const slot_consts = try ...
  errdefer slot_consts.deinit()

  const slot_state = try ...
  errdefer slot_state.deinit() // not needed, but symmetrical ig

  break :blk Element{ slot_state, slot_consts }
};
```

src/replay/trackers.zig
```diff
@@ -59,16 +62,20 @@ pub const SlotTracker = struct {
...
+        allocator: Allocator,
+        rc: ReferenceCounter = .init,
 
-        fn toRef(self: *Element) Reference {
-            return .{
-                .constants = &self.constants,
-                .state = &self.state,
-            };
+        pub fn toRef(self: *Element) ?Reference {
+            if (!self.rc.acquire()) return null;
```
kprotty: when can this fail?

src/replay/trackers.zig
```diff
@@ -114,18 +131,24 @@ pub const SlotTracker = struct {
...
         self.wg.wait();
         allocator.destroy(self.wg);
 
-        var slots = self.slots;
-        for (slots.values()) |element| element.destroy(allocator);
+        var slots_lg = self.slots.write();
+        const slots = slots_lg.mut();
+        for (slots.values()) |element| element.destroy();
         slots.deinit(allocator);
+        slots_lg.unlock();
```
kprotty: nit: can `defer unlock()`

src/replay/trackers.zig
```diff
@@ -134,16 +157,24 @@ pub const SlotTracker = struct {
...
+
+        defer tracy.plot(u32, "slots tracked", @intCast(slots.count()));
 
-        try self.slots.ensureUnusedCapacity(allocator, 1);
+        try slots.ensureUnusedCapacity(allocator, 1);
         const elem = try allocator.create(Element);
         elem.* = slot_init;
-        self.slots.putAssumeCapacity(slot, elem);
+        const gop = slots.getOrPutAssumeCapacity(slot);
+        if (gop.found_existing) gop.value_ptr.*.releaseRef();
```
kprotty, line 170: real

src/consensus/progress_map.zig
prestonsn: I see, on error path`slot_consts` hasn't been moved into `slot_element` yet. https://github.com/Syndica/sig/pull/1254/commits/546c2d1796436da8ea541536349a3bbdf28fed4b

src/replay/trackers.zig
prestonsn: ye good catch, can't fail for element that still in the `slots` map. Calling `toRef` implies either the read or write lock is held. made this an `assert` instead. https://github.com/Syndica/sig/pull/1254/commits/8f619a29542c78aa1ce22dade795a8e35b8921c9

src/replay/trackers.zig
prestonsn: :saluting_face: https://github.com/Syndica/sig/pull/1254/commits/deab244731f754432416e9bc6a54a3b050a96d5f

src/replay/trackers.zig
prestonsn: gop 😫

kprotty (approved):

src/core/epoch_tracker.zig
```diff
@@ -423,36 +505,35 @@ const RootedEpochBuffer = struct {
     buf: [4]?*const EpochInfo = @splat(null),
     root: Atomic(Epoch) = .init(0),
 
-    fn deinit(self: *const RootedEpochBuffer, allocator: Allocator) void {
+    fn deinit(self: *const RootedEpochBuffer, _: Allocator) void {
```
yewman: If no longer using the allocator, remove the arg

src/core/epoch_tracker.zig
```diff
@@ -490,12 +574,11 @@ const UnrootedEpochBuffer = struct {
 
     pub const MAX_FORKS = 4;
 
-    fn deinit(self: *const UnrootedEpochBuffer, allocator: Allocator) void {
+    fn deinit(self: *const UnrootedEpochBuffer, _: Allocator) void {
```
yewman: Same here

yewman (commented):
Lgtm, just one small nit

src/core/epoch_tracker.zig
prestonsn: sweet, thanks! https://github.com/Syndica/sig/pull/1254/changes/c64b011e37a742abbd6a03c6eab6d284e6f6b8fa

src/core/epoch_tracker.zig
prestonsn: https://github.com/Syndica/sig/pull/1254/changes/c64b011e37a742abbd6a03c6eab6d284e6f6b8fa

src/consensus/progress_map.zig
```diff
@@ -1247,26 +1247,30 @@ test "ForkProgress.init" {
...
+        break :blk .{
+            .constants = consts,
+            .state = state,
+            .allocator = allocator,
+        };
     };
+    defer slot_element.constants.deinit(allocator);
+    defer slot_element.state.deinit(allocator);
+    const slot_info: replay.trackers.SlotTracker.Reference = slot_element.toRef();
+    defer slot_info.release();
```
kprotty: Is there a possible double/invalid-free here? `defer release()` -> last rc calls destroy():
* where destroy() calls `allocator.destroy(element_ptr)` but the element here is on the stack
* where destroy() frees constants & state, which are also freed above

Is the `slot_info` here used down below? If not, could be removed entirely

prestonsn: This is 100% an invalid free. unfortunately `slot_info` is used later in this unit test. I think the simplest fix for this, since the `Element` for this test is stack allocated, is to just bypass the ref counting stuff. https://github.com/Syndica/sig/pull/1254/commits/b32929843f263601f4b7000984215722ed76836c

src/core/epoch_tracker.zig
```diff
@@ -423,36 +505,35 @@ const RootedEpochBuffer = struct {
     buf: [4]?*const EpochInfo = @splat(null),
     root: Atomic(Epoch) = .init(0),
```
yewman: No need for this to be an atomic since it is now wrapped in a RwMux

src/core/epoch_tracker.zig
```diff
@@ -423,36 +505,35 @@ const RootedEpochBuffer = struct {
...
 
-    fn deinit(self: *const RootedEpochBuffer, allocator: Allocator) void {
+    fn deinit(self: *const RootedEpochBuffer) void {
```
yewman: This is my error, but we should fix it here 😛 The stack copy means setting the maybe_entry.* = null is a noop. Changing to take a mutable pointer and iterating on the buffer directly fixes this. The same applies to the UnrootedEpochBuffer
```suggestion
    fn deinit(self: *RootedEpochBuffer) void {
```

prestonsn: Good catch tho, fixed. https://github.com/Syndica/sig/pull/1254/commits/ec59ed713d07728a6aed63fdf9421532faaebcc7

prestonsn: github is degraded rn hence the ci issues

src/core/epoch_tracker.zig
prestonsn: true! fixed: https://github.com/Syndica/sig/pull/1254/commits/4aa12c0c0d873fca7a9c46c58287735a2c100580

yewman (approved):

PR #1256, opened 2026-02-25
feat(rpc): getBlockHeight, getTransactionCount, getHighestSnapshotSlot, getEpochInfo & getLatestBlockhash

hamza-syndica:
## Summary

- Implement five new RPC methods: `getBlockHeight`, `getTransactionCount`, `getHighestSnapshotSlot`, `getEpochInfo`, and `getLatestBlockhash`
- `getHighestSnapshotSlot` returns `null` for now since accounts-db v2 doesn't have a relevant implementation yet

### DIFF
--- a/src/rpc/methods.zig
+++ b/src/rpc/methods.zig
@@ -1629,33 +1629,6 @@ pub const RpcHookContext = struct {
             .delinquent = dlinqt,
         };
     }
-
-    pub fn getBalance(
-        self: RpcHookContext,
-        allocator: std.mem.Allocator,
-        params: GetBalance,
-    ) !GetBalance.Response {
-        const config: common.CommitmentSlotConfig = params.config orelse .{};
-        const resolved = try self.resolveSlot(config.commitment, config.minContextSlot);
-        defer resolved.ref.release();
-        const slot_reader = self.account_reader.forSlot(&resolved.ref.constants().ancestors);
-
-        // Look up account
-        const maybe_account = try slot_reader.get(allocator, params.pubkey);
-
-        const lamports: u64 = if (maybe_account) |account| blk: {
-            defer account.deinit(allocator);
-            break :blk account.lamports;
-        } else 0;
-
-        return .{
-            .context = .{
-                .slot = resolved.slot,
-                .apiVersion = ClientVersion.API_VERSION,
-            },
-            .value = lamports,
-        };
-    }
 };
 
 pub const StaticHookContext = struct {
@@ -1727,7 +1700,6 @@ fn testRpcHookContext(slot_tracker: *sig.replay.trackers.SlotTracker) RpcHookCon
     return .{
         .slot_tracker = slot_tracker,
         .epoch_tracker = undefined, // not used by getBlockHeight/getTransactionCount/getHighestSnapshotSlot
-        .account_reader = .noop,
     };
 }
 
@@ -1738,7 +1710,6 @@ fn testRpcHookContextWithEpochTracker(
     return .{
         .slot_tracker = slot_tracker,
         .epoch_tracker = epoch_tracker,
-        .account_reader = .noop,
     };
 }
 


ajw221 (approved):

src/rpc/test_serialize.zig
```diff
@@ -179,7 +181,23 @@ test GetGenesisHash {
...
+    try testRequest(.getHighestSnapshotSlot, .{},
+        \\{"jsonrpc":"2.0","id":1,"method":"getHighestSnapshotSlot","params":[]}
+    );
+    try testResponse(GetHighestSnapshotSlot, .{ .result = .{
+        .full = 100,
+        .incremental = 110,
+    } },
+        \\{"jsonrpc":"2.0","result":{"full":100,"incremental":110},"id":1}
+    );
+    try testResponse(GetHighestSnapshotSlot, .{ .result = .{
```
yewman, line 315: What about the error case? i.e. No snapshot

src/rpc/methods.zig
```diff
@@ -738,6 +751,139 @@ pub const RpcHookContext = struct {
...
+        // https://github.com/anza-xyz/agave/blob/v3.1.8/rpc/src/rpc.rs#L270-L285
+        const commitment = config.commitment orelse .finalized;
+        const slot = self.slot_tracker.getSlotForCommitment(commitment);
+
+        if (config.minContextSlot) |min_slot| {
+            if (slot < min_slot) return error.RpcMinContextSlotNotMet;
+        }
+
+        const slot_ref = self.slot_tracker.get(slot) orelse return error.SlotNotAvailable;
+
```
yewman: This block is used in many methods, and given the missing min context slot in the previous PR, perhaps this is justification for extracting it into a method on the RpcHookContext. What are your thoughts?

hamza-syndica: Yeah, https://github.com/Syndica/sig/pull/1256

src/rpc/test_serialize.zig
hamza-syndica: https://github.com/Syndica/sig/pull/1256

src/rpc/methods.zig
hamza-syndica: there are other PRs which will also need these common functions, I will update those once it's on main

src/rpc/methods.zig
```diff
@@ -730,12 +745,143 @@ pub const RpcHookContext = struct {
...
+    }
+
+    /// [agave] https://github.com/anza-xyz/agave/blob/v3.1.8/rpc/src/rpc.rs#L955-L958
+    pub fn getBlockHeight(
+        self: RpcHookContext,
+        _: std.mem.Allocator,
+        params: GetBlockHeight,
+    ) !GetBlockHeight.Response {
+        const config = params.config orelse common.CommitmentSlotConfig{};
+        const resolved = try self.resolveSlot(config.commitment, config.minContextSlot);
```
Sobeston, line 1420: nit: types on left preferred, i.e. `const config: common.CommitmentSlotConfig = params.config orelse .{};`

src/rpc/methods.zig
```diff
@@ -730,12 +745,143 @@ pub const RpcHookContext = struct {
...
+
+    /// for the time being we will return null
+    /// since accounts-db v2 don't have relevant implementation
+    pub fn getHighestSnapshotSlot(
+        _: RpcHookContext,
+        _: std.mem.Allocator,
+        _: GetHighestSnapshotSlot,
+    ) !GetHighestSnapshotSlot.Response {
+        return null;
+    }
```
Sobeston, line 1445: Yeah we don't generate snapshots at the moment. We could return the slot of the snapshot which we loaded from? But this method probably isn't useful yet

src/rpc/methods.zig
```diff
@@ -730,12 +745,143 @@ pub const RpcHookContext = struct {
...
+        const config = params.config orelse common.CommitmentSlotConfig{};
+        const resolved = try self.resolveSlot(config.commitment, config.minContextSlot);
+        defer resolved.ref.release();
+
+        // Read blockhash queue under lock
+        const bq, var bq_lock = resolved.ref.state().blockhash_queue.readWithLock();
+        defer bq_lock.unlock();
+
+        const last_hash = bq.last_hash orelse return error.SlotNotAvailable;
+
```
Sobeston: Where possible it's a good idea to reduce the duration locks are held
```suggestion
        const last_hash = hash: {
            const bq, var bq_lock = resolved.ref.state().blockhash_queue.readWithLock();
            defer bq_lock.unlock();
            
            break :hash bq.last_hash orelse return error.SlotNotAvailable;
        };
```

src/rpc/methods.zig
```diff
@@ -730,12 +745,143 @@ pub const RpcHookContext = struct {
...
+            .absoluteSlot = resolved.slot,
+            .blockHeight = resolved.ref.constants().block_height,
+            .transactionCount = resolved.ref.state().transaction_count.load(.monotonic),
+        };
+    }
+
+    /// [agave] https://github.com/anza-xyz/agave/blob/v3.1.8/rpc/src/rpc.rs#L2352-L2365
+    pub fn getLatestBlockhash(
+        self: RpcHookContext,
+        allocator: std.mem.Allocator,
```
Sobeston: if this function assumes an arena is passed in, it's good to make that obvious e.g. `arena: std.mem.Allocator`

src/rpc/methods.zig
Sobeston: (resolve this comment once read, not requesting a change here)

src/rpc/methods.zig
hamza-syndica: https://github.com/Syndica/sig/pull/1256/commits/73c5c6fd220ca73e937285a17810d71f27639808

src/rpc/methods.zig
hamza-syndica: https://github.com/Syndica/sig/pull/1256/commits/73c5c6fd220ca73e937285a17810d71f27639808

src/rpc/methods.zig
hamza-syndica: https://github.com/Syndica/sig/pull/1256/commits/73c5c6fd220ca73e937285a17810d71f27639808

PR #1257, opened 2026-02-25
fix(ci): update docs CI Zig version

Rexicon226:


.github/workflows/docs.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       - name: setup-zig
         uses: mlugg/setup-zig@v2
         with:
-          version: 0.14.1
+          version: 0.15.1
```
yewman: Haha I was literally about to open the same PR. Why not use 0.15.2? Also there are a bunch of other places in the codebase still referencing 0.14.1, could we address those in this PR as well?

Rexicon226: No reason, tbh forgot that there was a 0.15.2. I've updated the rest of the relevant references to 0.14.

Sobeston (approved):

PR #1260, opened 2026-02-26
feat(rpc): add jrpc websockets

jbuckmccready:
## Summary

Adds jrpc websocket layer on top of `webzockets`. Events are mocked, this doesn't integrate with Sig, PR for integration kept separate for easier review, integration PR is here: https://github.com/Syndica/sig/pull/1277

## High Level Data Flow

```mermaid
flowchart TD
  subgraph PT["Producer thread in Sig"]
    E1["Generate slot, logs, account, etc. events"]
    E2["Push EventMsg into inbound channel"]
  end

  subgraph LT["Loop thread xev"]
    L1["Async wakeup drains inbound events"]
    L2["Match events against active subscription keys"]
    L3["Submit SerializeJob for each matching queue"]
    L4["Drain CommitMsg from workers"]
    L5["Commit payload into NotifQueue"]
    L6["Wake handlers that have pending notifications"]

    subgraph HT["Per connection handler"]
      H1["Prioritize pending rpc responses"]
      H2["Read next payload from each subscribed queue"]
      H3["Batch payloads up to max batch bytes"]
      H4["Write batch to network socket"]
    end
  end

  subgraph WT["Serialization worker pool"]
    W1["Serialize JSON RPC notification body"]
    W2["Attach WebSocket text frame header"]
    W3["Publish CommitMsg"]
  end

  E1 --> E2 --> L1 --> L2 --> L3 --> W1 --> W2 --> W3 --> L4 --> L5
  L6 --> H2
  H1 --> H2 --> H3 --> H4
  L5 -.->|"reserved commit path for ordered events"| L6
  L5 -.->|"direct commit path for unordered events"| L6
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,9 +15,6 @@ executors:
   linux-executor:
     machine: true
     resource_class: syndica/r7iz-fleet
-  python-executor:
-    docker:
-      - image: cimg/python:3.10
   macos-executor:
     macos:
       xcode: 15.4.0
@@ -34,7 +31,7 @@ jobs:
       - run:
           name: Download Zig
           command: |
-            sudo apt update && sudo apt install -y wget unzip xz-utils
+            sudo apt-get update -y && sudo apt-get install -y wget unzip xz-utils
             wget https://ziglang.org/download/0.15.2/zig-<< parameters.target >>-0.15.2.tar.xz
             tar xf zig-<< parameters.target >>-0.15.2.tar.xz
             mkdir -p workspace
@@ -44,65 +41,7 @@ jobs:
           paths:
             - "*"
 
-  lint:
-    executor: linux-executor
-    steps:
-      - checkout
-      - attach_workspace:
-          at: workspace
-      - run:
-          name: Lint
-          command: |
-            workspace/zig/zig fmt --check src/ build.zig
-            cd conformance/
-            ../workspace/zig/zig fmt --check src/ build.zig
-
-  check_style:
-    executor: python-executor
-    steps:
-      - checkout
-      - run:
-          name: Install dependencies
-          command: pip install --upgrade pip
-      - run:
-          name: Check style
-          command: python scripts/style.py --check src
-      - run:
-          name: Check style in conformance
-          command: python scripts/style.py --check conformance/src
-      - run:
-          name: Check style in v2
-          command: python scripts/style.py --check v2/src
-
-  v2_checks:
-    executor: linux-executor
-    steps:
-      - checkout
-      - attach_workspace:
-          at: workspace
-      - run:
-          name: CI
-          command: |
-            sudo apt update && sudo apt install wget -y
-            ./scripts/proxy_workaround.sh workspace/zig/zig
-            cd v2/
-            ../workspace/zig/zig build ci --summary all
-
-  check_docs:
-    executor: python-executor
-    steps:
-      - checkout
-      - run:
-          name: Install dependencies
-          command: pip install --upgrade pip
-      - run:
-          name: Check docs
-          command: python docs/check.py ./
-      - run:
-          name: Check conformance commits
-          command: conformance/scripts/verify-commits.sh
-
-  build_linux:
+  build_and_test_linux:
     executor: linux-executor
     steps:
       - checkout
@@ -113,7 +52,7 @@ jobs:
       - run:
           name: Build
           command: |
-            sudo apt update && sudo apt install wget -y
+            sudo apt-get update -y && sudo apt-get install wget -y
             ./scripts/proxy_workaround.sh workspace/zig/zig
             workspace/zig/zig build -Denable-tsan=true -p workspace/zig-out -Dcpu=x86_64_v3 --summary all
       - save_cache:
@@ -121,10 +60,13 @@ jobs:
           paths:
             - .zig-cache
             - ~/.cache/zig
-      - persist_to_workspace:
-          root: workspace
-          paths:
-            - "zig-out/bin/test"
+      - run:
+          name: Build Tests
+          # Disable network-accessing tests for this job, which behave badly on circleci
+          command: workspace/zig/zig build test -Dno-run -Dlong-tests -Dcpu=x86_64_v3 -Denable-tsan=true -Dno-network-tests --summary all
+      - run:
+          name: Run Tests
+          command: zig-out/bin/test 2>&1 | cat
 
   build_linux_release:
     executor: linux-executor
@@ -137,7 +79,7 @@ jobs:
       - run:
           name: Build
           command: |
-            sudo apt update && sudo apt install wget -y
+            sudo apt-get update -y && sudo apt-get install wget -y
             ./scripts/proxy_workaround.sh workspace/zig/zig
             workspace/zig/zig build sig fuzz -Dno-run -Denable-tsan=false -Doptimize=ReleaseSafe -Dcpu=x86_64_v3 -p workspace/zig-out-release --summary all
       - run:
@@ -228,53 +170,6 @@ jobs:
             - .zig-cache
             - ~/.cache/zig
 
-  build_and_test_hashmap:
-    executor: linux-executor
-    steps:
-      - checkout
-      - attach_workspace:
-          at: workspace
-      - restore_cache:
-          key: linux-x86_64-0.15.2-{{ checksum "build.zig" }}-{{ checksum "build.zig.zon" }}-selfhosted
-      - run:
-          name: Build Tests
-          command: workspace/zig/zig build test -Dno-run -Dlong-tests -Denable-tsan=true -Dledger=hashmap -Dcpu=x86_64_v3 -Dfilter="ledger" --summary all
-      - run:
-          name: Run Tests
-          command: zig-out/bin/test 2>&1 | cat
-
-  test_linux:
-    executor: linux-executor
-    steps:
-      - checkout
-      - attach_workspace:
-          at: workspace
-      # Restore the cache in order to have access to the files which the DWARF info
-      # is referencing when dumping stack traces.
-      - restore_cache:
-          key: linux-x86_64-0.15.2-{{ checksum "build.zig" }}-{{ checksum "build.zig.zon" }}-selfhosted
-      - run:
-          name: Build Tests
-          # Disable network-accessing tests for this job, which behave badly on circleci
-          command: workspace/zig/zig build test -Dno-run -Dlong-tests -Dcpu=x86_64_v3 -Denable-tsan=true -Dno-network-tests --summary all
-      - run:
-          name: Run Tests
-          command: zig-out/bin/test 2>&1 | cat
-
-  test_webzockets:
-    executor: linux-executor
-    steps:
-      - checkout
-      - attach_workspace:
-          at: workspace
-      - restore_cache:
-          key: linux-x86_64-0.14.1-{{ checksum "build.zig.zon" }}-selfhosted
-      - run:
-          name: Build and Run Webzockets Tests
-          command: |
-            cd src/rpc/webzockets
-            ../../../workspace/zig/zig build test -Dcpu=x86_64_v3 --summary all
-
   test_webzockets_macos:
     executor: macos-executor
     steps:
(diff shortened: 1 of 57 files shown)


src/rpc/jrpc_websockets/handler.zig
```diff
@@ -0,0 +1,615 @@
...
+        // Batch available notifications into one send. Cap batch size to avoid blocking the event
+        // loop when a single handler has a large backlog of pending notifications.
+        self.send_state.notif_batch_buf.clearRetainingCapacity();
+        var batch_count: u64 = 0;
+        var batch_bytes: u64 = 0;
+
+        var start_idx: usize = 0;
+        var first_pick = true;
+        while (first_pick or batch_bytes < self.ctx.max_batch_bytes) {
+            first_pick = false;
```
kprotty: Could the `first_pick` variable be removed? If `max_batch_bytes is 0 on first iteration, does it still need to entire the loop?

src/rpc/jrpc_websockets/NotifQueue.zig
```diff
@@ -0,0 +1,1395 @@
...
+pub fn subscriberCount(self: *const NotifQueue) usize {
+    return self.subscribers.items.len;
+}
+
+/// Register a subscriber.
+///
+/// Callers must not register the same handler pointer more than once.
+/// Returns `error.TooManySubscribers` when count exceeds `u32`.
+pub fn addSubscriber(self: *NotifQueue, ptr: *JRPCHandler) !void {
+    if (self.subscribers.items.len == std.math.maxInt(u32)) {
```
kprotty, line 232: This will effectively never be hit.

src/rpc/jrpc_websockets/metrics.zig
```diff
@@ -0,0 +1,120 @@
...
+    /// Total bytes in JSON-RPC responses accepted for sending.
+    responses_sent_bytes: u64 = 0,
+    /// Notifications skipped due to subscriber lag (IndexOverwritten).
+    skipped_notifications: u64 = 0,
+    /// Send errors (unexpected errors, not WriteBusy/InvalidState).
+    send_errors: u64 = 0,
+    /// Total payload buffers freed (last Rc reference released).
+    payloads_freed: std.atomic.Value(u64) = std.atomic.Value(u64).init(0),
+    /// Total wall time spent freeing payload buffers (ns).
+    payload_free_wall_ns: std.atomic.Value(u64) = std.atomic.Value(u64).init(0),
```
kprotty, line 69: why are these tracked?

src/rpc/jrpc_websockets/protocol.zig
```diff
@@ -0,0 +1,691 @@
...
+    const raw = try account.data.readAllAllocate(allocator);
+    defer allocator.free(raw);
+
+    const b64 = std.base64.standard.Encoder;
+
+    const result: EncodedAccountData = switch (encoding) {
+        // TODO: integrate jsonParsed encoding
+        .base64, .jsonParsed => blk: {
+            const buf = try allocator.alloc(u8, b64.calcSize(raw.len));
+            const encoded = b64.encode(buf, raw);
```
kprotty, line 155: add assert that encoded.len == buf.len, otherwise frees could be invalid

src/rpc/jrpc_websockets/ws_request.zig
```diff
@@ -0,0 +1,706 @@
...
+        });
+    }
+
+    pub const Dynamic = struct {
+        jsonrpc: ?[]const u8 = null,
+        id: Id,
+        method: ?[]const u8 = null,
+        params: ?[]const std.json.Value = null,
+
+        pub const ParseDiagnostic = union {
```
kprotty: use tagged union: `union(enum)`

src/rpc/jrpc_websockets/protocol.zig
```diff
@@ -0,0 +1,691 @@
...
+const types = @import("types.zig");
+const methods = @import("methods.zig");
+const NotifPayload = sig.sync.RcSlice(u8);
+
+const Id = sig.rpc.request.Id;
+/// Responses are small and infrequent, done on IO loop, we just reuse an ArrayList
+const ResponseBuffer = std.ArrayList(u8);
+
+/// Prepend a WebSocket text frame header to the response buffer contents.
+fn prependFrameHeader(
```
kprotty, line 15: why this + connection.sendRaw() over e.g. sendText() in the ws lbirary?

src/rpc/jrpc_websockets/serialize.zig
```diff
@@ -0,0 +1,106 @@
...
+            .{ .payload = p }
+        else |err|
+            .{ .serialize_error = err };
+
+        const end_ns = std.time.nanoTimestamp();
+        const serialize_ns_i = end_ns - start_ns;
+        const serialize_ns: u64 = if (serialize_ns_i > 0)
+            @intCast(serialize_ns_i)
+        else
+            0;
```
kprotty: cleaner to create a u64 and use saturating subtraction:
```zig
const end_ns: u64 = @intCast(std.time.nanoTimestamp());
const serialize_ns = start_ns -| end_ns;
```

src/rpc/jrpc_websockets/serialize.zig
```diff
@@ -0,0 +1,106 @@
...
+    fn deinit(self: *SerializeTask) void {
+        self.job.event_data.deinit(self.allocator);
+        self.allocator.destroy(self);
+    }
+
+    fn execute(task: *ThreadPool.Task) void {
+        const self: *SerializeTask = @alignCast(@fieldParentPtr("task", task));
+        defer self.deinit();
+
+        const start_ns = std.time.nanoTimestamp();
```
kprotty: nit: we should be using `std.time.Instant` instead as its higher precision, monotonic, and not changeable by the user. But atp its a lot to change and probably doesnt matter.

src/rpc/jrpc_websockets/serialize.zig
```diff
@@ -0,0 +1,106 @@
...
+            result.deinit(self.allocator);
+            return;
+        };
+
+        const already_pending = self.notify_pending.swap(true, .release);
+        if (!already_pending) {
+            self.loop_async.notify() catch {
+                _ = self.notify_pending.swap(false, .release);
+            };
+        }
```
kprotty: This handling should be a method in `RuntimeContext`, so that this job doesnt have to store its individual fields. This `SerializeTask` could also move there too.

src/rpc/jrpc_websockets/metrics.zig
```diff
@@ -0,0 +1,120 @@
+const std = @import("std");
+
+// TODO: use actual metrics patterns for Sig.
+
+/// Runtime metrics counters for the JRPC WebSocket broadcast server.
+pub const Metrics = struct {
```
kprotty, line 6: This file could be deleted, its not really exposed to the client.

src/rpc/jrpc_websockets/handler.zig
jbuckmccready: Hmm, this code got modified quite a bit over time, but currently the `first_pick` is just ensuring at least one message will be written even if configured `max_batch_bytes` is 0 (forcing only 1 message per write regardless of size). Simplified and preserved the `do while` logic: https://github.com/Syndica/sig/commit/d1be878657a89ac695a89698126d8193bc67cb37

src/rpc/jrpc_websockets/NotifQueue.zig
jbuckmccready: Yep, I just didn't want to let it do very bad things if somehow it does. I can remove it and just leave comment, not sure what preferred approach is.

src/rpc/jrpc_websockets/metrics.zig
jbuckmccready: Used in smoke test/benchmarks. Should be removed/replaced by Sig metrics in integration PR or post Sig refactor.

src/rpc/jrpc_websockets/protocol.zig
jbuckmccready: This code just temporary for mock/benchmark, to be fully replaced by work done by @prestonsn for account encoding.

jbuckmccready: https://github.com/Syndica/sig/commit/155271923a081509584739f35e1d75b73b403800

src/rpc/jrpc_websockets/ws_request.zig
jbuckmccready: Copied from https://github.com/Syndica/sig/blob/jbuckmccready/jrpc-ws/src/rpc/request.zig#L58-L58
Updated in both places: https://github.com/Syndica/sig/commit/8a516999b6c49b6ad406976797ff3d20332e00d8

Need to do cleanup/improvement pass of RPC server generally I think and combine/share some of this code (I put in TODOs).

src/rpc/jrpc_websockets/protocol.zig
jbuckmccready: It batch writes websocket messages to TCP socket to get much better performance, and to in the future support offloading websocket per messsage deflate compression to threadpool.

- `NotifQueue` payloads are: `[ws_header|ws_payload]` as a single contiguous buffer, created by thread pool threads.
- TCP write (`sendRaw`) is passed a concatenated buffer: `[ws_header|ws_payload|ws_header|ws_payload|...]`.

In the future with ws compression then:
- `NotifQueue` payloads are: `[ws_header|compressed_payload]` as a single contiguous buffer, created by threadpool threads.
- TCP write (`sendRaw`) is passed a concatenated buffer:
`[ws_header|compressed_payload|ws_header|compressed_payload|...]`.

src/rpc/jrpc_websockets/metrics.zig
jbuckmccready: Used in smoke test/benchmarks.

src/rpc/jrpc_websockets/serialize.zig
jbuckmccready: https://github.com/Syndica/sig/commit/69214ff9c41aa21056c3b73671fb2597add9f08b

src/rpc/jrpc_websockets/serialize.zig
jbuckmccready: https://github.com/Syndica/sig/commit/2d808975fdaf024199ddfc153956ae0173acb4f8

src/rpc/jrpc_websockets/serialize.zig
jbuckmccready: https://github.com/Syndica/sig/commit/7e1725624f530138f6293fa6232101ab2a82eac2

src/rpc/jrpc_websockets/NotifQueue.zig
kprotty: its fine

src/rpc/jrpc_websockets/protocol.zig
kprotty: 1. is the perf worth it over relying on Nagle to batch?

2. thought the future would be a rewrite in v2

jbuckmccready: 1. Short answer yes. From my experience Nagle is never useful for websockets. It's impossible to use if you care about latency at all (adds up to 40ms), and is not even good for batching due to being generic (easier to batch optimally for the particular use case). Much better is to just have a linger interval for batch flushing on the user side if throughput is an issue. Had to test/experiment this for previous infra stuff for ws broadcasting to land on 128kB batch size with 5ms linger interval and Nagle disabled.
2. Not sure which parts will be rewritten, but I think packing framed messages into a buffer for batched tcp write will be the best approach for anything wanting to broadcast on websockets. I don't see anything changing about this part of the architecture if websockets is a protocol that needs to be supported. The main changes I anticipate would be in other things like adding linger flush interval, batching threadpool tasks, integrate proper metrics, further optimizing batching into the notification queue payloads, multiple IO threads, etc.

kprotty: 1. makes sense
 
2. v2 will have a different concurrency (custom threading) and IO (no xev) model, so most of this likely wont be kept re:future

for the purpose of this PR though, you can resolve this comment thread, then I can approve.

PR #1267, opened 2026-02-27
feat(rpc): impl getProgramAccounts

prestonsn:
Implements [getProgramAccounts](https://solana.com/docs/rpc/http/getprogramaccounts). Queries accounts_db for all accounts where the owner is the pubkey specified in the request. 

Requires scanning Rooted and Unrooted. First, rooted is scanned, then depending on the requests specified commitment level (`.confirmed` and `.processed`), Unrooted is scanned too. The account writes found in unrooted are then used to replace the equivalent ones from the initial Rooted scan. This way, we're returning the _current_ state of the accounts, matching Agave's behavior. 

The new `getByOwner` query is placed behind a runtime flag (since adding the index does incur overhead on every write to Rooted): 

```bash
❯ ./zig-out/bin/sig   validator -c testnet --rpc-enable-owner-index
```

### Performance 

Tracy was added to see how long each scan made by the RPC thread in order to service a full getProgramAccounts request takes. The test request used was a getProgramAccounts request for the vote program (`Vote111111111111111111111111111111111111111`) on testnet. This responds with ~15000 vote accounts (77MiB response size) when no filters are used. For testing, base64 encoding and `.confirmed` commitment was specified in the request. 

#### Without the index: 

image

~10s of time spent scanning Rooted. 

#### With the index: 

image

Rooted scan down to ~100ms. 

### Testing 

| What's Tested | Accounts | Response Size |
|---|---|---|
| Unfiltered (Config program) | 1,252 | 1.3 MB |
| Non-existent program | 0 | - |
| dataSize filter (643, 8) | 1,193 / 4 | - |
| memcmp filter | 43 | - |
| Combined memcmp + dataSize (Token program) | 10,596 | 4.7 MB |
| dataSlice (offset/length variations) | 1,252 | - |
| withContext wrapper | 4 | - |
| Encodings (base64, base64+zstd, jsonParsed) | 4 | - |
| Commitment (finalized, confirmed, processed) | 4 | - |
| jsonParsed + dataSlice combo | 4 | - |
| Invalid program IDs (empty, short, long) | - | - |
| Invalid filters (bad type, bad base58, negative offset) | - | - |
| Filter limits (4 accepted, 5 rejected) | 0 | - |
| memcmp bytes too large (200 chars) | - | - |
| minContextSlot (current, future) | 4 | - |
| Vote program unfiltered (big request) | 15,401 | 77.1 MB |

### DIFF
--- a/src/rpc/hook_contexts/Account.zig
+++ b/src/rpc/hook_contexts/Account.zig
@@ -504,7 +504,8 @@ pub fn getProgramAccounts(
         });
     }
 
-    if (config.sortResults orelse false) {
+    // [agave] https://github.com/anza-xyz/agave/blob/v3.1.8/rpc/src/rpc.rs#L3361
+    if (config.sortResults orelse true) {
         const z = tracy.Zone.init(@src(), .{ .name = "rpc.gPA.sort" });
         defer z.deinit();
         std.mem.sortUnstable(GetProgramAccounts.Value, results.items, {}, struct {


src/accountsdb/Rooted.zig
```diff
@@ -176,6 +205,81 @@ pub fn get(
     };
 }
 
+threadlocal var owner_pk: Pubkey = undefined;
+pub fn getByOwner(self: *Rooted, owner: *const Pubkey) OwnerIterator {
+    const stmt: *sql.sqlite3_stmt = if (get_by_owner_stmt) |stmt| stmt else blk: {
+        const query =
+            \\ SELECT address, lamports, data, owner, executable, rent_epoch
+            \\ FROM entries WHERE owner = ?;
```
jbuckmccready: Unneeded optimization at this point (maybe just  throw in a TODO comment): it would be nice to have the `getProgramAccount` parameters passed in and applied via dynamic query. It would allow applying filters and only fetching smaller data slices. Also noticed later we just discard zero sized accounts so they can be filtered out at DB level as well.

src/rpc/hook_contexts/Account.zig
```diff
@@ -41,33 +45,24 @@ pub fn getAccountInfo(
...
-        // Currently, `error.Base58DataTooLarge` propagates to hooks.zig's generic error mapper,
-        // which produces a non-deterministic positive error code via `@intFromError` and the raw
-        // error name as the message.
-        try account_codec.encodeStandard(allocator, account, encoding, config.dataSlice);
+    // TODO: [agave conformance] When base58 encoding is requested and account data exceeds
+    // 128 bytes, Agave returns JSON-RPC error code -32600 (InvalidRequest) with the message:
+    // "Encoded binary (base 58) data should be less than 128 bytes, please use Base64 encoding."
+    // Currently, `error.Base58DataTooLarge` propagates to hooks.zig's generic error mapper,
+    // which produces a non-deterministic positive error code via `@intFromError` and the raw
+    // error name as the message.
```
jbuckmccready, line 74: We should add an issue to track this, the RPC code should be mapping whatever errors it can get to actual JRPC errors.

src/rpc/hook_contexts/Account.zig
```diff
@@ -105,3 +100,129 @@ pub fn getBalance(
...
+                sig.core.Account{
+                    .lamports = entry.lamports,
+                    .data = .{ .unowned_allocation = entry.data },
+                    .owner = entry.owner,
+                    .executable = entry.executable,
+                    .rent_epoch = entry.rent_epoch,
+                };
+            if (account.lamports == 0) continue;
+            // Owner may have changed in a newer unrooted slot; skip if no longer ours.
+            if (!account.owner.equals(&params.program_id)) continue;
```
jbuckmccready: The control flow here is weird: we only need to check `account.owner.equals(&params.program_id)` if there was a newer version loaded, but this control flow always checks it.

src/accountsdb/Rooted.zig
prestonsn: That's a great idea

prestonsn: I double checked agave curious if they do such a thing, but they too just read out all the accounts data. Could be a good optimization we do down the line (https://github.com/anza-xyz/agave/blob/2717084afeeb7baad4342468c27f528ef617a3cf/accounts-db/src/accounts_db.rs#L3662). 

Added the TODO comment in for now: https://github.com/Syndica/sig/pull/1267/commits/3b46a24bccb319d543b3bd397af4b0b93380f39b

src/rpc/hook_contexts/Account.zig
prestonsn: Agreed, I do have a draft PR that adds proper error conformance: https://github.com/Syndica/sig/pull/1268

It isn't just error codes that currently mismatch, but also that some errors that agave returns are dynamically typed structs at times. It's a little annoying to handle cleanly in zig, but that PR adds an approach we can follow. 

Created a tracking issue for this though: https://github.com/Syndica/sig/issues/1294?reload=1

src/rpc/hook_contexts/Account.zig
prestonsn: Agreed, fixed this by moving the owner equality check from the RPC handler to the OwnerQuery.next() call. Now it only applies the check to resoved entries from `Db.get()` (i.e the unrooted overrides): https://github.com/Syndica/sig/pull/1267/commits/fa4ce9e1bcc1fb30f5f5cbceca0b4b080fbe9cb1

src/rpc/hook_contexts/Account.zig
```diff
@@ -26,9 +29,10 @@ pub fn getAccountInfo(
     // [agave] Default commitment is finalized:
     // https://github.com/anza-xyz/agave/blob/v3.1.8/rpc/src/rpc.rs#L348
     const commitment = config.commitment orelse .finalized;
-    // [agave] Default encoding in Agave is `Binary` (legacy base58):
+    // [agave] Default encoding in AGave is `Binary` (legacy base58):
```
Sobeston: AGave

src/rpc/hook_contexts/Account.zig
```diff
@@ -225,3 +220,81 @@ pub fn getTokenSupply(
...
+
+    var results = std.ArrayListUnmanaged(GetProgramAccounts.Value){};
+    defer {
+        for (results.items) |v| v.account.data.deinit(arena);
+        results.deinit(arena);
+    }
+
+    while (try iter.next()) |entry| {
+        const pubkey, const account = entry;
+        defer account.deinit(arena);
```
Sobeston: Worth noting that using an arena in a loop like this may cause requests to use large amounts of memory. Let's get this in before fixing that though

src/rpc/server/basic.zig
```diff
@@ -394,16 +403,19 @@ fn handleRpcRequest(
...
                 method,
                 @field(rpc_request.method, @tagName(method)),
             ) catch |e| switch (e) {
                 error.MethodNotImplemented => {
+                    handler_zone.deinit();
                     try sendFinalMethodNotFound(request, logger, method, rpc_request.id);
                     return;
                 },
             };
+            handler_zone.deinit();
```
Sobeston: Might be better to use a block for the zone to make the deinit less error prone using a defer. Fine for now

build.zig
```diff
@@ -565,6 +574,7 @@ fn generateFeatureSetId(b: *Build) Build.LazyPath {
     // import a library in the future.
     const gen = b.addExecutable(.{
         .name = "gen_feature_set_id",
+        .use_llvm = true,
         .root_module = b.createModule(.{
             .target = b.graph.host,
```
Sobeston, line 569: was this change intended?

src/utils/types.zig
```diff
@@ -349,6 +349,16 @@ pub fn eqlCustom(a: anytype, b: @TypeOf(a), comptime config_: EqlConfig) bool {
...
+    if (T == std.json.Value) {
+        // std.json.Value contains ObjectMap with many-item pointers that
+        // generic eql cannot recurse into. Fall back to JSON string comparison.
+        const stringify = std.json.stringifyAlloc;
+        const a_str = stringify(std.heap.page_allocator, a, .{}) catch return false;
+        defer std.heap.page_allocator.free(a_str);
+        const b_str = stringify(std.heap.page_allocator, b, .{}) catch return false;
+        defer std.heap.page_allocator.free(b_str);
+        return std.mem.eql(u8, a_str, b_str);
+    }
```
Sobeston: This feels pretty dirty, ideally we'd have a better Value.eql. 

Is this intended for usage in tests only? If so we should guard for that

src/accountsdb/Unrooted.zig
```diff
@@ -125,6 +126,34 @@ pub fn get(
...
+/// The caller is responsible for resolving each pubkey to its latest account state
+/// which avoids borrowing account data across lock boundaries.
+pub fn getByOwner(
+    self: *Unrooted,
+    allocator: std.mem.Allocator,
+    owner: Pubkey,
+    ancestors: *const Ancestors,
+) !PubkeyMap(void) {
+    var map = PubkeyMap(void){};
+    errdefer map.deinit(allocator);
```
Sobeston:
```suggestion
    var map: PubkeyMap(void) = .empty;
    errdefer map.deinit(allocator);
```

src/rpc/hook_contexts/Account.zig
prestonsn: FIxed: https://github.com/Syndica/sig/pull/1267/commits/799aaccaf36a1f706f962928e5647f11bfee9213

src/accountsdb/Unrooted.zig
prestonsn: https://github.com/Syndica/sig/pull/1267/commits/a396c4a97a63938e700bf585960381026098aa04

src/utils/types.zig
prestonsn: Should not be in the code, was due to messy rebase: https://github.com/Syndica/sig/pull/1267/commits/c1fed3d103dac5d76485a292d09eb2a1e0a7fc85

build.zig
prestonsn: Nope, that was part of some unrelated debugging (messed up rebase at some point): https://github.com/Syndica/sig/pull/1267/commits/c1fed3d103dac5d76485a292d09eb2a1e0a7fc85

Thanks!

src/rpc/server/basic.zig
prestonsn: https://github.com/Syndica/sig/pull/1267/commits/5b93747f23e0e250f74eb1a86d713599e9122e38

Good catch, thanks!

src/rpc/hook_contexts/Account.zig
prestonsn: Agreed! The happy (skip accounts due to filter) path has LIFO allocation-deallocation pattern, so the arena shouldn't be growing for skipped accounts. But the accounts that end up getting selected do use a lot more memory than they should (allocation for read data from accounts db, and maybe an allocation for serialized data depending upon serialization format). 

Anyways, recent feedback from @yewman has been to remove deallocs for now and rely on the arena: https://github.com/Syndica/sig/pull/1267/commits/061381b45011dee5a9b9fe180a872f46955ec5b6

Sobeston (dismissed):

Sobeston (approved):

src/accountsdb/Db.zig
```diff
@@ -172,6 +173,80 @@ pub fn slotModifiedIterator(self: *Db, slot: Slot) ?SlotModifiedIterator {
...
+        self.unrooted_set.deinit(self.allocator);
+    }
+};
+
+pub fn ownerQuery(self: *Db, owner: *const Pubkey, ancestors: *const Ancestors) !OwnerQuery {
+    var unrooted_set = try self.unrooted.getByOwner(
+        self.allocator,
+        owner.*,
+        ancestors,
+    );
```
dnut: it's a bit awkward that we need to lock + scan over the unrooted store repeatedly. once here before looking at rooted, then once again for every single relevant account in the unrooted store. why don't you just accumulate all the accounts here up front in the values of this hashmap, instead of having void values? then in `next` can just iterate over this map first, and then the rooted map after.

src/rpc/hook_contexts/Account.zig
```diff
@@ -455,3 +451,76 @@ fn messageToRuntimeTransaction(
...
+        defer z.deinit();
+        break :blk try slot_reader.getByOwner(arena, &params.program_id);
+    };
+    defer iter.deinit();
+
+    var results = std.ArrayListUnmanaged(GetProgramAccounts.Value){};
+
+    while (try iter.next()) |entry| {
+        const pubkey, const account = entry;
+        if (account.lamports == 0) continue;
```
dnut: This filtering should happen in accountsdb. It's the contract of an account store. If an account has zero lamports, it doesn't exist and shouldn't be returned.

src/rpc/hook_contexts/Account.zig
```diff
@@ -455,3 +451,76 @@ fn messageToRuntimeTransaction(
...
+        const data_slice: []const u8 = switch (account.data) {
+            .unowned_allocation => |d| d,
+            .owned_allocation => |d| d,
+            else => @panic("gPA: unexpected AccountDataHandle variant"),
+        };
```
dnut: why is this here? encodeAccount uses the iterator so it should be compatible with any variant.

src/rpc/hook_contexts/Account.zig
```diff
@@ -455,3 +451,76 @@ fn messageToRuntimeTransaction(
...
+
+    if (config.sortResults orelse false) {
+        const z = tracy.Zone.init(@src(), .{ .name = "rpc.gPA.sort" });
+        defer z.deinit();
+        std.mem.sortUnstable(GetProgramAccounts.Value, results.items, {}, struct {
+            fn lessThan(_: void, a: GetProgramAccounts.Value, b: GetProgramAccounts.Value) bool {
+                return std.mem.order(u8, &a.pubkey.data, &b.pubkey.data) == .lt;
+            }
+        }.lessThan);
+    }
```
dnut: i don't see sortResults in the api spec

build.zig
```diff
@@ -27,6 +27,7 @@ pub const Config = struct {
     use_llvm: bool,
     error_tracing: ?bool,
     long_tests: bool,
+    rpc_enable_owner_index: bool,
```
dnut: Why is this a build option? Runtime options are more flexible so generally preferable, unless there is a specific requirement or benefit of hardcoding it at comptime

src/accountsdb/Rooted.zig
```diff
@@ -176,6 +206,80 @@ pub fn get(
...
+            const len: usize = @intCast(sql.sqlite3_column_bytes(self.stmt, 2));
+            if (len == 0) break :blk &.{};
+            const ptr: [*]const u8 = @ptrCast(sql.sqlite3_column_blob(self.stmt, 2));
+            break :blk ptr[0..len];
+        };
+        const owner: Pubkey = .{
+            .data = @as([*]const u8, @ptrCast(sql.sqlite3_column_blob(self.stmt, 3)))[0..32].*,
+        };
+        const executable: bool = sql.sqlite3_column_int(self.stmt, 4) != 0;
+        const rent_epoch: u64 = @bitCast(sql.sqlite3_column_int64(self.stmt, 5));
```
dnut: isn't this identical to the logic in `get`? can this be a helper function?

src/rpc/filters.zig
```diff
@@ -0,0 +1,526 @@
...
+        if (obj.get("memcmp")) |val| {
+            return .{ .memcmp = try Memcmp.jsonParseFromValue(allocator, val, options) };
+        }
+        // [agave] The value for tokenAccountState is ignored — only key presence matters.
+        if (obj.contains("tokenAccountState")) {
+            return .tokenAccountState;
+        }
+
+        return error.UnexpectedToken;
+    }
```
dnut, line 82: why is this necessary? is this not the default parsing logic?

src/rpc/filters.zig
```diff
@@ -0,0 +1,526 @@
...
+                };
+                break :blk result;
+            },
+        };
+
+        return .{
+            .offset = offset,
+            .bytes = decoded,
+        };
+    }
```
dnut, line 160: My instinct would have been to parse first to this struct:
```zig
struct {
    offset: usize,
    bytes: []const u8,
    encoding: enum { base58, base64 },
}
```
then convert the bytes. might have been less boilerplate. your approach is fine too.

src/accountsdb/Db.zig
prestonsn: Yeah I was certainly over complicating things here. My thinking was that I wanted to ensure the response had the latest states of the accounts, hence the use of `Db.get()` to fetch the actual state of the account from `Unrooted` if present.  The other aspect I wanted to avoid was copying out account data while holding the Unrooted's entry locks by just copying out only the pubkeys, and calling `Db.get()` for each of them later when needed, minimizing time spent with the `SlotIndex` locks taken. 

But I don't think this really matters as much. I made the above change you suggested and re-ran my tests on testnet and things are working as expected: https://github.com/Syndica/sig/pull/1267/commits/8fd69243468237c3f102bf146dc4214daec7d11c

src/rpc/hook_contexts/Account.zig
prestonsn: That makes sense, moved it into each of the stores' queries: https://github.com/Syndica/sig/pull/1267/commits/2235f067e4f1c76977305aec912efe6cd79e8c19

src/rpc/hook_contexts/Account.zig
prestonsn: Oops, I forgot `AccountDataHandle` has a pretty nice api for this sorta stuff: https://github.com/Syndica/sig/pull/1267/commits/63601879baa5149cbc7280e1968c74ed414499b2

src/rpc/hook_contexts/Account.zig
prestonsn: Yeah it's weird that it's not documented anywhere, but it's present in agave's implementation:
* https://github.com/anza-xyz/agave/blob/v3.1.8/rpc-client-types/src/config.rs#L169
* https://github.com/anza-xyz/agave/blob/2717084afeeb7baad4342468c27f528ef617a3cf/rpc/src/rpc.rs#L3356-L3365

build.zig
prestonsn: No good reason: https://github.com/Syndica/sig/pull/1267/commits/f7c00fee2b5acf9c4f957a6529041295a0e2516c

src/accountsdb/Rooted.zig
prestonsn: Yup, was able to consolidate: https://github.com/Syndica/sig/pull/1267/commits/77316eac4f6e4298b74551e62791ecde76005864

src/rpc/filters.zig
prestonsn: Yeah this logic was necessary to support an inconsistency that would otherwise arise if we use default parsing logic, ex:

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "getProgramAccounts",
  "params": [
    "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA",
    {
      "filters": [
        {"tokenAccountState": null}
      ]
    }
  ]
}
```

This is what clients could send (serde deserializes unit types from json null). The default `std.json` union parser would reject null for a void field (seems it requirs the value to be an object `{}`)

prestonsn: I did improve the doc comment above this code tho: https://github.com/Syndica/sig/pull/1267/commits/8a4b67ef88b01a4fa6663fb5c63925c839464d7d

src/rpc/filters.zig
prestonsn: I tried this out, but I run into memory leaks (that get caught by the testing allocator). I believe this has to do with  `std.json.innerParseFromValue`'s handling of `[]const u8` always allocates a copy, so if a later field fails to parse (failure case testing),, it does not free the memory used for the slice copy. 

But this is just on the failure path, and really just requires updating all the unit tests...

src/rpc/hook_contexts/Account.zig
```diff
@@ -456,3 +451,70 @@ fn messageToRuntimeTransaction(
...
+            slot_reader,
+            config.dataSlice,
+        );
+        try results.append(arena, .{
+            .pubkey = pubkey,
+            .account = .from(account, data),
+        });
+    }
+
+    if (config.sortResults orelse false) {
```
hamza-syndica: agave: `sort_results.unwrap_or(true)`

https://github.com/anza-xyz/agave/blob/v3.1.8/rpc/src/rpc.rs#L3361

prestonsn: good catch, thanks! https://github.com/Syndica/sig/pull/1267/commits/2f575878f3ddc41afed03c695326a3b5cea915e0

src/accountsdb/Db.zig
dnut: Originally you were just doing shallow copies so there is not really a performance penalty for copying Account vs Pubkey, they're about the same size.

Now with the deep copy, you may want to consider the performance impact. But batching it actually reduces the total time holding the lock. It just concentrates it with one expensive hold instead of spreading it out over many holds, which might be good or bad. Hard to tell without profiling it. This is fine for now. If we really want to optimize this, there are other approaches that can avoid copying the account data.

src/rpc/hook_contexts/Account.zig
dnut: we should probably guarantee that the rooted store does not have any accounts with 0 lamports rather than needing this SQL filter. could put a debug assertion about it. not in scope for this pr though, this works well enough

src/rpc/hook_contexts/Account.zig
dnut: if it's not in the spec, we don't need to support it. but it's nice to have

PR #1270, opened 2026-03-02
feat(rpc): getMinimumBalanceForRentExemption, getStakeMinimumDelegation & getEpochSchedule

hamza-syndica:
## Summary

Add implementations for RPC methods: `getMinimumBalanceForRentExemption`, `getStakeMinimumDelegation`, and `getEpochSchedule`.

### DIFF
--- a/src/rpc/hook_contexts/Consensus.zig
+++ b/src/rpc/hook_contexts/Consensus.zig
@@ -320,19 +320,9 @@ pub fn getMinimumBalanceForRentExemption(
     }
 
     const config = params.config orelse common.CommitmentSlotConfig{};
-    // [agave] Default commitment is finalized:
-    // https://github.com/anza-xyz/agave/blob/v3.1.8/rpc/src/rpc.rs#L348
-    const commitment = config.commitment orelse .finalized;
-
-    const slot = self.slot_tracker.getSlotForCommitment(commitment);
-    if (config.minContextSlot) |min_slot| {
-        if (slot < min_slot) return error.RpcMinContextSlotNotMet;
-    }
-
-    // Get slot reference to access rent collector
-    const ref = self.slot_tracker.get(slot) orelse return error.SlotNotAvailable;
-    defer ref.release();
-    const rent = ref.constants().rent_collector.rent;
+    const resolved = try self.resolveSlot(config.commitment, config.minContextSlot);
+    defer resolved.ref.release();
+    const rent = resolved.ref.constants().rent_collector.rent;
 
     // [agave] https://github.com/anza-xyz/agave/blob/v3.1.8/runtime/src/bank.rs#L2719-L2720
     // minimum_balance returns 0 for empty accounts, but agave returns max(1, min_balance)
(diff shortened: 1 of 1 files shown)


src/rpc/methods.zig
```diff
@@ -889,6 +909,87 @@ pub const RpcHookContext = struct {
...
+        if (params.data_len > sig.runtime.program.system.MAX_PERMITTED_DATA_LENGTH) {
+            return error.InvalidRequest;
+        }
+
+        const config = params.config orelse common.CommitmentSlotConfig{};
+        // [agave] Default commitment is finalized:
+        // https://github.com/anza-xyz/agave/blob/v3.1.8/rpc/src/rpc.rs#L348
+        const commitment = config.commitment orelse .finalized;
+
+        const slot = self.slot_tracker.getSlotForCommitment(commitment);
```
yewman: Are we missing a min context slot check here?

src/rpc/methods.zig
```diff
@@ -889,6 +909,87 @@ pub const RpcHookContext = struct {
...
+        self: RpcHookContext,
+        _: std.mem.Allocator,
+        params: GetStakeMinimumDelegation,
+    ) !GetStakeMinimumDelegation.Response {
+        const config = params.config orelse common.CommitmentSlotConfig{};
+        // [agave] Default commitment is finalized:
+        // https://github.com/anza-xyz/agave/blob/v3.1.8/rpc/src/rpc.rs#L348
+        const commitment = config.commitment orelse .finalized;
+
+        const slot = self.slot_tracker.getSlotForCommitment(commitment);
```
yewman: Are we missing min context slot check here?

src/rpc/test_serialize.zig
```diff
@@ -225,7 +226,19 @@ test GetLeaderSchedule {
 
 // TODO: test getMaxRetransmitSlot()
 // TODO: test getMaxShredInsertSlot()
-// TODO: test getMinimumBalanceForRentExemption()
+
+test "GetMinimumBalanceForRentExemption" {
```
yewman:
```suggestion
test GetMinimumBalanceForRentExemption {
```

src/rpc/test_serialize.zig
```diff
@@ -225,7 +226,19 @@ test GetLeaderSchedule {
...
+    const GetMinimumBalanceForRentExemption = methods.GetMinimumBalanceForRentExemption;
```
yewman: Import this at the top level with others

src/rpc/methods.zig
hamza-syndica: https://github.com/Syndica/sig/pull/1270/changes/f92dfaabc0bb2a4319916618136d5599d8a127a1

src/rpc/methods.zig
hamza-syndica: https://github.com/Syndica/sig/pull/1270/changes/f92dfaabc0bb2a4319916618136d5599d8a127a1

src/rpc/test_serialize.zig
hamza-syndica: https://github.com/Syndica/sig/pull/1270/changes/f92dfaabc0bb2a4319916618136d5599d8a127a1

src/rpc/test_serialize.zig
hamza-syndica: https://github.com/Syndica/sig/pull/1270/changes/f92dfaabc0bb2a4319916618136d5599d8a127a1

yewman (approved):

Sobeston (approved):

src/rpc/methods.zig
```diff
@@ -892,6 +912,95 @@ pub const RpcHookContext = struct {
...
+        const epoch_schedule = &self.epoch_tracker.epoch_schedule;
+
+        return .{
+            .slotsPerEpoch = epoch_schedule.slots_per_epoch,
+            .leaderScheduleSlotOffset = epoch_schedule.leader_schedule_slot_offset,
+            .warmup = epoch_schedule.warmup,
+            .firstNormalEpoch = epoch_schedule.first_normal_epoch,
+            .firstNormalSlot = epoch_schedule.first_normal_slot,
+        };
+    }
```
ajw221: can we precompile this value and move it into `StaticHookContext` since there's no work being done?

was just wondering if this would be possible since `getInflationRewards` calls `getEpochSchedule` and it would be convenient not to duplicate code

hamza-syndica: I am not sure to be honest. Over to you.

yewman: Whilst the epoch schedule is effectively constant at the moment there may be instances in the future where it needs to change during runtime. For example there have been discussions about reducing the slots per epoch, how that looks in practice I am uncertain but perhaps it is justification for not moving this to static hook context. Thoughts @dnut ?

Sobeston: > can we precompile this value and move it into StaticHookContext since there's no work being done?

You may be right

Personally at this point I would opt for the path of less thought - I'm also not *100%* sure that it'll always be safe to make static

Sobeston (approved):

src/rpc/methods.zig
dnut: I would just leave this as is for now.

The epoch schedule is allowed to vary between clusters based on their unique genesis configs. It's not something that can be defined at compile time.

In terms of being mutated over the lifetime of the cluster, I'm not sure if that's supported by agave. `leader_schedule_slot_offset` is the only field that could potentially be mutated without breaking a lot of stuff.

Sobeston (approved):

PR #1271, opened 2026-03-02
feat(rpc): getInflationGovernor, getInflationRate & getClusterNodes

hamza-syndica:
## Summary

New RPC methods:

- getInflationGovernor
- getInflationRate
- getClusterNodes

### DIFF
--- a/src/cmd.zig
+++ b/src/cmd.zig
@@ -1680,6 +1680,9 @@ fn validator(
     else
         null;
 
+    var prioritization_fee_cache: sig.rpc.hook_contexts.PrioritizationFeeCache = .EMPTY;
+    defer prioritization_fee_cache.deinit(allocator);
+
     var replay_service_state: ReplayAndConsensusServiceState = try .init(allocator, .{
         .app_base = &app_base,
         .account_store = .{ .accounts_db = &new_db },
@@ -1691,6 +1694,7 @@ fn validator(
         .voting_enabled = voting_enabled,
         .vote_account_address = maybe_vote_pubkey,
         .stop_at_slot = cfg.stop_at_slot,
+        .prioritization_fee_cache = &prioritization_fee_cache,
     });
     defer replay_service_state.deinit(allocator);
 
@@ -1703,6 +1707,7 @@ fn validator(
 
     try app_base.rpc_hooks.set(allocator, sig.rpc.hook_contexts.LedgerHookContext{
         .ledger = &ledger,
+        .epoch_schedule = loaded_snapshot.genesis_config.epoch_schedule,
         .epoch_tracker = &epoch_tracker,
         .commitments = &replay_service_state.replay_state.slot_tracker.commitments,
     });
@@ -1711,6 +1716,11 @@ fn validator(
         .slot_tracker = &replay_service_state.replay_state.slot_tracker,
         .account_reader = replay_service_state.replay_state.account_store.reader(),
     });
+
+    try app_base.rpc_hooks.set(
+        allocator,
+        sig.rpc.hook_contexts.PrioritizationFeeHookContext{ .cache = &prioritization_fee_cache },
+    );
 
     const replay_thread = try replay_service_state.spawnService(
         &app_base,
@@ -2667,15 +2677,6 @@ fn startGossip(
     try app_base.rpc_hooks.set(allocator, struct {
         info: ContactInfo,
 
-        pub fn getHealth(
-            _: @This(),
-            _: std.mem.Allocator,
-            _: anytype,
-        ) !sig.rpc.methods.GetHealth.Response {
-            // TODO: more intricate
-            return .ok;
-        }
-
         pub fn getIdentity(
             self: @This(),
             _: std.mem.Allocator,
@@ -2740,6 +2741,7 @@ const ReplayAndConsensusServiceState = struct {
             voting_enabled: bool,
             vote_account_address: ?Pubkey,
             stop_at_slot: ?Slot,
+            prioritization_fee_cache: ?*sig.rpc.hook_contexts.PrioritizationFeeCache = null,
         },
     ) !ReplayAndConsensusServiceState {
         var replay_state: replay.service.ReplayState = replay_state: {
@@ -2796,6 +2798,7 @@ const ReplayAndConsensusServiceState = struct {
                 .hard_forks = hard_forks,
                 .replay_threads = params.replay_threads,
                 .stop_at_slot = params.stop_at_slot,
+                .prioritization_fee_cache = params.prioritization_fee_cache,
             }, if (params.disable_consensus) .disabled else .enabled);
         };
         errdefer replay_state.deinit();
--- a/src/rpc/methods.zig
+++ b/src/rpc/methods.zig
@@ -76,7 +76,7 @@ pub const MethodAndParams = union(enum) {
     getMultipleAccounts: GetMultipleAccounts,
     getProgramAccounts: noreturn,
     getRecentPerformanceSamples: GetRecentPerformanceSamples,
-    getRecentPrioritizationFees: noreturn,
+    getRecentPrioritizationFees: GetRecentPrioritizationFees,
     getSignaturesForAddress: GetSignaturesForAddress,
     getSignatureStatuses: GetSignatureStatuses,
     getSlot: GetSlot,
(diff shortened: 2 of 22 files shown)


src/rpc/methods.zig
```diff
@@ -892,6 +920,180 @@ pub const RpcHookContext = struct {
...
         };
     }
+
+    pub fn getInflationGovernor(
+        self: RpcHookContext,
+        _: std.mem.Allocator,
+        params: GetInflationGovernor,
+    ) !GetInflationGovernor.Response {
+        const config = params.config orelse GetInflationGovernor.Config{};
+        const commitment = config.commitment orelse .finalized;
```
Sobeston: nit: prefer types on the left `const config: GetInflationGovernor.Config = params.config orelse .{};`

src/rpc/methods.zig
```diff
@@ -892,6 +920,180 @@ pub const RpcHookContext = struct {
...
+        defer gossip_lock.unlock();
+
+        // Get all contact infos matching our shred version
+        // We iterate through the gossip table contact infos directly to get full ContactInfo
+        var contact_info_iter = gossip_table.contactInfoIterator(0);
+        var result_list: std.ArrayListUnmanaged(common.RpcContactInfo) = .empty;
+        errdefer {
+            for (result_list.items) |item| {
+                freeRpcContactInfo(allocator, item);
+            }
```
Sobeston: would freeRpcContactInfo make more sense as RpcContactInfo.deinit?

src/rpc/methods.zig
```diff
@@ -892,6 +920,180 @@ pub const RpcHookContext = struct {
...
+                    "{d}.{d}.{d}",
+                    .{ v.major, v.minor, v.patch },
+                );
+                feature_set = v.feature_set;
+            }
+            errdefer if (version_str) |vs| allocator.free(vs);
+
+            // Build the RpcContactInfo response
+            const rpc_contact_info = common.RpcContactInfo{
+                .pubkey = try std.fmt.allocPrint(
```
Sobeston: nit: prefer types on the left, `const rpc_contact_info: common.RpcContactInfo = .{`

src/rpc/methods.zig
```diff
@@ -892,6 +920,180 @@ pub const RpcHookContext = struct {
...
+                    allocator,
+                    contact_info.getSocket(.tpu_forwards_quic),
+                ),
+                .tpuVote = try formatSocketAddr(allocator, contact_info.getSocket(.tpu_vote)),
+                .serveRepair = try formatSocketAddr(
+                    allocator,
+                    contact_info.getSocket(.serve_repair),
+                ),
+                .rpc = try formatSocketAddr(allocator, contact_info.getSocket(.rpc)),
+                .pubsub = try formatSocketAddr(allocator, contact_info.getSocket(.rpc_pubsub)),
```
Sobeston: Annoyingly this could cause leaks.

e.g. if this line fails
` .rpc = try formatSocketAddr(allocator, contact_info.getSocket(.rpc)),`

then all of the fields allocated before it wouldn't be cleaned up - there's no err defers.

Sobeston (changes requested):

src/rpc/methods.zig
hamza-syndica: https://github.com/Syndica/sig/pull/1271/commits/f2f5b4b48fc4a30c55c0ed07e3215079b5fb518d

src/rpc/methods.zig
hamza-syndica: https://github.com/Syndica/sig/pull/1271/commits/f2f5b4b48fc4a30c55c0ed07e3215079b5fb518d

src/rpc/methods.zig
hamza-syndica: https://github.com/Syndica/sig/pull/1271/commits/f2f5b4b48fc4a30c55c0ed07e3215079b5fb518d

src/rpc/methods.zig
hamza-syndica: https://github.com/Syndica/sig/pull/1271/commits/f2f5b4b48fc4a30c55c0ed07e3215079b5fb518d

src/rpc/methods.zig
```diff
@@ -1485,6 +1529,190 @@ pub const RpcHookContext = struct {
...
+            .total = inflation.total(slot_in_years),
+            .validator = inflation.validatorRate(slot_in_years),
+            .foundation = inflation.foundationRate(slot_in_years),
+            .epoch = epoch,
+        };
+    }
+
+    pub fn getClusterNodes(
+        self: RpcHookContext,
+        allocator: std.mem.Allocator,
```
yewman:
```suggestion
        arena: std.mem.Allocator,
```

src/rpc/methods.zig
```diff
@@ -1485,6 +1529,190 @@ pub const RpcHookContext = struct {
...
+            return error.ShredVersionNotAvailable;
+
+        const my_shred_version = my_shred_version_atomic.load(.monotonic);
+
+        const gossip_table, var gossip_lock = gossip_table_rw.readWithLock();
+        defer gossip_lock.unlock();
+
+        var contact_info_iter = gossip_table.contactInfoIterator(0);
+        var result_list: std.ArrayListUnmanaged(common.RpcContactInfo) = .empty;
+        errdefer {
```
yewman: All RPC methods are called using an arena allocator because it is required by the standard library json implementation. Freeing with an arena allocator is unusual in that if only actual frees memory if that memory was the most recent allocation, otherwise it is effictively a noop. Given that is the case, unless there is a specific reason to be freeing the last allocation, it is better to ommit frees because they are either unnecessary or a noop.

src/rpc/methods.zig
```diff
@@ -1485,6 +1529,190 @@ pub const RpcHookContext = struct {
...
+
+        return try result_list.toOwnedSlice(allocator);
+    }
+
+    /// Converts a gossip ContactInfo into RpcContactInfo. Returns null if the contact
+    /// should be skipped (shred version mismatch or invalid gossip address).
+    /// Caller owns the returned value and must call deinit on it (or append to a list
+    /// that is cleaned up on error).
+    fn contactInfoToRpc(
+        allocator: std.mem.Allocator,
```
yewman:
```suggestion
        arena: std.mem.Allocator,
```

src/rpc/methods.zig
```diff
@@ -1485,6 +1529,190 @@ pub const RpcHookContext = struct {
...
+        // Build RpcContactInfo incrementally so a single errdefer can clean up on any failure.
+        // deinit only frees non-null optional slices and pubkey, so it's safe for partial init.
+        var rpc_contact_info = common.RpcContactInfo{
+            .pubkey = try std.fmt.allocPrint(
+                allocator,
+                "{s}",
+                .{contact_info.pubkey.base58String().slice()},
+            ),
+        };
+        errdefer rpc_contact_info.deinit(allocator);
```
yewman: unnecessary or noop, consequently the deinit method on this struct is likely unused

src/rpc/methods.zig
hamza-syndica: I was not aware of this, updated: https://github.com/Syndica/sig/pull/1271/commits/676b44bebe2dc540337313d161277e5ecac5c7bf

@prestonsn I think you also need to update similar in your `getVoteAccounts`. I could've done that, but just letting you know as well for your context.

src/rpc/methods.zig
hamza-syndica: https://github.com/Syndica/sig/pull/1271/commits/676b44bebe2dc540337313d161277e5ecac5c7bf

src/rpc/methods.zig
hamza-syndica: https://github.com/Syndica/sig/pull/1271/commits/676b44bebe2dc540337313d161277e5ecac5c7bf

src/rpc/methods.zig
hamza-syndica: https://github.com/Syndica/sig/pull/1271/commits/676b44bebe2dc540337313d161277e5ecac5c7bf

src/rpc/methods.zig
prestonsn: Yup, thanks! I'll create a PR to move the methods in `methods.zig` out to a file under `hook_contexts` then update those for the arena usage.

src/cmd.zig
```diff
@@ -1694,9 +1694,16 @@ fn validator(
...
     defer replay_service_state.deinit(allocator);
 
+    const account_store = sig.accounts_db.AccountStore{
+        .accounts_db = &new_db,
+    };
+
     try app_base.rpc_hooks.set(allocator, sig.rpc.methods.RpcHookContext{
         .slot_tracker = &replay_service_state.replay_state.slot_tracker,
         .epoch_tracker = &epoch_tracker,
+        .account_reader = account_store.reader(),
```
yewman: Is this used?

src/rpc/methods.zig
```diff
@@ -1629,6 +1658,178 @@ pub const RpcHookContext = struct {
...
+        const my_shred_version_atomic = self.my_shred_version orelse
+            return error.ShredVersionNotAvailable;
+
+        const my_shred_version = my_shred_version_atomic.load(.monotonic);
+
+        const gossip_table, var gossip_lock = gossip_table_rw.readWithLock();
+        defer gossip_lock.unlock();
+
+        var contact_info_iter = gossip_table.contactInfoIterator(0);
+        var result_list: std.ArrayListUnmanaged(common.RpcContactInfo) = .empty;
```
yewman: As of 0.15 std.ArrayList is unmanaged :)
```suggestion
        var result_list: std.ArrayList(common.RpcContactInfo) = .empty;
```

src/cmd.zig
hamza-syndica: Yes, it's used by some other PR. The whole rebase thing messed up few stuffs. Removed from this PR anyway.

https://github.com/Syndica/sig/pull/1271/commits/e21d71dfec9fbb305cd94581d7f08beab23b2421

src/rpc/methods.zig
hamza-syndica: `ArrayListUnmanaged` is still used in a lot of other places as well, should I give a separate PR just to rename those?

Updated this one: https://github.com/Syndica/sig/pull/1271/commits/e21d71dfec9fbb305cd94581d7f08beab23b2421

src/rpc/methods.zig
prestonsn: For reference: https://github.com/Syndica/sig/pull/1293

yewman (approved):

PR #1275, opened 2026-03-03
feat(rpc): impl getTokenAccountsByOwner

prestonsn:
Implements [getTokenAccountsByOwner](https://solana.com/docs/rpc/http/gettokenaccountsbyowner) 

Adds `--rpc-enable-spl-token-owner-index` runtime CLI flag that adds a nullable `token_owner` column to the rooted accounts DB, populated from SPL Token account data bytes `32..64` on write, with a partial index (`WHERE token_owner IS NOT NULL`) for fast lookups. When disabled, falls back to a full table scan using `substr(data, 33, 32)` filtered to the Token and Token-2022 program IDs.

### Testing
| Test | Agave | Sig |
|---|---|---|
| programId / base64 | 3 accts | 3 accts |
| mint / base64 | 1 acct | 1 acct |
| programId / jsonParsed | 3 accts | 3 accts |
| programId / base64+zstd | 3 accts | 3 accts |
| programId / dataSlice(0,32) | 3 accts | 3 accts |
| programId / System Program owner | 39 accts | 39 accts |

### DIFF
--- a/src/accountsdb/Rooted.zig
+++ b/src/accountsdb/Rooted.zig
@@ -561,7 +561,21 @@ fn putWithoutTokenOwner(
     const stmt: *sql.sqlite3_stmt = if (put_stmt) |stmt| stmt else blk: {
         // Insert or update only if last_modified_slot is greater (excluded = VALUES)
         // https://sqlite.org/lang_upsert.html#examples
-        const query =
+        const query = if (self.enable_spl_token_owner_index)
+            \\INSERT INTO entries
+            \\(address, lamports, data, owner, executable, rent_epoch, last_modified_slot)
+            \\VALUES (?, ?, ?, ?, ?, ?, ?)
+            \\ON CONFLICT(address) DO UPDATE SET
+            \\  lamports=excluded.lamports,
+            \\  data=excluded.data,
+            \\  owner=excluded.owner,
+            \\  executable=excluded.executable,
+            \\  rent_epoch=excluded.rent_epoch,
+            \\  last_modified_slot=excluded.last_modified_slot,
+            \\  token_owner=NULL
+            \\WHERE excluded.last_modified_slot > entries.last_modified_slot
+            \\;
+        else
             \\INSERT INTO entries
             \\(address, lamports, data, owner, executable, rent_epoch, last_modified_slot)
             \\VALUES (?, ?, ?, ?, ?, ?, ?)
--- a/src/accountsdb/account_store.zig
+++ b/src/accountsdb/account_store.zig
@@ -328,7 +328,6 @@ pub const SlotAccountReader = union(enum) {
 
     pub fn getBySplTokenOwner(
         self: SlotAccountReader,
-        _: Allocator,
         token_owner: *const Pubkey,
     ) !OwnerIterator {
         return switch (self) {
--- a/src/rpc/hook_contexts/Account.zig
+++ b/src/rpc/hook_contexts/Account.zig
@@ -567,7 +567,7 @@ pub fn getTokenAccountsByOwner(
     var iter = blk: {
         const z = tracy.Zone.init(@src(), .{ .name = "rpc.gTABO.splTokenOwnerQuery" });
         defer z.deinit();
-        break :blk try slot_reader.getBySplTokenOwner(arena, &params.owner);
+        break :blk try slot_reader.getBySplTokenOwner(&params.owner);
     };
     defer iter.deinit();
 


src/accountsdb/Unrooted.zig
```diff
@@ -176,6 +177,57 @@ pub fn getByOwner(
...
+            gop.value_ptr.* = .{
+                index.slot,
+                .{
+                    .lamports = acc.lamports,
+                    .data = .{ .owned_allocation = data },
+                    .owner = acc.owner,
+                    .executable = acc.executable,
+                    .rent_epoch = acc.rent_epoch,
+                },
+            };
```
kprotty: could use AccountSharedDta.clone() here

src/rpc/hook_contexts/Account.zig
```diff
@@ -520,3 +522,98 @@ pub fn getProgramAccounts(
...
     return .{ .list = values };
 }
+
+/// [agave] https://github.com/anza-xyz/agave/blob/v3.1.8/rpc/src/rpc.rs#L2091-L2130
+pub fn getTokenAccountsByOwner(
+    self: AccountHookContext,
+    arena: std.mem.Allocator,
+    params: GetTokenAccountsByOwner,
+) !GetTokenAccountsByOwner.Response {
+    const zone = tracy.Zone.init(@src(), .{ .name = "rpc.gTABO" });
```
kprotty: nit: better to spell out the zone name

src/rpc/hook_contexts/Account.zig
```diff
@@ -520,3 +522,98 @@ pub fn getProgramAccounts(
...
+    {
+        const z = tracy.Zone.init(@src(), .{ .name = "rpc.gTABO.sort" });
+        defer z.deinit();
+        std.mem.sortUnstable(GetTokenAccountsByOwner.Value, results.items, {}, struct {
+            fn lessThan(
+                _: void,
+                a: GetTokenAccountsByOwner.Value,
+                b: GetTokenAccountsByOwner.Value,
+            ) bool {
+                return std.mem.order(u8, &a.pubkey.data, &b.pubkey.data) == .lt;
```
kprotty: could use `Pubkey.order`

src/accountsdb/Unrooted.zig
prestonsn: https://github.com/Syndica/sig/pull/1275/commits/37a5788619de61dc0dc0cccef4f592b4fa720d9a

Not sure if it's any cleaner. I think to really clean this up I'd have to use `AccountSharedData` over `Account` here:
 https://github.com/Syndica/sig/blob/cee831d3d83e465cc6229b396ab67e797c531094/src/accountsdb/Unrooted.zig#L130

src/rpc/hook_contexts/Account.zig
prestonsn: https://github.com/Syndica/sig/pull/1275/commits/37a5788619de61dc0dc0cccef4f592b4fa720d9a

Also fixed up getProgramAccount's

src/rpc/hook_contexts/Account.zig
prestonsn: https://github.com/Syndica/sig/pull/1275/commits/37a5788619de61dc0dc0cccef4f592b4fa720d9a

much cleaner, thanks! :saluting_face:

src/accountsdb/Unrooted.zig
prestonsn: I was being silly, cleaned it up: https://github.com/Syndica/sig/pull/1275/commits/3da5aeab868ad4d386a9fa81fe1bf6dbcc162d87

src/accountsdb/Rooted.zig
```diff
@@ -473,7 +575,13 @@ pub fn put(self: *Rooted, address: Pubkey, slot: Slot, account: AccountSharedDat
             \\WHERE excluded.last_modified_slot > entries.last_modified_slot
```
yewman, line 576: On conflict should we enforce that the token_owner is reset to null in case it was already set?
```suggestion
            \\  token_owner=NULL
            \\WHERE excluded.last_modified_slot > entries.last_modified_slot
```

src/accountsdb/account_store.zig
```diff
@@ -325,6 +325,20 @@ pub const SlotAccountReader = union(enum) {
             }
         }
     };
+
+    pub fn getBySplTokenOwner(
+        self: SlotAccountReader,
+        _: Allocator,
```
yewman: Why take an unused allocator here?

src/accountsdb/Rooted.zig
prestonsn: Good catch, thank you 🫡 https://github.com/Syndica/sig/pull/1275/commits/7a9dcb8dbb754d5756e796c000dc9d6e3966d132

had to add an `else` since `token_owner` column only exists in the schema when `--rpc-enable-spl-token-owner-index` flag is set. 

The alternative here is to just modify the base schema to always have the `token_owner` column. This is what @kprotty suggested in an earlier review. But my concern with that was the potential of adding storage overhead per row. But I think with a `NULLABLE` column like `token_owner` it's 1 byte per row -> ~1 MiB for a million accounts, so negligible overhead perhaps? It does remove all the if/else stuff (except for the index itself). 

I just didn't want to make too many changes to the core Rooted logic just for RPC. But I can go either ways on this, @kprotty also suggested the above.

src/accountsdb/account_store.zig
prestonsn: Cleaned up: https://github.com/Syndica/sig/pull/1275/commits/7a9dcb8dbb754d5756e796c000dc9d6e3966d132

src/accountsdb/Rooted.zig
yewman: I am happy with the conditional approach 👍

PR #1283, opened 2026-03-05
feat(v2): telemetry service

### LINKED ISSUE #1265: feat(observability): v2 observability service to handle logging + prometheus
create a solution for logging and prometheus in v2. this most likely needs a dedicated service for at least the prometheus server, but this service does not strictly need to encapsulate all behavior. create a design plan and then let's discuss it before moving forward with implementation.

we should be able to reuse the types from v1 by placing them in shared memory

InKryption:
Implements the fundamental building blocks for our core uses of metrics and logging. The MetricId & gauge primitives can be used to implement APIs for, gauges, counters, and variant counters; the histogram API is the main stand-out that needs slightly deeper integration - in the future there may be a way to decompose this into simpler abstractions, this is just the most straightforward approach.

The procedure for this service is as follows:
1. The service starts up, and waits until all other services have registered a logger and all the metrics they need.
2. Other services acquire a logger stream, and atomically append metrics into a set of shared memory regions, and notify when they have registered all required metrics, holding onto pointers into the shared memory region to communicate updates.
3. The service aggregates all registered metrics into a sorted list.
4. A loop processes all incoming logs and prints them to its stderr, and then serializes all metrics to the prometheus client whenever said client issues a GET metrics request (currently done just using `std.net` & `std.http`).

Closes #1265.


v2/src/common/linux/bpf.zig
```diff
@@ -73,6 +73,13 @@ pub fn printSleepExit(maybe_stderr: ?std.os.linux.fd_t) [28]sock_filter {
         allowSyscall(@intFromEnum(syscalls.close)) ++
         allowSyscall(@intFromEnum(syscalls.bind)) ++
         allowSyscall(@intFromEnum(syscalls.socket)) ++
+        // observability
+        allowSyscall(@intFromEnum(syscalls.getsockname)) ++
+        allowSyscall(@intFromEnum(syscalls.listen)) ++
+        allowSyscall(@intFromEnum(syscalls.accept4)) ++
+        allowSyscall(@intFromEnum(syscalls.pwritev)) ++
+        allowSyscall(@intFromEnum(syscalls.readv)) ++
+        allowSyscall(@intFromEnum(syscalls.sendmsg)) ++
```
Sobeston: let's trim these down once we can define per-service filters, all good for now

v2/lib/observability/log.zig
```diff
@@ -0,0 +1,485 @@
...
+        const day_secs = epoch_secs.getDaySeconds();
+
+        const year_and_day = epoch_day.calculateYearDay();
+        const month_and_day = year_and_day.calculateMonthDay();
+
+        try w.print("{[YYYY]d:0>4}-{[MM]d:0>2}-{[DD]d:0>2}", .{
+            .YYYY = year_and_day.year,
+            .MM = month_and_day.month.numeric(),
+            .DD = month_and_day.day_index + 1,
+        });
```
Sobeston, line 468: very nice

v2/src/common/observability.zig
```diff
@@ -0,0 +1,851 @@
...
+    /// A service should call this when they want to signal to the observability service
+    /// that it has acquired a logger stream, and has registered all desired metrics.
+    pub fn signalReady(self: Regions) void {
+        std.debug.assert(self.startup.pending_services.fetchSub(1, .release) != 0);
+    }
+
+    pub fn acquireLogger(
+        self: Regions,
+        /// Assumes `str.len <= log.MessageStream.Name.MAX_LEN`.
+        name: []const u8,
```
Sobeston: For functions that assume something, can we enforce that with an explicit assert?

v2/src/services/observability.zig
```diff
@@ -0,0 +1,406 @@
...
+    _ = start;
+}
+
+pub const name = .observability;
+pub const panic = start.panic;
+pub const std_options = start.options;
+
+pub const ReadWrite = struct {
+    /// NOTE: some of these actually represent floats, and `std.atomic.Value(T)`s.
+    histogram_data: []u64,
```
Sobeston: Would it make sense to use a bare union or something for these?

Sobeston (dismissed):
Looking great, just one 1 question and 1 nit

v2/src/common/observability.zig
InKryption: This says "assume" because of the call to `stream.name.init`, which also makes this "assumption", which is described as such because the slice it does on the destination buffer for the `@memcpy` already makes this assertion for us.
Do you think we should make it an explicit assertion regardless?

v2/src/services/observability.zig
InKryption: Could make sense to use an extern or packed union in order to retain layout; can't have auto layout types like bare unions in shared memory.

v2/src/common/observability.zig
Sobeston: Personally I prefer any assumptions like this to be specified upfront, even if a little redundant

InKryption: f3fd4f8b4b39d3c0239f732dad09a6da35e888f2

Sobeston (dismissed):

v2/src/services/observability.zig
InKryption: I gave this a try, but there wasn't really much benefit; the way it's used doesn't really ever actually interact directly with the elements, and segments of it are ptrCasted to slices of the actual type, not just single item pointers.

v2/lib/ipc/ring.zig
```diff
@@ -63,77 +67,274 @@ pub fn Ring(N: comptime_int, T: type) type {
...
-                    };
-                    position.value.store(position.value.raw +% n, .release);
-                }
+        fn asRef(self: *RingSelf) RingRef(T) {
+            return .{
+                .head = &self.head,
+                .tail = &self.tail,
+                .array = &self.array,
+            };
+        }
```
kprotty: could avoid storing multiple pointers and instead store one:
```zig
type RingHeader(T: type) = extern struct:
  head: Pos align(cache_line),
  tail: Pos align(cache_line),
  array: [0]T,

type Ring(N: comptime_int, T: type) = extern struct:
  header: RingHeader(T),
  buf: [N]T,

type RingRef(T: type) = struct:
  header: *RingHeader(T),
  n: usize,
```

v2/lib/ipc/ring.zig
```diff
@@ -63,77 +67,274 @@ pub fn Ring(N: comptime_int, T: type) type {
...
+            position.value.store(position.value.raw +% n, .release);
+        }
+
+        pub fn reader(self: *RingSlice(u8, .reader), buffer: []u8) Reader {
+            return .init(self, buffer);
+        }
+
+        pub fn writer(self: *RingSlice(u8, .writer), buffer: []u8) Writer {
+            return .init(self, buffer);
         }
```
kprotty, line 371: a buffer in front of Ring defeats idea behind it being zero-copy. Not sure if readers and writers are needed for v2 (see previous comment on single-shot contiguous-msg logging). If it truly is, could init these with first(), or second() if first is empty.

v2/lib/observability/log.zig
```diff
@@ -0,0 +1,486 @@
...
+        };
+    }
+};
+
+pub const MessageStream = extern struct {
+    name: Name,
+    ring: RingBuffer,
+
+    /// 100 MiB byte ring buffer.
+    pub const RingBuffer = Ring(100 * 1024 * 1024, u8);
```
kprotty: This could instead be a 100 MiB fixed buffer that can be written to multiple times, but consumed in one go (to avoid adding reader/writer to Ring + avoids dealing with wrap around).

v2/lib/telemetry/log.zig
```diff
@@ -0,0 +1,486 @@
...
+pub const MessageStream = extern struct {
+    name: Name,
+    ring: RingBuffer,
+
+    /// 100 MiB byte ring buffer.
+    pub const RingBuffer = Ring(100 * 1024 * 1024, u8);
+
+    pub const Name = extern struct {
+        len: Len,
+        buf: [MAX_LEN]u8,
```
kprotty, line 30: me when the BoundedArray

v2/src/common/observability.zig
```diff
@@ -0,0 +1,853 @@
...
+        }
+    };
+
+    pub fn appendGauge(
+        self: MetricAppender,
+        metric_id: MetricId,
+        comptime kind: GaugeKind,
+        init_value: kind.Type(),
+    ) *std.atomic.Value(kind.Type()) {
+        const gauge_index = self.gauges_end.fetchAdd(1, .acq_rel);
```
kprotty: could this be `monotonic`

v2/src/common/observability.zig
```diff
@@ -0,0 +1,853 @@
...
+    };
+
+    /// Writes a value into the histogram.
+    pub fn observe(
+        self: *const Histogram,
+        /// Must be f64 or int
+        value: anytype,
+    ) void {
+        const float: f64 = if (@typeInfo(@TypeOf(value)) == .int) @floatFromInt(value) else value;
+        const shard_sync = self.incrementCount(.acquire); // acquires lock. must be first step.
```
kprotty: method could be inlined since used once.

v2/lib/telemetry.zig
```diff
@@ -0,0 +1,853 @@
...
+    /// Returns the state from before this operation, which was replaced by this operation.
+    fn incrementCount(self: *const Histogram, comptime ordering: std.builtin.AtomicOrder) ShardSync {
+        return @bitCast(self.shard_sync.fetchAdd(1, ordering));
+    }
+
+    /// Makes the hot shard cold and vice versa.
+    /// Returns the state from before this operation, which was replaced by this operation.
+    fn flipShard(self: *const Histogram, comptime ordering: std.builtin.AtomicOrder) ShardSync {
+        const shard_sync: ShardSync = .{ .shard = 1 };
+        const data = self.shard_sync.fetchAdd(@bitCast(shard_sync), ordering);
```
kprotty, line 611: neat trick keeping the bit at the end

v2/src/common/observability.zig
```diff
@@ -0,0 +1,853 @@
...
+    pub fn swapOutSnapshot(self: *const Histogram) SnapshotReader {
+        // Make the hot shard cold. Some writers may still be writing to it,
+        // but no more will start after this.
+        const shard_sync = self.flipShard(.monotonic);
+        const cold_shard = &self.shards[shard_sync.shard];
+        const hot_shard = &self.shards[shard_sync.shard +% 1];
+
+        // Wait until all writers are done writing to the cold shard
+        // TODO: maybe switch to a condvar? see: `std.Thread.Condition`
+        while (cold_shard.count.cmpxchgStrong(shard_sync.count, 0, .acquire, .monotonic)) |_| {
```
kprotty: should this instead be a `while load(count, acq) != shared_sync.count`? loads are faster than rmws (no frequent cache line invalidation)

v2/src/common/observability.zig
```diff
@@ -0,0 +1,853 @@
...
+            // Acquire on success: keeps shard usage after.
+        }
+
+        // Now the cold shard is totally cold and unused by other threads.
+        // - read the cold shard's data
+        // - zero out the cold shard.
+        // - write the cold shard's data into the hot shard.
+        const cold_shard_sum = cold_shard.sum.swap(0.0, .monotonic);
```
kprotty: nit: could be a load + store. No contention atp, but rmws are slower even serially

v2/src/common/observability.zig
```diff
@@ -0,0 +1,853 @@
...
+    }
+
+    /// Swaps in the hot shard for the cold shard, such that it can be viewed as a snapshot of
+    /// recent state.
+    /// The returned struct is used to access this state iteratively, before resetting the cold
+    /// shard's state, so it can be swapped out again in the future.
+    pub fn swapOutSnapshot(self: *const Histogram) SnapshotReader {
+        // Make the hot shard cold. Some writers may still be writing to it,
+        // but no more will start after this.
+        const shard_sync = self.flipShard(.monotonic);
```
kprotty: should be `acq_rel`:
- `rel`:  the memory zero/reset of the shard being swapped in happens-before `observe`s see the new shared with `acq`
- `acq`: flip happens-before `cold_shard.count.load`s below

v2/src/main.zig
```diff
@@ -82,6 +101,56 @@ pub fn main() !void {
                 .{ .instance = .{ .service = .shred_receiver } },
             },
         },
+        .{
```
kprotty: 1. these obs regions seem like they could be merged into one. Not sure if it's worth making the region distinction for the logs & histograms over the counters, even if the observability doesnt write to the counters.

2. instead of having regions for each type of metric, what about having a service define the set of metric types it needs into one struct, then share that with the observability service & that relevant service?

v2/build.zig
```diff
@@ -4,127 +4,166 @@ const Build = std.Build;
 pub fn build(b: *Build) !void {
     const target = b.standardTargetOptions(.{});
     const optimize = b.standardOptimizeOption(.{});
+    const force_llvm = b.option(bool, "force-llvm", "Force usage of LLVM.");
+    const artifact_opts: ExeOutput.InitOptions = .{
+        .no_bin = b.option(bool, "no-bin", "Don't install artifacts implied by specified steps.") orelse false,
+        .no_run = b.option(bool, "no-run", "Don't execute run steps implied by the specified steps.") orelse false,
```
kprotty, line 14: IIRC, the run step wasnt enabled by default like in v1

v2/lib/telemetry.zig
InKryption: yeah, not my trick, just ported from the original code

v2/build.zig
InKryption: The behavior here is such that when you run `zig build thing`, it will run `thing`. If it were inversed like you imply it should be, you would have to do `zig build thing -Dno-run=false`. This doesn't affect `zig build install`, which doesn't depend on any run steps.

kprotty: Have been doing `zig build` on its own, which doesnt run. Are there other intended build targets besides that and `zig build test`?

InKryption: `zig build` on its own is equivalent to `zig build install`; I'm not sure I really understand the question. To clarify the intent here: by default, when you specify steps that depend on a run step, it will run that run step. The way you disable that step from running is by specifying `-Dno-run`, which is useful when you just want it to build, and you specify `-Dno-run -Dno-bin` in conjunction when you just want to run semantic analysis on it (ie for `--watch` and/or `-fincremental`, for example). This is exactly how it works in v1 too.

v2/src/common/observability.zig
InKryption: 1b91a073a4da4cd0f96a1fef8df7527db622bb1f

v2/src/common/observability.zig
InKryption: 1b91a073a4da4cd0f96a1fef8df7527db622bb1f

v2/src/common/observability.zig
InKryption: 1b91a073a4da4cd0f96a1fef8df7527db622bb1f
I'll take your word for it

v2/src/common/observability.zig
InKryption: 1b91a073a4da4cd0f96a1fef8df7527db622bb1f

v2/src/common/observability.zig
InKryption: 1b91a073a4da4cd0f96a1fef8df7527db622bb1f

v2/lib/telemetry.zig
```diff
@@ -0,0 +1,854 @@
...
+        const cold_shard = &self.shards[shard_sync.shard];
+        const hot_shard = &self.shards[shard_sync.shard +% 1];
+
+        // Wait until all writers are done writing to the cold shard
+        while (true) {
+            const current_cold_count = cold_shard.count.load(.acquire);
+            if (current_cold_count != shard_sync.count) continue;
+            std.debug.assert(current_cold_count == shard_sync.count);
+            cold_shard.count.store(0, .monotonic);
+        }
```
yewman, line 480: Infinite loop

v2/src/common/observability.zig
```diff
@@ -0,0 +1,854 @@
...
+pub fn Logger(comptime scope_str: []const u8) type {
+    return struct {
+        sink: log.MessageSink,
+        max_level: log.Level,
+        const LoggerSelf = @This();
+
+        pub const scope = scope_str;
+
+        pub const noop: LoggerSelf = .{
+            .ring = null,
```
yewman:
```suggestion
            .sink = .noop,
```

v2/src/common/observability.zig
```diff
@@ -0,0 +1,854 @@
...
+            const LoggerOther = Logger(@TypeOf(logger).scope);
+            return LoggerOther.withScope(logger, scope);
+        }
+
+        pub fn withScope(
+            self: LoggerSelf,
+            comptime new_scope: []const u8,
+        ) Logger(new_scope) {
+            return .{
+                .ring = self.ring,
```
yewman:
```suggestion
                .sink = self.sink,
```

v2/src/services/observability.zig
```diff
@@ -0,0 +1,406 @@
...
+            );
+            const stderr = &stderr_fw.interface;
+            defer stderr.flush() catch {};
+            outer: for (log_streams) |*log_stream| {
+                // stream out at most 16 logs at a time from each service.
+                // limit chosen arbitrarily.
+                inner: for (0..16) |_| {
+                    switch (try streamLogs(stderr, log_stream.name.slice(), log_stream)) {
+                        .empty => continue :outer,
+                        .sent => break :inner,
```
yewman:
```suggestion
                        .sent => {},
```

v2/services/telemetry.zig
```diff
@@ -0,0 +1,406 @@
...
+
+    while (true) {
+        {
+            var stderr_buf: [4096]u8 = undefined;
+            var stderr_fw: std.fs.File.Writer = .init(
+                .{ .handle = start.panic_state.stderr },
+                &stderr_buf,
+            );
+            const stderr = &stderr_fw.interface;
+            defer stderr.flush() catch {};
```
yewman, line 75: flushed on line 86

v2/src/common/observability.zig
```diff
@@ -0,0 +1,854 @@
...
+                    std.debug.assert(label_index < label_count - 1);
+                    break;
+                },
+                '}' => {
+                    std.debug.assert(label_index == label_count - 1);
+                    break;
+                },
+                else => {},
+            }
+            if (r.buffered()[index] == ',') break;
```
yewman: redundant

v2/src/common/observability.zig
```diff
@@ -0,0 +1,854 @@
...
+    }
+};
+
+/// Data that's only relevant during startup.
+pub const Startup = extern struct {
+    /// The port to listen on for the prometheus client.
+    /// Mutating this is after initialization illegal.
+    port: u16,
+    /// The maximum log level to emit.
+    /// Mutating this is after initialization illegal.
```
yewman, line 19:
```suggestion
    /// Mutating this after initialization illegal.
```

v2/lib/telemetry.zig
InKryption: good catch
fa93fb2ba9cdf7ef3b9efb082a41c404115b7ac3

v2/src/common/observability.zig
InKryption: fbcd3a4a334d620ddb702eaad8940397bcb827f7

v2/src/common/observability.zig
InKryption: fbcd3a4a334d620ddb702eaad8940397bcb827f7

v2/services/telemetry.zig
InKryption: This defer is moreso a last-effort attempt to flush any remaining buffered data in the event that any code returns an error in between here and the flush on line 86; if we're going to crash and burn, nice to potentially still have the last few logs before that happens.

v2/src/services/observability.zig
InKryption: 59b75b3b764541f5ccf7ad152c2394d33e0cc07e

v2/src/common/observability.zig
InKryption: 59b75b3b764541f5ccf7ad152c2394d33e0cc07e

v2/src/common/observability.zig
InKryption: 59b75b3b764541f5ccf7ad152c2394d33e0cc07e

v2/src/services/observability.zig
InKryption: 59b75b3b764541f5ccf7ad152c2394d33e0cc07e

v2/src/common/observability.zig
```diff
@@ -0,0 +1,854 @@
...
+        // but no more will start after this.
+        const shard_sync = self.flipShard(.acq_rel);
+        const cold_shard = &self.shards[shard_sync.shard];
+        const hot_shard = &self.shards[shard_sync.shard +% 1];
+
+        // Wait until all writers are done writing to the cold shard
+        while (true) {
+            const current_cold_count = cold_shard.count.load(.acquire);
+            if (current_cold_count == shard_sync.count) break;
+            std.debug.assert(current_cold_count == shard_sync.count);
```
dnut: This will never be true

v2/lib/ipc/ring.zig
kprotty: being reworked in https://github.com/Syndica/sig/pull/1346
feel free to resolve

v2/init/services.zon
```diff
@@ -5,27 +5,58 @@
...
+            },
+        },
+        .{
+            .name = .observability,
+            .regions = .{
+                .{ .name = .obs_startup, .access = .readonly },
+                .{ .name = .obs_id_mem, .access = .readonly },
+                .{ .name = .obs_gauges, .access = .readonly },
+                .{ .name = .obs_histogram_data, .access = .rw },
+                .{ .name = .obs_log_streams, .access = .rw },
```
kprotty: Since these only differ in `readonly` for the observability service on `startup, id_mem, gauges`, could these all be coalesced into the same `.rw` shared memory type? Or at least two (`startup, id_mem, gauges`) + (`histogram_data`, `log_streams`)?

v2/config/testnet.zig.zon
```diff
@@ -9,4 +9,7 @@
     .shred_network = .{
         .recv_port = 8002,
     },
+    .observability = .{
```
dnut:
```suggestion
    .telemetry = .{
```

dnut: We agreed on this name change right? If so let's just get it taken care of now.

v2/init/main.zig
```diff
@@ -89,6 +107,24 @@ pub fn main() !void {
...
+        },
+        .obs_gauges = .{
+            .max_elements = 4096 * 2,
+        },
+        .obs_histogram_data = .{
+            .max_elements = 4096 * 3,
+        },
+        .obs_log_streams = .{
+            .max_log_streams = service_instances.len - 1,
+        },
```
dnut: In some cases I can see where you would never want to use the full name of the service. Like you wouldn't litter the code with `observability.logger` every time you want to log. In that case it would be justified to just alias `logger` directly. But if you had a short abbreviation like "obs" then you can justify prefixing it as `obs.logger` to give *more* information than you would without abbreviations. That's the one case where I can sympathize with using application-scoped abbreviations (though it's a slippery slope)

But in some cases it makes sense to use the full service name regardless of whether you have an abbreviation available. I think the high-level architecture definition of the validator is one of those cases.

v2/lib/lib.zig
```diff
@@ -7,6 +7,18 @@ pub const shred = @import("shred.zig");
...
+comptime {
+    _ = net;
+    _ = crypto;
+    _ = solana;
+    _ = shred;
+    _ = gossip;
+    _ = ipc;
+    _ = linux;
+    _ = observability;
+}
```
kprotty, line 26: did refAllDecls in the effects PR. Should we be more explicit with this in general?

v2/services/telemetry.zig
```diff
@@ -0,0 +1,492 @@
...
+                try streamLogs(.{
+                    .output = stderr,
+                    .service_name = log_stream.name.slice(),
+                    .log_messages_buffer = log_stream.swap_buffer.swap(),
+                });
+            }
+            try stderr.flush();
+        }
+
+        const conn = server.accept() catch |err| switch (err) {
```
kprotty, line 86: is there any way to set a timeout on the http connection's internal send/recvs? e.g. `setsockopt(conn.fd, SO_RCV_TIMEO)`

v2/init/main.zig
```diff
@@ -89,6 +107,24 @@ pub fn main() !void {
...
+            .max_bytes = 4096 * 16,
+        },
+        .obs_gauges = .{
+            .max_elements = 4096 * 2,
+        },
+        .obs_histogram_data = .{
+            .max_elements = 4096 * 3,
+        },
+        .obs_log_streams = .{
+            .max_log_streams = service_instances.len - 1,
```
dnut: Why are there separate regions for different metrics? It feels overly granular to express this at this level. We shouldn't need to change the application architecture just because a service starts or stops using a histogram. It should just be expressed that it does or does not need need to publish observability/telemetry data. I would have just made a single region and divided it within the client and server code.

v2/config/testnet.zig.zon
```diff
@@ -9,4 +9,7 @@
...
+        .port = 12345,
```
kprotty, line 13: nit: should it use the standard 9090 prometheus port?

v2/init/services.zig
```diff
@@ -94,15 +95,13 @@ const RequiredRegion = struct {
...
 
 pub const SharedRegionInstances = blk: {
     var fields: []const std.builtin.Type.StructField = &.{};
     for (service_region_fields) |r| {
-        const RegionType = @TypeOf(@field(
-            @as(Region, undefined),
-            @tagName(@field(services_zon.regions, r.name)),
-        ));
+        const region_tag = @field(services_zon.regions, r.name);
+        const RegionType = @FieldType(Region, @tagName(region_tag));
```
kprotty, line 104: I need to use `@FieldType` more often

dnut (changes requested):

v2/lib/lib.zig
InKryption: I don't have a super strong preference; at some point it will cease to matter once zig reworks unit test discovery, so it's a temporary fixture in my mind.

v2/services/telemetry.zig
InKryption: I don't think we want to block on this? How long prometheus takes to issue a request will depend on the refresh rate you set in the client.

v2/init/main.zig
InKryption: Well, there's only 1 region for the vast majority of metrics, and then an additional one for histograms. Everything else was divided as it is mostly for the sake of simplicity; can probably figure out a way to collapse these together, though would prefer to keep RW and RO regions separate.

v2/config/testnet.zig.zon
InKryption: 8ddc59125b284f393967dfcc386186b2da3f417b

v2/init/main.zig
InKryption: 8ddc59125b284f393967dfcc386186b2da3f417b

v2/config/testnet.zig.zon
InKryption: I believe there's other stuff we have set up that already presumes our prometheus stuff will go over 12345? Care to weigh in @dnut?

yewman: I don't think changing legacy usage of 12345 is much of an issue. I'd go with 9090

v2/init/services.zon
InKryption: d009f1de3c2f6da23f59ff2ee67cef60e25ab888
Tried to consolidate into two separate regions, but ended up being more complicated than it was worth. I'm fine with putting it all in RW for now, with the hope of hoisting telemetry above the service interface later, where this could probably be implemented with fewer compromises.

v2/init/main.zig
InKryption: d009f1de3c2f6da23f59ff2ee67cef60e25ab888

https://github.com/Syndica/sig/pull/1283#discussion_r3023709102

v2/config/testnet.zig.zon
dnut: We should not use 9090. Almost any other number would be fine though.

9090 is **not** the port number used for exposing data *to* prometheus. It's the port number used for exposing data *from* prometheus to other services, like grafana. Using 9090 would be misleading and cause potential port number collisions.

Here's the v1 architecture, with each service listing the port numbers it exposes.
```mermaid
flowchart LR
    subgraph sig["sig"]
        SIGM1["metrics :12345"]
        SIGM2["geyser metrics :12355"]
        SIGLOG["logs"]
    end

    subgraph Docker["Observability Stack"]
        PROM["Prometheus :9090"]
        NODE["Node Exporter :9100"]
        LOKI["Loki :3100"]
        PYRO["Pyroscope :4040"]
        GRAF["Grafana :3000"]
        ALLOY["Alloy :3200"]
    end

    PROM --> SIGM1
    PROM --> SIGM2
    PROM --> NODE

    ALLOY --> SIGLOG
    LOKI --> ALLOY

    GRAF --> PROM
    GRAF --> LOKI
    GRAF --> PYRO
```

There is no "correct" port number for exposing prometheus metrics. Prometheus never assumes metrics are on a specific port. You always have to specify it when you tell prometheus what to connect to.

The closest thing to a standard would be port 9100, because that's what the Node Exporter uses to expose its metrics. But using 9100 would collide with Node Exporter without much benefit.

dnut: In terms of selecting a port number, initially I suggested 9101, but then I realized incrementing up from 9100 would likely be a common pattern and could lead to more collisions. Turns out HAProxy uses 9101 for its metrics.

But frankly it does not matter that much. It would be nice to avoid a collision with node exporter since we do use node exporter, but even that is not a huge issue since it runs in docker.

I'll propose 9110. It's a simple increment up from 9100. I couldn't find any other tools that use this port number. Let's do 9110 unless there is a good reason to use a different number.

v2/init/services.zon
dnut: >  hoisting telemetry above the service interface 

what does this mean?

v2/services/net.zig
```diff
@@ -14,19 +15,35 @@ pub const name = .net;
...
+
 pub const ReadWrite = struct {
     gossip_pair: *Pair,
     shred_pair: *Pair,
+    tel: *tel.Region,
 };
 
-pub fn serviceMain(rw: ReadWrite) !noreturn {
-    try mainInner(&.{ rw.gossip_pair, rw.shred_pair });
+pub fn serviceMain(_: ReadOnly, rw: ReadWrite) !noreturn {
```
dnut, line 26: why add `_: ReadOnly,`?

v2/services/telemetry.zig
```diff
@@ -0,0 +1,491 @@
...
+        ctx: MetricIdHashCtx,
+        key: api.MetricId,
+    ) u32 {
+        _ = ctx;
+        return @truncate(key.hash());
+    }
+};
+
+fn collectMetrics(
+    gpa: std.mem.Allocator,
```
dnut: Let's use `allocator` unless there's some requirement for the allocator to be constrained to something more specific than the Allocator interface.

v2/lib/telemetry.zig
```diff
@@ -0,0 +1,966 @@
...
+        const hot_shard = &self.shards[shard_sync.shard +% 1];
+
+        // Wait until all writers are done writing to the cold shard
+        while (true) {
+            const current_cold_count = cold_shard.count.load(.acquire);
+            if (current_cold_count == shard_sync.count) {
+                cold_shard.count.store(0, .monotonic);
+                break;
+            }
+            std.debug.assert(current_cold_count < shard_sync.count);
```
dnut, line 479: isn't this a compare and swap loop?

```suggestion
        // Wait until all writers are done writing to the cold shard
        while (cold_shard.count.cmpxchgWeak(shard_sync.count, 0, .acquire, .monotonic)) |_| {
```

I guess it doesn't *need* to happen atomically. Once you reach the count, you can reasonably assume that it will stop changing. Is there an advantage of breaking it down into separate steps though?

v2/lib/telemetry.zig
```diff
@@ -0,0 +1,966 @@
...
+                return null;
+            }
+            defer self.current_bucket_index += 1;
+
+            const upper_bound = self.upper_bounds[self.current_bucket_index];
+            const cold_bucket = &self.cold_shard_buckets[self.current_bucket_index];
+            const hot_bucket = &self.hot_shard_buckets[self.current_bucket_index];
+
+            const count = cold_bucket.swap(0, .monotonic);
+            _ = hot_bucket.fetchAdd(count, .monotonic);
```
dnut, line 556: This is an interesting approach to put it in nextBucket. It's a little confusing though. Swapping the snapshot is not a complete operation. It leaves the hot shard partially incorrect, until you call nextBucket to completion. You better not have an error or exit early while handling the snapshot.

Why not just finish updating the hot shard in swapSnapshot? That gives cleaner guarantees with fewer assumptions about usage patterns across multiple methods. Swapping the shard actually finishes the job of swapping the shard.

v2/lib/telemetry.zig
```diff
@@ -0,0 +1,966 @@
...
+        const backing_element_count =
+            bucket_count + // upper_bounds
+            1 + // shard_sync
+            2 * elements_per_shard // shards
+        ;
+        return backing_element_count;
+    }
+
+    pub const BucketFromElementsCountError = error{ TooFewElements, InvalidFlexibleCount };
+    pub fn bucketFromElementsCount(total_elements: u64) BucketFromElementsCountError!u64 {
```
dnut, line 582: This pattern always trips me up. I find it very hard to understand this while skimming the file. I have to pause to read carefully to understand that this is separately a const and a function.
```suggestion
    pub const BucketFromElementsCountError = error{ TooFewElements, InvalidFlexibleCount };

    pub fn bucketFromElementsCount(total_elements: u64) BucketFromElementsCountError!u64 {
```

I think we should have a rule that function definitions should always have a new line before the next decl

v2/lib/telemetry.zig
```diff
@@ -0,0 +1,966 @@
...
+
+        pub fn Type(comptime kind: GaugeKind) type {
+            return switch (kind) {
+                .int => u64,
+                .float => f64,
+            };
+        }
+    };
+
+    pub fn appendGauge(
```
dnut: This is a nice primitive but I'd like to have some helpers on top of this for common patterns like Counter and maybe VariantCounter.

v2/services/net.zig
```diff
@@ -14,19 +15,35 @@ pub const name = .net;
...
+    const logger = rw.tel.acquireLogger(@tagName(name), "main");
+
+    const metric_appender = rw.tel.metricAppender();
+    _ = metric_appender;
```
dnut: So we don't have any usages of metrics anywhere? Why not add a packet counter as a POC?

v2/services/telemetry.zig
```diff
@@ -0,0 +1,491 @@
+/// This service consumes information from other services,
+/// and sends them to an aggregator (prometheus).
+const tel = @This();
```
dnut, line 3: There's a lot going on in this file. I thought we decided to have service files as just a thin lightweight execution layer to that loops to run components that are implemented in lib.

v2/services/net.zig
InKryption: See b5330e1e1c1f4d541d79ffe16ba1966d40d25f4b

Originally when I had all the data in separate regions, I simplified usage by making it possible to have non-pointer struct fields in `ReadOnly` and `ReadWrite`, which would be descended into to have their pointer fields populated with the relevant regions, and then providing a `Regions` struct which encapsulated all of the regions exposed by the telemetry service. While doing that, I also simplified the logic by making both parameters & their respective type declarations required.

This PR no longer takes direct advantage of these changes, but I believe they improve clarity and flexibility.

v2/services/telemetry.zig
InKryption: `gpa` simply implies that it is intended to be a general purpose allocator (ie it is expected that all allocated memory can be correctly & exhaustively freed). It's the standard name for this purpose in the standard library, and increasingly across the ecosystem.

v2/lib/telemetry.zig
InKryption: See King's comments:
* https://github.com/Syndica/sig/pull/1283#discussion_r2962801319
* https://github.com/Syndica/sig/pull/1283#discussion_r2962804158

v2/lib/telemetry.zig
InKryption: Can't do it all in `swapSnapshot` without allocating memory to store the progressions of `cumulative_count`.

Though if we hit a hard error in telemetry that would cause it to exit and make us unable to complete iteration of the snapshot, the consistency of the histogram data ceases to matter anyway.

That said, consistency of the histogram data is still ensured by calling `snapshot_reader.release()`, which in all existing usages is `defer`'d.

v2/lib/telemetry.zig
InKryption: Have no strong feelings about this

v2/services/telemetry.zig
kprotty: reads/writes already block by default. Meant rather to set a deadline for it to unblock. When running in a server, there will be random connections from internet probers which could halt the telemetry service.

v2/services/telemetry.zig
dnut: Just because another codebase uses the name doesn't mean it's a good name. I've noticed its usage in std for quite a while and have been disappointed by the poor naming choice. It's also far from a standard, the name `allocator` is also widely used.

The name `allocator` already communicates the full intended meaning. I see no advantage that `gpa` has over allocator. I only see disadvantages:
- The name `gpa` is esoteric: It's an acronym that requires extra knowledge to understand what it means compared to just understanding what an allocator is.
- The name `gpa` is inaccurate: It implies that any old allocator won't do, it must be a "general purpose" allocator, whatever that implies. But in fact it really doesn't matter what allocator is passed in. It just needs to be something that adheres to the allocator interface. So why not call it "allocator"?

v2/lib/telemetry.zig
dnut: The current approach is fine and I can see that it's going to work. I don't think you need to change this if you think the current approach is better. I just thought it's a bit easier to reason about the code when a single function call handles the transition fully. It's not super important to achieve this.

> Can't do it all in swapSnapshot without allocating memory to store the progressions of cumulative_count.

I don't know what you mean by the progressions. What's wrong with this?
```zig
for (cold_shard.buckets, hot_shard.buckets) |cold_bucket, hot_bucket| {
    _ = hot_bucket.fetchAdd(cold_bucket.load(.monotonic), .monotonic);
}
_ = hot_shard.sum.fetchAdd(cold_shard_sum, .monotonic);
_ = hot_shard.count.fetchAdd(shard_sync.count, .monotonic);
```

I still would expect to need a cleanup process to zero out the cold shard after reading the snapshot. I was mainly just trying to finalize the hot shard all at once so you don't need to include the hot shard at all as part of the SnapshotReader. As it stands now, I found the code confusing -- why would I need the hot shard to read the snapshot? I had to read a lot of surrounding code to make sense of it.

v2/services/telemetry.zig
kprotty: for 1. `gpa: Allocator` already hints its an allocator, It's a standard in other [zig](https://github.com/search?q=repo%3Aghostty-org%2Fghostty%20gpa&type=code) [codebases](https://github.com/search?q=repo%3Atigerbeetle%2Ftigerbeetle%20gpa&type=code), and is shorter to fit in the column limits

for 2. "general purpose" means you can alloc/free without any hidden invariants. Over e.g. `arena: Allocator`. No different to `allocator` in this regard, just shorter.

v2/lib/telemetry.zig
dnut: (not needed in this pr)

v2/services/net.zig
InKryption: 352c92368593581b3bdda2310ae4628aea21c20b

v2/lib/telemetry.zig
InKryption: I ended up implementing the basis of a higher level framework for this: 352c92368593581b3bdda2310ae4628aea21c20b

v2/services/telemetry.zig
dnut: Well if everyone on the team prefers gpa, let's go with that. I remain thoroughly unconvinced, but there are bigger priorities than coming to consensus on every argument behind minor stylistic choices. If we can at least establish a convention that we use consistently, I can live with it.

Anyway, here are my final arguments on the topic. The benefit of saving characters is so negligible in my mind that it holds no weight against using the clearest possible naming. The usage in other zig codebases is also not a benefit. It just means the name is *less* esoteric, but it's still esoteric compared to "allocator," which is still a con.

Let's do a poll and see what the rest of the team thinks. I'd like to avoid spending any more time debating this. We can just go with the majority vote.

v2/lib/telemetry.zig
dnut: nice

v2/config/testnet.zig.zon
dnut: I should have commented more directly on this:

> I believe there's other stuff we have set up that already presumes our prometheus stuff will go over 12345? Care to weigh in @dnut?

This is true but it's trivial to change it if we believe there's a better port number. Also it's built mainly for v1, so there's actually an argument for using a different port number because the metrics will be different.

v2/services/telemetry.zig
dnut: To be clear, I'm not demanding a change here right now. I'm just feeling an inconsistency between this code and some of the design discussions I've been in. So I'd like to gain some clarity/agreement about what service files are supposed to contain before moving on.  @yewman and @Sobeston do you have any opinions on this?

Personally I like the idea of keeping service files lean and embedding the components they run within the lib folder. And also making sure the lib folder itself is also well enough organized to easily distinguish components from shared data structures. But I can see room for various reasonable approaches here. I just want to make sure we're all on the same page about the high-level architecture.

v2/services/telemetry.zig
InKryption: d8b49b50d9da302bd33772a8aaf946f6d4fee817

v2/config/testnet.zig.zon
InKryption: I tried to change it to this, and tried to update the existing machinery to use the new port, but couldn't manage to get it to work

dnut: What existing machinery? There is no existing machinery that needs to change. v2 and v1 can use different port numbers.

InKryption: As in the docker container which configures Prometheus to listen on port 12345.

InKryption: I figured out the problem, for some reason the docker container container didn't restart properly.

v2/services/telemetry.zig
yewman: I think a vote is a good solution 👍

v2/config/testnet.zig.zon
dnut: The docker container is for v1, so it's not relevant. I would leave all of that untouched. v2 can use a different port number.

dnut (dismissed):

v2/services/telemetry.zig
InKryption: bcd5a1cd695918062d9bd83241150b62dfbd562a

dnut (dismissed):

dnut (dismissed):

dnut (approved):

kprotty (approved):

PR #1336, opened 2026-03-20
feat(jrpc-ws): add voteSubscribe method

prestonsn:
## Summary

Wire up `voteSubscribe` end-to-end, connecting the vote listener's gossip vote processing to the existing websocket subscription infrastructure.

Our implementation matches agave's, which sources notifications from both gossip and replay vote transactions, but dedupes both down to produce a unique stream of votes for RPC. Agave's `ClusterInfoVoteListener` merges gossip and replay vote channels into a single processing loop (https://github.com/anza-xyz/agave/blob/727d1043267a9c6bf24d67d2a614721b7a969615/core/src/cluster_info_vote_listener.rs#L594).

Deduplication occurs here, only the first arrival triggers `rpc_subscriptions.notify_vote()` (https://github.com/anza-xyz/agave/blob/master/core/src/cluster_info_vote_listener.rs#L654-L659). 

Gossip votes are chained first in the iterator, so they typically win the race. (https://github.com/anza-xyz/agave/blob/master/rpc/src/rpc_subscriptions.rs#L811). I think the comment there is a bit confusing, it is accurate from the sense that gossip should be propagating unique votes prior to replay seeing them (typically). But not entirely true in edge-case scenarios maybe?  

## Main changes

- emit vote events from `vote_listener.zig` to the RPC `EventSink` when new gossip votes are observed
- add `handleVoteEvent` in `Runtime.zig` to fan out vote notifications to all vote subscribers
- implement `serializeVoteNotification` producing Agave-compatible `voteNotification` JSON-RPC payloads
- register .vote subscription kind across types, sub_map, handler (unsubscribe), and request parsing
- thread event_sink through vote listener Senders and consensus core params
- update sub_map comment to document .vote uses .direct commit policy (no commitment tracking)
- remove `voteSubscribe`/`voteUnsubscribe` from the "not implemented" error path in `ws_request.zig`

## Known behavior difference
Sig emits ~5x more vote notifications than Agave. Agave's cursor-based CRDS polling ([crds.rs](https://github.com/anza-xyz/agave/blob/v3.1.8/gossip/src/crds.rs#L349-L360), [recv_loop](https://github.com/anza-xyz/agave/blob/v3.1.8/core/src/cluster_info_vote_listener.rs#L252-L268)) coalesces intermediate vote replacements between 100ms polls;

Sig's push-based `LocalMessageBroker.publish()` delivers each one individually. All notifications are accurate and unique, our subscribers just get finer-grained data. Matching Agave's rate would require switching to cursor-based polling in the vote pipeline. Not something that's in-scope for RPC compat. We still deliver all the data 

Separate PR up that adds a coalescing approach similar to Agave: https://github.com/Syndica/sig/pull/1350

## Testing

- add integration test: subscribe, inject vote event, verify notification fields (votePubkey, slots, hash, timestamp), unsubscribe
- Tested with the[ solana pub sub client impl](https://docs.rs/solana-pubsub-client/latest/solana_pubsub_client/nonblocking/index.html) on testnet

### DIFF
--- a/src/consensus/vote_listener.zig
+++ b/src/consensus/vote_listener.zig
@@ -21,6 +21,8 @@ const OptimisticConfirmationVerifier =
 const SlotTracker = sig.replay.trackers.SlotTracker;
 const EpochTracker = sig.core.EpochTracker;
 
+const VoteEventData = sig.rpc.jrpc_websockets.types.VoteEventData;
+
 const Logger = sig.trace.Logger("vote_listener");
 
 pub const SlotDataProvider = struct {
@@ -84,6 +86,8 @@ pub const Senders = struct {
     bank_notification: ?*sig.sync.Channel(BankNotification),
     /// TODO: same advisory as on `bank_notification` wrt hooking into RPC.
     subscriptions: RpcSubscriptionsStub,
+    // TODO: consolidate with the above stub.
+    event_sink: ?*sig.rpc.jrpc_websockets.types.EventSink = null,
 
     const test_setup_enabled = @import("builtin").is_test;
 
@@ -439,7 +443,7 @@ fn listenAndConfirmVotes(
     vote_processing_time: ?*VoteProcessingTiming,
     latest_vote_slot_per_validator: *sig.utils.collections.PubkeyMap(Slot),
     metrics: VoteListenerMetrics,
-) std.mem.Allocator.Error![]const ThresholdConfirmedSlot {
+) ![]const ThresholdConfirmedSlot {
     var replay_votes_buffer: std.ArrayListUnmanaged(vote_parser.ParsedVote) = .empty;
     defer replay_votes_buffer.deinit(allocator);
     try replay_votes_buffer.ensureTotalCapacityPrecise(allocator, 4096);
@@ -487,7 +491,7 @@ fn filterAndConfirmWithNewVotes(
     gossip_vote_txs: []const vote_parser.ParsedVote,
     replayed_votes: []const vote_parser.ParsedVote,
     metrics: VoteListenerMetrics,
-) std.mem.Allocator.Error![]const ThresholdConfirmedSlot {
+) ![]const ThresholdConfirmedSlot {
     const root_slot = slot_data_provider.rootSlot();
 
     var diff: SlotsDiff = .EMPTY;
@@ -785,7 +789,7 @@ fn trackNewVotesAndNotifyConfirmations(
     new_optimistic_confirmed_slots: *std.ArrayListUnmanaged(ThresholdConfirmedSlot),
     is_gossip_vote: bool,
     latest_vote_slot_per_validator: *sig.utils.collections.PubkeyMap(Slot),
-) std.mem.Allocator.Error!void {
+) !void {
     if (vote.isEmpty()) return;
     const root = slot_data_provider.rootSlot();
 
@@ -905,7 +909,13 @@ fn trackNewVotesAndNotifyConfirmations(
 
                 // Note gossip votes will always be processed because those should be unique
                 // and we need to update the gossip-only stake in the `VoteTracker`.
-                return;
+                //
+                // We break here instead of returning to preserve is_new_vote from a prior
+                // iteration so that post-loop notification logic still fires.
+                //
+                // This matches agave's control flow:
+                // [agave] https://github.com/anza-xyz/agave/blob/v3.1.8/core/src/cluster_info_vote_listener.rs#L560-L571
+                break;
             }
 
             is_new_vote = is_new;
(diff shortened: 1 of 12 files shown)


src/rpc/jrpc_websockets/Runtime.zig
```diff
@@ -578,6 +581,35 @@ fn handleSlotTransition(
...
+) void {
+    for (self.sub_map.entries.items) |*entry| {
+        if (entry.key.method != .vote) continue;
+
+        // Clone the vote slots for each subscription's serialize job since
+        // each job takes ownership of the data.
+        const cloned_slots = self.allocator.dupe(u64, vote_data.slots) catch |err| {
+            self.logger.err().logf("failed to clone vote slots for subscription: {}", .{err});
+            continue;
+        };
```
jbuckmccready: I don't think it actually needs to allocate to duplicate here since voteSubscribe has no parameters there can only be one multiplexed subscription in the sub map.

You'll need to take ownership of the slots slice similar to what it does with the accounts data in slot frozen event.

Or if you don't want to change the code then just throw in TODO(perf) in here to remark on this, either way it should have a comment for future reference.

src/rpc/jrpc_websockets/sub_map.zig
```diff
@@ -71,9 +71,8 @@ pub const RPCSubMap = struct {
     /// retained. On new entry, heap fields are deep-copied into the map's allocator.
     ///
     /// Queue commit path policy:
-    /// - `.slot`, `.root`, `.program`, `.account`, and `.logs` use `.reserved`
-    ///   to preserve event order
-    /// - all other methods use `.direct`
+    /// - `.slot`, `.root`, `.program`, and `.account` use `.reserved` to preserve event order
+    /// - all other methods (`.logs`, `.vote`, `.signature`) use `.direct`
```
jbuckmccready: `.logs` actually uses `.reserved` - I think we should just get rid of this part of the comment, it's noisy to maintain and has no real value. We can update the switch on the method to be exhaustive and comment there regarding the importance of ordering for the particular methods.

prestonsn: Good catch, and agreed. Also breaks compilation here if future events get added: https://github.com/Syndica/sig/pull/1336/commits/3071181c97330656c4f37c852dad14aac9edde46

src/rpc/jrpc_websockets/Runtime.zig
prestonsn: Yeah this is a good point. Matches the pattern established with the accounts data: https://github.com/Syndica/sig/pull/1336/commits/a3b8a769a8f264e02b850a68a8bb3d0b5473fb0a

src/rpc/jrpc_websockets/sub_map.zig
jbuckmccready: >             // voteSubscribe and signatureSubscribe each multiplex to a single stream, so they
>             // can commit serialized payloads directly.

Reserved vs direct has to do with whether the events should be strictly ordered the same as they arrived over the event sink rather than it being single vs multi subscription streams.

prestonsn: https://github.com/Syndica/sig/pull/1336/commits/91425de7cb4c4700c510b9746559abafb85a56ad

jbuckmccready (dismissed):

src/rpc/jrpc_websockets/types.zig
```diff
@@ -607,6 +623,54 @@ pub const SignatureNotificationData = struct {
     }
 };
 
+pub const VoteEventData = struct {
+    vote_pubkey: Pubkey,
+    slots: []const u64,
+    hash: Hash,
+    timestamp: ?i64,
+    signature: Signature,
+    arena: std.heap.ArenaAllocator,
```
jbuckmccready: It seems weird to use ArenaAllocator here when it's just a slice of u64. I updated the event type to take in allocator for deinit to a support events allocating if needed in slotUpdatesSubscribe: https://github.com/Syndica/sig/pull/1338

Could either update this PR with the change or update VoteEvent when merging the slotUpdates PR.

jbuckmccready: Or if it doesn't quite align then we could just have field for the allocator rather than an arena.

prestonsn: Transplanted the `allocator` changes from the `slotUpdateSubscribe` PR: https://github.com/Syndica/sig/pull/1336/commits/0deb900efd5563c5c45b0ba6432b989d6b56372c

jbuckmccready (dismissed):

src/shred_network/collector/shred_receiver.zig
Sobeston, line 309: Would be nice to assert that all signature items were written to

src/consensus/vote_listener.zig
```diff
@@ -934,7 +944,37 @@ fn trackNewVotesAndNotifyConfirmations(
...
+        if (senders.event_sink) |event_sink| {
+            var vote_event = VoteEventData.initOwned(
+                event_sink.allocator(),
+                vote_pubkey,
+                vote_slots,
+                last_vote_hash,
+                vote.timestamp(),
+                vote_transaction_signature,
+            ) catch return;
+            errdefer vote_event.deinit(event_sink.allocator());
```
Sobeston: The catch return here seems odd - is this to silently ignore OutOfMemory errors, or something else?

src/consensus/vote_listener.zig
```diff
@@ -934,7 +944,37 @@ fn trackNewVotesAndNotifyConfirmations(
...
+
+            try event_sink.send(.{ .vote = vote_event });
+        }
```
Sobeston, line 976: We're allocating with the event sink allocator, sending that event, then it looks like this is cleaned up by handleInboundEvent, which uses self.allocator to free.

If these allocators are different this is illegal, presumably so far they've been the same allocators

Sobeston (changes requested):
Needs a rebase, a few risky bits:

src/shred_network/collector/shred_receiver.zig
prestonsn: agreed, think this should do it: https://github.com/Syndica/sig/pull/1336/commits/0f0c76312a355d6263ee2cae1b8c10156204bc8e

src/consensus/vote_listener.zig
prestonsn: https://github.com/Syndica/sig/pull/1336/commits/754cad4ee0d467bc9fdaf048f323778bcd29053e

Should not have been ignoring the allocator error here. Switched to just propagate the error up and restrict return error set. If this alloc fails here, bigger issues afoot

src/consensus/vote_listener.zig
jbuckmccready: Yeah I think this just didn't get updated to what it should be in previous PRs.

The intent is to have the events all be using the sink allocator (channel's), we need to adjust the receive side to use it as well to ensure no mismatch. This snuck in because originally all events that needed to allocate just had an arena in them so it was tracked by the arena (`deinit` on the event type didn't even take in an allocator).

prestonsn: Good catch, fixed this up: https://github.com/Syndica/sig/pull/1336/commits/b96aed6d53ba23a37e8a8079955b62caeaab3945

Reference to commit with changes we moved into this PR (I didn't copy all the changes properly): https://github.com/Syndica/sig/pull/1338/commits/e2676fc886b5d42438b28c5f5f40ee109567c341

Sobeston (approved):

jbuckmccready (approved):

PR #1339, opened 2026-03-23
feat(shred-network/replay) reworked shred receiver + new block assembly pipeline

Sobeston:
This PR reworks the shred receiver service, and introduces the Replay service. 

The replay service can now build blocks from data given by shred network - this PR doesn't intend to have any transaction execution

Working so far
- Shred validation
- Fec set tracking
- Fec set reconstruction
- Building Merkle Forest of reconstructed fec sets
- Received first complete blocks

Not working yet:
- Eviction from merkle forest (block assembly)
- Repair

Also adds:
- A left-child right-sibling tree
- A mem pool
- Docgen step

Working nicely on testnet:
image


v2/lib/collections/pool.zig
```diff
@@ -0,0 +1,156 @@
+const std = @import("std");
+const collections = @import("../collections.zig");
```
kprotty: theres a util.zig. Want to move it to lib/utils? or is it best to keep collections in their own thing (if so, should ring be moved here?)

v2/lib/collections.zig
```diff
@@ -0,0 +1,30 @@
...
+    _ = std.testing.refAllDecls(@This());
+}
+
+pub const Pool = @import("collections/pool.zig").Pool;
+pub const LCRSTree = @import("collections/lcrs_tree.zig").LCRSTree;
+
+pub fn Id(IdInt: type) type {
+    const expected_idx_ints: []const type = &.{ u8, u16, u32, u64 };
+    if (std.mem.indexOfScalar(type, expected_idx_ints, IdInt) == null)
+        @compileError("Unexpected integer type");
```
kprotty: a `switch (IdInt)` here seems cleaner

v2/lib/collections.zig
```diff
@@ -0,0 +1,30 @@
...
+
+    return enum(IdInt) {
+        null = std.math.maxInt(IdInt),
```
kprotty: Will zero always be the invalid sentinel? It seems specific to pool usage.

v2/lib/shred.zig
```diff
@@ -1,2219 +1,565 @@
...
-        };
+    pub fn eql(a: *const FecSetId, b: *const FecSetId) bool {
+        return (a.slot == b.slot and a.fec_set_idx == b.fec_set_idx);
     }
 
-    pub fn retransmitterSignature(self: Shred) !Signature {
-        return switch (self) {
-            inline .data, .code => |s| s.retransmitterSignature(),
-        };
+    pub fn compare(a: *const FecSetId, b: *const FecSetId) std.math.Order {
```
kprotty: use stdlib convention of `order` here? can also switch similar calls in Pubkey to compare instead

v2/lib/shred.zig
```diff
@@ -1,2219 +1,565 @@
...
 
-        fn chainedMerkleRoot(self: Self) !Hash {
-            return layout.getChainedMerkleRoot(self.payload) orelse error.InvalidPayloadSize;
+        /// returns a code variant as a data variant (or vice versa), preserving its fields
+        pub fn swapType(self: Variant) Variant {
+            // swaps bit 4 and 5, swaps bit 6 and 7
+            return .{
+                .inner = ((self.inner & 0x50) << 1) |
+                    ((self.inner & 0xA0) >> 1) |
+                    (self.inner & 0x0F),
```
kprotty: someone optimized this

v2/lib/shred.zig
```diff
@@ -1,2219 +1,565 @@
...
-    if (offset + 1 == tree.len) {
-        return .{
-            .bytes = try allocator.realloc(proof, write_cursor),
-            .len = write_cursor / merkle_proof_entry_size,
+            break :sizes .{
+                effective_size - header_size - payload_size - trailer_size,
+                payload_size,
+            };
+        } else sizes: {
+            const zero_padding_size = 0;
```
kprotty: is this intentional?

v2/services/replay.zig
```diff
@@ -0,0 +1,1103 @@
...
+    /// slots.
+    /// For example, accountsdb may want to hold onto data for blocks for a while longer before it
+    /// flushes to disk.
+    /// Such subsystems may want to hold onto their own root?
+    consensus_rooted_block: BlockId,
+
+    // // NOTE: It would probably better to traverse the tree from a given root instead of using a map,
+    // // as it would be easier to make lock-free. Currently if made multithreaded, value mutations
+    // // could race with map inclusion.
+    // last_merkle_map: LastMerkleMap,
```
kprotty: worth removing since already commented out?

v2/services/replay.zig
```diff
@@ -0,0 +1,1103 @@
...
+    // chained-merkle-hash -> node
+    orphan_map: OrphanMap,
+
+    const capacity = 2048;
+
+    // keyed by merkle root
+    const OrphanMap = std.ArrayHashMapUnmanaged(void, *MerkleNode, OrphanContext, true);
+
+    // keyed by chained merkle root
+    const MerkleMap = std.ArrayHashMapUnmanaged(void, *MerkleNode, MerkleContext, true);
```
kprotty, line 360: we should prob eventually migrate to fixed-sized hashmaps & pinned-init them (instead of using the scratch_mem + fba trick)

v2/services/replay.zig
```diff
@@ -0,0 +1,1103 @@
...
+                        break;
+                    } else {
+                        elem |= 0x80;
+                        writer.writeByte(elem) catch return error.EncodeFailed;
+                        byte_count += 1;
+                    }
+                }
+            }
+            return .{ .value_count = max_count, .byte_count = byte_count };
+        }
```
kprotty: This looks like `lib.gossip.bincode.VarInt`, used by `lib.gossip.bincode.ShortVec`. Worth moving bincode.zig out of `lib/gossip` and into `lib/solana`

v2/services/replay.zig
```diff
@@ -0,0 +1,1103 @@
...
+    const CompiledInstruction = struct {
+        program_id_index: u8,
+        accounts: []u8,
+        data: []u8,
+
+        const bk_config: bk.Codec(CompiledInstruction) = .standard(.tuple(.{
+            .program_id_index = .fixint,
+            .accounts = .from(shortVec(u8, bk.StdCodec(u8).fixint.codec)),
+            .data = .from(shortVec(u8, bk.StdCodec(u8).fixint.codec)),
+        }));
```
kprotty: These in lib.gossip (Vote struct). Maybe worth making a lib/solana/transaction.zig with Vote + AddressLookup + various Message types in there.

kprotty: These in lib.gossip (Vote struct)

v2/services/shred_receiver.zig
```diff
@@ -23,208 +67,71 @@ pub const panic = start.panic;
...
-
-        verifyShred(packet, &ro.config.leader_schedule, &verified_roots) catch |err| {
-            std.log.debug("failed to verify shred: {}", .{err});
+    const idle_src = @src();
+    var maybe_idle_zone: ?tracy.Zone = tracy.Zone.init(idle_src, .{ .name = "idle" });
+    while (true) {
+        const packet = packet_iter.next() orelse {
+            if (maybe_idle_zone == null) {
+                maybe_idle_zone = tracy.Zone.init(idle_src, .{ .name = "idle" });
+            }
```
kprotty: what about moving the loop inside + zone outside:
```zig
while (true) {
  {
    const idle_zone = tracy.Zone.init();
    defer idle_zone.deinit();
    while (packet_iter.peek() == null) continue;
  }
  while (packet_iter.next()) |p| {
    const recv_zone = tracy.Zone.init();
    defer recv_zone.deinit();
    handle(p);
  }
}
```

v2/build.zig
```diff
@@ -127,5 +142,55 @@ pub fn build(b: *Build) !void {
...
+                .root_module = b.createModule(.{
+                    .target = target,
+                    .optimize = .Debug,
+                    .root_source_file = b.path("scripts/gen_docs_entry.zig"),
+                }),
+                .use_llvm = false,
+            }),
+        );
+
+        inline for (&.{ doc_service_modules, doc_service_modules }) |module_list| {
```
kprotty: why are there two instances of same module_list?

v2/lib/collections/lcrs_tree.zig
```diff
@@ -0,0 +1,47 @@
...
+    const needed_fields: []const []const u8 = &.{ "parent", "child", "sibling" };
+
+    const NodeId = Id(IdInt);
+
+    for (needed_fields) |field| {
+        if (!@hasField(Node, field)) {
+            @compileLog("missing field", Node, field);
+            continue;
+        }
+        if (@TypeOf(@field(@as(Node, undefined), field)) != NodeId)
```
InKryption:
```suggestion
        if (@FieldType(Node, field) != NodeId)
```

v2/lib/collections.zig
```diff
@@ -0,0 +1,30 @@
+const std = @import("std");
+
+comptime {
+    _ = std.testing.refAllDecls(@This());
+}
+
+pub const Pool = @import("collections/pool.zig").Pool;
+pub const LCRSTree = @import("collections/lcrs_tree.zig").LCRSTree;
+
+pub fn Id(IdInt: type) type {
```
InKryption: I don't really see the point of generalizing this. If there were two pieces of code both using `Id(u32)` for something, they wouldn't be unique types, kind of defeating the point of a type-safe int wrapper.

v2/lib/shred/reed_solomon_table.zig
```diff
@@ -0,0 +1,260 @@
+pub const log: [256]u8 = .{ 0, 0, 1, 25, 2, 50, 26, 198, 3, 223, 51, 238, 27, 104, 199, 75, 4, 100, 224, 14, 52, 141, 239, 129, 28, 193, 105, 248, 200, 8, 76, 113, 5, 138, 101, 47, 225, 36, 15, 33, 53, 147, 142, 218, 240, 18, 130, 69, 29, 181, 194, 125, 106, 39, 249, 185, 201, 154, 9, 120, 77, 228, 114, 166, 6, 191, 139, 98, 102, 221, 48, 253, 226, 152, 37, 179, 16, 145, 34, 136, 54, 208, 148, 206, 143, 150, 219, 189, 241, 210, 19, 92, 131, 56, 70, 64, 30, 66, 182, 163, 195, 72, 126, 110, 107, 58, 40, 84, 250, 133, 186, 61, 202, 94, 155, 159, 10, 21, 121, 43, 78, 212, 229, 172, 115, 243, 167, 87, 7, 112, 192, 247, 140, 128, 99, 13, 103, 74, 222, 237, 49, 197, 254, 24, 227, 165, 153, 119, 38, 184, 180, 124, 17, 68, 146, 217, 35, 32, 137, 46, 55, 63, 209, 91, 149, 188, 207, 205, 144, 135, 151, 178, 220, 252, 190, 97, 242, 86, 211, 171, 20, 42, 93, 158, 132, 60, 57, 83, 71, 109, 65, 162, 31, 45, 67, 216, 183, 123, 164, 118, 196, 23, 73, 236, 127, 12, 111, 246, 108, 161, 59, 82, 41, 157, 85, 170, 251, 96, 134, 177, 187, 204, 62, 90, 203, 89, 95, 176, 156, 169, 160, 81, 11, 245, 22, 235, 122, 117, 44, 215, 79, 174, 213, 233, 230, 231, 173, 232, 116, 214, 244, 234, 168, 80, 88, 175 };
```
InKryption, line 1: Seems like a good case for a zon file?

v2/lib/collections/pool.zig
```diff
@@ -0,0 +1,156 @@
...
+        pub fn indexToPtr(self: *const Self, item_id: ItemId) *Item {
+            std.debug.assert(item_id != .null);
+            return @ptrCast(&self.buf[item_id.index().?]);
+        }
+
+        pub fn ptrToIndex(self: *const Self, item: *Item) ItemId {
+            const node: [*]const Node = @ptrCast(item);
+            const base: [*]const Node = self.buf;
+            return @enumFromInt(node - base);
+        }
```
InKryption: Maybe worth it to assert it's in bounds:
```suggestion
        pub fn ptrToIndex(self: *const Self, item: *Item) ItemId {
            std.debug.assert(@ptrFromInt(item) >= self.buf);
            std.debug.assert(@ptrFromInt(item) < self.buf + self.len);
            const node: [*]const Node = @ptrCast(item);
            const base: [*]const Node = self.buf;
            return @enumFromInt(node - base);
        }
```

v2/lib/shred/receiver.zig
```diff
@@ -0,0 +1,751 @@
...
+
+        var done: DoneSets = try .init(allocator, done_capacity);
+        errdefer done.deinit(allocator);
+
+        return .{
+            .in_progress = in_progress,
+            .done = done,
+
+            .root_slot = 0,
+            .max_slot = std.math.maxInt(Slot),
```
InKryption, line 46: Is there a reason to init this to maxint instead of an actual expected max slot?

v2/lib/shred/receiver.zig
```diff
@@ -0,0 +1,751 @@
...
+
+        tracy.frameMarkNamed("finished FEC sets");
+
+        return .fec_set_finished;
+    }
+
+    pub const NonErrorStatus = union(enum) {
+        unfinished_fec_set: struct {
+            // 0..=31 (if it had 32, it would be finished)
+            total_shreds_received: std.math.IntFittingRange(0, FecSetCtx.fec_shred_cnt - 1),
```
InKryption: Waiter, one ranged int pls

v2/lib/shred/receiver.zig
```diff
@@ -0,0 +1,751 @@
...
+    merkle_root: Hash,
+
+    // https://github.com/firedancer-io/firedancer/blob/ecd2d6d8f5b9f926d0b9aa9360efe36ea1550ad6/src/ballet/reedsol/fd_reedsol.h#L23
+    // https://github.com/solana-foundation/specs/blob/main/p2p/shred.md
+
+    // There's now a max of 32+32 shreds
+    // https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0317-enforce-32-data-shreds.md
+    const data_shreds_max = 32;
+    const code_shreds_max = 32;
+    pub const fec_shred_cnt = 32;
```
InKryption: Could we just call this `count` lol

v2/lib/shred/receiver.zig
```diff
@@ -0,0 +1,751 @@
...
+        std.debug.assert(data_recv <= data_shreds_max);
+        const code_recv: u8 = @intCast(self.code_shreds_received.count());
+        std.debug.assert(code_recv <= code_shreds_max);
+
+        return data_recv + code_recv;
+    }
+};
+
+fn hashSignature(a: *const Signature) u32 {
+    return @bitCast((a.r[0..2] ++ a.s[0..2]).*);
```
InKryption, line 372: I'm surprised you can concatenate runtime-known array pointers

v2/lib/shred/receiver.zig
```diff
@@ -0,0 +1,751 @@
...
+    // https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0317-enforce-32-data-shreds.md
+    const data_shreds_max = 32;
+    const code_shreds_max = 32;
+    pub const fec_shred_cnt = 32;
+
+    fn totalShredsReceived(self: *const FecSetCtx) u8 {
+        const data_recv: u8 = @intCast(self.data_shreds_received.count());
+        std.debug.assert(data_recv <= data_shreds_max);
+        const code_recv: u8 = @intCast(self.code_shreds_received.count());
+        std.debug.assert(code_recv <= code_shreds_max);
```
InKryption, line 365: Annoying that stdlib's staticbitset's `count` just returns an unconstrained usize.

v2/lib/shred/receiver.zig
```diff
@@ -0,0 +1,751 @@
...
+
+    const Eviction = std.PriorityQueue(Pool.ItemId, QueueContext, QueueContext.compare);
+    const Pool = lib.collections.Pool(FecSetCtx, u32);
+    const Queue = std.PriorityQueue(Pool.ItemId, QueueContext, QueueContext.compare);
+
+    // Key from signature-hash, rather than merkle root, as it's equivalent for lookup and we don't
+    // have to compute it.
+    const SignatureMap = std.ArrayHashMapUnmanaged(void, *FecSetCtx, SignatureContext, true);
+
+    fn init(allocator: std.mem.Allocator, capacity: u32) !InProgressSets {
```
InKryption, line 392: Might be worth documenting that the capacity is for the total number of fec sets & also the queues & pool - the fact that those are all one in the same seems a bit arbitrary to me, but lmk if there's actually some higher level reasoning I just don't see.

v2/lib/shred/receiver.zig
```diff
@@ -0,0 +1,751 @@
...
+        return for (self.signature_map.values()) |fec_set_ctx| {
+            const pool_id = self.ctx_pool.ptrToIndex(fec_set_ctx);
+            const idx = pool_id.index().?;
+
+            if (self.ids[idx].eql(&id)) break true;
+        } else false;
+    }
+
+    fn assertCounts(self: *const InProgressSets) void {
+        std.debug.assert(self.signature_map.count() == self.eviction.items.len);
```
InKryption, line 527: I don't fully understand this invariant. Why is the eviction count always equal to the number of fec sets/signatures?

v2/lib/shred/receiver.zig
```diff
@@ -0,0 +1,751 @@
...
+    signature_map: SignatureMap,
+    eviction: Eviction,
+
+    const Eviction = std.PriorityQueue(Pool.ItemId, QueueContext, QueueContext.compare);
+    const Pool = lib.collections.Pool(FecSetCtx, u32);
+    const Queue = std.PriorityQueue(Pool.ItemId, QueueContext, QueueContext.compare);
+
+    // Key from signature-hash, rather than merkle root, as it's equivalent for lookup and we don't
+    // have to compute it.
+    const SignatureMap = std.ArrayHashMapUnmanaged(void, *FecSetCtx, SignatureContext, true);
```
InKryption, line 390: I love adapter patterns.

v2/lib/shred/reed_solomon.zig
```diff
@@ -0,0 +1,327 @@
...
+    for (0..data_count) |r| {
+        sub_matrix[r] = encoding_matrix[valid_indices[r]];
+    }
+
+    // Invert sub_matrix via Gaussian elimination on augmented [sub_matrix | identity]
+    var aug: [data_count][data_count * 2]u8 = undefined;
+    for (0..data_count) |r| {
+        for (0..data_count) |c| {
+            aug[r][c] = sub_matrix[r][c];
+            aug[r][data_count + c] = if (r == c) @as(u8, 1) else @as(u8, 0);
```
InKryption: Are these `@as`s necessary? Thought the result location type was propagated through branches now.

v2/lib/shred/reed_solomon.zig
```diff
@@ -0,0 +1,327 @@
...
+    }
+
+    // Forward elimination
+    for (0..data_count) |r| {
+        if (aug[r][r] == 0) {
+            for (r + 1..data_count) |r_below| {
+                if (aug[r_below][r] != 0) {
+                    const tmp = aug[r];
+                    aug[r] = aug[r_below];
+                    aug[r_below] = tmp;
```
InKryption: Porque no
```suggestion
                    std.mem.swap([data_count * 2]u8, &aug[r], &aug[r_below]);
```

v2/lib/shred/reed_solomon.zig
```diff
@@ -0,0 +1,327 @@
...
+            if (fec_set_ctx.code_shreds_received.isSet(i)) {
+                @memcpy(&leader_sig, fec_set_ctx.code_shreds_buf[i][0..Signature.SIZE]);
+                have_sig = true;
+                break;
+            }
+        }
+    }
+
+    // Collect pointers to erasure shards for the 32 valid indices
+    var shard_ptrs: [data_count][*]const u8 = undefined;
```
InKryption: Any particular reason to store pointers instead of slices here? Would be nice to know if there's any bug in this that causes an OOB access.

v2/lib/shred/reed_solomon.zig
```diff
@@ -0,0 +1,327 @@
...
+        }
+
+        // For data shreds, erasure shard starts at offset 64 (after signature)
+        // and ends at headers_size + capacity. We compute it the same way.
+        // The erasure shard for data shreds covers bytes [64 .. 64 + shard_len]
+        const dest_start = Signature.SIZE; // 64
+        const dest_end = dest_start + shard_len;
+        if (dest_end > Packet.capacity) return;
+
+        var dest = dest_packet[dest_start..dest_end];
```
InKryption:
```suggestion
        const dest = dest_packet[dest_start..dest_end];
```

v2/lib/shred.zig
```diff
@@ -1,2219 +1,565 @@
...
-            _: bk.Config,
-            value_count: usize,
-            decoded_count: *usize,
-            _: void,
-        ) bk.DecodeSkipError!void {
-            try reader.discardAll(value_count);
-            decoded_count.* = value_count;
-        }
+        const merkle_proof_ptr: [*]const MerkleProofNode =
+            @ptrCast(buffer[0..].ptr + merkle_offset);
```
InKryption:
```suggestion
            @ptrCast(buffer.ptr + merkle_offset);
```

v2/lib/shred.zig
```diff
@@ -1,2219 +1,565 @@
...
-            self.resigned == true)
-                0xb0
-            else
-                return error.UnknownShredVariant;
-        return big_end | self.proof_size;
-    }
+    // The payload of a data shred. Asserts shred is a data shred.
+    pub fn dataPayload(shred: *const Shred) []const u8 {
+        std.debug.assert(shred.variant.isData());
+        const buffer: *const Packet.Buffer = @ptrCast(@alignCast(shred));
```
InKryption: is the aligncast here necessary?

v2/lib/shred.zig
```diff
@@ -1,2219 +1,565 @@
...
+        const merkle_protected_size = if (shred.variant.isData())
+            data_merkle_protected_size
         else
-            packet.size;
-    }
+            code_merkle_protected_size;
 
-    pub fn getSlot(shred: []const u8) ?Slot {
-        return getInt(Slot, shred, OFFSET_OF_SHRED_SLOT);
+        return @as(*const Packet.Buffer, @ptrCast(@alignCast(shred)))[Signature.SIZE..][0..merkle_protected_size];
```
InKryption: Here too, is this aligncast necessary?

v2/lib/shred.zig
```diff
@@ -1,2219 +1,565 @@
...
-        const start = index * merkle_proof_entry_size;
-        const end = start + merkle_proof_entry_size;
-        var entry: MerkleProofEntry = undefined;
-        @memcpy(&entry, self.bytes[start..end]);
-        return entry;
-    }
+    // This is combined with fragments from other shreds in the erasure set to
+    // reconstruct a collection of entries.
+    pub fn erasureFragment(shred: *const Shred) ?[]const u8 {
+        const buffer: *const Packet.Buffer = @ptrCast(@alignCast(shred));
```
InKryption: Is this aligncast necessary?

v2/lib/shred.zig
```diff
@@ -1,2219 +1,565 @@
...
-                const byte = v.toByte() catch return error.EncodeFailed;
-                try writer.writeByte(byte);
-            }
-            return .{
-                .value_count = max_count,
-                .byte_count = max_count,
-            };
-        }
+    fn merkleProofNodes(shred: *const Shred) []const MerkleProofNode {
+        const buffer: *const Packet.Buffer = @ptrCast(@alignCast(shred));
```
InKryption: Ditto on the aligncast

v2/scripts/gen_docs_entry.zig
```diff
@@ -0,0 +1,79 @@
...
+
+    const service_names_str = args[1];
+    const module_names_str = args[2];
+    const output_file_path = args[3];
+
+    const file = try std.fs.cwd().createFile(output_file_path, .{});
+    var buf: [4096]u8 = undefined;
+    var writer_state = file.writer(&buf);
+    const writer = &writer_state.interface;
+    defer writer.flush() catch @panic("failed to flush");
```
InKryption: If we fail for some other reason before flushing, I don't think there's any point in obfuscating the original error with this panic
```suggestion
    defer writer.flush() catch {};
```

v2/services/shred_receiver.zig
```diff
@@ -1,18 +1,62 @@
...
+//!       interleaved we would drop the 2nd FEC set. In this case we may have to get shreds from
+//!       the missing set from repair. I expect this to be rare, and this behaviour matches
+//!       Firedancer's.
+//!
+//!       Once we complete a FEC set, we are free to build another FEC set of that very same
+//!       FecSetId; the downstream service must be fully aware of equivocation.
+//!
+//!       Once repair is implemented, shreds from it should bypass these checks and enter the map,
+//!       even if there is already an instance of that FecSetId.
+//!
```
InKryption, line 48: Note: top level doc comments like this are gonna be removed, I think it would be good if we preemptively started using the new endorsed pattern, which is this:
```zig
/// The doc comment
const filename_or_something = @This();
```

v2/services/replay.zig
```diff
@@ -0,0 +1,1103 @@
...
+                }
+            }
+            decoded_count.* = value_count;
+        }
+
+        pub const free = null;
+    });
+
+    // For slices that have a compact_u16 length rather than a u64
+    fn shortVec(comptime Element: type, comptime element: bk.Codec(Element)) bk.Codec([]Element) {
```
InKryption: Fwiw in latest binkode version there's now `sliceWithLen` that effectively does this. Could all be replaced with `.standard(.sliceWithLen(u16, .varint, element))`.

v2/services/replay.zig
```diff
@@ -0,0 +1,1103 @@
...
+    try std.testing.expectEqual(.waiting_for_child, a_result);
+    try std.testing.expectEqual(.waiting_for_parent_and_child, c_result);
+    try std.testing.expectEqual(.chain_incomplete, b_result);
+    try std.testing.expectEqual(.chain_complete, d_result);
+}
+
+const bincode = struct {
+    // This isn't quite a bincode fixed int, nor a varint. It's some custom Solana thing used in
+    // Agave's `short_vec`. I think it's supposed to be smaller than a fixed or varint for a u16.
+    const compact_u16: bk.Codec(u16) = .implement(void, void, struct {
```
InKryption: Isn't this just `.standard(.varint)`?

InKryption (changes requested):

v2/lib/collections/pool.zig
Sobeston: I'm not picky. Maybe IPC contains IPC-compatible collections like rings, and this one contains the others?

v2/lib/collections.zig
Sobeston: > Will zero always be the invalid sentinel

MaxInt, not 0. I haven't found a good reason to use anything else, but this could be made configurable later

v2/services/replay.zig
Sobeston: I'm pretty sure this is a bit different to bincode's varint

v2/services/replay.zig
Sobeston: This uses the weird compact_u16, which afaik isn't a bincode varint, edit: see below comment

v2/services/shred_receiver.zig
Sobeston: Interesting, can I get a link for that? Happy to do it

v2/lib/shred.zig
Sobeston: f9efe4e287b03ea1de7182d524adcfde40f2f46b

v2/lib/shred.zig
Sobeston: Initially yes, but ended up removing it in 521f9ac7686bc61d8853a8851357f6fadb6ac06b

v2/services/replay.zig
Sobeston: Removed in 521f9ac7686bc61d8853a8851357f6fadb6ac06b

v2/services/shred_receiver.zig
Sobeston: 521f9ac7686bc61d8853a8851357f6fadb6ac06b

v2/build.zig
Sobeston: Fixed in [469bac9217b32e9be7aa15f3017a2d9bcd922034](https://github.com/Syndica/sig/blob/469bac9217b32e9be7aa15f3017a2d9bcd922034/v2/build.zig)

v2/lib/shred/receiver.zig
Sobeston: The reasoning is that at startup we don't necessarily know what to expect, or at least it isn't the shred receiver's decision.

Once something notifies us of the expected slot range (presumably rooted slot, and rooted slot + some heuristic), we then filter on it. This could be done first by the snapshot loader, and later by consensus, or something like that

v2/lib/shred/receiver.zig
Sobeston: u5 is actually perfect here, which is what is being used, I just wanted to make it extra explicit

v2/lib/shred/receiver.zig
Sobeston: Yep

v2/lib/shred/receiver.zig
Sobeston: They are all one and the same, i.e.

The pool is where the contexts are stored, only items inside the pool are ever in the eviction queue

v2/lib/shred/receiver.zig
Sobeston: Every fec set inside the pool can be found by the signature map + every fec set inside the pool may be evicted (and is therefore in the eviction queue).

Any mismatch is indicative of an eviction/insertion bug

v2/lib/shred/receiver.zig
Sobeston: Yeah I'm using them very aggressively for more intrusive data structures.

e.g. a single MerkleNode in the pool may be some combination of these at any time:
- a node in the merkle map
- a node in the orphan map
- a node in a linked list (pool's freelist)
- a node in a tree

I could maybe go for ArrayHashMapUnmanaged(void, void) in the MerkleForest but it feels less clear?

v2/lib/shred.zig
Sobeston: all align-cast free now b246c47b1d686fa459c5f8cbcfe54709e71b2490

v2/lib/shred.zig
Sobeston: all align-cast free now b246c47b1d686fa459c5f8cbcfe54709e71b2490

v2/lib/shred.zig
Sobeston: all align-cast free now b246c47b1d686fa459c5f8cbcfe54709e71b2490

v2/lib/shred.zig
Sobeston: all align-cast free now b246c47b1d686fa459c5f8cbcfe54709e71b2490

v2/lib/shred.zig
Sobeston: b246c47b1d686fa459c5f8cbcfe54709e71b2490

v2/lib/collections/pool.zig
Sobeston: Added in b246c47b1d686fa459c5f8cbcfe54709e71b2490

v2/lib/collections.zig
Sobeston: This is true, how about if the Id came from the Pool generic, and we used that in the tree?
Would make it clearer that the point is that it's a pool index + would make the type unique per pool type

v2/lib/collections/lcrs_tree.zig
Sobeston: b246c47b1d686fa459c5f8cbcfe54709e71b2490

v2/lib/collections.zig
Sobeston: 521f9ac7686bc61d8853a8851357f6fadb6ac06b

v2/lib/shred/reed_solomon.zig
Sobeston: 7c8e629977ccf4c2aac2ae684395994221c41bfb

v2/lib/shred/reed_solomon.zig
Sobeston: 7c8e629977ccf4c2aac2ae684395994221c41bfb

v2/lib/shred/reed_solomon.zig
Sobeston: 7c8e629977ccf4c2aac2ae684395994221c41bfb

v2/lib/shred/reed_solomon.zig
Sobeston: switched to intFromBool 7c8e629977ccf4c2aac2ae684395994221c41bfb

v2/services/replay.zig
Sobeston: Okay, the difference is that it's enforced to be a u16, can we do that with binkode?

v2/lib/shred/receiver.zig
Sobeston: `fec_shred_count`? Or do you mean unifying the 3 constants

v2/lib/shred/reed_solomon_table.zig
Sobeston: If I make this into a zon file, then everything becomes a tuple rather than an array, right? This doesn't compile

Sobeston:
Let's maybe sync on the bincode stuff

v2/scripts/gen_docs_entry.zig
Sobeston: removed aed381de5ec9b3828fb516df03d4bbf76737a625

v2/services/shred_receiver.zig
InKryption: Apparently I misremembered; the proposal is not actually accepted yet: https://codeberg.org/ziglang/zig/issues/30132

v2/services/replay.zig
InKryption: ~~Not sure what you mean by "enforced to be a u16"?~~ Ah, I think I see, it is not allowed to be bigger than a u16. Wouldn't that still kind of logically follow from both encoding and decoding though? If a varint > 2^16-1 is decoded, it will overflow (I imagine that it would do much the same for compact_u16). And when you encode it, it would obviously not exceed the range of the u16 value.

v2/lib/shred/receiver.zig
InKryption: Yeah, `fec_shred_count`

v2/lib/shred/reed_solomon_table.zig
InKryption: I suppose that's true; wouldn't be much of a problem if we had that `name: [_]T = ...` syntax being pushed by mlugg, but I guess for now it's fine

v2/lib/collections.zig
InKryption: Yeah, I would prefer that.
Note: make sure to do `comptime { _ = PoolSelf; }` inside its definition to make sure it is unique per pool type, otherwise it'll only be generic over the int still if you just plop it in there as is.

v2/services/replay.zig
InKryption: Well, even if it is the case that varint can't be used for this compact_u16 codec, can still use `.standard(.sliceWithLen(u16, compact_u16, element))`.

v2/lib/collections.zig
Sobeston: https://github.com/Syndica/sig/pull/1339/commits/a12be0fe8d3290e6aed3e1dc2e24491ef2c89edf

v2/services/replay.zig
Sobeston: https://docs.rs/solana-short-vec/latest/solana_short_vec/

Wait, no, it actually is different to bincode. It just so happens that values 0 <= 127 are represented the same. 

*bincode::config::VarintEncoding*
> Encoding an unsigned integer v (of any type excepting u8) works as follows:

    If u < 251, encode it as a single byte with that value.
    If 251 <= u < 2**16, encode it as a literal byte 251, followed by a u16 with value u.
    If 2**16 <= u < 2**32, encode it as a literal byte 252, followed by a u32 with value u.
    If 2**32 <= u < 2**64, encode it as a literal byte 253, followed by a u64 with value u.
    If 2**64 <= u < 2**128, encode it as a literal byte 254, followed by a u128 with value u.

*short_vec*
```rs
        assert_good_deserialized_value(0x0000, &[0x00]);
        assert_good_deserialized_value(0x007f, &[0x7f]);
        assert_good_deserialized_value(0x0080, &[0x80, 0x01]);
        assert_good_deserialized_value(0x00ff, &[0xff, 0x01]);
        assert_good_deserialized_value(0x0100, &[0x80, 0x02]);
        assert_good_deserialized_value(0x07ff, &[0xff, 0x0f]);
        assert_good_deserialized_value(0x3fff, &[0xff, 0x7f]);
        assert_good_deserialized_value(0x4000, &[0x80, 0x80, 0x01]);
        assert_good_deserialized_value(0xffff, &[0xff, 0xff, 0x03]);
```
cc @kprotty

v2/lib/collections.zig
InKryption: Nit: `IdSelf` seems superfluous here since you could just refer to `ItemId` directly, but not a huge thing, resolve this at will whether you choose to change that or not.

v2/services/replay.zig
InKryption: Ah that's good to know; so it shall stay.

v2/services/replay.zig
Sobeston: Works nicely https://github.com/Syndica/sig/pull/1339/commits/3e22ce4baba4a47e2271de95b186359ac9b961fe

v2/lib/collections.zig
Sobeston: removed 48a10d244fa77baad187a781000f8493874db7a4

v2/lib/shred/receiver.zig
Sobeston: 48a10d244fa77baad187a781000f8493874db7a4

v2/services/replay.zig
Sobeston: Moved the stuff to solana.bincode 3e22ce4baba4a47e2271de95b186359ac9b961fe

Sobeston: went with lib.solana.bincode for now 3e22ce4baba4a47e2271de95b186359ac9b961fe

v2/services/replay.zig
Sobeston: Turns out ShortVec/short_u16 definitely isn't a bincode varint

InKryption (dismissed):

v2/lib/shred.zig
```diff
@@ -1,2219 +1,560 @@
...
+    /// to determine which kind of shred we have.
+    /// It would be great to represent this as a packed struct, but Agave decided that the legacy
+    /// Code and Data shreds should be defined like this:
+    ///
+    /// pub enum ShredType {
+    ///     Data = 0b1010_0101,
+    ///     Code = 0b0101_1010,
+    /// }
+    pub const Variant = extern struct {
+        inner: u8,
```
dnut: Go ahead and use a packed struct if legacy shreds are the only thing in its way.

Legacy shreds stopped being used years ago. We have never supported them at all in v1. They are also no longer supported by Agave. They deleted the code for legacy shreds. We don't need any code to consider legacy shreds whatsoever. We'd just be creating technical debt that puts us behind the other validator client implementations.

v2/lib/shred.zig
```diff
@@ -1,2219 +1,560 @@
...
-        /// agave: retransmitter_signature
-        fn retransmitterSignature(self: Self) !Signature {
-            const offset = try retransmitterSignatureOffset(self.common.variant);
-            const end = offset + Signature.SIZE;
-            if (self.payload.len < end) return error.InvalidPayloadSize;
-            return .fromBytes(self.payload[offset..][0..64].*);
-        }
+        // upper 4 bits
+        const legacy_data = 0xA0; // deprecated
+        const legacy_code = 0x50; // deprecated
```
dnut:
```suggestion
```

v2/lib/shred.zig
```diff
@@ -1,2219 +1,560 @@
...
+        const merkle_data = 0x80; // deprecated by SIMD-0313
+        const merkle_code = 0x40; // deprecated by SIMD-0313
```
dnut: I'm not as sure about this as I am about legacy shreds, but I think the same idea probably applies.


```suggestion
```

dnut (changes requested):

v2/lib/shred.zig
```diff
@@ -1,2219 +1,560 @@
...
-/// agave: retransmitter_signature_offset and get_retransmitter_signature_offset
-fn retransmitterSignatureOffset(variant: ShredVariant) !usize {
-    if (!variant.resigned) {
-        return error.InvalidShredVariant;
-    }
-    return try proofOffset(variant.constants(), variant) +
-        variant.proof_size * merkle_proof_entry_size;
-}
+            // TODO: support the upcoming feature shredXP8xLjJWp1AWh3gAFsFn4GSH1vohhCMDHw5koU, which
+            // drops data shreds with data_complete=true that aren't the last data shred in the set.
```
dnut, line 235: if shred receiver needs feature activations, that means it also needs to be fork-aware. this is a bit inconvenient

v2/lib/shred/receiver.zig
```diff
@@ -0,0 +1,750 @@
...
+    }
+
+    // TODO: report return values to observability
+    // TODO: report back equivocating shreds, so that we can construct and send out duplicate proofs
+    pub fn processPacket(
+        state: *Receiver,
+        leader_schedule: *const lib.solana.LeaderSchedule,
+        network_shred_version: u16,
+        packet: *const Packet,
+        deshred_writer: *DeshredRing.Iterator(.writer),
```
dnut, line 69: What do you think about using the effects interface pattern, like king implemented for gossip?

Alternatively, if this is the only effect, this could accept `*DeshreddedFecSet`, and the caller could manage its allocation.

v2/lib/shred/receiver.zig
```diff
@@ -0,0 +1,750 @@
...
+    // we store the first seen, and make sure later shreds have the same one
+    merkle_root: Hash,
+
+    // https://github.com/firedancer-io/firedancer/blob/ecd2d6d8f5b9f926d0b9aa9360efe36ea1550ad6/src/ballet/reedsol/fd_reedsol.h#L23
+    // https://github.com/solana-foundation/specs/blob/main/p2p/shred.md
+
+    // There's now a max of 32+32 shreds
+    // https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0317-enforce-32-data-shreds.md
+    const data_shreds_max = 32;
+    const code_shreds_max = 32;
```
dnut, line 358: 32+32 was always the max. What changed with this simd is that now 32+32 is the only valid size. I would remove the word "max" from these names and the comment can indicate that every set is exactly 32+32

dnut (commented):
I like what I see so far. This is better than v1 in a lot of ways.

The data structures feel a bit like a puzzle. You need to look at several of them to see how they all piece together. But I see the reasoning behind it.

I have a few comments. I'll keep reviewing throughout the day and may submit another review later.

v2/lib/solana/bincode.zig
```diff
@@ -0,0 +1,466 @@
...
+        .instructions = .from(shortVec(CompiledInstruction, CompiledInstruction.bk_config)),
+        .address_table_lookups = .from(shortVec(AddressLookup, AddressLookup.bk_config)),
+    }));
+};
+
+const VersionedMessage = union(enum) {
+    // first byte & 0x80 == 0
+    legacy: LegacyMessage,
+    // first byte & 0x80 != 0
+    v0: V0Message,
```
dnut, line 138: There's no need for this to be a tagged union because v0 and legacy messages contain the same data. They can be represented with the same struct. We put a lot of thought into this for v1 after initially implementing it as a tagged union and then realizing a struct would be easier to use without any downsides. The only difference is that v0 messages include a version byte in the serialization format, and legacy messages do not. But you don't need a tagged union to represent this. If you want round-trip serialization, you only need a version field.

Likewise you can avoid all the naming complexity with `Versioned*`, `Legacy*`, `V0*`, etc. and simply stick to `Transaction` and `Message`. You can also eliminate the `MessageHeader` struct since it can be flattened into `Message` since that would be the only usage.

v2/lib/solana/bincode.zig
```diff
@@ -0,0 +1,466 @@
+//! Types and encoders/decoders for protocol-defined data using bincode
```
dnut, line 1: Isn't this basically the whole point of the `solana` folder? I think `bincode.zig` is a confusing name because it can become a catch-all for every type in the solana folder. Also that name makes it seem like it's a bincode library. I think a better name here would be `transaction.zig`. You could also split out some of the integer and collection serializers into something called `primitives.zig` or `bincode_primitives.zig` if you really want to emphasize the bincode-specific aspect.

v2/services/shred_receiver.zig
```diff
@@ -1,19 +1,58 @@
...
+//!       in-progress FEC set due to a mismatch (i.e. when the Signature or Merkle root are
+//!       different) are dropped under equivocation.
+//!
+//!       For example, if the shreds of two mismatching FEC sets of the same FecSetId came in
+//!       interleaved we would drop the 2nd FEC set. In this case we may have to get shreds from
+//!       the missing set from repair. I expect this to be rare, and this behaviour matches
+//!       Firedancer's.
+//!
+//!       Once we complete a FEC set, we are free to build another FEC set of that very same
+//!       FecSetId; the downstream service must be fully aware of equivocation.
```
dnut, line 44: The code will refuse to build conflicting sets until the first one is completed, and then suddenly it's perfectly happy to do it. I know we discussed this previously, but I just dont see the purpose of this arbitrary sequencing constraint.

What harm of reconstructing the second one disappears as soon as the first one is completed? What benefit of reconstructing the second one only appears after the first one is completed?

The main problem I see here is if the first set is partially constructed but never completed, we'll never be able to receive the conflicting set.

If the goal is to prevent DoS, why not use per-slot limits? That seems more straightforward and more effective.

v2/lib/shred/receiver.zig
```diff
@@ -0,0 +1,750 @@
...
+            "slot: {}, idx: {}",
+            .{ fec_set_id.slot, fec_set_id.fec_set_idx },
+        );
+        zone.text(str);
+
+        const fec_set_ctx = if (state.in_progress.getFecSetCtx(
+            &shred.signature,
+        )) |fec_set_ctx| existing_set: {
+            // fec set is already being built
+            @branchHint(.likely); // ~31/32 expected
```
dnut: Isn't this actually only about 1/2 the time? In the other case, it says "likely finished already" which implies the context is removed after reconstruction. Reconstruction occurs after half of the shreds have been received. My assumption is that a validator would typically receive nearly every shred.

v2/lib/shred/receiver.zig
```diff
@@ -0,0 +1,750 @@
...
+
+        const new_done: *DoneItem = self.done_pool.indexToPtr(new_pool_id);
+        new_done.* = .{ .id = id, .signature_hashed = hashSignature(signature) };
+        self.eviction.add(new_pool_id) catch unreachable;
+        const entry = self.done_map.getOrPutAssumeCapacityAdapted(&id, done_ctx);
+        std.debug.assert(!entry.found_existing);
+        entry.value_ptr.* = new_done;
+    }
+
+    fn doneSignatureHash(
```
dnut: This name is very confusing to me. I had no idea what it means until I read the whole function. And even then, I don't think it communicates anything useful to callers. I think a simple name like `lookup` would be clearer.

v2/lib/shred/receiver.zig
```diff
@@ -0,0 +1,750 @@
...
+                // all other conflicting shreds are dropped until the in-progress fec set is built.
+                //
+                // This is intention as it stops the leader from producing many equivocating shreds
+                // that would a) fill up our in-progress map, and b) starve our CPU from shred
+                // verification.
+                //
+                // TODO: once repair is implemented, repaired shreds should skip these checks to
+                // allow conflicting fec sets to be inside the in-progress map. We will need to do
+                // this to reliably repair when equivocation is detected.
+                .mismatching_signature => return error.EquivocationDifferentHashForSameFecSetId,
```
dnut, line 170: Returning an error here means it will never reconstruct a conflicting set, even after the first one is completed.

v2/lib/shred/receiver.zig
```diff
@@ -0,0 +1,750 @@
...
+
+                .data_shreds_buf = undefined,
+                .code_shreds_buf = undefined,
+            };
+
+            break :new_set fec_set_ctx;
+        };
+
+        // in the case that we just acquired a fec set, it is critical that we do not leak it
+        errdefer comptime unreachable;
```
dnut, line 217: What do you mean by "leak"?

v2/lib/shred/receiver.zig
```diff
@@ -0,0 +1,750 @@
...
+            if (state.in_progress.containsId(fec_set_id)) {
+                // NOTE: see above note.
+                return error.EquivocationMatchingFecSetWithDifferentSignatureAlreadyInProgress;
+            }
+
+            // This is the first shred of a new in-progress fec set.
+            const slot_leader = leader_schedule.get(shred.slot) orelse return error.UnknownLeader;
+
+            var shred_merkle_root: Hash = undefined;
+            try shred.merkleRoot(&shred_merkle_root);
```
dnut, line 183: I only see this being called for the first shred received in the set. How are you verifying that the other shreds belong in the merkle tree?

v2/lib/shred/receiver.zig
Sobeston: Any objects allocated from a pool must be destroyed, i.e. returned to the pool. Otherwise the pool will fill up with stuff we're not using anymore

v2/lib/shred/receiver.zig
Sobeston: The merkle tree is protected by the shred's signature (which is the same across a fec set), which our in-progress map is keyed by. 

i.e. finding an in-progress set means we've already matched with a previously calculated merkle root.

v2/services/shred_receiver.zig
Sobeston: > we'll never be able to receive the conflicting set.

We should bypass this check for shreds coming in from repair (when implemented)

> I just dont see the purpose of this arbitrary sequencing constraint.

It simplifies state - I'm prioritising simplicity and, in the case of a malicious leader, preferring resilience over optimal performance.

> If the goal is to prevent DoS, why not use per-slot limits? 

This is possible, but it opens more questions and complicates some state

v2/lib/shred/receiver.zig
Sobeston: I think I wrote this when I was checking the done map first? But I don't unconditionally check done any more, so this comment/hit is stale

So yes it's probably something like 31/64 assuming no packet loss

v2/lib/solana/bincode.zig
Sobeston: > Isn't this basically the whole point of the solana folder? 

Well it's part of it, the solana library contains other things (hashes, pubkeys, clusters, etc).

> Also that name makes it seem like it's a bincode library. I think a better name here would be transaction.zig. You could also split out some of the integer and collection serializers into something called primitives.zig or bincode_primitives.zig if you really want to emphasize the bincode-specific aspect.

This has all been rewritten anyway in my replay branch (sobe/v2-replay)

v2/lib/shred/receiver.zig
Sobeston: I believe anything up to 67+67 used to be legal, you can still see it in some code, e.g. from FD
```c
#define FD_REEDSOL_DATA_SHREDS_MAX   (67UL)
#define FD_REEDSOL_PARITY_SHREDS_MAX (67UL)
```

> every set is exactly 32+32

from our perspective, 32+32 is the max as we will(/should) never receive more than that

v2/lib/shred/receiver.zig
Sobeston: > What do you think about using the effects interface pattern

I don't think it's much of a value-add here, all of the parameters are easily pluggable already

v2/lib/shred.zig
Sobeston: I believe only rooted features "count" for the shred receiver, I don't see it making sense any other way. I think we'd only need it to send shred receiver an updated featureset before the next epoch boundary?

v2/lib/solana/bincode.zig
Sobeston: Makes sense, will do this in the replay branch based off this

v2/lib/shred/receiver.zig
Sobeston: When we have repair, shreds from it should bypass the check

v2/lib/shred/receiver.zig
Sobeston: renamed https://github.com/Syndica/sig/pull/1339/commits/85b9d556e0a33605f57c024c434a8b59c3899d60

v2/lib/shred/receiver.zig
Sobeston: changed comment https://github.com/Syndica/sig/pull/1339/commits/85b9d556e0a33605f57c024c434a8b59c3899d60

v2/lib/shred.zig
Sobeston: removed code for deprecated shred types https://github.com/Syndica/sig/pull/1339/commits/85b9d556e0a33605f57c024c434a8b59c3899d60

v2/lib/shred.zig
Sobeston: removed code for deprecated shred types https://github.com/Syndica/sig/pull/1339/commits/85b9d556e0a33605f57c024c434a8b59c3899d60

v2/lib/shred.zig
Sobeston: removed code for deprecated shred types https://github.com/Syndica/sig/pull/1339/commits/85b9d556e0a33605f57c024c434a8b59c3899d60

v2/lib/shred/receiver.zig
dnut: Sure, it is the max, but it's also the min. It's the exact value. It doesn't make sense to call it a "max" when it's just the exact value.

v2/lib/solana/bincode.zig
dnut: Why not this branch?

v2/lib/shred/receiver.zig
dnut: It's not enough to see that a shred contains a signature you've already verified. You still need to check that the shred actually belongs to the corresponding merkle tree. The signature only needs to be verified once, but the merkle proof must be verified for every shred.

If you skip this, a malicious node can trick you. The leader builds a valid merkle tree over shreds `[A, B, ...]`, computes root `MR`, and signs it as `S`, attaching `S` to each shred. Another node creates a fake shred `B'`, which is *not* in that tree, but they attach the same signature `S` and send it to you.

You receive `A`, verify its merkle proof and signature, and accept `MR` and `S` as valid. Later, you receive `B'`. Since it carries the same signature `S`, you might assume it belongs to the same set, but that's wrong. `B'` would produce a different root if you actually checked its merkle proof. Verifying the proof for every single shred is the only way to prevent this attack.

v2/lib/shred/receiver.zig
dnut: How about my second proposal then? It simplifies this function and makes it less opinionated about where its data comes from. The caller can manage the ring buffer. Sure it's not hard to create a ring buffer for a test, but why bother when you can easily avoid the unnecessary dependency?

v2/lib/shred/receiver.zig
Sobeston: Good catch, fixed https://github.com/Syndica/sig/pull/1339/commits/80bd0b5e26a300c4ac0cea0b87070b09aaba4226

v2/lib/solana/bincode.zig
Sobeston: All the bincode stuff is rewritten in the replay branch, it'd mean pulling more changes into this one

v2/lib/shred/receiver.zig
Sobeston: > but it's also the min

Not from our perspective, we will basically never have 32 code shreds or 32 data shreds. We keep track of 1<=x<=31 shreds overall, completing on the 32nd. For us the min number of shreds to be concerned with is 1

dnut: OK I see. The root of the disagreement is in what we are interpreting this to be the maximum of:
- the number of shreds that we've received from the fec set -- *your interpretation*
- the number of shreds that are in the fec set -- *my interpretation*

I need to take a closer look at the code

v2/lib/shred.zig
```diff
@@ -1,2219 +1,497 @@
...
+    ///
+    pub const Variant = packed struct(u8) {
+        merkle_count: u4,
+        kind: enum(u4) {
+            // all supported kinds
+            merkle_code_chained = 0x6,
+            merkle_code_chained_resigned = 0x7,
+            merkle_data_chained = 0x9,
+            merkle_data_chained_resigned = 0xB,
+            _,
```
dnut: Why is this necessary? Shreds are required to be one of the four specified variants. If it's not one of those, that means it's not a valid shred, so it should fail deserialization and be discarded immediately. Supporting this adds code complexity and weakens the type, opening the door to future bugs.

v2/lib/shred.zig
```diff
@@ -1,2219 +1,497 @@
...
-                .payload = payload,
+        // [firedancer] https://github.com/firedancer-io/firedancer/blob/9f7770af997a1443e7903113fc03ca1ce3b0ad73/src/ballet/shred/fd_shred.c#L16
+        // Legacy (non-merkle), and non-chained shreds are deprecated
+        // https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0313-drop-unchained-merkle-shreds.md
+        fn isSupported(self: Variant) bool {
+            return switch (self.kind) {
+                _ => false,
+                else => true,
             };
         }
```
dnut: should be implied by being deserialized without error

v2/lib/shred.zig
```diff
@@ -1,2219 +1,497 @@
...
-        fn id(self: *const Self) ShredId {
-            return .{
-                .slot = self.common.slot,
-                .index = self.common.index,
-                .shred_type = self.common.variant.shred_type,
+        fn isResigned(self: Variant) bool {
+            return switch (self.kind) {
+                .merkle_data_chained_resigned, .merkle_code_chained_resigned => true,
+                .merkle_data_chained, .merkle_code_chained => false,
+                _ => unreachable, // unsupported variant
```
dnut: runtime assertion could be guaranteed at comptime

v2/lib/shred.zig
dnut: hmm i suppose that makes sense at some level because features are only supposed to be updated at epoch boundaries. but you can have forks that extend over epoch boundaries. during those few slots there may be some disagreements about features.

i'd like a protocol spec for the timing of feature activations, and then make sure every client enforces these rules. right now it all feels a bit fuzzy

feel free to resolve this conversation so it doesn't block merge

v2/lib/shred.zig
Sobeston: We don't deserialise, and we check that is a known variant immediately

v2/lib/shred/receiver.zig
dnut: I don't see a downside to the pointer based approach. I think it's a cleaner api so I say go for it unless you have an objection.

v2/lib/shred.zig
Sobeston: If we tell Zig that there's only X possible variants, and we cast a value (which is not one of this variants) into said enum type, this is illegal behavior. 

By using an exhaustive enum, we can explicitly guard against unknown/invalid variants (rather than asserting that there aren't any)

v2/lib/shred.zig
Sobeston: Same as above, we *don't* deserialise

v2/lib/shred.zig
Sobeston: Same as above

Sobeston: These assertions are also genuinely unreachable, we always call isSupported before.

v2/lib/shred/receiver.zig
Sobeston: It's tempting, but there's also not a clear way to do this while remaining optimal. Currently I acquire the buffer exactly when I need it, and (more importantly) send it off as soon as it is safe to do so. 

Acquiring the buffer "early" wouldn't be harmful, but notifying downstream late costs us latency

v2/lib/solana/bincode.zig
dnut: i see the replay branch still has this code

v2/lib/shred/receiver.zig
dnut: i just don't follow what kind of leak could happen if you return an error after this. you created an item and already put it in the in_progress map. it's going to sit there and be equally recyclable even if you return with an error immediately on this line. it's not any more or less likely to leak just because you return early from this function.

the problem with returning early is that the data is going to be malformed. i think what this errdefer is actually helping to ensure is that the item you just allocated is going to be properly initialized by the time this function returns

v2/lib/shred.zig
dnut: >  we check that is a known variant immediately

Sure, I get that, but it's not great when a type's methods rely on assumptions about how the type will be used in other places. "you must call function X before function Y or the whole program crashes." Sometimes we make assumptions like that for performance, but if if it can be avoided without cost, that's better. People who work on this code in the future may not realize this requirement, and they may introduce bugs because of that. It seems obvious to you now, but things like this can fall apart as a codebase evolves.

> We don't deserialise

You're right. It's not "deserialization" but just a pointer cast. Regardless my point still stands. There is a deserialization-like process that has a slice of bytes as its input and a pointer to a shred as its output. That process should be responsible for ensuring the type is well-formed.

Currently you ensure the type meets a basic standard of being well-formed by making the type weak enough to ensure that any backing bytes are actually a legal representation of the type. That's a start, but it's not actually well-formed from a "this type is usable" perspective.

A stronger approach would be to have a your "deserialization" function (that calls ptrCast) check the offset of the variant to make sure it's legal. Then it can safely cast the variant as an exhaustive enum. This allows the type to be stronger and eliminates unnecessary assumptions and maintenance risks.

v2/services/shred_receiver.zig
dnut: How does it simplify state? The constraint is imposed on top of state that's already capable of tracking conflicting simultaneous fec sets. If anything it seems to me that it makes state more complex because you need this distinction between in-progress vs completed fec sets in order to support the constraint. I feel like it would simplify the state to consolidate it.

You understand this code better than I do, so if this seems wrong to you, we can go with your approach. I need some more time to study the code and flesh out my thoughts. My idea is not a high priority right now.

v2/lib/collections/lcrs_tree.zig
```diff
@@ -0,0 +1,46 @@
...
+        }
+        pub fn setParent(ctx: Context, node: *Node, parent: *Node) void {
+            _ = .{ ctx, node, parent };
+        }
+        pub fn setChild(ctx: Context, node: *Node, child: *Node) void {
+            _ = .{ ctx, node, child };
+        }
+        pub fn setSibling(ctx: Context, node: *Node, sibling: *Node) void {
+            _ = .{ ctx, node, sibling };
+        }
```
InKryption, line 28: I know this is how it's been done for everything else, but just want to note/ask, couldn't we could be doing it like this instead:
```suggestion
    lib.util.assertInterface(Context, struct {
        pub const parentOf: fn (ctx: Context, node: *const Node) *Node = undefined;
        pub const childOf: fn (ctx: Context, node: *const Node) *Node = undefined;
        pub const siblingOf: fn (ctx: Context, node: *const Node) *Node = undefined;
        pub const setParent: fn (ctx: Context, node: *Node, parent: *Node) void = undefined;
        pub const setChild: fn (ctx: Context, node: *Node, child: *Node) void = undefined;
        pub const setSibling: fn (ctx: Context, node: *Node, sibling: *Node) void = undefined;
```
would have the additional guarantee that it guarantees nothing could be calling these stub functions (without it being a compile error).

Sobeston: Hm that's a good point, I've been having mixed feelings about this Tree API honestly, considering removing it in a follow up. Really there's only a few spots we'll use trees like this (merkle forest, blocks, maybe epochs?), might not be worth the generic

v2/lib/shred.zig
Sobeston: Yeah I understand, we *could* have two entirely separate types for shreds (i.e. before checking variant, and after), but I don't think it adds any value here.

We don't need to add an additional type that will be used for about 3 lines and not used anywhere else, and anyone working on this code can very clearly see the assertion being made here. Either we duplicate code or add a weird generic that's more confusing than the assertion already in

This type doesn't really have to be super ergonomically usable for anything but the shred receiver anyway.

If you feel strongly about this then you can open a follow-up PR

v2/lib/solana/bincode.zig
Sobeston: Oh yep I went back on that, but either way this code should be pretty ephemeral, this should be replaced by more custom deserialisation logic soon.

v2/services/shred_receiver.zig
Sobeston: We already have a distinction between in-progress and completed for other reasons, this is already very simple

v2/lib/shred/receiver.zig
Sobeston: There's some points where, if errors could occur, we could get bad state resulting in a leak. 
e.g. if we "finish" a fec set, but don't actually remove it from in progress.

> the problem with returning early is that the data is going to be malformed. 

This is also true

I think in a follow-up we may want to handle an error in there unfortunately, so I will have to write a proper errdefer path

v2/lib/shred.zig
dnut: I agree about not adding another type. I'm just suggesting doing the check before the ptrCast instead of after. It would actually simplify the code.

v2/lib/collections/lcrs_tree.zig
InKryption: I would be happy with that; always find tree abstractions be kinda iffy, even moreso than linked lists.

v2/lib/shred.zig
Sobeston: d6b6df852e25d82aebd25b1bd6cda8a83a2e937d

v2/lib/collections/lcrs_tree.zig
Sobeston: Cool, I think I'll remove this in the replay PR

v2/lib/shred.zig
dnut: I guess the 4-bit aspect of it makes it more awkward to do the validation than i imagined, but I still think this is an improvement

v2/services/shred_receiver.zig
dnut: alright i'll give this some more thought and decide if i want to revisit it after merge

dnut (dismissed):

InKryption (dismissed):

dnut (approved):

InKryption (approved):

PR #1344, opened 2026-03-25
feat(jrpc-ws): add blockSubscribe method (streaming)

prestonsn:
Implements [blockSubscribe](https://solana.com/docs/rpc/websocket/blocksubscribe) JRPC Web Socket method

### Testing Done

via the solana pubsub crate: 

| Test | Args | Result |
|------|------|--------|
| block default | `cargo run -- block` | PASS - ~93 blocks/20s, ~630 txs/block |
| block signatures | `cargo run -- block --details signatures` | PASS - sigs populated, txs=0 |
| block none | `cargo run -- block --details none` | PASS - no tx/sig data, metadata only |
| block base64 | `cargo run -- block --encoding base64` | PASS - 116 blocks/20s |
| block base58 | `cargo run -- block --encoding base58` | PASS - 116 blocks/20s |
| block jsonParsed | `cargo run -- block --encoding json-parsed` | PASS - parsed vote instructions visible |
| block confirmed | `cargo run -- block --commitment confirmed` | PASS - 80 blocks/20s |
| block finalized | `cargo run -- block --commitment finalized` | PASS - 43 blocks/20s (~half of confirmed) |
| block filter pubkey | `cargo run -- block --filter 11111111111111111111111111111111 --details signatures` | PASS - ~2-23 sigs/block (filtered) |
| block raw JSON | `cargo run -- block --raw --details none` | PASS - full JSON with blockHeight, rewards, previousBlockhash |
| slot subscribe | `cargo run -- slot` | PASS - 13 updates/15s |
| block jsonParsed raw | `cargo run -- block --encoding json-parsed --raw` | PASS - parsed towerSync lockouts visible |


src/rpc/jrpc_websockets/types.zig
```diff
@@ -7,8 +7,18 @@ const ws_request = @import("ws_request.zig");
...
-const TransactionError = sig.ledger.transaction_status.TransactionError;
+const TransactionError =
+    sig.ledger.transaction_status.TransactionError;
+const InnerInstructions =
+    sig.ledger.transaction_status.InnerInstructions;
+const TransactionTokenBalance =
+    sig.ledger.transaction_status.TransactionTokenBalance;
+const LoadedAddresses =
+    sig.ledger.transaction_status.LoadedAddresses;
+const TransactionReturnData =
```
prestonsn: clean up wrapped

src/replay/Committer.zig
```diff
@@ -481,7 +431,226 @@ fn writeTransactionStatus(
...
+        .inner_instructions = entry.inner_instructions,
+        .log_messages = entry.log_messages,
+        .pre_token_balances = entry.pre_token_balances,
+        .post_token_balances = entry.post_token_balances,
+        .rewards = null,
+        .loaded_addresses = entry.loaded_addresses,
+        .return_data = entry.return_data,
+        .compute_units_consumed = entry.compute_units_consumed,
+        .cost_units = null,
+    };
```
prestonsn, line 689: address the null fields

src/replay/Committer.zig
```diff
@@ -73,13 +73,14 @@ pub fn commitTransactions(
...
     var priority_fees: u64 = 0;
 
-    var maybe_logs_batch_event: ?jrpc_types.SlotTransactionLogs = null;
-    defer if (maybe_logs_batch_event) |*logs_batch_event| {
-        logs_batch_event.deinit();
-    };
-    var batch_log_entries: ArrayListUnmanaged(jrpc_types.TransactionLogsEntry) = .{};
+    // When an event sink is present, build a rich
+    // transaction batch event (replaces the old logs-only
+    // event). The arena owns all TransactionEntry data.
```
jbuckmccready: I prefer to avoid these comments that reference code that no longer exists.

src/replay/Committer.zig
```diff
@@ -184,17 +166,67 @@ pub fn commitTransactions(
...
+                    try writeTransactionStatus(
+                        temp_allocator,
+                        ledger,
+                        slot,
+                        transaction,
+                        tx_result.*,
+                        transaction_index,
+                        self.account_store,
+                    );
+                }
```
jbuckmccready: Instead of fallback we're having errors like this just hard fail/propagate, even for RPC specific things. Can see discussion here: https://github.com/Syndica/sig/pull/1317#discussion_r3021702216

src/replay/Committer.zig
```diff
@@ -456,7 +426,224 @@ fn writeTransactionStatus(
...
+/// `TransactionEntry` that can be cached and reused for both
+/// ledger writes and notification construction.
+fn buildTransactionEntry(
+    arena: Allocator,
+    temp_allocator: Allocator,
+    transaction: ResolvedTransaction,
+    tx_result: ProcessedTransaction,
+    is_vote: bool,
+    account_store: ?SlotAccountStore,
+) !jrpc_types.TransactionEntry {
```
jbuckmccready, line 493: We should have a tracy zone for this function.

src/replay/Committer.zig
```diff
@@ -456,7 +426,224 @@ fn writeTransactionStatus(
...
+    defer mint_cache.deinit();
+    for (tx_result.writes.constSlice()) |*written| {
+        if (written.account.data.len >= spl_token.MINT_ACCOUNT_SIZE) {
+            if (spl_token.ParsedMint.parse(
+                written.account.data[0..spl_token.MINT_ACCOUNT_SIZE],
+            )) |mint| {
+                mint_cache.put(
+                    written.pubkey,
+                    mint.decimals,
+                ) catch {};
```
jbuckmccready: Shouldn't just throw away error here.

src/replay/Committer.zig
```diff
@@ -456,7 +426,224 @@ fn writeTransactionStatus(
...
+    ) catch null;
+
+    const post_raw = collectPostTokenBalances(transaction, tx_result);
+    const post_token_balances = spl_token.resolveTokenBalances(
+        arena,
+        post_raw,
+        &mint_cache,
+        FallbackAccountReader,
+        mint_reader,
+    ) catch null;
```
jbuckmccready: Shouldn't discard errors here.

src/replay/service.zig
```diff
@@ -616,37 +616,72 @@ fn freezeCompletedSlots(state: *ReplayState, results: []const ReplayResult) !boo
...
+                        };
+                        // Ownership transferred to event.
+                        distributed = .empty();
+                    } else |err| {
+                        state.logger.err().logf(
+                            "failed to materialize modified accounts for frozen slot {}: {}",
+                            .{ slot, err },
+                        );
+                    }
                 }
```
jbuckmccready, line 856: Bad merge/rebase, accounts send and error handling should match the target branch.

src/replay/service.zig
```diff
@@ -616,37 +616,72 @@ fn freezeCompletedSlots(state: *ReplayState, results: []const ReplayResult) !boo
...
+                // Free rewards if not transferred.
+                distributed.deinit();
```
jbuckmccready: This should be in a defer to keep things clear and avoid bugs.

src/rpc/jrpc_websockets/integration_tests/logs_subscription_tests.zig
```diff
@@ -43,7 +43,11 @@ fn parseLogsNotification(
 }
 
 const types = lib.types;
-const TransactionError = sig.ledger.transaction_status.TransactionError;
+const TransactionError =
+    sig.ledger.transaction_status.TransactionError;
+const LoadedAddresses =
+    sig.ledger.transaction_status.LoadedAddresses;
```
prestonsn: fix wrapping

src/rpc/jrpc_websockets/Runtime.zig
```diff
@@ -58,6 +58,9 @@ threadpool_wg: std.Thread.WaitGroup,
 /// Optional callback run on the loop thread whenever the async wake fires.
 /// Used by integrations to drain additional loop-thread-only queues.
 wakeup_hook: ?WakeupHook = null,
+/// Ledger handle for block subscriptions. Optional because not all deployments
+/// have a ledger available (e.g. tests).
+ledger: ?*sig.ledger.Ledger = null,
```
jbuckmccready: This is unused, all data from ledger is sent in now via events.

src/rpc/jrpc_websockets/Runtime.zig
```diff
@@ -329,19 +335,16 @@ fn handleInboundEvent(
...
-    // .slot_frozen and .logs cases take ownership of the inner data
+    // .slot_frozen and .transaction_batch cases take ownership of the inner data
     var e = event;
     defer e.deinit();
 
     switch (e) {
-        .logs => |*log_data| {
-            // NOTE: logs events only populate the slot cache; publication happens on later
-            // slot transitions. Replay may send multiple log batches for the same slot before
-            // `.slot_frozen`, and the cache accumulates them until the slot is published.
```
jbuckmccready, line 341: Useful comment was nuked.

src/rpc/jrpc_websockets/Runtime.zig
```diff
@@ -707,6 +737,47 @@ fn deletedAccountPlaceholder() sig.core.Account {
...
 
+fn enqueueBlockJob(
+    self: *Runtime,
+    entry: *sub_map_mod.MapEntry,
+    slot: Slot,
+    cached: *const SlotStateCache.CachedSlot,
+    task_batch: *ThreadPool.Batch,
+) void {
+    const bp = entry.key.params.block;
+    const parent_slot = cached.parent orelse slot -| 1;
```
jbuckmccready: Seems like this should log an error on the else branch since parent should always be populated at this point.

src/rpc/jrpc_websockets/Runtime.zig
```diff
@@ -707,6 +737,47 @@ fn deletedAccountPlaceholder() sig.core.Account {
...
+
+    var arena = std.heap.ArenaAllocator.init(self.allocator);
+    const block = cached.buildConfirmedBlock(arena.allocator(), parent_slot) catch |err| {
```
jbuckmccready, line 888: This should have a TODO(perf) comment on it, similar to other jobs that have to allocate we should have ref counter on the arena slot cached structure to avoid all this work on the IO loop.

src/rpc/jrpc_websockets/Runtime.zig
```diff
@@ -717,7 +788,7 @@ fn publishLogsSubscriptionForEntry(
     const logs_filter = entry.key.params.logs.filter;
     var log_entries = cached_slot.logEntriesIterator();
     while (log_entries.next()) |log_event| {
-        if (!logsEventMatchesFilter(logs_filter, log_event)) {
+        if (!logsEventMatchesFilter(logs_filter, &log_event)) {
```
jbuckmccready: Naming here is confusing, it's a log entry, then it's a log event, but the type for the function parameter is a TransactionBatchView. I think it's supposed to be a batch view iterator etc.

src/rpc/jrpc_websockets/Runtime.zig
```diff
@@ -960,6 +1033,11 @@ fn handleCommitMsg(self: *Runtime, msg: types.CommitMsg) void {
     switch (msg.result) {
         .serialize_error => {
             self.metrics.serialize_errors += 1;
+            // Cancel the reserved slot so later entries can
+            // still commit (avoids blocking the prefix).
+            if (msg.index) |idx| {
+                q.cancelReservation(idx);
+            }
```
jbuckmccready, line 1184: I don't there's much to do done at this point but this does conflict with the signatureSubscribe PR (https://github.com/Syndica/sig/pull/1345).

src/rpc/jrpc_websockets/SlotStateCache.zig
```diff
@@ -6,10 +6,18 @@ const sig = @import("../../sig.zig");
 const types = @import("types.zig");
 
 const Slot = sig.core.Slot;
+const Reader = sig.ledger.Reader;
```
jbuckmccready: This alias weird, never used directly so it might as well just have the full path to types used to avoid confusion (we never use the Reader).

src/rpc/jrpc_websockets/SlotStateCache.zig
```diff
@@ -56,40 +73,244 @@ pub const CachedSlot = struct {
...
+        entry_idx: usize = 0,
+
+        pub const View = struct {
+            signature: sig.core.Signature,
+            err: ?sig.ledger.transaction_status.TransactionError,
+            is_vote: bool,
+            logs: []const []const u8,
+            mentioned_pubkeys: []const sig.core.Pubkey,
+
+            pub fn toOwnedNotificationData(
```
jbuckmccready: Naming here is weird when it's returning notification specific for logs. Relates to naming in Runtime.zig.

src/rpc/jrpc_websockets/SlotStateCache.zig
```diff
@@ -193,8 +414,10 @@ pub fn onTipChanged(
     };
 }
 
-/// Freeze is the first point where the slot's final modified accounts are stable,
-/// so cache them once here and reuse them for later confirmed/finalized publication.
+/// Freeze is the first point where the slot's final
+/// modified accounts are stable, so cache them once here
+/// and reuse them for later confirmed/finalized
+/// publication.
```
jbuckmccready: Excessive wrapping.

src/rpc/jrpc_websockets/SlotStateCache.zig
```diff
@@ -203,28 +426,42 @@ pub fn onSlotFrozen(
     const slot_state = try self.getOrPutSlot(allocator, slot_data.slot);
     const duplicate_frozen = slot_state.state.frozen;
     if (duplicate_frozen) {
-        // TODO: what about repair/replay over same slot? Could be frozen twice for same
-        // slot number. For now just overwrite the cached data and log it.
+        // TODO: what about repair/replay over same slot?
+        // Could be frozen twice for same slot number. For
+        // now just overwrite the cached data and log it.
```
jbuckmccready: Excessive wrapping.

src/rpc/jrpc_websockets/SlotStateCache.zig
```diff
@@ -203,28 +426,42 @@ pub fn onSlotFrozen(
...
+
     if (duplicate_frozen) {
         // TODO: for now just not emitting anything, this needs to be addressed by tracking more
         // than just slot numbers as a larger refactor in Sig
         return .{};
     }
 
-    // Maybe other commitments for slot were waiting for frozen, so try all of them.
+    // Maybe other commitments for slot were waiting
+    // for frozen, so try all of them.
```
jbuckmccready: Excessive wrapping.

src/rpc/jrpc_websockets/SlotStateCache.zig
```diff
@@ -236,23 +473,19 @@ pub fn onSlotFrozen(
...
-        log_event.deinit();
+    if (batch_event.entries.len == 0) {
+        batch_event.deinit();
         return;
     }
 
-    const slot_state = try self.getOrPutSlot(allocator, log_event.slot);
-    // TODO: what about repair/replay over same slot? We can receive multiple log batches for the
-    // same slot during normal execution and append them, but if the slot is replayed with
-    // different contents we still only key by slot number and do not handle that case.
```
jbuckmccready, line 253: Comment nuked.

src/rpc/jrpc_websockets/SlotStateCache.zig
```diff
@@ -591,92 +795,6 @@ test "duplicate frozen overwrites cached slot data" {
     try std.testing.expectEqual(4, cached.root);
 }
 
-test "transaction logs batch ownership transfers into cache" {
```
jbuckmccready, line 607: Basic tests were removed with no equivalent replacement.

src/rpc/jrpc_websockets/protocol.zig
```diff
@@ -304,6 +304,125 @@ pub fn serializeLogsNotification(
...
+/// Convert WS TransactionDetails to the HTTP/common TransactionDetails
+/// used by encodeBlockWithOptions.
+fn toCommonTxDetails(td: methods.TransactionDetails) sig.rpc.methods.common.TransactionDetails {
+    return switch (td) {
+        .full => .full,
+        .accounts => .accounts,
+        .signatures => .signatures,
+        .none => .none,
+    };
+}
```
jbuckmccready: Why do we need these? Seems like we should be able to use existing stuff from `getBlock`?

src/rpc/hooks.zig
```diff
@@ -90,7 +90,7 @@ pub const Hooks = struct {
             const method = if (@hasField(Method, decl.name))
                 @field(Method, decl.name)
             else
-                @compileError("No RPC method named: " ++ decl.name);
+                continue;
```
jbuckmccready: What is this change for/about?

jbuckmccready (changes requested):

src/rpc/hooks.zig
prestonsn: This has to do with changing the visibility of `encodeBlockWithOptions()` in `Ledger.zig`. Making it `pub` caused this to fail to compile here due to the way this code parses `pub` methods from the hook source files (expects them to have the fixed RPC method function signature). 

I wanted to avoid a larger diff, but I think having the compile error for malformed rpc methods is more correct. 

Lifted the `encodeBlockWithOptions()` and downstream fns it calls into its own file: https://github.com/Syndica/sig/pull/1344/commits/eeced50238c1f596797c91418aeff9e7da13a9dd

src/replay/Committer.zig
prestonsn: https://github.com/Syndica/sig/pull/1344/commits/eeced50238c1f596797c91418aeff9e7da13a9dd

src/rpc/jrpc_websockets/Runtime.zig
prestonsn: https://github.com/Syndica/sig/pull/1344/commits/eeced50238c1f596797c91418aeff9e7da13a9dd

src/rpc/jrpc_websockets/SlotStateCache.zig
prestonsn: https://github.com/Syndica/sig/commit/eeced50238c1f596797c91418aeff9e7da13a9dd

src/rpc/jrpc_websockets/SlotStateCache.zig
prestonsn: https://github.com/Syndica/sig/commit/eeced50238c1f596797c91418aeff9e7da13a9dd

src/rpc/jrpc_websockets/SlotStateCache.zig
prestonsn: https://github.com/Syndica/sig/commit/eeced50238c1f596797c91418aeff9e7da13a9dd

src/rpc/jrpc_websockets/SlotStateCache.zig
prestonsn: https://github.com/Syndica/sig/commit/eeced50238c1f596797c91418aeff9e7da13a9dd

src/rpc/jrpc_websockets/SlotStateCache.zig
prestonsn: https://github.com/Syndica/sig/pull/1344/changes/059ad3ba87c46bf157e7513d6d28437dc1ce35dd

src/rpc/jrpc_websockets/Runtime.zig
prestonsn: https://github.com/Syndica/sig/pull/1344/changes/059ad3ba87c46bf157e7513d6d28437dc1ce35dd

src/rpc/jrpc_websockets/protocol.zig
prestonsn: https://github.com/Syndica/sig/pull/1344/changes/059ad3ba87c46bf157e7513d6d28437dc1ce35dd

src/replay/service.zig
prestonsn: https://github.com/Syndica/sig/pull/1344/commits/7de08ac709b05c3d34792ff9639ba611bea30d1d

src/replay/service.zig
prestonsn: https://github.com/Syndica/sig/pull/1344/commits/7de08ac709b05c3d34792ff9639ba611bea30d1d

src/replay/Committer.zig
prestonsn: https://github.com/Syndica/sig/pull/1344/commits/7de08ac709b05c3d34792ff9639ba611bea30d1d

src/replay/Committer.zig
prestonsn: Makes sense, and I think this was due to bad rebase: https://github.com/Syndica/sig/commit/eeced50238c1f596797c91418aeff9e7da13a9dd

src/replay/Committer.zig
prestonsn: https://github.com/Syndica/sig/pull/1344/commits/7de08ac709b05c3d34792ff9639ba611bea30d1d

src/rpc/jrpc_websockets/Runtime.zig
prestonsn: https://github.com/Syndica/sig/pull/1344/commits/0692e73a6a082619c2d39fe4b7d4a109993ceb6e

src/rpc/jrpc_websockets/Runtime.zig
prestonsn: https://github.com/Syndica/sig/pull/1344/commits/0692e73a6a082619c2d39fe4b7d4a109993ceb6e

src/rpc/jrpc_websockets/Runtime.zig
prestonsn: oops, due to rebase: https://github.com/Syndica/sig/pull/1344/commits/0692e73a6a082619c2d39fe4b7d4a109993ceb6e

src/rpc/jrpc_websockets/SlotStateCache.zig
prestonsn: https://github.com/Syndica/sig/pull/1344/commits/0692e73a6a082619c2d39fe4b7d4a109993ceb6e

src/replay/Committer.zig
prestonsn: https://github.com/Syndica/sig/pull/1344/commits/0692e73a6a082619c2d39fe4b7d4a109993ceb6e

src/rpc/jrpc_websockets/Runtime.zig
prestonsn: I see, i can fix this conflict after https://github.com/Syndica/sig/pull/1345 is merged

src/rpc/jrpc_websockets/SlotStateCache.zig
prestonsn: oops, re-added the tests: https://github.com/Syndica/sig/pull/1344/commits/a24bacf8ef4d9a6f282e5544b230d968f3853b6c

src/replay/Committer.zig
```diff
@@ -214,82 +227,20 @@ pub fn commitTransactions(
...
-    transaction: ResolvedTransaction,
-    tx_err: ?TransactionError,
-    maybe_log_collector: ?LogCollector,
-    is_vote: bool,
-) !void {
-    const log_collector = maybe_log_collector orelse return;
-    const log_messages = try cloneLogMessages(allocator, log_collector);
-    if (log_messages.len == 0) {
-        return;
-    }
```
prestonsn: Since we're producing all data (not just logs), I had to move these skips to the subscription consumer side (as opposed to here, replay producer side): https://github.com/Syndica/sig/pull/1344/commits/b9a2c39da74a3c95230176fbe81ed97fa830631f

cc @jbuckmccready

src/replay/Committer.zig
jbuckmccready: I don't see the change in that commit but it's changed in the PR, so resolving this.

src/ledger/transaction_status.zig
```diff
@@ -288,13 +295,15 @@ pub const TransactionStatusMetaBuilder = struct {
         if (count == 0) return &.{};
 
         const messages = try allocator.alloc([]const u8, count);
-        errdefer allocator.free(messages);
+        var i: usize = 0;
+        errdefer {
+            for (messages[0..i]) |m| allocator.free(m);
+            allocator.free(messages);
+        }
```
jbuckmccready: Another comment

src/replay/Committer.zig
```diff
@@ -459,48 +479,327 @@ fn writeTransactionStatus(
...
+        var have_data = false;
+
+        if (findInWrites(writes, pubkey)) |written| {
+            owner = written.account.owner;
+            if (written.account.data.len >= spl_token.TOKEN_ACCOUNT_SIZE) {
+                @memcpy(&data_buf, written.account.data[0..spl_token.TOKEN_ACCOUNT_SIZE]);
+                have_data = true;
+            }
+        } else if (account_store_reader) |reader| {
+            const account = reader.get(allocator, pubkey) catch continue orelse continue;
```
jbuckmccready: Throwing away error here seems awkward.

src/replay/Committer.zig
```diff
@@ -459,48 +479,327 @@ fn writeTransactionStatus(
...
+    for (transaction.accounts.items(.pubkey), 0..) |pubkey, idx| {
+        // Skip token program IDs themselves since they are programs and not token accounts.
+        // [agave] https://github.com/anza-xyz/agave/blob/v3.1.2/svm/src/transaction_balances.rs#L98
+        if (spl_token.isTokenProgram(pubkey)) continue;
+
+        // Read the account's final state. Check transaction writes
+        // first (updated/rolled-back accounts), then fall back to
+        // the account store (unmodified accounts).
+        // [agave] https://github.com/anza-xyz/agave/blob/v3.1.2/svm/src/account_loader.rs#L237-L255
+        var owner: Pubkey = undefined;
```
jbuckmccready: This should maybe be an optional rather than undefined so it's easy to sanity check it's been set before being used later.

src/replay/Committer.zig
```diff
@@ -826,18 +1130,211 @@ test "commitTransactions emits empty transaction logs batch when execution has n
...
+    // Regression test: for failed transactions, writes contain only
+    // rollback accounts (fee payer / nonce). Token accounts that
+    // were loaded but not modified must still appear in
+    // postTokenBalances, read from the account store.
+    //
+    // [agave] https://github.com/anza-xyz/agave/blob/v3.1.2/svm/src/transaction_balances.rs#L88
+    const allocator = std.testing.allocator;
+    const ids = sig.runtime.ids;
+
+    var prng = std.Random.DefaultPrng.init(42);
```
jbuckmccready: Should use `std.testing.random_seed`.

src/replay/Committer.zig
```diff
@@ -826,18 +1130,211 @@ test "commitTransactions emits empty transaction logs batch when execution has n
...
+    try std.testing.expect(
+        entry.program_id.equals(&token_program),
+    );
+}
+
+test "collectPostTokenBalances: success tx reads token account from writes" {
+    const allocator = std.testing.allocator;
+    const ids = sig.runtime.ids;
+
+    var prng = std.Random.DefaultPrng.init(43);
```
jbuckmccready: Should use `std.testing.random_seed`.

src/replay/freeze.zig
```diff
@@ -257,21 +280,31 @@ fn distributeTransactionFees(
...
+        _ = capitalization.fetchSub(burn, .monotonic);
+        return .{
+            .rewards = keyed_rewards,
+            .num_partitions = num_partitions,
+            .arena = arena,
+        };
     }
 
     _ = capitalization.fetchSub(burn, .monotonic);
+    return .empty();
```
jbuckmccready: Not part of your changes but maybe re-arrange so there's not a large nested block for `payout > 0`. E.g., `if payout == 0 { early return stuff... }`.

src/replay/service.zig
```diff
@@ -785,35 +785,62 @@ fn freezeCompletedSlots(state: *ReplayState, results: []const ReplayResult) !boo
...
-                    last_entry_hash,
-                    state.ledger,
-                ));
+
+                // Capture previous blockhash before freeze
+                // inserts the current one into the queue.
+                const prev_blockhash = blk: {
+                    var q = slot_info.state().blockhash_queue.read();
+                    defer q.unlock();
+                    break :blk q.get().last_hash orelse sig.core.Hash.ZEROES;
```
jbuckmccready: Previous blockhash should always be present at this point I think? I think it can just return error if it's not present as something is seriously wrong.

src/replay/service.zig
```diff
@@ -785,35 +785,62 @@ fn freezeCompletedSlots(state: *ReplayState, results: []const ReplayResult) !boo
...
+                            .root = slot_tracker.state_root.load(.monotonic),
+                            .accounts = accounts,
+                            .blockhash = last_entry_hash,
+                            .previous_blockhash = prev_blockhash,
+                            .block_height = slot_info.constants().block_height,
+                            .block_time = state.ledger.db.get(
+                                state.allocator,
+                                schema.blocktime,
+                                slot,
+                            ) catch null,
```
jbuckmccready: Is this the right thing to do if it errors here? Looks odd but maybe it's best we can do.

src/replay/service.zig
```diff
@@ -1973,6 +2000,12 @@ test "freezeCompletedSlots emits slot_frozen event with slot metadata" {
             try std.testing.expectEqual(1, slot_frozen.slot);
             try std.testing.expectEqual(0, slot_frozen.parent);
             try std.testing.expectEqual(0, slot_frozen.root);
+            // Block metadata enrichment (Step 5).
+            try std.testing.expectEqual(Hash.ZEROES, slot_frozen.blockhash);
```
jbuckmccready: Can we have it compare with actual value here? `Hash.ZEROES` is the fail case right now.

src/replay/service.zig
```diff
@@ -1973,6 +2000,12 @@ test "freezeCompletedSlots emits slot_frozen event with slot metadata" {
...
+            try std.testing.expect(slot_frozen.block_height != null);
+            // block_time is read from ledger (clock.unix_timestamp);
+            // the test ledger has no blocktime entry, so it's null.
+            try std.testing.expectEqual(null, slot_frozen.block_time);
```
jbuckmccready: Similar to the `Hash.ZEROES` assert.

src/rpc/jrpc_websockets/integration_tests/logs_subscription_tests.zig
```diff
@@ -52,6 +53,26 @@ fn signatureWithFill(fill: u8) Signature {
     return sig_value;
 }
 
+fn cloneTransactionError(
```
jbuckmccready: Bad rebase I think, this is a method now on `TransactionError`.

src/rpc/jrpc_websockets/Runtime.zig
```diff
@@ -346,6 +346,18 @@ fn handleInboundEvent(
                 );
             };
         },
+        .transaction_batch => |*batch_data| {
```
jbuckmccready, line 339: Bad rebase I think: this event is fine but the `.logs` event needs to be removed, this one replaces it.

src/rpc/jrpc_websockets/Runtime.zig
```diff
@@ -872,16 +960,18 @@ fn publishLogsSubscriptionForEntry(
     }
 }
 
-fn logsEventMatchesFilter(
+const TransactionBatchView = SlotStateCache.CachedSlot.TransactionBatchIterator.View;
```
jbuckmccready: Move to top of file.

src/rpc/jrpc_websockets/SlotStateCache.zig
```diff
@@ -75,21 +95,256 @@ pub const CachedSlot = struct {
...
+                self.entry_idx = 0;
+            }
+            return null;
+        }
+    };
+
+    /// Deep-copy all cached transaction and block data into `arena`, producing a standalone
+    /// `VersionedConfirmedBlock` that can be handed to a worker thread. Returns `null` when block
+    /// metadata is incomplete (not yet frozen).
+    pub fn buildConfirmedBlock(
```
jbuckmccready, line 145: Not part of this PR but making comment for future reference: we're just brute forcing this for every subscription, but we should be able to do something like lazy init and share across subscriptions by fetching the lazy init from the threadpool - first thread that accesses it does the build, or some variation of that.

src/rpc/jrpc_websockets/types.zig
```diff
@@ -370,22 +444,25 @@ pub const InboundEvent = union(enum) {
...
             .vote => |vote_data| {
                 var data = vote_data;
                 data.deinit();
             },
             .slot_frozen => |slot_data| {
-                var accounts = slot_data.accounts;
-                accounts.deinit();
+                var data = slot_data;
+                data.accounts.deinit();
+                data.distributed_rewards.deinit();
```
jbuckmccready: Maybe should be a deinit on the `SlotFrozenEvent`, this looks funky.

jbuckmccready (changes requested):

src/rpc/jrpc_websockets/Runtime.zig
prestonsn: https://github.com/Syndica/sig/pull/1344/commits/23cbf0dcb3eaa572a20bdd1e392b4ea1bb4937e6

src/replay/Committer.zig
prestonsn: https://github.com/Syndica/sig/pull/1344/commits/23cbf0dcb3eaa572a20bdd1e392b4ea1bb4937e6

src/replay/Committer.zig
prestonsn: https://github.com/Syndica/sig/pull/1344/commits/23cbf0dcb3eaa572a20bdd1e392b4ea1bb4937e6

src/rpc/jrpc_websockets/Runtime.zig
prestonsn: Oops, yeah there was a bunch of dead code left over: https://github.com/Syndica/sig/pull/1344/commits/4251a6426039f09b6278ef769f3aa6cc18e75b95

src/rpc/jrpc_websockets/integration_tests/logs_subscription_tests.zig
prestonsn: Yup: https://github.com/Syndica/sig/pull/1344/commits/4251a6426039f09b6278ef769f3aa6cc18e75b95

src/replay/Committer.zig
prestonsn: https://github.com/Syndica/sig/pull/1344/commits/4d9831a9cb7e7d08328c7a129616c23d1ddd64f8

src/replay/service.zig
prestonsn: https://github.com/Syndica/sig/pull/1344/commits/4d9831a9cb7e7d08328c7a129616c23d1ddd64f8

src/replay/service.zig
prestonsn: https://github.com/Syndica/sig/pull/1344/commits/4d9831a9cb7e7d08328c7a129616c23d1ddd64f8

I switched things over to use `try` instead since `db.get()` returns `!?i64`. This way we should be propagating actually db errors up, while still passing null for missing entries (shouldn't really happen, but matches agave behavior).

src/replay/Committer.zig
jbuckmccready: It should always be set right? I think it can just error rather than continue.

src/replay/freeze.zig
prestonsn: https://github.com/Syndica/sig/pull/1344/commits/a2ca43a2da9457186a510cd219ccccc3c65ddf3e

src/rpc/jrpc_websockets/types.zig
prestonsn: https://github.com/Syndica/sig/pull/1344/commits/a2ca43a2da9457186a510cd219ccccc3c65ddf3e

src/replay/service.zig
prestonsn: https://github.com/Syndica/sig/pull/1344/commits/872bbe5cff79ed43cea69c75015d2369c23cabf3

src/replay/service.zig
prestonsn: https://github.com/Syndica/sig/pull/1344/commits/872bbe5cff79ed43cea69c75015d2369c23cabf3

src/replay/Committer.zig
prestonsn: https://github.com/Syndica/sig/pull/1344/changes/4d9831a9cb7e7d08328c7a129616c23d1ddd64f8

src/replay/Committer.zig
prestonsn: right, doesn't make sense to `continue` here: https://github.com/Syndica/sig/pull/1344/commits/8508ed53d7a68baad1c03e961cf2c17b43121988

src/replay/Committer.zig
```diff
@@ -459,48 +479,328 @@ fn writeTransactionStatus(
...
+            &data_buf,
+        ) orelse continue;
+
+        result.append(.{
+            .account_index = @intCast(idx),
+            .mint = parsed.mint,
+            .owner = parsed.owner,
+            .amount = parsed.amount,
+            .program_id = owner_pubkey,
+        }) catch break; // BoundedArray full; should never happen.
```
Sobeston: Would it be safe to do catch unreachable here?

src/replay/freeze.zig
```diff
@@ -202,12 +203,34 @@ fn finalizeState(allocator: Allocator, params: FinalizeStateParams) !void {
...
+/// Rewards and partition info produced by fee distribution.
+/// Owns an arena that backs the `rewards` slice. The arena can be handed off
+/// to another thread (e.g. the websocket runtime) for deferred deallocation.
+pub const DistributedRewards = struct {
+    rewards: []const sig.ledger.meta.Reward = &.{},
+    num_partitions: ?u64 = null,
+    arena: std.heap.ArenaAllocator,
+
+    pub fn empty() DistributedRewards {
+        return .{ .arena = std.heap.ArenaAllocator.init(std.heap.page_allocator) };
```
Sobeston: Not a fan of this one, is this intended to be used like this?

src/replay/freeze.zig
```diff
@@ -221,57 +244,67 @@ fn distributeTransactionFees(
...
+        .reward_type = .fee,
+        .commission = null,
+    };
 
-        try ledger.db.put(sig.ledger.schema.schema.rewards, slot, .{
-            .rewards = keyed_rewards,
-            .num_partitions = num_partitions,
-        });
-    }
+    var arena = std.heap.ArenaAllocator.init(allocator);
```
Sobeston: more of a general comment (feel free to resolve this one), but we might be overdoing it on the arenas

src/rpc/jrpc_websockets/types.zig
```diff
@@ -545,6 +600,72 @@ pub const TransactionLogsEntry = struct {
...
+    slot: u64 = 0,
+    /// Transaction entries, arena-owned.
+    entries: []const TransactionEntry = &.{},
+    arena: std.heap.ArenaAllocator,
+
+    pub fn empty() SlotTransactionBatch {
+        return .{
+            .slot = 0,
+            .entries = &.{},
+            .arena = std.heap.ArenaAllocator.init(std.heap.page_allocator),
```
Sobeston: We should basically never use the page_allocator, even if wrapped. In part because it bypasses the instrumented allocations - this wouldn't be seen by tracy (which our GPA is instrumented with), and likely wouldn't be seen by leak detectors for C code either. This also makes the code less portable

src/rpc/jrpc_websockets/types.zig
```diff
@@ -545,6 +600,72 @@ pub const TransactionLogsEntry = struct {
...
 
+/// A batch of full transaction metadata for a single slot, streamed from the Committer at commit time. Carries
+/// everything needed to build block, log, and future transaction notifications from cache (no ledger read).
+/// Owns all referenced data through an arena allocator.
+pub const SlotTransactionBatch = struct {
+    /// Slot this batch belongs to.
+    slot: u64 = 0,
+    /// Transaction entries, arena-owned.
+    entries: []const TransactionEntry = &.{},
+    arena: std.heap.ArenaAllocator,
```
Sobeston: If we are going to store arenas, we can store the arena allocator state, which is smaller as it doesn't contain the allocator used by the arena

Sobeston (changes requested):

src/replay/Committer.zig
prestonsn: I think you're right. The original code (prior to changes in this PR) looked like this: 
https://github.com/Syndica/sig/blob/463f4ee079b14bbb2daf34e3e1fdd40d162447c2/src/replay/Committer.zig#L490-L498

Updated to use `unreachable`: https://github.com/Syndica/sig/pull/1344/commits/54521186605d8cbfebafcd13f4591c1ac3f541e6

src/replay/freeze.zig
prestonsn: Swapped over to storing `ArenaAllocator.State {}` so that drops the need for a sentinel value, eliminates this entirely: https://github.com/Syndica/sig/pull/1344/commits/5efb812cbd9d7e33b933275c4ad04a434b8074bd

src/replay/freeze.zig
prestonsn: Yeah, mostly sticking with established pattern for the websocket jrpc stuff. Likely not what we're going to have long term once everything is just streamed out into an RPC tile/service in v2.

src/rpc/jrpc_websockets/types.zig
prestonsn: I think this is fair. However as I'm working through the changes for this I'm noticing that this weakens the "correct allocator" contract a bit. Storing the `ArenaAllocator` does have the advantage of preventing users calling `deinit()` from passing in the wrong backing allocator. 

But this doesn't really change any of the original assumptions. All this code requires things to use the same thread-safe allocator anyways. Getting chopped in v2 :saluting_face: 

https://github.com/Syndica/sig/pull/1344/commits/e5d3ba0e921b79229b5d418d2a642f2a9faa9954

src/rpc/jrpc_websockets/types.zig
prestonsn: https://github.com/Syndica/sig/pull/1344/commits/e5d3ba0e921b79229b5d418d2a642f2a9faa9954

src/replay/Committer.zig
```diff
@@ -77,15 +77,11 @@ pub fn commitTransactions(
 
     // When an event sink is present, build a rich transaction batch event.
     // The arena owns all TransactionEntry data.
-    var maybe_batch_event: ?jrpc_types.SlotTransactionBatch = null;
-    defer if (maybe_batch_event) |*ev| ev.deinit();
+    const maybe_event_sink = self.event_sink;
+    var owns_batch_arena = maybe_event_sink != null;
+    var batch_arena = std.heap.ArenaAllocator.init(persistent_allocator);
```
jbuckmccready: I think this needs to use the event sink allocator, that's how it deallocates it on the receive side.

jbuckmccready: This was bug I had going back to logsSubscribe PR, works out OK currently since it's the same global allocator but the pattern for all the events should be to use the event sink allocator (which is the channel's allocator). We should probably double check this across all the event types.

jbuckmccready (dismissed):

src/replay/Committer.zig
prestonsn: Good catch, fixed the immediate ones and evaluated the other events. I don't see issues with those (would love to move away from this pattern in v2 in general, managing all the allocators is a bit hairy): https://github.com/Syndica/sig/pull/1344/commits/e02248a3fa3d754d42ab0605c9c5b7148d85a44d

I also found another spot where we were allocating into replay's allocator, but freeing on channel's (only affects for the `DistributedRewards` type: https://github.com/Syndica/sig/pull/1344/commits/ba699748c328ca827ba7dba710808fdf64d7626e)

src/replay/service.zig
```diff
@@ -794,26 +807,51 @@ fn freezeCompletedSlots(state: *ReplayState, results: []const ReplayResult) !boo
...
-                        .accounts = accounts,
-                    } });
+                    errdefer accounts.deinit(event_sink.allocator());
+
+                    // Transfer ownership: rewards were already allocated on
+                    // event_sink.allocator() via rewards_allocator, so we
+                    // can hand them off directly.
+                    const event_distributed = distributed;
+                    distributed = .empty();
+
```
Sobeston, line 838: We have a defer deinit on `distributed`, but in this path we set `distributed` to .empty, which would cause the deinit to do nothing. Is this a potential leak?

src/replay/Committer.zig
```diff
@@ -459,48 +475,328 @@ fn writeTransactionStatus(
...
+                @memcpy(&data_buf, written.account.data[0..spl_token.TOKEN_ACCOUNT_SIZE]);
+                have_data = true;
+            }
+        } else if (account_store_reader) |reader| {
+            const account = try reader.get(allocator, pubkey) orelse continue;
+            defer account.deinit(allocator);
+            owner = account.owner;
+            if (account.data.len() >= spl_token.TOKEN_ACCOUNT_SIZE) {
+                _ = account.data.read(0, &data_buf);
+                have_data = true;
```
Sobeston, line 762: Could assert that you read TOKEN_ACCOUNT_SIZE bytes

src/replay/service.zig
prestonsn: True, I think what I added was a leak on the `send` error path since `distributed` gets set to `.empty()` before the `try EventSink.send()`

Fixed by using (what I previously had) an `errdefer .deinit()`:  https://github.com/Syndica/sig/pull/1344/commits/bbba4b2632dcecc06f55d07fc233f9d1aaec9609

src/replay/Committer.zig
prestonsn: https://github.com/Syndica/sig/pull/1344/commits/bb24406d03f219dfb39ac89cebb3062182ef9e7a

jbuckmccready (approved):

Sobeston (approved):

PR #1345, opened 2026-03-25
feat(jrpc-ws): implement signatureSubscribe method

jbuckmccready:
## Summary

Implements WebSocket `signatureSubscribe`:

- `receivedSignature` notifications from shred ingestion
- final `signatureNotification` delivery based on commitment/status cache
- auto-unsubscribe after final delivery
- ordering guarantees so received notifications are emitted before final ones

## Main Changes

- Added `signatureSubscribe` request/config parsing, subscription keying, and `signatureNotification` serialization.
- Extended the WS runtime to:
  - react to received-signature events
  - look up landed signatures via `StatusCache` across blockhashes
  - enqueue a terminal/final notification for processed/confirmed/finalized commitments
- Updated `NotifQueue` / handler flow to support sticky final notifications and cleanup after final delivery.
- Emitted received-signature events from `ShredReceiver` when completed data sets are inserted, and plumbed the event sink through startup/service wiring.
- Extended `StatusCache` to retain tx errors and support commitment-agnostic lookup via `getOwnedStatusAnyBlockhash`.
- Moved transaction error deep-copy logic into `TransactionError.clone()` and reused it where status/log data is copied.

## Adjacent Changes

- Fix off by one bugs in `ShredReceiver.zig` (only affects code path for new signature events, which is why it hasn't been detected/an issue up until now).
- Add doc comments clarifying exclusive end ranges in ledger `Reader.zig`.

## Testing

- Added/updated unit tests and new `signatureSubscribe` integration tests.
- Tested on testnet with all different parameters.


src/core/status_cache.zig
```diff
@@ -136,8 +188,11 @@ pub const StatusCache = struct {
 
         const lookup_key: [CACHED_KEY_SIZE]u8 = key[key_index..][0..CACHED_KEY_SIZE].*;
 
+        const owned_err = if (maybe_err) |err| try err.clone(allocator) else null;
+        errdefer if (owned_err) |err| err.deinit(allocator);
+
```
prestonsn: I think it's a bit subtle, but is there a double-free path for `owned_err` here?

if this line fails: https://github.com/Syndica/sig/blob/27836e032923dd162943acdad14070f48803293c/src/core/status_cache.zig#L198

or if this line fails: https://github.com/Syndica/sig/blob/27836e032923dd162943acdad14070f48803293c/src/core/status_cache.zig#L207

Then, `errdefer` frees `owned_err`, but it's potentially still in the `forks` list:

https://github.com/Syndica/sig/blob/27836e032923dd162943acdad14070f48803293c/src/core/status_cache.zig#L195

(I think the append copies the struct by value, so they share same heap ptr?)

Very unlikely path anyways, i think only impacts `.BorshIoError` variant (plus needs the inserts/appends to hit OOMs)

prestonsn (commented):
Just one minor error path to consider in status_cache's insert(), looks good otherwise 🫡

src/core/status_cache.zig
jbuckmccready: https://github.com/Syndica/sig/pull/1345/changes/f54ce6941abca58e9bb4b1697a21bc00c2cb9489

prestonsn (dismissed):

InKryption (dismissed):

prestonsn (approved):

dnut (approved):

PR #1366, opened 2026-04-06
fix(accountsdb): add dedicated RPC reader

ajw221:
### Intent

- Fix accounts db contention caused by RPC read locks being acquired on heavy methods like `getSupply`

### Implementation

- Added optional `reader_rooted` field to `accountsdb.Db` to be used as a separate connection in queries if RPC is enabled
- Updated `accountsdb.Db` queries to conditionally use new `reader_rooted` field if available
- Added `enableWalMode` to `accountsdb.Rooted` for enabling WAL mode when RPC is enabled and WAL is chosen to not be disabled

### Ramifications

- Currently any `getSupply` request greatly slows down the validator to 1+ second slot times until the request has completed, which could hold the validator's slot times hostage for as long as the request is inflight. This could also be the case for any and all other RPC methods also using the same reading pattern as `getSupply`.

### Tests

- Tested `getSupply` with and without these changes to ensure there is no loss in slot times when a heavy accounts db RPC request is made


yewman (dismissed):

Sobeston (dismissed):

InKryption (dismissed):

src/cmd.zig
```diff
@@ -686,6 +687,16 @@ const Cmd = struct {
                     "at the cost of increased storage and slower writes. When disabled, " ++
                     "falls back to a full table scan - default: false",
             },
+            .disable_wal_mode = .{
```
kprotty: Now that concurrent reader db connections are the default, should WAL mode also be default / removed as a CLI setting?

ajw221: [904f4d8](https://github.com/Syndica/sig/pull/1366/commits/904f4d88dacb4e12821decce82c8b6ad85c3b934)

kprotty (approved):

yewman (approved):

PR #1367, opened 2026-04-06
fix(vote_listener): improve `trackNewVotesAndNotifyConfirmations`

InKryption:
Bring implementation up to par with agave's by removing the `accumulate_intermediate_votes` condition & related code.

Also amend a discrepancy where 0 was used as a default stake value for epochs we had no stake information on, where agave simply skipped over those epochs.

Inspired by https://github.com/anza-xyz/agave/pull/10594

### DIFF
--- a/src/consensus/vote_listener.zig
+++ b/src/consensus/vote_listener.zig
@@ -21,8 +21,6 @@ const OptimisticConfirmationVerifier =
 const SlotTracker = sig.replay.trackers.SlotTracker;
 const EpochTracker = sig.core.EpochTracker;
 
-const VoteEventData = sig.rpc.jrpc_websockets.types.VoteEventData;
-
 const Logger = sig.trace.Logger("vote_listener");
 
 pub const SlotDataProvider = struct {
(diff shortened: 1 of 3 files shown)


Sobeston:
Failing style checks

src/consensus/vote_listener.zig
```diff
@@ -962,20 +920,23 @@ fn trackNewVotesAndNotifyConfirmations(
...
+            };
+            errdefer allocator.free(vote_slots);
+
+            try event_sink.send(.{ .vote = .{
+                .vote_pubkey = vote_pubkey,
+                .slots = vote_slots,
+                .hash = last_vote_hash,
+                .timestamp = vote.timestamp(),
+                .signature = vote_transaction_signature,
+            } });
```
prestonsn, line 933: This could be due to rebase, but it looks like the constructed `VoteEventData` struct is now changed to be constructed by allocating `vote_slots` with `std.mem.Allocator` instead of how it was before: using the channel (`event_sink.allocator()`) via calling `VoteEventData.initOwned`: https://github.com/Syndica/sig/blob/5c0cad5b8731cd181ef840fbe7e2c0a4641fa521/src/rpc/jrpc_websockets/types.zig#L946-L962

src/consensus/vote_listener.zig
```diff
@@ -805,122 +872,15 @@ fn trackNewVotesAndNotifyConfirmations(
...
-        const slot = vote_slots[rev_index];
-        if (slot <= root) continue;
+    const slot_count = vote.slotCount();
+    // Track all vote slots for propagated check (iterates from most recent to oldest)
+    for (0 + 1..slot_count + 1) |fwd_index| {
+        const rev_index = slot_count - fwd_index;
+        const slot = vote.getSlot(rev_index);
 
         // if we don't have stake information, ignore it
         // Pull stakes via slot_data_provider helpers
```
prestonsn: Should these be moved to where relevant? (maybe artifact of github diff I'm looking at in web UI)

src/consensus/vote_listener.zig
```diff
@@ -796,6 +796,73 @@ fn trackNewVotesAndNotifyConfirmations(
...
 
+    const is_new_vote: bool = if (last_vote_slot <= root) false else track_last_vote_slot: {
+        // Track the last vote slot for optimistic confirmation
+        const epoch = slot_data_provider.getSlotEpoch(last_vote_slot);
+
+        const total_stake = slot_data_provider.getEpochTotalStake(epoch) orelse {
+            // If we don't have stake information, ignore it
+            break :track_last_vote_slot false;
+        };
+        const stake = slot_data_provider.getDelegatedStake(last_vote_slot, vote_pubkey).?; // see above
```
prestonsn, line 805: This makes sense as is, so feel free to ignore. But is this preferable?
```suggestion
        const stake = slot_data_provider.getDelegatedStake(last_vote_slot, vote_pubkey) orelse {
            break :track_last_vote_slot false;
        };
```

src/consensus/vote_listener.zig
prestonsn: It's likely all just fine since we're using the std allocator throughout, just breaks semantics a bit since we had it use the channel allocator intentionally to ensure deallocation on receiver thread uses the same allocator.

src/consensus/vote_listener.zig
```diff
@@ -805,122 +872,15 @@ fn trackNewVotesAndNotifyConfirmations(
...
-            break :blk false;
-        }
-    };
-
-    var is_new_vote = false;
-    // If slot is before the root, ignore it. Iterate from latest vote slot to earliest.
-    for (0 + 1..vote_slots.len + 1) |fwd_index| {
-        const rev_index = vote_slots.len - fwd_index;
-        const slot = vote_slots[rev_index];
-        if (slot <= root) continue;
```
prestonsn, line 835: it looks like this check `if (slot <= root) continue;` gets dropped with changes, is this intended?

Sobeston (commented):
happy to approve when preston's comments have been addressed

src/consensus/vote_listener.zig
InKryption: If the second one returns null despite the former returning non-null, that would indicate something is very wrong, so we would want it to panic.

src/consensus/vote_listener.zig
InKryption: Good catch
b076128e2a0971ba5f6dacc1405c1f6390182040

src/consensus/vote_listener.zig
InKryption: Just removed it, the comment no longer makes sense
b076128e2a0971ba5f6dacc1405c1f6390182040

src/consensus/vote_listener.zig
InKryption: b076128e2a0971ba5f6dacc1405c1f6390182040

prestonsn (dismissed):
LGTM 🫡

Sobeston (dismissed):

prestonsn (approved):

Sobeston (approved):

PR #1393, opened 2026-04-17
feat(v2): rooted db 

### LINKED ISSUE #1162: perf(accountsdb): new rooted store
replace sqlite with something else. king tried several mmap based approaches which seems like a dead end. currently working on buffer pool based approach.

### LINKED ISSUE #1386: feat(v2): atomic/shared mem pool
We already have a mem pool, implemented in https://github.com/Syndica/sig/pull/1339. 

It's very important for performance that we have a mem pool that can be shared between services.

kprotty:
Implements:
* Loading Manifest, StatusCache, and Accounts from a `snapshot.tar.zst` fetched from the snapshot-service.
* Creating an on-disk (rooted) KV database file for the accounts using the loaded snapshot.
* Database loads from an existing db file instead of a snapshot if found.
* Database can recover from partial put() queries (`beginTransaction`, `commitTransaction`).
* Pool of in-memory accounts (unrooted) which the database can populate from get() queries.
* Ring buffer for replay-service to accounts-db-service to do get() requests and return pool indexes.

Additionally:
* Changes shared-region mapping to use `MAP_POPULATE` after init, during `resolveArgs` for a given service:
  * Rooted DB uses a large amount of memory (8GB testnet, 64GB mainnet) as a hashmap to do random accesses for put/get. During snapshot loading, each access is almost a guaranteed cache-miss. The hashmap optimizes for such, but not having MAP_POPULATE means each miss is also a page-fault, slowing down snapshot loading by 2x.
* Fix replay not using telemetry:
  * Originally to address https://github.com/Syndica/sig/pull/1469. But given that's now merged, it just changes replay's `std.log.info` to use the proper `logger`
* Change `lib.clock` to use `posix.clock_gettime` to avoid linux syscall:
  * Potentially addressed by https://github.com/Syndica/sig/pull/1473
* `zig build lint` not working on windows
  `zig build -Dtarget=x86_64-native` to cross-compile should ideally support running on all target hosts


v2/lib/accounts_db.zig
```diff
@@ -0,0 +1,41 @@
...
+
+pub const AccountLookups = extern struct {
+    in: lib.ipc.Ring(256, Request),
+    out: lib.ipc.Ring(256, Result),
+
+    pub const Request = Pubkey;
+    pub const Result = extern struct {
+        pubkey: Pubkey,
+        account_index: AccountPool.Index, // .invalid_index if not found
+    };
```
prestonsn, line 34: is it intentional that there is no id/gen attached to request for lookups (only pubkey)? 

if multiple lookups for the same pubkey are outstanding to the caller, then caller could treat both requests as interchangeable I think.  But maybe this is intentional since this case doesn't matter for replay/enforced against in some way. In which case could this be documented?

v2/lib/accounts_db/rooted.zig
```diff
@@ -0,0 +1,805 @@
...
+    lookup_nodes: [max_active_lookups]LookupNode,
+
+    journal: Journal,
+    io: union {
+        reader: FileReader(.{ .buffer_size = buffer_size, .block_size = block_size }),
+        writer: FileWriter(.{ .buffer_size = buffer_size, .block_size = block_size }),
+    },
+
+    const buffer_size = 32 * 1024 * 1024;
+    const block_size = 128 * 1024;
```
prestonsn: very nit: could use the new `MemorySize` here?

https://github.com/Syndica/sig/blob/ef4ba033a9038f3239301b5f7b5cfd5d7c38c31e/v2/init/main.zig#L64

v2/lib/accounts_db/pool.zig
```diff
@@ -0,0 +1,223 @@
...
+        64,
+        256,
+        1024,
+        4096,
+        16 * 1024,
+        64 * 1024,
+        1 * 1024 * 1024,
+        4 * 1024 * 1024,
+        10 * 1024 * 1024,
+    };
```
prestonsn, line 104: also very nit: can we use `MemorySize` here?

https://github.com/Syndica/sig/blob/ef4ba033a9038f3239301b5f7b5cfd5d7c38c31e/v2/init/main.zig#L64

v2/lib/accounts_db.zig
kprotty: If replay needs metadata to distinguish it, we could adjust it for that here & in Rooted (LookupNode & LookupResult). But ATM, replay doesnt even have a lookup path.

v2/lib/clock.zig
```diff
@@ -32,13 +32,9 @@ pub fn wallclock(comptime unit: Unit) u64 {
...
-    var ts: linux.timespec = undefined;
-    const ret = linux.clock_gettime(clock_id, &ts);
-
-    switch (linux.E.init(ret)) {
-        .SUCCESS => {},
-        else => |err| std.debug.panic("clock_gettime failed: {}", .{err}),
-    }
+    const ts = std.posix.clock_gettime(clock_id) catch |err| {
+        std.debug.panic("clock_gettime failed: {}", .{err});
+    };
```
Sobeston, line 43: Let's avoid libc where possible, but this should have been solved by https://github.com/Syndica/sig/pull/1473

v2/lib/accounts_db/pool.zig
```diff
@@ -0,0 +1,223 @@
...
+                    while (self.locked.load(.monotonic) != 0) std.atomic.spinLoopHint();
+                    if (self.locked.fetchOr(1, .acquire) & 1 == 0) break;
+                }
+            }
+        }
+
+        fn release(self: *Lock, _: *Holder) void {
+            self.locked.store(0, .release);
+        }
+    };
```
Sobeston: is there a reason for this specific lock implementation here?

v2/lib/accounts_db/table.zig
```diff
@@ -0,0 +1,207 @@
...
+            0x1cad21f72c81017c,
+            0xdb979083e96dd4de,
+            0x1f67b3b7a4a44072,
+        };
+        const seed = self.seed;
+        const in: [4]u64 = @bitCast(pubkey.data);
+        const a: [2]u64 = @bitCast(@as(u128, in[0] ^ (s[0] +% seed)) * (in[1] ^ (s[1] -% seed)));
+        const b: [2]u64 = @bitCast(@as(u128, in[2] ^ (s[2] +% seed)) * (in[3] ^ (s[3] -% seed)));
+        return @bitCast([2]u64{ a[0] ^ a[1], b[0] ^ b[1] });
+    }
```
Sobeston: I guess 2 questions
- would it be good to expose this code for reuse? 
- wondering why it returns a u128, this is quite big for a hash, but maybe it is reasonable for such a large set of keys

Sobeston (commented):
looking good so far, still working my way through it

v2/lib/accounts_db/rooted.zig
```diff
@@ -0,0 +1,805 @@
...
+            try self.io.reader.advance(block_size);
+
+            // validate journal
+            if (self.journal.magic != .valid) {
+                logger.err().logf("invalid journal magic - expected:{x}, found:{x}", .{
+                    @intFromEnum(@TypeOf(self.journal.magic).valid),
+                    @intFromEnum(self.journal.magic),
+                });
+                return;
+            }
```
prestonsn, line 213: Do we want to return `error.InvalidJournal` (or some other concrete error) here instead of just return?

v2/init/main.zig
```diff
@@ -53,6 +55,29 @@ const Config = struct {
         known_validators: []const []const u8,
     };
 
+    const AccountsDb = struct {
+        file: []const u8,
+        rooted: MemorySize,
+        unrooted: MemorySize,
+
+        // For nicer initialization of constants (instead of x * 1024 * !024 * 1024)
```
prestonsn:
```suggestion

        // For nicer initialization of constants (instead of x * 1024 * 1024 * 1024)
```

v2/services/accounts_db.zig
```diff
@@ -0,0 +1,121 @@
...
+            if (try rooted.queueRead(.from(logger), pubkey)) {
+                replay_in.markUsed();
+            }
+        }
+        if (replay_out.peek()) |result| {
+            if (try rooted.pollRead(.from(logger))) |res| {
+                result.* = res;
+                replay_out.markUsed();
+            }
+        }
```
prestonsn, line 131: I'm not 100% sure, but from what I understand from the ring impl: `peek()` does not advance pos, nor does `markUsed()`. Are we missing calling `.next()`?

as per the doc comment on `markUsed()`:

https://github.com/Syndica/sig/blob/ef4ba033a9038f3239301b5f7b5cfd5d7c38c31e/v2/lib/ipc/ring.zig#L75-L82

prestonsn (commented):
This is pretty sick! Still working my way through it all

v2/lib/accounts_db/pool.zig
kprotty: Was experimenting with a queuelock style impl. Having a holder type allowed for that.

Switched to a simple bool just to make sure it worked for now. Are you thinkin of moving this to its own file? IIRC pool should be the only place where we have to resort to a blocking lock.

v2/lib/accounts_db/table.zig
kprotty: * Probably good to move it into `Pubkey` (taking the seed as a parameter)
* Felt u64 would be too conservative in case of collisions, especially since this is not a cryptographic hash

v2/lib/accounts_db/pool.zig
```diff
@@ -0,0 +1,223 @@
...
+            std.math.divCeil(Index, @sizeOf(Account) + sc_data_size, index_scale) catch unreachable;
+
+        // try bump from our own bucket's stolen block, if any
+        if (bucket.stolen.block != invalid_index) {
+            idx = bucket.stolen.block + bucket.stolen.offset;
+            if (idx + idx_bump < bucket.stolen.len) {
+                bucket.stolen.offset += idx_bump;
+                return idx;
+            }
+        }
```
prestonsn: I'm a little confused on this bit. Here's my understanding of what's going on: 

```
idx_bump = the size of the allocation we want (in 8-byte units, but everything here operating on that scale so not too important)

bucket.stolen.block = global pool index where stolen block starts
bucket.stolen.offset = local offset within stolen block; end-of-occupied/start-of-free 
                       after bucket.stolen.block (offset to free region in stolen block)
bucket.stolen.len = total local length of the stolen block (max size for this specific stolen block)
```

So, from the code then:

```zig
idx = bucket.stolen.block + bucket.stolen.offset;
```

means

```
idx = global pool index where free region inside stolen block begins
```

So then the `if` condition here means:

```zig
 if (idx + idx_bump < bucket.stolen.len) { ... }
```

```
if (the globl pool index for start of free region in stolen block) + (alloc size we want) < (bucket's size) { ... }
```

So it seems like we're comparing two different things here. Seems to be comparing a global pool index against a local stolen block's length (not an index)?

prestonsn: I guess maybe this was supposed to check purely based on stolen block's local coords instead? something like this (no longer mixes global and local indices):

```zig
if (bucket.stolen.offset + idx_bum <= bucket.stolen.len) { ... }
```

running through a simple example:

```
idx_bump = 4
bucket.stolen.blockk = 1000 
bucket.stolen.offset = 4 
bucket.stolen.len = 20 
```

The allocation should fit since: `bucket.stolen.offset + idx_bump <= bucket.stolen.len`
 => `4 + 4 <= 20`

But with the current condition in code: `idx + idx_bump < bucket.stolen.len` 
=> `(bucket.stolen.block + bucket.stolen.offset) + idx_bump < bucket.stolen.len` 
=> `(1000 + 4) + 4 < 20` 
=> `1008 < 20` 

So I think net effect is we would reject allocation even though there is room in stolen block?

v2/lib/accounts_db/pool.zig
```diff
@@ -0,0 +1,223 @@
...
+            const target = &self.buckets[target_sc_idx];
+
+            // try steal from its own stolen bump allocating region
+            if (target.stolen.block != invalid_index) {
+                idx = target.stolen.block + target.stolen.offset;
+                if (idx + idx_bump < target.stolen.len) {
+                    target.stolen.offset += idx_bump;
+                    return idx;
+                }
+            }
```
prestonsn: Similar concern as here: https://github.com/Syndica/sig/pull/1393/changes#r3320035025

v2/lib/accounts_db/pool.zig
```diff
@@ -0,0 +1,223 @@
...
+                    .block = idx,
+                    .offset = idx_bump,
+                    .len = std.math.divCeil(
+                        Index,
+                        @sizeOf(Account) + target_sc_data_size,
+                        index_scale,
+                    ) catch unreachable,
+                };
+                return idx;
+            }
```
prestonsn: I get that this trying to steal from another size class's free list. my understanding here is that `target.free_list` points to a free block (for the `target` size class). 

So setting `bucket.stolen` like below is basically marking that blck as not being used/stolen by `bucket`: 

https://github.com/Syndica/sig/blob/698dae51327c99907538c5a5cda331f638868e82/v2/lib/accounts_db/pool.zig#L183-L191

But I don't see/understand where we actually remove `idx` from `target.free_list`? 

So should this path also pop from `target.fre_list` first, similar to how it's done in the happy-path: 

https://github.com/Syndica/sig/blob/698dae51327c99907538c5a5cda331f638868e82/v2/lib/accounts_db/pool.zig#L137-L144

It seems like the same block can remain in `target.free_list` despite being set to some other `bucket.stolen.block`

prestonsn (commented):
Reviewed more of the pool implementation itself. Looks great, but I have a couple questions

v2/lib/accounts_db.zig
prestonsn: Fairs. can easily be added later if we need it

v2/lib/accounts_db/pool.zig
kprotty: `.len` is in 8-byte units. The check there is to make sure the allocation would fit in that stolen block range, so it should indeed be `<=` instead of `<`

v2/lib/accounts_db/pool.zig
kprotty: yup, this needs to update `target.free_list` once stealing it.

v2/lib/accounts_db/pool.zig
prestonsn: ye true on the `<=` over the `<`, but main confusion is a little separate from that. 

Even with <=, it is still comparing `idx + idx_bump <= bucket.stolen.len`

but idx is an index to global mem pool while bucket.stolen.len is the length of the stolen block

Unless I'm mistaken and bucket.stolen.len is actually the end index (into global pool) for where the stolen block region ends.

v2/lib/accounts_db/rooted.zig
kprotty: `block_size` is checked directly in Rooted for a few places (Journal & padding sector alignment). This would turn those into `block_size.toBytes()` which feels more awkward.

v2/lib/accounts_db/pool.zig
Sobeston: Seems fine, maybe just add a comment on why this is here / why you're not using some std lock

v2/lib/accounts_db/table.zig
Sobeston: You can move it into Pubkey if you think that's appropriate? Would be nice to get some reasoning commented

v2/lib/accounts_db/pool.zig
```diff
@@ -0,0 +1,223 @@
...
+    pub const Account = extern struct {
+        ref_count: std.atomic.Value(u32),
+        pubkey: Pubkey,
+        owner: Pubkey,
+        lamports: u64,
+        rent_epoch: Epoch,
+        data: packed struct(u32) {
+            executable: bool,
+            len: u31,
+        } align(1),
```
Sobeston: outdated comment? It says all fields align(1) but they're not all set as such

v2/lib/accounts_db.zig
```diff
@@ -0,0 +1,40 @@
...
+pub const AccountPool = @import("accounts_db/pool.zig").AccountPool;
+pub const Rooted = @import("accounts_db/rooted.zig").Rooted;
+pub const Table = @import("accounts_db/table.zig").Table;
+
+pub const RootedConfig = extern struct {
+    file_len: u32,
+    file_path: [std.fs.max_path_bytes]u8,
+
+    memory_len: usize,
+    memory: [0]u8, // VLA [0..memory_len]
```
Sobeston: More of a passing thought than a review comment, but we will be using this pattern a bunch, would be interesting to have a safer generic way of representing VLAs

v2/lib/accounts_db/table.zig
```diff
@@ -0,0 +1,207 @@
...
+const tracy = @import("tracy");
+
+const Pubkey = lib.solana.Pubkey;
+const Slot = lib.solana.Slot;
+
+const FastDiv = lib.util.FastDiv;
+
+pub const Table = struct {
+    seed: u64, // runtime-provided hashing seed (if wanting to help lower collision attacks)
+    count: u32, // can represent 4-billion accounts (would need re-design after that)
```
Sobeston: Might be a good idea to make these kinds of limitations more visible? Because by the time we're approaching 4 billion accounts on the network, this may be forgotten

v2/lib/accounts_db/table.zig
```diff
@@ -0,0 +1,207 @@
...
+            self.count += @intFromBool(new_write); // bump for empty inserts
+
+            const ptr = if (newer_slot) &ptrs[i][0] else &stub;
+            ptr.* = batch.entries[i];
+        }
+    }
+
+    pub fn get(self: *const Table, pubkey: *const Pubkey) Value {
+        const hash = self.hashPubkey(pubkey);
+        std.debug.assert(hash != 0);
```
Sobeston: I see this assert in a few places, would it make more sense inside the hash function?

v2/lib/fio/reader.zig
```diff
@@ -0,0 +1,231 @@
...
+                    continue;
+                },
+            };
+            sqe.prep_read(
+                self.file.handle,
+                self.buffer[@as(u64, data.block_idx) * block_size + data.read ..][0..block_size],
+                @as(u64, data.disk_block) * block_size + data.read,
+            );
+            sqe.user_data = @bitCast(data);
+            _ = try self.ring.submit(); // SQPOLL makes this fast if done frequently enough
```
Sobeston, line 187: image

With SQPOLL, did you `THINK_LONG_AND_HARD_ABOUT_THIS_NO_REALLY`

v2/services/accounts_db.zig
```diff
@@ -0,0 +1,223 @@
...
+    }
+}
+
+
+// const features = lib.solana.features;
```
Sobeston: dead code

v2/lib/accounts_db/rooted.zig
kprotty: https://github.com/Syndica/sig/pull/1393/commits/04c9afff44ebf45708d65a4e6ed45a920d6f35a9

v2/lib/accounts_db/pool.zig
kprotty: https://github.com/Syndica/sig/pull/1393/commits/04c9afff44ebf45708d65a4e6ed45a920d6f35a9 (stealing index fix addressed later)

v2/services/accounts_db.zig
kprotty: https://github.com/Syndica/sig/pull/1393/commits/a4ce8026bd4ab33d1feb6e0aebf78db0a6717f33

v2/init/main.zig
kprotty: https://github.com/Syndica/sig/pull/1393/commits/a4ce8026bd4ab33d1feb6e0aebf78db0a6717f33

v2/services/accounts_db.zig
kprotty: this was an unintended merge of https://github.com/Syndica/sig/pull/1487

Will keep the code in this PR, but clean it up proper

v2/lib/accounts_db/table.zig
kprotty: its a sanity check thats used for the algs after (put & get use different algs). 

Enforced it at runtime by sacrificing 1-bit of entropy  https://github.com/Syndica/sig/pull/1393/commits/e93aa69383a9cda8f177967bc15a5bee1c2cabb5

v2/lib/accounts_db/table.zig
kprotty: Added an assert after incrementing `count` in the put path: https://github.com/Syndica/sig/pull/1393/commits/e93aa69383a9cda8f177967bc15a5bee1c2cabb5

v2/lib/accounts_db/pool.zig
kprotty: https://github.com/Syndica/sig/pull/1393/commits/e93aa69383a9cda8f177967bc15a5bee1c2cabb5

v2/lib/accounts_db/pool.zig
kprotty: changed to `start/end` ranges instead: https://github.com/Syndica/sig/pull/1393/commits/e93aa69383a9cda8f177967bc15a5bee1c2cabb5

v2/lib/accounts_db/table.zig
kprotty: Decided not to move into Pubkey since hashing usually wants u64 but this is u128.

Referred back to the corresponding comment: https://github.com/Syndica/sig/pull/1393/commits/e93aa69383a9cda8f177967bc15a5bee1c2cabb5

v2/lib/accounts_db/pool.zig
kprotty: https://github.com/Syndica/sig/pull/1393/commits/e93aa69383a9cda8f177967bc15a5bee1c2cabb5

v2/lib/fio/reader.zig
kprotty: Thought for 2h14m

It's not a bug — It's an intended feature.

The reading of snapshot-data / existing db-file-data should happen concurrently to the processing (decompression, table puts) of already read data. This can be achieved through dedicated services/threads doing those over ring buffers, or through io_uring doing it concurrently. 

With such a design, every new concurrently block being read-from / written-to disk needs a ring submission (`io_uring_enter` syscall) to get started while we do CPU work. Proper use of io_uring is to batch up the submissions for the same syscall, but this increases latency as it means the IO can't be started until enough has been accumulated. 

The current approach with SQPOLL instead starts the IO immediately after queueing `(FileReader/FileWriter).submit` and relies on the sq thread being awake to immediately process the submission, avoiding the enter syscall for each `block_size` of a submission. The use of `ring.copy_cqes(0)` in the completion poll() path also avoids an enter syscall by spinning on the CQE ring instead of blocking.

v2/lib/accounts_db/pool.zig
```diff
@@ -0,0 +1,234 @@
...
+            return idx;
+        }
+
+        fn bumpStolenBlock(self: *Bucket, idx_bump: Index) ?Index {
+            const idx = self.stolen.start;
+            if (idx + idx_bump > self.stolen.end) return null;
+
+            self.stolen.start = idx + idx_bump;
+            return idx;
+        }
```
prestonsn, line 49: These changes look really good, so much cleaner to read too. 

My only concern around `bumpStolenBlock` usage is potential overflow due to `idx + idx_bump`  when combined with `invalid_index = maxInt(u32)`

on init, new bucket is initialized to: 

```
bucket.stolen.start = invalid_index; // maxInt(u32)
bucket.stolen.end = invalid_index; //maxInt(u32)
```

Then later, when `alloc()` is called, it calls this: 

https://github.com/Syndica/sig/blob/e93aa69383a9cda8f177967bc15a5bee1c2cabb5/v2/lib/accounts_db/pool.zig#L169-L172

So on that first normal allocation, `bumpStolenBlock` does `maxInt(u32) + idx_bump` 

I think a sanity check at the top of function could help for this case?

```zig 
if (self.stolen.start == invalid_index) return null;
```

but I'm not sure if this is valid since debug build should have caught this?

prestonsn: or I think even a debug assert would do here

v2/lib/accounts_db/rooted.zig
prestonsn: I don't think this was changed (unless my shit is bugged)

https://github.com/Syndica/sig/blob/e93aa69383a9cda8f177967bc15a5bee1c2cabb5/v2/lib/accounts_db/rooted.zig#L206-L213

v2/lib/accounts_db/table.zig
```diff
@@ -0,0 +1,214 @@
...
+        for (0..PARALLEL_SCAN) |i| {
+            // empty curr has .slot=0 : always overwritten
+            const newer_slot = @as(Entry, @bitCast(curr[i])).slot <= batch.entries[i].slot;
+
+            const new_write = newer_slot and batch.entries[i].slot > 0;
+            self.count += @intFromBool(new_write); // bump for empty inserts
+
+            const ptr = if (newer_slot) &ptrs[i][0] else &stub;
+            ptr.* = batch.entries[i];
+        }
```
prestonsn: I think there's a small bug here where `self.count` is incremented for updates to existing accounts and not just for new entries to Table.

I think the `batch.entries[i].slot > 0` bit filters out zeroed/unused batch lanes, but I don't think that tels us whether `curr[i]` was empty right? I think an explicit check for `curr[i].hash == 0` seems needed to see if we're actually tracking a new account, or just updating an existing one 

(assuming here that `self.count` is number of entries, not number of updates/writes)

prestonsn: basically it seems like `self.count` increments when an existing account but at a newer slot shows up

v2/lib/accounts_db/table.zig
```diff
@@ -0,0 +1,214 @@
...
+                const collided = @reduce(.Or, (empty_mask | eq_mask) == zero);
+                if (collided) has_collisions = true;
+
+                // probe next item if collided
+                ptrs[i] += @intFromBool(collided);
+                if (ptrs[i] == self.entries.ptr + self.entries.len) ptrs[i] = self.entries.ptr;
+                curr[i] = @bitCast(ptrs[i][0]);
+            }
+            if (!has_collisions) break;
+        }
```
prestonsn: is it possible that different pubkeys end up choosing the same empty entry?

let say pubkey A is lane 0, and pubkey B is lane 1

if hash(A) -> reduced to idx 100 and hash(B) -> reduced to idx 100 as well, then at the end of phase 3:
```
ptrs[0] = &entries[100]
ptrs[1] = &entries[100]
```
So then in phase 4:

```
lane 0 writes A to entries[100]
lane 1 writes B to entries[100]
```

So then i think A is lost?

prestonsn: hmm i also think it could happen due to two pubkeys landing on same probe destination? (not necessary for just hashes to match)

prestonsn: also, is it possible for the same pubkey appearing twice in one batch maybe messes up the slot odering? If `A` appears twice in a flush, then there's no check in phase 4 to correctly record the highest of the A writes in the flush batch (there's enforcement for slot ordering on `curr[i]`, but not against writes frm inside the same batch).

kprotty: fixed phase-4 to check for overwrite linearly

https://github.com/Syndica/sig/pull/1393/commits/50cd8d2911450f537343d221fcd8a242c6e6aa0b

also fix zero-pubkey handling: 

https://github.com/Syndica/sig/pull/1393/commits/687f321e03bb5124cc85957583757e2ecaa94d86

v2/lib/accounts_db/table.zig
kprotty: added table-ptr-empty check for count increment

https://github.com/Syndica/sig/pull/1393/commits/50cd8d2911450f537343d221fcd8a242c6e6aa0b

v2/lib/accounts_db/table.zig
kprotty: Updated version here (no more assert, just u64 count with u64 index reduction): https://github.com/Syndica/sig/pull/1393/commits/50cd8d2911450f537343d221fcd8a242c6e6aa0b

v2/lib/accounts_db/rooted.zig
kprotty: missed in commit. This has it https://github.com/Syndica/sig/pull/1393/commits/ce0a6e041fa29790c85ac90f072a7be83b269034

v2/lib/accounts_db/pool.zig
kprotty: https://github.com/Syndica/sig/pull/1393/commits/ce0a6e041fa29790c85ac90f072a7be83b269034

v2/lib/accounts_db/rooted.zig
```diff
@@ -0,0 +1,808 @@
...
+        logger: tel.Logger("Rooted.put"),
+        slot: Slot,
+        pubkey: Pubkey,
+        owner: Pubkey,
+        lamports: u64,
+        rent_epoch: Epoch,
+        executable: bool,
+        data_len: usize,
+        data_reader: anytype, // anything with a std.Io.Reader.readSliceAll API
+    ) !void {
```
InKryption, line 438: Probably makes sense to use a parameter struct here, considering so many of these parameters are actually just ints (could take the data_reader param as second or third arg, mostly concerned about most things between slot and data_len).

v2/lib/accounts_db/rooted.zig
```diff
@@ -0,0 +1,808 @@
...
+
+            try self.queueWrite(.from(logger), data_len, data_reader);
+        }
+    }
+
+    fn queueWrite(
+        self: *Rooted,
+        logger: tel.Logger("Rooted.queueWrite"),
+        len: usize,
+        reader: anytype, // anything that impls an interface like std.Io.Reader
```
InKryption: Would be good to make these into doc comments, so that they can appear in docgen and tooling and all that.

v2/lib/accounts_db/rooted.zig
```diff
@@ -0,0 +1,808 @@
...
+        );
+        sqe.user_data = @bitCast(data);
+        _ = try self.ring.submit(); // SQPOLL makes this fast if done frequently enough
+    }
+
+    fn pollCompletedReads(self: *Rooted, logger: tel.Logger("Rooted.pollCompletedReads")) !void {
+        var cqes: [max_active_lookups]std.os.linux.io_uring_cqe = comptime blk: {
+            @setRuntimeSafety(false);
+            break :blk undefined;
+        };
```
InKryption: Just fyi this will not prevent the `0xAA` memset. When an `undefined` value is `comptime`-known, the compiler doesn't forget about it, even if you `@setRuntimeSafety(false)`, because that's only really applicable to runtime codegen. So this is just creating a comptime `undefined` in an unsafe scope, and then breaking it out into a safe scope, making it set `0xAA` bytes.

I did a little bit of testing, and this seems to accomplish it in the way you'd want:
```
inline fn initUndefUnchecked(comptime T: type) T {
    @setRuntimeSafety(false);
    var init: T = undefined;
    _ = &init;
    return init;
}
```
Optimizer obviously elides the `init` variable. And since it's a runtime-known `undefined`, the compiler can't propagate this knowledge across boundaries (you can also do this as an inline block, it's just the assigning to a `var` first that's important).

v2/lib/accounts_db/rooted.zig
```diff
@@ -0,0 +1,808 @@
...
+            try self.submitRead(.from(logger), data); // continue the read process
+        }
+    }
+
+    const ReadError =
+        error{ ReadFailed, EndOfStream } || // I/O error
+        error{InvalidRead} || // Corrupted data error
+        @typeInfo( // io_uring.submit() error
+            @typeInfo(@TypeOf(std.os.linux.IoUring.submit)).@"fn".return_type.?,
+        ).error_union.error_set;
```
InKryption, line 737: I don't think we really want to just blindly propagate all of submit's errors, do we? Think it would be better if we allocated just one error for it, ie `SubmitFail` or something.

v2/lib/fio/reader.zig
```diff
@@ -0,0 +1,233 @@
...
+        }
+
+        fn poll(self: *Self, logger: tel.Logger("FileReader.poll")) !void {
+            const zone = tracy.Zone.init(@src(), .{ .name = "FileReader.poll" });
+            defer zone.deinit();
+
+            var cqes: [num_blocks]std.os.linux.io_uring_cqe = blk: {
+                @setRuntimeSafety(false);
+                break :blk undefined; // avoid needless memset
+            };
```
InKryption: Ditto on this attempt to avoid the memset.

v2/lib/fio/writer.zig
```diff
@@ -0,0 +1,216 @@
...
+        }
+
+        fn poll(self: *Self, logger: tel.Logger("FileWriter.poll")) !void {
+            const zone = tracy.Zone.init(@src(), .{ .name = "FileWriter.poll" });
+            defer zone.deinit();
+
+            var cqes: [num_blocks]std.os.linux.io_uring_cqe = blk: {
+                @setRuntimeSafety(false);
+                break :blk undefined; // avoid needless memset
+            };
```
InKryption: Aqui tambien.

InKryption (commented):
Just a few observations and suggestions.

v2/lib/accounts_db/rooted.zig
kprotty: If io_uring does fail, would be helpful to know why. Could add error logging at that site, before returning up `error.SubmitFail`, but theres multiple sites to add which can trigger said error (`ring.submit` and `ring.copy_cqes`, if either does an `io_uring_enter`)

InKryption: I mean, even if you `catch return error.Foo`, the original error that caused the `catch` sticks in the error return trace.

v2/lib/fio/writer.zig
kprotty: https://github.com/Syndica/sig/pull/1393/commits/da975db13a031a470e4ed1b2e5d63cef7fe853d4

v2/lib/fio/reader.zig
kprotty: https://github.com/Syndica/sig/pull/1393/commits/da975db13a031a470e4ed1b2e5d63cef7fe853d4

v2/lib/accounts_db/rooted.zig
kprotty: https://github.com/Syndica/sig/pull/1393/commits/da975db13a031a470e4ed1b2e5d63cef7fe853d4

InKryption: There's no reason to do it for global variables, they have no runtime init, and last I checked the compiler is already smart enough to zero out global `undefined` inits.

v2/lib/accounts_db/table.zig
```diff
@@ -0,0 +1,239 @@
...
+        while (true) {
+            var had_collision = false;
+            for (0..PARALLEL_SCAN) |i| {
+                const item_key = item_keys[i];
+                const empty_mask = @select(u8, keys[i] == zero, ~zero, zero);
+                const eq_mask = @select(u8, keys[i] == item_key, ~zero, zero);
+
+                // check for collisions (!eq and !empty = populated with unrelated entry)
+                const collided = @reduce(.Or, (empty_mask | eq_mask) == zero);
+                if (collided) had_collision = true;
```
prestonsn: I might be missing something here, but is this collision check meant to be performed on the whole key (all key bytes at once) and not byte-by-byte like how it is now?

I think as written there's a case where the masks don't work as expected:

```
zero:        00 00 00 00
keys[i]:     00 aa 00 bb
item_key:    11 aa 22 bb
```

`keys[i]` is neither `zero` nor equal to `item_key`, so this should be a collision right? But since each byte of `keys[i]` is either zero OR equal to corrsponding `item_key` byte,  `empty_mask | eq_mask` in the reduce would be non zero in every byte lane, making `collided = false`

thoughts on comparing the whole key? (I'm not sure on impact on perf here, might be bad):


```suggestion
                const item_key = item_keys[i];
                const is_empty = @reduce(.And, keys[i] == zero);
                const is_equal = @reduce(.And, keys[i] == item_key);
                const collided = !is_empty and !is_equal;
                if (collided) had_collision = true;
```

v2/lib/accounts_db/table.zig
prestonsn: The changes here look good to me :saluting_face: 

I had a follow-up on the new masking here: https://github.com/Syndica/sig/pull/1393/changes#r3337070594

v2/lib/accounts_db/table.zig
```diff
@@ -0,0 +1,239 @@
...
+        std.debug.assert(slot <= std.math.maxInt(u32));
+        std.debug.assert(!value.isEmpty());
+
+        // We use a zero pubkey in the map to denote empty Entry.
+        // If actualy inserting to the zero pubkey, update the one in the table instead.
+        if (pubkey.isZeroed()) {
+            @branchHint(.unlikely); // zero pubkey is System Program. Should rarely be written.
+            if (slot >= self.zero_entry.slot) self.zero_entry.value = value;
+            return;
+        }
```
prestonsn, line 98: With new changes here, should this be also update `self.zero_entry.slot`?


```suggestion
        if (pubkey.isZeroed()) {
            @branchHint(.unlikely); // zero pubkey is System Program. Should rarely be written.
            if (slot >= self.zero_entry.slot) {
                self.zero_entry.value = value;
                self.zero_entry.slot = @intCast(slot);
            }
            return;
        }
```

v2/lib/accounts_db/rooted.zig
prestonsn: Looks like this change got reverted by 687f321e03bb5124cc85957583757e2ecaa94d86 somehow, maybe for good reason?

v2/lib/accounts_db/table.zig
prestonsn: Actually, I think there's still slight gap. With latest serialize changes to phase 4, the "same pubkey A appearing multiple times in the batch" case is handled well, but I don't think it handles two different pubkeys choosing the same entry in `entries`. 

I still think it's possible that an earlier lane filled this `ptrs[i]` with a different pubkey, and a later lane only rechecks slot freshness (via `item_slot >= ptrs[i][0].slot`). 

I think there should also be a check for `ptrs[i][0].key.equals(&item_key)` before being able to trust `item_slot >= ptrs[i][0].slot`

if `ptrs[i][0]` now belongs to a different pubkey due to an earlier lane's write, then a later lane's should re-probe instead of overwriting solely based on slot

v2/lib/accounts_db/rooted.zig
```diff
@@ -0,0 +1,797 @@
...
+
+        node.file_offset = entry.offset;
+        try self.submitRead(.from(logger), .{
+            .lookup_idx = lookup_idx,
+            .reading = .header,
+            .read = 0,
+        });
+
+        return true;
+    }
```
prestonsn, line 615: maybe this is intentional, but it seems like we pop a node with `  self.free_lookups = node.next;` then do fallible stuff via `try self.account_pool.alloc(...)` and `try self.submitRead(...)`

are we missing `errdefer` here? (should likely never happen i think/perhaps this is irrecoverable anyways)

v2/lib/accounts_db/rooted.zig
kprotty: I mean rather that this results in `catch |e| switch (e) { ..all io uring error cases => return error.Foo }` for 3 call sites atm: two `ring.submit()`s and a `ring.copy_cqes()`, which seems bloated

v2/lib/accounts_db/rooted.zig
kprotty: https://github.com/Syndica/sig/pull/1393/commits/851dcb22244efbb1be5f1f78eda40fe6a315dbc1 + https://github.com/Syndica/sig/pull/1393/commits/dbad4bb9167bcf7a211a2ea61beddba217ffe30a

v2/lib/accounts_db/rooted.zig
kprotty: https://github.com/Syndica/sig/pull/1393/commits/013cc982fcf506b7bcadf61eb3be837968540c16

v2/lib/accounts_db/rooted.zig
kprotty: https://github.com/Syndica/sig/pull/1393/commits/013cc982fcf506b7bcadf61eb3be837968540c16

v2/lib/accounts_db/table.zig
kprotty: https://github.com/Syndica/sig/pull/1393/commits/2b7ae67ad364fcf4c02da433437413af7d3d1de6

v2/lib/accounts_db/table.zig
kprotty: https://github.com/Syndica/sig/pull/1393/commits/2b7ae67ad364fcf4c02da433437413af7d3d1de6

v2/lib/accounts_db/table.zig
kprotty: https://github.com/Syndica/sig/pull/1393/commits/2b7ae67ad364fcf4c02da433437413af7d3d1de6

InKryption (dismissed):

v2/lib/accounts_db/rooted.zig
kprotty: https://github.com/Syndica/sig/pull/1393/commits/bb450dc24a53e7346d8899e01d0af24760575d6e

v2/services/accounts_db.zig
kprotty: https://github.com/Syndica/sig/pull/1393/commits/bb450dc24a53e7346d8899e01d0af24760575d6e

v2/lib/clock.zig
kprotty: need it now https://github.com/Syndica/sig/pull/1393/commits/dcc52f36d2f3216788b714dc63071631c2f423e5

v2/lib/accounts_db/table.zig
prestonsn: Very clean :saluting_face:

v2/lib/accounts_db/table.zig
prestonsn: fully resolved sir :saluting_face:

v2/lib/accounts_db/rooted.zig
prestonsn: Clean :saluting_face:

v2/lib/solana/snapshot.zig
```diff
@@ -0,0 +1,804 @@
...
+            // defer zone.deinit();
+
+            const entries = self.entries;
+            var idx = (slot *% HASH_MULT) >> @intCast(@as(u7, 64) - @ctz(entries.len));
+            while (true) {
+                const e = &entries[idx];
+                idx = (idx +% 1) & (entries.len - 1);
+                if (@as(u128, @bitCast(e.*)) == @as(u128, @bitCast(Entry.empty))) continue;
+                if (e.slot == slot) return e;
+            }
```
prestonsn, line 373: likely need to return `null` here for non-existent `slot` param

v2/lib/fio/reader.zig
```diff
@@ -0,0 +1,231 @@
...
+
+            std.debug.assert(self.io_inflight < num_blocks);
+            self.io_inflight += 1;
+
+            const sqe = while (true) break self.ring.get_sqe() catch |err| switch (err) {
+                error.SubmissionQueueFull => {
+                    _ = try self.ring.submit();
+                    continue;
+                },
+            };
```
prestonsn, line 178: it looks like we modify `block_state.status = .reading;` then could potentially fail on `self.rin.submit()`. Maybe need an `errdefer` for this?

prestonsn: oh I guess same with `self.io_inflight += 1;`

v2/lib/fio/writer.zig
```diff
@@ -0,0 +1,213 @@
...
+
+            std.debug.assert(self.io_inflight < num_blocks);
+            self.io_inflight += 1;
+
+            const sqe = while (true) break self.ring.get_sqe() catch |err| switch (err) {
+                error.SubmissionQueueFull => {
+                    _ = try self.ring.submit();
+                    continue;
+                },
+            };
```
prestonsn, line 165: similar to the `Reader.submit()` feedback, but with `writing_mask` and `self.io_inflight` modifications. Potentially need an errdefer here?

v2/lib/solana/snapshot.zig
```diff
@@ -0,0 +1,804 @@
...
+            while (self.account_file_len == 0) {
+                @branchHint(.unlikely);
+
+                const tar_file = (try self.tar_iter.next()) orelse return null;
+                const split = std.mem.indexOfScalar(u8, tar_file.name, '.') orelse
+                    return error.InvalidAccountFileName;
+                if (split + 1 >= tar_file.name.len)
+                    return error.InvalidAccountFileName;
+
+                const slot = std.fmt.parseInt(u64, tar_file.name["accounts/".len..split], 10) catch
```
prestonsn, line 590: (nit) is it worth also checking that `std.mem.startsWith(u8, tar_file.name, "accounts/')` since that is assumed here?

Then can check that `split <= "accounts/".len`

Can just mark this resolved tho, hard to be totally defensive against malformed snapshot.

v2/lib/fio/reader.zig
kprotty: fio types are considered invalid after returning an error (they do IO concurrently & that isnt recoverable)

prestonsn: fairs, going to resolve this then

v2/lib/solana/snapshot.zig
kprotty: SnapshotIter doesnt check the manifest tar_file name either. I think its fine for now; as long the file has the right data (which is runtime validated)

v2/lib/solana/snapshot.zig
kprotty: https://github.com/Syndica/sig/pull/1393/commits/7628fa304fb33bad06a72e9bfdf800b39eae1351

prestonsn (approved):
LGTM 🫡

Sobeston (approved):

InKryption (approved):

PR #1397, opened 2026-04-20
fix(vote): align commission and withdraw checks with agave

yewman:


conformance/build.zig
```diff
@@ -88,14 +87,142 @@ pub fn build(b: *Build) void {
...
+        if (val.len >= 2 and val[0] == '"' and val[val.len - 1] == '"') {
+            return b.dupe(val[1 .. val.len - 1]);
+        }
+        return b.dupe(val);
+    }
+    return null;
+}
+
+// FetchProtosStep – downloads .proto files from GitHub at a pinned commit
+const FetchProtosStep = struct {
```
Rexicon226: Keep in mind that,
1. custom build steps will be disallowed/broken in the future
2. downloading implicitly within the build script is generally a bad idea
3. this is what the build.zig.zon is for!! If you want, try expressing the commits within the build.zig.zon.

yewman: Makes sense, I will apply the fix in the original PR :) [fix(conformance): update protobuf dep to enforce field ordering](https://github.com/Syndica/sig/pull/1395#top)

InKryption (approved):

hamza-syndica (approved):

Issue #1401, opened 2026-04-21
feat(v2): document how to create a new service


prestonsn:
### Problem

Creating a new service is not documented

### Proposal

Little walkthough in README for how to create a new dummy service, setup regions, and read/write to data in regions

Sobeston:
closed by #1402

PR #1402, opened 2026-04-22
docs(v2): add tutorial for creating a new v2 service to v2/README.md

prestonsn:
Adds a small section to README that walks through how v2 service architecture is organized and how to add a new service called `foo`. 

https://github.com/Syndica/sig/issues/1401

### DIFF
--- a/docs/docusaurus/docs/code/v2.md
+++ b/docs/docusaurus/docs/code/v2.md
@@ -38,3 +38,216 @@ advertise a turbine receive port.
 1. Build with `-Denable-tracy`
 
 2. Use `.sandboxing_mode = .threaded` in your config
+
+# Adding a New Service
+
+v2 runs each service as an isolated process. Services cannot allocate memory or open files at runtime. 
+Instead, they communicate through shared memory: fixed-size buffers created by the parent process before any
+service is spawned. Each service declares which regions it needs and whether it needs read-only or read-write access.
+
+This walkthrough adds a service called `foo` that reads a config region
+and logs through telemetry.
+
+```mermaid
+graph TD
+    subgraph "<i>init/</i> (parent process)"
+        main["<i>main.zig</i><br/>parses config, sets up<br/>region init args"]
+        zon["<i>services.zon</i><br/>declares region/service wiring"]
+        svc["<i>services.zig</i><br/>creates + initializes regions,<br/>spawns service processes"]
+    end
+
+    subgraph "<i>lib/</i> (validator components)"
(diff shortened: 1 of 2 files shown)


jbuckmccready:
Mermaid diagram doesn't seem to render in Github: https://github.com/Syndica/sig/blob/7a7e4bc357243d7136a65586e9e53e3e9dad9b53/docs/docusaurus/docs/code/v2.md

EDIT: it loaded after refreshing, so just transient bug with Github.

docs/docusaurus/docs/code/v2.md
```diff
@@ -37,4 +37,217 @@ advertise a turbine receive port.
...
++            .name = .foo,
++            .regions = .{
++                .{ .name = .foo_config, .access = .readonly },
++                .{ .name = .telemetry, .access = .rw },
++            },
++        },
+     },
+```
+
+The parent process needs to how to initialize the region before spawning
```
jbuckmccready, line 170:
```suggestion
The parent process needs to know how to initialize the region before spawning
```

docs/docusaurus/docs/code/v2.md
```diff
@@ -37,4 +37,217 @@ advertise a turbine receive port.
...
-2. Use `.sandboxing_mode = .threaded` in your config
\ No newline at end of file
+2. Use `.sandboxing_mode = .threaded` in your config
+
+# Adding a New Service
+
+v2 runs each service as an isolated process. Services cannot allocate memory or open files at runtime. 
+Instead, they communicate through shared memory: fixed-size buffers created by the parent process before any
+service is spawned. Each service declares which regions it needs and whether it needs read-only or read-write access.
+
```
Sobeston:
```suggestion

v2 runs each service as an isolated process. Services typically cannot allocate memory or open files at runtime; any syscalls used must be whitelisted, as they are denied by default. 
Instead, they communicate through shared memory using fixed-size buffers created by the init process before any
service is spawned. Each service declares which regions it needs and whether it needs read-only or read-write access.
```

Sobeston (commented):
Looking good, just wanted to touch up this paragraph

docs/docusaurus/docs/code/v2.md
prestonsn: Oops, thanks! :saluting_face:  https://github.com/Syndica/sig/pull/1402/commits/8fc2bdac8da5e3cffca1436533e6c26291af4d80

docs/docusaurus/docs/code/v2.md
prestonsn: fax :saluting_face:

Sobeston (approved):

jbuckmccready (approved):

PR #1403, opened 2026-04-22
fix(ci): handle exit more gracefully (ensure archive deletion)

InKryption:


ci/run-and-update-service/handle-exit.sh
```diff
@@ -18,22 +18,27 @@ fi
...
-    # upload artifact to S3 and clean up local state
-    . /etc/sig.conf  # get S3_BUCKET
-    if [ -z "${S3_BUCKET:-}" ]; then
-        log error "S3_BUCKET is not set, cannot upload validator archive"
-        exit 1
-    fi
-    aws s3 cp "$BASE_DIR/validator-${timestamp}.tar.zst" "s3://$S3_BUCKET/ci-crashes/validator-${timestamp}.tar.zst"
-    rm "$BASE_DIR/validator-${timestamp}.tar.zst"
+    upload_file $archive_path || true
+    rm $archive_path
```
dnut: this will not run if a single command in this script has a non-zero exit code since we have `set -e` (regardless of whether the command is in a function). also if `exit 1` in upload_file is hit. this rm really needs to be handled like a defer, which in bash is a bit awkward to handle.

dnut (changes requested):
i don't see any semantic change here. i think this is just a refactor. the file would still fail to be deleted under the same conditions

ci/run-and-update-service/handle-exit.sh
dnut: you could use a trap for this, or you could put `|| { rm $archive_path && exit 1; }` for every command after the file is created

InKryption: a973de66eedce0c58ca6b00779b4abe6de49c1c8

ci/run-and-update-service/handle-exit.sh
```diff
@@ -24,16 +24,20 @@ if [ -d "$BASE_DIR/validator" ]; then
 
     # compress validator state to a tarball
     cp "$BASE_DIR/logs/sig.log" "$BASE_DIR/validator/"
-    timestamp="$(date -u +"%Y-%m-%dT%H:%M:%S.%3NZ")"
-    tar --zstd -cf "$BASE_DIR/validator-${timestamp}.tar.zst" -C "$BASE_DIR" validator
+    local timestamp="$(date -u +"%Y-%m-%dT%H:%M:%S.%3NZ")"
+    local archive_name="validator-${timestamp}.tar.zst";
+    local archive_path="$BASE_DIR/$archive_name"
```
dnut: `local` can only be used in a function. the script will crash on these lines

ci/run-and-update-service/handle-exit.sh
```diff
@@ -24,16 +24,20 @@ if [ -d "$BASE_DIR/validator" ]; then
...
+    tar --zstd -cf $archive_path -C "$BASE_DIR" validator
     rm -rf "$BASE_DIR/validator"
+    trap "rm -f $archive_path"
```
dnut: this is invalid, trap needs a signal name at the end.

also this will override the other trap above. we should keep that behavior too

dnut (changes requested):
please test this before we merge

ci/run-and-update-service/handle-exit.sh
InKryption: How does this one look 1f7cfa91d8d48f9dff9e5c774771fd89a5180343

ci/run-and-update-service/handle-exit.sh
```diff
@@ -25,15 +25,19 @@ if [ -d "$BASE_DIR/validator" ]; then
     # compress validator state to a tarball
     cp "$BASE_DIR/logs/sig.log" "$BASE_DIR/validator/"
     timestamp="$(date -u +"%Y-%m-%dT%H:%M:%S.%3NZ")"
-    tar --zstd -cf "$BASE_DIR/validator-${timestamp}.tar.zst" -C "$BASE_DIR" validator
+    archive_name="validator-${timestamp}.tar.zst";
+    archive_path="$BASE_DIR/$archive_name"
+    tar --zstd -cf $archive_path -C "$BASE_DIR" validator
     rm -rf "$BASE_DIR/validator"
+    trap "rm -f $archive_path" EXIT
```
jbuckmccready: This will replace existing `trap` which logs for observability of failures. Can do something like this maybe (replacing the original `trap` set on line 11):

```sh
   archive_path=""

   on_exit() {
       local e=$?
       set +e

       if [ -n "$archive_path" ]; then
           rm -f -- "$archive_path"
       fi

       if [ "$e" -ne 0 ]; then
           log error "handle-exit.sh exited with code $e. see sig.service logs"
       fi

       exit "$e"
   }

   trap on_exit EXIT
```

jbuckmccready (changes requested):

ci/run-and-update-service/handle-exit.sh
InKryption: What do you think of this? 5d9513daa7cfa2e74dbd3cfdea5ceb7fb2720bd9

ci/run-and-update-service/handle-exit.sh
InKryption: I think this one'll do it: a5cac73aad80d6aff9097e47ced6d65039ae77e2

ci/run-and-update-service/handle-exit.sh
jbuckmccready: Looks good, only minor nit I'd suggest is removing the `rm $archive_path` at the end since the trap will execute the `rm` on clean exit anyways.

InKryption: a5cac73aad80d6aff9097e47ced6d65039ae77e2

jbuckmccready (approved):

dnut (approved):

PR #1405, opened 2026-04-22
feat(v2): snapshot download

prestonsn:
Uses io_uring to race multiple candidate snapshot downloads concurrently, sticking with the fastest one after it crosses a certain threshold bytes written. The rather large number of lines of code has to do with the large number of checks/asserts in each state transition handler, and creating some of the necessary pieces to make/parse HTTP requests/responses, minimizing dependence on std.

- [x] io_uring based race for downloading multiple snapshots concurrently
- [x] `known_list` whitelist configuration via `testnet.zig.zon`. Only gossip pears in the whitelist are used in the download race. Supports wildcard (allow all) via specifying `["*"]` in the list:  https://github.com/Syndica/sig/blob/8d2ade3136a157552cb3f73812667149209223b9/v2/config/testnet.zig.zon#L20
- [x] Refactor sub-phase states for `Probe` and `Download` into `enum union`

### How it works

1. Streams in peers' slot, hash, identity pk, and socketaddr from gossip service.
2. Check whitelist first to see if the peer's in the list (if the list is present).
3. Dedupes peers by `socketaddr` and tracks `socketaddr -> (slot, hash, pubkey, probe status, latency)`.
4. New or updated peers enter the probe state machine:
   - `connecting`: opens a TCP connection to the peer.
   - `sending`: sends an HTTP HEAD req for snapshot file: `HEAD /snapshot-{slot}-{hash}.tar.zst HTTP/1.1`
   - `receiving`: reads the HTTP response, checks for `200`, and parses `Content-Length`.
5. Successful probes become download candidates with their content length and observed latency.
6. The download race starts candidates and advances each download through:
   - connect and send HTTP GET request
   - read/validate headers and write any extra body bytes already received
   - splice snapshot bytes from socket to file through a pipe
   - fsync the completed file
7. Once a winner is selected or completes, loser downloads are cancelled/cleaned up and the final snapshot file is made available.

Note: There are still a few TODOs around the download code that I've left unaddressed for now (mostly around code-org, clean-ups, using free lists instead of iterating, and other minor optimizations). The core functionality of the race, whitelisting of gossip peers, finalizing the download path, cleanup of losers, etc. is complete and tested. The largest piece of low-hanging fruit would be the re-opening of a TCP socket to the peer when upgrading it from the probe phase to the download phase. I've created an issue to track these here: https://github.com/Syndica/sig/issues/1448

### Sample Run

```zig
ubuntu@o-use1-sig-001:~/sig/v2$ zig build run -Doptimize=ReleaseSafe  -- config/testnet.zig.zon
info: config: .{
    .sandboxing_mode = .sandboxed,
    .cluster = .testnet,
    .leader_schedule_file = "schedule.txt",
    .gossip = .{ .port = 8001 },
    .shred_network = .{ .recv_port = 8002 },
    .telemetry = .{ .port = 12345 },
    .snapshot = .{ .folder = "./validator" },
    .known_validators = .{ "*" },
}
info: Initialising: .net_pair
info: Initialised: Region `net_pair` shared with [ net_0 (rw), shred_receiver_0 (rw), ]
info: Initialising: .shred_recv_config
info: Initialised: Region `shred_recv_config` shared with [ shred_receiver_0 (ro), ]
info: Initialising: .net_pair
info: Initialised: Region `net_pair` shared with [ net_0 (rw), gossip_0 (rw), ]
info: Initialising: .gossip_config
info: Initialised: Region `gossip_config` shared with [ gossip_0 (ro), ]
info: Initialising: .telemetry
info: Initialised: Region `telemetry` shared with [ net_0 (rw), gossip_0 (rw), shred_receiver_0 (rw), telemetry_0 (rw), snapshot_0 (rw), ]
info: Initialising: .snapshot_source_ring
info: Initialised: Region `snapshot_source_ring` shared with [ gossip_0 (rw), snapshot_0 (rw), ]
info: Initialising: .snapshot_config
info: Initialised: Region `snapshot_config` shared with [ snapshot_0 (ro), ]
info: Starting Service `shred_receiver_0`, pid: 581451
info: Starting Service `net_0`, pid: 581452
info: Starting Service `gossip_0`, pid: 581453
info: Starting Service `telemetry_0`, pid: 581454
info: Starting Service `snapshot_0`, pid: 581455
time=2026-05-14T17:39:26.758Z level=info scope=main service=net message="binding 0.0.0.0:8002"
time=2026-05-14T17:39:26.758Z level=info scope=main service=net message="binding 0.0.0.0:8001"
time=2026-05-14T17:39:26.758Z level=info scope=main service=shred_receiver message="Waiting for shreds on port 8002"
time=2026-05-14T17:39:26.758Z level=info scope=main service=gossip message="Gossip started on :8001 as 9Arn2juTs56iRqghNuouPEG6x6jxkuknjXQ8wE2qATvs:
        shred_version:57087
        entrypoints:{ 35.203.170.30:8001, 109.94.99.177:8001 }"
time=2026-05-14T17:39:26.759Z level=info scope=snapshot service=snapshot message="snapshot path ./validator"
time=2026-05-14T17:39:28.645Z level=err scope=processMessage service=gossip message="pong from untracked peer 5yXtKhQ6YTGy6iR9EhfZqvPjyxGfvwPjAv8DKNcLWTVM:64.130.61.108:9001"
time=2026-05-14T17:39:30.067Z level=warn scope=snapshot service=snapshot message="probe recv bad status from=QwHmim3t55tdrnTK2PUtzWcA41rtLPU72pwcHyXiHSs status=HTTP/1.1 405 Method Not Allowed"
time=2026-05-14T17:39:30.999Z level=info scope=snapshot service=snapshot message="download winner from=3RoP1jj8dKwGyDK7kXh5BaudvQVzZfMjVqGvn91wKUcs addr=149.28.33.168:8899"
time=2026-05-14T17:39:32.063Z level=info scope=snapshot service=snapshot message="download progress from=3RoP1jj8dKwGyDK7kXh5BaudvQVzZfMjVqGvn91wKUcs completed=16.9% speed=328.87MiB/s"
time=2026-05-14T17:39:33.063Z level=info scope=snapshot service=snapshot message="download progress from=3RoP1jj8dKwGyDK7kXh5BaudvQVzZfMjVqGvn91wKUcs completed=26.0% speed=455.64MiB/s"
time=2026-05-14T17:39:34.064Z level=info scope=snapshot service=snapshot message="download progress from=3RoP1jj8dKwGyDK7kXh5BaudvQVzZfMjVqGvn91wKUcs completed=35.3% speed=470.72MiB/s"
time=2026-05-14T17:39:35.064Z level=info scope=snapshot service=snapshot message="download progress from=3RoP1jj8dKwGyDK7kXh5BaudvQVzZfMjVqGvn91wKUcs completed=44.3% speed=453.19MiB/s"
time=2026-05-14T17:39:36.064Z level=info scope=snapshot service=snapshot message="download progress from=3RoP1jj8dKwGyDK7kXh5BaudvQVzZfMjVqGvn91wKUcs completed=52.7% speed=427.76MiB/s"
time=2026-05-14T17:39:37.065Z level=info scope=snapshot service=snapshot message="download progress from=3RoP1jj8dKwGyDK7kXh5BaudvQVzZfMjVqGvn91wKUcs completed=61.1% speed=422.40MiB/s"
time=2026-05-14T17:39:38.065Z level=info scope=snapshot service=snapshot message="download progress from=3RoP1jj8dKwGyDK7kXh5BaudvQVzZfMjVqGvn91wKUcs completed=69.1% speed=404.49MiB/s"
time=2026-05-14T17:39:39.066Z level=info scope=snapshot service=snapshot message="download progress from=3RoP1jj8dKwGyDK7kXh5BaudvQVzZfMjVqGvn91wKUcs completed=76.7% speed=382.43MiB/s"
time=2026-05-14T17:39:40.066Z level=info scope=snapshot service=snapshot message="download progress from=3RoP1jj8dKwGyDK7kXh5BaudvQVzZfMjVqGvn91wKUcs completed=84.7% speed=403.33MiB/s"
time=2026-05-14T17:39:41.066Z level=info scope=snapshot service=snapshot message="download progress from=3RoP1jj8dKwGyDK7kXh5BaudvQVzZfMjVqGvn91wKUcs completed=94.5% speed=493.97MiB/s"
time=2026-05-14T17:39:42.066Z level=info scope=snapshot service=snapshot message="download progress from=3RoP1jj8dKwGyDK7kXh5BaudvQVzZfMjVqGvn91wKUcs completed=100.0% speed=279.60MiB/s"
time=2026-05-14T17:39:43.067Z level=info scope=snapshot service=snapshot message="download progress from=3RoP1jj8dKwGyDK7kXh5BaudvQVzZfMjVqGvn91wKUcs completed=100.0% speed=0.00MiB/s"
time=2026-05-14T17:39:43.075Z level=info scope=snapshot service=snapshot message="download complete slot=408299544 path=./validator/snapshot-408299544-AusbokkrQa4pgSZZ5iHpb3duXHjw4QdLzRdkCVb4rCCa.tar.zst"
time=2026-05-14T17:39:43.075Z level=info scope=snapshot service=snapshot message="snapshot download completed slot=408299544 hash=AusbokkrQa4pgSZZ5iHpb3duXHjw4QdLzRdkCVb4rCCa path=./validator/snapshot-408299544-AusbokkrQa4pgSZZ5iHpb3duXHjw4QdLzRdkCVb4rCCa.tar.zst"
```


v2/config/testnet.zig.zon
```diff
@@ -12,4 +12,10 @@
     .telemetry = .{
         .port = 12345,
     },
+    .snapshot = .{
+        .folder = "./validator",
+    },
+
+    // NOTE: currently all validators are allowed as snapshot sources, but this can be configured to a specific set of pubkeys.
+    .known_validators = .{"*"},
```
kprotty: should this be moved into the `.snapshot=.{}` config item?

v2/config/testnet.zig.zon
```diff
@@ -12,4 +12,10 @@
     .telemetry = .{
         .port = 12345,
     },
+    .snapshot = .{
+        .folder = "./validator",
```
kprotty, line 17: probably create a `v2/.gitignore` to ignore this folder

v2/services/snapshot.zig
```diff
@@ -0,0 +1,90 @@
...
+    gossip_to_snapshot: *SnapshotSourceRing,
+};
+
+pub fn serviceMain(ro: ReadOnly, rw: ReadWrite) !noreturn {
+    const logger = rw.tel.acquireLogger(@tagName(name), "snapshot");
+    const metrics = rw.tel.metricAppender().appendFields(Metrics, Metrics.fields_config);
+
+    rw.tel.signalReady();
+
+    const snapshot_dir = ro.config.folder_buffer[0..ro.config.folder_len];
```
kprotty, line 37: snapshot_dir should `std.fs.cwd().openDir()` and passed around as a file handle. (move `makeDir` before the open, or merge makeDir & open into the same "createOrOpen" type call if that's possible).

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2632 @@
...
+const Pubkey = lib.solana.Pubkey;
+const Slot = lib.solana.Slot;
+const Hash = lib.solana.Hash;
+const IoUring = std.os.linux.IoUring;
+const SnapshotSourceRing = lib.snapshot.SnapshotSourceRing;
+const KnownValidators = lib.snapshot.SnapshotConfig.KnownValidators;
+
+const IO_URING_ENTRIES = 256;
+const MAX_DRAIN: u8 = 64;
+const GOSSIP_DRAIN_INTERVAL: std.os.linux.kernel_timespec = .{ .sec = 0, .nsec = 100_000_000 };
```
kprotty: couldnt gossip be drained continuously (also without a max)?

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2632 @@
...
+        metrics: Metrics,
+        logger: tel.Logger("snapshot"),
+    ) !Downloader {
+        return .{
+            .ring = try IoUring.init(IO_URING_ENTRIES, 0),
+            .gossip_iter = gossip_to_snapshot.get(.reader),
+            .dedupe_map = DedupeMap{},
+            .dedupe_alloc = dedupe_alloc,
+            .known_validators = known_validators,
+            .probe_conns = .{ProbeConn.empty()} ** MAX_CONCURRENT_PROBES,
```
kprotty: use `@splat(v)` instead of `.{v} ** len`

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2632 @@
...
+            .active_probes = 0,
+            .timeout_pending = false,
+            .download_conns = .{DownloadConn.empty()} ** MAX_DOWNLOAD_RACERS,
+            .active_downloads = 0,
+            .download_race = DownloadRace.empty(),
```
kprotty: can use decl-literals: `.empty()` instead of `T.empty()`

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2632 @@
...
+        }
+
+        self.ring.deinit();
+    }
+
+    pub fn run(self: *Downloader) !DownloadResult {
+        var cqes: [256]std.os.linux.io_uring_cqe = undefined;
+
+        // drain messages from gossip service immediately. This also submits the first timeout for drain interval.
+        try self.handleGossipDrainTimeout();
```
kprotty: if moved to inside the `while (true)` , gossip can avoid submitting io_uring timeouts. Should then change `submit_and_wait(1)` to be `submit_and_wait(0)` - so it can spin on uring & check gossip again instead of getting blocked in the kernel (spinning on uring is also just generally faster, per some snapshot-loading experiments)

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2632 @@
...
+        break :blk config;
+    };
+};
+
+pub const DedupeMap = std.array_hash_map.ArrayHashMapUnmanaged(
+    Address,
+    PeerState,
+    std.array_hash_map.AutoContext(Address),
+    true,
+);
```
kprotty: Is this potentially unbounded? Should ideally switch to a custom fixed-sized map that does its own eviction, even a flat array would probably work.

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2632 @@
...
+        return .{
+            .primary = self.ring.get_sqe() catch unreachable,
+            .timeout = self.ring.get_sqe() catch unreachable,
+        };
+    }
+
+    fn reserveSqe(self: *Downloader) !*std.os.linux.io_uring_sqe {
+        const capacity: u32 = @intCast(self.ring.sq.sqes.len);
+        if (capacity - self.ring.sq_ready() < 1) return error.SubmissionQueueFull;
+        return self.ring.get_sqe() catch unreachable;
```
kprotty: isnt this just `self.ring.get_sqe()`

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2632 @@
...
+        if (self.download_race.phase == .idle) {
+            self.download_race = DownloadRace.empty();
+            self.download_race.phase = .racing;
+            self.download_race.slot = slot;
+            self.download_race.hash = hash;
+        }
+
+        // Only accept candidates matching the current race target.
+        if (self.download_race.slot != slot) return;
+        if (!std.meta.eql(self.download_race.hash, hash)) return;
```
kprotty: shouldnt it allow racing for snapshots of different slots/hashes? there could be a faster node either from a new slot (or even an older slot)

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2632 @@
...
+        if (active.fd >= 0) std.posix.close(active.fd);
+        if (active.file_fd >= 0) std.posix.close(active.file_fd);
+        if (active.pipe_rd >= 0) std.posix.close(active.pipe_rd);
+        if (active.pipe_wr >= 0) std.posix.close(active.pipe_wr);
+
+        // If this download was cancelled/failed delete its temp file.
+        if (result != .succeeded) {
+            var tmp_buf: [std.fs.max_path_bytes:0]u8 = undefined;
+            // TODO: we gotta store this bruh.
+            const tmp_path = formatTempSnapshotPath(
```
kprotty: nah, its cheap to create

kprotty (commented):
While reading the code, realized the logic could be simplified & memory more compact when modeled as dataflow into each data structure:
```rs
// gossip -> dedup into this open-addressing hashmap.
// ignores the peers' pubkey since only cares about slot/hash on addr.
// if exists & fresh, update slot/hash to newer one
// if exists & probing: either drop it, or maybe add a peer.val.next_slothash current prober can retry with
// if exists & (dl or probe) failed: update slot/hash if newer & mark as fresh again
peers: [MAX_PEERS]{
    key: Addr,
    val: { slot_hash, latency: MAX, state:{fresh/probing/failed}, next: ?peer_idx }
}
// when peer inserted, push to this linked list
fresh_list: peer_idx
// when < MAX_PROBES active, pop from fresh_list & start new probe
probes: [MAX_PROBES]{ http_fd, send_buf, peer_idx, state:{send/recv/idle} }
// when probe finishes, update its latency & min-heap insert into this (key=peers[idx].latency)
dl_queue: [MAX_PROBES]{ peer_idx }
// when < MAX_DOWNLOADERS active, pop from dl_queue & start download
downloaders: [MAX_DOWNLOADERS]{
    peer_idx
    http_fd, send_buf, recv_buf, state:{send/recv/download/idle},
    file_fd, remaining,
    // recv(socket, recv_buf) -> pwrite(file, recv_buf). no need for pipe, the copy is fine
}
// when dl finishes, set result & io_cancel all active downloaders & probers
// when dl fails, fail its peers[idx]. If peers is full & all failed, reset peers array/map
```

v2/config/testnet.zig.zon
prestonsn: True, https://github.com/Syndica/sig/pull/1405/commits/7efdb231ce481b6f665bae42a03788bc05ad5183

v2/config/testnet.zig.zon
prestonsn: https://github.com/Syndica/sig/pull/1405/commits/ed90f0427998077e30ceb62792de3031d73d7e04

v2/lib/snapshot/download.zig
prestonsn: Yeah it could. Think I over thought it there and was worried about it just spinning receiving entries from gossip. Removed and testing things: https://github.com/Syndica/sig/pull/1405/commits/44a02f50eac2d218e88835379d36e6bee2015e67

v2/lib/snapshot/download.zig
prestonsn: https://github.com/Syndica/sig/pull/1405/commits/21f45fe0da94b8d11c03490746f58aec457b3ba3

v2/lib/snapshot/download.zig
prestonsn: https://github.com/Syndica/sig/pull/1405/commits/0d7183a864f51bf40bbeeb33326062d4dec48a0b

v2/lib/snapshot/download.zig
prestonsn: Ahh interesting! Made the change here: https://github.com/Syndica/sig/pull/1405/commits/44a02f50eac2d218e88835379d36e6bee2015e67

v2/lib/snapshot/download.zig
prestonsn: True https://github.com/Syndica/sig/pull/1405/commits/dc4a1d2ae1c5c541e4ba7596e8c3c3f75123ca6b

v2/init/main.zig
```diff
@@ -20,6 +20,7 @@ const Config = struct {
     shred_network: ShredNetwork,
 
     telemetry: Telemetry,
+    snapshot: Snapshot,
```
InKryption: Very small nit, but might make more sense to keep telemetry at the bottom, since it's not really part of the core group of validator components like how the snapshot service is.

v2/init/main.zig
```diff
@@ -34,6 +35,15 @@ const Config = struct {
     const Telemetry = struct {
         port: u16,
     };
+
+    const Snapshot = struct {
+        folder: []const u8,
+        known_validators: [][]const u8,
```
InKryption: Why is this mutable?

v2/init/services.zig
```diff
@@ -194,12 +194,22 @@ pub const Region = union(enum) {
 
     telemetry: tel.Region.Info,
 
+    snapshot_config: struct {
+        folder_path: []const u8,
+        cluster: lib.solana.Cluster,
+        known_validators: [][]const u8,
```
InKryption: Same question, why is this mutable?

v2/init/services.zig
```diff
@@ -194,12 +194,22 @@ pub const Region = union(enum) {
 
     telemetry: tel.Region.Info,
 
+    snapshot_config: struct {
```
InKryption: Same comment about keeping telemetry at the bottom.

v2/init/services.zig
```diff
@@ -240,6 +250,63 @@ pub const Region = union(enum) {
...
+                                "\"*\" takes precedence, ignoring the rest.",
+                            .{},
+                        );
+                    }
+                    data.known_validators_allow_all = true;
+                    // NOTE: we zero out known_validators_len to make it clear that no validator pubkeys were provided.
+                    data.known_validators_len = 0;
+                } else {
+                    data.known_validators_allow_all = false;
+                    data.known_validators_len = @intCast(cfg.known_validators.len);
```
InKryption, line 291: Mightn't it be simpler to just treat `known_validators = .{}` as the sign for "allow all validators"? Can add a log that notifies the operator that the set is empty, and that all validators will be allowed, and otherwise just deal with the specific set.

v2/init/services.zon
```diff
@@ -39,5 +43,13 @@
                 .{ .name = .telemetry, .access = .rw },
             },
         },
+        .{
+            .name = .snapshot,
+            .regions = .{
+                .{ .name = .snapshot_config, .access = .readonly },
+                .{ .name = .telemetry, .access = .rw },
+                .{ .name = .gossip_to_snapshot, .access = .rw },
```
InKryption: Same comment about telemetry order (don't forget to also change the field order in the parameter struct).

v2/init/services.zon
```diff
@@ -23,6 +26,7 @@
                 .{ .name = .net_to_gossip, .access = .rw },
                 .{ .name = .gossip_config, .access = .readonly },
                 .{ .name = .telemetry, .access = .rw },
+                .{ .name = .gossip_to_snapshot, .access = .rw },
```
InKryption: Here too

v2/services/snapshot.zig
InKryption: Just want to echo King's comment here about passing this around as a file handle.

v2/services/snapshot.zig
```diff
@@ -0,0 +1,90 @@
...
+            else => |e| return e,
+        };
+
+        logger.info().logf("snapshot path {s}", .{snapshot_dir});
+
+        if (try download.findExistingSnapshot(snapshot_dir)) |existing| {
+            break :result .{ .already_exists = existing };
+        }
+
+        var dedupe_fba = std.heap.FixedBufferAllocator.init(&dedupe_map_buf);
```
InKryption: Would be a good place to use that local static buffer pattern I showed you.

v2/lib/lib.zig
```diff
@@ -13,6 +7,7 @@ pub const ipc = @import("ipc.zig");
 pub const linux = @import("linux.zig");
 pub const util = @import("util.zig");
 pub const telemetry = @import("telemetry.zig");
+pub const snapshot = @import("snapshot.zig");
```
InKryption: Here too

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+const PROBE_TIMEOUT_SECS: i64 = 3;
+const PROBE_TIMEOUT: std.os.linux.kernel_timespec = .{ .sec = PROBE_TIMEOUT_SECS, .nsec = 0 };
+
+pub const MAX_DOWNLOAD_RACERS: u8 = 16;
+const MAX_DOWNLOAD_CANDIDATES: u8 = 64;
+const DOWNLOAD_RACE_THRESHOLD_PCT: u64 = 10;
+const SPLICE_CHUNK: u32 = 1 * 1024 * 1024;
+const NO_OFFSET: u64 = std.math.maxInt(u64);
+const DOWNLOAD_TIMEOUT_SECS: i64 = 3;
+const DOWNLOAD_TIMEOUT: std.os.linux.kernel_timespec = .{ .sec = DOWNLOAD_TIMEOUT_SECS, .nsec = 0 };
```
InKryption: Why are `MAX_CONCURRENT_PROBES` and `MAX_DOWNLOAD_RACERS` pub, but not `MAX_DOWNLOAD_CANDIDATES` private? Prefer to use a consistent "measure" for what should be publicized, like either of the following:
* Private unless specifically needed by external code
* Private unless it is indirectly accessible and/or needed by external code.

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+    active_offset: u16,
+    /// Timestamp captured when the current io_uring op was submitted.
+    /// Used to measure SQE to CQE received latency.
+    op_start_time: std.time.Instant,
+
+    fn init(active_offset: u16) LinkedTimeoutOp {
+        return .{
+            .active_offset = active_offset,
+            .timed_out = false,
+            .op_start_time = std.time.Instant.now() catch unreachable,
```
InKryption: Maybe want to make this a panic instead of unreachable, just to be safe.

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+    const ProbeConnecting = struct {
+        op: LinkedTimeoutOp,
+    };
+
+    const ProbeSending = struct {
+        op: LinkedTimeoutOp,
+        /// stores the http HEAD request pre-formatted to check
+        send_buf: [256]u8,
+        /// lenfth of the formatted HTTP HEAD request.
+        send_len: u16,
```
InKryption: Feel free to ignore if you prefer the "native-sized" integer type here, but could be more precise here with `send_len: std.math.IntFittingRange(0, 255)`.

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+
+        return &active.phase.sending;
+    }
+
+    fn enterReceiving(self: *ProbeConn) *ProbeReceiving {
+        const active = &self.lifecycle.active;
+        switch (active.phase) {
+            .sending => {
+                active.phase = .{ .receiving = .{
+                    .op = undefined,
```
InKryption: Is there any reason for these two functions to not just accept an `offset: u16` parameter in order to initialize the `op` field?

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+
+pub const PeerState = struct {
+    from: Pubkey,
+    slot: Slot,
+    hash: Hash,
+    probe_status: ProbeStatus,
+    latency_ms: u32,
+
+    pub fn eql(self: PeerState, other: PeerState) bool {
+        return self.slot == other.slot and std.meta.eql(self.hash, other.hash);
```
InKryption: I banish thine usage of `std.meta`
```suggestion
        return self.slot == other.slot and self.hash.eql(&other.hash);
```

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+
+        return &active.phase.splicing_out;
+    }
+
+    fn enterFsyncing(self: *DownloadConn) *DownloadFsyncing {
+        const active = &self.lifecycle.active;
+        switch (active.phase) {
+            .splicing_out => {
+                active.phase = .{ .fsyncing = .{
+                    .op_start_time = undefined,
```
InKryption: Same comment about the `op` (and in this case also the `op_start_time`) fields; can't we just initialze them here?

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+};
+
+pub const DownloadResult = union(enum) {
+    already_exists: ExistingSnapshot,
+    downloaded: CompletedDownload,
+    failed: DownloadFailureReason,
+};
+
+// TODO: Can be smaller than u8.
+const Op = enum(u8) {
```
InKryption: I don't understand the TODO. If it can be smaller than a u8, why isn't it?

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+        for (@typeInfo(Metrics).@"struct".fields) |field| {
+            if (field.type == tel.Histogram) {
+                @field(config, field.name) = .{
+                    .id_override = null,
+                    .upper_bounds = IO_LATENCY_BOUNDS,
+                };
+            }
+        }
+        break :blk config;
+    };
```
InKryption, line 737: I'll look into adding some way to overlay a default value like this.

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+};
+
+pub const DedupeMap = std.array_hash_map.ArrayHashMapUnmanaged(
```
InKryption: Why not just `std.ArrayHashMapUnmanaged`? The full qualification seems unnecessary.

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+    PeerState,
+    std.array_hash_map.AutoContext(Address),
+    true,
+);
+
+pub const Downloader = struct {
+    ring: IoUring,
+    gossip_iter: SnapshotSourceRing.Iterator(.reader),
+    dedupe_map: DedupeMap,
+    dedupe_alloc: std.mem.Allocator,
```
InKryption: Naming this as to be specifically for `dedupe_map` seems like a dangerous slippery slope to me; we've been here before with `gossip_data_allocator`, and the end-result isn't pretty. I think we should default to just having one general purpose allocator here - I would also remove this as a field, and instead just pass a `gpa: std.mem.Allocator` to the methods that need to allocate.

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+        dedupe_alloc: std.mem.Allocator,
+        known_validators: KnownValidators,
+        snapshot_dir: []const u8,
+        metrics: Metrics,
+        logger: tel.Logger("snapshot"),
+    ) !Downloader {
+        return .{
+            .ring = try IoUring.init(IO_URING_ENTRIES, 0),
+            .gossip_iter = gossip_to_snapshot.get(.reader),
+            .dedupe_map = DedupeMap{},
```
InKryption: Don't use deprecated default field inits
```suggestion
            .dedupe_map = .empty,
```

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+            if (!self.probe_conns[i].isUnused()) {
+                self.finishProbe(@intCast(i), .failed);
+            }
+        }
+
+        for (0..self.download_conns.len) |i| {
+            if (!self.download_conns[i].isUnused()) {
+                self.finishDownload(@intCast(i), .cancelled);
+            }
+        }
```
InKryption:
```suggestion
        for (&self.probe_conns, 0..) |*probe_conn, i| {
            if (!probe_conn.isUnused()) {
                self.finishProbe(@intCast(i), .failed);
            }
        }

        for (&self.download_conns, 0..) |*download_conn, i| {
            if (!download_conn.isUnused()) {
                self.finishDownload(@intCast(i), .cancelled);
            }
        }
```

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+
+            _ = try self.ring.submit_and_wait(0);
+            const n = try self.ring.copy_cqes(&cqes, 0);
+            if (n == cqes.len) {
+                self.metrics.io_uring_cqe_batch_fulls_total.increment(1);
+            }
+
+            for (cqes[0..n]) |cqe| {
+                // A previoud cqe in this copied batch may have set the final result. Stop before
+                // handling any more completions.
```
InKryption:
```suggestion
                // A previous cqe in this copied batch may have set the final result. Stop before
                // handling any more completions.
```

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+        errdefer std.posix.close(fd);
+
+        probe.gen +%= 1;
+        probe.lifecycle = .{ .active = .{
+            .state = .{
+                .fd = fd,
+                .addr = addr,
+                .from = peer.from,
+                .net_addr = net_addr,
+                .start_time = std.time.Instant.now() catch unreachable,
```
InKryption: Same comment about panicking to be safe

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+                .slot = peer.slot,
+                .hash = peer.hash,
+            },
+            .phase = .{ .connecting = .{
+                .op = undefined,
+            } },
```
InKryption, line 1124: Unclear to me when this will be defined? A comment explaining when that should be, and an explanation for why it will be before anything reads it, would be appreciated.

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+    /// NOTE: assumes SnapshotService is the only producer (thread) mutating the ring.
+    fn reserveLinkedSqes(self: *Downloader) !LinkedSqes {
+        // NOTE: it is safe to get capacity this way due to single event producing thread.
+        const capacity: u32 = @intCast(self.ring.sq.sqes.len);
+        const ready = self.ring.sq_ready();
+
+        std.debug.assert(ready <= capacity);
+
+        if (capacity - ready < 2) {
+            return error.SubmissionQueueFull;
```
InKryption: Explicit error set pls, preferably one which you then explicitly propagate where relevant.

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+                // Same addr + same slot/hash. Do not retry failed probes here.
+                // A failed probe is terminal for this snapshot candidate. It will
+                // be retried only if slot/hash changes (the update path above).
+                self.metrics.snapshot_sources_deduped.increment(1);
+            }
+        }
+        if (drained) self.gossip_iter.markUsed();
+    }
+
+    fn startProbe(self: *Downloader, addr: Address, peer: *PeerState) !void {
```
InKryption: Since this is `catch`'d, would be good to describe its error set so it's obvious what we're catching (the Submission error + the socket error).

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+            if (p.isUnused()) break .{ @intCast(i), p };
+        } else return;
+
+        // create tcp socket
+        const net_addr = addr.toNetAddress();
+        const fd = try std.posix.socket(
+            net_addr.any.family,
+            std.posix.SOCK.STREAM | std.posix.SOCK.NONBLOCK,
+            0,
+        );
```
InKryption: In relation to the comment about the error set, I think it would be useful to do `catch return error.SocketOpenFailed` instead of `try`.

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+    }
+
+    /// Enqueues a pair of SQEs to connect and perform the tcp handshake along with a timeout.
+    fn queueProbeConnectWithTimeout(
+        self: *Downloader,
+        index: u8,
+        fd: std.posix.fd_t,
+    ) !void {
+        const probe = &self.probe_conns[index];
+        const active = probe.activePtr() orelse unreachable;
```
InKryption, line 1194: Doc comment about the assumption of this `orelse unreachable` would be good.

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+    ) !void {
+        const probe = &self.probe_conns[data.index];
+
+        var send_data = data;
+        send_data.op = .probe_send;
+        send_data.offset = offset;
+
+        const timeout_data = send_data.timeout();
+
+        const active = probe.activePtr() orelse unreachable;
```
InKryption, line 1230: Same here (doc comment about the unreachable's assumption).

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+    ) !void {
+        const probe = &self.probe_conns[data.index];
+
+        var recv_data = data;
+        recv_data.op = .probe_recv;
+        recv_data.offset = offset;
+
+        const timeout_data = recv_data.timeout();
+
+        const active = probe.activePtr() orelse unreachable;
```
InKryption, line 1277: Another comment about the unreachable's assumption.

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+                    active.from, @tagName(cqe.err()),
+                });
+            },
+        }
+    }
+
+    fn handleProbeConnect(self: *Downloader, data: UserData, cqe: std.os.linux.io_uring_cqe) void {
+        const probe = self.getProbeForCqe(data) orelse return;
+        const active = probe.activePtr() orelse return;
+        if (probe.lifecycle.active.phase != .connecting) {
```
InKryption: Should this not just be
```suggestion
        const active = probe.activePtr() orelse return;
        if (active.phase != .connecting) {
```
?
(Same for a bunch of other code adjacent to this)

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+    }
+
+    /// Enqueues a pair of SQEs to send the HTTP HEAD request along with a timeout for the corresponding probe.
+    /// After which we update the probe's phase to `sending`.
+    fn queueProbeSendWithTimeout(
+        self: *Downloader,
+        data: UserData,
+        /// used to track partial completions.
+        offset: u16,
+    ) !void {
```
InKryption: This should also specify an explicit error set, since it's explicitly caught.

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+            self.probeNextPending();
+            return;
+        }
+
+        self.onOpComplete(.probe_send, sending.op.op_start_time);
+
+        // Track send progress across partial completions via user_data offset.
+        std.debug.assert(cqe.res > 0);
+        std.debug.assert(cqe.res <= sending.send_len - data.offset);
+        const new_offset = data.offset + @as(u16, @intCast(cqe.res));
```
InKryption:
```suggestion
        const new_offset: u16 = @intCast(data.offset + cqe.res);
```

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+            self.logger.warn().logf("probe recv eof from={f}", .{active.from});
+            self.finishProbe(data.index, .failed);
+            self.probeNextPending();
+            return;
+        }
+
+        self.onOpComplete(.probe_recv, receiving.op.op_start_time);
+
+        // Track recv progress across partial completions via user_data offset.
+        const new_offset = data.offset + @as(u16, @intCast(cqe.res));
```
InKryption:
```suggestion
        const new_offset: u16 = @intCast(data.offset + cqe.res);
```

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+        // Parse Content-Length from headers.
+        const content_len = parseContentLength(response[0 .. header_end + 4]) orelse {
+            self.logger.warn().logf("probe missing/zero content-length from={f}", .{active.from});
+            self.finishProbe(data.index, .failed);
+            self.probeNextPending();
+            return;
+        };
+
+        // Compute and store latency.
+        const elapsed_ns = (std.time.Instant.now() catch unreachable).since(active.start_time);
```
InKryption: Should also panic instead of unreachable

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+                    .failed => .failed,
+                    .sq_full => .pending,
+                };
+            }
+        }
+
+        std.posix.close(active.fd);
+        const old_gen = probe.gen;
+        self.probe_conns[probe_index] = .empty();
+        self.probe_conns[probe_index].gen = old_gen;
```
InKryption:
```suggestion
        const probe = &self.probe_conns[probe_index];
        // If this fn gets called on a probe that's .unused, then return immidiately.
        const active = probe.activePtr() orelse return;
        if (self.dedupe_map.getPtr(active.addr)) |peer| {
            // NOTE: We gaurd the update here since gossip can update peer in dedupe map while io_uring was completing this probe.
            // If it was updated underneath us, don't update things here since another/newer probe was alread issued to io_uring.
            if (peer.slot == active.slot and std.meta.eql(peer.hash, active.hash)) {
                peer.probe_status = switch (result) {
                    .succeeded => .succeeded,
                    .failed => .failed,
                    .sq_full => .pending,
                };
            }
        }

        std.posix.close(active.fd);
        const old_gen = probe.gen;
        probe.* = .empty();
        probe.gen = old_gen;
```

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+        const conn = &self.download_conns[data.index];
+
+        if (conn.isUnused() or conn.gen != data.gen) return;
+
+        const active_download = switch (conn.lifecycle) {
+            .unused => return,
+            .active => |*active| active,
+        };
+        const expected_phase = data.op.expectedDownloadPhase() orelse unreachable;
+        if (std.meta.activeTag(active_download.phase) != expected_phase) return;
```
InKryption: `expected_phase` is already the enum, there's no reason to use `meta.activeTag` here.
```suggestion
        if (active_download.phase != expected_phase) return;
```

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+        const conn = self.getDownloadForCqe(data) orelse return;
+        const active = conn.activePtr() orelse return;
+        if (conn.lifecycle.active.phase != .connecting) {
+            self.logger.warn().logf(
+                "unexpected download connect cqe from={f} phase={s}",
+                .{ active.from, conn.phaseName() },
+            );
+            return;
+        }
+        const connecting = &conn.lifecycle.active.phase.connecting;
```
InKryption:
```suggestion
        const active = conn.activePtr() orelse return;
        if (conn.lifecycle.active.phase != .connecting) {
            self.logger.warn().logf(
                "unexpected download connect cqe from={f} phase={s}",
                .{ active.from, conn.phaseName() },
            );
            return;
        }
        const connecting = &active.phase.connecting;
```
?

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+        const conn = self.getDownloadForCqe(data) orelse return;
+        const active = conn.activePtr() orelse return;
+        if (conn.lifecycle.active.phase != .sending) {
+            self.logger.warn().logf(
+                "unexpected download send cqe from={f} phase={s}",
+                .{ active.from, conn.phaseName() },
+            );
+            return;
+        }
+        const sending = &conn.lifecycle.active.phase.sending;
```
InKryption:
```suggestion
        const active = conn.activePtr() orelse return;
        if (conn.lifecycle.active.phase != .sending) {
            self.logger.warn().logf(
                "unexpected download send cqe from={f} phase={s}",
                .{ active.from, conn.phaseName() },
            );
            return;
        }
        const sending = &active.phase.sending;
```
?

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+            self.finishDownload(data.index, .failed);
+            self.startPendingRacers();
+            return;
+        }
+
+        self.onOpComplete(.download_send, sending.op.op_start_time);
+
+        std.debug.assert(cqe.res > 0);
+        std.debug.assert(cqe.res <= sending.send_len - data.offset);
+        const new_offset = data.offset + @as(u16, @intCast(cqe.res));
```
InKryption:
```suggestion
        const new_offset: u16 = @intCast(data.offset + cqe.res);
```

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+        sqes.timeout.user_data = timeout_data.encode();
+
+        sending.op = LinkedTimeoutOp.init(offset);
+    }
+
+    fn queueDownloadRecvHeadersWithTimeout(
+        self: *Downloader,
+        data: UserData,
+        offset: u16,
+    ) !void {
```
InKryption: Should also get an explicit error set

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+            return;
+        }
+
+        // Detect extra body bytes received after the header terminator.
+        const body_start: u16 = @intCast(header_end + 4);
+        const extra_len: u16 = new_offset - body_start;
+
+        // If we got body bytes during this header parse phase, start writing em to the file.
+        if (extra_len > 0) {
+            _ = conn.enterWritingExtra(body_start, extra_len);
```
InKryption, line 1845: Doesn't this discard a pointer to something that still needs to have the `op` field initialized?

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+        std.debug.assert(buf_end <= writing_extra.recv_len);
+
+        sqe.prep_write(
+            active.file_fd,
+            writing_extra.recv_buf[buf_start..buf_end],
+            active.bytes_written + written,
+        );
+        sqe.user_data = write_data.encode();
+
+        writing_extra.op_start_time = std.time.Instant.now() catch unreachable;
```
InKryption: Panic instead of unreachable

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+        sqes.timeout.prep_link_timeout(&DOWNLOAD_TIMEOUT, 0);
+        sqes.timeout.user_data = timeout_data.encode();
+
+        reading_headers.op = LinkedTimeoutOp.init(offset);
+    }
+
+    /// handles HTTP edge case where the header recv can read past `\r\n\r\n` and already consume some snapshot body bytes into recv_buf.
+    fn queueDownloadWriteExtra(self: *Downloader, data: UserData, written: u16) !void {
+        const conn = &self.download_conns[data.index];
+        const active = conn.activePtr() orelse unreachable;
```
InKryption: Should get an explicit error set, and a doc comment about the assumption implied by the `unreachable` here.

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+        if (self.download_race.phase == .idle) {
+            self.download_race = .empty();
+            self.download_race.phase = .racing;
+            self.download_race.slot = slot;
+            self.download_race.hash = hash;
+        }
+
+        // Only accept candidates matching the current race target.
+        if (self.download_race.slot != slot) return;
+        if (!std.meta.eql(self.download_race.hash, hash)) return;
```
InKryption:
```suggestion
        if (!self.download_race.hash.eql(&hash)) return;
```

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+        if (self.download_race.phase != .racing) return;
+
+        while (self.active_downloads < MAX_DOWNLOAD_RACERS) {
+            const candidate_index = self.nextUnstartedCandidate() orelse return;
+
+            self.startDownloadRacer(candidate_index) catch |err| {
+                self.logger.warn().logf("failed to start download racer from={f} err={}", .{
+                    self.download_race.candidates[candidate_index].from, err,
+                });
+                self.download_race.candidates[candidate_index].started = true;
```
InKryption:
```suggestion
                const candidate = &self.download_race.candidates[candidate_index];
                self.logger.warn().logf("failed to start download racer from={f} err={}", .{
                    candidate.from, err,
                });
                candidate.started = true;
```

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+                .net_addr = net_addr,
+                .content_len = candidate.content_len,
+                .bytes_written = 0,
+                .pipe_pending = 0,
+                .op_seq = 0,
+                .cancel_requested = false,
+            },
+            .phase = .{ .connecting = .{
+                .op = undefined,
+            } },
```
InKryption, line 2551: Also not clear to me when this is defined

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+            @as(u64, 1),
+            active.content_len * DOWNLOAD_RACE_THRESHOLD_PCT / 100,
+        );
+
+        if (active.bytes_written < threshold) return;
+
+        self.download_race.phase = .winner_selected;
+        self.download_race.winner_index = index;
+        self.download_race.progress_sample = .{
+            .time = std.time.Instant.now() catch unreachable,
```
InKryption: unreachable -> panic

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+    }
+
+    fn maybeLogWinnerProgress(self: *Downloader) void {
+        if (self.download_race.phase != .winner_selected) return;
+
+        const winner_index = self.download_race.winner_index orelse return;
+        const active = self.download_conns[winner_index].activePtr() orelse return;
+        if (active.content_len == 0) return;
+
+        const now = std.time.Instant.now() catch unreachable;
```
InKryption: unreachable -> panic

v2/lib/snapshot/download.zig
InKryption: Yeah this fine

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2602 @@
...
+    }
+
+    // TODO: Maybe move these into some kinda effects interface struct for tests?
+    fn onOpTimeout(self: *Downloader, _: Op) void {
+        self.metrics.io_uring_timeouts_total.increment(1);
+    }
+
+    fn onOpComplete(self: *Downloader, op: Op, op_start: std.time.Instant) void {
+        const elapsed_ns: f64 = @floatFromInt(
+            (std.time.Instant.now() catch unreachable).since(op_start),
```
InKryption: unreachable -> panic

InKryption (changes requested):

InKryption (changes requested):
A slew of comments (many repetitions of the same basic change request).

v2/services/snapshot.zig
prestonsn: https://github.com/Syndica/sig/pull/1405/commits/fbe757f7654c74fd9027f983a71af7cbf000d2aa

Swapped over to use this: https://ziglang.org/documentation/0.15.2/std/#std.fs.Dir.makeOpenPath

v2/lib/snapshot/download.zig
kprotty: we should probably have a `lib.clock` for this (+ realtime timestamps used in gossip)

v2/lib/snapshot/download.zig
prestonsn: Yeah, good point. Way I had it was new slot+hash was raced only if existing one's race totally failed.

Changed the behavior to instead store slot+hash on `DownloadCandidate`, this allows race to accept candidates for different snapshot targets concurrently (instead of only for one pre-selected/global slot+hash target): https://github.com/Syndica/sig/pull/1405/commits/74010745845f9bcbc1efaf809518a984184e19cd

v2/lib/snapshot/download.zig
prestonsn: Removed the comments: https://github.com/Syndica/sig/pull/1405/changes/fbe757f7654c74fd9027f983a71af7cbf000d2aa

v2/lib/snapshot/download.zig
prestonsn: Yeah, swapped out for custom map: https://github.com/Syndica/sig/pull/1405/commits/c2c42a8029d185300e57d13779f171a20b547600

v2/init/services.zig
prestonsn: True, def simpler that way. But I wanted to make the opt-in for trusting all gossip peers explicit. I just prefer the explicitness in configs, rather than having to infer/trust default behavior.  

That said I can go either ways on this.

v2/lib/snapshot/download.zig
prestonsn: That's fair. I ended up removing the map entirely and using a custom one instead to avoid the edge-case of running out of memory with a std map: https://github.com/Syndica/sig/pull/1405/commits/c2c42a8029d185300e57d13779f171a20b547600

v2/lib/snapshot/download.zig
prestonsn: This is a good point, i guess another reason to remove the reliance on std's map, especially for this use case. The custom array backed map added instead doesn't allocate dynamically, so the allocator param got removed altogether: https://github.com/Syndica/sig/pull/1405/commits/c2c42a8029d185300e57d13779f171a20b547600

v2/services/snapshot.zig
prestonsn: 100% should have refactored to use that strat for grouping the static bufs under one label. Ended up removing the FBA usage entirely instead: https://github.com/Syndica/sig/pull/1405/commits/c2c42a8029d185300e57d13779f171a20b547600

v2/lib/snapshot/download.zig
prestonsn: That would be sweet! 

It would also be nice if we could support labels on histograms too. Where each label is really just an entire histogram (similar to[`HistogramVec`](https://docs.rs/prometheus/latest/prometheus/type.HistogramVec.html)). That way each of child histograms can just share the same bucket distributions. 

I think it's also overall a smaller payload for encoding when serving Prometheus snapshots than if each histogram was it's own distinct one in the payload (though likely only slightly).

v2/lib/snapshot/download.zig
prestonsn: Fixed this by adding a .empty() to the new DedupeMap: https://github.com/Syndica/sig/pull/1405/commits/3f69a0507af9a08b8f816371c3bb31a781cd781e

v2/init/main.zig
prestonsn: That's cleaner, moved things around here: https://github.com/Syndica/sig/pull/1405/commits/87d674e1cb0b629ebe591d9f6ca72ef25f1381c6

v2/init/main.zig
prestonsn: Should not have been: https://github.com/Syndica/sig/pull/1405/commits/87d674e1cb0b629ebe591d9f6ca72ef25f1381c6

v2/init/services.zig
prestonsn: Also should not have been: https://github.com/Syndica/sig/pull/1405/commits/87d674e1cb0b629ebe591d9f6ca72ef25f1381c6

v2/init/services.zig
prestonsn: Fixed telemetry order here: https://github.com/Syndica/sig/pull/1405/commits/87d674e1cb0b629ebe591d9f6ca72ef25f1381c6

v2/init/services.zon
prestonsn: Fixed telemetry order here: https://github.com/Syndica/sig/pull/1405/commits/87d674e1cb0b629ebe591d9f6ca72ef25f1381c6

v2/init/services.zon
prestonsn: Fixed telemetry order here: https://github.com/Syndica/sig/pull/1405/commits/87d674e1cb0b629ebe591d9f6ca72ef25f1381c6

v2/lib/lib.zig
prestonsn: Fixed telemetry order here: https://github.com/Syndica/sig/pull/1405/commits/87d674e1cb0b629ebe591d9f6ca72ef25f1381c6

v2/lib/snapshot/download.zig
prestonsn: This is good advice, thanks. I just forgot to make them private again after a refactor where I moved more code out of `services/` and into `lib/snapshot/download.zig`. Fixed: https://github.com/Syndica/sig/pull/1405/commits/87d674e1cb0b629ebe591d9f6ca72ef25f1381c6

v2/lib/snapshot/download.zig
prestonsn: Fixed as part of these changes: https://github.com/Syndica/sig/pull/1405/commits/87d674e1cb0b629ebe591d9f6ca72ef25f1381c6

v2/lib/snapshot/download.zig
prestonsn: oops, fixed here: https://github.com/Syndica/sig/pull/1405/commits/87d674e1cb0b629ebe591d9f6ca72ef25f1381c6

v2/lib/snapshot/download.zig
prestonsn: Oops, fixed this here: https://github.com/Syndica/sig/pull/1405/commits/4edfb162a1e22de11134b3e84973c30fbad8ad98

v2/lib/snapshot/download.zig
prestonsn: :skull: fax, fixed: https://github.com/Syndica/sig/pull/1405/commits/4edfb162a1e22de11134b3e84973c30fbad8ad98

v2/lib/snapshot/download.zig
prestonsn: I have quite forsaken the use of `std.meta`: https://github.com/Syndica/sig/commit/4edfb162a1e22de11134b3e84973c30fbad8ad98

v2/lib/snapshot/download.zig
prestonsn: Fixed as part of these changes: https://github.com/Syndica/sig/pull/1405/commits/4edfb162a1e22de11134b3e84973c30fbad8ad98

v2/lib/snapshot/download.zig
prestonsn: Fixed as part of these changes: https://github.com/Syndica/sig/pull/1405/commits/4edfb162a1e22de11134b3e84973c30fbad8ad98

v2/lib/snapshot/download.zig
prestonsn: Fixed as part of these changes: https://github.com/Syndica/sig/pull/1405/commits/4edfb162a1e22de11134b3e84973c30fbad8ad98

v2/services/snapshot.zig
```diff
@@ -0,0 +1,90 @@
...
+            const path = std.fmt.bufPrint(
+                &path_buf,
+                "{f}",
+                .{std.fs.path.fmtJoin(&.{ snapshot_dir, snapshot.name() })},
+            ) catch unreachable;
+            logger.info().logf("snapshot download completed slot={d} hash={f} path={s}", .{
+                snapshot.slot,
+                snapshot.hash,
+                path,
+            });
```
kprotty: instead of creating new path_buf, can print directly: `path={s}/{f}` ++ `.{ snapshot_dir, snapshot }` (add `fn format(..)` method on this `snapshot` type so it writes its file name when printed instead of returning a large buf)

v2/lib/linux.zig
```diff
@@ -291,6 +291,19 @@ pub const bpf = struct {
             allowSyscall(@intFromEnum(syscalls.readv)) ++
             allowSyscall(@intFromEnum(syscalls.sendmsg)) ++
             allowSyscall(@intFromEnum(syscalls.setsockopt)) ++
+            // snapshot
+            allowSyscall(@intFromEnum(syscalls.io_uring_setup)) ++
+            allowSyscall(@intFromEnum(syscalls.io_uring_enter)) ++
+            allowSyscall(@intFromEnum(syscalls.io_uring_register)) ++
+            allowSyscall(@intFromEnum(syscalls.mmap)) ++
+            allowSyscall(@intFromEnum(syscalls.munmap)) ++
```
kprotty, line 298: dont need `io_uring_register` (thats for registered buffer pooling) and `mmap/munmap`

v2/lib/snapshot/download.zig
```diff
@@ -0,0 +1,2850 @@
...
+    name_buf: [std.fs.max_path_bytes]u8,
+    name_len: usize,
+
+    slot: Slot,
+    hash: Hash,
+
+    pub fn name(self: *const CompletedDownload) []const u8 {
+        return self.name_buf[0..self.name_len];
+    }
+};
```
kprotty: Passing/copying 4kb types around isnt great. These should instead be stored as just their slot + hash, then have a custom `format()` method to write themselves as the filename when printing (to open, it prints to a local buf once, then uses that)

v2/lib/snapshot/download.zig
prestonsn: Swapped this out for our own linux clock helper instead: https://github.com/Syndica/sig/pull/1405/commits/b4a147f2561ac1d7eae3272e598627639e617242

v2/lib/snapshot/download.zig
prestonsn: Agreed, added a linux clock helper under `lib.clock`. This also moves the panic to exactly where the issue is (i.e platform does not support `MONOTONIC` for whatever reason, not a situation we care about): https://github.com/Syndica/sig/pull/1405/commits/b4a147f2561ac1d7eae3272e598627639e617242

v2/lib/snapshot/download.zig
prestonsn: True, made improvements to this as part of these changes: https://github.com/Syndica/sig/pull/1405/commits/b4a147f2561ac1d7eae3272e598627639e617242

v2/lib/snapshot/download.zig
prestonsn: I went ahead and added a `.empty()` for `LinkedTimeoutOp` to better indicate that the timeout op state is just being constructed for now, but it's actual state is not set until `initOffset` is called on it: https://github.com/Syndica/sig/pull/1405/commits/b4a147f2561ac1d7eae3272e598627639e617242

v2/lib/snapshot/download.zig
prestonsn: True, the pointer was getting discarded. Good catch! 

https://github.com/Syndica/sig/pull/1405/commits/b4a147f2561ac1d7eae3272e598627639e617242 changes things so that the th phase state is fully initialized before being entered.

v2/lib/snapshot/download.zig
prestonsn: Now changed this to be initialized, with actual values set subsequently:  https://github.com/Syndica/sig/pull/1405/commits/b4a147f2561ac1d7eae3272e598627639e617242

v2/lib/snapshot/download.zig
prestonsn: Changed with the new clock: https://github.com/Syndica/sig/pull/1405/commits/b4a147f2561ac1d7eae3272e598627639e617242

v2/lib/snapshot/download.zig
prestonsn: Changed with the new clock: https://github.com/Syndica/sig/pull/1405/commits/b4a147f2561ac1d7eae3272e598627639e617242

v2/lib/snapshot/download.zig
prestonsn: Changed with the new clock: https://github.com/Syndica/sig/pull/1405/commits/b4a147f2561ac1d7eae3272e598627639e617242

v2/lib/snapshot/download.zig
prestonsn: Remove with: https://github.com/Syndica/sig/pull/1405/commits/b4a147f2561ac1d7eae3272e598627639e617242

v2/lib/snapshot/download.zig
prestonsn: Removed with: https://github.com/Syndica/sig/pull/1405/commits/b4a147f2561ac1d7eae3272e598627639e617242

v2/lib/snapshot/download.zig
prestonsn: Changed so that code now initializes those fields at phase creation (we now use `.op = .empty()` and .op_start_time = lib.clock.monotonic(.ns)`: https://github.com/Syndica/sig/pull/1405/changes/b4a147f2561ac1d7eae3272e598627639e617242

v2/lib/snapshot/download.zig
prestonsn: integrated the new `lib.clock` with gossip and telemetry where relevant: https://github.com/Syndica/sig/pull/1405/commits/7c79006d5593100d1137865beda6cd98518e0ba8

v2/lib/snapshot/download.zig
prestonsn: Swapped to explicit set: https://github.com/Syndica/sig/pull/1405/commits/25d2de54ee321f1525f3bf02a1a7e8159232576c

v2/lib/snapshot/download.zig
prestonsn: Swapped to explicit set: https://github.com/Syndica/sig/pull/1405/commits/25d2de54ee321f1525f3bf02a1a7e8159232576c

v2/lib/snapshot/download.zig
prestonsn: Agreed, swapped to explicit set: https://github.com/Syndica/sig/pull/1405/commits/25d2de54ee321f1525f3bf02a1a7e8159232576c

v2/lib/snapshot/download.zig
prestonsn: Documented each of these lines: https://github.com/Syndica/sig/pull/1405/commits/abf7ffdc0d41050bc0dc8627b509d1deba0facc2

v2/lib/snapshot/download.zig
prestonsn: https://github.com/Syndica/sig/pull/1405/commits/abf7ffdc0d41050bc0dc8627b509d1deba0facc2

v2/lib/snapshot/download.zig
prestonsn: https://github.com/Syndica/sig/pull/1405/commits/abf7ffdc0d41050bc0dc8627b509d1deba0facc2

v2/lib/snapshot/download.zig
prestonsn: https://github.com/Syndica/sig/pull/1405/commits/25d2de54ee321f1525f3bf02a1a7e8159232576c

v2/lib/snapshot/download.zig
prestonsn: https://github.com/Syndica/sig/pull/1405/commits/25d2de54ee321f1525f3bf02a1a7e8159232576c

v2/lib/snapshot/download.zig
prestonsn: Explicit errors: https://github.com/Syndica/sig/pull/1405/commits/25d2de54ee321f1525f3bf02a1a7e8159232576c
Doc comments for `unreachable`: https://github.com/Syndica/sig/pull/1405/commits/abf7ffdc0d41050bc0dc8627b509d1deba0facc2

v2/lib/snapshot/download.zig
prestonsn: Removed all use of std.meta.eql: https://github.com/Syndica/sig/commit/4edfb162a1e22de11134b3e84973c30fbad8ad98

v2/lib/snapshot/download.zig
prestonsn: Yeah, the names I gave those types are bit confusing. But those are distinct types. 

I changed things a bit and I've made `.activePtr` return a pointer to the underlying `ActiveDownload` struct (instead of just one field on that struct). Should make things more clear/readable: https://github.com/Syndica/sig/pull/1405/commits/463dcb1529f7f0d00cd178ab56bdbb291befce5a

v2/lib/snapshot/download.zig
prestonsn: Cleaned up a bit (those aren't the same types, but is confusing so refactored it): https://github.com/Syndica/sig/pull/1405/commits/463dcb1529f7f0d00cd178ab56bdbb291befce5a

v2/lib/snapshot/download.zig
prestonsn: Cleaned up a bit (those aren't the same types, but is confusing so refactored it): https://github.com/Syndica/sig/pull/1405/commits/463dcb1529f7f0d00cd178ab56bdbb291befce5a

v2/lib/snapshot/download.zig
prestonsn: These changes should also help simplify things:  https://github.com/Syndica/sig/pull/1405/commits/463dcb1529f7f0d00cd178ab56bdbb291befce5a

v2/lib/snapshot/download.zig
prestonsn: Addressed as part of this change: https://github.com/Syndica/sig/commit/25d2de54ee321f1525f3bf02a1a7e8159232576c

v2/lib/snapshot/download.zig
prestonsn: That's pretty neat! I kinda like specifying the exact type over this. I did swap over to a u8: https://github.com/Syndica/sig/pull/1405/commits/d2e3373bac1616772a7affba967765278525eebf

v2/lib/snapshot/download.zig
prestonsn: good shout, fixed this up here: https://github.com/Syndica/sig/pull/1405/commits/d2e3373bac1616772a7affba967765278525eebf

v2/init/services.zig
InKryption: After meditating on it a bit, I realize I don't feel that strongly about it; if you tnink it's useful enough to justify the small complexity, then it's fine.

v2/lib/snapshot/download.zig
InKryption: Isn't `u8` wrong here, since it can't actually represent `len = capacity = 256`?

v2/lib/linux.zig
prestonsn: Makes sense for `io_uring_register`, but `mmap`/`munmap` are used for constructor/destructor for `IoUring` (for the submission/completion queues themselves):
* constructor: https://ziglang.org/documentation/0.15.2/std/#std.os.linux.IoUring.init_params
* destructors that the ring calls: https://ziglang.org/documentation/0.15.2/std/#std.os.linux.IoUring.SubmissionQueue.deinit

removed `io_uring_register`: https://github.com/Syndica/sig/pull/1405/commits/153dfcdf966a2229642db070d96a6b090f1ab72f

v2/lib/snapshot/download.zig
InKryption: Would also still like to swap away from propagating `std.posix` errors directly to doing `catch return error.<SpecificName>`.

v2/lib/snapshot/download.zig
prestonsn: Ye good call, removed them. Added the `format()` methods: https://github.com/Syndica/sig/pull/1405/commits/7aeec62a50907128ce930ff60cbfd73df47480b3

v2/services/snapshot.zig
prestonsn: Handled as part of this change: https://github.com/Syndica/sig/pull/1405/commits/7aeec62a50907128ce930ff60cbfd73df47480b3

v2/lib/snapshot/download.zig
prestonsn: Gotcha, I wanted the granularity for logging, but I think that requires more annoying change. Swapped over to just returning our own for now: https://github.com/Syndica/sig/pull/1405/commits/9ccb7f5d035253dbdf130a4ca538faf93e7ebc1b

v2/lib/snapshot/download.zig
prestonsn: Yes `u8` was totally wrong lol. Went with `std.math.IntFittingRange`: https://github.com/Syndica/sig/pull/1405/commits/9ccb7f5d035253dbdf130a4ca538faf93e7ebc1b

v2/lib/linux.zig
kprotty: ah, true

kprotty (approved):

InKryption (approved):

PR #1410, opened 2026-04-23
fix(v2): error dump for threaded mode

### LINKED ISSUE #1377: feat(v2): report service crashes in non-sandboxed mode
Currently the crash handling/reporting code seems broken in non-sandboxed mode; services can crash silently. The fix is probably simple but I've yet to work on it. 

error traces are returned fine, but panics are lost 

The source of the problem:
```zig
fn threadEntry(
    entry_point: ServiceEntrypoint,
    service: Service,
    args: lib.ipc.ResolvedArgs,
    service_idx: u16,
    finished_idx: *std.atomic.Value(u16),
    reset_event: *std.Thread.ResetEvent,
) void {
    switch (service) {
        inline else => |svc| tracy.setThreadName("svc: " ++ @tagName(svc)),
    }

    entry_point(args); // <- crashes

    // never happens
    reset_event.set(); 
    finished_idx.store(service_idx, .seq_cst);
}
```

jbuckmccready:
Fixes panics silently killing service threads when running in threaded mode by ensuring monitoring thread is notified in panic handler. Closes https://github.com/Syndica/sig/issues/1377

Also this fixes ordering bug where thread was notified before setting value:

Before:
```zig
    reset_event.set();
    finished_idx.store(service_idx, .seq_cst);
```

After:
```zig
    finished_idx.store(service_idx, .seq_cst);
    reset_event.set();
```

Note `panic_state` is global per service and not global to Sig, even in threaded mode, as it is compiled into each service library separately.

Tested:
- Added panic to `serviceMain` in `v2/services/net.zig` before this PR's changes to confirm silent failure and Sig continues running
- Added this PR's changes and ran to confirm things still run as expected with no panic
- Aded this PR's changes and Added panic to `serviceMain` in `v2/services/net.zig` to confirm panic now gets reported and Sig process exits.


v2/lib/ipc.zig
```diff
@@ -11,6 +11,12 @@ pub const ResolvedArgs = extern struct {
     ro: [max_regions]?[*]align(page_size_min) const u8,
     ro_len: [max_regions]usize,
 
+    /// Only set when services are running as threads inside init.
+    thread_crash_ctx: ?*anyopaque,
+    thread_crash_fn: ?ThreadCrashFn,
+    service_idx: u16,
```
Sobeston: Not sure why we're using anyopaque here?

v2/init/services.zig
```diff
@@ -763,8 +792,12 @@ fn spawnServiceNoSandbox(
     service_idx: u16,
     finished_idx: *std.atomic.Value(u16),
     reset_event: *std.Thread.ResetEvent,
+    thread_crash_ctx: *const ThreadCrashContext,
 ) !std.Thread {
-    const resolved_args = try resolveArgs(exit, stderr, regions);
+    var resolved_args = try resolveArgs(exit, stderr, regions);
+    resolved_args.thread_crash_ctx = @ptrCast(@constCast(thread_crash_ctx));
+    resolved_args.thread_crash_fn = signalThreadCrash;
+    resolved_args.service_idx = service_idx;
```
Sobeston: we should probably never use constCast - it also doesn't make sense in this case as we know that it may be mutated

Sobeston (changes requested):
This does work nicely, thanks for this. My only issue is how the ThreadContext is being handled / passed around

v2/lib/ipc.zig
jbuckmccready: Got over zealous, should have marked draft. Changed to use actual type for argument and moved the type to where it actually is used to avoid unneeded import cycle.
https://github.com/Syndica/sig/pull/1410/changes/d9c40417b395c398c53c05537d485b4582f98d9c

v2/init/services.zig
jbuckmccready: Yeah, it was half assed. I changed it so the function arg is now const pointer, fields are pointers to things that are mutated, but the pointed to struct shouldn't ever be mutated under this approach.

https://github.com/Syndica/sig/commit/d9c40417b395c398c53c05537d485b4582f98d9c

It's a tiny struct so maybe it should just passed by value. `* const T` vs `T` argument for small structs always seems to come up as a meaningless point of optionality. Often compiles to same thing if inlined, and if not inlined then is not called enough to matter. I thought I had escaped it going from C++ to Rust with moves and explicit Copy trait, but it haunts me still.

prestonsn (dismissed):
Clean 🫡 , same trace doesn't show up in main.

```bash
❯ zig build run -- config/testnet.zig.zon
info: config: .{ .sandboxing_mode = .threaded, .cluster = .testnet, .leader_schedule_file = { 115, 99, 104, 101, 100, 117, 108, 101, 46, 116, 120, 116 }, .gossip = .{ .port = 8001 }, .shred_network = .{ .recv_port = 8002 }, .telemetry = .{ .port = 12345 } }
info: Initialising: .net_pair
info: Initialised: Region `net_pair` shared with [ net_0 (rw), shred_receiver_0 (rw), ]
info: Initialising: .shred_recv_config
info: Initialised: Region `shred_recv_config` shared with [ shred_receiver_0 (ro), ]
info: Initialising: .net_pair
info: Initialised: Region `net_pair` shared with [ net_0 (rw), gossip_0 (rw), ]
info: Initialising: .gossip_config
info: Initialised: Region `gossip_config` shared with [ gossip_0 (ro), ]
info: Initialising: .telemetry
info: Initialised: Region `telemetry` shared with [ net_0 (rw), gossip_0 (rw), shred_receiver_0 (rw), telemetry_0 (rw), ]
error: Service `net_0` (pid: 0) panicked with message: test: threaded panic notification
error: Error trace:
error: Stack trace:
time=2026-04-24T00:19:26.978Z level=info scope=main service=shred_receiver message="Waiting for shreds on port 8002"
time=2026-04-24T00:19:26.978Z level=info scope=main service=net message="binding 0.0.0.0:8002"
time=2026-04-24T00:19:26.978Z level=info scope=main service=net message="binding 0.0.0.0:8001"
time=2026-04-24T00:19:26.978Z level=info scope=main service=gossip message="Gossip started on :8001 as BN5JxE85PMX49UuFVmH9yQAYfMUJ7y1ooedChfoK8pTo:
        shred_version:57087
        entrypoints:{ 35.203.170.30:8001 }"
/home/preston/dev/sig/v2/services/net.zig:119:9: 0x12c36a0 in mainInner (net)
        @panic("test: threaded panic notification");
        ^
/home/preston/dev/sig/v2/services/net.zig:33:18: 0x12c0d3a in serviceMain (net)
    try mainInner(
                 ^
/home/preston/dev/sig/v2/init/start_service.zig:105:21: 0x12c03dc in serviceMain (net)
    root.serviceMain(ro, rw) catch |err| {
                    ^
/home/preston/dev/sig/v2/init/services.zig:779:16: 0x121c965 in threadEntry (sig-init)
    entry_point(args);
               ^
/home/preston/.config/Code/User/globalStorage/ziglang.vscode-zig/zig/x86_64-linux-0.15.2/lib/std/Thread.zig:509:13: 0x11f1b4b in callFn__anon_33862 (sig-init)
            @call(.auto, f, args);
            ^
/home/preston/.config/Code/User/globalStorage/ziglang.vscode-zig/zig/x86_64-linux-0.15.2/lib/std/Thread.zig:1381:30: 0x11cc0d1 in entryFn (sig-init)
                return callFn(f, self.fn_args);
                             ^
/home/preston/.config/Code/User/globalStorage/ziglang.vscode-zig/zig/x86_64-linux-0.15.2/lib/std/os/linux/x86_64.zig:119:5: 0x11f1bd1 in clone (sig-init)
    asm volatile (
    ^
```

v2/init/services.zig
Sobeston: I think T would be a bit cleaner here but either is fine

v2/lib/ipc.zig
```diff
@@ -2,6 +2,11 @@ const std = @import("std");
 
 pub const Ring = @import("ipc/ring.zig").Ring;
 
+pub const ThreadCrashContext = struct {
+    finished_idx: *std.atomic.Value(u16),
+    reset_event: *std.Thread.ResetEvent,
+};
```
Sobeston: This should be an extern struct, as should be any structs that we point to (which ResetEvent isn't)

This is because the services and the init process are different compile units, which are allowed to have different layouts for normal structs

Unfortunately using ResetEvent isn't sound for this reason

Sobeston (changes requested):

v2/lib/ipc.zig
jbuckmccready: Oh, hmm yeah. I didn't really consider how deep the multiple compilations affects pretty much everything in the code base.

The opaque pointer does work by keeping the cast and dereference inside of `services.zig` which is only in the main compilation unit, so maybe that's fine for now.

My understanding now is with this approach anything referenced by a service will be compiled into different compilation unit, so the mental model is very different from typical project. The implication is any types shared (read or write) across services must have explicit layout for ABI compatibility.

jbuckmccready: Changed back to opaque pointer arg and added some doc comments: https://github.com/Syndica/sig/pull/1410/changes/503058c4a6d5cec8fa651861098f15decd5d1779

I experimented with having context just be globals in `services.zig` which simplified the function pointer, and removed need to pass the opaque pointer around, but it made the code more confusing as it added some obfuscation for how the data moved through the control flow.

Sobeston (approved):

prestonsn (approved):

PR #1411, opened 2026-04-24
refactor(rpc): split Consensus hook context into Replay and Gossip

dnut:
### Problem

The methods in Consensus hook context do not return consensus data.

### Purpose

This is a pure refactor serving two goals:
- naming accuracy
- dependency clarity/minimization

### Implementation

First, getClusterNodes is pure gossip data, so I moved it into a Gossip hook context, which cleanly isolates the dependencies.

Second, the remaining methods are returning slot- and epoch-specific state that's managed by replay. This data is actually more runtime-oriented. There's very little consensus-specific data being provided here.

### DIFF
--- a/src/cmd.zig
+++ b/src/cmd.zig
@@ -1822,12 +1822,15 @@ fn validator(
             .epoch_schedule = &epoch_tracker.epoch_schedule,
         });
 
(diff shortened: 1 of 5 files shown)


prestonsn (approved):
LGTM 🫡

ajw221 (approved):

PR #1406, opened 2026-04-22
fix(runtime): fuzzing failures for sol_prog_bpf_loader_v3_sig_structured_diff

ajw221:
### Intent

- Fix `sol_prog_bpf_loader_v3_sig_structured_diff` octane mismatch

### Implementation

- Added `text_section_len` tracking to `Executable`
- Added missing ELF text section checking in `Executable.verify`

### Ramifications

N/A

### Tests

- Ran conformance and test fixtures against `sol_prog_bpf_loader_v3_sig_structured_diff` successfully

PR #1416, opened 2026-04-27
features: bls pubkey rekey

Rexicon226:
This will fix the unknown feature errors in:
https://app.circleci.com/pipelines/github/Syndica/sig/9474/workflows/afd5d201-f182-4fbb-ba19-ccfc6c8c7ddf/jobs/152633

Reference: https://github.com/anza-xyz/agave/pull/12190
Also keep in mind that there was a little issue noticed with it, https://github.com/anza-xyz/agave/pull/12200, which was merged just one commit after the rekey, but it is considered to be within the rekey.


Rexicon226:
Now that I think about it, this won't actually pass, as it'll need the test-vectors updates, which themselves require #1406, but feel free to merge this in.

yewman (approved):

prestonsn (approved):

PR #1418, opened 2026-04-28
feat(v2): replay

### LINKED ISSUE #1189: feat(replay): v2 service

Sobeston:
includes
- some shred receiver fixes
- a new exec process, with an async task request/response pattern
- allocation of blocks
- correct chaining of fec sets
- incremental deserialisation over blocks
- some replay progress tracking

image

```
(replay) info: Slot 458717152 (@enumFromInt(33)) complete! (24/24)
(replay) info: Slot 458717154 (@enumFromInt(35)) complete! (20/20)
(replay) info: Slot 458717155 (@enumFromInt(36)) complete! (38/38)
(replay) info: Slot 458717156 (@enumFromInt(37)) complete! (125/125)
(replay) info: Slot 458717157 (@enumFromInt(38)) complete! (57/57)
(replay) info: Slot 458717158 (@enumFromInt(39)) complete! (57/57)
(replay) info: Slot 458717159 (@enumFromInt(40)) complete! (56/56)
(replay) info: Slot 458717160 (@enumFromInt(41)) complete! (32/32)
(replay) info: Slot 458717161 (@enumFromInt(42)) complete! (29/29)
(replay) info: Slot 458717162 (@enumFromInt(43)) complete! (60/60)
(replay) info: Slot 458717163 (@enumFromInt(44)) complete! (64/64)
(replay) info: Slot 458717164 (@enumFromInt(45)) complete! (32/32)
(replay) info: Slot 458717166 (@enumFromInt(47)) complete! (21/21)
```


v2/lib/collections/pool.zig
```diff
@@ -1,5 +1,167 @@
...
+                const head = self.free_list.load(.acquire);
+                if (head == .null) return error.OutOfSpace;
+
+                const new_node: *Node = &self.buf()[head.index().?];
+                const next = new_node.next_free;
+
+                // these atomics may be wrong, but king will smell this and fix it before it gets
+                // in main
+                if (self.free_list.cmpxchgWeak(head, next, .acq_rel, .acquire) != null)
+                    continue;
```
kprotty: l'indeed:
```rs
free_list : [ a -> b -> c -> null ]
t1.create:
  head = load() // a
  next = head.next // b

t2.create:
  head = load() // a
  cas(a, b) orelse return // free_list : [ b -> c -> null ]
t2.create (again):
  ... cas(b, c) orelse return // free_list : [ c -> null ]
t2.destroy(a):
 ... cas(c, a) orelse return // free_list : [a -> c -> null ]

t1.create (resumed):
  cas(a, b) orelse return // free_list [ a -> b?? -> ?] [ c -> null ]
```

Either need to make the pool single-creator, or add a lock internally. Can it be single-create here? (make the pool not generic)

v2/lib/solana/bincode_2.zig
```diff
@@ -0,0 +1,98 @@
+const std = @import("std");
```
kprotty: bincode_2 (remove lib/solana/bincode & inline this to place if u must)

v2/lib/solana/transaction.zig
```diff
@@ -0,0 +1,127 @@
...
+const tracy = @import("tracy");
+const lib = @import("../lib.zig");
+
+const bincode = lib.solana.bincode_2;
+
+const Signature = lib.solana.Signature;
+const Pubkey = lib.solana.Pubkey;
+const Hash = lib.solana.Hash;
+
+pub const Entry = struct {
```
kprotty, line 11: ideally, make a deserializer for this directly on the struct to avoid bincode. Even if it still takes in an fba+reader (see lib/solana/snapshot.zig in #1393 )

v2/lib/replay.zig
```diff
@@ -0,0 +1,77 @@
+const std = @import("std");
+const solana = @import("solana.zig");
+const collections = @import("collections.zig");
+const ipc = @import("ipc.zig");
+
+comptime {
+    _ = std.testing.refAllDecls(@This());
+}
+
+pub const TransactionPool = collections.SharedPool([1232]u8, 1_000_000);
```
kprotty: thick one

v2/lib/replay.zig
```diff
@@ -0,0 +1,77 @@
...
+
+    request_kind: RequestKind,
+    data: extern union {
+        transaction_execution: extern struct {
+            block_idx: BlockRef,
+            tx_idx: TransactionPool.ItemId,
+        },
+        transaction_signature_verify: extern struct {
+            tx_idx: TransactionPool.ItemId,
+        },
```
kprotty: thoughts on shortening these variant names? `txn_exec`, `txn_sig_verify`

v2/lib/replay.zig
```diff
@@ -0,0 +1,77 @@
...
+
+pub const ExecResponse = extern struct {
+    task_id: u64, // user-provided, arbitrary, for the caller's tracking
+
+    request_kind: RequestKind,
+    data: extern union {
+        transaction_execution: extern struct {
+            block_idx: BlockRef,
+            tx_idx: TransactionPool.ItemId,
+            success: bool,
```
kprotty: should replace `success: bool` with something like `res: TransactionResult` to give replay reason for the failure, similar to how it was done in v1 with `ReplaySlotFuture.setError`. Ideally, the error code type should be just as small (u8/u16 backing)

v2/lib/TopologySchema.zig
```diff
@@ -397,6 +397,8 @@ pub fn Bind(
             regions: []const SharedRegion,
             region_memfds: []const lib.linux.memfd.RW,
         ) std.mem.Allocator.Error!ServiceMap {
+            @setEvalBranchQuota(5000);
```
kprotty: i did this too, but 100_000. Just as a heads up.

v2/build.zig
```diff
@@ -61,9 +61,17 @@ pub fn build(b: *Build) !void {
             "not silently accepted.",
     ) orelse (optimize == .Debug);
 
+    const debug_skip_shred_checks = b.option(
+        bool,
+        "debug-skip-shred-checks",
```
kprotty, line 73: maybe rename to `debug-skip-verify` for general verification skipping

kprotty (commented):
partial review for now

v2/init/services.zon
```diff
@@ -45,4 +48,12 @@
             .{ .name = .main, .access = .rw },
         },
     },
+    .{
+        .name = .exec,
+        .regions = .{
+            .{ .name = .exec_req_response, .access = .rw },
+            .{ .name = .transaction_pool, .access = .rw },
+            .{ .name = .block_pool, .access = .rw },
```
InKryption: The access modifiers here don't line up with the ReadOnly & ReadWrite fields in `services/exec.zig`.

v2/lib/shred/receiver.zig
```diff
@@ -366,6 +416,63 @@ pub const FecSetCtx = extern struct {
 
         return data_recv + code_recv;
     }
+
+    // prep for recover64
+    // TODO: clean this code up, it's bad, e.g. we should probably ban constcast
```
InKryption: I don't see a constCast in here?

v2/lib/replay.zig
```diff
@@ -0,0 +1,77 @@
+const std = @import("std");
+const solana = @import("solana.zig");
+const collections = @import("collections.zig");
+const ipc = @import("ipc.zig");
+
+comptime {
+    _ = std.testing.refAllDecls(@This());
+}
```
InKryption: No real point in this

v2/lib/telemetry.zig
```diff
@@ -317,6 +318,8 @@ pub fn Logger(comptime scope_str: []const u8) type {
                     comptime fmt_str: []const u8,
                     args: anytype,
                 ) void {
+                    tracy.print("{s}: " ++ fmt_str, .{@tagName(self.level)} ++ args);
```
InKryption: Maybe not super important, but this would break any fmt messages that use positional indices.

v2/services/exec.zig
```diff
@@ -0,0 +1,87 @@
...
+    while (true) {
+        const request: *const lib.replay.ExecRequest = request_reader.next() orelse continue;
+        defer request_reader.markUsed();
+        defer deserial_fba.reset();
+
+        const zone = tracy.Zone.init(@src(), .{});
+        defer zone.deinit();
+        zone.name(switch (request.request_kind) {
+            inline else => |val| @as([:0]const u8, @tagName(val)),
+        });
```
InKryption: What's this about?

v2/services/replay.zig
```diff
@@ -23,273 +106,901 @@ pub const std_options = start.options;
...
-        };
+            try w.writeAll(read_slice);
+            self.deserial_state.pos_offset += read_slice.len;
+            return read_slice.len;
+        }
+
+        fn rebase(r: *std.Io.Reader, capacity: usize) !void {
+            _ = .{ r, capacity };
+            @panic("rebase is illegal");
+        }
```
InKryption: Why is rebase illegal? Feel like I'm missing something.

InKryption (commented):
Just a few high level comments

v2/services/exec.zig
Sobeston: Just setting the zone name based on the kind of the request (without duplicate code, or adding extra nesting)

InKryption: why wouldn't just `@tagName(request.request_kind)` directly work?

v2/lib/collections/pool.zig
Sobeston: Hm so far the intended users of this only have one thread that does all create/destroy anyway

How about I just remove the atomics, and document the limitation?

v2/build.zig
Sobeston: I think this would be best to do later maybe, because at the moment this only affects some shred checks + I'm a bit wary of removing checks generally. Not sure which other checks we'd want to rid get of

v2/lib/solana/transaction.zig
Sobeston: I've gone for custom/direct on the replay side, but left this generic version for tx exec, might do it on both sides

v2/lib/collections/pool.zig
Sobeston: removed atomics for now https://github.com/Syndica/sig/pull/1418/commits/1750aa3e4e73a26c329bf7790a7377d562d68bd4

v2/services/exec.zig
Sobeston: Not sure why I wrote that fixed in, 95969109198c7c10bef5d7eda3df58082bec48e2

v2/lib/telemetry.zig
Sobeston: Is there a nicer way of doing this? I guess it doesn't matter for now

v2/lib/replay.zig
Sobeston: removed this

v2/lib/shred/receiver.zig
Sobeston: outdated comment, removed it in 95969109198c7c10bef5d7eda3df58082bec48e2

v2/init/services.zon
Sobeston: fixed in 95969109198c7c10bef5d7eda3df58082bec48e2

v2/lib/replay.zig
Sobeston: shortened 95969109198c7c10bef5d7eda3df58082bec48e2

v2/lib/replay.zig
Sobeston: 95969109198c7c10bef5d7eda3df58082bec48e2 reduced the size for now

v2/lib/solana/bincode_2.zig
Sobeston: removed bincode_2

v2/services/replay.zig
Sobeston: This was an artifact of my weird io.Reader, stopped using it in https://github.com/Syndica/sig/pull/1418/commits/9d18ea2ffb36b6629af6e11294679b6f7ffa5e39

v2/lib/replay.zig
Sobeston: Added basic one here 9d18ea2ffb36b6629af6e11294679b6f7ffa5e39
let's add fields as we go?

v2/lib/telemetry.zig
InKryption: Possible fix:
```suggestion
                    switch (self.level) {
                        inline else => |ilevel| {
                            tracy.print(@tagName(ilevel) ++ ": " ++ fmt_str, args);
                        },
                    }
```

InKryption (dismissed):

v2/build.zig
```diff
@@ -168,7 +175,6 @@ pub fn build(b: *Build) !void {
         .optimize = optimize,
         .imports = &.{
             .{ .name = "base58", .module = base58_mod },
```
prestonsn, line 185: I see that we've removed `binkode`, but it looks like we still import/use it here: 

https://github.com/Syndica/sig/blob/ae63c80ee7a9476e5c07f3cd614964ec4c97cfce/v2/lib/solana/signature.zig#L9

I'm also curious why this did not cause the build to fail, but maybe has something to do with zig's build cache? 

not sure if we want to remove it fully, or if keeping this import was intended for something

v2/services/replay.zig
```diff
@@ -1,20 +1,96 @@
+//! Completed FEC (Forward Error Corerectrion) sets flow into this service from the shred receiver.
```
prestonsn:
```suggestion
//! Completed FEC (Forward Error Correction) sets flow into this service from the shred receiver.
```

prestonsn: Also I really appreciate the detailed doc comment here, great context!

v2/services/replay.zig
```diff
@@ -1,20 +1,96 @@
...
+//!
+//! Each FEC set contains its own Merkle Root, and its Chained Merkle Root, with the Chained Merkle
+//! Root refering to the FEC set of its parent, i.e. the previous FEC set.
```
prestonsn:
```suggestion
//! Root referring to the FEC set of its parent, i.e. the previous FEC set.
```

v2/services/replay.zig
```diff
@@ -28,280 +104,951 @@ pub const std_options = start.options;
...
+            while (offset < n_bytes) {
+                if (mode == .copy) {
+                    @memcpy(out[offset..next_copy.len], next_copy);
+                }
+                self.deserial_state.pos_offset += next_copy.len;
+
+                offset += next_copy.len;
+                try self.nextNode();
+                next_copy = self.currentReadableSlice();
+            }
```
prestonsn, line 797: I think there's a subtle issue here for the chained fec set case (where a block is spread over 2+ fec sets)

Right now `@memcpy(out[offset..next_copy.len], next_copy);` uses `next_copy.len` as the end of the outpt slice. 

But once `offset > 0` (i.e after the first frag.), I think the destination range becomes wrong?

prestonsn: thoughts on something like this (with early break to skip going to another node after the requested read is done)?


```suggestion
            while (offset < n_bytes) {
                const copy_len = @min(next_copy.len, n_bytes - offset);

                if (mode == .copy) {
                    @memcpy(out[offset..][0..copy_len], next_copy[0..copy_len]);
                }

                self.deserial_state.pos_offset += copy_len;
                offset += copy_len;

                if (offset == n_bytes) break;

                try self.nextNode();
                next_copy = self.currentReadableSlice();
            }
```

prestonsn (commented):
Looking good to me so far. Still getting through the base replay code, but wanted to bring up a couple things i noticed

v2/services/replay.zig
Sobeston: Nice catch, can confirm this fixes the bug: https://github.com/Syndica/sig/pull/1418/commits/0b32ec5fb86abe38150eb60284b57e1075b37778

v2/services/replay.zig
Sobeston: fixed some typos https://github.com/Syndica/sig/pull/1418/commits/266fc5de5d9f8a90301cbdca3dc4088bed394d6e

v2/services/replay.zig
Sobeston: fixed some typos https://github.com/Syndica/sig/pull/1418/commits/266fc5de5d9f8a90301cbdca3dc4088bed394d6e

v2/build.zig
Sobeston: This code was never really referenced, so it didn't cause a compile error

removed https://github.com/Syndica/sig/pull/1418/commits/be582bbba9e64ad0d7196368fbbfdabd35cebae9

v2/lib/telemetry.zig
Sobeston: went for it: https://github.com/Syndica/sig/pull/1418/commits/ebc35949e26d1f028d5f37ba0b88ab55a51756d6

InKryption (dismissed):

v2/services/replay.zig
```diff
@@ -28,280 +104,954 @@ pub const std_options = start.options;
...
-    ) !BlockTree {
-        const pool_buf = try allocator.alloc(Node, capacity);
-        errdefer allocator.free(pool_buf);
+    // Iterate over all child nodes, setting their parent node, and deleting any bad child nodes
+    {
+        var maybe_child_node: ?*MerkleNode = children_head;
+        while (maybe_child_node) |child_node| {
+            std.debug.assert(child_node.parent == .null);
+
+            defer maybe_child_node = child_node.sibling.ptr(&forest.pool);
```
prestonsn: it seems like this `defer` line would read `child_nde.sibling` after `child_node` may have already been destroyed on the invalid child path due to this at the bottom of the loop: 

https://github.com/Syndica/sig/blob/4e2829b7973389f022d211d95d7f267f0f527267/v2/services/replay.zig#L430

v2/services/replay.zig
```diff
@@ -28,280 +104,954 @@ pub const std_options = start.options;
...
+                self.n_transactions_left.? -= 1;
+
+                const post_state = self.*;
+
+                const tx_bytes_read = diff(pre_state, post_state, merkle_pool);
+
+                var tx_reader = pre_state.getReader(merkle_pool);
+                try tx_reader.advanceBytes(.copy, tx_buf[0..tx_bytes_read], {});
+                self.next_read = .transaction;
+                return tx_buf[0..tx_bytes_read];
```
prestonsn, line 1030: Thoughts on some runtime enforcement on `tx_bytes_read`? I guess i'm not seeing anything defending against this directly, but maybe this is handled elsewhere, but something like: 

```zig
if (tx_bytes_read > tx_buf.len) return error.TransactionTooLarge;
```

v2/services/exec.zig
```diff
@@ -0,0 +1,86 @@
...
+
+                var reader =
+                    std.io.Reader.fixed(ro.replay_transaction_pool.indexToConstPtr(data.tx_idx));
+
+                const transaction: lib.solana.transaction.VersionedTransaction =
+                    try lib.solana.bincode.read(
+                        &deserial_fba,
+                        &reader,
+                        lib.solana.transaction.VersionedTransaction,
+                    );
```
prestonsn, line 64: Does replay service need to be informed of this (unlikely) decode failure via an `ExecResponse` in the future? (just curious if that's the plan, not code change stuff for this pr)

v2/build.zig
prestonsn: ahh i c

prestonsn (dismissed):

v2/services/replay.zig
prestonsn: (can also be a TODO comment, or feel free to resolve if it's kinda enforced elsewhere and I missed it)

v2/services/exec.zig
Sobeston: Right now we're just hard-crashing the validator if this fails, i.e. assuming that transactions can be deserialised.

We *should* report this failure back to replay (which should invalidate the whole block), however we don't yet have the reporting or block invalidation code

v2/services/replay.zig
Sobeston: https://github.com/Syndica/sig/pull/1418/commits/4ca921e72dbfc8f2bbf6362db086fe78459b0604

added this, might as well
we will need to handle invalid blocks more thoroughly later

v2/services/replay.zig
Sobeston: Nice find, I just removed the defer https://github.com/Syndica/sig/pull/1418/commits/4ca921e72dbfc8f2bbf6362db086fe78459b0604

v2/services/replay.zig
prestonsn: Gotcha :saluting_face:

v2/services/exec.zig
prestonsn: Clean, thanks for the context :saluting_face:

prestonsn (approved):

dnut (approved):

PR #1423, opened 2026-04-29
feat(runtime,conformance): simd-0268: raise cpi nesting limit to 8

ajw221:
https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0268-raise-cpi-nesting-limit.md

### DIFF
--- a/src/runtime/transaction_context.zig
+++ b/src/runtime/transaction_context.zig
@@ -81,7 +81,7 @@ pub const TransactionContext = struct {
     /// Only set if a precompile is present and the move precompiles to svm feature is enabled
     instruction_datas: ?[]const []const u8 = null,
 
-    instruction_stack: InstructionStack,
+    instruction_stack: std.ArrayList(InstructionContext),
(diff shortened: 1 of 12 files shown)


src/runtime/transaction_context.zig
```diff
@@ -103,7 +107,7 @@ pub const TransactionContext = struct {
 
     pub const InstructionStack = std14.BoundedArray(
         InstructionContext,
-        MAX_INSTRUCTION_STACK_DEPTH,
+        MAX_INSTRUCTION_STACK_DEPTH_SIMD_0268,
```
yewman: We should change this to an array and use the runtime determined max instruction stack depth. A good opportunity to remove the BoundedArray usage which is deprecated since 0.15

ajw221: [87948b4](https://github.com/Syndica/sig/commit/87948b492ada0e367893f77bb7e969ad144316dd)

PR #1426, opened 2026-04-29
fix(conformance): runner passed, failed, skipped output granularity

ajw221:
### Intent

- Re-add pass/failed/skipped totals for each fixtures subfolder to the new zig test runner
- Update invalid example commands in conformance README

### Implementation

- Updated conformance test runner to show totals for each subfolder
- Fixed invalid example commands in conformance README

### Ramifications

- Any output with the test runner will not immediately show where any failures are, just the total passed/failed/skipped, whereas previously these numbers were being shown for each subfolder

### Tests

N/A


hamza-syndica (dismissed):
NIT:

I always forget to do build the lib file, can we do that part automatically?

Like: before running the fixture it should automatically run `zig build solfuzz_sig` too?

It might be upto the choice/decision. Just leaving it here for thought.

yewman (dismissed):

conformance/src/main.zig
```diff
@@ -28,37 +28,122 @@ pub fn main() !void {
...
         const out_buf = try allocator.alloc(u8, exec.output_buffer_size);
         defer allocator.free(out_buf);
         switch (try exec.execFixture(allocator, &lib, args[1], out_buf)) {
-            .pass => passed_count = 1,
-            else => |r| {
-                std.debug.print("{s}: {}\n", .{ args[1], r });
-                failed_count = 1;
-            },
+            .pass => total_passed = 1,
+            .missing_entrypoint => total_skipped = 1,
```
dnut: this should be considered a failure. otherwise we risk missing bugs in ci.

we should only classify a test as skipped if it was explicitly requested by the user or the programmer to skip the test. this is not supported currently so i would just hardcode skips to 0.

conformance/src/main.zig
```diff
@@ -28,37 +28,122 @@ pub fn main() !void {
...
+        for (group_dirs.items) |gd| {
+            const rel = relativeTo(gd, input_prefix);
+            max_name_len = @max(max_name_len, rel.len);
+        }
+
+        // Run each group and print immediately
+        std.debug.print("\n", .{});
+        for (group_dirs.items) |group_path| {
+            const display_name = relativeTo(group_path, input_prefix);
+            const stats = exec.execDir(allocator, &lib, group_path) catch |e| {
```
dnut: breaking up execDir into a separate execution per folder degrades the performance of the test runner. if you want to format the output with a separate line per folder, that's fine, but you should run all of them at once and then process the results after it all completes

likewise you can simplify the code too. you don't need any code to walk the directories. execDir already handles this. you can just take the stats returned by execDir and sum them, grouping by dirname

dnut (changes requested):
I like the idea but there is another approach that is simpler and faster

conformance/src/main.zig
ajw221: [69ff33a](https://github.com/Syndica/sig/pull/1426/commits/69ff33afce07cf70acaee9c1cbf5f80392b894d0)

conformance/src/main.zig
ajw221: [69ff33a](https://github.com/Syndica/sig/pull/1426/commits/69ff33afce07cf70acaee9c1cbf5f80392b894d0)

conformance/src/main.zig
```diff
@@ -28,37 +28,100 @@ pub fn main() !void {
...
+    var total_failed: usize = 0;
     const stat = try std.fs.cwd().statFile(args[1]);
+
     if (stat.kind == .directory) {
         const stats = try exec.execDir(allocator, &lib, args[1]);
-        for (stats.fix_paths, stats.results) |input_path, result| switch (result) {
-            .pass => passed_count += 1,
-            else => {
-                failed_count += 1;
-                std.debug.print("{s}: {}\n", .{ input_path, result });
```
dnut, line 40: by deleting this, you've removed the only way that we could discover which tests actually failed. the summary is nice, but i think printing the individual test failures is more important

conformance/src/main.zig
```diff
@@ -28,37 +28,100 @@ pub fn main() !void {
...
+            total_passed += passed;
+            total_failed += failed;
+        }
     } else {
         const out_buf = try allocator.alloc(u8, exec.output_buffer_size);
         defer allocator.free(out_buf);
         switch (try exec.execFixture(allocator, &lib, args[1], out_buf)) {
-            .pass => passed_count = 1,
-            else => |r| {
-                std.debug.print("{s}: {}\n", .{ args[1], r });
```
dnut, line 49: while the print is not as important here as it is when running multiple fixtures, it still provided some information that is now hidden, i.e. what kind of failure it was

conformance/src/main.zig
ajw221: [e1e525b](https://github.com/Syndica/sig/pull/1426/commits/e1e525b0d5053e24b137cedff7944a8a5352fa5b)

conformance/src/main.zig
ajw221: [e1e525b](https://github.com/Syndica/sig/pull/1426/commits/e1e525b0d5053e24b137cedff7944a8a5352fa5b)

dnut (approved):

hamza-syndica (approved):

PR #1427, opened 2026-04-30
refactor(features): refactor features and resolve unkown ids

yewman:
- Update feature pubkeys which had been rekeyed and causing unknown id errors in conformance testing
- Add metadata to entries in the features.zon to convey information about there activation and implementation status
- Warn about unkown feature ids in conformance only if they are not in our current feature set and not known reverted keys.


hamza-syndica (dismissed):

ajw221 (dismissed):

ajw221 (dismissed):

hamza-syndica (dismissed):

src/core/features.zig
```diff
@@ -7,63 +7,108 @@ const EpochSchedule = sig.core.EpochSchedule;
...
+    name: [:0]const u8,
+    pubkey: [:0]const u8,
+    description: [:0]const u8,
+    status: union(enum) {
+        /// The feature has been removed from the feature set and its implementation has been removed from source.
+        /// Fuzzing: reverted features may appear in historical regression fixtures.
+        reverted,
+        /// This feature is contained in the feature set and its implementation is not complete.
+        /// Fuzzing: not advertised as a supported feature for fuzzing (i.e. it will not be toggled on).
+        staged,
```
dnut: `staged` is ambiguous and potentially misleading. Usually I think of staging as a testing phase that comes after implementation is complete but before deployment to production. Conceptually that word maps more closely to the `supported` state. I prefer `unsupported` for this state.

src/core/features.zig
```diff
@@ -7,63 +7,108 @@ const EpochSchedule = sig.core.EpochSchedule;
...
+        /// This feature is contained in the feature set and its implementation is complete.
+        /// Fuzzing: advertised as a supported feature for fuzzing (i.e. it may be toggled on and off).
+        supported,
+        /// This feature has been implemented and is included in the feature set.
+        /// Fuzzing: advertised as a hardcoded feature for fuzzing (i.e. it is always toggled on).
+        hardcoded,
+        /// This feature has been implemented and is included in the feature set.
+        /// Fuzzing: advertised as a hardcoded feature for fuzzing (i.e. it is always toggled on).
+        persisted,
```
dnut: These two have exactly the same comment and the names also don't help clarify the distinction. I remember our conversation, so I know what these mean, but no one else will be able to figure it out from this.

I suggest using different names and adding more explanation into the comments.

Rename `hardcoded` to `activated`: Think about when this stage is actually useful. We shouldn't stop fuzzing the deactivated path until after it's activated. After activation, there's no point in fuzzing the deactivated state any more. The whole point of this stage is to modify the fuzzing behavior after the feature is activated. 

Rename `persisted` to `hardcoded`. Hardcoded should be the final stage. This is actually exactly what the word "hardcoded" refers to. The feature is no longer dynamically parameterized. It's hardcoded as active.

`activated` ends up being treated as if its hardcoded for fuzzing purposes. But it's not actually hardcoded. We're lying to the fuzzing engine to get the fuzzing coverage we want.

```suggestion
        /// This feature has been implemented and is included in the feature set.
        /// Sig can be parameterized by this feature's activation, and can function properly
        /// with the feature either enabled or disabled, but we are no longer interested in fuzzing
        /// with the feature disabled because it is already activated on mainnet.
        /// Fuzzing: advertised as a hardcoded feature for fuzzing (i.e. it is always toggled on).
        activated,
        /// This feature has been implemented and is included in the feature set.
        /// Sig no longer supports parameterization of this feature. Sig is hardcoded to assume
        /// that the code is always activated, and cannot properly handle it being deactivated.
        /// Fuzzing: advertised as a hardcoded feature for fuzzing (i.e. it is always toggled on).
        hardcoded,
```

src/core/features.zig
```diff
@@ -7,63 +7,108 @@ const EpochSchedule = sig.core.EpochSchedule;
...
+/// ZonInfo represents the metadata for a feature, including its name, pubkey, description, status, and an optional note.
+/// It is used to record the history and current state of all features, including those that have been reverted or staged for future activation.
+const ZonInfo = struct {
+    name: [:0]const u8,
+    pubkey: [:0]const u8,
+    description: [:0]const u8,
+    status: union(enum) {
+        /// The feature has been removed from the feature set and its implementation has been removed from source.
+        /// Fuzzing: reverted features may appear in historical regression fixtures.
+        reverted,
```
dnut, line 27: Originally you had this at the end which I thought makes sense because it's a terminal state. Not important though.

dnut (changes requested):

src/core/features.zig
yewman: What about reverted, unsupported, supported, hardcoded_for_fuzzing, and hardcoded. I would prefer not to use `activated` as it the activation <-> hardcoded_for_fuzzing relationship may not always hold. Also I agree with your earlier statement that this should reflect 'sig' state, not cluster feature statuses.

src/core/features.zig
yewman: 44209f3f

src/core/features.zig
yewman: 44209f3f

dnut (dismissed):

hamza-syndica (dismissed):

ajw221 (approved):

prestonsn (approved):

PR #1428, opened 2026-04-30
feat(dev-setup): updated conformance to support more platforms

hamza-syndica:
## Summary

Updated the conformance's nix setup to run on macOS or arm64 linux machines.


yewman (approved):

dnut (approved):
The layering of `let` expressions is a bit confusing, but I don't see an easy way to reduce the layering. it's worth it to get more platforms supported

PR #1429, opened 2026-04-30
chore(runtime): revert stake_minimum_delegation_for_rewards

yewman:
Not ready for activation due to flux in agave.


hamza-syndica (commented):
Agave still has the stake_minimum_delegation_for_rewards gate in `partitioned_epoch_rewards/calculation.rs:385`

I know that you have already written in the PR description. But, shouldn't we mark this PR as draft?

PR #1432, opened 2026-04-30
feat(v2:telemetry): log filters

### LINKED ISSUE #1389: feat(v2): per-service log filtering
Allow telemetry to configure the log level per service instead of globally.

The idea is to support showing debug/trace logs of a specific service while debugging, without having to filter through the logs of other services.

InKryption:
* Adds `fatal` log level, which should be reserved for absolutely critical failures that cannot be recovered from.
* Adds `log_level` to the config to define what the default log level should be in the absence of any filter.
* Replaces the singular `max_log_level` cli arg with a filter syntax, which can also still be written as just a log level.
* Allows filtering by both or either service & scope.

### DIFF
--- a/v2/lib/telemetry.zig
+++ b/v2/lib/telemetry.zig
@@ -39,8 +39,8 @@ pub const Region = extern struct {
     pub const Info = extern struct {
         /// The port to listen on for the prometheus client.
         port: u16,
-        /// The maximum log level to emit.
-        max_log_level: log.Level,
+        /// The length of the encoded log filters byte string.
+        log_filters_len: u32,
 
         /// Number of other services excluding the telemetry service (of which there is presumably only instance).
         /// This is also the maximum number of log streams to support.
@@ -58,6 +58,9 @@ pub const Region = extern struct {
             var size: usize = 0;
             size += @sizeOf(Region);
 
+            size = std.mem.alignForward(usize, size, @alignOf(u8));
+            size += self.log_filters_len;
+
             size = std.mem.alignForward(usize, size, @alignOf(u64));
             size += self.gauges_len * @sizeOf(u64);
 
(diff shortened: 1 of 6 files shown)


v2/lib/telemetry.zig
```diff
@@ -277,7 +309,6 @@ pub fn Logger(comptime scope_str: []const u8) type {
                     comptime fmt_str: []const u8,
                     args: anytype,
                 ) void {
-                    if (@intFromEnum(self.level) > @intFromEnum(self.logger.max_level)) return;
```
prestonsn, line 280: All changes look good to me 🫡 

I think the only thing that stands out a bit is with the removal of this check here. It seems like now, despite the setup of the filter and `max_level`, all logs will be encoded and placed into the shmem region, regardless of desired level. Maybe this adds more overhead for debug logs/logs we don't care about in 90% of cases. 

I mostly wanted to bring this up due to the presence of pre-existing debug log usage like this: https://github.com/Syndica/sig/blob/469215c91857c500e90576f5077babf858e99e6c/v2/services/shred_receiver.zig#L91-L101

Which logs every shred at debug level, where this early level-check-and-return would prevent from performing any additional work. 

Likely not a concern, but perhaps there's a clean way to keep such an early return?

prestonsn (approved):

v2/lib/telemetry.zig
prestonsn: Correction on my part: shouldn't increase Ring traffic at all. Likely not a concern and is a totally worthwhile tradeoff.

InKryption: I had looked at this from a few angles, but I couldn't really see a clean way to achieve it; it would require each service to do the following: have access to the log filters list (which would require decoding the log filters string at startup, which has its own complexities), and then iterate through the filters every time it wants to emit a log to see if there's a match; I think offloading these details and computational costs to the telemetry service makes the most sense.

prestonsn: Thanks for the context here, and agreed on all points 🫡

v2/lib/telemetry/log.zig
```diff
@@ -19,6 +21,23 @@ pub const Level = enum(u8) {
...
+            fatal = @intFromEnum(Level.fatal),
+            @"error" = @intFromEnum(Level.err),
+            warn = @intFromEnum(Level.warn),
+            info = @intFromEnum(Level.info),
+            debug = @intFromEnum(Level.debug),
+            trace = @intFromEnum(Level.trace),
+        };
+        const str_text = std.meta.stringToEnum(Text, str) orelse return null;
+        return @enumFromInt(@intFromEnum(str_text));
+    }
```
jbuckmccready, line 36: Not familiar with background to this, why is there this split across enums with `err` and `error`? Is this because we want the level to be "error" but don't want to deal with `@"error"` in the code? Kind of confusing having the enum label `err` not match the logging output/filtering.

InKryption: It's mostly that, yes, but also because it mirrors the actual names of the methods used in code. `logger.err().logf(...)` -> `logger.entry(.err).logf(...)`.

v2/lib/telemetry.zig
jbuckmccready: This is maybe not an immediately pressing issue but this approach to filtering feels counter intuitive to log filtering. Filtering is usually helpful in allowing for more expensive formatting or verbose logging within a scope that can be compile time filtered out, or cheaply filtered at runtime, without formatting or copying, which makes it viable to retain verbose log lines for quick debugging.

E.g., maybe have 5 services and having all of them output verbose logs would overwhelm the system, but narrowing it to be verbose for one module inside one service is completely reasonable and useful for debugging.

I think we could load the log filtering state into shared memory that the services can read for filtering on call side. Also seems like we could apply comptime call side filtering via compilation inputs.

Some more ideas:
Rust tracing library works by performing call site registration, when a logging site is called it loads an associated byte:

```rust
   match self.interest.load(Ordering::Relaxed) {
       0 => Interest::never(),
       1 => Interest::sometimes(),
       2 => Interest::always(),
       _ => self.register(),
   }
```

The fist time it's called `register()` initializes the interest byte according to active subscribers/filters, and after that it's just a relaxed load and branch for filtering unless it's `interest::sometimes()` in which case it does active filtering (e.g., maybe filtering by logging value or span according to dynamic subscribers, but this is just a side note, I don't think we need this kind of log arg/input based filtering).

v2/lib/telemetry/log.zig
dnut: At least one type of ugly is required:

1. internal inconsistency: logging an error message invokes `err()` whereas the logger internally reads/writes text using `error`
2. awkward syntax: logging an error message invokes `.@"error"()`
3. external inconsistency: error messages are labelled as "err" instead of the more conventional "error"

I would rank my choices in that order, but it's pretty arbitrary.

jbuckmccready (approved):

PR #1434, opened 2026-05-03
ci: migrate circleci to github

dnut:


.github/workflows/check_linux.yml
```diff
@@ -0,0 +1,236 @@
...
+      - uses: mlugg/setup-zig@v2
+        with:
+          version: 0.15.2
+          cache-key: selfhosted-${{ hashFiles('build.zig', 'build.zig.zon') }}
+
+      - name: Build
+        run: |
+          sudo apt-get update -y && sudo apt-get install wget -y
+          ./scripts/proxy_workaround.sh zig
+          zig build -Denable-tsan=true -p workspace/zig-out --summary all
```
prestonsn, line 136: I think we've dropped `-Dcpu=x86_64_v3` here (was in the `.circleci/config.yml`

Maybe this was intentional?

prestonsn: I think this applies to other `zig build *` lines

.circleci/config.yml
```diff
@@ -1,363 +0,0 @@
...
-      - checkout
-      - attach_workspace:
-          at: workspace
-      - restore_cache:
-          key: macos-aarch64-0.14.1-{{ checksum "build.zig.zon" }}
-      - run:
-          name: Build and Run Webzockets Tests
-          command: |
-            ulimit -Hn unlimited
-            ulimit -Sn unlimited
```
prestonsn, line 154: I think these got removed by accident? But maybe these are set by default in GitHub actions?

prestonsn: Looks like these got removed across all the macos workflows

prestonsn: Nvm, no longer needed due to accountsdb not opening a million files anymore

.github/workflows/check_linux.yml
dnut: This was intentional. It would only be necessary if we are building on one machine and running on a separate one. This is previously how the ci worked, but we now build and run on the same machine.

InKryption (approved):

prestonsn (approved):

ajw221 (approved):

PR #1435, opened 2026-05-04
feat(replay): simd-0444: Relax program data account check in migration

hamza-syndica:
https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0444-relaxing-program-data-account-requirements.md


Sobeston (approved):
I think @yewman wants us to fuzz a bit locally before merging this kind of thing, have you been able to do that?

Otherwise this seems good - this implementation seems to match Agave +  no major style issues.

yewman (approved):

PR #1437, opened 2026-05-07
fix(ci): disable tsan on macos CI

yewman:


prestonsn (dismissed):

ajw221 (dismissed):

.github/workflows/check_linux.yml
```diff
@@ -136,7 +136,7 @@ jobs:
           zig build -Denable-tsan=true -p workspace/zig-out --summary all
 
       - name: Build Tests
-        run: zig build test -Dno-run -Dlong-tests -Denable-tsan=true -Dno-network-tests
+        run: zig build test -Dno-run -Dlong-tests -Denable-tsan=true -Dno-network-tests -Dcpu=x86_64_v3
```
dnut: why are you adding this cpu here?

.github/workflows/check_macos.yml
```diff
@@ -34,7 +34,7 @@ jobs:
           cache-key: macos-${{ hashFiles('build.zig', 'build.zig.zon') }}
 
       - name: Build Check
-        run: zig build -Denable-tsan=true --summary all -Dno-run -Dno-bin
+        run: zig build -Denable-tsan=false --summary all -Dno-run -Dno-bin
```
dnut, line 37: What's the benefit of disabling tsan? This means we'll lose a signal about bugs

dnut (changes requested):
These changes seem harmful. What's the goal?

.github/workflows/check_macos.yml
yewman: Just trying to fix broken CI: https://github.com/Syndica/sig/actions/runs/25499962638/job/74829812964

.github/workflows/check_linux.yml
yewman: First attempt. Still looking into the root cause of seemingly intermittent CI failures: https://github.com/Syndica/sig/actions/runs/25499962638/job/74829812964

dnut: That failure happened on a macos runner though so this won't affect it.

yewman: Wrong link: https://github.com/Syndica/sig/actions/runs/25512554434/job/74874884337

dnut: this means we have a race. why would building for a different cpu fix the race?

.github/workflows/check_macos.yml
yewman: Yes we will lose signal about bugs specific to building on macos, but we will still run tsan for linux ci tests. Given v2 only supports linux I don't think there is much risk in disabling tsan for macos tests

prestonsn (dismissed):

.github/workflows/check_linux.yml
yewman: Not necessarily, it could be a false positive. Building for a different CPU can fix false positive tsan issues by using different instructions and atomic operations. Also this is a tsan issue stemming from RocksDB which is not tsan instrumented in our build (sanitize_c = off), so tsan has partial visibility which can make it more likely to produce false positives. I am still reading up on this confirm and improve my understanding, but it is my view that there is very little risk to merging this right now and it would be great to have CI working. 
[c4fbd3f](https://github.com/Syndica/sig/pull/1437/commits/c4fbd3f1682046f951b7630a1b2ffa9ba37736b1)

dnut: I ran the tests 20 times in ci with the code on main and didn't see this tsan warning a single time.

Sobeston (approved):
If you're sure that we're getting false positives on macos, let's go for it.

Even if this isn't a false positive, it's not like we have the time to fix rocksdb for macos...

.github/workflows/check_linux.yml
yewman: [fd879f1](https://github.com/Syndica/sig/pull/1437/commits/fd879f1734b9b14a0dc28cd29da3c76f8424c40c)

dnut (approved):

PR #1440, opened 2026-05-08
test(v2): kcov

dnut:
I added kcov support for v2 by modifying build.zig to run kcov whenever the `test` step runs and the `-Dkcov` flag is enabled. I modified ci to run it and include it in the codecov reports.

In ci, I stopped using docker to run kcov because it is unnecessary and difficult to work with. All we need is a kcov binary, but upstream does not produce binaries and the version in ubuntu is too old (which is why we used docker previously), so I forked kcov at https://github.com/Syndica/kcov to auto-build release binaries, which we can download and use in ci.


yewman (dismissed):

prestonsn (dismissed):

prestonsn (dismissed):

.github/workflows/check_linux.yml
```diff
@@ -31,25 +31,45 @@ jobs:
...
+      - name: Collect v1 coverage
+        run: bash scripts/kcov_ci.sh
 
-      - name: Configure docker user/mount defaults
+      - name: Collect v2 coverage
         run: |
-          echo "DOCKER_UID_GID=$(id -u):$(id -g)" >> "$GITHUB_ENV"
-          printf 'DOCKER_MOUNT=-v %s:%s -w %s -e HOME=%s -u %s\n' "$PWD" "$PWD" "$PWD" "$PWD" "$(id -u):$(id -g)" >> "$GITHUB_ENV"
+          cd v2
+          zig build test -Dkcov --summary all
```
Sobeston, line 57: How much time does this add to CI overall?

v2/build.zig
```diff
@@ -8,11 +8,12 @@ const test_install_dir: Build.Step.InstallArtifact.Options.Dir = .{
...
     const target = b.standardTargetOptions(.{});
     const optimize = b.standardOptimizeOption(.{});
-    const use_llvm = b.option(
+    const use_kcov = b.option(bool, "kcov", "Use kcov to run the tests.") orelse false;
+    const use_llvm: ?bool = b.option(
         bool,
         "use-llvm",
         "Force usage of LLVM (currently ignored for some artifacts).",
-    );
+    ) orelse if (use_kcov) true else null;
```
Sobeston: might be clearer to just error if  kcov==true and usellvm!=true

.github/workflows/check_linux.yml
dnut: Doesn't seem to make much difference at this point:

30 seconds running this step: https://github.com/Syndica/sig/actions/runs/26434204922/job/77813354815
43 seconds running `zig build ci` on main: https://github.com/Syndica/sig/actions/runs/26659498077/job/78578149213

v2/build.zig
dnut: This means the user would be required to specify `-Duse-llvm` whenever they use kcov. Why not just have kcov imply llvm will be used by default?

Now if you had said this instead:

> might be clearer to just error if kcov==true and usellvm==false

That seems a little more reasonable to me. That woud protect against someone setting `-Dkcov -Duse-llvm=false` explicitly on the command line. It would be pretty weird for someone to do that, but yeah an error probably makes the most sense in that case. That could look something like this: 7bbb96e2109e7e43b7c48f391f9eec3524882436

Why don't we just hardcode use_llvm to true everywhere? Why do we ever need to use the self-hosted backend?

.github/workflows/check_linux.yml
Sobeston: Okay, not so bad

v2/build.zig
Sobeston: > Why don't we just hardcode use_llvm to true everywhere?

I mean, true, for now basically everything has to use llvm anyway. I think it would be cleaner if we
1. have use-llvm as default true
2. error if `!llvm and kcov`

cc @InKryption

InKryption: I think use-llvm=true by default makes sense for now.

dnut: is this what you have in mind? 75327e256acdbd6cc39fb78630185e7af37f31be

prestonsn (approved):

Sobeston (approved):

PR #1441, opened 2026-05-08
feat(v2): add lint checks and test inclusion enforcement

jbuckmccready:
## Summary

Adds lint executable to `v2` written in Zig to replace `style.py`, Zig is used to take advantage of `std.Ast`.

Initial lints:
- `line_length`: works the same as `style.py` which allows longer comments but otherwise requires 100 column limit.
- `unused_declarations`: similar to unused imports from `style.py`, but now also checks all private declarations rather than just imports and uses `std.Ast` to find identifiers.
- `test_inclusion`: manages a formatted `comptime` block with `_ = import(...);` for each file in associated directory (see `test_inclusion.zig` doc comment for details).

Linting supports both default `--check` and optional `--fix`. `--fix` will attempt to fix lint issues automatically (applies only to `unused_declarations` and `test_inclusion` for now).  `--fix` always checks no uncommitted changes before modifying (can be overridden with `--force`). Fix will run `zig fmt` on files if any changes are made.

Adds steps to `build.zig`, from inside the `v2` directory:
```sh
zig build lint # runs lint check
zig build lint -- --fix # changes files to fix lint issues if possible
zig build lint -- --fix --force # applies fixes even if uncommitted changes detected
zig build lint-test # runs unit tests for lint tool
```

Made changes to CI to run the lint and also fixed dependency fetching (`proxy_workaround.sh` is needed to avoid errors when job runs, `zig build` seems to fail at fetching).

## Implementation Notes

(Some of this is after updates)

- Allocations are just leaked since it's a short lived CLI tool, this simplifies the code and makes execution faster, this could maybe be improved to use temporary arenas for scratch to reduce total memory use over an execution run but avoided for now as it's very little memory and simpler to just leak everywhere
- All files are found, loaded, and parsed before running through the lint rules - done for simplicity of the code
- Lint rules applying fixes/edits do so by updating the the loaded in memory file such that lint rules that run afterward have the updated file
- Files that are fixed/changed are written back to disk after all lint rules are executed
- Entirely single threaded for now, but should be easy to make parallel if needed by splitting the steps of file loading, parsing, and file level lint rules across threads (main work done and no contention for these steps)


v2/lint/main.zig
```diff
@@ -0,0 +1,237 @@
...
+    defer _ = debug_allocator.deinit();
+    const allocator = debug_allocator.allocator();
+
+    const parse_result = cli.parseArgs(allocator) catch |err| {
+        switch (err) {
+            error.InvalidArguments => {},
+            else => {
+                std.debug.print("unexpected lint CLI error: {s}\n", .{@errorName(err)});
+                if (@errorReturnTrace()) |trace| std.debug.dumpStackTrace(trace.*);
+            },
```
kprotty: main already prints out an stack trace when an error is returned

v2/lint/main.zig
```diff
@@ -0,0 +1,237 @@
...
+        _ = @import("line_length.zig");
+        _ = @import("test_inclusion.zig");
+        _ = @import("unused_declarations.zig");
+    }
+}
+
+const project_paths = [_][]const u8{ "build.zig", "init", "lib", "lint", "services" };
+const test_inclusion_roots = [_][]const u8{ "lib/lib.zig", "lint/main.zig" };
+
+/// Runs v2 lint and exits with 0 for no diagnostics, 1 for diagnostics, and 2 for CLI or internal
```
kprotty, line 24: why care about the error codes?

v2/lint/main.zig
```diff
@@ -0,0 +1,237 @@
...
+    if (ctx.config.ruleEnabled(.test_inclusion)) {
+        for (test_inclusion_roots) |root| {
+            try test_inclusion.lint(ctx, root, &files);
+        }
+    }
+
+    if (ctx.config.mode == .fix) {
+        try files.writeChanged();
+        try files.fmtChanged(ctx.allocator);
+    }
```
kprotty, line 105: like zig fmt, all rules should be enabled. no need to configure them from cli

verbose can be replaced with printing out the files it writes chages to

v2/lint/main.zig
```diff
@@ -0,0 +1,237 @@
...
+fn ensureFixModeCleanAtPath(ctx: *const core.Context, cwd: []const u8) !void {
+    if (ctx.config.mode != .fix or ctx.config.force) {
+        return;
+    }
+    if (try hasUncommittedChanges(ctx.allocator, cwd)) {
+        return error.UncommittedChanges;
+    }
+}
+
+fn hasUncommittedChanges(allocator: std.mem.Allocator, cwd: []const u8) !bool {
```
kprotty, line 117: i run zig fmt on uncommitted changes already (i assume most do, with it ran on saving to a file)

is this check necessary?

v2/lint/test_inclusion.zig
```diff
@@ -0,0 +1,1198 @@
...
+//! must contain managed test imports for `ed25519.zig`. Missing companion files are reported.
+//!
+//! Managed block format:
+//!
+//! ```zig
+//! comptime {
+//!     if (@import("builtin").is_test) {
+//!         _ = @import("foo.zig");
+//!     }
+//! }
```
kprotty, line 23: this should just be a test block instead of comptime-if:
```zig
test _ {
  _ = @import("foo.zig");
}
```

v2/lint/core.zig
```diff
@@ -0,0 +1,389 @@
...
+    pub fn addDiagnostic(
+        self: *Context,
+        path: []const u8,
+        line: usize,
+        column: usize,
+        rule: cli.Rule,
+        message: []const u8,
+    ) !void {
+        try self.addDiagnosticId(path, line, column, rule.id(), message);
+    }
```
kprotty, line 133: this helper seems unnecessary

v2/lint/core.zig
```diff
@@ -0,0 +1,389 @@
...
+
+    pub fn get(self: *const SourceFiles, path: []const u8) ?*const SourceFile {
+        const index = self.indexOf(path) orelse return null;
+        return &self.items.items[index];
+    }
+
+    pub fn getMut(self: *SourceFiles, path: []const u8) ?*SourceFile {
+        const index = self.indexOf(path) orelse return null;
+        return &self.items.items[index];
+    }
```
kprotty: a single mutable `.get()` thats mutable should work

v2/lint/core.zig
```diff
@@ -0,0 +1,389 @@
...
+    pub fn fmtChanged(self: *const SourceFiles, allocator: Allocator) !void {
+        for (self.items.items) |*file| {
+            if (!file.has_changes) {
+                continue;
+            }
+            try runZigFmt(allocator, file.path);
+        }
+    }
+
+    fn indexOf(self: *const SourceFiles, path: []const u8) ?usize {
```
kprotty: `std.sort.binarySearch`

v2/lint/cli.zig
```diff
@@ -0,0 +1,106 @@
...
+            config.deinit(allocator);
+            return .help;
+        } else if (std.mem.startsWith(u8, arg, "-")) {
+            std.debug.print("unknown flag: {s}\n", .{arg});
+            return error.InvalidArguments;
+        } else {
+            std.debug.print("unexpected argument: {s}\n", .{arg});
+            return error.InvalidArguments;
+        }
+    }
```
kprotty, line 58: could be simplified to `lint (check/help)?`, where its always fix & force unless `check/help` and verbose isnt needed

v2/lint/main.zig
jbuckmccready: Bug tracking, internal error -> problem in linter/environment.

v2/lint/main.zig
jbuckmccready: Related to exit code control.

v2/lint/main.zig
jbuckmccready: Yeah I think it's needed, it will delete unused private declarations (functions, constants, etc.) that you may want to keep or save elsewhere.

v2/lint/test_inclusion.zig
jbuckmccready: I don't like this because it adds a test entry for a nonexistent test, so then for the summary like this:

```
 == test block only import ==
 1/2 test_block_only_import.test_0...OK
 2/2 foo.test.foo test...OK
 All 2 tests passed.
```
`test_0` doesn't exist, it's just due to the `test` block that was added for inclusion, it also shows 2 tests passed when there was only 1 test.

v2/lint/core.zig
jbuckmccready: https://github.com/Syndica/sig/pull/1441/changes/fbba4b41e9e780520938a4734b78da5ce6b138f1

v2/lint/core.zig
jbuckmccready: Maybe there's a better way to organize it. The main "API" is add diagnostic for rule (enum), but there is the other signature for accepting a string since a parse error is also a diagnostic but doesn't come from a rule.

v2/lint/core.zig
jbuckmccready: I can change it, it will cause other function that currently accepts `*const SourceFiles` to require `* SourceFiles` even though it doesn't mutate anything, actually only needs a contains check: https://github.com/Syndica/sig/blob/fbba4b41e9e780520938a4734b78da5ce6b138f1/v2/lint/line_length.zig#L30-L44

v2/lint/cli.zig
jbuckmccready: I guess this relates to other comment on no split between `fix` and `check`.

v2/lint/main.zig
jbuckmccready: https://github.com/Syndica/sig/pull/1441/changes/08750bd2059451d5db265beeea7dcd343a6231d4

v2/lint/core.zig
kprotty: since `self.items.items` refers to a mutable slice that doesnt depend on the mutability of `self`, it can be `get(self: *const SourceFiles) *T`

v2/lint/main.zig
kprotty: fair, although that should be what `lint check` is for

v2/lib/crypto/ed25519.zig
```diff
@@ -28,6 +23,21 @@ const has_avx512 = builtin.cpu.arch == .x86_64 and
...
+}
+comptime {
+    if (@import("builtin").is_test) {
+        // lint: skip ed25519/avx512.zig
+        _ = @import("ed25519/generic.zig");
+        _ = @import("ed25519/pippenger.zig");
+        _ = @import("ed25519/straus.zig");
+        _ = @import("ed25519/wycheproof.zig");
+    }
+}
```
ajw221, line 48: was wondering why this can't be done instead?
```suggestion
test {
    // lint: skip ed25519/avx512.zig
    _ = @import("ed25519/generic.zig");
    _ = @import("ed25519/pippenger.zig");
    _ = @import("ed25519/straus.zig");
    _ = @import("ed25519/wycheproof.zig");
}
```

ajw221 (changes requested):

v2/lint/core.zig
jbuckmccready: Ah right const is always shallow.

https://github.com/Syndica/sig/pull/1441/changes/00dc759ff2810f7e4dd6cb0e5b1c8ed1515c4c00

v2/lib/crypto/ed25519.zig
jbuckmccready: https://github.com/Syndica/sig/pull/1441#discussion_r3211094872

v2/lint/test_inclusion.zig
dnut: I agree with Buck. I prefer a bit of extra boilerplate to avoid misleading output.

.github/workflows/check_linux.yml
```diff
@@ -156,7 +156,7 @@ jobs:
       - uses: mlugg/setup-zig@v2
         with:
           version: 0.15.2
-          cache-key: selfhosted-${{ hashFiles('build.zig', 'build.zig.zon') }}
+          cache-key: selfhosted-${{ hashFiles('build.zig', 'build.zig.zon', 'v2/build.zig', 'v2/build.zig.zon') }}
```
InKryption, line 160: Adding v2's build.zig{,.zon} here seems contradictory with the removal of the python script running in them just below in this same file? Unless I'm misunderstanding something.

v2/lib/gossip/node.zig
```diff
@@ -384,8 +383,9 @@ pub fn GossipNode(comptime Effects: type) type {
                     var prune_len: usize = 0;
 
                     for (push.values.items) |value| {
-                        const key, const duplicates =
-                            (try self.insertValue(.from(logger), now, .push, value)) orelse continue;
+                        const maybe_insert =
+                            try self.insertValue(.from(logger), now, .push, value);
+                        const key, const duplicates = maybe_insert orelse continue;
```
InKryption: Nit: this would be more accurately named something like `maybe_key_value`.

v2/lib/gossip/node.zig
```diff
@@ -819,7 +823,8 @@ pub fn GossipNode(comptime Effects: type) type {
             now: u64,
             value: GossipValue,
         ) !void {
-            const key, _ = (try self.insertValue(.from(logger), now, .us, value)) orelse unreachable;
+            const maybe_insert = try self.insertValue(.from(logger), now, .us, value);
+            const key, _ = maybe_insert orelse unreachable;
```
InKryption: Ditto

v2/lint/core.zig
```diff
@@ -0,0 +1,361 @@
...
+            try files.items.append(allocator, file);
+        }
+
+        return files;
+    }
+
+    pub fn deinit(self: *SourceFiles) void {
+        for (self.items.items) |*file| file.deinit(self.allocator);
+        self.items.deinit(self.allocator);
+    }
```
InKryption: Seems like this would be simplified by just using an arena.

v2/lint/core.zig
```diff
@@ -0,0 +1,361 @@
...
+    for (diagnostics) |diag| {
+        try out.print("{s}:{d}:{d}: error[{s}]: {s}\n", .{
+            diag.path,
+            diag.line,
+            diag.column,
+            diag.rule_id,
+            diag.message,
+        });
+    }
+}
```
InKryption, line 164: Should either just use `std.debug.print`, or accept an `*std.Io.Writer`.

v2/lint/core.zig
```diff
@@ -0,0 +1,361 @@
...
+        },
+        else => {},
+    }
+}
+
+pub fn collectDirRecursive(
+    allocator: Allocator,
+    dir_path: []const u8,
+    paths: *std.ArrayList([]const u8),
+) anyerror!void {
```
InKryption: Any benefit of all this over `std.fs.Dir.walk`?

v2/lint/main.zig
```diff
@@ -0,0 +1,233 @@
...
+    try unused_declarations.lint(ctx, file);
+}
+
+test "parse errors report diagnostic and skip file-level rules" {
+    const allocator = std.testing.allocator;
+    const path = "lint/.tmp_parse_error_test.zig";
+    const source = "const x = ; // 12345678901234567890123456789012345678901234567890" ++
+        "12345678901234567890123456789012345678901234567890\n";
+    try std.fs.cwd().writeFile(.{ .sub_path = path, .data = source });
+    defer std.fs.cwd().deleteFile(path) catch {};
```
InKryption: Preferably use `std.testing.tmpDir`. If this `lint` path is something that's also hard-coded elsewhere in the code I've not seen, I think it would be better to pass this directory as a parameter somewhere instead.

v2/lint/main.zig
```diff
@@ -0,0 +1,233 @@
...
+fn tempRepoPaths(
+    allocator: std.mem.Allocator,
+    sub_path: []const u8,
+) !struct { repo: []u8, v2: []u8 } {
+    const repo = try std.fmt.allocPrint(allocator, ".zig-cache/tmp/{s}/repo", .{sub_path});
+    errdefer allocator.free(repo);
+    const v2 = try std.fmt.allocPrint(allocator, "{s}/v2", .{repo});
+    errdefer allocator.free(v2);
+    return .{ .repo = repo, .v2 = v2 };
+}
```
InKryption: `.zig-cache` isn't a hard-coded directory, it can be overriden with environment variables and on the command line, which I do on occasion for a bunch of reasons. This should be inlined into its one call-site, and refactored to make direct use of the `tmpDir` handle instead of performing this brittle path-based logic.

v2/lint/test_inclusion.zig
```diff
@@ -0,0 +1,1250 @@
...
+    imports: std.ArrayList([]const u8) = .empty,
+};
+
+const managed_test_inclusion_wrapper = "if (@import(\"builtin\").is_test) {";
+const obsolete_ref_all_decls = "std.testing.refAllDecls(@This())";
+
+pub fn lint(
+    ctx: *core.Context,
+    root_file: []const u8,
+    source_files: *core.SourceFiles,
```
InKryption: Why by mutable ref if it's not being mutated.

v2/lint/core.zig
```diff
@@ -0,0 +1,361 @@
...
+        return self.ast.errors.len != 0;
+    }
+
+    pub fn replaceSource(
+        self: *SourceFile,
+        allocator: Allocator,
+        source: [:0]u8,
+        fix_name: []const u8,
+    ) !void {
+        errdefer allocator.free(source);
```
InKryption: I find this type of transfer of ownership to be rather confusing, especially without a doc comment. Should at minimum add a doc comment, but ideally would move this errdefer to the caller, and add a comment describing the transfer of ownership, with the asterisk that ownership is only transferred upon a successful return of the function.

Note: this code, much like the other code I commented on, and much of the rest of this code, could be simplified by using an arena allocator.

v2/lint/test_inclusion.zig
```diff
@@ -0,0 +1,1250 @@
...
+            companion_path,
+            .{ .line = 1, .column = 1 },
+            "missing managed test inclusion block",
+        );
+        return;
+    };
+
+    if (std.meta.eql(@as([]const []const u8, block.imports.items), expected_imports)) {
+        return;
+    }
```
InKryption: Are these pointers expected to to be directly equal? If not, take note that `std.meta.eql` does not follow pointers. If yes, a doc comment explaining why, and why this is being done this way, would be appreciated.

v2/lint/unused_declarations.zig
```diff
@@ -0,0 +1,727 @@
...
+    const path = "lint/.tmp_unused_declarations_fix.zig";
+    const source =
+        \\const Writer = fmt.Writer;
+        \\const fmt = std.fmt;
+        \\const alias = std;
+        \\const std = @import("std");
+        \\
+    ;
+    try std.fs.cwd().writeFile(.{ .sub_path = path, .data = source });
+    defer std.fs.cwd().deleteFile(path) catch {};
```
InKryption: Should also be using `std.testing.tmpDir`

v2/services/telemetry.zig
```diff
@@ -1,7 +1,5 @@
 /// This service consumes information from other services,
 /// and sends them to an aggregator (prometheus).
-const tel = @This();
-
```
InKryption, line 4: This declaration was being used for the doc comment above it. Removing it means the doc comment now applies to the one just underneath it; in other cases, this would be a malformed code change, for example:
```zig
/// comment
const foo = 0;
test {}
```
removing `foo` here means the comment will seemingly apply to the unit test, which isn't allowed syntax.

v2/lint/test_inclusion.zig
```diff
@@ -0,0 +1,1250 @@
...
+
+    for (expected) |import_path| {
+        if (containsString(skipped, import_path)) continue;
+        try filtered.append(allocator, import_path);
+    }
+
+    return filtered;
+}
+
+fn findManagedBlocks(allocator: Allocator, source: []const u8) !std.ArrayList(ManagedBlock) {
```
InKryption: Any reason you didn't use `std.zig.Ast` here? Seems like it would be among the ideal usages.

InKryption (changes requested):
A general comment: I think the code is a bit over-complicated from memory management for complicated nested allocation, and seems to be trying to amortize these allocations into central points, but I think given the short-lived nature of this utility's execution, it would benefit greatly from just using an arena allocator and constructing things on-demand in more places (e.g. parsing into `std.zig.Ast` as needed, instead of having to do it all up-front [unless there's some specific design reason for that]).

.github/workflows/check_linux.yml
jbuckmccready: The V2 CI checks are run in this job, this fixes an existing bug where they were not included. I removed the python script since this tool replaces it entirely for v2.

v2/lib/gossip/node.zig
jbuckmccready: https://github.com/Syndica/sig/pull/1441/changes/0dacdc8bfb08aecaa6c152fb2a397fe2818146df

v2/lib/gossip/node.zig
jbuckmccready: https://github.com/Syndica/sig/pull/1441/changes/0dacdc8bfb08aecaa6c152fb2a397fe2818146df

v2/lint/core.zig
jbuckmccready: Yeah definitely.

https://github.com/Syndica/sig/pull/1441/changes/0e679f1d065e6f70cf0b631851664f15b4aa3cd5

v2/lint/core.zig
jbuckmccready: https://github.com/Syndica/sig/pull/1441/changes/141ca3b9ae9f1a9ee9b2afa5d7b59a4da93fdc53

v2/lint/core.zig
jbuckmccready: Honestly I did not look at the `std` for this. I've been avoiding the `std` library for anything that isn't primitive (like file IO) and keeping it as minimal as possible. The docs are mostly non-existent and API is usually not thorough (usually proof of concept), so often times it feels like more work to use as I end up having to just read the code and deal with the specifics anyways (easier to just read and maintain the code inline for exactly what is needed), and it has the added downside of the API possibly changing in future Zig versions which adds more dev time.

I took a look at the API, it doesn't support filtering out directories, so it can end up walking things like `.zig-cache` - which shouldn't affect things now for how things are organized, but it seems sensible to skip known build artifacts to allow the tool to "just work" without having to waste time dealing with it in the future if you clone a repo into a subdir, or the lint is given different paths.

v2/lint/main.zig
jbuckmccready: https://github.com/Syndica/sig/pull/1441/changes/60000e27326fa83dcd4be1034b3553c8a254779d

v2/lint/main.zig
jbuckmccready: https://github.com/Syndica/sig/pull/1441/changes/60000e27326fa83dcd4be1034b3553c8a254779d

v2/lint/test_inclusion.zig
jbuckmccready: Left over from old method use that assumed const not shallow.

https://github.com/Syndica/sig/pull/1441/changes/6bcfe7ff12ed7bb6263afb8e0e6aaf2bd2103322

v2/lint/core.zig
jbuckmccready: https://github.com/Syndica/sig/pull/1441/changes/d7e0a6f4e174f08d3f46524f5b0f83ad04765c00

jbuckmccready: Regarding arena for the whole thing, that's probably a better approach. I was keeping it less "leaky" initially and the mental model more scoped to the types - but just using top level arena that grows is more aligned with the tool.

The current approach:
1. Read and parse all the files up front for passing into lints
2. Run lints on files, if a lint is applying a change/fix then it can refresh/update the file and AST so the next lints work with the updated file as expected (this frees memory used for the old version of the file and AST)
3. Write all files to disk that have been changed by lints

I think it could be same approach but just use arena and don't bother freeing anything as it replaces/makes changes (all ownership is scoped to the top level arena for files).

v2/lint/test_inclusion.zig
jbuckmccready: Yeah this code is wrong. I wasn't sure how slice was identified by `@typeInfo` and when making a test program with const slices the pointers ended up being equal for different declarations due to optimizations for compile time constants (put in same place of program memory).

jbuckmccready: I just removed the short circuiting entirely, it was a check that was left over from old code structure, acting as a silly micro optimization at this point.

https://github.com/Syndica/sig/pull/1441/changes/41973a4e692254116dfb0b4ed9b904a3f2a6bf99

v2/lint/unused_declarations.zig
jbuckmccready: https://github.com/Syndica/sig/pull/1441/changes/26cdc254ffcab12ccef8d891e756ba02f51dccc5

v2/services/telemetry.zig
jbuckmccready: Oops, I added logic to the lint fix to avoid this type of mess up (avoids autofix if there is comment directly above the unused declaration), but I added it after running the lint fix and missed this goof.

I fixed this particular file to use file level doc comment and I made changes to the lint logic to ensure it doesn't just silently skips fixing when it can't, it now reports diagnostics for when it can't. Also simplified the code a bit and added a test case for same line multiple unused.

https://github.com/Syndica/sig/pull/1441/changes/cdcecdddc2dad9d8ee6b8562ef58b43d024e07e4

v2/lint/test_inclusion.zig
jbuckmccready: I looked into it a bit but it felt like it made the code harder to read/understand because the pattern is very exact and not arbitrarily nested. It seemed to add abstraction without simplifying the code.

v2/lint/core.zig
InKryption: For the call-sites that now expect an arena, they should probably use `arena: std.mem.Allocator`, to adjust expectations.

v2/services/telemetry.zig
InKryption: Just a note that this decl-for-doc-comment pattern may become required in the future if the proposal to remove top level doc comments is accepted. Change is fine for now for uniformity.

v2/lint/core.zig
InKryption: That's fair; I forgot that `walkSelectively` was in 0.15. Can maybe update it to use that later.

v2/lint/core.zig
jbuckmccready: https://github.com/Syndica/sig/pull/1441/commits/1138c450f7ef1eea35c47b86cf08e6be7855f1e6

v2/lint/test_inclusion.zig
kprotty: IMO, the output isnt misleading (one looks for specific tests, not the total tests, since the `-Dfilter` isnt accurate and usually adds unrelated tests in anyways).

But its not that important. Feel free to resolve.

v2/lint/main.zig
jbuckmccready: I like having it extra explicit for these types of global destructive edits, but not a hill I need to die on if the team wants it different. 

Current approach (matches clippy fix behavior):
- `lint check`: just runs check and reports issues
- `lint fix`: applies all fixes if there are no uncommitted changes (clean state)
- `lint fix --force`: applies all fixes even if there are uncommitted changes

`lint fix` is very destructive, it will delete unused private functions entirely.

kprotty: i see. Personally, i run `zig fmt` on staged changes instead of committed changes (to avoid creating a separate `zig fmt` commit, but that sometimes slips through). Thoughts on downgrading this check from "uncomitted" to "unstaged"?

jbuckmccready: My concern with distinguishing on staged is I think that can lead to even more confusion or problems depending on the git knowledge and/or tools used (it's less clear how to revert changes, it's easy to accidentally stage a file which loses the diff, etc.).

The reason I'm emphasizing this to be different from `zig fmt` is that it is highly destructive, as in `rm -rf` your functions and structs, `zig fmt` will only make formatting changes so there's no risk of actual code loss.

For scenario you described this should work to avoid additional commit noise (git tools I have used all support this amend workflow as well):
```sh
git commit -m "..."
lint fix
git add ...
git commit --amend --no-edit
```

kprotty: Realized its not a replacement for `zig fmt`, but more for `scripts/style.py` with fixing over just checking

kprotty (dismissed):

InKryption (approved):

kprotty (approved):

ajw221 (approved):

PR #1442, opened 2026-05-11
feat(v2:telemetry): more metrics & grafana dashboards

### LINKED ISSUE #1385: feat(v2): dashboards from telemetry data
We currently have a telemetry service with can be used to "export" various metrics.

However currently we aren't making much use of it in other services, and we don't have a dashboard or anything set up  for them.

InKryption:


v2/lib/telemetry.zig
```diff
@@ -243,125 +243,184 @@ pub fn Logger(comptime scope_str: []const u8) type {
...
-                        .logger = self.logger,
-                        .level = self.level,
-                        .entries = self.entries ++ .{new_entry},
-                    };
-                }
+/// Equivalent to `Logger`, but with a runtime `scope`.
+/// This is intended as the parameter of "leaf" or "branch" functions that want to inherit the
+/// caller's scope; it is for this reason that it can only convert to and from an actual scoped
+/// logger, and not another `AnyLogger`.
+pub const AnyLogger = struct {
```
dnut: Is this abstraction needed yet? I don't see any usages of `AnyLogger` outside this file.

My preference would be for a log message's scope to reflect the location where it is emitted. That matches how loggers commonly behave across languages, and it gives readers a stable, predictable signal about where the log originated.

I agree there can be value in adding tracing context, especially for general-purpose functions called from many places. I just would not overload `scope` to serve both purposes, since that makes the field ambiguous for people reading the logs. I would rather keep `scope` as the source location and add a separate contextual field for tracing or caller-specific information. That preserves a conventional and precise meaning for `scope` while still supporting the additional context this is aiming for.

Given that, I think it would be simpler to remove this abstraction for now and keep `scope` static. Supporting both static and dynamic scope adds another layer of complexity that callers and readers need to reason about. We can add dynamic context fields once there is a clear need for them.

InKryption: I originally did have a change that used this, but it was swept away after a rebase; the main reason I introduced it was because there was a general-purpose function in gossip which was called from many contexts, and in fact its behaviour depended explicitly on who the caller was (it had an enum that directly said who the caller was).

But yeah, given it isn't needed, I have no issue removing it, long live YAGNI.

InKryption: Dropped the commit

v2/lib/time.zig
```diff
@@ -0,0 +1,88 @@
...
+    /// This is not guaranteed to be monotonic or steadily increasing, but for
+    /// most implementations it is.
+    /// Returns `error.Unsupported` when a suitable clock is not detected.
+    pub fn now() Instant {
+        const clock_id: std.posix.CLOCK = .BOOTTIME;
+        const ts = std.posix.clock_gettime(clock_id) catch |err| {
+            std.debug.panic("clock_gettime unsupported: {}", .{err});
+        };
+        return .{ .timestamp = ts };
+    }
```
prestonsn, line 23: I like the `Instant` and `Timer` abstractions a lot. I've got a couple questions tho: 
1.  Is there a reason why we don't use `.MONOTONIC` here? last i checked, the std lib uses it too (but perhaps that's part of the reason why its getting overhauled?)
2. Could we integrate/use `lib.clock` here in some way? It has the warmup path already setup. Wondering if this separation was intended

v2/lib/gossip/node.zig
```diff
@@ -1132,24 +1229,53 @@ pub fn GossipNode(comptime Effects: type) type {
...
+        fn sendMessage(self: *Self, addr: std.net.Address, msg: GossipMessage) void {
             const packet: *Packet = self.config.effects.writePacket();
             packet.addr = addr;
 
             var writer: std.Io.Writer = .fixed(&packet.data);
-            try bincode.write(&writer, msg);
+            bincode.write(&writer, msg) catch |err| switch (err) {
+                error.Deprecated => unreachable, // we should not be sending deprecated types
+                error.WriteFailed => unreachable, // we should not send values that are too large
+            };
```
prestonsn, line 1244: I like this change, do you think it would be useful to also track these cases in gossip's metrics?

v2/lib/snapshot/download.zig
```diff
@@ -649,51 +649,51 @@ const UserData = packed struct(u64) {
...
-    snapshot_sq_fulls: tel.Counter,
-    snapshot_download_candidates: tel.Counter,
-    snapshot_downloads_started: tel.Counter,
-    snapshot_downloads_failed: tel.Counter,
-    snapshot_downloads_cancelled: tel.Counter,
-    snapshot_downloads_succeeded: tel.Counter,
-    snapshot_download_bytes_written: tel.Counter,
+    sources_received: tel.Counter,
+    sources_deduped: tel.Counter,
+    sources_new: tel.Counter,
```
prestonsn, line 654: Thank you for cleaning up my mess 🫡

v2/lib/snapshot/download.zig
```diff
@@ -713,10 +713,12 @@ pub const Metrics = struct {
 
     // TODO: This was to not have a big fat anon. struct with the same fields repeated. But perhaps there's a better way?
     pub const fields_config = blk: {
-        var config: tel.metric.FieldsConfig(Metrics) = .{};
+        var config: tel.metric.FieldsConfig(Metrics) = .{
+            .prefix = "snapshot",
```
prestonsn, line 717: 📝

v2/metrics/grafana/dashboards/gossip.json
```diff
@@ -0,0 +1,841 @@
...
+      "id": 6,
+      "panels": [],
+      "title": "Incoming",
+      "type": "row"
+    },
+    {
+      "datasource": {
+        "type": "prometheus",
+        "uid": "PBFA97CFB590B2093"
+      },
```
prestonsn, line 41: I found these hard-coded uids for the datasource a little annoying to deal with when setting things up local. Maybe we could update [datasource.yml ](https://github.com/Syndica/sig/blob/ink/more-metrics-and-grafana/v2/metrics/grafana/datasources/datasource.yml)?

I ended up adding this to that datasource.yml file, and updating each of the above datasource config structs in every dashboard to use the same uid:

```yml
    uid: sig-prometheus
```

prestonsn (commented):
LGTM overall, just a few minor things I noticed

v2/lib/gossip/node.zig
prestonsn: actually never mind, hard failing/panicking is enough of a "metric". These cases should be fatal anyways

v2/lib/time.zig
InKryption: On point 1, not sure iiuc. Are you saying stdlib uses MONOTONIC? Because that may be true for something else, but the Instant abstraction uses BOOTTIME on linux. If you're suggesting replacing BOOTTIME to stray from what stdlib was using, then sure I suppose, I think it probably doesn't matter a huge amount in our case.
On point 2, yeah, I implemented this before lib.clock got implemented. I'll give this another pass and probably move this into there.

v2/lib/gossip/node.zig
InKryption: Well, if these cases were hit, it would crash; the idea is that they're not reachable. This part of the changeset was mostly to help narrow down gossip's errors for telemetry purposes.

v2/lib/time.zig
prestonsn: on 1, I see your point actually. I was mostly curious as why we chose BOOTTIME here. I checked the zig std lib and was totally wrong on stating it uses MONOTONIC, I thought it did for some reason. I'm cool with sticking with BOOTTIME for this

v2/metrics/grafana/dashboards/gossip.json
InKryption: I'll be real I'm not fully familiar with this aspect of grafana, the only material I could find on it said I should leave it be (could be misunderstanding something, I didn't find much good material).
Think you could outline the exact changes? You could commit them directly to this branch if you think they're valuable.

dnut: @prestonsn do you handwrite these files? I've always built dashboards in the grafana gui and then exported the json without even looking at it.

InKryption: Yeah, I'm the same as Drew here in this respect. I looked at the output, but only to verify there wasn't anything obviously wrong or strange, never edited them directly.

prestonsn: Yeah that's typically what I do too, but for some reason directly importing these dashboards didn't work (the error I kept seeing was "Datasource PBFA97CFB590B2093 not found" despite grafana being able to talk to prometheus). I can look into this more today

prestonsn: Can confirm after setting up docker compose that the source id issue is no longer a concern (doesn't need to get setup again on a new host once configured).

dnut (approved):

prestonsn (approved):

PR #1446, opened 2026-05-12
chore(conformance): revert supported but incomplete features

yewman:
This PR includes some improvements to feature comments and descriptions relating them to SIMD's. The core changes relate to reverting some partially or unimplemented features from `supported` to `unsupported` to prevent excessive octane failures. Once the features have been implemented and locally fuzzed, we can update their status to `supported` once again.
`enable_sbpf_v3_deployment_and_execution`: .supported -> .unsupported
`virtual_address_space_adjustments`: .supported -> .unsupported

### DIFF
--- a/src/core/features.zon
+++ b/src/core/features.zon
@@ -179,11 +179,11 @@
         .description = "fail libsecp256k1_verify if count appears wrong",
         .status = .hardcoded,
         .note =
-        \\\ TODO: Low Priority
-        \\\ Inactive on all clusters.
-        \\\ Hardcoded for compatibility.
-        \\\ Marked as hardcoded for fuzzing in solfuzz-agave and firedancer.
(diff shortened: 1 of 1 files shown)


InKryption (approved):

src/core/features.zon
```diff
@@ -633,7 +638,7 @@
     .{
         .name = "cap_accounts_data_size_per_block",
         .pubkey = "qywiJyZmqTKspFg2LeuUHqcA5nNvBgobqb9UprywS9N",
-        .description = "TODO: Add description",
+        .description = "",
```
dnut, line 641: is this an improvement? do we explicitly want no description here? i think a description would still be nice to have

the same idea applies in other empty descriptions

yewman: The only features which have no description are reverted features. I don't think a description provides any value in these instances

dnut (approved):

PR #1447, opened 2026-05-13
fix(cpi): remove unwrap in cpi region updates

yewman:
WIP - currently being tested

fixes #1445 - bug revealed after `virtual_address_space_adjustments` activation on testnet

### DIFF
--- a/src/vm/syscalls/cpi.zig
+++ b/src/vm/syscalls/cpi.zig
@@ -1155,10 +1155,19 @@ pub fn invokeSigned(AccountInfo: type) sig.vm.SyscallFn {
                                 else
                                     .constant;
 
-                                const data = if (account_data_direct_mapping)
+                                const data: []u8 = if (account_data_direct_mapping)
                                     callee_account.account.data
                                 else
-                                    region.hostSlice(state).?[0..callee_account.account.data.len];
+                                    // TODO: Fix this!!
+                                    // After `virtual_address_space_adjustments` memory regions may be switched between writable/non-writable
+                                    // during CPI. Given our current MemoryState and HostMemory couples the writability of a region to the slice type
+                                    // we are unable to promote a constant region to a mutable region. In practice during runtime all region host memory
+                                    // is allocated so constCast should okay however this is not something we should rely on. We need to rework our
+                                    // HostMemory and MemoryState implementation to decouple the writability of a region from the slice type.
+                                    switch (region.host_memory) {
+                                        .mutable => |slice| slice,
+                                        .constant => |slice| @constCast(slice),
+                                    }[0..callee_account.account.data.len];
 
                                 region.* = .init(state, data, region.vm_addr_start);
                             },
--- a/src/runtime/program/bpf_loader/execute.zig
+++ b/src/runtime/program/bpf_loader/execute.zig
@@ -242,7 +242,8 @@ fn handleExecutionResult(
             if (deplete_cu_meter and err_kind != .Syscall)
                 compute_meter.* = 0;
             if (virtual_address_space_adjustments and err == error.AccessViolation)
-                std.debug.print("TODO: Handle AccessViolation: {s}\n", .{@errorName(err)});
+                // std.debug.print("TODO: Handle AccessViolation: {s}\n", .{@errorName(err)});
+                return error.AccessViolation;
             return err;
         },
     }


src/vm/syscalls/cpi.zig
```diff
@@ -1158,7 +1158,7 @@ pub fn invokeSigned(AccountInfo: type) sig.vm.SyscallFn {
                                 const data = if (account_data_direct_mapping)
                                     callee_account.account.data
                                 else
-                                    region.hostSlice(state).?[0..callee_account.account.data.len];
+                                    @constCast(region.constSlice())[0..callee_account.account.data.len];
```
dnut: `@constCast` is not safe to use unless you can guarantee that it was actually mutable memory that was passed in when the region was initialized. A safer approach would require a mut pointer to be used at initialization, but force any accesses to use a const pointer, and provide a method for the explicit conversion for this use case.

For example:

```diff
diff --git a/src/vm/memory.zig b/src/vm/memory.zig
index 3e36144c..f1de4910 100644
--- a/src/vm/memory.zig
+++ b/src/vm/memory.zig
@@ -211,7 +211,7 @@ pub const MemoryState = enum {
 
 const HostMemory = union(MemoryState) {
     mutable: []u8,
-    constant: []const u8,
+    constant: []u8,
 
     fn getSlice(self: HostMemory, comptime state: MemoryState) !state.Slice() {
         if (self != state) return error.AccessViolation;
@@ -226,13 +226,13 @@ pub const Region = struct {
     vm_gap_shift: std.math.Log2Int(u64),
     vm_addr_end: u64,
 
-    pub fn init(comptime state: MemoryState, slice: state.Slice(), vm_addr: u64) Region {
+    pub fn init(comptime state: MemoryState, slice: []u8, vm_addr: u64) Region {
         return initGapped(state, slice, vm_addr, 0);
     }
 
     pub fn initGapped(
         comptime state: MemoryState,
-        slice: state.Slice(),
+        slice: []u8,
         vm_addr: u64,
         vm_gap_size: u64,
     ) Region {
@@ -267,6 +267,16 @@ pub const Region = struct {
         }
     }
 
+    /// This should only be used when we've verified that a region that was
+    /// originally classified as readonly should actually be permitted to be
+    /// mutable. This is a very rare special case, use with caution.
+    pub fn upgradeToMutable(self: Region) []u8 {
+        switch (self.host_memory) {
+            .constant => |constant| return constant,
+            .mutable => |mutable| return mutable,
+        }
+    }
+
     fn translate(
         self: Region,
         comptime state: MemoryState,
diff --git a/src/vm/syscalls/cpi.zig b/src/vm/syscalls/cpi.zig
index 482bd51d..162add8f 100644
--- a/src/vm/syscalls/cpi.zig
+++ b/src/vm/syscalls/cpi.zig
@@ -1158,7 +1158,7 @@ pub fn invokeSigned(AccountInfo: type) sig.vm.SyscallFn {
                                 const data = if (account_data_direct_mapping)
                                     callee_account.account.data
                                 else
-                                    @constCast(region.constSlice())[0..callee_account.account.data.len];
+                                    region.upgradeToMutable()[0..callee_account.account.data.len];
 
                                 region.* = .init(state, data, region.vm_addr_start);
                             },
```

The code actually does not compile with this patch because there are many usages of Region that pass in data that is actually const. Using `@constCast` to write to any of those pointers has undefined behavior.

ajw221: were these changes tested?

> `@constCast` is not safe to use unless you can guarantee that it was actually mutable memory that was passed in when the region was initialized. A safer approach would require a mut pointer to be used at initialization, but force any accesses to use a const pointer, and provide a method for the explicit conversion for this use case.
> 
> For example:
> 
> ```diff
> diff --git a/src/vm/memory.zig b/src/vm/memory.zig
> index 3e36144c..f1de4910 100644
> --- a/src/vm/memory.zig
> +++ b/src/vm/memory.zig
> @@ -211,7 +211,7 @@ pub const MemoryState = enum {
>  
>  const HostMemory = union(MemoryState) {
>      mutable: []u8,
> -    constant: []const u8,
> +    constant: []u8,
>  
>      fn getSlice(self: HostMemory, comptime state: MemoryState) !state.Slice() {
>          if (self != state) return error.AccessViolation;
> @@ -226,13 +226,13 @@ pub const Region = struct {
>      vm_gap_shift: std.math.Log2Int(u64),
>      vm_addr_end: u64,
>  
> -    pub fn init(comptime state: MemoryState, slice: state.Slice(), vm_addr: u64) Region {
> +    pub fn init(comptime state: MemoryState, slice: []u8, vm_addr: u64) Region {
>          return initGapped(state, slice, vm_addr, 0);
>      }
>  
>      pub fn initGapped(
>          comptime state: MemoryState,
> -        slice: state.Slice(),
> +        slice: []u8,
>          vm_addr: u64,
>          vm_gap_size: u64,
>      ) Region {
> @@ -267,6 +267,16 @@ pub const Region = struct {
>          }
>      }
>  
> +    /// This should only be used when we've verified that a region that was
> +    /// originally classified as readonly should actually be permitted to be
> +    /// mutable. This is a very rare special case, use with caution.
> +    pub fn upgradeToMutable(self: Region) []u8 {
> +        switch (self.host_memory) {
> +            .constant => |constant| return constant,
> +            .mutable => |mutable| return mutable,
> +        }
> +    }
> +
>      fn translate(
>          self: Region,
>          comptime state: MemoryState,
> diff --git a/src/vm/syscalls/cpi.zig b/src/vm/syscalls/cpi.zig
> index 482bd51d..162add8f 100644
> --- a/src/vm/syscalls/cpi.zig
> +++ b/src/vm/syscalls/cpi.zig
> @@ -1158,7 +1158,7 @@ pub fn invokeSigned(AccountInfo: type) sig.vm.SyscallFn {
>                                  const data = if (account_data_direct_mapping)
>                                      callee_account.account.data
>                                  else
> -                                    @constCast(region.constSlice())[0..callee_account.account.data.len];
> +                                    region.upgradeToMutable()[0..callee_account.account.data.len];
>  
>                                  region.* = .init(state, data, region.vm_addr_start);
>                              },
> ```
> 
> The code actually does not compile with this patch because there are many usages of Region that pass in data that is actually const. Using `@constCast` to write to any of those pointers has undefined behavior.

yewman: No just a suggestion from the looks of it. It is valid though, and unfortunately changing out memory state to use const slices for both types affects a lot of code

PR #1449, opened 2026-05-14
feat(runtime): implement AccessViolationHandler for SIMD-0460

ajw221:
**Summary**

- Complete implementation of the access violation handler for SIMD-0460.
- Our current Region implementation requires a `@constCast` to convert from a non-writable to a writable region in `invokeSigned` since the underlying memory for non-writable Region's is stored as a `[]const u8`. This is okay as a temporary fix since the underlying memory is allocated however it is not a good long term solution and needs to be replaced by updating our representation of host memory. 

Fixes: https://github.com/Syndica/sig/issues/1445

**TODO**
- [] complete local fuzzing

### DIFF
--- a/src/core/features.zon
+++ b/src/core/features.zon
@@ -1539,7 +1539,7 @@
         .name = "virtual_address_space_adjustments",
         .pubkey = "7VgiehxNxu53KdxgLspGQY8myE6f7UokaWa4jsGcaSz",
         .description = "SIMD-0460: Virtual Address Space Adjustments",
-        .status = .unsupported,
+        .status = .supported,
         .note =
         \\ TODO: High Priority 
         \\ Pending testnet activation.
--- a/src/core/features.zon
+++ b/src/core/features.zon
@@ -1540,15 +1540,6 @@
         .pubkey = "7VgiehxNxu53KdxgLspGQY8myE6f7UokaWa4jsGcaSz",
         .description = "SIMD-0460: Virtual Address Space Adjustments",
         .status = .supported,
-        .note =
-        \\ TODO: High Priority 
-        \\ Pending testnet activation.
-        \\ Implementation has been complete for some time, it is being temporarily reverted to unsupported to 
-        \\ because it needs to be audited and reactivated on octane. It is currently not hardcoded for fuzzing in 
-        \\ firedancer or agave so can be switched off to restrict fuzzing in octane until audit and local fuzzing is
-        \\ completed. The current implementation was ported from our previous implementation of 'stricter_abi_and_runtime_constraints' and
-        \\ needs a once over to ensure correctness.
-        ,
     },
     .{
         .name = "account_data_direct_mapping",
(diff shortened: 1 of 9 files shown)


yewman (dismissed):

src/runtime/program/bpf_loader/execute.zig
```diff
@@ -241,14 +279,175 @@ fn handleExecutionResult(
...
+            //   `MAX_PERMITTED_DATA_INCREASE` zero bytes after the account
+            //   payload. Extending the slice into that padding is safe and
+            //   matches the buffer's contents (zeros = zero-extension).
+            if (ctx.direct_mapping) {
+                region.host_memory = .{ .mutable = account.data };
+            } else {
+                const ptr = region.host_memory.mutable.ptr;
+                region.host_memory = .{ .mutable = ptr[0..new_len] };
+            }
+            region.vm_addr_end = region.vm_addr_start +| @as(u64, new_len);
```
kprotty, line 388: Double checking that is is right: https://github.com/anza-xyz/agave/blob/v4.0/transaction-context/src/transaction.rs#L519-L543

particularly that the mutability of the region is changed inside the `if` instead of outside like in agave

yewman (commented):
This AccessViolation remap path can retain stale `last_access_violation` metadata from an earlier handled access. If the handler grows/remaps the account and execution continues, a later unrelated `AccessViolation` can still be remapped using that stale state, which returns the wrong final `InstructionError` (`InvalidRealloc` in the regression instead of preserving the generic `AccessViolation`). I have implemented to mark access violations as `handled` to prevent remapping from stale metadata and an associated regression test `harnew/SIMD-0460-patch`

src/runtime/program/bpf_loader/execute.zig
ajw221: [81922f6](https://github.com/Syndica/sig/pull/1449/commits/81922f6058e37e4b609db20853e68f48b02a32f6)

src/core/features.zon
```diff
@@ -1539,7 +1539,7 @@
         .name = "virtual_address_space_adjustments",
         .pubkey = "7VgiehxNxu53KdxgLspGQY8myE6f7UokaWa4jsGcaSz",
         .description = "SIMD-0460: Virtual Address Space Adjustments",
-        .status = .unsupported,
+        .status = .supported,
         .note =
         \\ TODO: High Priority 
         \\ Pending testnet activation.
```
hamza-syndica: I think you can remove these TODO: too

ajw221: [c3c7683](https://github.com/Syndica/sig/pull/1449/commits/c3c768324bdd1e1370b46324b89bf842f84a6428)

kprotty (approved):

src/core/features.zon
dnut: > We went looking everywhere, but couldn’t find those commits.
> 
> Sometimes commits can disappear after a force-push. Head back to the [latest changes here](https://github.com/Syndica/sig/pull/1449/files).

personally i stop rebasing my prs once they enter review, to make it easier for reviewers to follow the progress

dnut (approved):

PR #1459, opened 2026-05-19
feat(v2:init): re-design the topology abstractions

InKryption:
`services.zon` now no longer features a `regions` field, instead comprised entirely by service entries.
The type tags previously described by the `regions` field are now expressed through the `Bind` function, where each tag of a union represents a distinct set of inter-process region equivalences.
Extracts a bunch of functionality out to be generic over the topology schema, literally parameterized by type `TopologySchema`, a thin wrapper over the `services.zon` data.
Also does a few other things to bring everything in line.


Sobeston (dismissed):
Ignoring the merge conflicts, this looks like a step in the right direction + can't see any flaws

v2/init/main.zig
```diff
@@ -159,15 +159,208 @@ pub fn main() !void {
...
+    }),
+
+    .telemetry = .initMany(&.{
+        .@"telemetry:main",
+        .@"net:telemetry",
+        .@"gossip:telemetry",
+        .@"shred_receiver:telemetry",
+        .@"snapshot:telemetry",
+    }),
+}));
```
kprotty, line 208: why define it here instead of in the service.zon? couldnt this be auto-generated from the latter?

v2/lib/TopologySchema.zig
```diff
@@ -0,0 +1,981 @@
...
+    return struct {
+        const bound = @This();
+
+        pub const RegionTag = @typeInfo(Region).@"union".tag_type.?;
+        pub const BindingsMap = topo.RegionBindingsMap(RegionTag);
+        pub const bindings_map = bindings_map_init;
+
+        fn regionSize(r: Region) usize {
+            return r.size();
+        }
```
kprotty, line 112: could this be inlined/moved to all call-sites?

v2/init/main.zig
InKryption: The idea is to have `services.zon` just describe what services actually exist, without defining the relationships between them, so that a test runner can specify an entirely different topology without having to just re-declare the schema.

InKryption: For example, a test could have a topology defined using just the `gossip` and a service which is just defined directly by the test service runner, with a pre-filled `.from_net` region, and a `.source_to_snapshot` region which just returns directly to the runner.

v2/lib/TopologySchema.zig
InKryption: Well, it could be, I did this for interface clarity; the compile error for a `Region` that doesn't have these methods defined/has them defined incorrectly will happen in just one place, and close to the top, instead of in potentially multiple locations.

v2/init/main.zig
kprotty: wanst thinking of test-runner case. makes sense

kprotty (approved):

Sobeston (approved):

PR #1460, opened 2026-05-19
fix(octane): sol_prog_secp256k1_sig_structured_diff

hamza-syndica:
## Lineages Result

| Fix | main |
|--------|--------|
| image | image |

## Regular Conformance Result
```
Summary:
	Passed:  43427
	Failed:  1
	Skipped: 0

Detailed test results saved to env/test-outputs/
```

## Added fixed fixtures to `test-fixtures`

https://github.com/Syndica/test-fixtures/commit/7ca4cc98751fd033a97c18a5b21b9f028883cb8e


yewman (commented):
These fixtures are failing because they are legacy fixtures from octane which have features which sig does not currently support on main. This happened as a result of deprecating support for `virtual_address_space_adjustments` while we fix the implementation and run local fuzzing. It is always good to check the status of a feature in features.zon if you are implementing associated logic. If it is unsupported it should not trigger octane crashes (as it should never be a candidate for activation during fuzzing) and needs to be looked into further. 

Now that [feat(runtime): implement AccessViolationHandler for SIMD-0460 (](https://github.com/Syndica/sig/commit/488fc26f9d58fc820f8a188a08c7499281d34b15) is merged, lets recheck these features against main. If they pass, we can close this PR, otherwise rebase on main and apply the appropriate fix

yewman (approved):

ajw221 (approved):

PR #1462, opened 2026-05-19
feat(rpc): sendTransaction - return NodeUnhealthy when unhealthy and skip_preflight == false

dnut:
Preflight checks can never be perfectly reliable because they always represent an older version of the state compared to when the transaction will actually execute. The older the state is, the less reliable the check is. Regardless, clients still use preflight checks as a basic sanity check because they don't expect very much to change between the time the transaction is sent and when it is included in a block.

However, if the validator is caught far behind, and the state is very old, the preflight check provides a weaker signal about whether the transaction will succeed. At a certain level of staleness, clients may prefer to abort the transaction rather than rely on a simulation based on very old data.

This threshold is arbitrary, and it's not specified in the Solana RPC API specification. Anyway, I implemented it the same way as Agave for consistency: If preflight checks are enabled, we check the health first. If the validator is caught behind by 128 slots or more, return an error instead of forwarding the transaction. If preflight checks are skipped, so is the health check.

Here you can see where agave implements this in the same way: https://github.com/anza-xyz/agave/blob/2a61a3ecd417b0515c0b2f322d0128394f20626b/rpc/src/rpc.rs#L3867-L3886

---

I split getHealth into its own struct to simplify the dependencies so it can be used elsewhere. Previously it was in the Ledger struct which doesn't make sense, as health checks have nothing to do with the ledger.

### DIFF
--- a/src/cmd.zig
+++ b/src/cmd.zig
@@ -1899,6 +1899,10 @@ fn validator(
             allocator,
             sig.rpc.hook_contexts.PrioritizationFeeHookContext{ .cache = &prioritization_fee_cache },
         );
+
+        try app_base.rpc_hooks.set(allocator, sig.rpc.hook_contexts.HealthChecker{
+            .commitments = &replay_service_state.replay_state.commitments.?,
+        });
     }
 
     const replay_thread = try replay_service_state.spawnService(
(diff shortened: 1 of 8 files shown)


yewman (approved):
Looks good to me 👍

ajw221 (approved):

PR #1467, opened 2026-05-20
feat(runtime): sbpf-v3 and conformance

### LINKED ISSUE #1492: feat(runtime): sBPF v3

yewman:
# Overview

Makes sig conformant with SBPF v3, gated behind the `enable_sbpf_v3_deployment_and_execution` feature. Brings the VM, ELF loader, and surrounding runtime up to anza-xyz/sbpf v0.14.4 / agave v4.0.0 behavior.

## Strategy

The goal of this PR is conformance, not new functionality: every behavioral change is chosen to match agave/sbpf exactly under the v3 feature gate, while keeping pre-v3 behavior unchanged.

Implementation proceeded in two phases:

1. **Base implementation** of the three v3 SIMDs (0178 static syscalls, 0189 stricter ELF + rodata at vaddr 0, 0377 JMP32 opcodes), the supporting version-gating cleanup (0166 / 0174), and an `elf.zig` refactor aligning sig's parser with agave's two-buffer relocation model.
2. **A series of small fixup commits** driven by conformance gaps surfaced during fuzzing — per-invocation bump allocator, chained self-CPI in `pushInstruction`, ProgramData owner check, v3 Upgrade executable check, etc. Each is a targeted divergence fix rather than a feature.

## Validation

Conformance was established by **local solfuzz fuzzing**, with each lineage run for **20+ hours**:

- `elf_structured` — ELF loader / verifier
- `instr vm` — instruction-level VM execution

The branch passes all current test-vectors (one fixture is marked failing pending a test-vector refresh — it passes against `solfuzz-agave/a7120577cc16`). `zig build test` is green at branch tip.

## Follow-up work

- Improve ELF unit-test coverage in `src/vm/elf.zig` and `src/vm/tests.zig`.
- Improve ELF compilation tooling under `data/test-elfs/` / `scripts/gen_svm_elfs.sh` so v3 fixtures can be regenerated in-tree.
- Upgrade to the latest `solfuzz-agave` and refresh the pinned test-vectors, which will let the currently-skipped fixture be re-enabled.

## References

- sbpf v0.14.4: https://github.com/anza-xyz/sbpf/tree/v0.14.4
- agave v4.0.0: https://github.com/anza-xyz/agave/tree/v4.0.0
- SIMDs 0166, 0174, 0178, 0189, 0377


src/vm/elf.zig
```diff
@@ -221,43 +228,43 @@ const ElfIdent = extern struct {
...
     bytes: []u8,
     config: Config,
     loader: *const SyscallMap,
     version: sbpf.Version,
 ) !Executable {
-    // [agave] https://github.com/anza-xyz/sbpf/blob/v0.12.2/src/elf.rs#L607
-    var parsed = try Elf64.parse(bytes);
+    // [sbpf] https://github.com/anza-xyz/sbpf/blob/v0.20.0/src/elf.rs#L660
+    const unrelocated_bytes = try allocator.dupe(u8, bytes);
+    defer allocator.free(unrelocated_bytes);
```
jbuckmccready, line 248: - `relocate` call mutates the bytes it is given in `parseLenient`, but `parseLenient` didn't make a copy of `bytes` before this PR changes, is it actually required to make a clone of the bytes now? If it is then please add comment for why we're duplicating the bytes, at a glance the anza code is doing things for memory alignment so it's not clear why/what's going on.
- Somewhat related: `bytes` are not mutated in `parseStrict` but passing in `[]u8` and not `const []u8`, this is can be tightened up for clarity I think.

src/vm/elf.zig
```diff
@@ -748,7 +744,7 @@ const Elf64 = struct {
         return text_section;
     }
 
-    /// [agave] https://github.com/anza-xyz/sbpf/blob/v0.12.2/src/elf.rs#L990
+    /// [sbpf] https://github.com/anza-xyz/sbpf/blob/58c47586d70b3d2f1da6c4ff25dd0a3f53a979b6/src/elf.rs#L966-L967
```
jbuckmccready, line 754: Related to other comment on `parseLenient`, if `relocate` is mutating `bytes` and just reading from `self` then it should be `self: *const Elf64` instead of `*Elf64`. There should be more clarity here about what is read vs write using `const` and/or doc comments.

src/vm/elf.zig
```diff
@@ -797,22 +793,21 @@ const Elf64 = struct {
...
                     const sh_dynsym = self.shdrs[dynsymtab];
                     const symbol_table = std.mem.bytesAsSlice(
                         elf.Elf64_Sym,
-                        try safeSlice(bytes, sh_dynsym.sh_offset, sh_dynsym.sh_size),
+                        try safeSlice(self.bytes, sh_dynsym.sh_offset, sh_dynsym.sh_size),
                     );
-
-                    const addr_slice = try safeSlice(bytes, imm_offset, 4);
-                    const ref_addr = std.mem.readInt(u32, addr_slice[0..4], .little);
-                    // Make sure the relocation is referring to a symbol that's in the symbol table.
```
jbuckmccready, line 809: Seems like some unneeded diff here, consts moved higher up than where needed and comment deleted.

src/vm/executable.zig
```diff
@@ -110,46 +110,21 @@ pub const Executable = struct {
     };
 
     /// [agave] https://github.com/anza-xyz/sbpf/blob/v0.13.0/src/verifier.rs#L227
-    pub fn verify(
-        self: *const Executable,
-        loader: *const SyscallMap,
-    ) VerifierError!void {
+    pub fn verify(self: *const Executable) VerifierError!void {
```
jbuckmccready, line 113: This was changed and therefore no longer matches the [agave] doc comment linked code.

src/vm/memory.zig
```diff
@@ -371,23 +378,76 @@ pub const Region = struct {
...
+    ///     A caller-supplied slot-0 region is allowed.
     fn init(
         allocator: std.mem.Allocator,
         regions: []const Region,
         version: sbpf.Version,
         config: exe.Config,
     ) (error{OutOfMemory} || InitError)!AlignedMemoryMap {
-        for (regions, 1..) |reg, index| {
-            if (reg.vm_addr_start >> VIRTUAL_ADDRESS_BITS != index) {
+        var max_slot: u64 = 0;
```
jbuckmccready, line 412: Stale `agave` doc comment link on `AlignedMemoryMap` but there's changes in init and mentions of v3.

src/vm/elf.zig
yewman: 1. Added comment explaining
2. Good catch, changed to []cont u8
c860318740095

src/vm/elf.zig
yewman: Error propagation comes into play here. Moving the `safeSlice` call above the `return error.UnknownSymbol` is relevant for conformance. I didn't add a specific comment detailing this because if we document all changes relating to error propagation fixes comments will become bloated noise. Within the SVM we should always assume fallible statements are ordered deliberately and correctly until proven otherwise by analysis or fuzzing :)

As for the comment removal, I removed it because I don't think it adds additional information beyond what is expressed by the statement itself `if (reloc.r_sym() >= symbol_table.len) return error.UnknownSymbol;`

src/vm/executable.zig
yewman: cd312601851

src/vm/memory.zig
yewman: cd312601851

src/vm/elf.zig
yewman: c8603187

yewman: Generally I think some of this could be refactored for clarity, but better left for a follow up in my opinion

jbuckmccready (approved):

ajw221 (approved):

PR #1470, opened 2026-05-21
refactor(runtime): add interfaces in preparation for sharing with v2

### LINKED ISSUE #1490: refactor(runtime): Decouple transaction processor from v1 types using interfaces
AccountsDB and StatusCache

jbuckmccready:
## Summary

Refactors runtime execution to use interfaces in preparation of creating a shared directory (with runtime code) to be used from both existing V1 and new V2 development. Some other changes are made to decouple dependence of runtime on replay and accounts db.

Overarching goal: create `shared` source code that is imported from both existing v1 and new v2 for the runtime logic (at the boundary of [`loadAndExecuteTransaction`](https://github.com/Syndica/sig/blob/2e3a4c511f5ce819b67a8ca034964cf4695ab0ec/src/runtime/transaction_execution.zig#L180)). Moving to `shared` will be a separate PR on top of these changes.

### Interfaces

These interfaces are just an initial pass to decouple things for moving to `shared` and not indicative of a final form for v2. Code for the interfaces is in `src/runtime/execution_interfaces.zig`.

- Account reading (avoids bringing buffer pool, `Account`, and accounts DB related things into shared)
- Status checking (avoids `StatusCache` and `Ancestors`)
- Epoch stake reading (avoids `sig.core.EpochStakes`)

### Additional Refactoring

- Parts of `src/runtime/spl_token.zig` that depended on ledger types were moved to `src/ledger/token_balances.zig` to avoid needing to bring ledger dependencies into `shared`.
- Move `TransactionError` to core (to be moved into `shared`)
- Move `Account` <-> `AccountSharedData` conversions to dedicated file to not have to bring `Account` into `shared`
- Move `isWriteable` from core `transaction.zig` to replay as it depends on types in replay and not needed in `shared`
- Adjust runtime unit tests to use basic implementations of interfaces to avoid dependencies outside of runtime and core/shared

### Memory leak when transaction fee check fails:

[Alloc is here](https://github.com/Syndica/sig/blob/2e3a4c511f5ce819b67a8ca034964cf4695ab0ec/src/runtime/check_transactions.zig#L126-L132) and missing free/deinit on [early return path here](https://github.com/Syndica/sig/blob/2e3a4c511f5ce819b67a8ca034964cf4695ab0ec/src/runtime/check_transactions.zig#L172-L177).

This bug is fixed in this PR.

### getEpochStake syscall performs lookup in stake accounts but argument is vote account:

[Existing code here](https://github.com/Syndica/sig/blob/2e3a4c511f5ce819b67a8ca034964cf4695ab0ec/src/vm/syscalls/lib.zig#L391-L395) but should be:

```zig
registers.set(.r0, tc.epoch_stakes.stakes.vote_accounts.getDelegatedStake(vote_address.*));
```

Issue created to track this bug: https://github.com/Syndica/sig/issues/1481
This PR just has a TODO comment in place.

## Testing

- Ran on testnet


conformance/src/utils.zig
```diff
@@ -155,8 +157,9 @@ pub fn deinitTransactionContext(
     allocator.destroy(tc.feature_set);
     allocator.destroy(tc.vm_environment);
 
-    tc.epoch_stakes.deinit(allocator);
-    allocator.destroy(tc.epoch_stakes);
+    const epoch_stakes: *const EpochStakes = @ptrCast(@alignCast(tc.epoch_stake_reader.ctx));
```
Sobeston, line 160: This is a pretty strong sign that we're doing something wrong (alignCast especially), what's this for?

Sobeston: Same goes for the other callsites, can we avoid the anyopaque usage?

If we need to support multiple types, we can just go for generics; if we need to support only one type then we can just use that type directly. Not sure if I'm missing something

src/runtime/execution_adapters.zig
```diff
@@ -0,0 +1,189 @@
...
+    pub fn epochStakeReader(self: *const EpochStakeReaderAdapter) EpochStakeReader {
+        return .{
+            .ctx = self.epoch_stakes,
+            .totalStakeFn = totalStake,
+            .stakeForVoteAccountFn = stakeForVoteAccount,
+        };
+    }
+
+    fn totalStake(ctx: *const anyopaque) u64 {
+        const epoch_stakes: *const EpochStakes = @ptrCast(@alignCast(ctx));
```
Sobeston, line 53: What's the point in taking an anyopaque if we effectively assert it's an EpochStates anyway?

Sobeston (changes requested):

src/runtime/execution_adapters.zig
yewman: Maybe I am wrong but is this so that `totalStakeFn` and `stakeForVoteAccountFn` allow a different type to be used by an adapter that may be implemented in v2?

yewman: Perhapds we could remove the anyopaques and use a tagged union if the underlying context type differs between v1 and v2 as an alternative?

yewman (commented):
Looks good to me, once the anyopaue issue is resolved I am happy to approve 👍

conformance/src/utils.zig
jbuckmccready: This is v1 code that won't move to shared or be used by v2, all existing v1 code is adjusted/expects to just use the existing types for simplicity.

Sobeston: Okay I think I misunderstood this section, this seems fine

src/runtime/execution_adapters.zig
Sobeston: Misread some of the context, seems fine

yewman (approved):

src/ledger/token_balances.zig
```diff
@@ -0,0 +1,844 @@
...
+/// - account_reader: Reader to fetch mint accounts for decimals lookup
+///
+/// Returns a slice of TransactionTokenBalance that must be freed by the caller.
+/// Returns null if any mint lookup fails (graceful degradation).
+pub fn resolveTokenBalances(
+    allocator: Allocator,
+    raw_balances: spl_token.RawTokenBalances,
+    mint_decimals_cache: *MintDecimalsCache,
+    comptime AccountReaderType: type,
+    account_reader: AccountReaderType,
```
kprotty, line 32: couldnt this be an anytype?

jbuckmccready: Probably... I copied it exactly as it was, trying to avoid adjusting anything I don't need to. You can see original code deleted in the move in the related commit: https://github.com/Syndica/sig/pull/1470/changes/781f8d4a31c5ec00fcdfdaf592eac0b1b6139103

kprotty: ah, direct move

kprotty (approved):

Sobeston (approved):

PR #1471, opened 2026-05-21
fix(octane): sol_prog_vote_sig_structured_diff

ajw221:
# Intent

- Fix `sol_prog_vote_sig_structured_diff` lineage crashing

# Implementation

- Added signer deduplication to `InstructionInfo.getSigners` to prevent overflow on the `BoundedArray`, matching agave's current `get_signers` logic
- Updated `getVoteStateChecked` to peek the leading tag to return `InvalidAccountData`/`UninitializedAccount` early, preventing bincode from raising a divergent generic error or hitting an unexpected path on truncated zero-tagged vote buffers
- Updated the vote account data zeroing logic to zero in-place, without growing the account data from 3731 to 3762, preserving the original length and leaving `accounts_resize_delta` untouched, matching agave's current logic
- Updated `VoteStateV3.containsSlot` and `VoteStateV4.containsSlot` to use a rust port of `binary_search_by` to ensure unsorted vote lists produce the same membership results as agave
- Added u8 max value clamping `VoteStateV4.commission` to avoid panicking when `inflation_rewards_commission_bps` exceeds 25,500

# Ramifications

N/A

# Tests

- Added unit tests to ensure functionality doesn't regress

### DIFF
--- a/src/runtime/instruction_info.zig
+++ b/src/runtime/instruction_info.zig
@@ -168,21 +168,17 @@ pub const InstructionInfo = struct {
     }
 };
 
-fn pubkeyFromSeed(seed: u8) Pubkey {
-    var data: [Pubkey.SIZE]u8 = @splat(0);
-    data[0] = seed;
-    return .{ .data = data };
-}
-
 test "getSigners collects signer keys and excludes non-signers" {
     const allocator = std.testing.allocator;
+    var prng = std.Random.DefaultPrng.init(std.testing.random_seed);
+    const random = prng.random();
 
     var account_metas = InstructionInfo.AccountMetas{};
     defer account_metas.deinit(allocator);
 
-    const signer_a = pubkeyFromSeed(1);
-    const non_signer = pubkeyFromSeed(2);
-    const signer_b = pubkeyFromSeed(3);
+    const signer_a: Pubkey = .initRandom(random);
+    const non_signer: Pubkey = .initRandom(random);
+    const signer_b: Pubkey = .initRandom(random);
 
     try account_metas.append(allocator, .{
         .pubkey = signer_a,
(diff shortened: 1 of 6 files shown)


src/runtime/instruction_info.zig
```diff
@@ -160,3 +167,105 @@ pub const InstructionInfo = struct {
             return InstructionError.MissingAccount;
     }
 };
+
+fn pubkeyFromSeed(seed: u8) Pubkey {
+    var data: [Pubkey.SIZE]u8 = @splat(0);
+    data[0] = seed;
+    return .{ .data = data };
+}
```
yewman: This is not necessary, just use `Pubkey.initRandom`

src/runtime/instruction_info.zig
```diff
@@ -109,10 +109,17 @@ pub const InstructionInfo = struct {
...
-                signers.appendAssumeCapacity(account_meta.pubkey);
+        outer: for (self.account_metas.items) |account_meta| {
+            if (!account_meta.is_signer) continue;
+            // [agave] get_signers collects into a HashSet. account_metas may hold
+            // up to MAX_INSTR_ACCOUNTS (1094) entries with duplicate pubkeys, so
+            // dedupe to keep the result within the distinct-account bound
+            // (MAX_ACCOUNT_METAS) instead of overflowing the BoundedArray.
+            // [agave] https://github.com/anza-xyz/agave/blob/v4.0/transaction-context/src/instruction.rs#L253
+            for (signers.constSlice()) |existing| {
+                if (existing.equals(&account_meta.pubkey)) continue :outer;
```
yewman: Add a perf todo to come back to this. Or pass an allocator and use a map to collect signers

src/runtime/instruction_info.zig
ajw221: [6e0666c](https://github.com/Syndica/sig/pull/1471/commits/6e0666c35684fb9b1303b4b16ba5460f087f1148)

src/runtime/instruction_info.zig
ajw221: [cc62e6c](https://github.com/Syndica/sig/pull/1471/commits/cc62e6cdb10fa9346d9010d3ec7c206a94f96061)

yewman: Lets remove the bounded array and use an array hash map with to owned slice

ajw221: [85f0f1f](https://github.com/Syndica/sig/pull/1471/commits/85f0f1fcb6c72a0756a649acf868dd38a706fe1d)

yewman (approved):

src/runtime/program/vote/state.zig
```diff
@@ -39,6 +39,32 @@ pub const VOTE_CREDITS_GRACE_SLOTS: u8 = 2;
...
+        const mid = base + half;
+        // Rust: `base = if cmp == Greater { base } else { mid };`
+        // where `cmp` is `items[mid].slot.cmp(&candidate_slot)`.
+        if (items[mid].lockout.slot <= candidate_slot) {
+            base = mid;
+        }
+        size -= half;
+    }
+    return items[base].lockout.slot == candidate_slot;
+}
```
jbuckmccready, line 66: The Rust [`binary_search_by`](https://doc.rust-lang.org/std/primitive.slice.html#method.binary_search_by) result is unspecified if the slice isn't sorted.

> If the slice is not sorted or if the comparator function does not implement an order consistent with the sort order of the underlying slice, the returned result is unspecified and meaningless.

So I'm confused by this doc comment "so on unsorted vote lists (which fuzzer-generated vote
/// accounts may contain) they can disagree on whether a slot is present."

We're creating a specification on the unspecified Rust `std` `binary_search_by` which can change between Rust versions is my understanding of this? Is Agave behavior already coupled to Rust version that is used to compile? If so it seems like we should petition to have Agave decouple conformance from the Rust version used to compile, otherwise it is creating a moving target on a transitive dependency that explicitly states it is unspecified.

jbuckmccready: Polkadot ran into this years ago: https://users.rust-lang.org/t/polkadot-postmortem-kind-of-caused-by-rust-nightly-change/60380

hamza-syndica (approved):

src/runtime/program/vote/state.zig
jbuckmccready: Not a blocker for this PR, just something to investigate and bubble up.

PR #1473, opened 2026-05-21
fix(v2): warm clock path before service seccomp

prestonsn:
- Use libc `clock_gettime` for v2 clock reads, matching Firedancer’s logging clock path.
- Warm REALTIME/MONOTONIC clocks in `sig-init` before services install seccomp.
- Keep `clock_gettime` syscall blocked in services so unexpected fallback still faults (doesn't change seccomp filters)

## Why
`std.os.linux.clock_gettime` uses Zig’s own vDSO lookup state. v2 services enter through custom service entrypoints rather than `std.start`, so that state can be missing and Zig falls back to the raw syscall.  Hence this pr swaps `std.os.linux.clock_gettime` for `std.c.clock_gettime`

libc already initializes its auxv/vDSO state at process startup, and services inherit it upon getting spawned

This follows Firedancer’s model:

- Default logging clock uses `clock_gettime(CLOCK_REALTIME)`: https://github.com/firedancer-io/firedancer/blob/479de315a28b8f2ec662cfc6341699aae08792b8/src/util/log/fd_log.c#L361-L366
- QUIC tile warms the wallclock before sandboxing: https://github.com/firedancer-io/firedancer/blob/479de315a28b8f2ec662cfc6341699aae08792b8/src/disco/quic/fd_quic_tile.c#L485-L494

### DIFF
--- a/v2/init/start_service.zig
+++ b/v2/init/start_service.zig
@@ -41,6 +41,7 @@ comptime {
         if (@typeInfo(Return).error_union.payload != noreturn) @compileError("Invalid return type");
 
         @export(&serviceMain, .{ .name = "svc_main_" ++ @tagName(root.name) });
+        @export(&preSandboxInit, .{ .name = "svc_presandbox_init_" ++ @tagName(root.name) });
         @export(&handleSegfault, .{ .name = "svc_fault_handler_" ++ @tagName(root.name) });
     }
 }
@@ -81,6 +82,13 @@ fn serviceLog(
     tracy.print(fmt_string, args);
 
     nosuspend writer.interface.print(fmt_string, args) catch return;
+}
+
+fn preSandboxInit(params: lib.ipc.ResolvedArgs) callconv(.c) void {
+    // Service libraries do not enter through std.start, so restore the auxv
+    // pointer Zig's vDSO lookup needs before seccomp blocks clock_gettime.
+    std.os.linux.elf_aux_maybe = params.linux_auxv;
+    lib.clock.warmup();
 }
 
 fn serviceMain(params: lib.ipc.ResolvedArgs) callconv(.c) void {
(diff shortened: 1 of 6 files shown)


v2/lib/clock.zig
```diff
@@ -31,13 +38,18 @@ pub fn wallclock(comptime unit: Unit) u64 {
     return clockGetTime(.REALTIME, unit);
 }
 
+/// Initializes libc's clock path before services install seccomp.
+pub fn warmup() void {
+    _ = wallclock(.ns);
+    _ = monotonic(.ns);
```
kprotty: use `std.mem.doNotOptimizeAway` to be extra sure

prestonsn: ye good point: https://github.com/Syndica/sig/pull/1473/commits/531b4ce72b7dd2e2eb5921315feb1c9cca7801fd

thought maybe the optimizer can't see libc fns as pure and hence wont optimize em away

InKryption (dismissed):
I think it would be nice if we could avoid a hard dependency on libc using our own vdso, but can do that easily later.

v2/init/start_service.zig
```diff
@@ -83,6 +84,13 @@ fn serviceLog(
     nosuspend writer.interface.print(fmt_string, args) catch return;
 }
 
+fn preSandboxInit(params: lib.ipc.ResolvedArgs) callconv(.c) void {
+    // Service libraries do not enter through std.start, so restore the auxv
+    // pointer Zig's vDSO lookup needs before seccomp blocks clock_gettime.
+    std.os.linux.elf_aux_maybe = params.linux_auxv;
+    lib.clock.warmup();
+}
```
kprotty: do we need pre-sandbox-init? could this code be moved into `fn serviceMain` down below?

prestonsn: i guess no reason not to. I thought having dedicated init handler was clean in case more stuff in the future. I can remove it if preferred

kprotty: lets remove it until a service needs to do privileged service-specific things. Then it could be brought back & exposed.

prestonsn: aight removed it jit: https://github.com/Syndica/sig/pull/1473/commits/6ca464ed82ca096abc210d0af5d97b792d7f8b79

kprotty (approved):

InKryption (approved):

PR #1480, opened 2026-05-26
fix(rpc): return Agave-conformant BlockNotAvailable error

ajw221:
# Intent

- Update `BlockNotAvailable` rpc error to return an agave-conformant value containing the corresponding slot number

# Implementation

- Updated `BlockNotAvailable` codepaths to attach an agave-conformant JSON-RPC error (code `-32004`, message `"Block not available for slot {slot}"`) via a new thread-local `error_detail` slot that the `Hooks.set` wrapper consumes in place of the generic `mapError` fallback.

# Ramifications

- Currently RPC methods hitting codepaths using `BlockNotAvailable` may not be compatible with solana conformant clients

# Tests

- Created regression tests around new implementation
- Ran at the head of the chain with various RPC methods hitting `BlockNotAvailable`

### DIFF
--- a/src/rpc/hook_contexts/Ledger.zig
+++ b/src/rpc/hook_contexts/Ledger.zig
@@ -47,6 +47,7 @@ pub fn getBlock(
     self: LedgerHookContext,
     arena: Allocator,
     params: GetBlock,
+    error_detail: *?sig.rpc.response.Error,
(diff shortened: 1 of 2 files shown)


dnut (approved):

prestonsn (approved):
LGTM 🫡

PR #1482, opened 2026-05-26
fix(snapshot): use file-size-based bincode allocation limits

ajw221:
# Intent

- Fix snapshot load OOM currently occurring on testnet

# Implementation

- Replaced the hardcoded 2 GiB bincode `allocation_limit` for `Manifest` and `StatusCache` deserialization with a per-file budget of `file_size * 8` (saturating)

# Ramifications

- Currently can't load a snapshot on testnet

# Tests

- Successfully loaded snapshots locally

### DIFF
--- a/src/accountsdb/snapshot/data.zig
+++ b/src/accountsdb/snapshot/data.zig
@@ -515,15 +515,18 @@ pub const Manifest = struct {
         defer allocator.free(contents);
 
         var fbs = std.io.fixedBufferStream(contents);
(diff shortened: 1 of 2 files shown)


dnut (approved):
8x the snapshot size is likely to be way bigger than necessary in almost all cases because the vast majority of the snapshot is accounts. But it's pretty hard to actually predict the upper bound here and we don't want it to break because of a bad limit. I'm fine with just making it huge.

yewman (approved):

PR #1483, opened 2026-05-26
feat(scripts): add shred-stream tool

prestonsn:
Adds a very simple CLI tool to v2 called `shred-stream`. It streams shreds from an agave blockstore ledger over UDP for the purposes of testing replay v2.

- [x] Reads Agave ledgers
- [x] Streams data/code shreds in ledger order without reordering (same order as they are in ledger)
- [x] Supports bounded slot ranges and packet send-rate caps

Also added `.advertise_in_gossip = <bool>` to `shred_network`, which lets configs opt into publishing the configured shred receive port as the node's TVU address in gossip (or opt-out, as is useful to do when trying to test v2).

#### Example usage: 

```bash
zig build shred-stream -Doptimize=ReleaseFast -- --ledger ledger --target  8.8.8.8:8002 --start-slot 410010000 --end-slot 410010200 --rate-hz 5000
```

Where `--rate-hz` is an upper bound (in packets/shreds per second). It's pretty simple and just `std.Thread.sleep`s to stay under the configured rate. In the future, we could offload packet pacing to the kernel with `SO_TXTIME` if we need tighter timings

To get a leader schedule for v2 specific to a ledger: 

```
 solana leader-schedule --url testnet --epoch 961 > schedule.txt
```

Should resolve https://github.com/Syndica/sig/issues/1493

### DIFF
--- a/v2/config/testnet.zig.zon
+++ b/v2/config/testnet.zig.zon
@@ -8,6 +8,7 @@
     },
     .shred_network = .{
         .recv_port = 8002,
+        .advertise_in_gossip = true,
     },
     .telemetry = .{
         .port = 12345,
--- a/v2/config/testnet.zig.zon
+++ b/v2/config/testnet.zig.zon
@@ -5,10 +5,10 @@
     .leader_schedule_file = "schedule.txt",
     .gossip = .{
         .port = 8001,
+        .advertise_tvu_port = true,
     },
     .shred_network = .{
         .recv_port = 8002,
-        .advertise_in_gossip = true,
     },
     .telemetry = .{
         .port = 12345,
--- a/v2/scripts/shred_stream.zig
+++ b/v2/scripts/shred_stream.zig
@@ -133,15 +133,18 @@ pub fn main() !void {
 
     switch (parse_result) {
         .help => printHelp(),
-        .config => |config| try run(gpa, config),
+        .config => |config| {
+            var stdout_buf: [1024]u8 = undefined;
+            var stdout_writer = std.fs.File.stdout().writer(&stdout_buf);
+            const stdout = &stdout_writer.interface;
+
+            try run(gpa, stdout, config);
+            try stdout.flush();
+        },
     }
 }
 
-fn run(allocator: Allocator, config: Config) !void {
-    var stdout_buf: [1024]u8 = undefined;
-    var stdout_writer = std.fs.File.stdout().writer(&stdout_buf);
-    const stdout = &stdout_writer.interface;
-
+fn run(allocator: Allocator, stdout: *std.Io.Writer, config: Config) !void {
     const target = try std.net.Address.parseIpAndPort(config.target);
 
     var blockstore = try AgaveBlockstore.open(allocator, config.ledger);
@@ -256,8 +259,6 @@ fn run(allocator: Allocator, config: Config) !void {
         if (producer_failed.load(.monotonic)) return error.ProducerThreadFailed;
         if (net_thread_failed.load(.monotonic)) return error.NetThreadFailed;
     }
-
-    try stdout.flush();
 }
 
 const AgaveBlockstore = struct {
--- a/v2/scripts/shred_stream.zig
+++ b/v2/scripts/shred_stream.zig
@@ -182,7 +182,9 @@ fn run(allocator: Allocator, config: Config) !void {
             std.posix.SOCK.DGRAM | std.posix.SOCK.CLOEXEC,
             std.posix.IPPROTO.UDP,
         );
-        errdefer std.posix.close(sockfd);
+        // The net thread borrows this fd. Keep ownership here so every path joins
+        // the net thread before closing the fd exactly once.
+        defer std.posix.close(sockfd);
 
         var ring: StreamPacketRing = undefined;
         ring.init();
@@ -195,7 +197,16 @@ fn run(allocator: Allocator, config: Config) !void {
         var producer_progress: ProducerProgress = .{};
         var producer_result: ProducerThreadResult = .{};
 
-        const net_thread = try std.Thread.spawn(
+        var maybe_net_thread: ?std.Thread = null;
+        var maybe_producer_thread: ?std.Thread = null;
+        errdefer {
+            stop.store(true, .release);
+            producer_done.store(true, .release);
+            if (maybe_producer_thread) |thread| thread.join();
+            if (maybe_net_thread) |thread| thread.join();
+        }
+
+        maybe_net_thread = try std.Thread.spawn(
             .{},
             netThreadMain,
             .{
@@ -210,13 +221,8 @@ fn run(allocator: Allocator, config: Config) !void {
                 config.rate_hz,
             },
         );
-        errdefer {
-            stop.store(true, .release);
-            producer_done.store(true, .release);
-            net_thread.join();
-        }
 
-        const producer_thread = try std.Thread.spawn(
+        maybe_producer_thread = try std.Thread.spawn(
             .{},
             producerThreadMain,
             .{
@@ -230,7 +236,7 @@ fn run(allocator: Allocator, config: Config) !void {
             },
         );
 
-        monitorProgress(
+        try monitorProgress(
             stdout,
             &ring,
             &producer_done,
@@ -238,15 +244,12 @@ fn run(allocator: Allocator, config: Config) !void {
             &producer_progress,
             &net_progress,
             config.rate_hz,
-        ) catch |err| {
-            stop.store(true, .release);
-            producer_done.store(true, .release);
-            producer_thread.join();
-            net_thread.join();
-            return err;
-        };
-        producer_thread.join();
-        net_thread.join();
+        );
+
+        maybe_producer_thread.?.join();
+        maybe_producer_thread = null;
+        maybe_net_thread.?.join();
+        maybe_net_thread = null;
 
         try printProducerStats(stdout, producer_result.stats);
         try printNetThreadStats(stdout, net_thread_stats);
@@ -477,6 +480,7 @@ const ProgressSnapshot = struct {
     }
 };
 
+// Adapts the fallible net thread loop to std.Thread.spawn's void entry point.
 fn netThreadMain(
     ring: *StreamPacketRing,
     done: *std.atomic.Value(bool),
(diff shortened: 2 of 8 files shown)


v2/config/testnet.zig.zon
```diff
@@ -8,6 +8,7 @@
     },
     .shred_network = .{
         .recv_port = 8002,
+        .advertise_in_gossip = true,
     },
```
Sobeston, line 12: nit, isn't this config for gossip rather than config for shred-network

v2/scripts/shred_stream.zig
```diff
@@ -0,0 +1,1341 @@
...
+            if (slot < start_slot) return false;
+        }
+        if (self.end_slot) |end_slot| {
+            if (slot > end_slot) return false;
+        }
+        return true;
+    }
+
+    fn pastEndSlot(self: Config, slot: Slot) bool {
+        return if (self.end_slot) |end_slot| slot > end_slot else false;
```
Sobeston: nit, mildly redundant

v2/scripts/shred_stream.zig
```diff
@@ -0,0 +1,1341 @@
...
+        self.data_packets.store(stats.data_packets, .release);
+        self.code_packets.store(stats.code_packets, .release);
+        self.payload_bytes.store(stats.payload_bytes, .release);
+    }
+};
+
+const ProducerThreadResult = struct {
+    stats: ProducerStats = .{},
+    err: ?anyerror = null,
+};
```
Sobeston, line 454: nit: probably shouldn't be using anyerror

v2/scripts/shred_stream.zig
```diff
@@ -0,0 +1,1341 @@
...
+    if (std.fs.cwd().statFile(ledger_path)) |stat| {
+        if (stat.kind == .directory) return try allocator.dupe(u8, ledger_path);
+    } else |_| {}
+
+    std.debug.print("ledger path does not exist or is not a directory: {s}\n", .{ledger_path});
+    return error.InvalidLedgerPath;
+}
+
+const ColumnFamilyNames = struct {
+    allocator: Allocator,
```
Sobeston: Generally I would recommend against "managed" datatypes like this

v2/scripts/shred_stream.zig
```diff
@@ -0,0 +1,1341 @@
...
+    switch (parse_result) {
+        .help => printHelp(),
+        .config => |config| try run(gpa, config),
+    }
+}
+
+fn run(allocator: Allocator, config: Config) !void {
+    var stdout_buf: [1024]u8 = undefined;
+    var stdout_writer = std.fs.File.stdout().writer(&stdout_buf);
+    const stdout = &stdout_writer.interface;
```
InKryption: Just a thought: could move `stdout_writer` into main, and just pass in an `stdout: *std.Io.Writer` parameter; the call then can be:
```zig
var stdout_buf: [1024]u8 = undefined;
var stdout_writer = std.fs.File.stdout().writer(&stdout_buf);
const stdout = &stdout_writer.interface;
try run(gpa, stdout, config);
try stdout.flush();
```
which puts the final flush much closer to the area where the writer begints its lifetime. Slightly easier to validate at a glance.

v2/scripts/shred_stream.zig
```diff
@@ -0,0 +1,1341 @@
...
+    target: std.net.Address,
+    rate_hz: ?f64,
+) !void {
+    defer reader.markUsed();
+    errdefer |err| {
+        net_thread_error.* = err;
+        stats.send_errors += 1;
+        progress.store(stats.*);
+        stop.store(true, .release);
+    }
```
InKryption: Note: `errdefer |err|` is staged for removal (it already is in the latest version of the compiler), so should find a different way to accomplish this. Probably an easy-ish way to do so is to perform this post-error code in the wrapper function in the `catch`, before propagating the error.

v2/scripts/shred_stream.zig
```diff
@@ -0,0 +1,1341 @@
...
+            .produced_packets = produced_packets,
+            .sent_packets = sent_packets,
+            .producer_full_polls = producer_progress.full_polls.load(.acquire),
+            .sender_empty_polls = net_progress.empty_polls.load(.acquire),
+        };
+    }
+};
+
+// Adapts the fallible net thread loop to std.Thread.spawn's void entry point.
+fn netThreadMain(
```
InKryption: To be clear, `std.Thread.spawn` supports `!void` as a return type of a thread, it'll dump a stack trace on error return, much like returning an error from main does.

v2/scripts/shred_stream.zig
```diff
@@ -0,0 +1,1341 @@
...
+
+        stats.empty_polls += 1;
+        if (stats.empty_polls % 1024 == 0) {
+            progress.empty_polls.store(stats.empty_polls, .release);
+        }
+        std.atomic.spinLoopHint();
+    }
+}
+
+fn producerThreadMain(
```
InKryption, line 588: Just from a surface level evaluation, it seems like having this be a separate thread introduces a lot of code complexity. Does the rationale behind the existence of this have to do with trying to model this as if it were a service in some way? If yes, we can maybe leave it as it is for now until we concretize the details on how we'll do that. But otherwise - and correct me if I'm mistaken -, I think the code would benefit from moving this production inline to eliminate a lot of synchronization points.

v2/scripts/shred_stream.zig
```diff
@@ -0,0 +1,1341 @@
...
+        maybe_producer_thread.?.join();
+        maybe_producer_thread = null;
+        maybe_net_thread.?.join();
+        maybe_net_thread = null;
+
+        try printProducerStats(stdout, producer_result.stats);
+        try printNetThreadStats(stdout, net_thread_stats);
+
+        if (producer_result.err) |err| return err;
+        if (net_thread_error) |err| return err;
```
InKryption: Returning the errors of threads like this has little benefit; the error return traces will only have existed on those threads, dumping them in a meaningful way will have had to have taken place on those threads at some point before they exit (easiest way to do so being just returning the error directly from them, as I mentioned could be done in another comment). If you want to indicate they failed, I would turn these `foo_err: ?anyerror`s variables into just `foo_failed: std.atomic.Value(bool)` variables, and then just `if (foo_failed.load(.monotonic)) return error.FooFailed;`.

InKryption (changes requested):
Just a few comments. I appreciate the overall design; will be interesting to see how we can rework and integrate this into custom topology testing.

v2/config/testnet.zig.zon
prestonsn: True, I wanted it next to the port in question of being advertised. Thought it made sense that way, but it is modifying gossip behavior so made the change: https://github.com/Syndica/sig/pull/1483/changes/dd652514ac0d9048be906306e5efa69eac9dffa9

v2/scripts/shred_stream.zig
prestonsn: https://github.com/Syndica/sig/pull/1483/changes/eed7e49d6bc317ba2ee4b6a34fe4839dc3d287d7

v2/scripts/shred_stream.zig
prestonsn: I see that makes sense. I removed the use of `anyerror` and went with the suggested approach for atomic flag for failure reporting instead: https://github.com/Syndica/sig/pull/1483/commits/51eebae30d5718a6f30f9c3d1ff02c50a4947f80

v2/scripts/shred_stream.zig
prestonsn: gotcha, removed here: https://github.com/Syndica/sig/pull/1483/commits/51eebae30d5718a6f30f9c3d1ff02c50a4947f80

v2/scripts/shred_stream.zig
prestonsn: I see, went with your suggested approach and removed `errdefer |err|` here: https://github.com/Syndica/sig/pull/1483/commits/51eebae30d5718a6f30f9c3d1ff02c50a4947f80

v2/scripts/shred_stream.zig
prestonsn: Ahh I see, changes things to allow `std.Thread.spawn` to dump the error/trace if a thread fails as part of these changes: https://github.com/Syndica/sig/pull/1483/commits/51eebae30d5718a6f30f9c3d1ff02c50a4947f80

v2/scripts/shred_stream.zig
prestonsn: This is true. This is just the start of the tool and I initially had it all single threaded. The reason for the producer thread -> net thread is for allowing us to expand and add more complex ways of producing the shreds from the ledger. The idea being that the producer thread eventually manage it's own state, traverse the ledger, handles the generation of the test sequence on its own without having to worry about packet pacing (sending out the shreds at a user-determined max rate), and streaming the packets out as they become ready (selected and correctly sequenced for the test). 

This could totally be done with one thread, and I'll likely move to that in the future once we have more scenarios and we're getting some use out of the tool. Likely a bunch of the packet pacing stuff could be offloaded to the kernel, allowing to just have one thread that reads ledger and timestamps shreds to send in some sequence, handing them off to the kernel for the actual socket sends. 

For a test shred service in the future, we'd likely re-use certain parts of this tool (likely re-home the code into the service).

v2/scripts/shred_stream.zig
prestonsn: fax, also made `AgaveBlockstore` unmanaged: https://github.com/Syndica/sig/pull/1483/commits/37b5d83c8a49d62219a1770d8311db520fa984a7

v2/scripts/shred_stream.zig
prestonsn: Good point, went ahead and moved it: https://github.com/Syndica/sig/pull/1483/commits/dd4cc4e7c9595dd5aaa68af12f010f81d82b5686

Sobeston (approved):

v2/scripts/shred_stream.zig
prestonsn: I should also mention that this whole "rate limiting" thing goes away when we do move this functionality into a `shred_test` service (since we'll have back pressure via the ipc rings, and no longer spraying and praying over udp). So that should do away with the dedicated "send" thread soon.

InKryption (approved):

PR #1484, opened 2026-05-27
feat(v2): stop building tests in the default build step

### LINKED ISSUE #1417: feat(v2): stop building tests in the default build step
We already have `zig build test` and `zig build ci`, we don't need to be building all tests in the default step, it's noisy and slows us down

hamza-syndica:
Solves: #1417 (but adjusted as per new comments in this PR)

```sh
cd v2
rm -rf zig-out

zig build sig

ls zig-out/bin  

- sig-init
```


Sobeston (dismissed):
Nice, works well
just want a +1 from @InKryption before merging

edit: please ignore the copilot review stuff if it appears, miss clicked...

v2/build.zig
```diff
@@ -162,6 +162,11 @@ pub fn build(b: *Build) !void {
             .use_llvm = true,
         });
         const sig_init_out = addExeOutputs(b, sig_init_exe, run_step, artifact_opts, .{});
+        const sig_step = b.step("sig", "Build only the sig binary");
```
InKryption: Would prefer to keep all the top level steps listed out at the top like we've been doing; makes it easy to know all the top level steps just at a glance. Otherwise this looks good to me.

hamza-syndica: Definitely, thanks.

https://github.com/Syndica/sig/pull/1484/commits/3c1b9978f444edec106cac1ec3fb9ed64bb1827d

InKryption (approved):

Sobeston (approved):

PR #1485, opened 2026-05-27
feat(v2): move cryptography to its own library

### LINKED ISSUE #1388: feat(v2): move cryptography to its own library
This would allow us to unconditionally build it with ReleaseFast, which might make testing with even Debug builds on the rest of the project viable.

I've also noticed that sig verify for instances is ~2x faster in ReleaseFast over ReleaseSafe. I expect the same for other cryptography code.

It would also cut down on compile times, as it's very rarely modified and would be easily cached (+ reduces duplicate codegen, and increases codegen parallelism)

hamza-syndica:
Solves: #1388 

As per discussion with @Sobeston: The `reed-solomon` part got a rewrite in replay branch, it's okay to ignore that for now in this PR.


Sobeston (changes requested):
This doesn't do what you expect it to do; it needs to be a static library (i.e. its own compilation unit) for the benefits

v2/crypto/build.zig
```diff
@@ -0,0 +1,73 @@
+const std = @import("std");
+const Build = std.Build;
+
+pub fn build(b: *Build) void {
+    const target = b.standardTargetOptions(.{});
+    // Ignore consumer's optimize choice — always ReleaseFast for crypto.
+    _ = b.standardOptimizeOption(.{});
```
InKryption, line 7: I don't quite understand the point of doing this? If we're going to always use ReleaseFast, it should just not be consumed.

v2/crypto/header.zig
```diff
@@ -0,0 +1,294 @@
...
+    *const fn (*const Signature, *const Pubkey, [*]const u8, usize) callconv(.c) bool,
+    .{ .name = "sig_crypto_sig_verify" },
+);
+
+// -- Types -- //
+
+pub const ed25519 = struct {
+    // ed25519 functions are called internally by Signature.verify,
+    // which lives in the static library. No direct @extern needed here.
+};
```
InKryption, line 46: I think it would make sense to expose and move these extern decls into the `ed25519` namespace. Also: since you're using `@extern`, there's no need for the actual decl name to match the extern symbol identifier. So can rename these as `sig_crypto_sig_verify` -> `sigVerify` and so on.

Side note: we could `export` these symbols using camelCase, there's no inherent reason to use snake_case for functions just because they're linked.

v2/crypto/exports.zig
```diff
@@ -0,0 +1,73 @@
...
+    pubkey: *const Pubkey,
+    msg_ptr: [*]const u8,
+    msg_len: usize,
+) callconv(.c) bool {
+    self.verify(pubkey, msg_ptr[0..msg_len]) catch return false;
+    return true;
+}
+comptime {
+    @export(&sigVerify, .{ .name = "sig_crypto_sig_verify" });
+}
```
InKryption, line 73: It would seem clearer to me to simply mark these all as `export fn`? And as stated in the other comment, linker symbols don't have to be snake case.

v2/crypto/header.zig
```diff
@@ -0,0 +1,294 @@
...
+        var lens: [8]usize = undefined;
+        const n = @min(data.len, 8);
+        for (0..n) |i| {
+            ptrs[i] = data[i].ptr;
+            lens[i] = data[i].len;
+        }
+        var out: Hash = undefined;
+        sig_crypto_hash_init_many(&ptrs, &lens, n, &out);
+        return out;
+    }
```
InKryption, line 76: Doesn't this mean we're just ignoring `data[8..]`?
I think this is indicative of this API not making a huge amount of sense at this level. What we should probably do is expose the Sha state in the header, and have an `update` function equivalent to what we were using before. I imagine there are also other cases where we would want incremental hashing beyond just this.

v2/crypto/exports.zig
```diff
@@ -0,0 +1,73 @@
...
+    count: usize,
+    out: *Hash,
+) callconv(.c) void {
+    var slices: [8][]const u8 = undefined;
+    const n = @min(count, 8);
+    for (0..n) |i| {
+        slices[i] = ptrs[i][0..lens[i]];
+    }
+    out.* = Hash.initMany(slices[0..n]);
+}
```
InKryption, line 36: Re-doubling on my comment about ignoring `data[8..]`.

v2/crypto/build.zig
```diff
@@ -0,0 +1,73 @@
...
+        .root_source_file = b.path("lib.zig"),
+        .target = target,
+        .optimize = .ReleaseFast,
+        .imports = &.{
+            .{ .name = "base58", .module = base58_mod },
+            .{ .name = "binkode", .module = binkode_mod },
+            .{ .name = "tracy", .module = tracy_mod },
+            .{ .name = "build-options", .module = build_options_mod },
+        },
+    });
```
InKryption, line 43: I'm not sure where to put this comment, but I'll put it here:
I was confused by the structure of this library, initially thinking `exports.zig` had a circular definition with `header.zig`. This was because I didn't realize that the types are simply defined twice, once for the impl, and once for the header.

I think it would be preferable to have two directories that cleanly divide these: the implementation source, and the exposed header source.

I also think it would be good if we could avoid needing to define the types twice; I will describe one possible approach to doing this:

1. Have one module, which is just the base `ed25519` library.
2. Define `api.zig` (can bikeshed name), which defines the fundamental APIs in terms of arrays (no reference to any of the type wrappers like `Hash`), all with `callconv(.c)`, and all marked `pub`. Then there should be a `comptime` block which conditionally `export`s them. An easy condition for this would be `@This() == @import("root")`, so that the exports will only be evaluated if we're compiling it directly as the root of a compilation (ie, as a library). We could also give it a dedicated `build-options` module to import if we want to be more explicit.
3. Define the header, which will import the `api` module in order to implement the methods of the exposed types; this is what we expose to consumers.

v2/crypto/build.zig
```diff
@@ -0,0 +1,73 @@
...
+    // Self-reference so sub-files can @import("common") to reach the module root.
+    crypto_mod.addImport("common", crypto_mod);
+
+    // Static library (implementation — @export symbols)
+    const crypto_lib = b.addLibrary(.{
+        .name = "sig-crypto",
+        .linkage = .static,
+        .root_module = crypto_mod,
+    });
+    b.installArtifact(crypto_lib);
```
InKryption, line 53: We don't want to install the library; it can't meaningfully be used without its "header", which is the module, so it should only really be used through that.

It could make sense to install it if it had an associated C header file that could be used by another language for FFI, but we aren't doing that, and probably aren't going to do that in any near future, and certainly not if it remains an exclusive subsidiary of the sig repo.

v2/crypto/build.zig
```diff
@@ -0,0 +1,73 @@
...
+        .target = target,
+        .imports = &.{
+            .{ .name = "base58", .module = base58_mod },
+            .{ .name = "binkode", .module = binkode_mod },
+        },
+    });
+    _ = api_mod;
+
+    // Test step
+    const test_step = b.step("test", "Run crypto unit tests");
```
InKryption, line 67: Would prefer to keep steps defined at the top of the build.zig, as is precedent and convention - maybe we should add that to our style/conventions document.

v2/crypto/build.zig.zon
```diff
@@ -0,0 +1,30 @@
...
+    .paths = .{
+        "build.zig.zon",
+        "build.zig",
+        "lib.zig",
+        "ed25519.zig",
+        "ed25519",
+        "pubkey.zig",
+        "signature.zig",
+        "hash.zig",
+    },
```
InKryption, line 13: This seems like maybe a good reason to organize these into dedicated directories (this also relates to my previous comment about organizing the library into 3 parts).

v2/crypto/lib.zig
```diff
@@ -0,0 +1,25 @@
+const std = @import("std");
+
+comptime {
+    if (!@import("builtin").is_test)
+        std.debug.assert(@import("builtin").mode == .ReleaseFast);
+}
```
InKryption, line 6: Is there any reason for the `is_test` guard here, considering that we always compile the module with ReleaseFast, even for the test?

InKryption (changes requested):
I have more gripes, but the ones I outline here are a bit fundamental and will probably impact any other comments I may have.

PR #1499, opened 2026-05-29
refactor(runtime): move runtime and dependencies into shared

jbuckmccready:
## Summary

- Moves everything that `fn loadAndExecuteTransaction` requires as part of its exposed API to `shared`.
- Updates V1 `src` to import from shared.
- Split up `allocators.zig` to decouple things/avoid bringing more into shared.
- Split `time` directory/module to decouple things.

Other small adjustments (commit: https://github.com/Syndica/sig/pull/1499/changes/97d1b63dad765cb01808904a4b4f4a19ba939efc):
- Adjust `activationStateFromAccount` signature to avoid `Account` dependence.
- Move `ELF_DATA_DIR` constant into `shared/lib.zig` for decoupling.
- Define `DEFAULT_SLOTS_PER_YEAR` constant in `shared/core/rent_collector.zig` for decoupling.
- Small alias/constant use tweaks (e.g., use `sig.net.Packet.DATA_SIZE` instead of `sig.core.Transaction.MAX_BYTES` in `execute.zig` for decoupling)


ajw221 (dismissed):

ajw221 (approved):

prestonsn (approved):
Looking great! 🫡

PR #1500, opened 2026-05-29
fix(replay): skip slot_frozen event when parent is pruned

ajw221:
# Intent

- Prevent websocket event emitting codepath from taking down validator
```
time=2026-05-29T15:26:38.944Z level=info scope=replay_tower message="Can vote on: 411729832 96.2%"
time=2026-05-29T15:26:38.944Z level=info scope=replay_tower message="recorded vote in tower for slot 411729832, new root: 411729800"
time=2026-05-29T15:26:38.944Z level=info scope=fork_choice message="fork_choice: new root set to slot=411729800 hash=6epcrZGTe1wbuwa9NSkShJxVM5gzmeqxZWvusK5pX1PK"
time=2026-05-29T15:26:38.952Z level=info scope=replay message="slot state update from consensus: .{ .root = 411729800, .voted = 411729832, .optimistically_confirmed = null }"
time=2026-05-29T15:26:38.952Z level=info scope=replay message="Commitments" processed=411729832  confirmed=411729831  finalized=411729793  behind=0 
time=2026-05-29T15:26:38.952Z level=info scope=replay message="advanced in 62.455ms"
time=2026-05-29T15:26:38.953Z level=info scope=replay message="tracking new slot: 411729833"
time=2026-05-29T15:26:39.086Z level=info scope=replay message="3 active slots to replay: { 411729818, 411729796, 411729833 }"
time=2026-05-29T15:26:39.086Z level=info scope=replay message="slot is dead: 411729818"
time=2026-05-29T15:26:39.086Z level=info scope=replay message="got 4 entries for slot 411729796"
time=2026-05-29T15:26:39.086Z level=info scope=replay message="got 33 entries for slot 411729833"
time=2026-05-29T15:26:39.098Z level=info scope=replay message="finished replaying slot: 411729796"
time=2026-05-29T15:26:39.107Z level=info scope=replay.freeze message="froze slot 411729796 with hash 2TWjvWdJrFpQWJvyDKxGPKYYdZXTZLDg7cVHFWEYXuZg"
time=2026-05-29T15:26:39.107Z level=info scope=replay message="partially replayed slot: 411729833"
error: MissingParent
time=2026-05-29T15:26:39.107Z level=error scope=service_manager message="replay has errored: error.MissingParent null"
time=2026-05-29T15:26:39.107Z level=error scope=service_manager message="Signaling exit due to 1 errors from replay"
```

# Implementation

- Replaced `error.MissingParent` return with a break to skip event block

# Ramifications

- This isn't currently an issue but at the head of the chain this could occur under the right conditions

# Tests

- Add regression test verifying no crash and no spurious event

### DIFF
--- a/src/replay/service.zig
+++ b/src/replay/service.zig
@@ -835,14 +835,23 @@ fn freezeCompletedSlots(state: *ReplayState, results: []const ReplayResult) !boo
                 }
                 processed_a_slot = true;
 
-                if (state.event_sink) |event_sink| {
+                if (state.event_sink) |event_sink| skip_event: {
                     const slot_state = slot_info.state();
                     const parent_slot = slot_info.constants().parent_slot;
 
-                    const parent_tx_count = if (slot_tracker.get(parent_slot)) |parent_info| blk: {
-                        defer parent_info.release();
-                        break :blk parent_info.state().transaction_count.load(.monotonic);
-                    } else return error.MissingParent;
+                    const parent_info = slot_tracker.get(parent_slot) orelse {
+                        // Parent already pruned because root advanced past this slot mid-replay.
+                        // The slot-frozen event would not have a meaningful tx-count delta, so
+                        // skip it rather than crashing replay.
+                        state.logger.warn().logf(
+                            "skipping slot_frozen event for " ++
+                                "slot {}: parent {} no longer tracked (below consensus root)",
+                            .{ slot, parent_slot },
+                        );
+                        break :skip_event;
+                    };
+                    defer parent_info.release();
+                    const parent_tx_count = parent_info.state().transaction_count.load(.monotonic);
 
                     const num_processed_transactions =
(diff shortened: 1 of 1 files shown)


prestonsn (approved):
I think the changes look fine, and it's an improvement to make behavior under that missing parent edge-case consistent between when running sig with rpc and without.

The changes here do change the rpc api/contract a bit though since now with these changes, we would skip publishing slot_frozen event for the slot, which could mean websocket consumers (I think slotUpdateSubscribe would be affected, but also account/programSubscribe too since they use frozen to know when all accounts should be available for a slot) would miss data for that problem slot. 

I do think a better fix is to somehow snapshot the parent slot's transaction count (and maybe other state?) and store it on the child slot's data when it gets created somehow. But this is an edge-case and likely something we can come back to in RPC v2

yewman (approved):

PR #1505, opened 2026-06-01
feat(v2:topology,ipc,build): Further simplify/mould topology abstraction for testing, add "runner connection", POC integration test

### LINKED ISSUE #1387: feat(v2): integration test runner
Currently we have the startService code, which creates shared memory regions, and launches processes(/threads) for each service, however this is designed for usage with main.

We will soon want a way to test multiple services together, i.e. integration / end-to-end tests.

Ideally such a runner would let you easily specify which services you'd like to run, and let you pass in pre-populated memory regions to act as inputs, and memory regions to act as outputs. 

e.g. pre-fill a ring buffer with packet data, start up the shred-receiver service, wait for completion, and check that it created fec sets as expected.

Each service should, when all its input queues are empty and it has nothing to do, set an atomic bool (is-idle) to true. This would allow the test runner to watch for all services becoming idle and know when to halt.

InKryption:
This change set grew in scope as I encountered deficiencies that would end up as frustrating tech debt if not dealt with on-site.

## POC gossip "black box" test
Most of the groundwork is laid out for this to be done without too much additional effort. The main points of contention are:
* bikeshedding terminology. I've gone with "bbt" or "bb-test" for "blackbox testing".
* designing a streamlined API/process for implementing tests, potentially one which isn't just re-implementing absolutely everything every time?
* how heavily do we rely on building APIs in build.zig to aid the aforementioned goal vs developing "testing primitives" that we just use the build system to orchestrate; there are some current changes of interest in build.zig at the time of writing to kick off some discussion here, which goes down the route of adding some APIs to build.zig.

### Bind, Spawn, & Wait phases extracted and exposed
Where before the fundamental process for spawning and waiting the service topology was just `spawnAndWait(&toSharedRegions(regions_init))`, the code has been fairly heavily reworked to now consist of these steps:
1. Binding phase: call `serviceMap` (name retained from old internal API, but carries pretty different meaning and responsibilities now) to bind and initialize all the regions as represented by the Region union.
2. Spawning phase: call `spawnSandboxed` or `spawnNoSandbox`. This is pretty self explanatory, the basic change here is just spawning the services without immediately waiting, and instead returning all the necessary data to the caller (aka the "runner") to interact with the spawned services.
3. Waiting phase (and everything leading up to it): at the time of writing this part is still maybe a little bit of a WIP; it mostly does just exactly what the `*AndWait` portion of the original APIs did before, just as a standalone function. This may be insufficient, as for testing we will likely want a way to set a timeout in addition to the cooperative cancellation event (described further on in this description).

### Runner Connection
These changes also introduce the concept of a "runner connection", with "runner" here being used to refer to `sig-init`, or whatever main process is spawning the services as described by the selected topology (such as the current POC test runner).

This is represented at the IPC level with `lib.runner.Region`, which acts as the container for `lib.runner.Exit` (formerly `lib.ipc.Exit`), and for `lib.runner.Activity`; part of this is exposed directly to services as `lib.runner.Connection`, which for now just features an "activity" field of type `*lib.runner.Activity.ServiceView`. This is matched on the Runner side with a `lib.runner.Activity.RunnerView`; this "activity" link is used both for the service to signal whether it's idle or active, and for the runner to signal a cancellation event. The specific API constraints are documented in the code.

The connection is an additional parameter now passed to all services. At the time of writing, only a few of the services are actually signalling their activity to the runner.

In general, this "runner connection" parameter should be thought of as the direct line of communication between the runner and the service, and anything else that meets this criteria should probably also be placed here.

### Removed Service Instance Numbers
One of the biggest internal changes here is the removal of "service instance numbers". I concluded that they were specified at the wrong layer of abstraction; our ideal end goal will likely involve scaling the number of services of a given kind at runtime, but these instance numbers were specified at compile time, and at the same time were otherwise entirely inconsequential except for creating certain kinds of indirection. Their removal consolidates and simplifies many concepts in the topology code.

We will want to re-introduce this idea, but in a different form, likely in the form of runtime parameters specified either during binding (`serviceMap`), spawning (the `spawn` functions), or somewhere in between the two, with the idea being that the code can simply "duplicate" the already-created bindings and prepare for a multiplied number of the given services.

### Binding vs Region Terminology Distinction
Roughly, this PR refines a bunch of code to more-or-less more consistently use the terms "binding" or "region" where appropriate. The basic idea is that "region" refers to the view one service has of shared memory regions, and "binding" should refer to the top-down view as understood by the topology/runner. i.e, gossip has a "region" called `.from_net`, which is bound to the "binding" `.net_to_gossip`.

For some code this maybe gets a little fuzzy, specifically around where the regions are actually being materialized, but overall it should make sense.

### Other refactors, changes, additions
* `TopologySchema` -> `topology.Schema`, for consistency with everything else currently being namespaces.
* Added `util.ArrayEnumMap`, like `std.EnumMap`, but it preserves insertion order, which was important for our current `resolveArgs` & `start_service.zig` approach, which relies on regions being *assumed* to be supplied in the order specified by the schema (which currently lacks any real form of validation beyond the likely segfault/general protection exception/illegal instruction crash).
* Replaced usage of dynamic allocation in the topology code with fixed size buffers based on the compile-time schema (aided by the removal of service instance numbers).
* Split out "unbound" types in the generic `Bind` namespace into a separate generic `Unbound` namespace, instead of having a sort of hap-hazzard smattering of them being defined in the global namespace and being defined in `Bind`, and/or sometimes forwarded from the former to the latter.
* Centralized the gossip config region's init procedure for re-use in other runners.
* General QoL additions/changes, e.g. `memfd.RW.mmapStaticSize`.


v2/lib/telemetry/log.zig
```diff
@@ -428,6 +428,44 @@ pub const Filter = struct {
         }
     }
 
+    /// Returns the length that would be written by `parseListAndWriteBinary`.
+    pub fn calcParseListAndWriteBinaryLength(
+        default_log_level: Level,
+        str: []const u8,
+    ) ParseError!u64 {
+        const buffer_len = if (@inComptime()) str.len else 4096;
+        var buffer: [buffer_len]u8 = undefined;
```
Sobeston, line 437: fun

v2/build.zig
```diff
@@ -228,6 +243,30 @@ pub fn build(b: *Build) !void {
...
+    });
+    gossip_bbt_exe.linkLibrary(service_libs.get(.gossip));
+    gossip_bbt_exe.linkLibrary(service_libs.get(.telemetry));
+    gossip_bbt_exe.root_module.addAnonymousImport("schema", .{
+        .root_source_file = topoSchemaSubsetFile(
+            b,
+            "schema.zon",
+            .initMany(&.{ .gossip, .telemetry }),
+        ),
+    });
```
prestonsn, line 331: I think this is a really good poc. The overall orchestration feels clean. 

One _concern_ is with the amount of boilerplate needed to add blackbox tests in the future (likely something you've already considered for improvement). 

Could we generalize the test setup with something like this?

```zig
const BlackBoxTest = struct {
    name: []const u8,
    root_source_file: Build.LazyPath,
    services: std.EnumSet(topo_types.ServiceId),
}

(...)

inline for (black_box_tests) {
    addBlackBoxTest(b, bb_test_step, artifact_opts, runner_imports, service_libs, bbt);
}
```

where `addBlackBoxTest` could do basically what's in the block highlighted above (create the `"bbt-{name}"`, the runner module, link every service in `bbt.services`, do the scheme generation stuff based on `bbt.services`, and setup the exe to `bb-test`)

I'm not sure how this fits into having stuff in `build.zig` vs its own dedicated bbt lib for all the setup code. Maybe the lib is a bit cleaner? but I don't have too much of an opinion on this to be honest. I might need to mess around/add my own test to see how it plays out

v2/tests/gossip/main.zig
```diff
@@ -0,0 +1,174 @@
...
+
+    const activities = spawned.activityViews();
+
+    // wait for gossip and telemetry to go idle
+    blk: while (true) {
+        for (activities) |*view| {
+            if (view.isActive()) continue :blk;
+        }
+        break :blk;
+    }
```
prestonsn, line 85: maybe we want an explicit timeout here? (I can see this being useful across tests in the future, so maybe a timeout helper/struct to consolidate this could be clean)?

v2/lib/runner.zig
```diff
@@ -0,0 +1,244 @@
...
+        consecutive_idles: u32 = 0,
+
+        /// Signal idle; the `threshold` parameter will be used so that the actual idle signal
+        /// will only be sent if this function is called that many times consecutively without
+        /// signalling activity.
+        pub fn signalIdleAfterNCalls(self: *ServiceView, threshold: u32) error{Canceled}!void {
+            if (self.consecutive_idles > threshold) {
+                try Activity.signalIdle(.{
+                    .state = &self.activity.state,
+                });
```
Sobeston: Is this to work around race conditions; isn't signalling "free"

v2/lib/runner.zig
```diff
@@ -0,0 +1,244 @@
...
+            }
+            self.consecutive_idles += 1;
+        }
+
+        pub fn signalActive(self: *ServiceView) error{Canceled}!void {
+            try Activity.signalActive(.{
+                .state = &self.activity.state,
+            });
+            self.consecutive_idles = 0;
+        }
```
Sobeston, line 68: interesting how you've got a threshold for the idle signal but not for the active one

Sobeston (commented):
Looking nice, just got one/two questions

Sobeston (commented):
Looking nice, just got one/two questions

v2/tests/gossip/main.zig
```diff
@@ -0,0 +1,174 @@
...
+            // TODO: read this from identity file in signer service
+            .keypair = self_kp,
+            .turbine_recv_port = 8002,
+        },
+        // gossip -(source)-> snapshot
+        .gossip_source_to_snapshot = {},
+        // net -> gossip
+        .net_to_gossip = .{ .port = gossip_port },
+        .telemetry = .{
+            .port = 12345,
```
prestonsn, line 36: I think another pain point could be the port assignments. If there are multiple blackbox tests running concurrently, maybe it becomes a problem?

It would be nice if, for tests where the port is insignificant, we let the os decide based on what's available (though this seems really hard since it's up to the services to bind)?

v2/tests/gossip/main.zig
```diff
@@ -0,0 +1,174 @@
...
+        const packet = iter.next().?;
+        var fbw: std.Io.Writer = .fixed(&packet.data);
+        try lib.gossip.bincode.write(&fbw, gm);
+        packet.len = @intCast(fbw.end);
+    }
+
+    var spawned = try topology.spawnSandboxed(&service_map);
+
+    // var spawned: topology.NoSandbox = undefined;
+    // try topology.spawnNoSandbox(&spawned, &service_map);
```
prestonsn, line 75: This seems a bit tricky, but is there a clean way we could have bb tests be parametrized by their sandboxing somehow? I can see value in running the same test with and without sandboxing too somehow

prestonsn (commented):
This looks really good! I mostly have some questions + minor nits/things to consider. I'm going to mess around and create my own test to get a better feel

v2/tests/gossip/main.zig
prestonsn: (likely something for the future, feel free to resolve)

v2/lib/runner.zig
InKryption: It's not to work around race conditions so much as it is to "bias" the amount of actual idleness in the service has to be encountered before actually signal that it's really going idle. I.e., one call to this may not be indicative of that it's stalling, but several hundred consecutive calls may.

v2/lib/runner.zig
InKryption: Yeah, active is the "default" state in this exchange. Going idle means you're not receiving input, so you're telling the runner "I'm just sitting here". The moment you get any bit of input to process and work on, you are no longer idle.

v2/tests/gossip/main.zig
prestonsn: I think i misinterpreted actually. I guess `tests/gossip/main.zig` (for example) is simply where all the blackbox tests will be invoked from (i.e same port/other config state gets reused across tests)?

v2/tests/gossip/main.zig
InKryption: Yeah, true, current timeout mechanism only ensures termination after full idleness has been signalled.
A more generalized structure for all of this probably does make sense.

v2/build.zig
prestonsn: We discussed this. I'll create a follow-up pr with the `BlackBoxTest` strat on top of this. Leaning towards keeping things in `build.zig` for now. If we end up expanding test setup/configuration considerably enough, then we can easily move to a bbt lib

v2/tests/gossip/main.zig
prestonsn: Yeah I like that. Can also be in a followup PR if preferred, feel free to resolve at your discretion

v2/tests/gossip/main.zig
prestonsn: My mistake on this. The ports here are insignificant due to `net` service handling all networking (port acquisition)

v2/tests/gossip/main.zig
prestonsn: I think the option that is preferred here is to parametrize by `SandboxMode` at runtime to allow CI or others to run the bbts with/without sandboxing (would be nice if this is done runtime-dynamic to avoid having to compile separate sets of the bbts). But this is future scope, marking this as resolved :saluting_face:

prestonsn (dismissed):

v2/lib/runner.zig
Sobeston: I guess I'm just missing the point, shouldn't a service only signal idle if it 1) currently isn't doing anything 2) all input rings are empty. 

So whether the threshold value is 1 to 1,000,000 seems of little consequence to me. Isn't it fine to report active/idle millions of times per second?

InKryption: It's basically to avoid false positives for tests. There will inevitably be times even in a realistic environment where the service will observe zero inputs, and have nothing it's currently processing, and thus will signal that it's idle. If it were to propagate this signal immediately in a test, the runner would take this as a signal that it's completely finished, and should proceed to cancel it. With this "bias", it just minimizes the likelihood that a false positive will be reported to the runner.

Sobeston: Right, this is what I meant by working around race conditions.

Thought: we could have a bool that's true when these are all true:

1. We're currently doing nothing
2. All input ring buffers are empty
3. All *output* ring buffers are also empty

adding 3) might eliminate this race? I'd have to think about it

Anyway I'm thinking that maybe this threshold value shouldn't be exposed to the caller, hardcoding it might be the way to go. If I'm writing a service I have basically no idea what the "right" value should be because there's no real answer; if we're going to guess we might as well have one guess instead of many

InKryption: The idle signal essentially does(/should) represent the first two conditions being true. The third condition is a bit more tricky, since for testing it would require the runner to first spawn the services, and then drain the outputs while also checking the activity signal. I'll see if I can write something up for that.

prestonsn (dismissed):

v2/lib/runner.zig
InKryption: 7b26e33904c29207c4e368a70e84818904086b5c

I gave it a go, but at least as things are, it ends up being a bit overly-complicated.

Sobeston (dismissed):

prestonsn (dismissed):

prestonsn (approved):

Sobeston (approved):

PR #1509, opened 2026-06-02
fix: use symlinks instead of duplicating fixtures

yewman:
Fixes https://github.com/Syndica/sig/issues/1507

### DIFF
--- a/conformance/scripts/ci-run.sh
+++ b/conformance/scripts/ci-run.sh
@@ -18,7 +18,7 @@ comm -23 \
         -o -type f -name '*.fix' -printf '%P\n' | sort) \
     <(sort scripts/failing.txt) \
     | sed 's_^_env/test-vectors/_' \
-    | xargs -d '\n' cp -t env/split-fixtures/
+    | xargs -d '\n' ln -t env/split-fixtures/
 
 echo Running fixtures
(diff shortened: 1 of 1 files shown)


Sobeston (approved):

conformance/scripts/ci-run.sh
```diff
@@ -18,7 +18,7 @@ comm -23 \
         -o -type f -name '*.fix' -printf '%P\n' | sort) \
     <(sort scripts/failing.txt) \
     | sed 's_^_env/test-vectors/_' \
-    | xargs -d '\n' cp -t env/split-fixtures/
+    | xargs -d '\n' ln -t env/split-fixtures/
```
dnut, line 21: This is creating hard links, not symlinks. I actually tried this exact approach, but I couldn't run the script locally with nix because you can't hard link files owned by nix. It's valuable to be able to run the exact same code locally as what runs in CI. Of course there are other approaches for running this locally besides using nix, so it's not necessarily a deal breaker for hard links.

Anyway, this can work well if it actually uses symlinks. But it requires another small change to the test runner to allow fixtures to be symlinks.

dnut: see here: https://github.com/Syndica/sig/pull/1512

PR #1511, opened 2026-06-02
chore(ci): log disk spaces

hamza-syndica:
log disk spaces on ci to have more insight for "no space" issue


jbuckmccready (dismissed):

jbuckmccready (dismissed):

.github/workflows/check_linux.yml
```diff
@@ -22,6 +22,11 @@ jobs:
   test_kcov_linux:
     runs-on: ubuntu-latest
     steps:
+      - name: "Disk report: initial"
+        run: |
+          df -h /
+          sudo du -sh /usr/local/lib/android /usr/share/dotnet /opt/hostedtoolcache /usr/lib/google-cloud-sdk /opt/ghc 2>/dev/null | sort -rh
```
dnut, line 28: I don't think we should be limiting ourselves to specific folders. The whole idea is that we don't know what's going on, so maximum visibility is useful. Why not just do `sudo du /`? Same idea applies to all the du commands in this pr.

.github/workflows/check_linux.yml
```diff
@@ -33,12 +38,20 @@ jobs:
           version: 0.15.2
           cache-key: kcov-${{ hashFiles('build.zig', 'build.zig.zon', 'v2/build.zig', 'v2/build.zig.zon', 'shared/build.zig', 'shared/build.zig.zon') }}
 
+      - name: "Disk report: after zig setup"
+        run: df -h /
```
dnut: no du command here?

.github/workflows/check_linux.yml
hamza-syndica: the `sudo du /` will take so much time and so much logs, I think it's not ideal at all imho. But, I am open to hear your thoughts if you think otherwise.

.github/workflows/check_linux.yml
hamza-syndica: https://github.com/Syndica/sig/pull/1511/commits/1a668d8a02929ce2de1382eaf15277598b2a16c4

jbuckmccready (approved):

PR #1515, opened 2026-06-02
feat(conformance): upgrade to agave-v4.1.0-beta.1

ajw221:
# Intent

- Upgrade conformance to `agave-v4.1.0-beta.1`
- Address all fixture failures caused by upgrade 

# Implementation

- Removed `log` field setting from `createSyscallEffect` to match the latest `protosol` functionality where `log` is completely removed as a field
- Hoisted pubkey signer check before any nonce state mutation in `withdrawNonceAccount`, matching agave's check-signer-then-set-state ordering
- Removed `increase_cpi_account_info_limit`, `increase_cpi_account_info_limit` and `provide_instruction_data_offset_in_vm_r2` feature gates in favor of now hardcoding them to match v4.1 logic
- Bumped conformance pins to `agave v4.1.0-beta.1` and `protosol 8.2.0`, refreshing vendored fixtures and the failing-fixtures list accordingly
- Returned `InvalidAccountData` for variant `0` when deserializing V4 vote state, matching agave's guard against the uninitialized discriminant
- Cleared modified account data on CU exhaustion when running under direct mapping in the conformance harness, so failed executions don't leak partial writes into the resulting effects
- Always charge the CPI translation cost in `translateInstruction` regardless of early-exit paths, matching agave's unconditional metering
- Installed an access-violation handler in `vm_syscall` so memory faults are surfaced as syscall errors instead of aborting the conformance run

# Ramifications

- Current fixture failures are mainly being caused by conformance-related updates

# Tests

- Added debug logging in the conformance harness for fixtures that exercise reverted or unsupported feature gates, making triage of skipped/diverging fixtures easier
- Added regression tests covering the nonce-withdraw signer ordering and the V4 vote-state variant-0 / resize error guards
- Ran to the tip of the chain on testnet successfully
- Ran fuzzing multiple times and compared results to main to ensure no functionality regressed:
```
  REGRESSIONS:  0
  preexisting:  363759   (failures present on main too)
  noise:        582   (flaky + skipped + crash-replay errors)
  inputs:       364341   (divergences + crashes triaged)

  Target                    Div (br)  Div (base)  Crash (br)  Crash (base)  Regressions
  -----------------------  ---------  ---------  ---------  ---------  -----------
  programs.vote               176440     176440         48         48            0
  runtime.logs                     3          3          0          0            0
  txn                         187220     187220         44         44            0
  vm.syscalls.cpi                  4          4          0          0            0
  -----------------------  ---------  ---------  ---------  ---------  -----------
  TOTAL                       363667     363667         92         92            0
```


yewman (approved):
Looks good to me 👍

src/core/features.zig
```diff
@@ -49,6 +33,24 @@ pub const ZonInfo = struct {
     }
 };
 
+pub const Status = union(enum) {
```
InKryption, line 36: This isn't related to anything you've done, so can mark this as resolved or solve it at your discretion, but it's a bit strange that this is a union, when it could just be an enum.

src/core/features.zig
```diff
@@ -102,6 +104,16 @@ pub const pubkey_map: std.EnumArray(Feature, Pubkey) = blk: {
...
+/// Maps each `Feature` to its corresponding `Status`.
+pub const status_map: std.EnumArray(Feature, Status) = blk: {
+    @setEvalBranchQuota(features.len * 1000);
+    var s: std.enums.EnumFieldStruct(Feature, Status, null) = undefined;
+    for (@typeInfo(Feature).@"enum".fields, features) |field, feature| {
+        @field(s, field.name) = feature.status;
+    }
+    break :blk .init(s);
+};
+
```
InKryption, line 116: Might make sense to instead unify this with `pubkey_map`. Like:
```zig
Pub const Info = struct {
    pubkey: Pubkey,
    status: Status,
};

/// Maps each `Feature` to its corresponding `Status`.
pub const info_map: std.EnumArray(Feature, Info) = blk: {
    @setEvalBranchQuota(features.len * 1000);
    var s: std.enums.EnumFieldStruct(Feature, Info, null) = undefined;
    for (@typeInfo(Feature).@"enum".fields, features) |field, feature| {
        @field(s, field.name) = .{
            .pubkey = .parse(feature.pubkey),
            .status = feature.status,
        };
    }
    break :blk .init(s);
};
```
(not sure if there'd be any value in stuffing any of the other fields in `Info`).

InKryption (approved):
Just a couple of comments, everything else LGTM.

prestonsn (approved):

PR #1547, opened 2026-06-04
refactor(runtime): hardcode remove_accounts_delta_hash

hamza-syndica:
Hardcode `remove_accounts_delta_hash` — confirmed hardcoded in both Agave and Firedancer.

The accounts delta hash is no longer computed. The bank hash now always skips the delta merkle hash step. Removed dead code in freeze.zig and conformance/txn_execute.zig.

- **Agave**: [bb5a6e773d](https://github.com/anza-xyz/agave/commit/bb5a6e773d5f41388a962c5c4f96f5f2ef2209d0) — Remove SVM dependency on agave-feature-set (#5841)
- **Firedancer**: [a75575a412](https://github.com/firedancer-io/firedancer/commit/a75575a412b27869db72dc066d110967bc222d4a) — flamenco, features: clean up feature map (#8095)

Closes #1520


src/replay/freeze.zig
```diff
@@ -714,7 +709,7 @@ test "freezeSlot: trivial e2e merkle hash test" {
     distributed.deinit(allocator);
 
     try std.testing.expectEqual(
-        Hash.parse("8C4gpDhMz9RfajteNCf9nFb5pyj3SkFcpTs6uXAzYKoF"),
+        Hash.parse("9qWF7q5w5vxC3GX7g12VAFiLQ1XnFWU4qVEnNSDWoQxF"),
```
yewman: If this was testing the inactive path maybe just remove this test rather than changing the expectation, same for future. No need to keep tests which duplicate activated path tests

hamza-syndica: Shouldn't have done the force push anyway: https://github.com/Syndica/sig/pull/1547/commits/4f01d7193187c06c33d8f3f77b0b399358ab3d00

yewman (dismissed):

conformance/src/txn_execute.zig
ajw221, line 1172: this was the only location in which `deltaMerkleHash` was being called excluding tests, I think it's safe to say we can remove this function and its tests entirely?

ajw221 (changes requested):

conformance/src/txn_execute.zig
hamza-syndica: https://github.com/Syndica/sig/pull/1547/commits/ebd7ba324ec9d05aa770fe81bea357fdf6edafe7

yewman (approved):

ajw221 (approved):

PR #1559, opened 2026-06-05
feat(runtime): implement SIMD-0449 direct_account_pointers_in_program_input

yewman:
## Intent

- Implement [SIMD-0449](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0449-direct-account-pointers-in-program-input.md)
- After the program id in the ABIv1 program input region, the serializer emits padding zeros up to `BPF_ALIGN_OF_U128` followed by one little-endian `u64` per instruction account holding the `vm_addr` of that account's serialized record (the `NON_DUP_MARKER` byte).
- Duplicates reuse the original record's `vm_addr`, matching `serialize_parameters_for_abiv1`.

## Implementation

- `SerializedAccountMeta` gains a `vm_addr` field capturing the address of the account record's non-dup marker. Both `serializeParametersUnaligned` and `serializeParametersAligned` populate it; the field is also populated for duplicate metas via the existing clone path.
- `serializeParametersAligned` reserves `align_forward(size, BPF_ALIGN_OF_U128) - size` padding bytes plus `accounts.len * @sizeOf(u64)` at the end of the buffer when the flag is on, then writes the pointer array after the program id.
- ABIv0 (loader-v1) ignores the flag, matching agave's `serialize_parameters_for_abiv0`.
- Flag is threaded through `bpf_loader/execute.zig` and the two conformance entry points (`vm_interp.zig`, `vm_syscall.zig`); existing test/CPI call sites pass `false`.

## Gating

The trailing pointer array is only emitted when **all** of the following hold:
- `direct_account_pointers_in_program_input` (SIMD-0449) active
- the program is loaded by an ABIv1 loader (loader-v2 / loader-v3); ABIv0 is unconditionally skipped

## References

- SIMD: [0449-direct-account-pointers-in-program-input.md](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0449-direct-account-pointers-in-program-input.md)
- agave: `programs/bpf_loader/src/serialization.rs` (`serialize_parameters_for_abiv1`)
- firedancer: `src/flamenco/runtime/program/fd_bpf_loader_serialization.c`

## Tests

New unit test in serialize.zig — `direct_account_pointers_in_program_input emits trailing pointer array` — covering:

- ABIv1 with `virtual_address_space_adjustments` off and on (`direct_mapping = false`)
- ABIv0 (loader-v1) does **not** emit the array even when the flag is on
- Duplicate account (A, B, A) reuses the original meta's `vm_addr`
- Trailing array is `BPF_ALIGN_OF_U128`-aligned relative to `INPUT_START`
- Each emitted pointer (under vasa=false) lands on a `NON_DUP_MARKER` byte

Existing `serializeParameters` tests updated to pass `false` for the new parameter.

## Follow-ups

None.

### DIFF
--- a/conformance/src/vm_interp.zig
+++ b/conformance/src/vm_interp.zig
@@ -120,12 +120,17 @@ fn executeVmTest(
         .mask_out_rent_epoch_in_vm_serialization,
         slot,
     );
+    const direct_account_pointers_in_program_input = tc.feature_set.active(
+        .direct_account_pointers_in_program_input,
+        slot,
+    );
     var serialized = try serialize.serializeParameters(
         allocator,
         &ic,
         direct_mapping,
         virtual_address_space_adjustments,
         mask_out_rent_epoch_in_vm_serialization,
+        direct_account_pointers_in_program_input,
     );
     defer serialized.deinit(allocator);
     tc.serialized_accounts = serialized.account_metas;
--- a/shared/runtime/program/bpf/serialize.zig
+++ b/shared/runtime/program/bpf/serialize.zig
@@ -38,6 +38,10 @@ pub const SerializedAccount = union(enum) {
 
 /// [agave] https://github.com/anza-xyz/agave/blob/108fcb4ff0f3cb2e7739ca163e6ead04e377e567/program-runtime/src/invoke_context.rs#L182
 pub const SerializedAccountMeta = struct {
+    /// Address of the first byte of the serialized account record (the
+    /// `NON_DUP_MARKER`/duplicate-marker byte). Used by SIMD-0449 to emit
+    /// a trailing pointer array in the program input region.
+    vm_addr: u64,
     original_data_len: u64,
     vm_data_addr: u64,
     vm_key_addr: u64,
@@ -248,6 +252,7 @@ pub fn serializeParameters(
     direct_mapping: bool,
     virtual_address_space_adjustments: bool,
     mask_out_rent_epoch_in_vm_serialization: bool,
+    direct_account_pointers_in_program_input: bool,
 ) (error{OutOfMemory} || InstructionError)!SerializeReturn {
     if (ic.ixn_info.account_metas.items.len > InstructionInfo.MAX_ACCOUNT_METAS - 1) {
         return error.MaxAccountsExceeded;
@@ -304,6 +309,7 @@ pub fn serializeParameters(
             direct_mapping,
             virtual_address_space_adjustments,
             mask_out_rent_epoch_in_vm_serialization,
+            direct_account_pointers_in_program_input,
         );
 }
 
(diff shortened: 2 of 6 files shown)


jbuckmccready (dismissed):

conformance/src/vm_interp.zig
```diff
@@ -119,12 +119,17 @@ fn executeVmTest(
...
+        .direct_account_pointers_in_program_input,
+        slot,
+    );
     var serialized = try serialize.serializeParameters(
         allocator,
         &ic,
         direct_mapping,
         virtual_address_space_adjustments,
         mask_out_rent_epoch_in_vm_serialization,
+        direct_account_pointers_in_program_input,
```
ajw221, line 133: can we refactor the parameters for this function to just take in a `FeatureSet` and `Slot` and call these inside `serializeParameters`?

not sure if this is relevant to this PR, but thought it could be worth doing while we're here

shared/runtime/program/bpf/serialize.zig
```diff
@@ -304,6 +311,7 @@ pub fn serializeParameters(
             direct_mapping,
             virtual_address_space_adjustments,
             mask_out_rent_epoch_in_vm_serialization,
+            direct_account_pointers_in_program_input,
```
ajw221, line 312: same as question above

shared/runtime/program/bpf/serialize.zig
```diff
@@ -286,6 +291,8 @@ pub fn serializeParameters(
     }
 
     return if (is_loader_v1)
+        // SIMD-0449 only applies to ABIv1 (loader-v2/v3); ABIv0 path ignores
+        // the flag, matching agave's serialize_parameters_for_abiv0.
```
ajw221: would this be good to add as an additional doc comment on `serializeParametersUnaligned` as well?

I guess it makes sense here since this is the only place `serializeParametersUnaligned` is being called, feel free to ignore and resolve this if you want

conformance/src/vm_interp.zig
yewman: I think it is better to keep it as is, or add a `gates` parameter struct if anything. It keeps the dependencies for serialisation more explicit over passing the full feature set and slot as well

shared/runtime/program/bpf/serialize.zig
yewman: same as above

shared/runtime/program/bpf/serialize.zig
yewman: 7d34983ea572f8
Unnecessary comment, removed with some others as well

ajw221 (approved):

jbuckmccready (approved):

PR #1562, opened 2026-06-05
fix(octane): sol_txn_sig_structured_diff

ajw221:
# Intent

- Fix `agave-v4.1.0-beta.1` octane mismatches for the `sol_txn_sig_structured_diff` lineage

# Implementation

N/A

# Ramifications

N/A

# Tests

N/A

### DIFF
--- a/conformance/src/txn_execute.zig
+++ b/conformance/src/txn_execute.zig
@@ -313,52 +313,15 @@ fn executeTxnContext(
             //     @panic("background account hasher not implemented");
             // }
 
-            const account_reader_for_ancestors = account_store.reader().forSlot(&ancestors);
-            // Add builtin programs
-            for (builtin_programs.BUILTINS) |builtin_program| {
-                // If the feature id is not null, and the builtin program is not migrated, add
-                // to the builtin accounts map. If the builtin program has been migrated it will
-                // have an entry in accounts db with owner bpf_loader.v3.ID (i.e. it is now a BPF program).
-                // For fuzzing purposes, accounts db is currently empty so we do not need to check if
-                // the builtin program is migrated or not.
-                const builtin_is_bpf_program: bool = blk: {
-                    const account = try account_reader_for_ancestors.get(
-                        allocator,
-                        builtin_program.program_id,
-                    ) orelse break :blk false;
-                    defer account.deinit(allocator);
-                    break :blk account.owner.equals(&program.bpf_loader.v3.ID);
-                };
-
-                if (builtin_program.enable_feature_id != null or builtin_is_bpf_program) continue;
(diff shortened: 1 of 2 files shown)


yewman (dismissed):

hamza-syndica (approved):

conformance/src/txn_execute.zig
```diff
@@ -313,52 +312,25 @@ fn executeTxnContext(
...
+            // firedancer-io/agave, not upstream anza-xyz/agave), which only runs
+            // `apply_activated_features()`, populating the program cache via
+            // `add_active_builtin_programs()`, and deliberately skips
+            // `add_builtin_program_accounts()` (the latter is reached only from
+            // `compute_and_apply_genesis_features()`, which this path bypasses).
+            // Builtin and precompile program accounts are therefore not seeded
+            // into accounts-db here: each fixture is expected to carry any program
+            // account it invokes, and missing ones must surface as
+            // `TransactionError::ProgramAccountNotFound` during account loading
+            // rather than executing against a synthetic native-loader stub.
```
prestonsn, line 327: From what I understand, It feels like this comment isn't entirely accurate since our `executeTxnContext()` still calls `applyFeatureActivations()` (with `allow_new_activations` set to `false`):

https://github.com/Syndica/sig/blob/1688e4ae1ee46731f76e400445f4afa85b3aece7/conformance/src/txn_execute.zig#L288-L294

`applyuFeatureActivations()` then calls `applyBuiltinProgramFeatureTransitions`, which has this logic here: 

https://github.com/Syndica/sig/blob/1688e4ae1ee46731f76e400445f4afa85b3aece7/conformance/src/bank_methods.zig#L109-L129

But since `allow_new_activations` is false, `should_enable_on_transition` falls back to that else logic: `feature_set.active(enable_feature_id, slot);` (which could fall through to the account store population code below).

I think it's possible for a feature gated builtin whose feature is alredy active at the fixture slot to end up unintentionally populating `account_store` right? (despite comment stating "program accoutns are therefore not seeded ...")

Maybe this isn't possible at all given how fixtures are designed/run, but it is a bit inconsistent

prestonsn (changes requested):
Looks good, just one potential inconsistency to address

conformance/src/txn_execute.zig
prestonsn: Tracked, and we'll fix this in the future. Thinking we can make things 1:1 with firedancer-io/agave's logic here (totally separate out program cache registration from accountsdb seeding instead of how both are consolidated into the same fn in Sig).

prestonsn (approved):

PR #1566, opened 2026-06-07
refactor(runtime): hardcode abort_on_invalid_curve

dymchenkko:
Hardcode the abort_on_invalid_curve feature flag (SIMD-0137).

The flag is permanently active on mainnet. invalidError now always
returns SyscallError.InvalidAttribute for an unrecognised curve ID 
the old soft-error path that wrote 1 to r0 is removed. Updated
two tests and changed the feature status from hardcoded_for_fuzzing
to hardcoded.

Closes #1529

### DIFF
--- a/shared/vm/syscalls/ecc.zig
+++ b/shared/vm/syscalls/ecc.zig
@@ -48,13 +48,8 @@ pub const GroupOp = enum(u64) {
     }
 };
 
-fn invalidError(tc: *const TransactionContext, registers: *RegisterMap) !void {
-    if (tc.feature_set.active(.abort_on_invalid_curve, tc.slot)) {
-        return SyscallError.InvalidAttribute;
-    } else {
-        registers.set(.r0, 1);
-        return;
-    }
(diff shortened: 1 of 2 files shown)


shared/vm/syscalls/ecc.zig
```diff
@@ -48,13 +48,8 @@ pub const GroupOp = enum(u64) {
...
 
-fn invalidError(tc: *const TransactionContext, registers: *RegisterMap) !void {
-    if (tc.feature_set.active(.abort_on_invalid_curve, tc.slot)) {
-        return SyscallError.InvalidAttribute;
-    } else {
-        registers.set(.r0, 1);
-        return;
-    }
+fn invalidError(_: *const TransactionContext, _: *RegisterMap) !void {
+    return SyscallError.InvalidAttribute;
```
dnut, line 52: Is there a reason to keep this function? Why not just inline InvalidAttribute wherever this function is used?

dnut (commented):
Thanks for your contribution. Looks good, but I think I would take the simplification a step further

PR #1567, opened 2026-06-08
refactor(v2:topology,linux): Memfd & spawned state unifications

InKryption:
Unifies `linux.memfd.RW` & `linux.memfd.RO` into `linux.Memfd`; we only need the ability to open a temporary read-only file handle that's then memory-mapped with read-only permissions on the read-write version.

Unifies `Sandboxed` & `NoSandbox` into `Spawned`. Originally started as an attempt to unify these at the `clone3` level, but that grew far too complex, so the behavior simply bifurcates at the API level.

### DIFF
--- a/v2/lib/linux.zig
+++ b/v2/lib/linux.zig
@@ -1,5 +1,10 @@
 const std = @import("std");
 
+const linux = std.os.linux;
+const E = linux.E;
+const e = E.init;
+const page_size_min = std.heap.page_size_min;
+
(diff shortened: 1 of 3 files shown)


Sobeston (dismissed):

prestonsn (dismissed):

prestonsn (approved):
LGTM :saluting_face:

Sobeston (approved):

PR #1578, opened 2026-06-08
fix(ci): proxy_workaround.sh does not fetch all dependencies

dnut:
zig's dependency fetcher frequently fails to download dependencies in CI. We have `proxy_workaround.sh` as a workaround for this. It pre-fetches dependencies into the zig cache using `wget` before `zig build` runs.

The problem is that `proxy_workaround.sh` doesn't actually download all the dependencies, so we still rely on zig at some level, and likewise we still have some failing CI runs where zig fails to fetch.

Ultimately zig's http client is probably to blame and will need to be fixed in the standard library. But this will take many months or years before it goes live in a version of zig that we use to build sig, and we need a solution right now.

The immediate solution was to rewrite `proxy_workaround.sh` from scratch. It had many problems and the easiest solution was to just rewrite it. I named the new script `fetch-zig-deps.py`. Rather than doing a bunch of dirty hacks that only work half the time, we now have proper zon parsing, proper git cloning, etc.

Now this one script successfully fetches all the dependencies needed to build every build.zig in the repository.

## Verification

1. Delete your zig cache
2. Run the dependency-fetching script
3. Disconnect your network
4. Run `zig build` while offline

With the old `proxy_workaround.sh`, this does not work. With the new `fetch-zig-deps.py`, it does work.

## Details

`proxy_workaround.sh` makes a best-effort attempt, but it fails to download some dependencies.

First of all there are several issues apparent from `proxy_workaround.sh`'s stdout.

The script fails to download certain dependencies because the url does not match the expected pattern:
```
>>> Ignored https://github.com/wolfpld/tracy/archive/refs/tags/v0.13.0.zip, unable to resolve it!
>>> Ignored https://www.sqlite.org/2025/sqlite-amalgamation-3490200.zip, unable to resolve it!
>>> Ignored https://github.com/libarchive/bzip2/archive/refs/tags/bzip2-1.0.8.zip, unable to resolve it!
```

The script does not properly format the hashes for some of the transitive dependencies, specifying "N-V-" instead of the correct package name:

```
>> hash of https://github.com/facebook/zstd#794ea1b0afca0f020f4e57b6732332231fb23c70:
        N-V-__8AAKbffQDMlwdeMxqHyyAcA6AYoVIjJdaqhEYZpGFG
>> hash of https://github.com/Syndica/rocksdb#3793e721c7795b338d9d7808544b75db6bde3548:
        N-V-__8AAAKhcQL5PtDzqfPNCjk1LJCK8svO8KtC-f_3bj6z
>> hash of https://github.com/google/snappy#25e52c58fbf83ee40f4c9284f757f777f691f76f:
        N-V-__8AADcwMwCa__yeHfNDwYl4KEDbkj-VwAqY844Nblr8
>> hash of https://github.com/google/boringssl#535cc391915c37912ff9b3edb0cff9b3c5c77db0:
        N-V-__8AADGOTAg2qy2JMjVSleelXVWZhf9UTIjkuioQsv9-
>> hash of https://github.com/madler/zlib/archive/refs/tags/v1.3.1.tar.gz:
        N-V-__8AAB0eQwD-0MdOEBmz7intriBReIsIDNlukNVoNu6o
>> hash of https://github.com/bitcoin-core/secp256k1#e3a885d42a7800c1ccebad94ad1e2b82c4df5c65:
        N-V-__8AAGXcSgBY6xaM21Q70BszQyjyZBlP9Fgfj-UF9hG0
```

When running zig build offline, we immediately see some packages fail to download, indicating that `proxy_workaround.sh` has not successfully cached them. 

```sh
> zig build
sig/build.zig.zon:57:20: error: unable to connect to server: TemporaryNameServerFailure
            .url = "https://www.sqlite.org/2025/sqlite-amalgamation-3490200.zip",
                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
sig/build.zig.zon:65:20: error: unable to connect to server: TemporaryNameServerFailure
            .url = "https://github.com/libarchive/bzip2/archive/refs/tags/bzip2-1.0.8.zip",
                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
.cache/zig/p/zig_tracy-0.13.0-4TLLR1ZkAAAKZh0Gimq4P-Hyyd3b38XfFecg00aAqIaR/build.zig.zon:15:20: error: unable to connect to server: TemporaryNameServerFailure
            .url = "https://github.com/wolfpld/tracy/archive/refs/tags/v0.13.0.zip",
                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
.cache/zig/p/zstd-0.0.1-8QsqCTmhAAA-9ArKPxqvOt77ULmEr3xma_bbpeWm-b2u/build.zig.zon:7:20: error: unable to discover remote git server capabilities: TemporaryNameServerFailure
            .url = "git+https://github.com/facebook/zstd#794ea1b0afca0f020f4e57b6732332231fb23c70",
                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

There are layered issues at play here. Not only were transitive dependencies getting the wrong hash prefix, but they also had the wrong hash calculated. For example, `git+` dependencies were downloaded as tarballs from github and hashed as a whole. Zig's git fetcher chooses a different subset of files from the repo compared to github's tarballs. When their hash is not based on build.zig.zon, (such as facebook/zstd), it hashes bsaed on all files, which results in an incorrect hash being calculated


hamza-syndica (dismissed):
tested locally without network as per the steps, works nice.

.github/workflows/check_macos.yml
ajw221, line 1: wget is no longer being used so we should be able to remove `brew install wget`

ajw221 (dismissed):

.github/workflows/check_macos.yml
dnut: [a2167c5](https://github.com/Syndica/sig/pull/1578/commits/a2167c5bdebf811318c46d907412359f02d017dc)

jbuckmccready (dismissed):

jbuckmccready (approved):

hamza-syndica (approved):

PR #1586, opened 2026-06-10
refactor(runtime): use lamports per signature constant

### LINKED ISSUE #1133: Simplify fee handling
### Problem

Remove redundant data structures relating to transaction fees.

### Proposal

Right now there are a number of data structures which specify lamports_per_signature which are for the most part unused, leading to confusion and bug prone code. For example FeeRateGovernor is passed around throughout the runtime, but is not actually responsible for determining transaction fees. Instead transaction fees are effectively 'hardcoded' to 5000 lamports per signature. Agave uses the lamports per signature specified in the FeeStructure instead of the FeeRateGovernor. Rather than trying the implement solutions for functionality which 'may' be required in the future, we should simplify Sig by removing any fee paying code which is not explicitly required.

This issue was brought to attention in fixing the following bug: https://github.com/Syndica/sig/pull/1126/files#top

### Alternatives

_No response_

### Additional Context

_No response_

leviathanbeak:
Closes #1133 (or at least part of it)

I just focused on making `LAMPORTS_PER_SIGNATURE` constant and using it in places where it made sense.

Not sure does the original issue want more than that.

### DIFF
--- a/conformance/src/txn_execute.zig
+++ b/conformance/src/txn_execute.zig
@@ -695,7 +695,6 @@ fn executeTxnContext(
         // juggles the many different versions of lamports_per_signature.
         .next_lamports_per_signature = last_lamports_per_signature,
         .last_lamports_per_signature = last_lamports_per_signature,
-        .lamports_per_signature = 5000,
     };
 
     var failed_accounts = sig.runtime.account_loader.LoadedTransactionAccounts.Accounts{};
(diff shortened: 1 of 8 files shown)


ajw221 (changes requested):
The removal of `lamports_per_signature` as an field/argument may not be something we want to do given the current solana spec.

Keeping the fields/arguments wouldn't be just because agave is currently doing it as well as using a constant variable, but more so because their `FeeRateGovernor` struct definition the doc comment on `lamports_per_signature` [mentions](https://github.com/anza-xyz/solana-sdk/blob/fee-calculator@v3.2.2/fee-calculator/src/lib.rs#L48) that this amount can change over time.

A change to how this is calculated, could ultimately lead to needing to re-thread the parameter through each caller and function.

Replacing the hardcoded values with the new `LAMPORTS_PER_SIGNATURE` is definitely something we should keep here, since it eliminates the use of a magic number.

PR #1593, opened 2026-06-12
chore: bump conformance commits to agave-v4.1.0-beta.3

yewman:
- Bump test-vectors to head of main, and solfuzz-agave to head of agave-v4.1.0-beta.3
- Disable feature status logs for libsolfuzz_sig.so to reduce noise in CI
- Change `fixtures.txt` from an inclusive list to an exclusive list and rename to `excluded.txt`
    - It is easier to maintain what is **NOT** supported than what **IS** supported 
    - We were missing new syscall groups as a result not updating the 'inclusive' list 
- Create `list_excluded.py` so `ci_run.sh` and `run.py` fixture exclusion remains in sync
- Create issues for missing harness implementations:
    - shred: https://github.com/Syndica/sig/issues/1605
    - vm_serialisation: https://github.com/Syndica/sig/issues/1606
    - cost and block harness are not added as issues because they are not immediately addressable


ajw221 (dismissed):

conformance/scripts/ci-run.sh
```diff
@@ -10,12 +10,26 @@ cd $conformance_dir
...
+# failing files, then links the selected subset into split-fixtures.
 comm -23 \
     <(find "env/test-vectors/" \
-        -path 'env/test-vectors/block/*' -prune \
-        -o -type f -name '*.fix' -printf '%P\n' | sort) \
+        "${prune_args[@]}" \
+        -type f -name '*.fix' -printf '%P\n' | sort) \
     <(sort scripts/failing.txt) \
     | sed "s_^_$PWD/env/test-vectors/_" \
     | xargs -d '\n' ln -s -t env/split-fixtures/
```
dnut: This complexity defeats the purpose of using a shared script to ensure the file is parsed correctly. It's trivial to parse it correctly directly in this script.

```suggestion
comm -23 \
    <(find env/test-vectors/ -type f -name '*.fix' -printf '%P\n' | sort) \
    <(sort scripts/failing.txt) \
    | grep -vE "^$(grep -vE '^\s*(#|$)' scripts/excluded.txt | paste -sd'|')" \
    | sed "s_^_$PWD/env/test-vectors/_" \
    | xargs -d '\n' ln -s -t env/split-fixtures/
```

This handles comments, blank lines, and single trailing slashes correctly.

If someone leaves extra unintended characters (like multiple trailing slashes or extra whitespace) in one of the filters in `excluded.txt`, the filter won't work. It would run extra tests, which would likely fail, and the person who added the extra characters would need to clean it up. In a way this is like a lint, I actually like it. We shouldn't be leaving unintended characters. I would also make the python code work this way.

conformance/scripts/excluded.txt
```diff
@@ -0,0 +1,15 @@
+# Test-vector directories to skip when running conformance fixtures.
+#
+# Used by both scripts/ci-run.sh and run.py as the single source of truth for
+# which fixture categories are NOT supported by the sig conformance runner.
+# Paths are relative to env/test-vectors/. Any fixture whose path equals an
+# entry below or lives under one (matched at a directory boundary) is skipped.
+#
+# scripts/failing.txt is a separate, file-level CI denylist for individual
+# `.fix` files that are known to fail.
```
dnut: I think we should combine them into a single file. Fundamentally they're solving the same problem. Some tests fail, so we shouldn't run them. The same filtering logic for excluded.txt can apply to the failing fixtures. One list of exclusions is simpler. We can add comments for why certain tests or test folders are excluded.

dnut (changes requested):

conformance/scripts/excluded.txt
yewman: I had a similar thought but after thinking about it keeping failed and excluded separate is preferable in my view. excluded.txt is what we can't even execute, while failures.txt is what we should be passing but are not due to a bug which we have decided to deprioritise right now. failures.txt should be a top priority and almost always empty except under special circumstance. I think this distinction is relevant so it is my preference to keep it as is

conformance/scripts/ci-run.sh
yewman: Yeah fair enough: b92059bd09e048797e789fc616bc

conformance/scripts/excluded.txt
dnut: how about this:
- urgent_failures.txt
- low_priority_failures.txt

In the scripts, concatenate them and apply unified filtering logic.

yewman: The exclusions aren't really failures though, they are just fixtures which we don't currently have harnesses for. Failures only pertain to ci-run.sh, not run.py because nix run should execute everything that we support. The failures are purely to allow certain failing fixtures through ci

dnut: It is a failure. It's just a very specific type of failure which you've decided to segregate from the other failures and treat differently.

There are several categories of failures:
1. harness is not implemented at all
2. harness is broken, causing all fixtures to fail
3. harness has a bug that only causes *some* fixtures to fail
4. the runtime has a bug (potentially more subcategories)

You've decided that category 1 is a special type of failure that deserves
- custom filtering logic
- a separate file
- is automatically filtered by run.py

Here's what I see as the ideal setup:
- the same filtering logic applies to all filter files
- filtering is implemented by the test runner 
- `run.py` and `zig build run` should be consistent in their filtering approach (ideally we eliminate run.py because the zig code can be improved to replace it)
- filters can be specified on the command line, with a sane default (let's say exclusions.txt is the default)
- you can specify multiple filter files (or a folder of filters). so for example, ci_run.sh could look like `zig build run --exclude exclusions.txt failures.txt`

As far as what's in scope right now, I don't want to block the merge of the fixtures upgrade, so we can leave most of this for later. Can we just improve the naming?

Currently there are two files `exclusion.txt` and `failures.txt`. Every entry in each file represents both a failure and an exclusion. The naming is totally ambiguous. What about this:
- unimplemented_harnesses.txt
- misc_failures.txt

Also I would throw them in a folder called `exclusions/` or `failures/`. Not important at this point though.

yewman: I don't agree with unimplemented harnesses being failures. I also don't want to hold this up any longer and am happy to review a more thorough refactor of the approach to handling exclusion / failures as a follow up. I have renamed the files as requested. Lmk if there is anything else you would like changed in directly within this PR. 
d9e60cde33bad27e1143b76753d04df92b3bbf5b

dnut: I guess it depends on how you define "failure." Mine was "the test does not run successfully" but I think your definition of failure is more specific, something like "we expect the test to run successfully, but it does not." My definition is purely technical, whereas yours seems to involve more intent or human judgment. It would be good to get on the same page about this but it's not critical for this pr.

dnut (approved):

ajw221 (approved):

PR #1599, opened 2026-06-15
feat(runtime): implement SIMD-0357 validator_admission_ticket

yewman:
## Summary

Implements [SIMD-0357 — Validator Admission Ticket](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0357-validator-admission-ticket.md)

At every epoch boundary, when `validator_admission_ticket` is active, the vote-account snapshot is reduced via `Stakes.cloneAndFilterForVat` to the top `MAX_ALPENGLOW_VOTE_ACCOUNTS` (2000) accounts with a BLS pubkey, non-zero stake, and at least the rent-exempt minimum for `VoteStateV4`.

When `alpenglow` is *also* active, two additional things happen:
1. The minimum balance check in the filter bumps by `VAT_TO_BURN_PER_EPOCH` (1.6 SOL), guaranteeing every surviving account can pay.
2. `burnVat` debits `VAT_TO_BURN_PER_EPOCH` from each surviving account into the incinerator. The existing freeze-time incinerator zeroing handles capitalization.

The live `stakes_cache` stays unfiltered; only the per-epoch `epoch_stakes` snapshot is reduced. Downstream consumers (leader schedule, fork choice, progress map, replay tower, rewards) pick up the filtered set automatically.

The `alpenglow` feature is `.unsupported` in sig, so in production only the filter step runs once VAT activates; the burn stays dormant until SIMD-0326 lands. Conformance fixtures exercise the full filter+burn path by forcing both flags active.

## Testing

- Unit tests on `Stakes.cloneAndFilterForVat` mirroring agave's `vote_account.rs` cases (truncate, non-alpenglow, border tie, insufficient lamports, empty set).
- Integration tests on `getEpochStakes` and `burnVat` across each combination of the two feature gates.

### DIFF
--- a/src/core/stakes.zig
+++ b/src/core/stakes.zig
@@ -1163,8 +1163,11 @@ pub fn randomEpochStakes(
 
     return sig.replay.epoch_transitions.getEpochStakes(
         allocator,
+        0,
         options.epoch + 1,
         &stakes_cache,
+        &sig.runtime.sysvar.Rent.INIT,
+        &sig.core.FeatureSet.ALL_DISABLED,
     );
 }
 
--- a/src/core/stakes.zig
+++ b/src/core/stakes.zig
@@ -23,6 +23,7 @@ const Stake = StakeStateV2.Stake;
 const Meta = StakeStateV2.Meta;
 const VoteState = sig.runtime.program.vote.state.VoteState;
 const VoteStateV3 = sig.runtime.program.vote.state.VoteStateV3;
+const VoteStateV4 = sig.runtime.program.vote.state.VoteStateV4;
 const VoteStateVersions = sig.runtime.program.vote.state.VoteStateVersions;
 
 const RwMux = sig.sync.RwMux;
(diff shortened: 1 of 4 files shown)


hamza-syndica (dismissed):

src/core/stakes.zig
```diff
@@ -503,6 +545,97 @@ pub const VoteAccounts = struct {
...
+            // Per SIMD-0357 drop everything with stake <= the first truncated
+            // entry, so Pubkey ordering can never be the tiebreaker.
+            var write: usize = 0;
+            for (entries.items) |item| {
+                if (item.stake > floor_stake) {
+                    entries.items[write] = item;
+                    write += 1;
+                }
+            }
+            entries.shrinkRetainingCapacity(write);
```
dnut: this iterates over every single item which is unnecessary, you only need to check a few items. also this re-writes items into the same slot it already occupies. the write is a noop because the list is already sorted by stake.

```suggestion
            var num_to_keep: usize = max_vote_accounts;
            while (num_to_keep > 0) : (num_to_keep -= 1) {
                if (entries.items[num_to_keep - 1].stake > floor_stake) break;
            }
            entries.shrinkRetainingCapacity(num_to_keep);
```

dnut (dismissed):
looks good, just one small performance tweak i would make

src/core/stakes.zig
yewman: Good catch :) 5b6940f1e2aa2acfa18cfb0c42f6f34e6a6bd3e5

dnut (approved):

hamza-syndica (approved):

PR #1600, opened 2026-06-15
refactor(runtime): hardcode accounts_lt_hash

hamza-syndica:
Hardcode `accounts_lt_hash` feature (SIMD-0215) - confirmed hardcoded in both:

- **Agave**: Feature gate removed; lattice hash is now unconditionally computed
- **Firedancer**: [a75575a412](https://github.com/firedancer-io/firedancer/commit/a75575a412b27869db72dc066d110967bc222d4a) - flamenco, features: clean up feature map

Closes #1585

### DIFF
--- a/src/replay/freeze.zig
+++ b/src/replay/freeze.zig
@@ -119,12 +119,12 @@ pub fn freezeSlot(allocator: Allocator, params: FreezeParams) !DistributedReward
     const rewards_allocator = params.rewards_allocator orelse allocator;
     const distributed = try finalizeState(allocator, rewards_allocator, params.finalize_state);
 
-    const maybe_lt_hash, slot_hash.mut().* = try hashSlot(
+    const lt_hash, slot_hash.mut().* = try hashSlot(
         allocator,
         params.hash_slot,
         params.thread_pool,
     );
-    if (maybe_lt_hash) |lt_hash| params.accounts_lt_hash.set(lt_hash);
+    params.accounts_lt_hash.set(lt_hash);
 
     params.logger.info().logf(
         "froze slot {} with hash {s}",
@@ -445,7 +445,7 @@ pub fn hashSlot(
     allocator: Allocator,
     params: HashSlotParams,
     thread_pool: *sig.sync.ThreadPool,
-) !struct { ?LtHash, Hash } {
+) !struct { LtHash, Hash } {
     var zone = tracy.Zone.init(@src(), .{ .name = "hashSlot" });
     defer zone.deinit();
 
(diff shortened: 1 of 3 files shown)


src/replay/freeze.zig
ajw221: because this can no longer return a null `LtHash` we should update the return type

A related cleanup we should consider, whether it's in this PR or a followup one, is tightening `SlotState.accounts_lt_hash` from `Mux(?LtHash)` to `Mux(LtHash)`. With `hashSlot` always producing a real value, every production initializer of this field is provably non-null: `SlotState.GENESIS` uses `.IDENTITY`; `fromBankFields` is fed `manifest.bank_extra.accounts_lt_hash`, which is typed non-optional in `ExtraFields` and whose bincode reader either returns a real hash, defaults to `.IDENTITY` on EOF, or fails with `error.DeltaLtNotPresent`; and `fromFrozenParent` inherits the parent's value. The only remaining `null` is in one test call at `src/consensus/progress_map.zig:1255`, and the sole reader (`src/replay/service.zig:710`) already unwraps with `.?`. So this should reduce to: changing the field type and the `lt_hash: ?LtHash` parameter on the two `fromBankFields*` constructors, swapping that one test to `.IDENTITY`, and dropping the `.?` in `service.zig`.

also worth noting that the sibling field `SlotConstants.parent_lt_hash` (and `HashSlotParams.parent_lt_hash`) must stay `?LtHash`, `fromBankFields` legitimately sets it to null for the snapshot-loaded root slot, which is short-circuited by the already-frozen check in `freezeSlot` before `hashSlot` runs, which is the invariant `error.UnknownParentLtHash` exists to protect

ajw221, line 487: unrelated to this PR, but same as above

ajw221: because of the changes mentioned below, this would no longer be null nor would it need a null check

ajw221 (changes requested):

src/replay/freeze.zig
hamza-syndica: https://github.com/Syndica/sig/pull/1600/commits/f6e6fa5d3f2f8e65499ad6bda54e4e8691c62e40

hamza-syndica: https://github.com/Syndica/sig/pull/1600/commits/f6e6fa5d3f2f8e65499ad6bda54e4e8691c62e40

hamza-syndica: I think we have nothing to change here

yewman (approved):

ajw221 (approved):

PR #1601, opened 2026-06-15
fix(v2:scripts): add corrupt test mode to shred-stream tool

prestonsn:
Adds `corrupt` test mode, which selects shreds by `--seed`, `--count`, and `--shred-kind`, then flips 1 random bit in each of `--corrupt-bytes` distinct packet byte positions per selected shred. 

Also cleans up some shred selection logic.

### Example usage

```zig
zig build shred-stream -- \
  --ledger ledger \
  --target 30.161.17.131:8002 \
  --start-slot 410010000 \
  --end-slot 410010200 \
  --test-mode corrupt \
  --seed 0xdeadbeaf \
  --count 100
```

### DIFF
--- a/v2/scripts/shred_stream.zig
+++ b/v2/scripts/shred_stream.zig
@@ -25,6 +25,7 @@ const TestMode = enum {
     drop,
     late,
     duplicate,
+    corrupt,
 
     fn parse(raw: []const u8) ?TestMode {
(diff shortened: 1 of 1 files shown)


hamza-syndica (approved):

Sobeston (approved):
Seems fun!

PR #1604, opened 2026-06-16
refactor(v2): simplify topology

dnut:
I took a close look at the topology code, and found that it is extremely clever. In fact, it is *too* clever. I was confused by the layers of abstraction, such as the generics and the domain specific language that it invents and parses. Untangling it all in my mind took quite a while. I felt that we could accomplish the same thing with a simpler approach.

I took a crack at simplifying it by trying to move away from *"define a declarative config, parse it, pass functions into functions, and magically initialize everything behind the scenes with hidden metaprogramming"* and closer towards *"initialize data, pass it into a function."* It was difficult to eliminate generics and metaprogramming entirely due to the nature of the problem space. We're generically initializing arbitrary topologies after all. But I tried to reduce it to a minimal set of necessary abstractions. I'm sure my approach is flawed, but I'm satisfied with this as an initial simplification.

I tried to keep the top-level code aesthetically just as concise and declarative as the prior code. This was tricky since I've moved from a domain-specific configuration language to imperative code, but I think it worked out nicely, partly because I was able to reduce fragmentation.

Please take a close look at whatever you understood that the topology code was supposed to be achieving, to make sure I haven't eliminated any necessary features. I believe it's all still there, but I could use some more eyes on this.

### Ideas for future improvement

I replaced services.zon with services.zig. But we not should have a services dot anything. main.zig already defines the topology. services.xyz is just another separate file that does nothing except define the parameters for services. The services themselves should be the *only* place that lists their params - directly above serviceMain. Anything else violates the principle of locality and makes the code convoluted. The problem is, implementing this requires some rework of the rules about how different modules import from each other. It potentially weakens some guarantees we get about how code is compiled and organized. So I decided this is not in scope for the current refactor. I'll keep thinking about this to see if there's a nice solution.

The topology, linux, and init code should all be consolidated into an init module. It has no business being in a shared library. But there are two problems with doing this now. First of all this poses some challenges because it means topology.zig would need to be included in multiple modules, so I'd need to create another module to fix this. Furthermore, I have a broader vision that reorganizes all the folders in v2, and any module rework would be best understood in a single holistic rework that can be reviewed independently, and as a whole.


jbuckmccready (dismissed):
I find this is big improvement for being able to track how things are wired.

v2/init/main.zig
```diff
@@ -85,6 +89,21 @@ const Config = struct {
...
+const Topology = struct {
+    net: ServiceRegions(services.net),
+    gossip: ServiceRegions(services.gossip),
+    shred_receiver: ServiceRegions(services.shred_receiver),
+    replay: ServiceRegions(services.replay),
+    snapshot: ServiceRegions(services.snapshot),
+    accounts_db: ServiceRegions(services.accounts_db),
+    telemetry: ServiceRegions(services.telemetry),
+    exec: ServiceRegions(services.exec),
+};
```
prestonsn, line 105: I feel like this change now has 3 separate lists for the services in 3 different places, while the original had the (authoritative) [services.zon](https://github.com/Syndica/sig/blob/8bedf04d4c92a982fa3adb06bb83c4cec1ae86ef/v2/init/services.zon) and the service binding/region init details in [main.zig](https://github.com/Syndica/sig/blob/8bedf04d4c92a982fa3adb06bb83c4cec1ae86ef/v2/init/main.zig#L139), we now have this list highlighted above, a second in `build.zig`: 

https://github.com/Syndica/sig/blob/7b42f508e73b03bab2d47f2bb6fe19054c9a0210/v2/build.zig#L4-L15

And a third in the form of the `ServiceSpec`s in [services.zig](https://github.com/Syndica/sig/blob/dnut/v2/simplify-topology/v2/init/services.zig)

Is there some way to consolidate/derive the other lists from a single point-of-reference that could live in the new `init` module?

v2/lib/topology.zig
```diff
@@ -14,1090 +25,585 @@ const linux = std.os.linux;
...
+    mseal(@ptrCast(resolved.runner), @sizeOf(lib.runner.Region));
 
-            closeAllFdsExceptStderr(params.stderr.handle);
+    closeAllFdsExceptStderr(resolved.stderr);
 
-            // makes it impossible for the service to gain privileges
-            std.debug.assert(try std.posix.prctl(.SET_NO_NEW_PRIVS, .{ 1, 0, 0, 0 }) == 0);
+    std.debug.assert(try std.posix.prctl(.SET_NO_NEW_PRIVS, .{ 1, 0, 0, 0 }) == 0);
 
-            // install a basic seccomp filter that bans syscalls except write+sleep
```
prestonsn, line 938: nit: I found these code comments helpful (could even be improved here)

v2/lib/topology.zig
```diff
@@ -14,1090 +25,585 @@ const linux = std.os.linux;
...
 
-        /// Returns the segfault handler function exposed by the specified service.
-        pub fn getFaultHandlerFn(service: ServiceId) SigactionFn {
-            return switch (service) {
-                inline else => |s| @extern(SigactionFn, .{ .name = getFaultHandlerName(s) }),
-            };
-        }
+        /// Allocate a memfd of `size` bytes and mmap it RW.
+        pub fn sized(size: usize) !Region(T) {
+            std.log.info("Initializing region '{s}' ({} bytes)", .{ @typeName(T), size });
```
prestonsn, line 71: More of a nit, but since `Region(T)` is a bit more "generic" now, these logs could be updated a bit. They still use `@typeName(T)`, which kinda look uninformative in the logs now for simple regions (can't identify the region as easily as before, I think this follows from the other ["identity" related feedback](https://github.com/Syndica/sig/pull/1604#discussion_r3438762494))

```log
info: Initializing region 'net.Pair' (44171904 bytes)
info: Initializing region 'net.Pair' (44171904 bytes)
```
vs how they looked prior: 

```
info: Initialising: .net_to_shred
info: Initialised: Region `net_to_shred` shared with [ net (rw), shred_receiver (rw) ]
info: Initialising: .net_to_gossip
info: Initialised: Region `net_to_gossip` shared with [ net (rw), gossip (rw) ]
```

Adding the region sizes is a nice change though 🫡

prestonsn (commented):
These changes look really good to me, I don't see any major gaps between what we had prior and this revamp. Though I did want to raise some tradeoffs with this design (that could be addressed later on too)

v2/init/main.zig
```diff
@@ -136,332 +155,213 @@ pub fn main() !void {
...
+                .replay_transaction_pool = transaction_pool.finish(),
+                .block_pool = block_pool.finish(),
+                .exec_req_response = exec_req_response.finish(),
+                .tel = telemetry_region.finish(),
             },
-            .replay_account_lookups => {
-                std.debug.assert(buf.len == @sizeOf(lib.accounts_db.AccountLookups));
-                const data: *lib.accounts_db.AccountLookups = @ptrCast(buf);
-                data.init();
+        },
```
prestonsn, line 279: I think one meaningful invariant getting dropped with the removal of the original binding map is the "identity" of each service + region binding had.  

The new `ServiceRegions` checks that each servce field gets the right region type, but does not check that it's the right instance of the region (example: `net_to_gossip` and `nt_to_shred` are both `Region(std.net.Pair)`, so something like this would still compile: 

```zig
gossip.net_pair = net_to_shred.finish()
```

This mapping was handled by the DSL version, so leaves a little less room for mistakes:  

https://github.com/Syndica/sig/blob/8bedf04d4c92a982fa3adb06bb83c4cec1ae86ef/v2/init/main.zig#L222-L225

Though we could close this gap in the future/add more validation checks later

dnut: I don't see a gap or lost invariants

> does not check that it's the right instance of the region

The previous approach did not do this either. It was legal to do something like this:

```zig
    .net_to_shred = .initMany(&.{
        .@"net:to_gossip",
        .@"shred_receiver:from_net",
    }),
    .net_to_gossip = .initMany(&.{
        .@"net:to_shred",
        .@"gossip:from_net",
    }),
```

There was no way to know that "net:to_shred" is supposed to identity with "net_to_shred" unless you actually write it that way here when you wire up the topology. This is entirely up to the programmer to write the code with whatever bindings they please. The only difference is that now the binding is defined with ordinary struct initialization instead of a DSL.

This flexibility is actually desirable for testing. Originally we didn't even have binding definitions like this. We just used the same name "net_to_shred" and there was only one way to wire things up. This meant we couldn't define custom topologies for testing, there was only one valid topology. Trevor changed it to add this flexibility a few weeks ago. I'm leaving the same amount of flexibility here.

v2/init/main.zig
dnut: Regarding the enum in build.zig, I agree that's an unfortunate redundancy. It's also easy to eliminate, so I went ahead: 50c672f6db903166ff2009099cec1b7bbe0adfea

Regarding the other two, there are two ideas that I'd like to distinguish:
1. service definitions: *what services exist?*
2. topology definitions: *what combination of services should run in a particular context?*

Previously, services.zon fulfilled both role 1 in general, and role 2 for the main topology. This conflates two distinct ideas, which I think is confusing. Eventually we won't even have a topology that runs every service, so this will not work long term. We'll need separate lists regardless. It's only a coincidence that it happens to be the same list for now. Unifying them into a single list was more of a fragile hack than a meaningful consolidation.

For test topologies, this was already evident. The topology definition was implicitly generated by build.zig. I think this made the code pretty confusing because it is inconsistent with how the main topology works, and because it makes behavior implicit and non-obvious. You shouldn't have to look at build.zig to understand what threads are being spawned by a main function.

I separated the primary topology definition from the service definition, and I consolidated to a single approach to topology definitions that is consistent for both the primary and test topologies. I did this because I think it improves clarity and it will become a technical necessity.

v2/lib/topology.zig
dnut: I also noticed this loss of context from the logs. Basically the difference is that previously a label was passed into the create region function, and now there is no longer a label being passed in. This was implicit in the previous approach and came for free. Now it would need to be explicit in the form of an additional parameter.

v2/lib/topology.zig
dnut: I'll take a close look to identify any comments I deleted for code that I kept, and add them back

v2/init/services.zig
```diff
@@ -0,0 +1,96 @@
+const lib = @import("lib");
+const topology = lib.topology;
+
+const ServiceSpec = topology.ServiceSpec;
+
+pub const accounts_db: ServiceSpec = .{
```
InKryption, line 6: Comment on the file in general: I do like that this acts as a sort of "header" file/source of truth between the topology's view of services and the services themselves at the type level. Seems relatively easy to avoid accidentally referencing irrelevant specs, so lazy analysis will avoid us analyzing the parameters for other services.

v2/services/net.zig
```diff
@@ -15,13 +16,8 @@ pub const name = .net;
...
 
-pub const ReadOnly = struct {};
-
-pub const ReadWrite = struct {
-    gossip_pair: *Pair,
-    shred_pair: *Pair,
-    tel: *tel.Region,
-};
+pub const ReadOnly = services.net.ReadOnly;
+pub const ReadWrite = services.net.ReadWrite;
```
InKryption, line 20: General comment: the main downside to the `ServiceSpec` solution is that it gimps zls' ability to provide auto-completion for these types, since it won't/can't resolve types through composite values like these. That said, that's a deficiency with the tooling that should get better, and it's of minimal consequence, so I'm just putting this here to explicitly acknowledge it and say I'm fine with that.

v2/lib/topology.zig
```diff
@@ -14,1090 +25,585 @@ const linux = std.os.linux;
...
+///
+/// ```zig
+/// const my_service: ServiceSpec = .{
+///     ReadOnly: struct {
+///         config: *Config,
+///     },
+///     ReadWrite: struct {
+///         ring: *Ring,
+///     },
+/// };
```
InKryption, line 120:
```suggestion
/// const my_service: ServiceSpec = .{
///     .ReadOnly = struct {
///         config: *Config,
///     },
///     .ReadWrite = struct {
///         ring: *Ring,
///     },
/// };
```

v2/lib/topology.zig
```diff
@@ -14,1090 +25,585 @@ const linux = std.os.linux;
...
-        }
+// -- Service tracking -- //
+
+/// State for one spawned service: pid (sandboxed) or thread (threaded), plus the
+/// runner memfd we'll later map in the parent. Memfds and threads are owned by the parent.
+pub const Service = struct {
+    label: [:0]const u8,
+    runner_memfd: Memfd,
+    runner: *lib.runner.Region = undefined,
+    activity_view: lib.runner.Activity.RunnerView = undefined,
```
InKryption: I would prefer that these be set to `undefined` explicitly wherever the struct is initialized. Shouldn't be hidden away.

v2/lib/topology.zig
```diff
@@ -14,1090 +25,585 @@ const linux = std.os.linux;
...
+/// Result of `spawn`: every service in the topology, plus a `ThreadExit` used in
+/// threaded mode. The address of this struct must be stable for the lifetime of
+/// the services (threads point into it), which is why `spawn` takes it as an out
+/// parameter rather than returning it by value.
+pub const Children = struct {
+    mode: Mode,
+    services_buf: [max_services]Service = undefined,
+    services_len: usize = 0,
+    /// Used only in `.threaded` mode to wake up the parent on first service exit.
+    thread_exit: ThreadExit = .{},
```
InKryption: Children is only initialized in one place, and it should probably only ever be initialized in that one place, so I think these default field inits should just be moved there.

In general, I would prefer to avoid default field values where the fields are not independent (like they often are in parameter/configuration structs). The danger being that if one was to write `.{ .mode = .sandboxed, .services_len = 1 }`, you have now created an invalid variant which will not necessarily be spotted right away.

This is also the recommendation of the [zig langref itself](https://ziglang.org/documentation/master/#Faulty-Default-Field-Values).

v2/lib/topology.zig
```diff
@@ -14,1090 +25,585 @@ const linux = std.os.linux;
...
+            .label = svc_name,
+            .runner_memfd = runner_memfd,
+            .slot = .{ .sandboxed = try spawnSandboxed(svc_name, entrypoint, args) },
+        },
+        .threaded => blk: {
+            var threaded_args = try resolveArgs(.{
+                .runner_memfd = runner_memfd,
+                .rw = rw_buf[0..rw_count],
+                .ro = ro_buf[0..ro_count],
+            });
```
InKryption: Tiny nit: would maybe be nice to extract the bulk of the `threaded` branch out into its own `spawnThreaded` function for symmetry.

v2/lib/topology.zig
```diff
@@ -14,1090 +25,585 @@ const linux = std.os.linux;
...
+    buf: *[max_regions_per_service]Memfd,
+) usize {
+    var count: usize = 0;
+    inline for (@typeInfo(Layout).@"struct".fields) |field| {
+        const region = @field(regions, field.name);
+        buf[count] = region.memfd;
+        count += 1;
+    }
+    return count;
+}
```
InKryption, line 382: Moving this logic here and out of ipc.zig makes sense, thumbs up.

v2/init/main.zig
prestonsn: That makes a lot of sense, and agreed on the separation between topology definitions and service definitions. My main concern here was avoiding potential drifting between the individual service manifests and that's been fully addressed here.

v2/lib/topology.zig
```diff
@@ -14,1090 +25,585 @@ const linux = std.os.linux;
...
+/// Spawn every service described by `topology` in one batch. Each field of
+/// `topology` must be a `ServiceLayout(spec)`. Each field name is used to look up
+/// the service's main function (`svc_main_<name>`) and fault handler. Regions are
+/// passed in field declaration order, matching what `start_service.zig` expects.
+///
+/// `out` is treated as undefined on invocation, and it is initialized by this
+/// function. The caller's storage is what the threads (in threaded mode) hold
+/// pointers into, so it must outlive the services. After this returns, call wait,
+/// cancel, or isActive on `out`.
+pub fn spawn(out: *Children, mode: Mode, topology: anytype) !void {
```
InKryption: A couple of thoughts:
1. Could we define this as a method on `Children`? It is essentially the method used to initialize the structure, and most instances of "data structures initialized in place" do tend to define that as a method rather than a free function. I don't feel super strongly about this alone, but ties into second thought.

2. Overall I think the changes here to "flatten" a lot of what was fairly complex comptime logic is the correct move, but making these all completely independent free functions seems like a slight over-correction in the opposite direction; I think the API could be made more unified if we just made `Children` and its subsidiary methods generic over `topology`, instead of making each relevant method individually generic and then type-erased in `Children`. e.g.:
```zig
/// Each field of `Topology` must be a `ServiceLayout(spec)`. Each field name is
/// used to look up the service's main function (`svc_main_<name>`) and fault handler.
/// Regions are passed in field declaration order, matching what `start_service.zig` expects.
pub fn Children(comptime Topo: type) type {
    const topo_info = @typeInfo(Topo).@"struct";
    return struct {
        mode: Mode,
        services_buf: [topo_info.fields.len]Service,
        services_len: usize,
        /// Used only in `.threaded` mode to wake up the parent on first service exit.
        thread_exit: ThreadExit,
        const ChildrenSelf = @This();

        /// Count the number of times that RegionType appears within the topology.
        pub fn countRegionShares(comptime RegionType: type) usize {
            // -- snip -- //
        }

        pub fn spawn(self: *ChildrenSelf, mode: Mode, topology: Topo) !void {
            // -- snip -- //
        }

        fn spawnOne(
            comptime svc_name: [:0]const u8,
            layout: @FieldType(Topo, svc_name),
            mode: Mode,
            thread_exit: *ThreadExit,
            service_idx: u16,
        ) !Service {
            // -- snip -- //
        }

        fn collectRegions(
            comptime Layout: type,
            regions: Layout,
            buf: *[topo_info.fields.len]Memfd,
        ) usize {
            // -- snip -- //
        }

        // -- etc -- //
    };
}
```

Haven't thought about this next bit too thoroughly, but maybe this also helps in regaining (for "free") the lost log context pointed out by Preston in another comment?

Lmk your thoughts.

InKryption (changes requested):
Overall I think I like the direction of these changes; I did want to move away from the more complex declarative API, but at the time of my changes all of the concepts were still a bit tangled in my brain until after I had landed on the current design, and I think I this all does a good job of fully decoupling all of the relevant concepts.

v2/lib/topology.zig
prestonsn: Yeah that matches what I'm seeing. It would be nice to id them some way, I don't think it would be bad to somehow name the regions. Could also do this in the future, can resolve at your discretion :saluting_face:

v2/init/main.zig
prestonsn: I looked over the previous code again and you're right on this. Perhaps "invariant" was a bit too strong wordage, I think it's more of a lost guardrail/readability. I'm resolving this

v2/init/main.zig
```diff
@@ -136,332 +155,213 @@ pub fn main() !void {
...
+    snapshot_ready_to_accounts_db.ptr().init();
+
+    const unrooted_memory = config.accounts_db.unrooted.toBytes();
+    var account_pool: Region(lib.accounts_db.AccountPool) =
+        try .sized(@sizeOf(lib.accounts_db.AccountPool) + unrooted_memory);
+    account_pool.ptr().init(unrooted_memory);
+
+    var shreds_to_replay: Region(lib.shred.DeshredRing) = try .simple();
+    shreds_to_replay.ptr().init();
+
```
Sobeston, line 210: I feel like putting this all back in main is a bit of a downgrade, it will cause code duplication whenever someone starts up a custom topology

v2/build.zig
```diff
@@ -411,13 +422,14 @@ const RunnerModuleOptions = struct {
...
         tracy: *Build.Module,
+        services: *Build.Module,
     };
 };
 
 const BlackBoxTest = struct {
     name: []const u8,
     root_source_file: Build.LazyPath,
-    services: std.EnumSet(topo_types.ServiceId),
+    services: []const []const u8,
```
Sobeston, line 432: More of a general comment, but I feel that the generics we had were a strength, this reduces type safety

Sobeston (changes requested):
I'm okay with the general direction, just have a couple comments to add

v2/services/net.zig
dnut: It doesn't need to be a struct of types. It could also be a namespace struct with decls in it.

v2/lib/topology.zig
dnut: 75779c821176aec3710af4a425c18d14600be3d3

v2/lib/topology.zig
dnut: 75779c821176aec3710af4a425c18d14600be3d3

v2/lib/topology.zig
dnut: 75779c821176aec3710af4a425c18d14600be3d3

v2/lib/topology.zig
dnut: 75779c821176aec3710af4a425c18d14600be3d3

Preston's comment was about the code that runs during region initialization, and that code is still decoupled from the topology.

v2/lib/topology.zig
dnut: This commit (75779c821176aec3710af4a425c18d14600be3d3) removed the default field values, but didn't properly initialize the struct.

This commit improves the struct initialization: 5cf76d32d4943f90f04234dca141f1541762c126

v2/build.zig
dnut: It's still verified at build time, so does it actually make a difference? For example if you misspell a service, you'll get an error like this when you try to build it:
```
thread 338677 panic: unknown service 'teemetry'
```

I could change this back to an enum, but it involves a bit more magical feeling comptime code so I'm leaning towards the strings.

v2/init/main.zig
dnut: In the existing approach, every custom topology needs its own Region union with an init function to initialize each region, and a call to `serviceMap` with the initialization parameters. It's roughly equivalent to this code, just more fragmented. Actually I think this approach involves less duplication because it involves fewer redundant lists of the same regions within a single topology's initialization, which keeps the code more concise.

v2/lib/topology.zig
dnut: It's pretty easy to add another param to the function to print a name, but it clutters the code in main with repetition, making it a bit harder to read.

It's a small cost that I'd be willing to pay in many cases, but I'm just not seeing much benefit. The region name would only be used for this one log message and then it would be discarded. Almost all the regions use types that are specific to the region, so the log is already clear in most cases. For the few exceptions, net pairs, they appear in the same sequence in the function and the log, so you could trace it back if necessary.

I don't see a lot of situations where someone would need to trace back these logs. I think the main use of these logs will be for giving a rough idea how far sig's init process has gotten, which probably won't involve a lot of scrutiny on the details of the individual log messages

Anyway, I don't feel strongly about it. I was indecisive about this at first. If you see a lot of value to this, let me know.

v2/init/services.zig
dnut: I feel ambivalent about this actually.

I like the idea of each service's file defining its own API rather than fragmenting it into a separate file. It keeps the service definition consolidated and easy to understand in its own scope without jumping around.

But that comes with a cost of importing the service file, which contains code that should not be imported. This could potentially impact compile times (not sure), and it results in ambiguity about what code is used where. It would require some loosening of constraints we currently have in our build approach. So I decided it wasn't worth it, and kept it separate.

v2/lib/topology.zig
dnut: af4f15e3baa0823045a05cc4aa58ccdf332ad474

v2/init/services.zig
InKryption: Yeah, it's certainly a trade-off. Could go the extra mile and define a separate "header file" for each service, but then we lose the ability to iterate over them nearly as conveniently.

v2/services/net.zig
InKryption: It could be, but then that's a trade-off in complexity, since now instead of passing around a value parameter, you're passing around an amorphous type that needs to be validated to meet a certain "interface" instead, so it's a bit of a wash. I think the loss of auto-completion is a lesser evil than the code complexity.

v2/lib/topology.zig
prestonsn: Fair points, and that makes sense to me! I think it's easy for us to come back to it *if* we need it for something in the future. I like the new region size bytes to the output, so this is a nice upgrade overall

prestonsn (dismissed):

prestonsn (approved):

v2/build.zig
Sobeston: isn't this a runtime error rather than build error?

> I could change this back to an enum, but it involves a bit more magical feeling comptime code so I'm leaning towards the strings.

How much of a difference do you think it would make to the code?

v2/init/main.zig
Sobeston: Hm okay, I see that this is close to equivalent.

v2/build.zig
dnut: I was careful to say "build time" instead of "compile time" because it occurs during the runtime of the build script. Compiling and running the build script always happen together so I don't see a meaningful difference.

> How much of a difference do you think it would make to the code?

I think it would just be extra code to construct a std.builtin.Type.Enum and reify it with @Type.

Sobeston: >  it occurs during the runtime of the build script

Ah okay I missed that. Then it's no big deal.

Sobeston (approved):

InKryption (approved):

PR #1609, opened 2026-06-16
refactor(runtime): add recent blockhash checker interface

jbuckmccready:
After research and discussion regarding the blockhash queue we're realizing we don't need it at all for runtime execution, and plan for v2 is to implement the age check by just walking up the blocks and incrementing to determine the block height for validity.

High level plan for v2:
1. Populate blockhash history from snapshot in replay, this will provide blockhashes for block references (without actual block entries/transactions).
2. Blockhashes for blocks are stored in array that is indexable by block reference id
3. Implement the `RecentBlockhashChecker` interface by walking the block references and comparing with the transactions recent blockhash to find if it's within the max blockheight allowed.

This PR just refactors `loadAndExecuteTransaction` to use an interface function for the recent blockhash validity check, removing the dependence on `BlockhashQueue` type from `fn loadAndExecuteTransaction` in preparation for v2.


hamza-syndica (approved):

dnut (changes requested):
I think we should do #1631 instead of this.

PR #1610, opened 2026-06-16
fix(v2:telemetry): comptime log filter encoding length

prestonsn:
Fixes a small bug where comtime log filter encoding length could be undercounted by omitting header bytes and bytes still buffered in the temporary counting writer

### DIFF
--- a/v2/lib/telemetry/log.zig
+++ b/v2/lib/telemetry/log.zig
@@ -433,7 +433,10 @@ pub const Filter = struct {
         default_log_level: Level,
         str: []const u8,
     ) ParseError!u64 {
-        const buffer_len = if (@inComptime()) str.len else 4096;
+        const buffer_len = if (@inComptime())
+            str.len + (std.mem.count(u8, str, ",") + 2) * @sizeOf(Filter.Header)
+        else
+            4096;
         var buffer: [buffer_len]u8 = undefined;
         var discarding: std.Io.Writer.Discarding = .init(&buffer);
         parseListAndWriteBinary(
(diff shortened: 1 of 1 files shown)


v2/lib/telemetry/log.zig
```diff
@@ -433,7 +433,10 @@ pub const Filter = struct {
         default_log_level: Level,
         str: []const u8,
     ) ParseError!u64 {
-        const buffer_len = if (@inComptime()) str.len else 4096;
+        const buffer_len = if (@inComptime())
+            str.len + (std.mem.count(u8, str, ",") + 2) * @sizeOf(Filter.Header)
+        else
+            4096;
         var buffer: [buffer_len]u8 = undefined;
         var discarding: std.Io.Writer.Discarding = .init(&buffer);
```
dnut: What's the point of this change? Actually why are we even using a buffer here? I believe the optimal use of a discarding writer would use an empty buffer. Otherwise it's just doing an extra memcpy for no reason.
```suggestion
        var discarding: std.Io.Writer.Discarding = .init(&.{});
```

jbuckmccready: Yeah it seems like there shouldn't be any buffer here because all the calls for drain do is increment the count, there's no reason to buffer anything.

Looking at it a bit more the awkward thing from an API perspective is it will make the virtual drain call on every write if the buffer is zero length (even if it's just writing 1 byte at a time in a loop for megabytes) because the counting happens behind the virtual call which is not really optimal (in such a case I wouldn't be surprised if the memcpy to buffer is actually faster due to how it can be compiled). But none of that is relevant in this case.

prestonsn: Oops good catch, I think I payed too much attention to the `WriteFailed` panic (and thought it was a buffer sizing issue); didn't realize this was just counting bytes. Fixed it here: https://github.com/Syndica/sig/pull/1610/commits/f5e62fa2b00a377b37fd620814043c92e6b229d1

jbuckmccready (approved):

dnut (approved):

PR #1615, opened 2026-06-17
feat(v2:shred_recv): add Effects interface

prestonsn:
Adds an `Effects` interface similar to Gossip's for the `shred_receiver` service's `Receiver` loop. This is part of the effort for adding unit tests and formalizing the interface/externalities for shred ingestion/replay services.


Sobeston (dismissed):

v2/lib/shred/receiver.zig
```diff
@@ -14,372 +14,476 @@ const Signature = solana.Signature;
...
-            // fec set is already being built. This branch will be taken for 31/64 shreds (assuming
-            // zero packet loss).
-
-            // variant should match that of the first recorded shred in the fec set
-            if ((shred.variant.isData() and !shred.variant.eql(fec_set_ctx.data_variant)) or
-                (shred.variant.isCode() and !shred.variant.eql(fec_set_ctx.code_variant)))
+        // TODO: report return values to observability
+        // TODO: report back equivocating shreds, so that we can construct and send out duplicate proofs
+        pub fn processPacket(
+            state: *Self,
```
kprotty: why not `self`

v2/lib/shred/receiver.zig
```diff
@@ -14,372 +14,476 @@ const Signature = solana.Signature;
...
+            return .fec_set_finished;
         }
+    };
+}
 
-        state.done.setDone(&shred.signature, fec_set_id);
-        state.in_progress.removeFinishedSet(fec_set_ctx);
-
-        tracy.frameMarkNamed("finished FEC sets");
+pub const PacketError = error{
```
kprotty, line 397: this need to be explicit?

kprotty (commented):
makes sense. few nitpick questions

v2/lib/shred/receiver.zig
prestonsn: true, https://github.com/Syndica/sig/pull/1615/commits/5aa8ff55c1571c9b427e16f5d6abc6d8491bfa74

v2/lib/shred/receiver.zig
prestonsn: My intention here was the make all the possible failure scenarios for shred parse explicitly typed such that the "promised" failure states exposed in the `Effects` interface are well-known/defined (callback contract now has a finite error set as opposed to more generic `anyerror` :duck: )

kprotty: is this for PacketResult? Not sure `reportReceiverPacketResult` needs to be in the Effect interface itself. Its the return value of `processPacket`, not an internal action needed to occur. The observability reporting should be at the call site instead of through an Effect

kprotty: same for `reportShredParseResult`; could be its own error returned by processPacket that the caller checks

prestonsn: True, yeah I kinda grew the scope of the Effects from modelling just the "externalities" of a service, to also having hooks into results. I think when I was putting this together initially, I thought we could use the Effects interface to build out custom `Effects` implementations for things like the solfuzz-agave shred fuzzer (which would use the outputs from the Effects struct it hands to its build of `shred_service`). Likely not the path we're going down anyways. 

Removed `reportShredParseResult` + `reportReceiverPacketResult`, but kept the strongly-typed error enum since it's kinda nice to see all the possible failure cases in one place: https://github.com/Syndica/sig/pull/1615/commits/fcb1fb91fa162a85f6fb88adbba5bbc24749dc8b

kprotty (approved):

dnut (changes requested):
I already discussed this with preston, but I don't think we should have an effects interface for the shred receiver. A ring buffer is already an effects interface. This just piles another interface on top of an existing interface to create unnecessary abstractions and unnecessary complexity that make the code more difficult to understand. I'm open to further discussion with others but for now I don't think we should have this.

PR #1616, opened 2026-06-18
fix(vm): CALL_IMM collision handling to match Anza behavior

hamza-syndica:
Fixes CALL_IMM collision where a key exists in both the syscall and function registries. Sig now matches Anza: both registries are checked, and the function registry target PC is validated.

Closes #1560


Sobeston (dismissed):

shared/vm/interpreter.zig
```diff
@@ -459,18 +459,36 @@ pub const Vm = struct {
             self.registers.set(.pc, frame.return_pc);
         }
 
+        /// [agave] https://github.com/anza-xyz/sbpf/blob/v0.19.0/src/interpreter.rs#L552-L583
```
yewman: NIT: may as well permalink latest
```suggestion
        /// [agave] https://github.com/anza-xyz/sbpf/blob/v0.21.0/src/interpreter.rs#L552-L587
```

yewman (commented):
One nit otherwise looks good 👍

yewman (approved):

Sobeston (approved):

PR #1617, opened 2026-06-18
feat(replay/accountsdb) basic account fetching + unrooted store

Sobeston:
Will likely follow this up with some restructuring of replay + async account fetching

### DIFF
--- a/v2/init/main.zig
+++ b/v2/init/main.zig
@@ -266,6 +266,8 @@ pub fn main() !void {
         try .sized(@sizeOf(lib.accounts_db.AccountPool) + unrooted_memory);
     account_pool.ptr().init(unrooted_memory);
 
+    var replay_scratch: Region([lib.replay.scratch_buffer_size]u8) = try .simple();
+
     var shreds_to_replay: Region(lib.shred.DeshredRing) = try .simple();
     shreds_to_replay.ptr().init();
 
@@ -332,11 +334,14 @@ pub fn main() !void {
         .replay = .{
             .ro = .{},
             .rw = .{
+                .scratch_memory = replay_scratch.finish(),
                 .snapshot_metadata_in = snapshot_metadata.finish(),
                 .deshredded_in = shreds_to_replay.finish(),
                 .replay_transaction_pool = transaction_pool.finish(),
                 .block_pool = block_pool.finish(),
                 .exec_req_response = exec_req_response.finish(),
+                .account_pool = account_pool.finish(),
+                .account_lookups = replay_account_lookups.finish(),
                 .tel = telemetry_region.finish(),
             },
         },
--- a/v2/lib/replay.zig
+++ b/v2/lib/replay.zig
@@ -1,6 +1,7 @@
 const solana = @import("solana.zig");
 const collections = @import("collections.zig");
 const ipc = @import("ipc.zig");
+const accounts_db = @import("accounts_db.zig");
 
 // This is a bit large currently because of the unrooted store
 pub const scratch_buffer_size = 12 * 1024 * 1024 * 1024;
@@ -54,7 +55,7 @@ pub const ExecRequest = extern struct {
             block_idx: BlockRef,
             tx_idx: TransactionPool.ItemId,
             n_account_refs: u8,
-            account_ref_buf: [128]u32,
+            account_ref_buf: [128]accounts_db.AccountPool.Index,
         },
         txn_sig_verify: extern struct {
             tx_idx: TransactionPool.ItemId,
@@ -71,7 +72,7 @@ pub const ExecResponse = extern struct {
             block_idx: BlockRef,
             tx_idx: TransactionPool.ItemId,
             n_account_refs: u8,
-            account_ref_buf: [128]u32,
+            account_ref_buf: [128]accounts_db.AccountPool.Index,
             result: TxExecResult,
         },
         txn_sig_verify: extern struct { success: bool },
--- a/v2/lib/replay.zig
+++ b/v2/lib/replay.zig
@@ -4,7 +4,7 @@ const ipc = @import("ipc.zig");
 const accounts_db = @import("accounts_db.zig");
 
 // This is a bit large currently because of the unrooted store
-pub const scratch_buffer_size = 12 * 1024 * 1024 * 1024;
+pub const scratch_buffer_size = 4 * 1024 * 1024 * 1024;
 
 pub const TransactionPool = collections.SharedPool([1232]u8, 10_000);
 
--- a/v2/lib/replay.zig
+++ b/v2/lib/replay.zig
@@ -2,13 +2,16 @@ const solana = @import("solana.zig");
 const collections = @import("collections.zig");
 const ipc = @import("ipc.zig");
 
+// This is a bit large currently because of the unrooted store
+pub const scratch_buffer_size = 12 * 1024 * 1024 * 1024;
+
 pub const TransactionPool = collections.SharedPool([1232]u8, 10_000);
 
 pub const BlockPool = collections.SharedPool(Node, 1024);
 
 /// NOTE: this is what we use for referencing blocks. This is equivalent to the block's index
 /// our block mem pool. If you want what Agave calls the "Block ID", this is the merkle root of
-///  the last fec set.
+/// the last fec set.
 pub const BlockRef = BlockPool.ItemId;
 
 // TODO: large values (e.g. Hashes) should probably live elsewhere in memory to keep tree
@@ -50,6 +53,8 @@ pub const ExecRequest = extern struct {
         txn_exec: extern struct {
             block_idx: BlockRef,
             tx_idx: TransactionPool.ItemId,
+            n_account_refs: u8,
+            account_ref_buf: [128]u32,
         },
         txn_sig_verify: extern struct {
             tx_idx: TransactionPool.ItemId,
@@ -65,6 +70,8 @@ pub const ExecResponse = extern struct {
         txn_exec: extern struct {
             block_idx: BlockRef,
             tx_idx: TransactionPool.ItemId,
+            n_account_refs: u8,
+            account_ref_buf: [128]u32,
             result: TxExecResult,
         },
         txn_sig_verify: extern struct { success: bool },
--- a/v2/lib/replay.zig
+++ b/v2/lib/replay.zig
@@ -4,7 +4,7 @@ const ipc = @import("ipc.zig");
 const accounts_db = @import("accounts_db.zig");
 
 // This is a bit large currently because of the unrooted store
-pub const scratch_buffer_size = 4 * 1024 * 1024 * 1024;
+pub const scratch_buffer_size = 3 * 1024 * 1024 * 1024;
 
 pub const TransactionPool = collections.SharedPool([1232]u8, 10_000);
 
@@ -55,7 +55,7 @@ pub const ExecRequest = extern struct {
             block_idx: BlockRef,
             tx_idx: TransactionPool.ItemId,
             n_account_refs: u8,
-            account_ref_buf: [128]accounts_db.AccountPool.Index,
+            account_ref_buf: [128]accounts_db.AccountPool.AccountRef,
         },
         txn_sig_verify: extern struct {
             tx_idx: TransactionPool.ItemId,
@@ -72,7 +72,7 @@ pub const ExecResponse = extern struct {
             block_idx: BlockRef,
             tx_idx: TransactionPool.ItemId,
             n_account_refs: u8,
-            account_ref_buf: [128]accounts_db.AccountPool.Index,
+            account_ref_buf: [128]accounts_db.AccountPool.AccountRef,
             result: TxExecResult,
         },
         txn_sig_verify: extern struct { success: bool },
--- a/v2/services/exec.zig
+++ b/v2/services/exec.zig
@@ -26,7 +26,7 @@ pub fn serviceMain(runner: lib.runner.Connection, ro: ReadOnly, rw: ReadWrite) !
     var request_reader = rw.exec_req_response.request_ring.get(.reader);
     var response_writer = rw.exec_req_response.response_ring.get(.writer);
 
-    var deserialised_buf: [4096]u8 = undefined;
+    var deserialised_buf: [16 * 1024]u8 = undefined;
     var deserial_fba: std.heap.FixedBufferAllocator = .init(&deserialised_buf);
 
     while (true) {
--- a/v2/services/exec.zig
+++ b/v2/services/exec.zig
@@ -75,10 +75,14 @@ pub fn serviceMain(runner: lib.runner.Connection, ro: ReadOnly, rw: ReadWrite) !
                             .block_idx = data.block_idx,
                             .tx_idx = data.tx_idx,
                             .n_account_refs = data.n_account_refs,
-                            .account_ref_buf = data.account_ref_buf,
+                            .account_ref_buf = undefined,
                         },
                     },
                 };
+                @memcpy(
+                    response.data.txn_exec.account_ref_buf[0..data.n_account_refs],
+                    data.account_ref_buf[0..data.n_account_refs],
+                );
                 response_writer.markUsed();
             },
             .txn_sig_verify => return error.SigVerifyExecUnimpl,
--- a/v2/services/exec.zig
+++ b/v2/services/exec.zig
@@ -74,6 +74,8 @@ pub fn serviceMain(runner: lib.runner.Connection, ro: ReadOnly, rw: ReadWrite) !
                             .result = .{ .success = true },
                             .block_idx = data.block_idx,
                             .tx_idx = data.tx_idx,
+                            .n_account_refs = data.n_account_refs,
+                            .account_ref_buf = data.account_ref_buf,
                         },
                     },
                 };
--- a/v2/services/replay.zig
+++ b/v2/services/replay.zig
@@ -329,7 +329,7 @@ const Unrooted = struct {
 
         pub fn hash(ctx: MapCtx, key: *const Pubkey) u32 {
             _ = ctx;
-            // TODO: we can do better than this
+            // TODO: we can use a smaller hash without truncating for better performance
             return @truncate(key.hash(123));
         }
 
@@ -366,12 +366,10 @@ const Unrooted = struct {
         return self;
     }
 
-    // TODO:
-    // 1) *never* block the replay thread (remove this function)
-    // 2) introduce a basic transaction scheduler
-    // 3) add a prefetcher
-    // NOTE: caller is responsible for freeing the account
-    fn fetchBlocking(
+    /// Get an account purely from the unrooted store.
+    /// For internal/testing usage only.
+    /// NOTE: caller is responsible for freeing the account
+    fn fetch(
         self: *Unrooted,
         key: *const lib.solana.Pubkey,
 
@@ -381,11 +379,8 @@ const Unrooted = struct {
 
         // account storage
         account_pool: *lib.accounts_db.AccountPool,
-
-        // ring buffer pair for rooted lookups
-        rooted_lookups: *lib.accounts_db.AccountLookups,
     ) AccountRef {
-        const zone = tracy.Zone.init(@src(), .{ .name = "fetchBlocking" });
+        const zone = tracy.Zone.init(@src(), .{ .name = "Unrooted.fetch" });
         defer zone.deinit();
 
         var current = block.ptr(block_pool);
@@ -400,42 +395,73 @@ const Unrooted = struct {
                 const account = account_pool.getAccount(found_account_ref);
                 account.ref();
 
-                zone.text("unrooted");
+                zone.text("found");
                 return found_account_ref;
             }
 
             current = ancestor_block.parent.ptr(block_pool);
         }
 
-        var requester = rooted_lookups.in.get(.writer);
-        var response_queue = rooted_lookups.out.get(.reader);
-
-        const request_buf = requester.next() orelse @panic("out of space");
-        request_buf.* = key.*;
-        requester.markUsed();
-
-        // blocking the thread - do not do this
-        while (response_queue.peek() == null) : (std.atomic.spinLoopHint()) {}
-
-        const response = response_queue.next().?;
-        defer response_queue.markUsed();
-
-        std.debug.assert(response.pubkey.equals(key));
-
-        if (response.account_index == lib.accounts_db.AccountPool.invalid_index) {
-            zone.text("account not found");
-            return response.account_index;
-        }
-
-        const account = account_pool.getAccount(response.account_index);
-
-        std.debug.assert(account.ref_count.load(.monotonic) > 0);
-        std.debug.assert(account.pubkey.equals(key));
-
-        zone.text("rooted");
-        return response.account_index;
+        return lib.accounts_db.AccountPool.invalid_index;
     }
 };
+// TODO:
+// 1) *never* block the replay thread (remove this function)
+// 2) introduce a basic transaction scheduler
+// 3) add a prefetcher
+/// Gets an account, trying the unrooted store before asking rooted.
+/// NOTE: caller is responsible for freeing the account
+fn fetchBlocking(
+    unrooted: *Unrooted,
+    key: *const lib.solana.Pubkey,
+
+    // current block + pool for ancestor lookups
+    block: lib.replay.BlockRef,
+    block_pool: *lib.replay.BlockPool,
+
+    // account storage
+    account_pool: *lib.accounts_db.AccountPool,
+
+    // ring buffer pair for rooted lookups
+    rooted_lookups: *lib.accounts_db.AccountLookups,
+) AccountRef {
+    const zone = tracy.Zone.init(@src(), .{ .name = "fetchBlocking" });
+    defer zone.deinit();
+
+    const unrooted_account = unrooted.fetch(key, block, block_pool, account_pool);
+    if (unrooted_account != lib.accounts_db.AccountPool.invalid_index) {
+        zone.text("unrooted");
+        return unrooted_account;
+    }
+
+    var requester = rooted_lookups.in.get(.writer);
+    var response_queue = rooted_lookups.out.get(.reader);
+
+    const request_buf = requester.next() orelse @panic("out of space");
+    request_buf.* = key.*;
+    requester.markUsed();
+
+    // blocking the thread - do not do this
+    while (response_queue.peek() == null) : (std.atomic.spinLoopHint()) {}
+
+    const response = response_queue.next().?;
+    defer response_queue.markUsed();
+
+    std.debug.assert(response.pubkey.equals(key));
+
+    if (response.account_index == lib.accounts_db.AccountPool.invalid_index) {
+        zone.text("account not found");
+        return response.account_index;
+    }
+
+    const account = account_pool.getAccount(response.account_index);
+
+    std.debug.assert(account.ref_count.load(.monotonic) > 0);
+    std.debug.assert(account.pubkey.equals(key));
+
+    zone.text("rooted");
+    return response.account_index;
+}
 
 /// Finds a node's parent, and attaches the new node to it.
 ///
@@ -858,9 +884,10 @@ fn maybeContinueBlockExec(
                 inline else => |txn| txn.account_keys.items,
             };
 
-            for (account_keys) |k| {
-                held_accounts_buf[held_accounts] = unrooted.fetchBlocking(
-                    &k,
+            for (account_keys) |*k| {
+                held_accounts_buf[held_accounts] = fetchBlocking(
+                    unrooted,
+                    k,
                     block_ref,
                     block_pool,
                     account_pool,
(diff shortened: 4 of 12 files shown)


v2/lib/replay.zig
```diff
@@ -50,6 +53,8 @@ pub const ExecRequest = extern struct {
         txn_exec: extern struct {
             block_idx: BlockRef,
             tx_idx: TransactionPool.ItemId,
+            n_account_refs: u8,
+            account_ref_buf: [128]u32,
```
kprotty: `[128]AccountPool.Index`

v2/services/replay.zig
```diff
@@ -164,6 +170,13 @@ pub fn serviceMain(runner: lib.runner.Connection, _: ReadOnly, rw: ReadWrite) !n
             std.debug.assert(response.request_kind == .txn_exec); // others unimplemented
             const response_data = response.data.txn_exec;
 
+            for (response_data.account_ref_buf[0..response_data.n_account_refs]) |account_ref| {
+                if (account_ref == lib.accounts_db.AccountPool.invalid_index) continue;
```
kprotty: hm, wonder if this should be a type-safe Id value like your collections Pool data structure (so you can do `ref == .invalid`)

v2/services/replay.zig
```diff
@@ -276,13 +289,133 @@ pub fn serviceMain(runner: lib.runner.Connection, _: ReadOnly, rw: ReadWrite) !n
...
+        pub fn eql(ctx: MapCtx, key_a: *const Pubkey, _: void, key_b_idx: usize) bool {
+            const b_account_ref = ctx.map.values()[key_b_idx];
+            return key_a.equals(&ctx.pool.getAccount(b_account_ref).pubkey);
+        }
+    };
+
+    // TODO: we'll want to share this map eventually, we should make our own extern fixed-size map.
+    // NOTE: this is where almost all of replay's memory usage comes from, this could be improved
+    // with a better map implementation.
+    const Map = std.ArrayHashMapUnmanaged(void, AccountRef, MapCtx, true);
```
kprotty: use a custom hashmap instead, sized to the max account entries per slot. e.g.
```zig
type Map:
  array: [N]AccountPool.Index = @splat(.invalid)
  get(pool, pk) ?AccountPool.Index:
    i = pk.hash(0) % N // Pubkey has a hash method btw
    loop: // TODO: handle when full
      if array[i] == .invalid: return null
      if pk.eql(pool.getAccount(array[i]).pubkey): return array[i]
      i = (i + 1) % N // map doesnt need to delete iirc, so no tombstones
```

v2/services/replay.zig
```diff
@@ -276,13 +289,133 @@ pub fn serviceMain(runner: lib.runner.Connection, _: ReadOnly, rw: ReadWrite) !n
...
+
+        var requester = rooted_lookups.in.get(.writer);
+        var response_queue = rooted_lookups.out.get(.reader);
+
+        const request_buf = requester.next() orelse @panic("out of space");
+        request_buf.* = key.*;
+        requester.markUsed();
+
+        // blocking the thread - do not do this
+        while (response_queue.peek() == null) {}
```
kprotty: spinLoopHint

v2/services/replay.zig
```diff
@@ -276,13 +289,133 @@ pub fn serviceMain(runner: lib.runner.Connection, _: ReadOnly, rw: ReadWrite) !n
...
+        std.debug.assert(response.pubkey.equals(key));
+
+        if (response.account_index == lib.accounts_db.AccountPool.invalid_index) {
+            zone.text("account not found");
+            return response.account_index;
+        }
+
+        const account = account_pool.getAccount(response.account_index);
+
+        std.debug.assert(account.ref_count.load(.seq_cst) > 0);
```
kprotty: can be monotonic or even unordered

v2/services/replay.zig
```diff
@@ -674,23 +816,111 @@ fn maybeContinueBlockExec(
...
+                    if (account_ref == lib.accounts_db.AccountPool.invalid_index)
+                        @panic("missing address lookup table / TODO: handle bad blocks");
+
+                    const ALT_account: *lib.accounts_db.AccountPool.Account =
+                        account_pool.getAccount(account_ref);
+
+                    defer if (ALT_account.unref()) account_pool.free(account_ref);
+
+                    // NOTE: this is *not* going to be conformant; TODO: respect the ALT metadata
+                    const ALT_data = ALT_account.getData();
```
kprotty, line 984: ALT man

v2/services/replay.zig
Sobeston: Yeah I think we should do it, want me to do it in this PR?

v2/services/replay.zig
Sobeston: https://github.com/Syndica/sig/pull/1617/commits/9a6404c87bbbb761abfe77e7cd33eb6b60e71b86 changed to monotonic

v2/services/replay.zig
Sobeston: https://github.com/Syndica/sig/pull/1617/commits/9a6404c87bbbb761abfe77e7cd33eb6b60e71b86 added it

v2/lib/replay.zig
Sobeston: https://github.com/Syndica/sig/pull/1617/commits/9a6404c87bbbb761abfe77e7cd33eb6b60e71b86 changed

v2/services/replay.zig
Sobeston: I would prefer to do this in a follow-up; agree we should do it though

v2/services/replay.zig
Sobeston: ALTernative implementation

v2/services/replay.zig
kprotty: yea

v2/services/replay.zig
kprotty: seems small enough to do here (just `get` and `put`/`getOrPut` - whichever is needed atm). Avoids having to do the scratch_memory stuff

v2/init/main.zig
```diff
@@ -215,6 +217,8 @@ pub const topology = lib.topology.Bind(topology_schema, Region, .init(.{
     .accounts_db_config = .initOne(.@"accounts_db:config"),
     .snapshot_config = .initOne(.@"snapshot:config"),
 
+    .replay_scratch_memory = .initOne(.@"replay:scratch_memory"),
```
dnut: Why does this need to be a shared region?

The topology describes the high-level architecture and interaction between services. Using it to also reflect the private state of each service clutters the high-level code with service-internal implementation details that aren't relevant to the architecture.

`gossip` and `shred_receiver` already follow the pattern of keeping private memory defined within the service itself, so this also introduces an inconsistency.

Are you planning to make use of this region in another service?

v2/services/replay.zig
```diff
@@ -146,7 +152,7 @@ pub fn serviceMain(runner: lib.runner.Connection, _: ReadOnly, rw: ReadWrite) !n
             const zone = tracy.Zone.init(@src(), .{ .name = "idle" });
             defer zone.deinit();
 
-            while (true) {
+            while (true) : (std.atomic.spinLoopHint()) {
```
dnut, line 174: what if we start using cpu affinity to give each service its own core? do you think this spin loop hints serve any benefit at that point? maybe it would release the cpu for usage by kernel or other os threads. i wonder if that could be harmful due to context switching? probably a non-issue since we need to wait here anyway. just curious on your thoughts.

v2/services/replay.zig
```diff
@@ -164,6 +170,13 @@ pub fn serviceMain(runner: lib.runner.Connection, _: ReadOnly, rw: ReadWrite) !n
             std.debug.assert(response.request_kind == .txn_exec); // others unimplemented
             const response_data = response.data.txn_exec;
 
+            for (response_data.account_ref_buf[0..response_data.n_account_refs]) |account_ref| {
+                if (account_ref == lib.accounts_db.AccountPool.invalid_index) continue;
+
+                const account = rw.account_pool.getAccount(account_ref);
+                if (account.unref()) rw.account_pool.free(account_ref);
```
dnut, line 198: how do we guarantee that this doesn't race with another thread?

1. thread A unref() -> true
2. thread B ref()
3. thread A free()
4. thread B dereferences pointer

What other services could potentially be thread B? if it's only replay, this race can't happen. what about the service in charge of flushing accounts when a slot is finalized? what about a block production service?

v2/services/replay.zig
```diff
@@ -276,13 +289,133 @@ pub fn serviceMain(runner: lib.runner.Connection, _: ReadOnly, rw: ReadWrite) !n
...
+/// Holds the accounts mutated for each tracked Block.
+const Unrooted = struct {
+    const MapCtx = struct {
+        pool: *lib.accounts_db.AccountPool,
+        map: *const Map,
+
+        pub fn hash(ctx: MapCtx, key: *const Pubkey) u32 {
+            _ = ctx;
+            // TODO: we can do better than this
+            return @truncate(key.hash(123));
```
dnut: how should this be improved? it seems ok to me. this comment could use some more specificity on your intent, so future readers know what to do.

v2/services/replay.zig
```diff
@@ -276,13 +289,133 @@ pub fn serviceMain(runner: lib.runner.Connection, _: ReadOnly, rw: ReadWrite) !n
...
+        var self: Unrooted = undefined;
+
+        for (&self.maps, 0..) |*map, i| {
+            map.* = .empty;
+
+            errdefer {
+                for (self.maps[0 .. i + 1]) |*deinit_map| deinit_map.deinit(allocator);
+            }
+
+            try map.ensureTotalCapacity(allocator, max_mutations_per_block);
```
dnut: feel free to resolve this comment. you don't need to change anything. just feels a little too spaced out for my taste...
```suggestion
            map.* = .empty;
            errdefer for (self.maps[0 .. i + 1]) |*deinit_map| deinit_map.deinit(allocator);
            try map.ensureTotalCapacity(allocator, max_mutations_per_block);
```

v2/services/replay.zig
```diff
@@ -674,23 +816,111 @@ fn maybeContinueBlockExec(
...
+
+                    const indexes = switch (pass) {
+                        .write => lookup.writable_indexes.items,
+                        .read => lookup.readonly_indexes.items,
+                    };
+
+                    for (indexes) |account_idx| {
+                        if (account_idx > ALT_pubkeys.len)
+                            @panic("bad ALT lookup / TODO: handle bad blocks");
+                        const account_pk: *const Pubkey = &ALT_pubkeys[account_idx];
```
dnut: off by one?

```suggestion
                        if (account_idx >= ALT_pubkeys.len)
                            @panic("bad ALT lookup / TODO: handle bad blocks");
                        const account_pk: *const Pubkey = &ALT_pubkeys[account_idx];
```

we don't have any way to deal with an invalid block? what happens when sigverify fails?

v2/services/replay.zig
```diff
@@ -674,23 +816,111 @@ fn maybeContinueBlockExec(
...
+
+                        held_accounts_buf[held_accounts] = unrooted.fetchBlocking(
+                            account_pk,
+                            block_ref,
+                            block_pool,
+                            account_pool,
+                            rooted_lookups,
+                        );
+                        held_accounts += 1;
```
dnut: what guarantees that this won't go out of bounds? you can fit more than 128 accounts in a 1232 byte transaction using lookup tables.

v2/services/replay.zig
```diff
@@ -674,23 +816,111 @@ fn maybeContinueBlockExec(
...
+            // TODO: replace this with something custom, this is slow - we only need to extract the
+            // accounts (including ALT accounts) here.
+            var deserialised_buf: [4096]u8 = undefined;
+            var deserial_fba: std.heap.FixedBufferAllocator = .init(&deserialised_buf);
+            var reader = std.io.Reader.fixed(tx);
+            const transaction: lib.solana.transaction.VersionedTransaction =
+                try lib.solana.bincode.read(
+                    &deserial_fba,
+                    &reader,
+                    lib.solana.transaction.VersionedTransaction,
```
dnut: how did you come up with 4096 here? are you sure that can handle the worst case?

v2/services/replay.zig
```diff
@@ -674,23 +816,111 @@ fn maybeContinueBlockExec(
...
+            const address_lookups: []const lib.solana.transaction.AddressLookup =
+                switch (transaction.message) {
+                    .legacy => &.{},
+                    .v0 => |v0| v0.address_table_lookups.items,
+                };
+
+            const Pass = enum { read, write };
+
+            // looked up accounts are writable first, then readable
+            for (@as([]const Pass, &.{ .write, .read })) |pass| {
```
dnut: consistency nitpick

```suggestion
            const Pass = enum { write, read };

            // looked up accounts are writable first, then readable
            for (@as([]const Pass, &.{ .write, .read })) |pass| {
```

v2/services/replay.zig
```diff
@@ -674,23 +816,111 @@ fn maybeContinueBlockExec(
...
+            var held_accounts_buf: [128]AccountRef = undefined;
+            var held_accounts: u8 = 0;
+
+            const account_keys: []const Pubkey = switch (transaction.message) {
+                inline else => |txn| txn.account_keys.items,
+            };
+
+            for (account_keys) |k| {
+                held_accounts_buf[held_accounts] = unrooted.fetchBlocking(
+                    &k,
```
dnut: not sure how much this makes a difference but in my mind this seems a little more likely to avoid copies:
```suggestion
            for (&account_keys) |*k| {
                held_accounts_buf[held_accounts] = unrooted.fetchBlocking(
                    k,
```

v2/services/replay.zig
```diff
@@ -674,23 +816,111 @@ fn maybeContinueBlockExec(
...
+
+                    if (account_ref == lib.accounts_db.AccountPool.invalid_index)
+                        @panic("missing address lookup table / TODO: handle bad blocks");
+
+                    const ALT_account: *lib.accounts_db.AccountPool.Account =
+                        account_pool.getAccount(account_ref);
+
+                    defer if (ALT_account.unref()) account_pool.free(account_ref);
+
+                    // NOTE: this is *not* going to be conformant; TODO: respect the ALT metadata
```
dnut: can you clarify this comment?

v2/services/replay.zig
```diff
@@ -276,13 +289,133 @@ pub fn serviceMain(runner: lib.runner.Connection, _: ReadOnly, rw: ReadWrite) !n
...
+            return response.account_index;
+        }
+
+        const account = account_pool.getAccount(response.account_index);
+
+        std.debug.assert(account.ref_count.load(.monotonic) > 0);
+        std.debug.assert(account.pubkey.equals(key));
+
+        zone.text("rooted");
+        return response.account_index;
```
dnut: It seems wrong to read from the rooted store in a method on the unrooted store. It's unexpected because it doesn't follow the pattern implied by the name and the contents of this struct.

I think it makes more sense how we had it in v1: unrooted has a method that checks in unrooted only, and then another function that calls unrooted first and then rooted second.

v2/services/exec.zig
```diff
@@ -75,6 +75,8 @@ pub fn serviceMain(runner: lib.runner.Connection, ro: ReadOnly, rw: ReadWrite) !
                             .result = .{ .success = true },
                             .block_idx = data.block_idx,
                             .tx_idx = data.tx_idx,
+                            .n_account_refs = data.n_account_refs,
+                            .account_ref_buf = data.account_ref_buf,
                         },
                     },
                 };
```
dnut, line 81: this copies undefined data. not a big deal i guess, but it's unnecessary
```suggestion
                response.* = .{
                    .task_id = request.task_id,
                    .request_kind = .txn_exec,
                    .data = .{
                        .txn_exec = .{
                            .result = .{ .success = true },
                            .block_idx = data.block_idx,
                            .tx_idx = data.tx_idx,
                            .n_account_refs = data.n_account_refs,
                            .account_ref_buf = undefined,
                        },
                    },
                };
                @memcpy(
                    response.data.txn_exec.account_ref_buf[0..data.n_account_refs],
                    data.account_ref_buf[0..data.n_account_refs],
                );
```

v2/services/replay.zig
```diff
@@ -674,23 +816,111 @@ fn maybeContinueBlockExec(
...
-            defer exec_request_sender.markUsed();
             request.* = .{
                 .task_id = tx_index,
                 .request_kind = .txn_exec,
                 .data = .{
                     .txn_exec = .{
                         .block_idx = block_ref,
                         .tx_idx = tx_ref,
+                        .n_account_refs = held_accounts,
+                        .account_ref_buf = held_accounts_buf,
```
dnut: same as prior comment

dnut (changes requested):

v2/services/replay.zig
Sobeston: As far as I understand, this will not cause more frequent context switches or the thread to be released.

This just uses the `PAUSE` instruction, which in theory reduces power consumption a lot when spinning (+ probably less useless core-to-core memory traffic); I don't think the OS is aware of us using it.

v2/services/replay.zig
Sobeston: Implemented a basic version of this, but realised a bit of a problem, is it perfectly safe to use linear probing open-addressing hashmaps?

What level of oversizing would allow us to be extremely confident that it wouldn't fail

Another issue is that someone could maliciously write to a lot of accounts whose pubkeys hash to the same (or similar) value, causing probing to start to fail. I guess here we could generate a random seed on startup to mitigate it

v2/init/main.zig
Sobeston: The issue is that at the moment this is how we do all memory allocation. We can't just create a 12GiB global variable, it makes lld very unhappy

kprotty: similar linker limit was hit for the rooted table memory

v2/services/replay.zig
kprotty: if thread B has a ref to it, its ref-count should be 2, making thread A's unref bring it to 1 & not return true.

Should be aware of all the spots an account is stored & its lifecycle. Id imagine the account would be ref'd when stored into unrooted in the future.

v2/services/replay.zig
kprotty: Its safe. Hashing to same spot means probing takes a bit longer, not that it fails. The amount of accounts per slot in unrooted shouldnt realistically be more than a few hundred thousand at worst. At those levels, the more important question is memory allocation rather than probing (which is fast, see memory-level-parallelism strat in table.zig). I presume the lookup also only occurs during account fetch?

The rooted table also does linear probing and takes a hash seed at runtime. We should actually hook it up to a random source for sure.

v2/services/replay.zig
Sobeston: I've split them out https://github.com/Syndica/sig/pull/1617/commits/0f75bea6076c37eb52c56657a7cf08fec7fbe8a4

v2/services/replay.zig
Sobeston: clarified https://github.com/Syndica/sig/pull/1617/commits/0f75bea6076c37eb52c56657a7cf08fec7fbe8a4

v2/services/replay.zig
Sobeston: https://github.com/Syndica/sig/pull/1617/commits/0f75bea6076c37eb52c56657a7cf08fec7fbe8a4 changed

v2/services/replay.zig
Sobeston: changed https://github.com/Syndica/sig/pull/1617/commits/0f75bea6076c37eb52c56657a7cf08fec7fbe8a4

v2/services/replay.zig
Sobeston: Guessing, really we should do something else. I've bumped it to 16KiB for now.

v2/services/replay.zig
Sobeston: Added a panic https://github.com/Syndica/sig/pull/1617/commits/0f75bea6076c37eb52c56657a7cf08fec7fbe8a4

v2/services/replay.zig
Sobeston: Fixed off by one.

Yeah we don't have a way of getting rid of invalid blocks just yet, it needs some extra machinery. Of course there's many reasons which could trigger it.

v2/services/replay.zig
Sobeston: Replaced this all with the new map

v2/services/replay.zig
Sobeston: Yeah my assumption is that unrooted will store refs to accounts, and we also hold refs across exec.

v2/init/main.zig
Sobeston: with the new map I've lowered this to 3GiB instead of 12, even with some oversizing of the map, the normal arrayhashmap is pretty big.

(I think 2GiB may be the point where lld complains)

Sobeston: We could maybe just give every service its own "scratch" region implicitly, with some configurable size. Maybe that's the way to go.

v2/services/replay.zig
Sobeston: Yeah I overthought this, the new map is in

v2/services/replay.zig
Sobeston: done in https://github.com/Syndica/sig/pull/1617/commits/8c90083dca27f0e79f79cdcd7f5c9427d816ee7b

v2/services/exec.zig
Sobeston: Yeah we always send the whole buffer, even if it's likely that most of the buffer isn't used. This isn't ideal, but I want to have the data in the request message which are fixed size. Maybe we can find a more efficient way but I'm not worried about this

v2/services/replay.zig
Sobeston: same as above

v2/services/replay.zig
```diff
@@ -273,13 +286,212 @@ pub fn serviceMain(runner: lib.runner.Connection, _: ReadOnly, rw: ReadWrite) !n
...
+    const max_mutations_per_block = 367_535;
+
+    const max_blocks = lib.replay.BlockPool.capacity;
+
+    // 2MiB entries + 4B len
+    const Map = extern struct {
+        len: u32 = 0, // only used to assert `max_mutations_per_block` holds true
+        data: [N]AccountRef = @splat(.invalid), // 2MiB
+
+        const N = std.math.ceilPowerOfTwo(usize, max_mutations_per_block) catch unreachable;
```
kprotty: nit: find to be not a pow2 - just use `% N` instead of `& (N - 1)`. Compiler will turn it into muls/adds + shifts when comptime known.

v2/services/replay.zig
```diff
@@ -273,13 +286,212 @@ pub fn serviceMain(runner: lib.runner.Connection, _: ReadOnly, rw: ReadWrite) !n
...
+            new_account_ref: AccountRef,
+        ) AccountRef {
+            const zone = tracy.Zone.init(@src(), .{ .name = "Map.put" });
+            defer zone.deinit();
+
+            std.debug.assert(new_account_ref != .invalid);
+            const new_account = account_pool.getAccount(new_account_ref);
+            const pubkey: *const Pubkey = &new_account.pubkey;
+
+            const found_entry: *u32 = self.entry(seed, account_pool, pubkey);
```
kprotty: `*AccountRef`

also, when does a "replace" occur?

v2/services/replay.zig
```diff
@@ -273,13 +286,212 @@ pub fn serviceMain(runner: lib.runner.Connection, _: ReadOnly, rw: ReadWrite) !n
...
+        account_pool: *lib.accounts_db.AccountPool,
+    ) AccountRef {
+        const zone = tracy.Zone.init(@src(), .{ .name = "Unrooted.fetch" });
+        defer zone.deinit();
+
+        // TODO:
+        // 1) create randomly + secretly at startup, to avoid performance degradation from attackers
+        //    using pre-made keys to cause bad clustering
+        // 2) change the seed used per block to avoid possibility of worst-case clustering
+        const seed = 123;
```
kprotty: seed should be stored in Unrooted struct, passed at `init()`

v2/services/exec.zig
dnut: what's wrong with my suggestion? it solves both:
- doesn't send the whole buffer
- message is still fixed size

v2/services/replay.zig
Sobeston: https://github.com/Syndica/sig/pull/1617/commits/e38f249b687dcddf61310bfa438c62db8ca776a4

v2/services/replay.zig
Sobeston: removed the pow2 oversize https://github.com/Syndica/sig/pull/1617/commits/e38f249b687dcddf61310bfa438c62db8ca776a4

v2/services/replay.zig
Sobeston: added a note https://github.com/Syndica/sig/pull/1617/commits/e38f249b687dcddf61310bfa438c62db8ca776a4

v2/services/replay.zig
dnut: > if thread B has a ref to it, its ref-count should be 2

In my example, thread B attempts to acquire a reference *after* thread A releases its reference. The whole point is that thread B does *not* have a reference to it yet.

I think the point that both of you are making is that replay is the only thread that would increment or decrement reference counts (at least in the current implementation), so there is no possibility of a race. In other words, thread A and thread B are actually both the same thread, so it's not possible to interleave the steps in this way. (this is what I meant by "if it's only replay, this race can't happen.")

---

There is a real race that can happen in reference counting implementations if you're not careful. It can happen when multiple threads share access to a pointer that is reference counted by an integer behind the pointer, and each thread only increments the reference counter *after* it already has access to the pointer being reference counted.

This is common pattern when multiple threads all share access to a single map that contains a bunch of reference counted pointers. You may have 2 threads that can access the map but you'll keep the refcount only at 1 while the items are in the map, and each thread increments the refcount for a specific item only when it calls `get` on the map.

The solution is that you need to synchronize reads and writes from that data structure to ensure `get+increment` and `decrement+remove+free` operations each occur atomically, and cannot overlap with each other. In other words you need to disallow interleaving of operations like:
1. thread B get
2. thread A decrement+remove+free
3. thread B increment

Step 3 has undefined behavior in this example. A lock can solve this problem because it can ensure that no part of step 2 ever happens between 1 and 3.

---

Now getting back to our specific example here. As long as the replay thread is the only one touching the unrooted map, and it is in charge of all increments and removals from the map, there is no problem. But my point was that you may also have a leader thread reading from the map and incrementing the counter, or an account flushing thread that's also removing items from the map. If this happens, we'll need to lock accesses to the map.

As I've put some more thought into this, I realize this is a specific consideration about how to synchronize the map, and not as much a general problem with reference counting. So I think it's fine to resolve this conversation now. Of course if we start sharing the map with multiple threads, we'll need to think about how to make it thread safe, and the reference counts will naturally be a part of that thought process.

v2/services/replay.zig
dnut: in this case i think it's fine to have a panic because we basically have to rewrite the whole account fetching logic very soon regardless. so i'm satisfied resolving this concern with it as is.

typically though, i would prefer to figure out error handling before merging prs rather than leaving todos. i'd rather have one feature fully implemented than two features halfway implemented. when you're writing the code is when you have the best context about how the error should be handled. later on when someone comes back to address the todo, it's going to take a lot more time to get familiar with the context and implement it correctly. it's also easy to neglect all these cases and we end up with large swaths of code that are unpolished. rather than having a clear sense of "this code is done, this code isn't" it's more like "nothing is done." often, code that you thought was only a prototype ends up being what you go with in production and then it's frustrating when it's littered with technical debt

v2/init/main.zig
dnut: https://github.com/Syndica/sig/issues/1691

v2/services/replay.zig
Sobeston: Yeah a lot of this code is a pain to get right upfront, an eviction system should come before dealing with bad blocks

v2/services/exec.zig
Sobeston: Your suggestion might reduce the amount of memory copied into the buffer, but we're still copying the full size into the buffer and out again. Applying the change is fine, I just don't think it makes much of a difference

Sobeston: applied https://github.com/Syndica/sig/pull/1617/commits/20555c198d5c26fcfbb90a75809955e1895ec575

v2/services/replay.zig
Sobeston: applied dcb0a245315b1f5e2366feb103b0922688eef0f7

v2/services/exec.zig
dnut: i don't understand. it can reduce the amount copied while still copying the full size. that sounds like a contradiction. what's the distinction that i'm missing?

dnut (dismissed):
feel free to resolve my other comments

v2/services/replay.zig
kprotty: meant rather to store it outside the Map & pass it into map operations, instead of duplicating it for each map. But this is more of a nitpick.

Sobeston: Oh right, I do think that each map having its own seed is probably a good move going forward though?

Sobeston: https://github.com/Syndica/sig/pull/1617/commits/9625aa67f03c43a08ff3d534ce6135f66789d2de

kprotty (dismissed):

dnut (approved):

kprotty (approved):

v2/services/exec.zig
Sobeston: Sorry I think I misunderstood your message earlier, yes this does avoid copying the full size, all is good

PR #1619, opened 2026-06-22
feat(conformance): add vm_serialization harness

yewman:
Implements sol_compat_vm_serialize_execute_v1, mirroring agave's and solfuzz-agave's harness. Decodes an InstrContext, pushes a top-level instruction frame, runs sig's bpf serializeParameters, and emits the serialized memory's XxHash64 along with the input memory regions and per-account VM addresses.

Regenerates v1.pb.zig from protosol to pull in the VMInputMemoryRegion / VMSerializedAccountMetadata / VMSerializationEffects messages, and drops vm_serialization from unimplemented_harnesses.txt.

All 1000 vm_serialization test-vector fixtures pass.

Fuzzed locally on mixed instr corpus.

### DIFF
--- a/conformance/src/vm_serialize.zig
+++ b/conformance/src/vm_serialize.zig
@@ -23,8 +23,6 @@ pub export fn sol_compat_vm_serialize_execute_v1(
     in_ptr: [*]const u8,
     in_size: u64,
 ) i32 {
-    errdefer |err| std.debug.panic("err: {s}", .{@errorName(err)});
-
     var arena = std.heap.ArenaAllocator.init(std.heap.c_allocator);
     defer arena.deinit();
     const allocator = arena.allocator();
@@ -42,7 +40,10 @@ pub export fn sol_compat_vm_serialize_execute_v1(
     };
 
     var writer: std.Io.Writer.Allocating = .init(allocator);
-    try effects.encode(&writer.writer, allocator);
+    effects.encode(&writer.writer, allocator) catch |err| {
+        std.debug.print("effects.encode: {s}\n", .{@errorName(err)});
+        return 0;
+    };
     const result_bytes = writer.written();
 
     const out_slice = out_ptr[0..out_size.*];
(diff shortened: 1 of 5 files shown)


conformance/src/vm_serialize.zig
```diff
@@ -0,0 +1,140 @@
...
+///
+/// [agave] https://github.com/anza-xyz/agave/blob/master/svm/src/conformance/serialization.rs
+/// [solfuzz-agave] https://github.com/firedancer-io/solfuzz-agave/blob/agave-v4.1.0-beta.3/src/vm_serialization.rs
+pub export fn sol_compat_vm_serialize_execute_v1(
+    out_ptr: [*]u8,
+    out_size: *u64,
+    in_ptr: [*]const u8,
+    in_size: u64,
+) i32 {
+    errdefer |err| std.debug.panic("err: {s}", .{@errorName(err)});
```
InKryption: This errdefer is dead code, it can never be hit in a function that can't return an error (and this function will never be changed to return an error in the future, because it's exposed on the ABI boundary).

conformance/src/vm_serialize.zig
```diff
@@ -0,0 +1,140 @@
...
+        &tc,
+        program_id,
+        pb_instr_ctx.data,
+        pb_instr_ctx.instr_accounts.items,
+    );
+    defer instr_info.deinit(allocator);
+
+    // serializeParameters needs an active instruction frame so it can borrow
+    // the program account to decide between the aligned/unaligned layouts.
+    executor.pushInstruction(&tc, instr_info) catch unreachable;
```
InKryption: Is this actually unreachable? If it could occur by means of bad input, should probably be a panic instead. Otherwise a brief comment alluding to why it's unreachable would be good.

InKryption (commented):
Just the one comment, otherwise lgtm

conformance/src/vm_serialize.zig
yewman: Woops 🙃  6086afbc9384b1204dbd11b7b680bf420d75067a

conformance/src/vm_serialize.zig
yewman: Yeah good call 👍 
6086afbc9384b1204dbd11b7b680bf420d75067a

hamza-syndica (approved):

InKryption (approved):

PR #1620, opened 2026-06-22
feat(v2:shred_rx): add basic unit tests for processPacket

prestonsn:
Builds on top of the new `Effects` interface for v2's `shred_receiver` and adds some basic unit tests (based on what we have in v2, and what firedancer and agave have as basic smoke tests). Just the first of many more: 

- `shred.receiver: empty packet`: verifies malformed packets emit parse failure and failed packet effects.
- `shred.receiver: shred version mismatch`: verifies valid-layout packets can parse successfully but fail receiver validation.
- `shred.receiver: one shred (unfinished fec set)`: verifies a signed valid shred is acepted as an unfinished FEC set
- `shred.receiver: duplicate shred`: verifies reprocessing the same valid shred reports `shred_already_seen`.

### DIFF
--- a/v2/lib/shred/receiver.zig
+++ b/v2/lib/shred/receiver.zig
@@ -677,6 +677,275 @@ const InProgressSets = struct {
     };
 };
 
+const ReceiverTestEffects = struct {
+    parse_results: [8]bool = undefined,
+    parse_result_count: usize = 0,
+
+    packet_results: [8]PacketResult = undefined,
+    packet_result_count: usize = 0,
+
+    fec_completed_count: usize = 0,
+
+    pub fn reportShredParseResult(
+        self: *ReceiverTestEffects,
+        parses_as_chained: bool,
+    ) void {
+        self.parse_results[self.parse_result_count] = parses_as_chained;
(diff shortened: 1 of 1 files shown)


Sobeston (dismissed):
Nice!

v2/lib/shred/receiver.zig
```diff
@@ -783,6 +783,275 @@ const InProgressSets = struct {
...
+                .flags = .{
+                    .reference_tick = 0,
+                    .data_complete = false,
+                    .last_shred_in_slot = false,
+                },
+                .size = @offsetOf(Shred, "code_or_data") + @sizeOf(Shred.DataHeader),
+            },
+        },
+    };
+    return shred;
```
kprotty: the return value doesnt seem to be used, but otherwise looks good

prestonsn: good catch! removed it: https://github.com/Syndica/sig/pull/1620/commits/1ee42e0e344f93d889439da805d1ecabc05709d9

Sobeston (approved):

dnut (approved):

PR #1622, opened 2026-06-23
fix(v2:build): wire unit-test build step

prestonsn:
v2's `zig build unit-test` was basically inert and did nothing (running no tests), while `zig build test` ran all the unit tests as well as the black box tests. 

- This PR wires up `zig build unit-test` such that it now runs the unit tests (and nothing more)
- Allows for `zig build unit-test -Dkcov` to run unit tests under kcov if specified 
- Keeps `zig build test` as the main "umbrella" step that runs unit tests and black box tests (unchanged)

### DIFF
--- a/v2/build.zig
+++ b/v2/build.zig
@@ -112,7 +112,7 @@ pub fn build(b: *Build) !void {
             .install_subdir = "kcov",
         });
         install_dir.step.dependOn(&run.step);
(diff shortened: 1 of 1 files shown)


InKryption (approved):
Whoops, nice catch.

dnut (approved):

PR #1623, opened 2026-06-23
fix(replay): use fee_rate_governor burn_percent for tx fees

ajw221:
# Intent

- Fix current OOM crashes when running sig

# Implementation

- Updated `distributeTransactionFees` to calculate `burn` using hardcoded `DEFAULT_BURN_PERCENT`
- Updated all references where `DEFAULT_BURN_PERCENT`'s value could have been used instead of a magic value
- Updated `UiRent` to more appropriately match `v4.1` RPC API structure

# Ramifications

- Currently not possible to run the validator without immediately running into slot hash mismatches

# Tests

- Ran the validator and caught up to the tip of the chain

### DIFF
--- a/src/replay/epoch_transitions.zig
+++ b/src/replay/epoch_transitions.zig
@@ -1086,7 +1086,10 @@ test "applyFeatureActivations: basic activations" {
             sig.core.genesis_config.Inflation.PICO,
             env.slot_constants.inflation,
         );
-        try std.testing.expectEqual(sig.runtime.sysvar.DEFAULT_BURN_PERCENT, env.slot_constants.fee_rate_governor.burn_percent);
+        try std.testing.expectEqual(
+            sig.runtime.sysvar.DEFAULT_BURN_PERCENT,
+            env.slot_constants.fee_rate_governor.burn_percent,
+        );
         try std.testing.expectEqual(
             sig.runtime.sysvar.DEFAULT_BURN_PERCENT,
             env.slot_constants.rent_collector.rent.burn_percent,
@@ -1158,7 +1161,10 @@ test "applyFeatureActivations: basic activations" {
             sig.core.genesis_config.Inflation.FULL,
             env.slot_constants.inflation,
         );
-        try std.testing.expectEqual(sig.runtime.sysvar.DEFAULT_BURN_PERCENT, env.slot_constants.fee_rate_governor.burn_percent);
+        try std.testing.expectEqual(
+            sig.runtime.sysvar.DEFAULT_BURN_PERCENT,
+            env.slot_constants.fee_rate_governor.burn_percent,
+        );
         try std.testing.expectEqual(
             sig.runtime.sysvar.DEFAULT_BURN_PERCENT,
             env.slot_constants.rent_collector.rent.burn_percent,
--- a/src/replay/epoch_transitions.zig
+++ b/src/replay/epoch_transitions.zig
@@ -315,16 +315,21 @@ pub fn applyFeatureActivations(
         }
     }
 
+    // On pico/full inflation activation, Agave resets the fee and rent burn
+    // percents to their defaults (see Agave `Bank::apply_feature_activations`).
+    // NOTE: sig's tx-fee burn no longer reads `fee_rate_governor.burn_percent`
+    // (it uses the hard-coded `DEFAULT_BURN_PERCENT` in `distributeTransactionFees`);
+    // we keep these assignments for snapshot-parity with Agave's `SlotConstants`.
     if (new_activations.active(.pico_inflation, slot)) {
         slot_constants.inflation = .PICO;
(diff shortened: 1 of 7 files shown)


src/replay/freeze.zig
```diff
@@ -86,6 +86,7 @@ pub const FreezeParams = struct {
                 .capitalization = &state.capitalization,
                 .blockhash_queue = &state.blockhash_queue,
                 .rent = constants.rent_collector.rent,
+                .fee_burn_percent = constants.fee_rate_governor.burn_percent,
```
dnut: According to this [SIMD](https://github.com/solana-foundation/solana-improvement-documents/pull/465/changes), the burn rate should be hardcoded to 50%.

In agave, they have a method on the bank called [burn_rate](https://github.com/anza-xyz/agave/blob/427d6242609ce6500f4cdec3ead85cc45fab28a3/runtime/src/bank/fee_distribution.rs#L115) that just returns a constant (50%). It's used to [calculate the fees burned](https://github.com/anza-xyz/agave/blob/427d6242609ce6500f4cdec3ead85cc45fab28a3/runtime/src/bank/fee_distribution.rs#L102).

`FeeRateGovernor.burn_rate` is never read once in the entire agave codebase. It appears to have been [set](https://github.com/anza-xyz/agave/blob/427d6242609ce6500f4cdec3ead85cc45fab28a3/runtime/src/bank.rs#L6152-L6162) to the same constant used in `Bank::burn_rate` when a pico_inflation or full_inflation feature was activated. Likewise `FeeRateGovernor.burn_rate` should also be 50% in every snapshot (until the agave code changes).

So what do we use, the field from the snapshot, or the hardcoded value? Both of them seem acceptable at some level.

If the burn rate ever changes, and we use a hardcoded value, we'll definitely need to change the code to use the correct value. If we use the snapshot field, then we may only see a problem during the feature activation itself, but then the problem may resolve itself once we restart and get a newer snapshot (from after the feature activation) from the cluster. We'll have the wrong code for the feature activation, but it's code that may basically work well enough after that.

It's hard to anticipate what will happen as the protocol evolves, but using the snapshot field seems to be more likely to lead to subtle bugs that are hard to detect because they auto-correct after the first failure, whereas using a hardcoded value will simply cause a consistent failure for every single block. Using the snapshot field is a form of defensive programming, whereas using the hardcoded value is a form of offensive programming.

I lean towards the hardcoding approach because it's consistent with the SIMD, agave code, and it alerts us to the issue more loudly if we don't implement a SIMD that modifies this value, forcing us to fix it right away.

dnut (changes requested):

src/replay/freeze.zig
ajw221: [7ba856c](https://github.com/Syndica/sig/pull/1623/commits/7ba856ccf59dd7d7574fd6b3feeaa3744751a80d)

src/replay/epoch_transitions.zig
```diff
@@ -315,16 +315,21 @@ pub fn applyFeatureActivations(
...
+    // percents to their defaults (see Agave `Bank::apply_feature_activations`).
+    // NOTE: sig's tx-fee burn no longer reads `fee_rate_governor.burn_percent`
+    // (it uses the hard-coded `DEFAULT_BURN_PERCENT` in `distributeTransactionFees`);
+    // we keep these assignments for snapshot-parity with Agave's `SlotConstants`.
     if (new_activations.active(.pico_inflation, slot)) {
         slot_constants.inflation = .PICO;
-        slot_constants.fee_rate_governor.burn_percent = 50; // DEFAULT_BURN_PERCENT: 50% fee burn.
-        slot_constants.rent_collector.rent.burn_percent = 50; // 50% rent bur.
+        slot_constants.fee_rate_governor.burn_percent = sig.runtime.sysvar.DEFAULT_BURN_PERCENT;
+        slot_constants.rent_collector.rent.burn_percent = sig.runtime.sysvar.DEFAULT_BURN_PERCENT;
```
yewman: We can drop the `rent_collector.rent.burn_percent` assignments and update the comment as this field is no longer read by anything relevant
https://github.com/anza-xyz/agave/blob/8d843af0806ace2036d274afca929adf3d5f34dc/runtime/src/bank.rs#L6152-L6162

yewman (commented):
We should also address the fact that agave loads rent from the on-chain sysvar account after snapshot load. Issue created here: [rent_collector.rent is zeroes after snapshot load](https://github.com/Syndica/sig/issues/1629#top)
I am okay with this being addressed in a follow up

dnut (dismissed):

src/replay/epoch_transitions.zig
ajw221: [8ee381a](https://github.com/Syndica/sig/commit/8ee381a4b2b6584bcdae73cbd02b556e22490fb2)

dnut (approved):

yewman (approved):

PR #1624, opened 2026-06-23
feat(v2:replay): basic black box test

prestonsn:
* Adds a basic Black Box Test (bbt) that reads a single fec set from file, and asserts end state about replay & shred_receiver (currently just checks that the right number of transactions exec requests are dispatched to the exec service)
* Adds the new `Fixture` setup for loading test fixtures from file


v2/tests/replay/main.zig
```diff
@@ -0,0 +1,173 @@
...
+                .block_pool = block_pool_init,
+                .exec_req_response = exec_req_response_init,
+                .tel = telemetry_init,
+            },
+        },
+        .telemetry = .{
+            .ro = .{},
+            .rw = .{ .region = telemetry_init },
+        },
+    });
```
dnut: `finish` is idempotent so you can avoid this boilerplate. 

```suggestion
    const net_pair = try net_to_shred.finish().memfd.mmapStaticSize(.rw, lib.net.Pair, .{});
    defer std.posix.munmap(@ptrCast(net_pair));

    const exec_req_response = try exec_req_response_region.finish().memfd.mmapStaticSize(
        .rw,
        lib.replay.ExecReqResponse,
        .{},
    );
    defer std.posix.munmap(@ptrCast(exec_req_response));

    var spawned: topology.Children(Topology) = undefined;
    try spawned.spawn(.sandboxed, .{
        .shred_receiver = .{
            .ro = .{ .config = shred_recv_config.finish() },
            .rw = .{
                .tvu_socket = net_to_shred.finish(),
                .deshredded_out = shreds_to_replay.finish(),
                .tel = telemetry_region.finish(),
            },
        },
        .replay = .{
            .ro = .{},
            .rw = .{
                .deshredded_in = shreds_to_replay.finish(),
                .replay_transaction_pool = transaction_pool.finish(),
                .block_pool = block_pool.finish(),
                .exec_req_response = exec_req_response.finish(),
                .tel = telemetry_region.finish(),
            },
        },
        .telemetry = .{
            .ro = .{},
            .rw = .{ .region = telemetry_region.finish() },
        },
    });
```

you could even go a step further and write another `finish` function that's specifically for handing off the region to the caller instead of the topology initializer.

v2/tests/replay/main.zig
```diff
@@ -0,0 +1,173 @@
...
+
+    var count: u32 = 0;
+    while (lib.clock.monotonic(.ns) - start < timeout_ns) {
+        while (request_reader.next()) |request| {
+            try std.testing.expect(request.request_kind == .txn_exec);
+            count += 1;
+            try std.testing.expect(count <= expected_transaction_count);
+            if (count == expected_transaction_count) return;
+        }
+        std.atomic.spinLoopHint();
```
InKryption:
```suggestion
    while (lib.clock.monotonic(.ns) - start < timeout_ns) : (std.atomic.spinLoopHint()) {
        while (request_reader.next()) |request| {
            try std.testing.expect(request.request_kind == .txn_exec);
            count += 1;
            try std.testing.expect(count <= expected_transaction_count);
            if (count == expected_transaction_count) return;
        }
```

v2/tests/replay/fixtures/load.zig
```diff
@@ -0,0 +1,198 @@
...
+}
+
+// NOTE: Transplanted from v1's src/ledger/tests.zig shred fixture loader.
+inline fn readChunk(allocator: std.mem.Allocator, reader: *std.Io.Reader) !?[]const u8 {
+    var size_bytes: [8]u8 = undefined;
+    reader.readSliceAll(&size_bytes) catch |err| switch (err) {
+        error.EndOfStream => return null,
+        else => return err,
+    };
+    const size = std.mem.readInt(u64, &size_bytes, .little);
```
InKryption:
```suggestion
    const size = reader.takeInt(u64, .little) catch |err| switch (err) {
        error.EndOfStream => return null,
        else => return err,
    };
```

v2/tests/replay/fixtures/load.zig
```diff
@@ -0,0 +1,198 @@
...
+
+    return error.FecSetNotFound;
+}
+
+inline fn containsIndex(indices: []const u32, index: u32) bool {
+    for (indices) |candidate| {
+        if (candidate == index) return true;
+    }
+    return false;
+}
```
InKryption:
```suggestion
        try std.testing.expect(std.mem.indexOfScalar(u32, expected_indices, shred_index) != null);

        const seen = if (shred.variant.isData()) &seen_data else &seen_coding;
        const bit_index: usize = @intCast(shred_index);
        try std.testing.expect(!seen.isSet(bit_index));
        seen.set(bit_index);

        selected[selected_count] = packet;
        selected_count += 1;
        if (selected_count == selected.len) return selected;
    }

    return error.FecSetNotFound;
}
```

v2/services/replay.zig
```diff
@@ -111,173 +111,386 @@ var scratch_memory: [256 * 1024 * 1024]u8 = undefined;
...
+    rw.tel.signalReady();
+
+    var fba: std.heap.FixedBufferAllocator = .init(&scratch_memory);
+    const allocator = fba.allocator();
+
+    const Effects = struct {
+        exec_request_sender: *replay.ExecReqResponse.RequestRing.Iterator(.writer),
+        transaction_pool: *replay.TransactionPool,
+
+        const Self = @This();
```
InKryption: Maybe makes more sense to define this outside `serviceMain`?

v2/services/replay.zig
```diff
@@ -486,13 +703,15 @@ fn setChildBlockRef(
...
     };
+    effects.reportBlockRefAssigned(child, child.block_ref);
 }
 
 fn setChildTreeBlockRefs(
     parent: *MerkleNode,
     child: *MerkleNode,
     forest_pool: *MerkleForest.NodePool,
     block_pool: *lib.replay.BlockPool,
+    effects: anytype,
```
InKryption: Comments on these `effects: anytype` fields of the form:
```zig
/// `Replay(Effects)`
effects: anytype
```
would be appreciated.

v2/services/replay.zig
```diff
@@ -111,173 +111,386 @@ var scratch_memory: [256 * 1024 * 1024]u8 = undefined;
...
-    const logger = rw.tel.acquireLogger(@tagName(name), "main");
-    rw.tel.signalReady();
+const TransactionBuffer = struct {
+    ref: replay.TransactionPool.ItemId,
+    buf: *[1232]u8,
+};
 
-    var fba: std.heap.FixedBufferAllocator = .init(&scratch_memory);
-    const allocator = fba.allocator();
+pub fn Replay(comptime Effects: type) type {
```
InKryption: Wouldn't this make more sense in `lib.replay`? That's how the GossipNode equivalent is handled.

v2/services/replay.zig
```diff
@@ -111,173 +111,386 @@ var scratch_memory: [256 * 1024 * 1024]u8 = undefined;
...
-            zone.value(response.task_id);
+        pub fn reportBlockComplete(self: Effects, block_ref: replay.BlockRef) void {
+            _ = .{ self, block_ref };
+        }
+    });
 
-            std.debug.assert(response.request_kind == .txn_exec); // others unimplemented
-            const response_data = response.data.txn_exec;
+    return struct {
+        const Self = @This();
```
InKryption: I think we should stick to the predominant convention where we're using `const ThingSelf = @This();` (ie `const ReplaySelf = @This();` in this case), and putting it after the fields.

dnut (changes requested):

v2/tests/replay/main.zig
prestonsn: True, fixed here: https://github.com/Syndica/sig/pull/1624/commits/57bf2ec36e6e3e30b44f8a435d727a1c6a5c3ffa

v2/tests/replay/fixtures/load.zig
prestonsn: that was silly: https://github.com/Syndica/sig/pull/1624/commits/57bf2ec36e6e3e30b44f8a435d727a1c6a5c3ffa

v2/tests/replay/fixtures/load.zig
prestonsn: cleaned up here: https://github.com/Syndica/sig/pull/1624/commits/57bf2ec36e6e3e30b44f8a435d727a1c6a5c3ffa

v2/services/replay.zig
prestonsn: Added this type comment here: https://github.com/Syndica/sig/pull/1624/commits/57bf2ec36e6e3e30b44f8a435d727a1c6a5c3ffa

v2/services/replay.zig
prestonsn: https://github.com/Syndica/sig/pull/1624/commits/57bf2ec36e6e3e30b44f8a435d727a1c6a5c3ffa

v2/tests/replay/main.zig
prestonsn: Cleaned this up here: https://github.com/Syndica/sig/pull/1624/commits/6198abded32c6f4762c86f0b40cbd2aae982c0ec

v2/services/replay.zig
prestonsn: Moved out of serviceMain here: https://github.com/Syndica/sig/pull/1624/commits/6198abded32c6f4762c86f0b40cbd2aae982c0ec

v2/services/replay.zig
prestonsn: Yeah I agree with that. I didn't want to make too many changes in one PR, but here's how things look if moved into `lib/replay.zig`: https://github.com/Syndica/sig/pull/1624/commits/85b67ed3129dc579226b7507fc4bb2acd121b31d

v2/services/replay.zig
InKryption: Could remove the `ServiceEffectsSelf` there and just use the type itself directly now :P

prestonsn: True, good catch! Removed it here: https://github.com/Syndica/sig/pull/1624/commits/c382f48fb0b13d75327513ba0a8604d0ddfa447a

v2/tests/replay/main.zig
```diff
@@ -0,0 +1,164 @@
...
+fn resignPackets(
+    packets: *[fixture_loader.FEC_SHRED_COUNT]lib.net.Packet,
+    keypair: *const lib.gossip.KeyPair,
+) !void {
+    for (packets) |*packet| {
+        const shred = try lib.shred.Shred.fromPacketChecked(packet);
+        var merkle_root: lib.solana.Hash = undefined;
+        try shred.merkleRoot(&merkle_root);
+
+        const mutable_shred: *lib.shred.Shred = @ptrCast(packet);
```
jbuckmccready: For readability and compile time safety checking I think we should avoid `@ptrCast` for reinterpreting bytes outside of a dedicated helper. `fromPacketChecked` already exists, there can be a `fromPacketUnchecked` that just does the `@ptrCast`.

prestonsn: turns out Shred already had a helpful method for this: https://github.com/Syndica/sig/pull/1624/commits/862f754e45b77ac5a08acaa95a57fc8f294124e2

InKryption (dismissed):

v2/tests/replay/main.zig
jbuckmccready: Before it was reinterpret casting a `*Packet`, and now the code will reinterpret casts a `*Packet.Buffer`, but it works out the same because the buffer is the first field and that's all it needs. This change is even bigger win for readability than I realized.

v2/lib/replay.zig
```diff
@@ -74,3 +87,1352 @@ pub const ExecResponse = extern struct {
...
+    // Iterate over all child nodes, setting their parent node, and deleting any bad child nodes
+    {
+        var maybe_child_node: ?*MerkleNode = children_head;
+        while (maybe_child_node) |child_node| {
+            std.debug.assert(child_node.parent == .null);
+            maybe_child_node = if (child_node.sibling.opt()) |s| s.ptr(&forest.pool) else null;
+
+            if (node.id.mayFollowWith(&child_node.id)) {
+                @branchHint(.likely);
+                child_node.parent = .init(forest.pool.ptrToIndex(node));
```
jbuckmccready: Effects `reportFecSetAttached` should be called here I think.

v2/tests/replay/main.zig
```diff
@@ -0,0 +1,164 @@
...
+    var request_reader = exec_req_response.request_ring.get(.reader);
+    defer request_reader.markUsed();
+
+    var count: u32 = 0;
+    while (lib.clock.monotonic(.ns) - start < timeout_ns) : (std.atomic.spinLoopHint()) {
+        while (request_reader.next()) |request| {
+            try std.testing.expect(request.request_kind == .txn_exec);
+            count += 1;
+            try std.testing.expect(count <= expected_transaction_count);
+            if (count == expected_transaction_count) return;
```
jbuckmccready: This will exit the loop as soon as expected count is received so it wont catch if expected 4 but received 5 (exits at 4).

Seems somewhat challenging to make it deterministically catch the issue every time it runs (unavoidable with this type of integration/blackbox testing), but I wonder if we could do something like expected count *and* service is idle according to the runner?

v2/services/replay.zig
```diff
@@ -108,1228 +21,114 @@ pub const ReadWrite = services.replay.ReadWrite;
...
-    fn init(root_node: *const MerkleNode) BlockDeserialState {
-        std.debug.assert(root_node.block_ref != .null);
-        std.debug.assert(root_node.id.fec_set_idx == 0);
+    pub fn reportBlockRefAssigned(
+        self: ServiceEffects,
+        node: *const replay.MerkleNode,
+        block_ref: replay.BlockRef,
+    ) void {
+        _ = .{ self, node, block_ref };
+    }
```
dnut: I don't think we should be cluttering the replay implementation with a bunch of noop methods that it's expected to call, which have no actual functional purpose in a running validator. It resists refactoring, and it makes the code seem arbitrary.

I don't even see anywhere that these functions have implementations that do anything. So it's not clear to me what benefit we're getting from this. Can you just remove all of these unneeded functions?

Even if these were used in tests, I would still be very skeptical of this approach to testing. I would prefer to validate the existing effects of the code rather than demand it to produce extra effects just for tests. At a certain point you're validating test code instead of production code, which feels a bit pointless.

v2/lib/replay.zig
```diff
@@ -74,3 +87,1352 @@ pub const ExecResponse = extern struct {
...
+                .exec_states = exec_states,
+
+                .deshredded_iter = args.deshredded_in.get(.reader),
+                .exec_response_receiver = args.exec_req_response.response_ring.get(.reader),
+
+                .block_pool = args.block_pool,
+            };
+        }
+
+        pub fn run(self: *ReplaySelf, activity: *runner.Activity.ServiceView) !noreturn {
```
dnut: Ideally, this would be a `poll` function that returns after executing a single step. Services should manually write a loop to invoke poll repeatedly. This makes the component much more composable, which makes it easier to refactor and easier to test.

I don't like this labeled switch because it doesn't do anything unusual that actually benefits from a labeled switch. It's just being used here to reimplement common programming patterns (looping and function calls), but instead using an esoteric syntax that would confuse most readers. It's analogous to replacing function definitions with `goto`s, it's not a reasonable approach imo. I'd like to eliminate this labeled switch. It would be easy to reimplement this as a loop and functions, I just haven't prioritized it.

I understand this pattern was here before your change. You can leave it as is to keep this pr reasonably scoped.

dnut (changes requested):

v2/lib/replay.zig
```diff
@@ -74,3 +87,1352 @@ pub const ExecResponse = extern struct {
...
     success: bool,
 };
+
+const DeserialStates = [BlockPool.capacity]?BlockDeserialState;
+const BlockExecStates = [BlockPool.capacity]?BlockExecState;
+
+pub const TransactionBuffer = struct {
+    ref: TransactionPool.ItemId,
+    buf: *[1232]u8,
+};
```
Sobeston: Do we need this / isn't this just an index and a pointer to the same thing?

v2/lib/replay.zig
```diff
@@ -74,3 +87,1352 @@ pub const ExecResponse = extern struct {
...
+            _ = .{ self, child, parent };
+        }
+
+        pub fn reportBlockRefAssigned(
+            self: Effects,
+            node: *const MerkleNode,
+            block_ref: BlockRef,
+        ) void {
+            _ = .{ self, node, block_ref };
+        }
```
Sobeston: Maybe let's leave these for when we have implementations for them?

Sobeston (commented):
Bit of a tricky review, I'll come back to this, got some comments for now

v2/lib/replay.zig
prestonsn: Yeah I think you're right, for now I've removed the Effects interface entirely.

v2/services/replay.zig
prestonsn: Agreed that it's a lot of clutter. But I do think we'd want something that can easily communicate a service's outgoing effects/outputs. And I do want that interface to be "pluggable" for building detailed trace-based debuggers, fuzzing tools, and sequence tests (tests that cover a certain series of internal state changes occurred across multiple services to produce a certain output) Less "fine-grained" than unit tests, but fit well as improvements to the level of introspection currently provided by black box tests). Perhaps we can implement this through the telemetry service in the future. 

The stream of internal events could also be used to build out rich dashboards to visualize sig's performance/behavior. The interface descriptors could even be used to later generate detailed diagrams/documentation of the data flow between services, with much more specific control flow information than what's available from the topology descriptions alone.

But for now I'm removing it, and there's likely a better approach for this if we want something similar in the future.

v2/lib/replay.zig
prestonsn: Oops I think you're right. Mostly had this in place to clean up args passed to hooks. 

Removed all Effects changes to replay code tho

v2/lib/replay.zig
prestonsn: True, removed Effects from replay entirely

v2/lib/replay.zig
prestonsn: I'm in agreement with you, I'll get a separate PR up for this

v2/tests/replay/main.zig
prestonsn: Good point, I went ahead and used the runner's state to check for idleness before concluding the test. Part of these changes in this commit is also now mocking/handling accountsdb requests coming from replay to allow it to progress as expected from the test's perspective: https://github.com/Syndica/sig/pull/1624/commits/762bb61a36ff0a45d6ad6456aa80895e631a081a

prestonsn: To make things a little more readable, I separated out waiting for the desired final asserted test state to be reached from waiting for the system under test to go idle completely: https://github.com/Syndica/sig/pull/1624/commits/11176faaa92c6bc80284f19dedcca6ed4cfa6d44

v2/services/replay.zig
dnut: I like the idea of improving traceability of our code. We have metrics and logging so far. If you have other ideas about how to record information then I'd like to discuss it to understand what problem it solves relative to existing solutions, and come up with a standard format or approach for it.

I also would like our code to be self-documenting in terms of its outgoing effects. It's inputs are equally important. Both should be clearly visible when looking at a component. Defining it as an interface in the code (e.g. an "effects" struct) is one way to do that.

But that's not really what's going on here. These methods are not actually outgoing effects. They don't affect any other component. This is just a way to expose the inner workings of a single component.

Sure you can argue that as soon as anything becomes a form of telemetry, it is *technically* an effect on the telemetry service, but the conveyed information is not actually semantically meaningful to any other part of the validator. Also, modeling every single metric as a separate effect of the service is extremely detailed and noisy. It masks the signal of what actual interactions this component needs to have with other components beyond "there's a ton of inner workings we'd like to track". The effects interface is no longer a clear signal of the meaningful effects of the component. Using a single construct for both purposes defeats the goal of architectural clarity.

To take metrics as an example, we can define a metrics struct within the component that describes all the metrics it records, so you can see it all in one place, but if there's an effects interface describing the high level interactions between components, I think at most, this should be a single item communicating "this component has telemetry," not by creating one big struct that has a method for each message queue and a method for each metric.

Due to sig's isolated process based architecture, for almost all interactions between components, I think we'll already be using simple data types (such as message queues) that achieve all of the goals of the effects interface:
- clearly describes the inputs and outputs
- abstract away all implementation details of the other component
- are easy to construct in tests
- are not coupled to any other component (it's trivial to swap out the other side of a ring buffer)

So, this is already an interface. Adding an effects struct piles another layer of abstraction on top of it, but it's not clear to me that it delivers any new value that we don't have already. So I lean towards using the queues themselves to describe the component's interface, unless an extra abstraction layer actually communicates valuable information that the the existing data structures do not illustrate.

prestonsn: True! Yeah we could just use the rings and message passing we have. There are places though where we don't rely on queues for message passing output state (like replay's use of `TransactionPool`).  But that's simple enough to model too. 

I'm sure we can come up with something that perhaps unifies these existing abstractions, or at least _documents_ the interactions better in the code in the future (but nothing beats just reading all the code of course).

v2/tests/replay/fixtures/load.zig
```diff
@@ -0,0 +1,202 @@
...
+    defer file.close();
+
+    const contents = try file.readToEndAllocOptions(
+        allocator,
+        1024 * 1024,
+        null,
+        .@"1",
+        0,
+    );
+    defer allocator.free(contents);
```
InKryption: readToEndAlloc[Options], among other methods directly on File, is deprecated.
```suggestion
    var file_reader = file.reader(&.{});
    const file_size = try file_reader.getSize();
    if (file_size > 1024 * 1024) return error.FileTooBig;
    const contents = try allocator.allocSentinel(u8, file_size, 0);
    defer allocator.free(contents);
    try file_reader.interface.readSliceAll(contents);
```

v2/tests/replay/fixtures/load.zig
```diff
@@ -0,0 +1,202 @@
...
+
+        const manifest_dir = std.fs.path.dirname(index_fec_set.path) orelse {
+            return error.InvalidFixtureManifestPath;
+        };
+        const shreds_path = try std.fs.path.join(allocator, &.{
+            FIXTURE_DIR,
+            manifest_dir,
+            "shreds.bin",
+        });
+        defer allocator.free(shreds_path);
```
InKryption: Things like this are why I think it's a good idea to prefer passing around `Dir` and `File` handles instead of string-based paths.

v2/tests/replay/fixtures/load.zig
```diff
@@ -0,0 +1,202 @@
...
+        selected[selected_count] = packet;
+        selected_count += 1;
+        if (selected_count == selected.len) return selected;
+    }
+
+    return error.FecSetNotFound;
+}
+
+// NOTE: Transplanted from v1's src/ledger/tests.zig shred fixture loader.
+inline fn readChunk(allocator: std.mem.Allocator, reader: *std.Io.Reader) !?[]const u8 {
```
InKryption: Why `inline`?
```suggestion
fn readChunk(allocator: std.mem.Allocator, reader: *std.Io.Reader) !?[]const u8 {
```

v2/tests/replay/main.zig
```diff
@@ -0,0 +1,282 @@
...
+    // Phase 2: wait for services to go idle.
+    const idle_start = lib.clock.monotonic(.ns);
+    while (spawned.isActive()) {
+        if (lib.clock.monotonic(.ns) - idle_start >= idle_timeout_ns) {
+            return error.ServicesDidNotBecomeIdle;
+        }
+        try serviceAccountLookups(account_pool, &lookup_reader, &lookup_writer);
+        try drainReplayRequests(&request_reader, &count, expected_transaction_count);
+        std.atomic.spinLoopHint();
+    }
```
InKryption:
```suggestion
    while (spawned.isActive()) : (std.atomic.spinLoopHint()) {
        if (lib.clock.monotonic(.ns) - idle_start >= idle_timeout_ns) {
            return error.ServicesDidNotBecomeIdle;
        }
        try serviceAccountLookups(account_pool, &lookup_reader, &lookup_writer);
        try drainReplayRequests(&request_reader, &count, expected_transaction_count);
    }
```

v2/tests/replay/fixtures/load.zig
prestonsn: Replaced it here: https://github.com/Syndica/sig/pull/1624/commits/76bb3dc10c14076b63986841f84ba6c1761931ec

v2/tests/replay/fixtures/load.zig
prestonsn: Fair, swapped it over: https://github.com/Syndica/sig/pull/1624/commits/76bb3dc10c14076b63986841f84ba6c1761931ec

v2/tests/replay/fixtures/load.zig
prestonsn: Removed the inline hint: https://github.com/Syndica/sig/pull/1624/commits/76bb3dc10c14076b63986841f84ba6c1761931ec

v2/tests/replay/main.zig
prestonsn: Used continue expression here: https://github.com/Syndica/sig/pull/1624/commits/76bb3dc10c14076b63986841f84ba6c1761931ec

v2/tests/replay/fixtures/load.zig
InKryption: Remember: it's not a hint, it's a semantic guarantee, that destroys information the backend could otherwise use.

PR #1628, opened 2026-06-24
feat(v2): blockhash queue

kprotty:
Parses the blockhash_queue out of the snapshot, saves it into the accounts.db file for persistence, and sends it over to replay. Establishes a struct/pattern for passing snapshot metadata to replay (next is vote accounts / epoch schedule, then replay can create feature-set and leader-schedule).

Also hooks up most of the `spinLoopHint()s` to `lib.runner.Connection`

Also fixes a bug on main from Telemetry's service_count being off & preventing it from starting up to process logs.


v2/lib/accounts_db.zig
```diff
@@ -38,3 +40,40 @@ pub const AccountLookups = extern struct {
         self.out.init();
     }
 };
+
+pub const RuntimeMetadata = struct {
+    slot: std.atomic.Value(u64),
+    blockhash_queue: struct {
+        max_age: u64, // read after consuming all of hashes
+        hashes: lib.ipc.Ring(256, Hash),
+    },
```
InKryption: Quick comment, shouldn't this be `extern struct`?

v2/lib/topology.zig
```diff
@@ -99,7 +99,15 @@ pub fn Region(comptime T: type) type {
...
         /// Only create it using `finish`.
-        pub const Initialized = struct { memfd: Memfd };
+        pub const Initialized = struct {
+            memfd: Memfd,
+
+            // Phantom dependency on T so each `Region(T).Initialized` is a distinct type.
+            // Without this, the struct body is `{ memfd: Memfd }` regardless of T and Zig
+            // collapses every instantiation to the same type, which breaks `countRegionShares`
+            // (and any other code that filters topology fields by `Region(X).Initialized`).
+            pub const _ = T;
```
InKryption: Better way to express this is with `comptime { _ = T; }`

kprotty: Thought so originally, but it gets `error: pointless discard of function parameter`

InKryption: Ah, right, it is actually being used elsewhere in this code. Could instead use `_ = &T;`. Not ideal either, but I think still better than having a bogus decl.

InKryption (dismissed):

v2/init/main.zig
```diff
@@ -327,6 +331,7 @@ pub fn main() !void {
         .replay = .{
             .ro = .{},
             .rw = .{
+                .snapshot_metadata_in = snapshot_metadata.finish(),
```
dnut, line 334: if the data is consumed using ring buffers, maybe we should be providing different rings to each service to ensure they don't interfere with each other?

v2/lib/accounts_db.zig
```diff
@@ -38,3 +40,40 @@ pub const AccountLookups = extern struct {
         self.out.init();
     }
 };
+
+pub const RuntimeMetadata = extern struct {
+    slot: std.atomic.Value(u64),
+    blockhash_queue: extern struct {
+        max_age: u64, // read after consuming all of hashes
+        hashes: lib.ipc.Ring(256, Hash),
```
dnut, line 49: Hash alone is not sufficient information to determine which slot each hash goes with. there may be skipped slots.

also, the number of hashes is capped at 300 so why don't you just make this a simple array? i get the ring buffer idea for things like vote accounts, but the blockhash queue is well bounded in size.

v2/lib/accounts_db.zig
```diff
@@ -38,3 +40,40 @@ pub const AccountLookups = extern struct {
...
+
+    pub fn getSlotBlocking(self: *RuntimeMetadata, runner: lib.runner.Connection) !Slot {
+        const slot = while (true) {
+            const slot = self.slot.load(.acquire);
+            if (slot != invalid_slot) break slot;
+            try runner.activity.signalIdleSpinning();
+        };
+
+        try runner.activity.signalActive();
+        return slot;
```
dnut: your code is fine but seems feels more direct and clear to me:
```suggestion
        while (true) {
            const slot = self.slot.load(.acquire);
            if (slot != invalid_slot) {
                try runner.activity.signalActive();
                return slot;
            }
            try runner.activity.signalIdleSpinning();
        }
```
just a suggestion

v2/lib/accounts_db.zig
```diff
@@ -38,3 +40,40 @@ pub const AccountLookups = extern struct {
         self.out.init();
     }
 };
+
+pub const RuntimeMetadata = extern struct {
+    slot: std.atomic.Value(u64),
+    blockhash_queue: extern struct {
+        max_age: u64, // read after consuming all of hashes
```
dnut: any reason for this not be a doc comment?
```suggestion
        /// read after consuming all of hashes
        max_age: u64,
```

v2/lib/accounts_db.zig
kprotty: https://github.com/Syndica/sig/pull/1628/commits/714e853ee0df4bf50bb7d3bb2b47f47127ccd59e

v2/lib/topology.zig
kprotty: https://github.com/Syndica/sig/pull/1628/commits/f994c1d791e7505cd33974bce265cd5b8b45b5c1

v2/lib/accounts_db.zig
```diff
@@ -38,3 +40,40 @@ pub const AccountLookups = extern struct {
...
+        self.blockhash_queue.max_age = 0;
+        self.blockhash_queue.hashes.init();
+    }
+
+    /// Unblocks all getSlotBlocking() callers with the given slot value.
+    /// Can be called only once.
+    /// Should also only call after all other RuntimeMetadata fields are populated.
+    pub fn populateSlot(self: *RuntimeMetadata, slot: Slot) void {
+        std.debug.assert(slot != invalid_slot);
+        std.debug.assert(self.slot.swap(invalid_slot, .release) == invalid_slot);
```
dnut:
```suggestion
        std.debug.assert(self.slot.swap(slot, .release) == invalid_slot);
```

v2/services/replay.zig
```diff
@@ -126,12 +127,35 @@ pub fn serviceMain(runner: lib.runner.Connection, _: ReadOnly, rw: ReadWrite) !n
...
+        defer blockhashes_in.close();
+
+        var start_block: usize = 0;
+        while (true) {
+            const hashes = try blockhashes_in.getBufferBlocking(runner);
+            if (hashes.len == 0) break; // blockhashes_out closed their end
+
+            for (hashes) |*hash| {
+                blockhash_states[start_block] = hash.*;
+                start_block += 1;
```
dnut: this needs to be a valid BlockRef. the tree needs to be constructed in the block pool as these hashes are populated.

dnut (changes requested):

v2/init/main.zig
kprotty: i believe the only relationship will be accounts_db -> replay.

the slot readiness is a tmp measure. It should be replay -> shred_receiver {slot/LeaderSchedule}

v2/lib/accounts_db.zig
kprotty: Its 301 instead of 300 (>= max_age). Could hard-code as a 301 array with len, but accounts_db stores it as a stream of hash additions anyway; would need to create a local ring buffer in `Rooted.loadExisting` then copy it to the array. Having a ring here seems simpler.

The hash entries from snapshot dont indicate what slot they belong to. I think the exact slot discrepancy wont affect much; replay should never be reaching back to slots older than the rooted one for any block-aware data structure.

v2/services/replay.zig
kprotty: https://github.com/Syndica/sig/pull/1628/commits/e1a805b7a87d0c03ce6dcbd0e15369ddbe483f0e

v2/lib/accounts_db.zig
kprotty: https://github.com/Syndica/sig/pull/1628/commits/e1a805b7a87d0c03ce6dcbd0e15369ddbe483f0e

v2/lib/accounts_db.zig
kprotty: https://github.com/Syndica/sig/pull/1628/commits/e1a805b7a87d0c03ce6dcbd0e15369ddbe483f0e

v2/lib/accounts_db.zig
kprotty: https://github.com/Syndica/sig/pull/1628/commits/e1a805b7a87d0c03ce6dcbd0e15369ddbe483f0e

v2/services/replay.zig
dnut: This creates the blockpool node but it leaves it as undefined memory. It needs to be initialized to a valid state, otherwise traversing the block tree will not work. 

This is what needs to be initialized:
```zig
pub const Node = extern struct {
    parent: BlockRef.Optional = .null,
    child: BlockRef.Optional = .null,
    sibling: BlockRef.Optional = .null,
    slot: solana.Slot,
};
```

v2/lib/accounts_db.zig
dnut: > would need to create a local ring buffer in Rooted.loadExisting

I don't see what you mean. readExisting just takes a buffer and writes to it. You could just pass in a pointer to the array. Seems simpler than the current approach of acquiring a buffer from the ring.

kprotty: `readExisting` is for reading/skipping bytes from the `accounts.db` file. This is rather about `loadExisting` which reads a set of hash changes as deltas from Sections in the `accounts.db` file. Each delta can contain 0..N hashes.

When it loads from a snapshot, it creates a delta with 301 hashes. But for each slot rooted, there would be a new hash delta of size 1. When loading from an existing `accounts.db`, older deltas could be discovered in older sections  (making the total in the file >301), hence "create a local ring buffer in loadExisting" to collect only the latest after scanning through.

An alternative is to just write 301-sized hash deltas for every rooted slot, but that seems unnecessary. This instead just streams them all to replay, who then decides how many to keep.

v2/services/replay.zig
kprotty: The back-populated hashes dont necessarily need valid blocks. The pool could technically start at the `start_block` offset if we wanted (given normal pool traversal shouldnt go upwards/higher than the root).

Proper initialization of the tree (along with first_slot fixes & the like) is deferred. Want this PR to simply be getting the data to replay.

Add a TODO?

dnut: The block tree nodes need to point at each other to make the data reachable. The way this works now, none of the data will be usable, so it's pointless to even write it into the array.

A TODO is fine. Trevor can take this over.

```zig
// TODO: initialize the block tree nodes and descend new blocks off these
latest_block = try rw.block_pool.createId();
```

kprotty: https://github.com/Syndica/sig/pull/1628/commits/303621fcc5f57ee428b725a2f6e73ebdf00c74cb

dnut (approved):

PR #1631, opened 2026-06-25
refactor(runtime): unify the blockhash queue and status cache checks

dnut:
I refactored the usage of the blockhash queue and the status cache to be more unified in the runtime. They are now both checked together behind the existing `StatusChecker` interface.

This is to facilitate our intended design in v2, which requires them to be linked. I also think it makes the code clearer.

The current design has them separated, but they don't make sense in isolation - they only work when used together. They're already entangled today, just in a subtle and confusing way.

## Conceptual Background

The blockhash queue and the status cache function hand-in-hand to verify that a transaction is recent and not present in a recent block.

The blockhash queue is the collection of the last N blockhashes. The status cache is the subset of those blockhashes which have had at least one transaction that specified them as a recent blockhash, mapped to the set of those transaction's hashes. First you check the blockhash queue to see if the blockhash is recent. If so, we need to check the status cache to see if the transaction has already been processed. If so, the transaction must fail.

If it's not in the blockhash queue, then either the hash is not a blockhash, or both of these are true:
- the transaction is too old
- the status cache is unable to answer the question *"has this transaction been processed?"* (because it's too old)

At that point, there's no point in checking the status cache any more. We already know it won't be there. Then you can try to interpret the "recent_blockhash" field instead as a durable nonce.

Currently in the code, this is implemented by first checking the blockhash queue, then if that fails, trying it as a nonce, then checking the status cache regardless. Something like this:

```python
if recent_blockhash is not recent:
    check_and_advance_nonce(recent_blockhash as nonce)
return_error_if_in_status_cache(recent_blockhash, transaction_hash)
```

Checking the status cache after verifying that the hash is actually a durable nonce is a total waste of time - it can't possibly hit, because the `recent_blockhash` is not a blockhash. So this is actually equivalent to:

```python
if recent_blockhash is recent:
    return_error_if_in_status_cache(recent_blockhash, transaction_hash)
else:
    check_and_advance_nonce(recent_blockhash as nonce)
```

The latter is also more intuitive: two steps (interpret recent_blockhash as a blockhash, else as a nonce) instead of three (interpret as a blockhash, else as a nonce, then interpret as a blockhash again).

## What could be

There is no real need for the blockhash queue and the status cache to even be separate data structures. The status cache could in principle serve both responsibilities as long as you ensure it has an entry for every recent blockhash. The agave and sig v1 implementations just have this weird approach some arbitrary chunk of data has been extracted out of the status cache and put in its own data structure called the blockhash queue. Specifically, the blockhash queue is the minimal subset of information from an exhaustive status cache that's needed to determine if you should try to interpret the `recent_blockhash` field as a nonce.

But in a cleaner implementation, the status cache could consolidate all this data and provide a clear signal to the caller about whether the blockhash is recent and whether the transaction is in a recent block.

## This change

This PR refactors the runtime to facilitate the cleaner approach described in the prior section. For v1, it still uses the blockhash queue and the status cache behind the scenes, but it reorganizes the code to be ordered like the latter pseudocode example above, with the first two lines factored into a unified function that handles both the blockhash queue check and the status cache check. That function returns a 3-variant enum the caller switches on:

```python
switch(check_transaction_status(transaction, recent_blockhash))
    recent_and_unprocessed => execute transaction
    already_processed => fail transaction
    unknown_blockhash => check_and_advance_nonce(recent_blockhash as nonce)
```

## Justification for making this change now

In v2, we're planning to organize both the blockhash queue and status cache around the same block tree. This is essentially a unification of the status cache and the blockhash queue.

The code will traverse the block tree until it finds a block with the recent blockhash, then look at the same node to find the transactions that specify that block as its recent blockhash. This is a unified process where the second step depends on information from the first. The old decoupled approach doesn't accommodate this consolidation, but the new one does.

I've moved the blockhash queue inside the StatusChecker interface, and made that interface responsible for the entire check, because it's the simplest approach that allows us to be agnostic to how v1 and v2 differ in their:
- representations of the blockhash history
- representations of the transaction status
- implementations of the check

### DIFF
--- a/shared/runtime/execution_interfaces.zig
+++ b/shared/runtime/execution_interfaces.zig
@@ -95,17 +95,17 @@ pub const StatusChecker = struct {
         /// The transaction was already executed in a recent block on the
         /// current fork, so it is not legal to include it in the current block.
         already_processed,
-        /// The recent_blockhash count not be identified. It could not be
-        /// determined whether the transaction already exists in a block. This
-        /// means the recent_blockhash is too old or invalid, or the transaction
-        /// uses a durable nonce.
+        /// The recent_blockhash was not found, so it could not be determined
+        /// whether the transaction already exists in a block. This means the
+        /// recent_blockhash is too old or invalid, or the transaction uses a
+        /// durable nonce.
         unknown_blockhash,
     };
 
-    /// Checks recent blocks up to max_age to see a transaction it was already
+    /// Checks recent blocks up to max_age to see if a transaction was already
     /// processed in any of these blocks.
     ///
-    /// Does not locate transactions using a durable nonce. It will return
+    /// Does not locate transactions that use a durable nonce. It will return
     /// "unknown_blockhash" for those.
     pub fn check(
         self: StatusChecker,
--- a/shared/runtime/execution_interfaces.zig
+++ b/shared/runtime/execution_interfaces.zig
@@ -5,7 +5,6 @@ const sig = @import("../lib.zig");
 const AccountSharedData = @import("AccountSharedData.zig");
 const Hash = sig.core.Hash;
 const Pubkey = sig.core.Pubkey;
-const TransactionError = sig.core.transaction_error.TransactionError;
 
 pub const AccountLoadError = error{ OutOfMemory, AccountsDBError };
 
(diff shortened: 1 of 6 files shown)


shared/runtime/execution_interfaces.zig
```diff
@@ -83,17 +82,38 @@ pub const EpochStakeReader = struct {
...
+    ) Result,
+
+    pub const Result = enum {
+        /// The transaction is recent, and has never been executed in the
+        /// current fork, so it is legal to execute this in the current block.
+        recent_and_unprocessed,
+        /// The transaction was already executed in a recent block on the
+        /// current fork, so it is not legal to include it in the current block.
+        already_processed,
+        /// The recent_blockhash count not be identified. It could not be
```
InKryption:
```suggestion
        /// The recent_blockhash count could not be identified. It could not be
```

shared/runtime/execution_interfaces.zig
```diff
@@ -83,17 +82,38 @@ pub const EpochStakeReader = struct {
...
+        /// determined whether the transaction already exists in a block. This
+        /// means the recent_blockhash is too old or invalid, or the transaction
+        /// uses a durable nonce.
+        unknown_blockhash,
+    };
 
+    /// Checks recent blocks up to max_age to see a transaction it was already
```
InKryption:
```suggestion
    /// Checks recent blocks up to max_age to see if a transaction was already
```

InKryption (dismissed):
Just noticed a couple spelling errors reading the doc comments, otherwise this makes sense to me.

shared/runtime/execution_interfaces.zig
dnut: [4879fbd](https://github.com/Syndica/sig/pull/1631/commits/4879fbd7fcd691ab9466ca32fa0ecce6134ad344)

shared/runtime/execution_interfaces.zig
dnut: [4879fbd](https://github.com/Syndica/sig/pull/1631/commits/4879fbd7fcd691ab9466ca32fa0ecce6134ad344)

InKryption (approved):

yewman (approved):
Lgtm 👍

kprotty (approved):

PR #1632, opened 2026-06-26
feat(v2:shred): receiver primitives for shred harness

yewman:
Four independent primitives the `sol_compat_shred_parse_v1` conformance harness needs out of `Receiver`. Each is self-contained — review per commit.

No behavioural change to existing callers: new feature gates default to inactive, `Receiver.reset()` is opt-in, and the hoisted parse-time bound was already enforced one stage later.

### Commits

1. **`feat(v2:shred): split debug-skip flags into sig-verify and version-check`** — replaces the single `-Ddebug-skip-shred-checks` meta-flag with two orthogonal options (`-Ddebug-skip-shred-sig-verify`, `-Ddebug-skip-shred-version-check`). The harness needs to skip sig verify (synthetic shreds, no known leader) but keep version-check on.

2. **`feat(v2:shred): add Receiver.reset() for state-free reuse`** — in-place reset of `Receiver`, `InProgressSets`, `DoneSets`, and a new `Pool.reset()` that rebuilds the free list without touching the backing buffer. Lets the harness reuse a single `Receiver` across fixtures (a fresh one per call is too expensive — the ctx pool alone holds 4096 in-flight sets).

3. **`feat(v2:shred): reject shred-indices past max_shreds_per_slot`** — hoists the `slot_idx < 32768` bound into `Shred.fromPacketChecked` so the parse-level verdict (`reportShredParseResult`) matches Agave/FD. Previously the bound was enforced one frame too late — after the harness had already observed `parses_as_chained = true`. The `32768` literal is promoted to a single named `pub const max_shreds_per_slot`. Closes the last shred-fixture mismatch in `test-vectors/shred/fixtures` (1001/1001).

4. **`feat(v2:shred): discard unexpected data-complete shreds under feature gate`** — adds `Receiver.Features` with the `discard_unexpected_data_complete_shreds` activation slot (default `maxInt(Slot)` = inactive). When active for a slot, a data shred with `DATA_COMPLETE` set is only accepted when `slot_idx == fec_set_idx + FEC_SHRED_COUNT - 1`. Enforced in `processPacketInner` because it needs the runtime activation slot. Mirrors agave's [`ledger/src/shred/filter.rs`](https://github.com/anza-xyz/agave/blob/v4.1.0-rc.1/ledger/src/shred/filter.rs#L327-L342).

### DIFF
--- a/v2/lib/collections/pool.zig
+++ b/v2/lib/collections/pool.zig
@@ -240,6 +240,18 @@ pub fn Pool(Item: type, IdInt: type) type {
             };
         }
 
+        /// Rebuilds the free list in place, returning every item to the pool
+        /// without freeing or reallocating the backing buffer. The buffer
+        /// contents are clobbered.
+        pub fn reset(self: *PoolSelf) void {
+            const buf = self.buf[0..self.len];
+            for (buf[0 .. buf.len - 1], 0..) |*node, i| {
+                node.* = .{ .next_free = @enumFromInt(i + 1) };
+            }
+            buf[buf.len - 1] = .{ .next_free = .null };
+            self.free_list = @enumFromInt(0);
+        }
+
         // take head off free_list
         pub fn create(self: *PoolSelf) !*Item {
             const head = self.free_list.opt() orelse return error.OutOfSpace;
--- a/v2/lib/collections/pool.zig
+++ b/v2/lib/collections/pool.zig
@@ -244,12 +244,8 @@ pub fn Pool(Item: type, IdInt: type) type {
         /// without freeing or reallocating the backing buffer. The buffer
         /// contents are clobbered.
         pub fn reset(self: *PoolSelf) void {
-            const buf = self.buf[0..self.len];
-            for (buf[0 .. buf.len - 1], 0..) |*node, i| {
-                node.* = .{ .next_free = @enumFromInt(i + 1) };
-            }
-            buf[buf.len - 1] = .{ .next_free = .null };
-            self.free_list = @enumFromInt(0);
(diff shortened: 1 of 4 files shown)


InKryption (dismissed):

v2/lib/collections/pool.zig
```diff
@@ -222,6 +222,18 @@ pub fn Pool(Item: type, IdInt: type) type {
...
+        /// without freeing or reallocating the backing buffer. The buffer
+        /// contents are clobbered.
+        pub fn reset(self: *PoolSelf) void {
+            const buf = self.buf[0..self.len];
+            for (buf[0 .. buf.len - 1], 0..) |*node, i| {
+                node.* = .{ .next_free = @enumFromInt(i + 1) };
+            }
+            buf[buf.len - 1] = .{ .next_free = .null };
+            self.free_list = @enumFromInt(0);
+        }
```
Sobeston, line 249: this seems the same
```suggestion
        pub fn reset(self: *PoolSelf) void {
            const buf = self.buf[0..self.len];
            self.* = init(buf);
        }
```

Sobeston (dismissed):
Looks great, I've got one comment, resolve it if you feel like it

v2/lib/collections/pool.zig
yewman: 17d8772bd9c7d77feee045cd3318c8e4555f580c

hamza-syndica (approved):

Sobeston (approved):

prestonsn (approved):
Looks good! commits are unsigned for some reason

PR #1634, opened 2026-06-27
refactor(v2/build.zig): reorganize for clarity

dnut:
build.zig is the sole code file that describes our entire architecture at a high level, so it is especially important for it to be clear.

The old build.zig felt like a disorganized sea of modules, executables, libraries, tests, tools, and subcommands that is hard to glance at and make any sense of at a high level. It took me hours to walk through every line of code and absorb the entire build graph into my mind.

This refactor gives the file a basic high-level structure where additional structure would be valuable, especially around CLI definition, while keeping the inner build graphs simple, direct, and and flexible.

A primary goal of this refactor was to make the following clearer:

- What each CLI subcommand (top-level step) actually does. This should be apparent where the subcommands are defined, not by tracing through a massive build function and helper functions that configure build steps.
- What modules, libraries, and executables compose Sig, as opposed to the surrounding tools.
- How build processes are defined (independently of the CLI). A build process should purely describe what exists and how it is built. `no-bin`, `no-run`, and other CLI concepts should only decide which steps are wired up for a particular invocation. This refactor intentionally describes builds before subcommands to ensure this.

We should also avoid over-generalization. The structure of the inner build graph should not be too constrained. We need the flexibility to define arbitrary dependency graphs. This is why I've kept Sig's build process defined as one function that creates all of its modules, libraries, and executables with arbitrary and direct invocations of std.Build functions, rather than being split into artificial layers like "build all modules first, then build all executables." That kind of structure *would* be tidier, but it would sacrifice flexibility.

A few pieces do benefit from standardization to ensure consistent application of correct behavior:

- Unit tests: They should be as easy to enable as possible, they should all follow the same approach, and they all need to run together.
- Executables: Every executable can be compiled, installed, or run, and the existing `no-bin` / `no-run` configuration already applies uniformly to those behaviors. So it is useful to define an executable as the bundle of those three processes, then have one path for wiring any executable into the CLI.

### Semantic changes

While this is primarily a refactor, a few semantic changes were made. This was to achieve standardization over things that were previously applied inconsistently without a clear reason.
- `lint-test` is no longer a step. Its unit tests are absorbed into `unit-test` and `test` steps.
- `b.args` are now passed to all executables except unit tests.
- `use_llvm` cli argument is now used for all builds.

Inconsistencies make the code more complex, harder to understand, and harder to use correctly. If we're going to introduce inconsistencies for these things intentionally, it needs to be clearly documented with an explanation in the code. Feel free to leave a comment on the PR if you think any of these *do* need to be inconsistently applied. I can make a change and leave a comment in the code if necessary.

### DIFF
--- a/v2/build.zig
+++ b/v2/build.zig
@@ -19,32 +19,42 @@ pub fn build(b: *Build) !void {
 
     // -- CLI Commands (Top Level Steps) ----
 
-    // install -- configured when configuring other steps
+    // install (default step)
     const install_step = b.getInstallStep();
+    install_step.dependOn(sig.exe.installStep());
+    install_step.dependOn(tools.shred_stream.installStep());
+    install_step.dependOn(tools.lint.installStep());
+    for (unit_tests.tests.items) |exe| install_step.dependOn(exe.installStep());
+    for (tools.black_box_tests) |exe| install_step.dependOn(exe.installStep());
 
     // run
     const run_step = b.step("run", "Run sig");
-    sig.exe.addToSteps(config.exe, install_step, run_step);
+    sig.exe.addToStep(run_step);
 
     // sig
     const sig_step = b.step("sig", "Build only the sig binary, without running it.");
-    sig.exe.addToSteps(.{ .install = config.exe.install, .run = false }, install_step, sig_step);
+    sig_step.dependOn(sig.exe.installStep());
 
     // unit-test
     const unit_test_step = b.step("unit-test", "Run unit tests.");
-    unit_tests.addToSteps(config.exe, install_step, unit_test_step);
+    for (unit_tests.tests.items) |exe| unit_test_step.dependOn(exe.installStep());
+    if (unit_tests.kcov) |kcov| {
+        if (config.exe.run) unit_test_step.dependOn(kcov.save_results_step);
+    } else for (unit_tests.tests.items) |unit_test| {
+        if (unit_test.run) |run| unit_test_step.dependOn(&run.step);
+    }
 
     // bb-test
     const bb_test_step = b.step("bb-test", "Run black box tests.");
-    for (tools.black_box_tests) |bbt| bbt.addToSteps(config.exe, install_step, bb_test_step);
+    for (tools.black_box_tests) |bbt| bbt.addToStep(bb_test_step);
 
     // shred-stream
     const shred_stream_step = b.step("shred-stream", "Stream shreds from an Agave ledger");
-    tools.shred_stream.addToSteps(config.exe, install_step, shred_stream_step);
+    tools.shred_stream.addToStep(shred_stream_step);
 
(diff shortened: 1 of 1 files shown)


v2/build.zig
```diff
@@ -1,593 +1,605 @@
...
-        .install = install_opt,
-        .run = run_opt,
-    };
-}
+/// All executables can be compiled, installed, and run. It keeps things simple
+/// to construct all the steps together and let callers decide when to use them.
+const Executable = struct {
+    compile: *Build.Step.Compile,
+    install: *Build.Step.InstallArtifact,
+    run: *Build.Step.Run,
```
InKryption: The big problem with unconditionally adding these steps to the graph is that the associated compile step ends up being forced to enter codegen, which is unnecessary work for build-on-save, and any other use case that wants to run semantic analysis without doing useless work. (this is especially important with the uniform application of use_llvm to everything).

And just to be quite clear: even if the install/run step doesn't end up in the resulting step graph, simply calling `addRunArtifact` and/or `addInstallArtifact` on the compilation step to create them is what causes it to reach codegen.

This is why they were optional.

v2/build.zig
```diff
@@ -1,593 +1,605 @@
...
+        if (self.kcov) |kcov| {
+            const kcov_run = self.build.addSystemCommand(&.{
+                "kcov",
+                "--collect-only",
+                "--include-pattern=v2/",
+                "--exclude-pattern=.cache",
+            });
+            const output_dir = kcov_run.addOutputDirectoryArg("output");
+            kcov_run.addArtifactArg(unit_test);
+            kcov_run.has_side_effects = true;
```
InKryption, line 553: Why does kcov need to declare side effects?

InKryption (changes requested):

v2/build.zig
dnut: Well that's unfortunate. Here we are building a dependency graph to determine what steps to execute, and some of the steps execute partially regardless of whether anything depends on them.

d0dc0404274f21f91cf4e20190f6af969dd7ada3

InKryption: Yeah, it's a bit unfortunate. Making this smarter would require splitting up the Compile step across fully independent "Install" and "Run" steps in the build system itself; maybe that'll come in the future. That, or having a native concept of a "sema-only" mode for the compile step natively.

v2/build.zig
dnut: Normally the tests run every you call `zig build test`. They are never skipped due to their results being assumed to be cached.

When the tests are wrapped with kcov, they lose that. They are treated as cached after the first run. Setting this to true makes the unit test behavior consistent - always run the tests, regardless of whether kcov is enabled.

InKryption (approved):

Sobeston (approved):

PR #1635, opened 2026-06-28
fix(bloom): guard contains and add against a zero-capacity filter

valentynkit:
# Intent

Closes #1633. A gossip `PullRequest` can carry a bloom filter whose `keys` list is non-empty while its bit array has zero capacity. Building the pull response calls `Bloom.contains` on that filter, which reaches `pos`:

```zig
return hashAtIndex(bytes, hash_index) % @as(u64, self.bits.capacity());
```

When `capacity()` is zero this divides by zero, so one unauthenticated UDP packet aborts the node.

The zero-capacity state is reachable from the wire. `BitVec(T)` stores `bits` as an optional slice with a separate `len`, and `keys` deserializes independently, so `keys = [k]` with `bits = null` and `len = 0` is a valid encoding. Agave avoids this because its `Bloom` checks `bits.is_empty()` in `contains` and `add` and uses `checked_rem` in `pos`; Sig's `Bloom` omits that guard.

# Implementation

Return early from `contains` (false) and `add` (no-op) when the filter has no bits, before any key is hashed. A zero-capacity filter has no set bits, so it matches nothing, which is what `false` already means on the response path. The guard goes in `contains` and `add` rather than `pos` because a guarded `pos` would still hand an index to `bits.isSet`, which asserts `index < bit_length` on the empty bitset. Putting it in the primitive also covers every caller, not just the gossip path.

# Ramifications

The behavior change is limited to zero-capacity filters: `contains` returns `false` and `add` is a no-op instead of panicking. Filters that have capacity are untouched. The bug itself is a remote, unauthenticated, single-packet crash, with no amplification (the pull response is already bounded by the filter mask and `max_number_values`) and no memory unsafety.

A defense-in-depth option is to also reject such a filter in `GossipMessage.sanitize` (the empty `.PullRequest` arm), as #1565 does for EpochSlots, which would drop the packet and count it as invalid. I kept this PR at the bloom layer to match Agave; I can add the sanitize check as well if you'd prefer.

# Tests

`test "zero-capacity filter does not divide by zero"` in `src/bloom/bloom.zig` builds a filter with `n_bits = 0` and one key, then exercises `contains` and `add`. It aborts in `pos` without the guard and passes with it.

### DIFF
--- a/src/bloom/bloom.zig
+++ b/src/bloom/bloom.zig
@@ -50,6 +50,7 @@ pub const Bloom = struct {
     }
 
     pub fn add(self: *Bloom, key: []const u8) void {
+        if (self.bits.capacity() == 0) return;
         for (self.keys.items) |hash_index| {
             const i = self.pos(key, hash_index);
             if (!self.bits.isSet(i)) {
@@ -60,6 +61,7 @@ pub const Bloom = struct {
     }
 
     pub fn contains(self: *const Bloom, key: []const u8) bool {
+        if (self.bits.capacity() == 0) return false;
         for (self.keys.items) |hash_index| {
             const i = self.pos(key, hash_index);
             if (self.bits.isSet(i)) {
@@ -181,6 +183,15 @@ test "serialized bytes equal rust (one key)" {
     };
 
     try testing.expectEqualSlices(u8, &rust_bytes, bytes[0..bytes.len]);
+}
+
+test "zero-capacity filter does not divide by zero" {
+    var bloom = try Bloom.init(testing.allocator, 0, null);
+    defer bloom.deinit();
+
+    try bloom.addKey(1);
+    try testing.expect(!bloom.contains(&[_]u8{ 1, 2, 3 }));
+    bloom.add(&[_]u8{ 1, 2, 3 });
 }
 
 test "serialized bytes equal rust (multiple keys)" {
--- a/src/bloom/bloom.zig
+++ b/src/bloom/bloom.zig
@@ -203,10 +203,10 @@ test "deserialized zero-capacity filter does not divide by zero" {
     var buf: [10000]u8 = undefined;
     const out = try bincode.writeToSlice(buf[0..], bloom, .{});
 
-    var deserialized = try bincode.readFromSlice(testing.allocator, Bloom, out, .{});
+    var deserialized: Bloom = try bincode.readFromSlice(testing.allocator, Bloom, out, .{});
     defer bincode.free(testing.allocator, deserialized);
 
-    try testing.expectEqual(@as(u64, 0), deserialized.bits.capacity());
+    try testing.expectEqual(@as(usize, 0), deserialized.bits.capacity());
     try testing.expect(!deserialized.contains(&[_]u8{ 1, 2, 3 }));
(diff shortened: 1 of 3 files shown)


src/bloom/bloom.zig
```diff
@@ -183,6 +185,15 @@ test "serialized bytes equal rust (one key)" {
...
 
+test "zero-capacity filter does not divide by zero" {
+    var bloom = try Bloom.init(testing.allocator, 0, null);
+    defer bloom.deinit();
+
+    try bloom.addKey(1);
+    try testing.expect(!bloom.contains(&[_]u8{ 1, 2, 3 }));
+    bloom.add(&[_]u8{ 1, 2, 3 });
+}
+
```
ajw221, line 222: while this test _does_ cover the target codepath, I think it's worth adding another specifically around the bincode deserialization that would most likely be hit on a live validator run

ajw221 (changes requested):

src/bloom/bloom.zig
valentynkit: Thanks for feedback!
Will create follow up commit to handle this.

valentynkit: Added deserialization tests at three layers (defense in depth):

- the filter itself
- a full PullRequest GossipMessage
- filterSignedGossipDatas.

Each panics in pos without the guard.

Still getting familiar with Sig's conventions, so if this is more coverage than you'd want here, let me know and I'll trim.

src/bloom/bloom.zig
```diff
@@ -60,6 +61,7 @@ pub const Bloom = struct {
     }
 
     pub fn contains(self: *const Bloom, key: []const u8) bool {
+        if (self.bits.capacity() == 0) return false;
```
ajw221: I'm just now realizing if `self.keys` is empty _and_ capacity is 0, this will return `true`, which seems like a pre-existing bug we may want to account for here now that we're checking the capacity

src/gossip/pull_response.zig
```diff
@@ -147,3 +147,39 @@ test "gossip.pull_response: test filtering values works" {
...
+
+    var result = try filterSignedGossipDatas(
+        random,
+        std.testing.allocator,
+        &gossip_table,
+        &filter,
+        0,
+        100,
+    );
+    defer result.deinit();
```
ajw221, line 184: Currently the test ensures a panic didn't occur, but isn't ensuring this is functioning as intended. I think adding the assertion here would close the loop in terms of ensuring this test is full testing what it needs.

```suggestion
    defer result.deinit();
    
    try std.testing.expectEqual(@as(usize, 8), result.items.len);
```

ajw221 (changes requested):
some small nits, otherwise looks good!

src/gossip/pull_response.zig
valentynkit: Thanks, applied the fix.

src/bloom/bloom.zig
valentynkit: Yeah, i missed the case when keys are empty.
Agave contains returns false when keys.isempty || bits.isempty.
Extended the guard to cover empty keys too.

PR #1618, opened 2026-06-22
refactor(runtime): hardcode migrate_stake_program_to_core_bpf, migrate_config_program_to_core_bpf, migrate_address_lookup_table_program_to_core_bpf

hamza-syndica:
Closes #1539 #1540 #1582

PR #1641, opened 2026-06-30
refactor(runtime): remove dead native stake/ALT implementations

hamza-syndica:
Closes #1627 (PR 1)

Removes ~4800 lines of dead code after hardcoding the core-BPF migrations in #1618.

- Remove stake and address_lookup_table from NATIVE dispatch map
- Delete address_lookup_table/execute.zig
- Strip stake/lib.zig to module shell (keep ID, state, type re-exports)
- Delete SOURCE_ID from stake, ALT, and config


shared/runtime/program/lib.zig
```diff
@@ -56,19 +55,28 @@ const Program = struct {
 };
 
 // zig fmt: off
-pub const NATIVE = sig.utils.pht(Program, &.{
+const native_entries = [_]struct { sig.core.Pubkey, Program }{
```
hamza-syndica: @yewman could't configure pht for this updated map, it was failing if I kept using pht for this. Any suggestion on this?

hamza-syndica: https://github.com/Syndica/sig/pull/1641/commits/84af917d303af29405e4e32caefcb82b6cb0b442

shared/runtime/program/config/lib.zig
```diff
@@ -2,6 +2,5 @@ const sig = @import("../../../lib.zig");
 const Pubkey = sig.core.Pubkey;
 
 pub const ID: Pubkey = .parse("Config1111111111111111111111111111111111111");
-pub const SOURCE_ID: Pubkey = .parse("BuafH9fBv62u6XjzrzS4ZjAE8963ejqF5rt1f8Uga4Q3");
 
 pub const COMPUTE_UNITS = 450;
```
dnut: This can be removed

shared/utils/pht.zig
```diff
@@ -58,9 +58,17 @@ pub fn pht(V: type, entries: []const struct { Pubkey, V }) type {
             }
         };
 
-        fn hash(x: *const Pubkey, key: u8) Hash {
+        fn hash(x: *const Pubkey, key: u32) Hash {
             const int: u32 = @bitCast(x.data[unique..][0..window].*);
-            const result = @as(u64, key) *% int;
+            // SplitMix64 finalizer over (key, int) — `key *% int` alone has
+            // too little mixing to reliably produce a perfect hash for
+            // small entry sets.
```
dnut: we don't need comments to refer to the old implementation.

shared/runtime/program/address_lookup_table/lib.zig
```diff
@@ -20,9 +19,6 @@ pub const ProgramState = state.ProgramState;
 const Pubkey = sig.core.Pubkey;
 
 pub const ID: Pubkey = .parse("AddressLookupTab1e1111111111111111111111111");
-pub const SOURCE_ID: Pubkey = .parse("AhXWrD9BBUYcKjtpA3zuiiZG4ysbo6C6wjHo1QhERk6A");
-
-pub const execute = @import("execute.zig").execute;
 
 // https://github.com/anza-xyz/agave/blob/8116c10021f09c806159852f65d37ffe6d5a118e/programs/address-lookup-table/src/processor.rs#L23
 pub const COMPUTE_UNITS = 750;
```
dnut: this can be removed

dnut (changes requested):

yewman (changes requested):
A few missed opportunities for dead-code removal.

Stake (`shared/runtime/program/stake/state.zig`):
- `Meta.setLockup` (L36-58) — reachable only from the deleted executor; drops last `instruction.LockupArgs` use inside `state.zig`.
- `Stake.split` (L66-82) — dangling `program.StakeError` ref, no external callers.
- `Stake.deactivate` (L85-90) — dangling `program.StakeError` return type, no external callers.
- `Authorized.authorize` (L138-178) + `Authorized.check` (L120-134) — third `program.StakeError` site; `check` only called by `authorize`.
- `Authorized.equals` (L180-183) and `Lockup.equals` (L208-212) — zero callers (all `.equals(` hits elsewhere are `Pubkey.equals`).
- `StakeFlags.combine` (L109-111) — zero callers (`StakeFlags.EMPTY` stays, used by `parse_stake.zig`).
- Follow-up: after the above, `const instruction = @import("instruction.zig")` at `state.zig:L10` becomes unused and can go, and all three dangling `program.StakeError` references disappear without needing to add back a `StakeError` stub.

ALT (`shared/runtime/program/address_lookup_table/`):
- `pub const program = @import("lib.zig")` (`lib.zig:L10`) — bogus self-import, zero refs (same shape as the `pub const state = @import("stake/lib.zig")` this PR already removed from `program/lib.zig`).
- `pub const instruction = @import("instruction.zig")` (`lib.zig:L12`) — module alias, zero refs; `Instruction` is already re-exported directly on L14. (The parallel `pub const state = @import("state.zig")` at L11 must stay — live via `address_lookup_table.state.LOOKUP_TABLE_META_SIZE`.)
- `pub const COMPUTE_UNITS = 750` (`lib.zig:L24`) — zero internal or external refs; parallel to the `SOURCE_ID`s this PR did remove.
- `LookupTableMeta.new` (`state.zig:L60-64`) — zero callers; `LookupTableMeta` is always built via struct-literal init.
- `AccountIndex` scaffolding in `instruction.zig` — private `AccountIndexWithPayer`/`AccountIndexNoPayer` enums (L52-61) and the five `pub const AccountIndex = ...` decls on the variant structs (L76, L80, L87, L91, L95). The variant structs themselves stay (bincode payloads), but `.AccountIndex` on them is never referenced anywhere post-executor.

Also worth updating the PR body — it still says "Replace PHT with linear scan for NATIVE", but the final commit (`applied pht fix from Harold`) kept the PHT and instead swapped in a SplitMix64 finalizer.

shared/utils/pht.zig
hamza-syndica: https://github.com/Syndica/sig/pull/1641/commits/1e5e9f44cdd8673be82f47cac5aac8f834570b27

shared/runtime/program/address_lookup_table/lib.zig
hamza-syndica: https://github.com/Syndica/sig/pull/1641/changes/be8c04a6fbc9984078860a6dfad7a16c3efad581

shared/runtime/program/config/lib.zig
hamza-syndica: https://github.com/Syndica/sig/pull/1641/changes/4f1c11261b4ab8cb133c0f1e6f0b94b2709501ea

shared/runtime/program/address_lookup_table/instruction.zig
```diff
@@ -72,25 +61,15 @@ pub const CreateLookupTable = struct {
...
 pub const ExtendLookupTable = struct {
     new_addresses: []const Pubkey,
-
-    pub const AccountIndex = AccountIndexWithPayer;
 };
 
-pub const DeactivateLookupTable = struct {
-    pub const AccountIndex = AccountIndexNoPayer;
-};
+pub const DeactivateLookupTable = struct {};
```
yewman: We should remove these empty structs, no need for the instruction variants to carry an named empty struct

yewman (commented):
We should remove the empty structs. Additionally I think the address_lookup_table, and stake modules could be defined just within a lib.zig similar to the config program. Ultimately I think we should separate the native program interfaces from their validator implementations however that can be left for a follow up PR.

shared/runtime/program/address_lookup_table/instruction.zig
hamza-syndica: https://github.com/Syndica/sig/pull/1641/commits/dabcbb544aca47a4ce00d7ce3bc20b48cdcc27c1

hamza-syndica: https://github.com/Syndica/sig/pull/1641/commits/85b0250df934e367ccfd540cd3918bf710f3f323

Merged into one file(s)

yewman (approved):

dnut (approved):

PR #1642, opened 2026-06-30
feat(v2:replay,exec): implement status cache parsing & state

InKryption:


v2/lib/solana/snapshot.zig
```diff
@@ -695,19 +1091,53 @@ pub fn TarZstIter(comptime BufReader: type) type {
             }
         }
 
-        // std.Io.Reader-like API for Manifest/StatusCache.read()
+        /// Zero-sized std.Io.Reader-like mixin API for Manifest/StatusCache's `.read()`.
+        /// Should never be copied out of the iterator, its address is used to reference the
+        /// iterator.
+        pub const ReaderMixin = struct {
+            zst: void align(@alignOf(Self)) = {},
```
Sobeston, line 1146: I don't see a problem with it, just kinda curious how you came up with this

v2/lib/replay.zig
```diff
@@ -73,4 +76,217 @@ pub const ExecResponse = extern struct {
...
+
+    pub const Entry = extern struct {
+        transactions: TxSegmentNode.Id.Optional,
+    };
+
+    pub const TxSegmentPool = collections.SharedPool(
+        TxSegmentNode,
+        BlockPool.capacity * 1000 / TxSegmentNode.tx_per_segment,
+    );
+
```
Sobeston, line 126: Not sure I understand the sizing here, what is it derived from?

v2/lib/solana/snapshot.zig
InKryption: It was just something I remembered from one of the comments mlugg made on either the PR or issue about removing `usingnamespace`. In the example he was addressing the "generic mixin" use case, I just applied the same pattern, but concretely, and purely for the purposes of organization.

v2/lib/replay.zig
InKryption: It is essentially a block-to-transaction ratio of approximately 1:1000. The division by tx_per_segment is because the transaction keys come in chunks of of that amount.

I chose 1000 at random early on because it's a relatively high upper bound. Could probably set it to a lower value.

v2/lib/solana/transaction.zig
```diff
@@ -356,6 +359,18 @@ pub const VersionedMessage = union(enum) {
...
+        var buffer: [1232]u8 = undefined;
+        var hashing: std.Io.Writer.Hashing(std.crypto.hash.Blake3) = .init(&buffer);
+        hashing.hasher.update("solana-tx-message-v1");
+        self.bincodeWrite(&hashing.writer) catch |err| switch (err) {
+            error.WriteFailed => unreachable,
+        };
+        var result: Hash = .{ .data = undefined };
+        hashing.hasher.final(&result.data);
+        return result;
+    }
```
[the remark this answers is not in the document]
InKryption: This is nonsense, the VersionedMessage type's hash is computed using Blake3

v2/lib/replay.zig
Sobeston: Ah okay, maybe for now just leave a comment? Just note that this is an estimate + we should figure out the real upper bound later

v2/lib/solana/snapshot.zig
Sobeston: That's fair, also curious about the align(alignof(self)) part, does that do anything here?

InKryption: Yeah, it makes it so it's aligned the same as the containing struct, which means it doesn't need `@alignCast` to use `@fieldParentPtr`.

v2/lib/replay.zig
InKryption: Comment added

dnut: Blocks on mainnet rarely go under 1000 transactions (if ever). 10,000 probably isn't even safe to use here but it should probably work most of the time. I wouldn't go under 10,000 as a starting point.

here's an issue to track the task of figuring out the limits: https://github.com/Syndica/sig/issues/1740

Could you format the comment like this please:

```zig
TODO(1740): ...
```

Going forward, I think any TODOs that get merged to main should also correspond with an issue in the tracker

InKryption: c183e6b76e80b822f3468463069fbbfc25e9059d

v2/init/services.zig
```diff
@@ -16,6 +16,7 @@ pub const accounts_db = struct {
         config: *lib.accounts_db.RootedConfig,
         ready_snapshot_in: *lib.snapshot.SnapshotData,
         snapshot_metadata_out: *lib.accounts_db.RuntimeMetadata,
+        status_cache_updates_out: *lib.accounts_db.StatusCacheUpdates,
```
dnut: Why is this a separate region? I assumed it would be part of RuntimeMetadata. @kprotty what do you think?

v2/lib/replay.zig
```diff
@@ -85,4 +85,237 @@ pub const ExecResponse = extern struct {
 
 pub const TxExecResult = extern struct {
     success: bool,
+    tx_hash: solana.Hash,
 };
+
+/// Represents a minimal subset of what is known as `StatusCache` in agave.
+/// Stores the status of all active transactions that are potentially/recently executed across relevant forks.
```
dnut: The meaning of this struct is not self-evident, so I would put a little more context here. For example:

```suggestion
/// Stores hashes of transactions that have been recently executed. 
///
/// Represents a minimal subset of what is known as `StatusCache` in agave.
///
/// Transactions are organized according to the block they specify as their
/// recent blockhash. For each transaction, we store which blocks it was
/// executed in. This can be more than one when there are forks. Transactions
/// are removed from the cache when the block they specify as their recent
/// blockhash is evicted.
```

Also, comments won't always be exempt from the 100 char limit. Please try to keep them shorter if you can.

v2/lib/replay.zig
```diff
@@ -85,4 +85,237 @@ pub const ExecResponse = extern struct {
...
+    const exec_registry = params.exec_registry;
+    const recent_blockhash = params.recent_blockhash;
+    const tx_key = params.tx_key;
+    const max_age = params.max_age;
+    const block_ref = bhq.getBlockRefIfValidForAge(recent_blockhash, max_age).opt() orelse {
+        return .unknown_blockhash;
+    };
+    if (exec_registry.containsTransaction(block_ref, tx_key)) {
+        return .already_processed;
+    } else {
```
dnut, line 343: You also need to check if the transaction was executed on the current fork. If it was executed on a competing fork, then it's not already processed. It looks like you're not storing any fork information in the ExecutionRegistry. I would store the BlockRef where the transaction was executed, so you can traverse the tree to see if it's in the same fork as the current BlockRef.

v2/lib/replay.zig
```diff
@@ -85,4 +85,237 @@ pub const ExecResponse = extern struct {
...
+            const block = block_ref.constPtr(block_pool);
+            if (block.slot <= min_root) exec_registry.evictBlock(block_ref);
+        }
+    }
+};
+
+pub const BlockhashQueueView = struct {
+    latest_block: BlockRef,
+    pool: *const BlockPool,
+    hashes: *const [BlockPool.capacity]?solana.Hash,
```
dnut, line 290: why is this a struct? don't these only need to be bundled up as the function parameters to lookup a block hash?

v2/services/replay.zig
```diff
@@ -325,6 +354,21 @@ fn bootstrap(
...
+            .pool = block_pool,
+            .hashes = blockhash_states,
+        };
+        for (exec_reg_updates.items) |*update| {
+            const block_ref = bhq_view.getBlockRefIfValidForAge(
+                &update.status_map_entry.hash,
+                max_age,
+            ).opt() orelse continue;
+            try exec_registry.insert(block_ref, &update.status_map_entry.status.key_slice);
+        }
```
dnut: Why are you using an intermediary list when you can just read from the buffer and write it directly into the registry here, after the blockhashes are populated?

v2/services/replay.zig
```diff
@@ -325,6 +354,21 @@ fn bootstrap(
...
+        const bhq_view: replay.BlockhashQueueView = .{
+            .latest_block = root_block,
+            .pool = block_pool,
+            .hashes = blockhash_states,
+        };
+        for (exec_reg_updates.items) |*update| {
+            const block_ref = bhq_view.getBlockRefIfValidForAge(
+                &update.status_map_entry.hash,
+                max_age,
+            ).opt() orelse continue;
```
dnut: i would expect the status cache and the blockhash queue to be consistent. i'm leaning towards this returning an error because it means the snapshot is inconsistent. it should at least be logged.

v2/init/services.zig
InKryption: This is part of the awkwardness I had mentioned. Originally I wanted to integrate it into the snapshot metadata struct, but reading the metadata happens at an earlier step than when the block hash queue is read, in a way that would separate the writes of the snapshot metadata output in a way that's less than ideal.

v2/lib/replay.zig
InKryption: `containsTransaction` inherently checks whether or not it's on the `block_ref` fork, by way of `block_ref` referring to a given ancestry; ExecutionRegistry doesn't need to store anything about alternate forks, because the relationship with forking is established through the block pool.

v2/services/replay.zig
InKryption: Follow-up to the awkwardness mentioned before: on the other side, the status cache is what is encountered first, followed then by the blockhash queue. If we read the blockhash queue first, the ring for the status cache updates could fill up and block indefinitely without the other side consuming it in the same sequence. So currently it has to first read status cache updates, followed by the blockhash queue. But we need the blockhash queue populated before the execution registry can be populated. And thus I ended up having to buffer the status cache updates here.

v2/lib/replay.zig
InKryption: Just felt correct in the moment for consistency; I can inverse the relationship to be a free function & a struct of parameters, it's not a significant difference.

v2/services/replay.zig
InKryption: 15b01eeb5434cb33dd6f18c75636d66ae392d654

v2/lib/replay.zig
InKryption: 8d24826a16abd579c2cc6da3fcc629be60aa580b

v2/lib/replay.zig
dnut: The block ref you're talking about is the block ref for the recent blockhash, which is insufficient for determining the ancestry of the block that contains the transaction. You also need to track the block refs that contain the transaction.

v2/services/replay.zig
dnut: writing the blockhash queue first seems like a simpler solution

v2/init/services.zig
dnut: i don't see why it demands a separate memory region though. i think you could still achieve the same thing but with just one region

v2/lib/solana/snapshot.zig
```diff
@@ -519,53 +825,118 @@ fn discardVoteAccounts(r: anytype) !void {
...
-            self.account_file_slot = 0;
-            self.account_data_len = 0;
-            self.account_data_padding = 0;
-            return self;
+        fn sendStatusCacheUpdateBlocking(
+            runner: lib.runner.Connection,
+            updates: *lib.accounts_db.StatusCacheUpdates.Iterator(.writer),
+            update: lib.accounts_db.StatusCacheUpdate,
+        ) !void {
+            while (true) {
```
yewman: inf loop

v2/services/replay.zig
```diff
@@ -176,39 +181,47 @@ pub fn serviceMain(runner: lib.runner.Connection, _: ReadOnly, rw: ReadWrite) !n
...
+                            "Slot {f} ({}) complete! ({}/{})",
+                            .{
+                                rw.block_pool.indexToPtr(block_ref).slot,
+                                block_ref,
+                                exec_state.n_transactions_requested,
+                                exec_state.n_transactions_completed,
+                            },
+                        );
+                        try exec_registry.insert(
+                            response_data.block_idx,
```
yewman: This should be the block index for the block associated with the transactions recent blockhash not the currently executing block. `bhq.getBlockRefIfValidForAge(recent_blockhash, max_age).opt()`

v2/services/replay.zig
```diff
@@ -176,39 +181,47 @@ pub fn serviceMain(runner: lib.runner.Connection, _: ReadOnly, rw: ReadWrite) !n
...
+                        logger.info().logf(
+                            "Slot {f} ({}) complete! ({}/{})",
+                            .{
+                                rw.block_pool.indexToPtr(block_ref).slot,
+                                block_ref,
+                                exec_state.n_transactions_requested,
+                                exec_state.n_transactions_completed,
+                            },
+                        );
+                        try exec_registry.insert(
```
yewman, line 228: This only inserts for the last transaction in the block given the exec_state.finished() gate? Am I missing something?

v2/services/replay.zig
```diff
@@ -176,39 +181,47 @@ pub fn serviceMain(runner: lib.runner.Connection, _: ReadOnly, rw: ReadWrite) !n
...
 
-                const account = rw.account_pool.getAccount(account_ref);
-                if (account.unref()) rw.account_pool.free(account_ref);
-            }
+                        const account = rw.account_pool.getAccount(account_ref);
+                        if (account.unref()) rw.account_pool.free(account_ref);
+                    }
 
-            defer rw.replay_transaction_pool.destroyId(response_data.tx_idx);
+                    defer rw.replay_transaction_pool.destroyId(response_data.tx_idx);
```
yewman: Double errdefer?

v2/services/replay.zig
```diff
@@ -292,16 +305,32 @@ pub fn serviceMain(runner: lib.runner.Connection, _: ReadOnly, rw: ReadWrite) !n
...
+    exec_registry: *lib.replay.ExecutionRegistry,
     exec_states: *BlockExecStates,
     blockhash_states: *BlockHashStates,
 ) !void {
     var num_hashes: usize = 0;
     // Drain the blockhash queue into the block tree. accountsdb writes into
     // this ring blocks waiting for the reader (us).
     var root_block = bhq: {
+        const ExecutionRegistryUpdate = lib.accounts_db.StatusCacheUpdate;
+        var exec_reg_updates_buf: [512]ExecutionRegistryUpdate = undefined;
```
yewman: Where does 512 come from and are we confident that it won't be exceeded? 
Instead do you think the following could work?
```
var root_block = bhq: {
    // (1) drain blockhash queue FIRST
    var blockhashes_in = snapshot_metadata.blockhash_queue.hashes.getView(.reader);
    defer blockhashes_in.close();
    var last_block: ?BlockRef = null;
    while (true) {
        const hashes = try blockhashes_in.getBufferBlocking(runner);
        if (hashes.len == 0) break;
        for (hashes) |*hash| { … build chain, populate blockhash_states … }
        blockhashes_in.advance(hashes.len);
    }
    const root_block = last_block orelse return error.NoBlockhashesInSnapshot;

    // (2) construct the view — the state it needs is now complete
    const bhq_view: replay.BlockhashQueueView = .{
        .latest_block = root_block,
        .pool = block_pool,
        .hashes = blockhash_states,
    };

    // (3) stream status cache and insert as we go — no buffer needed
    var status_cache_updates = status_cache_updates_in.getView(.reader);
    defer status_cache_updates.close();
    while (true) {
        const updates = try status_cache_updates.getBufferBlocking(runner);
        if (updates.len == 0) break;
        for (updates) |*update| {
            const block_ref = bhq_view.getBlockRefIfValidForAge(
                &update.status_map_entry.hash, 300,
            ).opt() orelse {
                logger.err().logf("…", .{});
                return error.InconsistentSnapshotData;
            };
            try exec_registry.insert(block_ref, &update.status_map_entry.status.key_slice);
        }
        status_cache_updates.advance(updates.len);
    }

    break :bhq root_block;
};
```

v2/services/replay.zig
InKryption: Yeah, I fixed this in a recent slew of changes I just pushed

v2/lib/solana/snapshot.zig
InKryption: Whoop, good catch haha. But a lot of this code is probably going to go by the wayside after rebasing on King's changes.

v2/services/replay.zig
InKryption: Yeah, accidentally reverted the fix that I had for this during some finagling, it's back in place properly now.

v2/services/replay.zig
InKryption: Aye good catch, silly rebase mistake.

v2/services/replay.zig
InKryption: In current design can't do this, can't start consuming the blockhash queue ring until we've consumed the status cache update ring, since that's the order in which the other side starts transmitting them.
This too shall be smoothed over once King's snapshot metadata PR is merged

yewman: Perhaps a brief comment or TODO if thats the case would be good :)

InKryption: No longer exists in latest version of this changeset

PR #1645, opened 2026-06-30
test(shred-stream): no output from passing unit tests

dnut:
The shred-stream unit tests print a bunch of output when running the tests. This clutters the logs and makes it hard to understand what's going on when the tests run - especially when a test fails. Passing tests should never print anything. Printing is reserved for failing tests.

The shred-stream tool was already using a stdout writer in a few places. I propagated the pattern further until it encompassed all messages printed by the tests. The tests use a discarding writer. If they need to make assertions about the error message, this can also now be added easily by giving each test its own writer, but that's not in scope for this PR.

### DIFF
--- a/v2/scripts/shred_stream.zig
+++ b/v2/scripts/shred_stream.zig
@@ -140,13 +140,13 @@ const PartialConfig = struct {
     corrupt_bytes: ?usize = null,
     dry_run: bool = false,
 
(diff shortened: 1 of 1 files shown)


prestonsn (approved):
Great changes! Thank you for cleaning the logging up 🫡

yewman (approved):

PR #1660, opened 2026-07-01
feat(v2:shred): transaction sanitize and verify ticks

yewman:
## `v2/solana`: parse-level transaction + tick validation primitives

Adds three parse-time validation primitives the Solana protocol applies to every transaction and every slot's tick chain. Pure v2 library additions, no callers yet — the `sol_compat_shred_parse_v1` differential harness wires them in a later PR.

### Changes

- **`verify_ticks`** (verify_ticks.zig) — enforces the three slot-level tick invariants: exact tick count for the window, final entry is a tick once full, inter-tick hash counts equal `hashes_per_tick`. Returns `BlockError` with the five failure modes. Lives under solana because v2 `Entry` uses `bincode.Vec(VersionedTransaction)` where v1 uses `[]const Transaction`.

- **`VersionedTransaction.serializedSize`** — analytical bincode size (no round-trip, allocation-free). Backs the MTU rule `size > MAX_BYTES` (1232). Per-component size functions co-located with each type.

- **`VersionedTransaction.validateAccountLocks`** — pairwise-distinct check on static `account_keys`. Duplicates reject as `AccountLoadedTwice`. Static keys only; ALT-resolved entries must be re-checked post-resolution.

- **`VersionedTransaction.sanitize`** — 13 numbered structural invariants (signature/header agreement, instruction indices, ALT bookkeeping, …). Each documented inline and individually tested. MTU and duplicate-key checks deliberately live in the two methods above — treating them as `sanitize` errors would conflate parse and account-locking stages.

### DIFF
--- a/v2/lib/solana/transaction.zig
+++ b/v2/lib/solana/transaction.zig
@@ -22,6 +22,51 @@ pub const Entry = struct {
 pub const VersionedTransaction = struct {
     signatures: bincode.ShortVec(Signature),
     message: VersionedMessage,
+
+    /// Maximum on-the-wire size of a serialised transaction. Equal to the
+    /// Solana packet MTU; the network refuses anything larger.
+    pub const MAX_BYTES: usize = 1232;
+
+    /// Total bincode-serialised size in bytes (signatures + message).
+    ///
+    /// Computed analytically (compact-u16 length prefixes + fixed-size
+    /// fields + variable-size payloads) rather than by round-tripping the
+    /// structure through `bincode.write`. This also sidesteps the latent
+    /// `VersionedMessage.bincodeWrite` path (a `switch (self)` over a
+    /// pointer rather than the tag) which has never been exercised.
+    pub fn serializedSize(self: VersionedTransaction) usize {
+        var n: usize = 0;
+        n += compactU16Len(self.signatures.items.len);
+        n += 64 * self.signatures.items.len;
+        n += self.message.serializedSize();
+        return n;
+    }
+
+    /// Returns `true` iff the message's static `account_keys` are pairwise
+    /// distinct. Solana rejects transactions whose visible accounts contain
+    /// a duplicate (`AccountLoadedTwice`).
+    ///
+    /// Only the keys carried directly in the message are checked here;
+    /// entries resolved from address-lookup tables are invisible without
+    /// ALUT resolution and must be re-checked once those are loaded.
+    ///
+    /// O(n^2); `account_keys.len ≤ 128` by transaction MTU.
+    ///
+    /// [agave] https://github.com/anza-xyz/agave/blob/v4.1.0-rc.1/accounts-db/src/account_locks.rs#L143-L155
+    pub fn validateAccountLocks(self: VersionedTransaction) bool {
+        const keys = switch (self.message) {
+            .legacy => |m| m.account_keys.items,
+            .v0 => |m| m.account_keys.items,
+        };
+        var i: usize = 0;
+        while (i < keys.len) : (i += 1) {
+            var j: usize = i + 1;
+            while (j < keys.len) : (j += 1) {
+                if (std.mem.eql(u8, &keys[i].data, &keys[j].data)) return false;
+            }
+        }
+        return true;
+    }
 };
 
 pub const VersionedMessage = union(enum) {
@@ -96,6 +141,15 @@ pub const VersionedMessage = union(enum) {
             },
         }
     }
+
+    /// Total bincode-serialised size of the message in bytes, including the
+    /// `0x80` version-prefix byte for v0 messages.
+    pub fn serializedSize(self: VersionedMessage) usize {
+        return switch (self) {
+            .legacy => |m| m.serializedSize(),
+            .v0 => |m| 1 + m.serializedSize(),
+        };
+    }
 };
 
 pub const LegacyMessage = struct {
@@ -103,6 +157,15 @@ pub const LegacyMessage = struct {
     account_keys: bincode.ShortVec(Pubkey),
     recent_blockhash: Hash,
     instructions: bincode.ShortVec(CompiledInstruction),
+
+    pub fn serializedSize(self: LegacyMessage) usize {
+        var n: usize = 3; // MessageHeader (3 u8 fields)
+        n += compactU16Len(self.account_keys.items.len) + 32 * self.account_keys.items.len;
+        n += 32; // recent_blockhash
+        n += compactU16Len(self.instructions.items.len);
+        for (self.instructions.items) |ix| n += ix.serializedSize();
+        return n;
+    }
 };
 
 pub const V0Message = struct {
@@ -111,6 +174,19 @@ pub const V0Message = struct {
     recent_blockhash: Hash,
     instructions: bincode.ShortVec(CompiledInstruction),
     address_table_lookups: bincode.ShortVec(AddressLookup),
+
+    /// Size of the v0 message body, *excluding* the `0x80` version-prefix
+    /// byte (the prefix is owned by `VersionedMessage.serializedSize`).
+    pub fn serializedSize(self: V0Message) usize {
+        var n: usize = 3; // MessageHeader
+        n += compactU16Len(self.account_keys.items.len) + 32 * self.account_keys.items.len;
+        n += 32; // recent_blockhash
+        n += compactU16Len(self.instructions.items.len);
+        for (self.instructions.items) |ix| n += ix.serializedSize();
+        n += compactU16Len(self.address_table_lookups.items.len);
+        for (self.address_table_lookups.items) |alt| n += alt.serializedSize();
+        return n;
+    }
 };
 
 pub const MessageHeader = struct {
@@ -123,10 +199,32 @@ pub const CompiledInstruction = struct {
     program_id_index: u8,
     accounts: bincode.ShortVec(u8),
     data: bincode.ShortVec(u8),
+
+    pub fn serializedSize(self: CompiledInstruction) usize {
+        var n: usize = 1; // program_id_index
+        n += compactU16Len(self.accounts.items.len) + self.accounts.items.len;
+        n += compactU16Len(self.data.items.len) + self.data.items.len;
+        return n;
+    }
 };
 
 pub const AddressLookup = struct {
     account_key: Pubkey,
     writable_indexes: bincode.ShortVec(u8),
     readonly_indexes: bincode.ShortVec(u8),
+
+    pub fn serializedSize(self: AddressLookup) usize {
+        var n: usize = 32; // account_key (Pubkey)
+        n += compactU16Len(self.writable_indexes.items.len) + self.writable_indexes.items.len;
+        n += compactU16Len(self.readonly_indexes.items.len) + self.readonly_indexes.items.len;
+        return n;
+    }
 };
+
+/// Byte length of the compact-u16 (short-vec) length prefix for `n`.
+/// Continuation-bit scheme: < 0x80 → 1 byte, < 0x4000 → 2, else 3.
+fn compactU16Len(n: usize) usize {
+    if (n < 0x80) return 1;
+    if (n < 0x4000) return 2;
+    return 3;
+}
--- a/v2/lib/solana/transaction.zig
+++ b/v2/lib/solana/transaction.zig
@@ -12,6 +12,11 @@ pub const Entry = struct {
     num_hashes: u64,
     hash: Hash,
     transactions: bincode.Vec(VersionedTransaction),
+
+    /// An entry is a "tick" iff it carries no transactions.
+    pub fn isTick(self: Entry) bool {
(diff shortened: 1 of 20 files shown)


v2/lib/solana/transaction.zig
```diff
@@ -12,11 +12,200 @@ pub const Entry = struct {
...
+    /// Inclusive maximum number of top-level instructions per transaction.
+    pub const MAX_INSTRUCTIONS: usize = 64;
+
+    /// Total bincode-serialised size in bytes (signatures + message).
+    ///
+    /// Computed analytically (compact-u16 length prefixes + fixed-size
+    /// fields + variable-size payloads) rather than by round-tripping the
+    /// structure through `bincode.write`. This also sidesteps the latent
+    /// `VersionedMessage.bincodeWrite` path (a `switch (self)` over a
+    /// pointer rather than the tag) which has never been exercised.
```
InKryption: What is this in reference? I can't seem to find the `switch (self)` branch this is talking about.

v2/lib/solana/transaction.zig
```diff
@@ -12,11 +12,200 @@ pub const Entry = struct {
...
+            .legacy => |m| m.account_keys.items,
+            .v0 => |m| m.account_keys.items,
+        };
+        var i: usize = 0;
+        while (i < keys.len) : (i += 1) {
+            var j: usize = i + 1;
+            while (j < keys.len) : (j += 1) {
+                if (std.mem.eql(u8, &keys[i].data, &keys[j].data)) return false;
+            }
+        }
```
InKryption:
```suggestion
        for (0..keys.len - 1) |i| {
            for (i + 1..keys.len) |j| {
                if (std.mem.eql(u8, &keys[i].data, &keys[j].data)) return false;
            }
        }
```

v2/lib/solana/transaction.zig
```diff
@@ -12,11 +12,200 @@ pub const Entry = struct {
...
+    /// |  8 | per instruction: `0 < program_id_index < account_keys.len` |
+    /// |  9 | `address_table_lookups.len ≤ V0Message.MAX_ADDR_TABLE_LOOKUPS` (v0 only) |
+    /// | 10 | per ALT: `writable_indexes.len + readonly_indexes.len ≥ 1` |
+    /// | 11 | per ALT: each of `writable_indexes.len`, `readonly_indexes.len ≤ MAX_ACCOUNT_ADDRESSES - account_keys.len` |
+    /// | 12 | `account_keys.len + Σ ALT-loaded ≤ MAX_ACCOUNT_ADDRESSES` |
+    /// | 13 | max account index used by any instruction `< account_keys.len + Σ ALT-loaded` |
+    ///
+    /// [firedancer] https://github.com/firedancer-io/firedancer/blob/main/src/ballet/txn/fd_txn_parse.c
+    /// [agave] https://github.com/anza-xyz/agave/blob/v4.1.0-rc.1/transaction-view/src/sanitize.rs
+    pub fn sanitize(self: VersionedTransaction) bool {
```
InKryption: Would there be any value in changing this from a boolean to an enum? The enum could then represent each "invariant", with the relevant member being annotated with the doc comment here, acting as the "numbering" directly in the code, instead of having to manually maintain these "comment links".

v2/lib/solana/transaction.zig
```diff
@@ -12,11 +12,200 @@ pub const Entry = struct {
...
+
+        // (8) per-instruction: program_id_index is a static account, but
+        // not the fee payer. Also collect the max account index referenced
+        // for check 13.
+        var max_acct_idx: usize = 0;
+        for (instructions) |ix| {
+            const pid: usize = ix.program_id_index;
+            if (pid == 0 or pid >= acct_cnt) return false;
+            for (ix.accounts.items) |a| {
+                if (@as(usize, a) > max_acct_idx) max_acct_idx = a;
```
InKryption: Any reason to not simply use a `u8` here?
```suggestion
        var max_acct_idx: u8 = 0;
        for (instructions) |ix| {
            const pid: usize = ix.program_id_index;
            if (pid == 0 or pid >= acct_cnt) return false;
            for (ix.accounts.items) |a| {
                max_acct_idx = @max(max_acct_idx, a);
```

v2/lib/solana/transaction.zig
```diff
@@ -118,10 +341,335 @@ pub const CompiledInstruction = struct {
...
+        try self.instructions.append(self.allocator, .{
+            .program_id_index = program_id_index,
+            .accounts = .{ .items = accounts_copy },
+            .data = .{ .items = data_copy },
+        });
+    }
+
+    fn pushAlt(self: *Builder, writable: []const u8, readonly: []const u8) !void {
+        const w = try self.allocator.dupe(u8, writable);
+        const r = try self.allocator.dupe(u8, readonly);
```
InKryption, line 471:
```suggestion
        const w = try self.allocator.dupe(u8, writable);
        errdefer self.allocator.free(w);
        const r = try self.allocator.dupe(u8, readonly);
        errdefer self.allocator.free(r);
```

v2/lib/solana/transaction.zig
```diff
@@ -118,10 +341,335 @@ pub const CompiledInstruction = struct {
...
+    }
+
+    fn pushInstr(self: *Builder, program_id_index: u8, accounts: []const u8) !void {
+        const accounts_copy = try self.allocator.dupe(u8, accounts);
+        const data_copy = try self.allocator.alloc(u8, 0);
+        try self.instructions.append(self.allocator, .{
+            .program_id_index = program_id_index,
+            .accounts = .{ .items = accounts_copy },
+            .data = .{ .items = data_copy },
+        });
```
InKryption, line 465: No need to allocate 0 bytes, can initialize `[]T` with `&.{}`.
```suggestion
        const accounts_copy = try self.allocator.dupe(u8, accounts);
        errdefer self.allocator.free(accounts_copy);
        try self.instructions.append(self.allocator, .{
            .program_id_index = program_id_index,
            .accounts = .{ .items = accounts_copy },
            .data = .{ .items = &.{} },
        });
```

v2/lib/solana/transaction.zig
```diff
@@ -118,10 +341,335 @@ pub const CompiledInstruction = struct {
...
+    signatures: std.ArrayList(Signature) = .{},
+    header: MessageHeader = .{
+        .num_required_signatures = 1,
+        .num_readonly_signed_accounts = 0,
+        .num_readonly_unsigned_accounts = 1,
+    },
+    account_keys: std.ArrayList(Pubkey) = .{},
+    instructions: std.ArrayList(CompiledInstruction) = .{},
+    alts: std.ArrayList(AddressLookup) = .{},
+    is_v0: bool = false,
```
InKryption: With the way this is used, should maybe instead modeled as a parameter on the `build` method, like `builder.build(.v0)`.

v2/lib/solana/transaction.zig
```diff
@@ -118,10 +341,335 @@ pub const CompiledInstruction = struct {
...
+    allocator: std.mem.Allocator,
+    signatures: std.ArrayList(Signature) = .{},
+    header: MessageHeader = .{
+        .num_required_signatures = 1,
+        .num_readonly_signed_accounts = 0,
+        .num_readonly_unsigned_accounts = 1,
+    },
+    account_keys: std.ArrayList(Pubkey) = .{},
+    instructions: std.ArrayList(CompiledInstruction) = .{},
+    alts: std.ArrayList(AddressLookup) = .{},
```
InKryption: General note: let's get used to using the new init decls, because the default initializations are deprecated. i.e. `signatures: std.ArrayList(Signature) = .empty`.

InKryption (changes requested):

v2/lib/solana/verify_ticks.zig
```diff
@@ -0,0 +1,93 @@
...
+    if (params.hashes_per_tick > 0) {
+        var hash_count = params.tick_hash_count;
+        for (entries) |entry| {
+            hash_count = hash_count +| entry.num_hashes;
+            if (entry.isTick()) {
+                if (hash_count != params.hashes_per_tick) return error.InvalidTickHashCount;
+                hash_count = 0;
+            }
+        }
+        if (hash_count >= params.hashes_per_tick) return error.InvalidTickHashCount;
```
Sobeston: sanity check, agave takes a `tick_hash_count: &mut u64`, is it fine to just copy rather than mutate?

v2/lib/solana/transaction.zig
```diff
@@ -12,11 +12,200 @@ pub const Entry = struct {
...
+    /// the static caps, etc. Rejecting these here prevents malformed
+    /// transactions from reaching account-locking or message resolution
+    /// and keeps later stages from having to defend against shapes the
+    /// protocol forbids.
+    ///
+    /// Two related checks are intentionally not part of this method:
+    ///
+    ///   * MTU (`serializedSize ≤ MAX_BYTES`). Owned by the caller and
+    ///     usually enforced at the packet boundary; doing it again here
+    ///     would just be defensive.
```
Sobeston: How about we add an assert for this at the top?

v2/lib/solana/transaction.zig
```diff
@@ -12,11 +12,200 @@ pub const Entry = struct {
...
+        // not the fee payer. Also collect the max account index referenced
+        // for check 13.
+        var max_acct_idx: usize = 0;
+        for (instructions) |ix| {
+            const pid: usize = ix.program_id_index;
+            if (pid == 0 or pid >= acct_cnt) return false;
+            for (ix.accounts.items) |a| {
+                if (@as(usize, a) > max_acct_idx) max_acct_idx = a;
+            }
+        }
```
Sobeston: I think we're not fully supporting SIMD 406 here, not a huge deal as it's not active yet

https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0406-maximum-instruction-accounts.md

Sobeston (changes requested):

v2/lib/solana/transaction.zig
yewman: Thanks for highlighting, I should have fixed it directly. Added a round trip unit test and updated the switch 
8c495f7ca087d9f334083f0089e1c5c964771e13

v2/lib/solana/transaction.zig
yewman: fb08f9dc65770d7d7c3d677bb68753d0a9620d3f

v2/lib/solana/transaction.zig
yewman: fb08f9dc65770d7d7c3d677bb68753d0a9620d3f

v2/lib/solana/transaction.zig
yewman: fb08f9dc65770d7d7c3d677bb68753d0a9620d3f

v2/lib/solana/transaction.zig
yewman: fb08f9dc65770d7d7c3d677bb68753d0a9620d3f

v2/lib/solana/transaction.zig
yewman: fb08f9dc65770d7d7c3d677bb68753d0a9620d3f

v2/lib/solana/transaction.zig
yewman: 5b125ccef7f0bbdf9de7cee0975953faa4406271

v2/lib/solana/transaction.zig
yewman: Good call, I like it! 15ece62d0930dc93437ec44f6d2319a7d7b55b3a

v2/lib/solana/transaction.zig
yewman: If we use `std.debug.assert` we may as well just have the check and return an error, since it will run in debug and release. This brings back the concept of having a debug assert which is omitted in ReleaseSafe. What are your thoughts?

v2/lib/solana/transaction.zig
InKryption: Making it an actual error makes sense, that way can also get the stack trace for it in some contexts; not sure why that didn't occur to me lol.

InKryption (approved):

v2/lib/solana/verify_ticks.zig
yewman: Fair, it wasn't required for shred parse harness but likely will be in v2, updated to use mutable u64 as in v1. aac674bc5f0657270a0898ef67a616be2b467815

v2/lib/solana/transaction.zig
yewman: Here is the parse update we discussed. Lmk your thoughts c69859d7d95f38f2d5e0e029f9f6d4d2c503710a

InKryption (dismissed):

Sobeston (dismissed):

prestonsn (dismissed):

PR #1662, opened 2026-07-01
fix(logger): misc fixes and ergonomics improvements

dnut:
- fix pointer cast mut->const->mut compile error. this made it impossible to use the `field` function with mutable pointers
- remove quoted/literal distinction from logger api. only the presence of whitespace should determine whether fields are quoted. this should not be customizable by the user.
- use a default formatter for `.field()` and add `.fieldFmt()` for when you want to customize the formatter. previously this function was confusing to use. ergonomics is very important for code as widely used as a logger. the default formatters will be the right answer in almost every case. i doubt there will be many exceptions, if any.
- add `log` function for when there are no args
- start using `field` and `log` in a few places


v2/lib/telemetry.zig
```diff
@@ -277,18 +277,68 @@ pub fn Logger(comptime scope_str: []const u8) type {
...
             return struct {
                 logger: LoggerSelf,
-                level: log.Level,
-                entries: [entry_count]log.EntryField,
+                level: log_zig.Level,
+                entries: [entry_count]log_zig.EntryField,
                 const EntrySelf = @This();
 
+                /// Formats using the default formatter for the type, if it exists. To
+                /// customize the format, use `fromValue` or `fromFmt`.
```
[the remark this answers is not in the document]
dnut: [4895533](https://github.com/Syndica/sig/pull/1662/commits/48955338268ec40f202f4ee8fe1fb12fe21377e9)

v2/lib/telemetry/log.zig
```diff
@@ -592,66 +590,52 @@ pub const EntryField = struct {
...
+                const value: @Type(ptr_info) = if (is_comptime_known)
                     value_ptr
                 else
                     @ptrCast(@alignCast(ptr));
-                try w.writeAll(surround.str());
+                const has_whitespace = WhitespaceDetector.detect(fmt_str, .{value.*});
+                if (has_whitespace) try w.writeAll("\"");
                 try w.print(fmt_str, .{value.*});
-                try w.writeAll(surround.str());
+                if (has_whitespace) try w.writeAll("\"");
```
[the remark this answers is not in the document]
dnut: This is complex to optimize and I don't think it's worth prioritizing at this point.

v2/lib/telemetry/log.zig
```diff
@@ -660,26 +644,24 @@ pub const EntryValueFmt = struct {
...
                 ptr: *const anyopaque,
                 w: *std.Io.Writer,
             ) std.Io.Writer.Error!void {
                 const args: @TypeOf(args_ptr) = @ptrCast(@alignCast(ptr));
-                try w.writeAll(surround.str());
+                const has_whitespace = WhitespaceDetector.detect(fmt_str, args.*);
+                if (has_whitespace) try w.writeAll("\"");
                 try w.print(fmt_str, args.*);
-                try w.writeAll(surround.str());
+                if (has_whitespace) try w.writeAll("\"");
```
InKryption:
```suggestion
                if (has_whitespace) try w.writeByte('"');
try w.print(fmt_str, args.*);
                if (has_whitespace) try w.writeByte('"');
```

v2/lib/telemetry/log.zig
```diff
@@ -839,3 +821,100 @@ test Iso8601Fmt {
...
+
+    pub fn init() WhitespaceDetector {
+        return .{
+            .writer = .{
+                .vtable = &.{ .drain = drain },
+                .buffer = &.{},
+            },
+            .found = false,
+        };
+    }
```
InKryption, line 840:
```suggestion
    pub fn init(buffer: []u8) WhitespaceDetector {
        return .{
            .writer = .{
                .vtable = &.{ .drain = drain },
                .buffer = buffer,
            },
            .found = false,
        };
    }
```
Never any harm in taking a buffer, the caller can always just pass `&.{}` if they don't care. Though passing a buffer will help to amortize indirect calls.

v2/lib/telemetry/log.zig
```diff
@@ -839,3 +821,100 @@ test Iso8601Fmt {
...
+    {
+        var wd: WhitespaceDetector = .init();
+        try wd.writer.writeByte('\t');
+        try std.testing.expect(wd.found);
+    }
+    {
+        var wd: WhitespaceDetector = .init();
+        try wd.writer.writeByte('\n');
+        try std.testing.expect(wd.found);
+    }
```
InKryption:
```suggestion
    var wd: WhitespaceDetector = undefined;

    wd = .init();
    try wd.writer.writeAll("no-whitespace-here");
    try std.testing.expect(!wd.found);

    wd = .init();
    try wd.writer.writeAll("hello world");
    try std.testing.expect(wd.found);

    wd = .init();
    try wd.writer.writeAll("end ");
    try std.testing.expect(wd.found);

    wd = .init();
    try wd.writer.writeAll(" start");
    try std.testing.expect(wd.found);

    wd = .init();
    try wd.writer.print("value={d}", .{42});
    try std.testing.expect(!wd.found);

    wd = .init();
    try wd.writer.print("value = {d}", .{42});
    try std.testing.expect(wd.found);

    wd = .init();
    try wd.writer.print("value={s}", .{"hello world"});
    try std.testing.expect(wd.found);

    wd = .init();
    try wd.writer.writeByte('\t');
    try std.testing.expect(wd.found);

    wd = .init();
    try wd.writer.writeByte('\n');
    try std.testing.expect(wd.found);
```

InKryption (dismissed):

v2/lib/telemetry/log.zig
dnut: 5a2fffe99a58b5bbd13860a0a1da0b33a291987b

v2/lib/telemetry/log.zig
dnut: 5a2fffe99a58b5bbd13860a0a1da0b33a291987b

v2/lib/telemetry/log.zig
dnut: I don't see how using a buffer here could help. There's nothing expensive to amortize because it's a discarding writer. Using a buffer would introduce pointless copies for data that's being discarded. This also means that `found` won't be accurate until after calling `flush`. Accepting a buffer here just feels like a footgun to me.

PR #1663, opened 2026-07-02
fix(ci): grafana metrics configuration

ajw221:
# Intent

- Fix misconfigured grafana ci settings

# Implementation

- Removed `message` as a low-cardinality label
- Updated `log-error-alert` query expression to sum by `scope` for `count_over_time`

# Ramifications

- Currently alerts and logging are not properly configured in a way that meaningfully allows us to meaningfully observe the ci runner when an error occurs

# Tests

N/A

### DIFF
--- a/metrics/grafana/datasources/datasource.yml
+++ b/metrics/grafana/datasources/datasource.yml
@@ -1,13 +1,18 @@
 apiVersion: 1
 
+# NOTE: we need to pin the hardcoded UIDs used in the dashboard + alert_rules.yaml
+# to the datasources here, otherwise grafana will just assign a random UID on every
+# provision and every panel/alert will break with "Datasource not found".
 datasources:
   - name: Prometheus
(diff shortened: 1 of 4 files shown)


metrics/grafana/datasources/datasource.yml
```diff
@@ -16,5 +21,6 @@ datasources:
     editable: false
 
   - name: Pyroscope
+    uid: pyroscope
```
prestonsn, line 24: not an issue, but is there a reason we use a dedicated hardcoded uid for Pyroscope? (The others use `P8E80F9AEF21F6940`)?

prestonsn (commented):
Changes look good to me, I'm going to check if it :works-on-my-machine: before approving 

update: It works well, so just wanted to confirm the thing below before approving

dnut (approved):

prestonsn (approved):

PR #1664, opened 2026-07-02
fix(runtime): append ALT extend addresses after existing entries, allow full capacity

insky7:
# Intent

- fix two bugs in the native address lookup table program's `ExtendLookupTable`:
  - new addresses are written at the start of the address list, overwriting existing entries instead of appending after them
  - the capacity check rejects a valid extension to exactly 256 addresses

Fixes #1637
Fixes #1636

# Implementation

- offset each new address write by the existing address count, so `[A]` extended with `[B]` yields `[A, B]` instead of `[B, 0]`. match agave's `extend_from_slice`. only the `.len` of the deserialized table is used since the resize before the copy may reallocate the account datas buffer.
- changed the `new_table_addresses_len` check from `>=` to `>`, matching agave.

Both fixes are in the same PR because the second can't be tested without the first: extending a 255-entry table exercises the append path.

# Ramifications

- Extending a non-empty lookup table no longer corrupts the account data.
- Sig no longer rejects a 255 -> 256 extension that agave accepts.

# Tests

- `extend appends after an existing address`: extends a one-entry table, asserts the result is `[first, second]`. Fails on main with `DataMismatch`.
- `extend allows exactly 256 addresses`: extends a 255-entry table by one, asserts success with the address appended. Fails on main with `InvalidInstructionData`.

the existing extend test only covers extending an empty table, where the wrong offset (`i`) and the correct offset (`old_len + i`) coincide, which is why neither issue was caught.

/sig/shared/runtime/program/address_lookup_table$ zig build test -Dfilter="address-lookup-table" --summary all

Build Summary: 13/13 steps succeeded; 8/8 tests passed
test success
└─ run test shared 8 passed 36ms MaxRSS:6M
   └─ compile test shared Debug native cached 20ms MaxRSS:39M
      ├─ options cached
      ├─ run exe gen_feature_set_id (feature-set-id.zig) cached
      │  └─ compile exe gen_feature_set_id Debug native cached 12ms MaxRSS:38M
      ├─ translate-c cached 15ms MaxRSS:38M
      ├─ WriteFile cached
      ├─ compile lib secp256k1 Debug native cached 17ms MaxRSS:38M
      ├─ compile lib secp256k1 Debug native (reused)
      ├─ WriteFile table.zig cached
      │  ├─ run exe generator_chain cached
      │  │  └─ compile exe generator_chain Debug native cached 12ms MaxRSS:38M
      │  └─ run exe generator_chain (+1 more reused dependencies)
      └─ options cached

### DIFF
--- a/shared/runtime/program/address_lookup_table/execute.zig
+++ b/shared/runtime/program/address_lookup_table/execute.zig
@@ -694,7 +694,7 @@ test "address-lookup-table create" {
         },
         .{ .pubkey = unsigned_authority_address },
         .{ .pubkey = payer, .lamports = before_lamports, .owner = system_program.ID },
-        .{ .pubkey = program.ID, .owner = runtime.ids.NATIVE_LOADER_ID, .executable = true },
+        .{ .pubkey = ID, .owner = runtime.ids.NATIVE_LOADER_ID, .executable = true },
         .{
             .pubkey = runtime.program.system.ID,
             .owner = runtime.ids.NATIVE_LOADER_ID,
@@ -711,7 +711,7 @@ test "address-lookup-table create" {
         },
         .{ .pubkey = unsigned_authority_address },
         .{ .pubkey = payer, .lamports = after_lamports, .owner = system_program.ID },
-        .{ .pubkey = program.ID, .owner = runtime.ids.NATIVE_LOADER_ID, .executable = true },
+        .{ .pubkey = ID, .owner = runtime.ids.NATIVE_LOADER_ID, .executable = true },
         .{
             .pubkey = runtime.program.system.ID,
             .owner = runtime.ids.NATIVE_LOADER_ID,
@@ -803,7 +803,7 @@ test "address-lookup-table freeze" {
             .data = before_lookup_table,
         },
         .{ .pubkey = unsigned_authority_address },
-        .{ .pubkey = program.ID, .owner = runtime.ids.NATIVE_LOADER_ID, .executable = true },
+        .{ .pubkey = ID, .owner = runtime.ids.NATIVE_LOADER_ID, .executable = true },
         .{
             .pubkey = runtime.program.system.ID,
             .owner = runtime.ids.NATIVE_LOADER_ID,
@@ -819,7 +819,7 @@ test "address-lookup-table freeze" {
             .data = after_lookup_table,
         },
         .{ .pubkey = unsigned_authority_address },
-        .{ .pubkey = program.ID, .owner = runtime.ids.NATIVE_LOADER_ID, .executable = true },
+        .{ .pubkey = ID, .owner = runtime.ids.NATIVE_LOADER_ID, .executable = true },
         .{
             .pubkey = runtime.program.system.ID,
             .owner = runtime.ids.NATIVE_LOADER_ID,
@@ -912,7 +912,7 @@ test "address-lookup-table close" {
         },
         .{ .pubkey = unsigned_authority_address },
         .{ .pubkey = payer, .lamports = 0 },
-        .{ .pubkey = program.ID, .owner = runtime.ids.NATIVE_LOADER_ID, .executable = true },
+        .{ .pubkey = ID, .owner = runtime.ids.NATIVE_LOADER_ID, .executable = true },
         .{
             .pubkey = runtime.program.system.ID,
             .owner = runtime.ids.NATIVE_LOADER_ID,
@@ -929,7 +929,7 @@ test "address-lookup-table close" {
         },
         .{ .pubkey = unsigned_authority_address },
         .{ .pubkey = payer, .lamports = 100 },
-        .{ .pubkey = program.ID, .owner = runtime.ids.NATIVE_LOADER_ID, .executable = true },
+        .{ .pubkey = ID, .owner = runtime.ids.NATIVE_LOADER_ID, .executable = true },
         .{
             .pubkey = runtime.program.system.ID,
             .owner = runtime.ids.NATIVE_LOADER_ID,
@@ -1023,7 +1023,7 @@ test "address-lookup-table deactivate" {
             .data = before_lookup_table,
         },
         .{ .pubkey = unsigned_authority_address },
-        .{ .pubkey = program.ID, .owner = runtime.ids.NATIVE_LOADER_ID, .executable = true },
+        .{ .pubkey = ID, .owner = runtime.ids.NATIVE_LOADER_ID, .executable = true },
         .{
             .pubkey = runtime.program.system.ID,
             .owner = runtime.ids.NATIVE_LOADER_ID,
@@ -1039,7 +1039,7 @@ test "address-lookup-table deactivate" {
             .data = after_lookup_table,
         },
         .{ .pubkey = unsigned_authority_address },
-        .{ .pubkey = program.ID, .owner = runtime.ids.NATIVE_LOADER_ID, .executable = true },
+        .{ .pubkey = ID, .owner = runtime.ids.NATIVE_LOADER_ID, .executable = true },
         .{
             .pubkey = runtime.program.system.ID,
             .owner = runtime.ids.NATIVE_LOADER_ID,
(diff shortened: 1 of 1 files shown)


shared/runtime/program/address_lookup_table/execute.zig
```diff
@@ -1221,3 +1221,149 @@ test "address-lookup-table extend" {
...
+    const authority: Pubkey = .{ .data = [_]u8{0xA1} ** Pubkey.SIZE };
+    const payer: Pubkey = .{ .data = [_]u8{0xA2} ** Pubkey.SIZE };
+    const lookup_table_address: Pubkey = .{ .data = [_]u8{0xA3} ** Pubkey.SIZE };
+    const first_address: Pubkey = .{ .data = [_]u8{0x11} ** Pubkey.SIZE }; // already stored
+    const second_address: Pubkey = .{ .data = [_]u8{0x22} ** Pubkey.SIZE }; // appended
+
+    const meta_state: state.ProgramState = .{ .LookupTable = state.LookupTableMeta.new(authority) };
+
+    // the table already holds one address: first_address.
+    const before_lookup_table = try allocator.alloc(u8, LOOKUP_TABLE_META_SIZE + @sizeOf(Pubkey));
```
ajw221: while not currently reflected or enforced in the codebase, we should stick with one of these conventions instead of using them both interchangeably

```suggestion
    const before_lookup_table = try allocator.alloc(u8, LOOKUP_TABLE_META_SIZE + Pubkey.SIZE);
```

shared/runtime/program/address_lookup_table/execute.zig
```diff
@@ -1221,3 +1221,149 @@ test "address-lookup-table extend" {
...
+    defer allocator.free(before_lookup_table);
+    _ = try sig.bincode.writeToSlice(before_lookup_table[0..LOOKUP_TABLE_META_SIZE], meta_state, .{});
+    @memcpy(before_lookup_table[LOOKUP_TABLE_META_SIZE..], &first_address.data);
+
+    // after extend the table must hold [first, second].
+    const after_lookup_table = try allocator.alloc(u8, LOOKUP_TABLE_META_SIZE + 2 * @sizeOf(Pubkey));
```
ajw221: same as previously mentioned
```suggestion
    const after_lookup_table = try allocator.alloc(u8, LOOKUP_TABLE_META_SIZE + 2 * Pubkey.SIZE);
```

shared/runtime/program/address_lookup_table/execute.zig
```diff
@@ -1221,3 +1221,149 @@ test "address-lookup-table extend" {
...
+    defer allocator.free(after_lookup_table);
+    @memcpy(after_lookup_table[0..LOOKUP_TABLE_META_SIZE], before_lookup_table[0..LOOKUP_TABLE_META_SIZE]);
+    const after_addresses = after_lookup_table[LOOKUP_TABLE_META_SIZE..];
+    @memcpy(after_addresses[0..@sizeOf(Pubkey)], &first_address.data);
+    @memcpy(after_addresses[@sizeOf(Pubkey)..], &second_address.data);
```
ajw221: same as previously mentioned
```suggestion
    @memcpy(after_addresses[0..Pubkey.SIZE], &first_address.data);
    @memcpy(after_addresses[Pubkey.SIZE..], &second_address.data);
```

shared/runtime/program/address_lookup_table/execute.zig
```diff
@@ -1221,3 +1221,149 @@ test "address-lookup-table extend" {
...
+        .{ .is_signer = false, .is_writable = true, .index_in_transaction = 0 },
+        .{ .is_signer = true, .is_writable = false, .index_in_transaction = 1 },
+        .{ .is_signer = true, .is_writable = true, .index_in_transaction = 2 },
+        .{ .is_signer = false, .is_writable = false, .index_in_transaction = 3 },
+        .{ .is_signer = false, .is_writable = false, .index_in_transaction = 4 },
+    };
+
+    try expectProgramExecuteResult(
+        allocator,
+        @This().ID,
```
ajw221: although this is being used in existing tests in this file, the `@This()` is redundant and can be removed
```suggestion
        ID,
```

shared/runtime/program/address_lookup_table/execute.zig
```diff
@@ -1221,3 +1221,149 @@ test "address-lookup-table extend" {
...
+    var b: u8 = 1;
+    for (0..255) |_| {
+        const pk: Pubkey = .{ .data = [_]u8{b} ** Pubkey.SIZE };
+        @memcpy(addr[0..@sizeOf(Pubkey)], &pk.data);
+        addr = addr[@sizeOf(Pubkey)..];
+        b +%= 1;
+    }
+
+    // after extend the table holds all 255 original addresses plus new_address appended
+    const after_lookup_table = try allocator.alloc(u8, LOOKUP_TABLE_META_SIZE + 256 * @sizeOf(Pubkey));
```
ajw221: `state.LOOKUP_TABLE_MAX_ADDRESSES` should be used where possible to avoid using magic numbers
```suggestion
    // fill the table to one below the cap so that extending by a single address
    // lands exactly on state.LOOKUP_TABLE_MAX_ADDRESSES
    const max_addresses = state.LOOKUP_TABLE_MAX_ADDRESSES;
    const before_lookup_table = try allocator.alloc(u8, LOOKUP_TABLE_META_SIZE + (max_addresses - 1) * @sizeOf(Pubkey));
    defer allocator.free(before_lookup_table);
    _ = try sig.bincode.writeToSlice(before_lookup_table[0..LOOKUP_TABLE_META_SIZE], meta_state, .{});
    var addr = before_lookup_table[LOOKUP_TABLE_META_SIZE..];
    var b: u8 = 1;
    for (0..max_addresses - 1) |_| {
        const pk: Pubkey = .{ .data = [_]u8{b} ** Pubkey.SIZE };
        @memcpy(addr[0..@sizeOf(Pubkey)], &pk.data);
        addr = addr[@sizeOf(Pubkey)..];
        b +%= 1;
    }

    // after extend the table holds all original addresses plus new_address appended
    const after_lookup_table = try allocator.alloc(u8, LOOKUP_TABLE_META_SIZE + max_addresses * @sizeOf(Pubkey));
```

shared/runtime/program/address_lookup_table/execute.zig
```diff
@@ -1221,3 +1221,149 @@ test "address-lookup-table extend" {
...
+        Instruction{ .ExtendLookupTable = .{ .new_addresses = &.{second_address} } },
+        meta,
+        .{
+            .accounts = accounts,
+            .compute_meter = 9_999_999,
+            .sysvar_cache = .{ .clock = runtime.sysvar.Clock.INIT, .rent = runtime.sysvar.Rent.INIT },
+        },
+        .{
+            .accounts = expected_accounts,
+            .accounts_resize_delta = @as(i64, @sizeOf(Pubkey)),
```
ajw221: same as previously mentioned
```suggestion
            .accounts_resize_delta = @as(i64, Pubkey.SIZE),
```

ajw221 (changes requested):

shared/runtime/program/address_lookup_table/execute.zig
insky7: [ab76644](https://github.com/Syndica/sig/commit/ab766445b9bf807570765ca48178982caaab7a47)

shared/runtime/program/address_lookup_table/execute.zig
insky7: [ab76644](https://github.com/Syndica/sig/commit/ab766445b9bf807570765ca48178982caaab7a47)

shared/runtime/program/address_lookup_table/execute.zig
insky7: [ab76644](https://github.com/Syndica/sig/commit/ab766445b9bf807570765ca48178982caaab7a47)

shared/runtime/program/address_lookup_table/execute.zig
insky7: [ab76644](https://github.com/Syndica/sig/commit/ab766445b9bf807570765ca48178982caaab7a47)

shared/runtime/program/address_lookup_table/execute.zig
insky7: [ab76644](https://github.com/Syndica/sig/commit/ab766445b9bf807570765ca48178982caaab7a47)

shared/runtime/program/address_lookup_table/execute.zig
insky7: [ab76644](https://github.com/Syndica/sig/commit/ab766445b9bf807570765ca48178982caaab7a47)

ajw221 (approved):

prestonsn (approved):
Nice contribution, and thanks for the fix! I'm approving, but FYI that this program has been migrated to an on-chain BPF program, so it's practically dead and being removed by this upcoming PR: https://github.com/Syndica/sig/pull/1641

PR #1665, opened 2026-07-02
feat(v2:conformance): implement shred parse harness

yewman:
## `conformance/v2`: `sol_compat_shred_parse_v1` differential harness

Lands the first v2 conformance harness: `sol_compat_shred_parse_v1`, covering shred parsing, FEC reassembly, per-slot chained-merkle-root validation, tick verification, and per-transaction structural checks. Includes the build-system plumbing (`sig_v2` module + shred protobuf bindings + `-Dversion=v2` backend selector) that lets the conformance library target v2.

Depends on receiver primitives (#1632) and parse/tick primitives (#1660).

**Scope**: passes the upstream `test-vectors/shred/fixtures` corpus. Local fuzz divergences (chain validation, RS-recovered revalidation, dead-slot semantics, …) are handled in a follow-up receiver-hardening PR — `--input ec6e612e7eb4` is a known crash until that lands.


conformance/src/v2/shred_parse.zig
```diff
@@ -0,0 +1,653 @@
...
+    @memcpy(out_slice[0..effects_bytes.len], effects_bytes);
+    out_size.* = effects_bytes.len;
+}
+
+// ---------------------------------------------------------------------------
+// Effects: collects callback output into ArrayLists.
+// ---------------------------------------------------------------------------
+
+/// Mutable scratch shared between the Receiver callbacks and the harness body.
+const Effects = struct {
```
dnut: I don't think an `Effects` interface is the right approach. The code would be simpler and less coupled without it. I also don't think we should merge #1615. The shred receiver already has a clear and testable boundary that should be used directly without piling on more abstractions and indirection that make the code harder to understand.

Here's an example of what the shred harness could look like without `Effects`. It could use some more refinement but it passes all the test vectors. https://github.com/Syndica/sig/commit/619dfc8188ceeae114f2573a1f6720feaa4149f6

conformance/build.zig
```diff
@@ -24,6 +24,16 @@ pub fn build(b: *Build) void {
...
 
+    // `-Dversion` picks one root per build because v1 and v2 share several
+    // sources (tracy, zstd, features.zon, ...) and Zig 0.15 rejects one
+    // source file rooting two modules in a single compilation.
+    const ConformanceVersion = enum { v1, v2 };
+    const version = b.option(
+        ConformanceVersion,
+        "version",
+        "Which conformance backend to compile into libsolfuzz_sig: v1 or v2.",
+    ) orelse .v1;
```
dnut: I don't see the need for this v1/v2 distinction. The runtime has already been extracted from v1. Let's just stop using v1 in conformance. We can test the same code through v2 or shared now.

dnut (changes requested):

conformance/build.zig
yewman: The harnesses aren't just missing shared re-exports, they use v1 specific types whose v2 replacements may differ (i.e. `sig.core.{Account, Ancestors, EpochStakes, EpochStakesMap, GenesisConfig, FeeRateGovernor, HardForks}` plus `sig.accounts_db.Db.initTest` / `AccountStore`). I don't like the idea of moving all that into `shared/` just to avoid temporary conformance versioning so I'd rather keep the `-Dversion` split in this PR and migrate harnesses to v2 one at a time in follow-ups. If you have another way to drop the version I am happy to give it a go, if the change set is large though I would still rather it land as part of a follow up

conformance/src/v2/shred_parse.zig
yewman: Removed via rebasing stack on main

conformance/build.zig
dnut: OK I see why we need to keep a v1 dependency. It's not as clean as I had hoped. But I still don't understand why we need to compile in different modes. Why not depend on both v1 and v2 and create a single binary?

yewman: I tried that initially, but adding sig_v2 as a second dep alongside sig and merging entrypoints into one dylib doesn't work: Zig 0.15 rejects it because sig and sig_v2 each build their own tracy/zstd/base58/features.zon from the same sources in ~/.cache/zig/p/..., so those files end up rooting two modules in a single compilation (file exists in modules 'tracy' and 'tracy0'). I couldn't come up with a clean way of getting around that so when with the build flag and two binaries

dnut: wow that sounds like a huge flaw in the build system. you can't have two dependencies that each have the same dependencies? am i understanding this correctly?

yewman: Here is a minimal repro if you are interested. Run `zig build` from the app dir. Looking into a solution at the moment...
[zig-shared-dep-collision.zip](https://github.com/user-attachments/files/29750184/zig-shared-dep-collision.zip)

dnut: OK I see the problem.

For tracy, it's a footgun in the build system. Normally it should be fine to have the same dependency appear multiple times in the graph. The build system can discover this and de-duplicate the modules. But the build system gets confused when the dependencies are the same exact version (same source code) but with different build options. It fails to dedupe them and you end up with multiple modules building the same source files. The build system ideally would have a nice way to deal with this, but it doesn't. So the solution is that we need to make sure that tracy is always built with the same exact options. That will eliminate the issue there.

For the feature set, that one is actually just the simple case of having multiple modules including on the same file. That's a mistake. We shouldn't be doing that. `shared` should create the only modules that include features.zon or gen_feature_set_id.zig, and v1 and v2 should import it from the `shared` module. v2 has a problem now because it creates another feature set module like the one in shared.

We need to fix all of this stuff asap to enable it all be to compiled together, otherwise the runtime will not be usable in v2 at all.

yewman: That is better, thanks for raising it. Lmk your thoughts on the new approach 
726396c3e2d130d14acbe36ca9c195fdd5f6addc
e88dfacb52f8cf9846b2dd9e50555dce4f37b002

dnut (dismissed):

Sobeston (dismissed):

prestonsn (dismissed):

dnut (approved):

Sobeston (approved):

PR #1695, opened 2026-07-08
fix(v2:replay): fecset 0 bootstrap hack adjustment 

prestonsn:
Fixes a replay panic caused by v2 replay's current bootstrap hack assuming the first observed FEC set is the slot-start FEC (`fec_set_idx == 0`). This PR adjusts (but does not fully replace) the hack so replay only botstraps once an `idx=0` FEC set is available. When that FEC arrives, replay assigns it the initial `BlockRef` and propagates that `BlockRef` to any already-attached child FEC sets that arrived earlier.


Sobeston (approved):
We 100% need a proper "boot" process for replay, bit this is an improvement for now

PR #1698, opened 2026-07-08
docs(topology): fixup README.md and minor comments

jbuckmccready:
- Update README.md with some tidying and current state of things for topology, updated walkthrough for adding new service and tested locally it works
- Add doc comments and TODO referencing issue number for per serivce seccomp filters

### DIFF
--- a/docs/docusaurus/docs/code/v2.md
+++ b/docs/docusaurus/docs/code/v2.md
@@ -1,28 +1,38 @@
 # Generated Docs
 
-1. Run `zig build docs`
+1. Run `zig build docs`.
 
-2. Serve it with a HTTP server, e.g. `python -m http.server -b 127.0.0.1 8000 -d zig-out/docs`
+2. Serve the generated files, for example:
+
+```sh
+python -m http.server -b 127.0.0.1 8000 -d zig-out/docs
+```
 
 # Running Shred Receive
 
-1. Use Agave to create a leader schedule text file, e.g.
+1. Use Agave to create a leader schedule text file:
 
-`solana leader-schedule > schedule.txt`
+```sh
+solana leader-schedule > schedule.txt
+```
 
-2. Start up v2
+2. Start v2 with the checked in example runtime config:
 
-`zig build run -- config/testnet.zig.zon`
+```sh
+zig build run -- config/example.zon
+```
 
-# Running with tracy
+`config/example.zon` shows the fields accepted by the runtime config and example values for running the current v2 topology. It is loaded as the `Config` type in `init/main.zig`. It controls values such as ports, cluster, telemetry settings, file paths, and memory sizes. It does not define service topology.
 
-1. Build with `-Denable-tracy`
+# Running with Tracy
 
-2. Use `.sandboxing_mode = .threaded` in your config
+1. Build with `-Denable-tracy`.
+
+2. Use `.sandboxing_mode = .threaded` in your runtime config.
 
 # Linting
 
-Run V2 lint checks from this directory:
+Run v2 lint checks from this directory:
 
 ```sh
 zig build lint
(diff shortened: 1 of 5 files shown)


docs/docusaurus/docs/code/v2.md
```diff
@@ -34,58 +44,29 @@ Pass explicit linter args after `--`:
...
 
-Run lint unit tests:
+# Topology
 
-```sh
-zig build lint-test
-```
+v2 runs each service as an isolated process, or as a thread when `.sandboxing_mode = .threaded` is used. The parent process creates shared memory regions before spawning services, then maps each region into the services that need it. Services communicate through those shared regions, but some services also perform runtime I/O, such as telemetry sockets, snapshot download and reads, and accounts DB file access.
+
+Topology is defined in Zig, not ZON:
```
InKryption: This is an unnecessary reference to a detail that's no longer relevant.

docs/docusaurus/docs/code/v2.md
```diff
@@ -34,58 +44,29 @@ Pass explicit linter args after `--`:
...
+
+1. `init/services.zig` declares each service's `ReadOnly` and `ReadWrite` region schema.
+2. `init/main.zig` declares the `Topology` type.
+3. `init/main.zig` creates and initializes shared memory regions.
+4. `init/main.zig` passes initialized regions to `children.spawn(...)`.
+
+Changing the service graph requires changing `main.zig`. The runtime config file cannot select a different topology. `config/example.zon` shows how `Config` fields are set and loaded for the topology defined in `init/main.zig`.
```
InKryption: > Changing the service graph requires changing `main.zig`. The runtime config file cannot select a different topology.

This is technically descriptive of current behavior, but seems to imply that it's architecturally impossible to influence the graph at runtime, which is not true.

> `config/example.zon` shows how `Config` fields are set and loaded for the topology defined in `init/main.zig`.

I would say the config is not so much defined for the "topology" as it is the configuration for the services themselves; if we added fields to influence what the service graph should look like, this would be more accurate.

In general, I think this paragraph is a bit misleading, and/or subject to being out of date very soon. I'd maybe replace it with something more neutral, like "The service graph is defined by the runner; the services are configured with the config file, for which `config/example.zon` serves as an example".

docs/docusaurus/docs/code/v2.md
```diff
@@ -34,58 +44,29 @@ Pass explicit linter args after `--`:
...
+
+Different "mains" (entry points) can be used to create integration tests on services in isolation with a different topology that allows for minimization of setup.
```
InKryption: I would maybe rephrase this as "Integration testing between subsets of services is done using alternative runners that construct the minimal graph needed to do so".

docs/docusaurus/docs/code/v2.md
```diff
@@ -34,58 +44,29 @@ Pass explicit linter args after `--`:
...
-    click svc "https://github.com/Syndica/sig/tree/main/v2/init/services.zig" _blank
-    click lib "https://github.com/Syndica/sig/tree/main/v2/lib" _blank
-    click s1 "https://github.com/Syndica/sig/tree/main/v2/services" _blank
-```
+This walkthrough adds a service called `foo` that reads a config region and logs through telemetry.
 
-Start by defining the data layout that will live in shared memory. The struct must be `extern` to guarantee a stable memory layout across process
-boundaries (Zig's default struct layout is not guaranteed to be consistent across compilations). Since regions are mapped at different virtual addresses
-in each process, pointer values from one process are invalid in another. Use fixed-size buffers with length fields instead of pointers or slices.
+Start by defining the data layout that will live in shared memory. The struct must be `extern` so its layout is stable across process boundaries. Since regions are mapped at different virtual addresses in each process, pointer values from one process are invalid in another. Use fixed-size buffers with length fields instead of pointers or slices.
```
InKryption: > Start by defining the data layout that will live in shared memory. The struct must be `extern` so its layout is stable across process boundaries.

It would be more accurate to say something along the lines of "Types stored in shared memory must have a defined layout, so that they're consistent across binary/process boundaries; `extern` structs are suitable for this".

> Since regions are mapped at different virtual addresses in each process, pointer values from one process are invalid in another. Use fixed-size buffers with length fields instead of pointers or slices.

This is an overly-restrictive suggestion, and not actually reflective of everything we do. The more general statement would be "[...]. For this reason, flat memory layouts should be used to avoid this issue.". If you want to be descriptive of the possible approaches, can also add "Flat layouts with a dynamic amount of data can be achieved with inline fixed-sized buffers, or using flexible array members".

docs/docusaurus/docs/code/v2.md
```diff
@@ -126,129 +116,69 @@ pub const name = .foo;
...
     rw.tel.signalReady();
 
     const bar = ro.config.bar_buf[0..ro.config.bar_len];
     logger.info().logf("foo service started with bar: {s}", .{bar});
 
-    while (true) {}
+    try runner.activity.signalIdleImmediate();
+    while (true) : (std.atomic.spinLoopHint()) {
+        try runner.activity.checkCanceled();
+    }
```
InKryption: Not sure the best way to express/illustrate this, but it would be good to demonstrate the more typical use of `try runner.activity.signalIdleSpinning()` in place of `std.atomic.spinLoopHint()`, which is what it's intended to replace.

(Which, on that note, using spinLoopHint here is a bit odd, since in theory this is the main loop, not a busy loop).

InKryption (changes requested):

docs/docusaurus/docs/code/v2.md
jbuckmccready: Agreed, https://github.com/Syndica/sig/pull/1698/changes/ce25c98db3e59956eb6e4c54ef1c7260af1e8b57

docs/docusaurus/docs/code/v2.md
jbuckmccready: Adjusted wording: https://github.com/Syndica/sig/pull/1698/changes/ce25c98db3e59956eb6e4c54ef1c7260af1e8b57

docs/docusaurus/docs/code/v2.md
jbuckmccready: Rewrote to clarify more specifically the what's and why's of using extern and avoiding sharing pointer values directly across shared memory.

https://github.com/Syndica/sig/pull/1698/changes/ce25c98db3e59956eb6e4c54ef1c7260af1e8b57

docs/docusaurus/docs/code/v2.md
jbuckmccready: Adjusted wording and kept it focus to just how it works, the intent is to clearly explain how things work as a "guide" for anyone jumping in, details likely to change as we make improvements.

https://github.com/Syndica/sig/pull/1698/changes/ce25c98db3e59956eb6e4c54ef1c7260af1e8b57

docs/docusaurus/docs/code/v2.md
jbuckmccready: Fixed to use runner API appropriately and added comment to clarify: https://github.com/Syndica/sig/pull/1698/changes/ce25c98db3e59956eb6e4c54ef1c7260af1e8b57

InKryption (approved):

ajw221 (approved):

PR #1699, opened 2026-07-08
docs(pool): clarify shared vs in-process pool semantics

jbuckmccready:
Added some doc comments clarifying use/purpose of pool collections with simple example use.

Reason for change: accelerate comprehension and code read speed for those unfamiliar with purpose and intent.


Sobeston (approved):
This reads fine 👍 

note: it might make sense to delete the regular `Pool` later, but there's no point blocking this PR over that

prestonsn (approved):

PR #1702, opened 2026-07-09
feat(v2): send VersionedEpochStakes to replay

kprotty:


v2/lib/solana/epoch_schedule.zig
```diff
@@ -1,3 +1,5 @@
+// TODO: move this into core
```
dnut, line 1: what's core?

also it would be nice if any TODOs we merge to main to have a github issue number as in `TODO(1234):`

v2/lib/snapshot.zig
```diff
@@ -76,3 +76,63 @@ pub const ReadySnapshot = extern struct {
...
+        while (true) {
+            const slot = self.slot.load(.acquire);
+            if (slot != invalid_slot) {
+                try runner.activity.signalActive();
+                return slot;
+            }
+            try runner.activity.signalIdleSpinning();
+        }
+    }
+};
```
dnut, line 138: Moving away from ring buffers has made this struct easier to use, but it's still a bit esoteric. You have to know that you're supposed to read getSlotBlocking before reading other fields. It's an unusual design that's not clear from looking at the code and it requires reading the type's documentation to understand how to use it correctly. Personally I prefer when the code is more self explanatory. What do you think about something like this:


```zig
pub const SnapshotMetadata = extern struct {
    slot: u64,

    manifest: lib.solana.snapshot.Manifest,
    status_cache: lib.solana.snapshot.StatusCache,

    memory_len: usize,
    memory: [0]u8 align(16), // VLA for [0..memory_len]

    pub fn init(self: *SnapshotMetadata, memory_len: usize) void {
        self.memory_len = memory_len;
    }

    /// Gets a slice of the serializable memory of the SnapshotMetadata
    pub fn getSerializable(self: *SnapshotMetadata) []u8 {
        // skip the slot & return all bytes from manifest onwards
        comptime std.debug.assert(@offsetOf(SnapshotMetadata, "slot") == 0);
        comptime std.debug.assert(@offsetOf(SnapshotMetadata, "manifest") == 8);
        const header_size = @sizeOf(SnapshotMetadata) - 8;
        return @as([*]u8, @ptrCast(&self.manifest))[0 .. header_size + self.memory_len];
    }

    /// Returns the base pointer used to resolve `RelativeSlice`/`RelativeOffset`
    /// values inside `manifest` / `status_cache`.
    pub fn getMemory(self: *SnapshotMetadata) []u8 {
        return self.memory[0..].ptr[0..self.memory_len];
    }
};

/// Stored in the region
pub const SharedSnapshotMetadata = extern struct {
    is_ready: std.atomic.Value(bool),
    /// not safe to read until is_ready is true
    private: SnapshotMetadata,

    /// Unblocks all getBlocking() callers.
    /// Can be called only once.
    /// Should also only call after all other SnapshotMetadata fields are populated.
    pub fn finalize(self: *SharedSnapshotMetadata) void {
        std.debug.assert(self.is_ready.swap(true, .release) == false);
    }

    pub fn getBlocking(
        self: *SharedSnapshotMetadata,
        runner: lib.runner.Connection,
    ) !*SnapshotMetadata {
        while (true) {
            if (self.is_ready.load(.acquire)) {
                try runner.activity.signalActive();
                return &self.private;
            }
            try runner.activity.signalIdleSpinning();
        }
    }
};
```

v2/lib/accounts_db/rooted.zig
```diff
@@ -391,16 +373,42 @@ pub const Rooted = struct {
...
+            try self.queueWrite(.from(logger), metadata.len, &r);
+
+            // write padding if any
+            const pad_len = padded - metadata.len;
+            if (pad_len > 0) try self.queueWrite(.from(logger), pad_len, struct {
+                pub fn readSliceAll(_: @This(), b: []u8) !void {
+                    @memset(b, 0);
+                }
+            }{});
+        }
```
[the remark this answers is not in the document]
kprotty: Intentional to (naively) rewrite it on root without shifting over sectors on size change

v2/lib/solana/snapshot.zig
```diff
@@ -106,23 +182,37 @@ pub const Manifest = struct {
...
+        hashes_count: u32,
 
         pub const Entry = extern struct {
             hash: Hash,
             hash_index: u64,
         };
 
+        pub fn getHashes(self: *const BlockHashQueue) []const Hash {
+            return self.hashes[0..self.hashes_count];
+        }
```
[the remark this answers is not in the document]
kprotty: Zig already does this

v2/services/replay.zig
```diff
@@ -292,52 +292,50 @@ pub fn serviceMain(runner: lib.runner.Connection, _: ReadOnly, rw: ReadWrite) !n
...
+    // Acquire barrier — pairs with accounts_db's `populateSlot` (Release). Once
+    // this returns, `snapshot_metadata.manifest` / `.status_cache` / all their
+    // trailing FBA data are fully published and safe to read from this process.
+    const root_slot = try snapshot_metadata.getSlotBlocking(runner);
+    logger.info().logf("got the root slot from the snapshot: {}", .{root_slot});
 
-        const root_block = last_block orelse return error.NoBlockhashesInSnapshot;
+    const manifest = &snapshot_metadata.manifest;
+    const bhq = &manifest.bank_fields.blockhash_queue;
+    const hashes = bhq.hashes[0..bhq.hashes_count];
```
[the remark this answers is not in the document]
dnut: the u32 doesn't matter, but using `getHashes` makes sense

v2/lib/solana/snapshot.zig
```diff
@@ -8,6 +8,77 @@ const Pubkey = lib.solana.Pubkey;
...
+
+        pub fn from(base: [*]const u8, ptr: *const T) Self {
+            const o = @intFromPtr(ptr) - @intFromPtr(base);
+            std.debug.assert(o != 0);
+            std.debug.assert(o <= std.math.maxInt(u32));
+            return .{ .offset = @intCast(o) };
+        }
+
+        pub fn isNull(self: Self) bool {
+            return self.offset == 0;
```
dnut, line 64: Why does every type need to be nullable? I'd rather express this in the concrete type by using an optional type as T if necessary.

v2/lib/solana/snapshot.zig
```diff
@@ -106,23 +182,37 @@ pub const Manifest = struct {
...
     blockhash_queue: BlockHashQueue,
-
-    pub const BlockHashQueue = struct {
-        last_hash: ?Hash,
+    epoch_schedule: EpochSchedule,
+    inflation: Inflation,
+    stakes_cache: extern struct {
+        epoch: Epoch,
+        vote_accounts: RelativeSlice(Pubkey), // acc.data contains stake for the voter
+        stake_accounts: RelativeSlice(Pubkey), // acc.data contains voter pubkey + Delegation
```
dnut, line 193: Your intent is for replay's bootstrap process to fetch all these accounts from accountsdb before starting replay? Avoiding that would require sending 8 extra bytes per vote account, and 64 extra bytes per stake account. Is that a problem?

```
StakeDelegation {
    voter_pubkey:       Pubkey,   // 32   which vote account this stake delegates to
    stake:              u64,      //  8   delegated lamports
    activation_epoch:   Epoch,    //  8   warmup/cooldown math needs both epochs
    deactivation_epoch: Epoch,    //  8
    credits_observed:   u64,      //  8   reward = (voter_credits_now − credits_observed) * pts
}
```

v2/lib/solana/snapshot.zig
```diff
@@ -367,25 +500,152 @@ pub const AccountsDbFields = struct {
...
     /// - Alpenglow: the "double Merkle root": a Merkle root computed over the
     ///   sequence of per-FEC-set Merkle roots of the block's shreds.
     block_id: Hash,
 
+    pub const VersionedEpochStakes = extern struct {
+        epoch: Epoch,
+        total_stake: u64,
+        vote_accounts: RelativeSlice(VoteAccountEntry),
+        node_to_vote_accounts: RelativeSlice(NodeToVoterEntry),
+        epoch_authorized_voters: RelativeSlice(AuthToVoterEntry),
```
dnut, line 515: Do we actually need these two fields for anything?

v2/lib/solana/snapshot.zig
```diff
@@ -8,6 +8,77 @@ const Pubkey = lib.solana.Pubkey;
...
+        len: u32 = 0,
+
+        const Self = @This();
+
+        pub fn fromSlice(base: [*]const u8, s: []const T) Self {
+            const start = @intFromPtr(s.ptr) - @intFromPtr(base);
+            std.debug.assert(start <= std.math.maxInt(u32));
+            std.debug.assert(s.len <= std.math.maxInt(u32));
+            return .{ .offset = @intCast(start), .len = @intCast(s.len) };
+        }
```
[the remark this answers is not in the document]
kprotty: zig already has these asserts done by the `-` of both `@intFromPtr`

v2/lib/solana/snapshot.zig
```diff
@@ -22,7 +93,12 @@ fn readBool(r: anytype) !bool {
...
 }
 
-pub const StatusCache = struct {
+pub const StatusCache = extern struct {
+    /// Placeholder — the deserialized contents are currently discarded. Kept as an
+    /// extern struct so it can live inline in `SnapshotMetadata` and be persisted
+    /// alongside the `Manifest` in the rooted DB blob.
+    _reserved: u8 = 0,
+
     pub fn read(fba: *std.heap.FixedBufferAllocator, r: anytype) !StatusCache {
```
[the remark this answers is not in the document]
kprotty: It would be better to make `memory_len` align(1) instead of tuning StatusCache to how its stored in its parent.

But in this case, it's fine to persist padding bytes. As long as those padding byte sizes dont change.

v2/lib/solana/inflation.zig
```diff
@@ -0,0 +1,108 @@
+// TODO: move this into core
+
+const std = @import("std");
+
+/// Zig's `std.math.pow` may return a result that is off by up to one ULP, when comparing to glibc or musl's `pow()`.
+/// As these calculations affect consensus, that is an unacceptable difference for us, so we import libc's pow and
+/// use that. For reference:
+/// - `std.math.pow`: pow(0.85, 4.019250798563942) -> 7.805634650110366e-2
+/// - glibc/musl: pow(0.85, 4.019250798563942) -> 7.805634650110367e-2
+extern fn pow(f64, f64) f64;
```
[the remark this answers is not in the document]
kprotty: If it fails to link, we'll get a compile error so that is fine. It currently works since we have dependencies like zstd that link to libc anyways.

v2/lib/solana/epoch_schedule.zig
kprotty: `shared/core` folder root. `v2/lib/solana` still has its own duplicates of types from there

v2/lib/solana/snapshot.zig
kprotty: `RelativeOffset` is stored in Manifest/StatusCache and must be `extern`/region compatible. Zig's optionals (like slices) dont have a stable ABI

v2/lib/solana/snapshot.zig
kprotty: Yes. It'll have to fetch them from AccountsDB to verify that the data in the snapshot manifest matches the accounts on disk: https://github.com/anza-xyz/agave/blob/master/runtime/src/stakes.rs#L344

This struct just then contains what to fetch, where we'll use the account data directly instead of verifying fields that the snapshot manifest presented. Should we still verify those fields?

v2/lib/solana/snapshot.zig
kprotty: Yes, IIRC: https://github.com/Syndica/sig/blob/main/v1/src/replay/epoch_transitions.zig#L212

PR #1703, opened 2026-07-09
feat(replay): bootstrap block tree and merkle forest from the snapshot

dnut:
I requested 4 reviewers because a piece of this touches something each of you are already working very closely on. You don't need to review the entire thing, only the relevant piece. I've messaged each of you with the context.

---

Currently replay uses a hack to bootstrap the starting root of the block tree. It trusts the first fec set received, and it treats its slot-1 as the root.

This change uses the snapshot to properly bootstrap replay's block tree and merkle forest.

There is one caveat: some of the fields in replay's data structures are simply not possible to derive from a snapshot. We will never be able to populate this data, unless the snapshot format is changed to include more information. So I implemented this to use placeholder/null values for the following:
- slot number in the block tree for slots older than the starting root
- fields in the final fec set of the starting root:
    - chained_merkle_root
    - fec_set_idx
    - payload_len

I made `slot` optional, and left the others with placeholder values.

# Testing

I ran it with a testnet snapshot and ledger fed in with the shred-stream tool. Replay bootstrapped fine and started scheduling slots for execution, once I fixed a bug that got the correct block_id.

image

### DIFF
--- a/v2/lib/accounts_db.zig
+++ b/v2/lib/accounts_db.zig
@@ -41,11 +41,20 @@ pub const AccountLookups = extern struct {
     }
 };
 
+/// How to consume this struct:
+/// 1. read all ring buffers in the correct order (specify the correct order
+///    here if more are added, currently there is only one: the blockhash queue)
+/// 2. call getSlotBlocking
+/// 3. read other fields
 pub const RuntimeMetadata = extern struct {
     slot: std.atomic.Value(u64),
+    /// The merkle root of the last fec set (tower) or the root of roots (alpenglow)
+    block_id: Hash,
     blockhash_queue: extern struct {
         /// read after consuming all of hashes
         max_age: u64,
+        /// Accountsdb blocks until enough hashes are read from here to make
+        /// room for accountsdb to write all of its block hashes here.
         hashes: lib.ipc.Ring(256, Hash),
     },
 
@@ -67,6 +76,8 @@ pub const RuntimeMetadata = extern struct {
         std.debug.assert(self.slot.swap(slot, .release) == invalid_slot);
     }
 
+    /// Accountsdb writes the slot last, so you need to empty all the ring
+    /// buffers before calling this.
     pub fn getSlotBlocking(self: *RuntimeMetadata, runner: lib.runner.Connection) !Slot {
         while (true) {
             const slot = self.slot.load(.acquire);
--- a/v2/lib/accounts_db/rooted.zig
+++ b/v2/lib/accounts_db/rooted.zig
@@ -168,10 +168,9 @@ pub const Rooted = struct {
             padding,
             /// holds an actual account
             account,
-            /// holds a serialized addition to the blockhash queue
-            blockhashes,
-            /// holds a serialized addition to the block_id queue
-            block_ids,
+            /// holds the block_id followed by a serialized addition to the blockhash queue.
+            /// body layout: `{ block_id: Hash, [info.count]Hash }`
+            block_metadata,
             _, // TODO: add other types of sections
         },
         info: packed union {
@@ -306,7 +305,13 @@ pub const Rooted = struct {
                     });
                     n_puts += 1;
                 },
-                .blockhashes => {
+                .block_metadata => {
+                    // read block_id
+                    var block_id: Hash = undefined;
+                    try self.readExisting(.from(logger), (&block_id.data).ptr, @sizeOf(Hash));
+                    last_block_id = block_id;
+
+                    // read blockhashes
                     var num_hashes = header.info.count;
                     while (num_hashes > 0) {
                         // get a buffer to write hashes into
@@ -331,11 +336,6 @@ pub const Rooted = struct {
                         blockhash_writer.advance(take);
                         num_hashes -= take;
                     }
-                },
-                .block_ids => for (0..header.info.count) |_| {
-                    var block_id: Hash = undefined;
-                    try self.readExisting(.from(logger), (&block_id.data).ptr, @sizeOf(Hash));
-                    last_block_id = block_id;
                 },
                 _ => return error.InvalidSector,
             }
@@ -438,26 +438,34 @@ pub const Rooted = struct {
             }
         }
 
-        { // write the current blockhash queue
+        { // write the block_id + current blockhash queue
             const blockhash_queue = &snapshot_iter.manifest.bank_fields.blockhash_queue;
             self.journal.blockhash_max_age = std.math.lossyCast(u32, blockhash_queue.max_age);
 
+            const block_id = snapshot_iter.manifest.extra_fields.block_id;
             const num_hashes = blockhash_queue.hashes.count;
             const hashes = blockhash_queue.hashes.array[0..num_hashes];
 
-            // write blockhashes header
+            // write block_metadata header
             const header: SectorHeader = .{
-                .type = .blockhashes,
+                .type = .block_metadata,
                 .info = .{ .count = @intCast(num_hashes) },
             };
 
             var r = std.Io.Reader.fixed(std.mem.asBytes(&header));
             try self.queueWrite(.from(logger), @sizeOf(SectorHeader), &r);
 
+            // write block_id
+            r = std.Io.Reader.fixed(std.mem.asBytes(&block_id));
+            try self.queueWrite(.from(logger), @sizeOf(Hash), &r);
+
+            // write blockhashes
             r = std.Io.Reader.fixed(std.mem.sliceAsBytes(hashes));
             try self.queueWrite(.from(logger), num_hashes * @sizeOf(Hash), &r);
 
-            // send over as metadata
+            runtime_metadata.block_id = block_id;
+
+            // send blockhashes over as metadata
             var blockhash_writer = runtime_metadata.blockhash_queue.hashes.getView(.writer);
             defer {
                 runtime_metadata.blockhash_queue.max_age = self.journal.blockhash_max_age;
@@ -474,23 +482,6 @@ pub const Rooted = struct {
                 blockhash_writer.advance(take);
                 i += take;
             }
-        }
-
-        { // write the block_id sector for this block
-            const block_id = snapshot_iter.manifest.extra_fields.block_id;
-
-            const header: SectorHeader = .{
-                .type = .block_ids,
-                .info = .{ .count = 1 },
-            };
-
-            var r = std.Io.Reader.fixed(std.mem.asBytes(&header));
-            try self.queueWrite(.from(logger), @sizeOf(SectorHeader), &r);
-
-            r = std.Io.Reader.fixed(std.mem.asBytes(&block_id));
-            try self.queueWrite(.from(logger), @sizeOf(Hash), &r);
-
-            runtime_metadata.block_id = block_id;
         }
 
         // write the slot to commit the RuntimeMetadata stuff
--- a/v2/lib/accounts_db/rooted.zig
+++ b/v2/lib/accounts_db/rooted.zig
@@ -164,9 +164,14 @@ pub const Rooted = struct {
 
     const SectorHeader = packed struct(u64) {
         type: enum(u3) {
-            padding, // padding data to get file aligned to block_size for writing
-            account, // holds an actual account
-            blockhashes, // holds a serialized addition to the blockhash queue
+            /// padding data to get file aligned to block_size for writing
+            padding,
+            /// holds an actual account
+            account,
+            /// holds a serialized addition to the blockhash queue
+            blockhashes,
+            /// holds a serialized addition to the block_id queue
+            block_ids,
             _, // TODO: add other types of sections
         },
         info: packed union {
@@ -258,6 +263,7 @@ pub const Rooted = struct {
         var timer = try std.time.Timer.start();
         var n_puts: usize = 0;
         var n_bytes_read: usize = 0;
+        var last_block_id: ?Hash = null;
 
         // read sectors until EOF
         while ((try self.io.reader.getBuffer(.from(logger))).len > 0) {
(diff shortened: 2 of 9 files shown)


v2/services/exec.zig
```diff
@@ -48,7 +48,7 @@ pub fn serviceMain(runner: lib.runner.Connection, ro: ReadOnly, rw: ReadWrite) !
             .txn_exec => {
                 const data = &request.data.txn_exec;
 
-                const slot = data.block_idx.constPtr(ro.block_pool).slot;
+                const slot = data.block_idx.constPtr(ro.block_pool).slot.opt().?;
                 zone.value(slot);
                 tracy.plot(u48, "exec slot", @intCast(slot));
```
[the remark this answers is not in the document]
dnut: I don't see why we would ever create an execution request for a block that's older than the snapshot root. seems pointless to handle this

v2/lib/accounts_db.zig
```diff
@@ -41,11 +41,20 @@ pub const AccountLookups = extern struct {
...
+/// How to consume this struct:
+/// 1. read all ring buffers in the correct order (specify the correct order
+///    here if more are added, currently there is only one: the blockhash queue)
+/// 2. call getSlotBlocking
+/// 3. read other fields
 pub const RuntimeMetadata = extern struct {
     slot: std.atomic.Value(u64),
+    /// The merkle root of the last fec set (tower) or the root of roots (alpenglow)
+    block_id: Hash,
     blockhash_queue: extern struct {
```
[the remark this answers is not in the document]
dnut: @kprotty does this actually matter? it's garbage data regardless, until externally initialized with the correct hash

v2/lib/solana/snapshot.zig
```diff
@@ -533,12 +538,17 @@ pub const ExtraFields = struct {
...
-            error.EndOfStream => {},
-            else => |e| return e,
-        };
+        //
+        // in agave, this field is optional on the deserialize side, but it's
+        // actually always populated. we do not need to handle the null case,
+        // and it's actually not even possible for replay to work if this is
+        // null. so we treat this as a required field.
+        var block_id: [32]u8 = undefined;
+        try r.readSliceAll(&block_id);
```
Sobeston: Would it be easy to check, and panic if it's null?

v2/services/exec.zig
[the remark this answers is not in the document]
dnut: I added a comment explaining this

v2/lib/accounts_db/rooted.zig
```diff
@@ -348,6 +359,7 @@ pub const Rooted = struct {
                 self.io.reader.io_stalled = 0;
             }
         }
+        runtime_metadata.block_id = last_block_id orelse return error.NoBlockIDs;
```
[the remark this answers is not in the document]
dnut: i think the error is sufficiently informative. if we're going to provide operator advice for this kind of error, it should probably be expressed at a higher level of abstraction. for example, more broadly, if there's any error resuming from an old accountsdb, then delete your db and start again from a snapshot

v2/lib/accounts_db.zig
```diff
@@ -41,11 +41,20 @@ pub const AccountLookups = extern struct {
...
 
+/// How to consume this struct:
+/// 1. read all ring buffers in the correct order (specify the correct order
+///    here if more are added, currently there is only one: the blockhash queue)
+/// 2. call getSlotBlocking
+/// 3. read other fields
 pub const RuntimeMetadata = extern struct {
     slot: std.atomic.Value(u64),
+    /// The merkle root of the last fec set (tower) or the root of roots (alpenglow)
+    block_id: Hash,
```
Sobeston, line 52: Nice that agave is producing this, I thought we'd have to wait for SIMD-0333

v2/lib/util.zig
```diff
@@ -82,6 +82,56 @@ pub fn assertInterface(comptime InterfaceType: type, comptime ContractStruct: ty
...
+    const Int = @as(?type, switch (@typeInfo(T)) {
+        .int => T,
+        .@"enum" => |info| info.tag_type,
+        .@"struct" => |s| s.backing_integer,
+        else => null,
+    }) orelse @compileError("Unsupported type for Optional(_): " ++ @typeName(T));
+
+    return enum(Int) {
+        null = std.math.maxInt(Int),
+        _,
```
Sobeston, line 95: Neat, can we give it a different name maybe? PackedOptional or something?

v2/lib/solana/snapshot.zig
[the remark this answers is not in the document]
dnut: [3b3ed04](https://github.com/Syndica/sig/pull/1703/commits/3b3ed045b103ef5521848c3a74328c72729f04ef)

Sobeston (dismissed):
Seems about as correct as we can do it right now, happy to merge assuming you've tested it

just a nit, a question, and a comment

v2/lib/solana/snapshot.zig
dnut: Actually right before you posted this I changed this to return a specific error when this is null, instead of EndOfStream which is what you'd see otherwise: [3b3ed04](https://github.com/Syndica/sig/pull/1703/commits/3b3ed045b103ef5521848c3a74328c72729f04ef)

v2/lib/accounts_db.zig
dnut: Yep, they're hardcoding it to `Some()`

```rust
        let block_id = Some(bank_fields.block_id);
        let bank_fields_to_serialize = (
            SerializableVersionedBank::from(bank_fields),
            SerializableAccountsDb::<'_> {
                slot,
                account_storage_entries: self.snapshot_storages,
                bank_hash_stats,
            },
            ExtraFieldsToSerialize {
                lamports_per_signature,
                unused_incremental_snapshot_persistence: None,
                unused_epoch_accounts_hash: None,
                versioned_epoch_stakes,
                accounts_lt_hash,
                block_id,
            },
```
https://github.com/anza-xyz/agave/blob/4fc49623410a3af05b37116645d59f18107f6161/runtime/src/serde_snapshot.rs#L674-L689

v2/lib/util.zig
dnut: [480bb93](https://github.com/Syndica/sig/pull/1703/commits/480bb934308846c483299be81217864e61a2a2c3)

prestonsn (changes requested):
Just one question about bootstrap's parent block setting behavior, looks good otherwise!

v2/lib/util.zig
```diff
@@ -82,6 +82,56 @@ pub fn assertInterface(comptime InterfaceType: type, comptime ContractStruct: ty
     }
 }
 
+/// A tightly packed optional value for data that can be represented in memory
+/// as an integer.
+///
+/// This is only safe to use if you can be certain that it will never need to
+/// represent the maxInt for the backing integer.
+pub fn PackedOptional(T: type) type {
```
prestonsn: This is pretty cool. More of a nit: Are there places where we log this type? in which case do we want a little `format(..)` impl for this?

v2/services/replay.zig
```diff
@@ -305,6 +257,106 @@ pub fn serviceMain(runner: lib.runner.Connection, _: ReadOnly, rw: ReadWrite) !n
...
+                    .parent = .init(parent_block),
+                };
+                if (parent_block) |p| p.ptr(block_pool).child = .init(latest_block);
+                blockhash_states[latest_block.?.index()] = hash.*;
+                num_hashes += 1;
+            }
+            blockhashes_in.advance(hashes.len);
+            parent_block = latest_block;
+        }
+
```
prestonsn, line 325: I could be wrong, but if the `blockhashes_in` buffer had block hashes A, B, C, and D. Then due to `var parent_block: ?lib.replay.BlockRef = null;`, the subsequent `.parent = .init(parent_block),` in inside the loop would then set: 

```zig
A.parent = null
B.parent = null
C.parent = null
D.parent = null 
// Then after the loop:
parent_block = latest_block; // D
```
But `C.parent` would still be null (and `D.parent` would not be C?)

dnut: good catch, i put it in the wrong loop

dnut: [48cfdf9](https://github.com/Syndica/sig/pull/1703/commits/48cfdf90ed8046dbd1e99d12fc200c5ac100a39d)

v2/lib/util.zig
dnut: good point, this is actually being printed already

[5918df1](https://github.com/Syndica/sig/pull/1703/commits/5918df1e87f4474f7aac0e4e6bad52c6004b7b22)

v2/lib/accounts_db/rooted.zig
```diff
@@ -164,9 +164,14 @@ pub const Rooted = struct {
...
-            account, // holds an actual account
-            blockhashes, // holds a serialized addition to the blockhash queue
+            /// padding data to get file aligned to block_size for writing
+            padding,
+            /// holds an actual account
+            account,
+            /// holds a serialized addition to the blockhash queue
+            blockhashes,
+            /// holds a serialized addition to the block_id queue
+            block_ids,
```
kprotty: could these be merged into the same section? where its body is `{ block_id, [info.count]BlockHashes }`

v2/lib/solana/snapshot.zig
```diff
@@ -533,12 +538,20 @@ pub const ExtraFields = struct {
...
 
         // block_id: NullOnEof(Hash)
-        r.discardAll(32) catch |err| switch (err) {
-            error.EndOfStream => {},
+        //
+        // in agave, this field is optional on the deserialize side, but it's
+        // actually always populated. we do not need to handle the null case,
+        // and it's actually not even possible for replay to work if this is
+        // null. so we treat this as a required field.
+        var block_id: [32]u8 = undefined;
```
kprotty:
```suggestion
        var block_id: Hash = undefined;
```

v2/lib/accounts_db.zig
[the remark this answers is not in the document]
kprotty: doesnt matter. can zero it out if you'd like

v2/services/replay.zig
prestonsn: Nice! I like the new test

prestonsn (dismissed):

v2/lib/util.zig
```diff
@@ -82,6 +82,96 @@ pub fn assertInterface(comptime InterfaceType: type, comptime ContractStruct: ty
...
```
InKryption: I think it would be nice to encode the sentinel value as an explicit part of the type, ie `PackedOptional(u32, std.math.maxInt(u32))`, but I'm not going to make a hard stand on this given we currently don't really actually need it.

v2/lib/util.zig
```diff
@@ -82,6 +82,96 @@ pub fn assertInterface(comptime InterfaceType: type, comptime ContractStruct: ty
...
+    const Int = @as(?type, switch (@typeInfo(T)) {
+        .int => T,
+        .@"enum" => |info| info.tag_type,
+        .@"struct" => |s| s.backing_integer,
+        else => null,
+    }) orelse @compileError("Unsupported type for PackedOptional(_): " ++ @typeName(T));
```
InKryption: It seems a bit superfluous to support packed structs for this, given an all-ones packed struct is a lot more common than a maxInt(Int).

v2/services/replay.zig
```diff
@@ -305,6 +257,105 @@ pub fn serviceMain(runner: lib.runner.Connection, _: ReadOnly, rw: ReadWrite) !n
...
+        .chained_merkle_root = .ZEROES, // cannot be determined from the snapshot
+        .id = .{
+            .slot = root_slot,
+            .fec_set_idx = 0, // cannot be determined from the snapshot
+        },
+        .data_complete = true,
+        .slot_complete = true,
+        .payload_len = 0, // cannot be determined from the snapshot
+        .payload_buf = undefined,
+    }, forest, block_pool)).?;
```
InKryption: Small style nit:
```suggestion
    const root_node = try insertFecSet(logger, &.{
        .merkle_root = snapshot_metadata.block_id,
        .chained_merkle_root = .ZEROES, // cannot be determined from the snapshot
        .id = .{
            .slot = root_slot,
            .fec_set_idx = 0, // cannot be determined from the snapshot
        },
        .data_complete = true,
        .slot_complete = true,
        .payload_len = 0, // cannot be determined from the snapshot
        .payload_buf = undefined,
    }, forest, block_pool) orelse unreachable;
```
`orelse unreachable` stands out a bit more here, and it lets you get rid of the parentheses.

InKryption (dismissed):
I have some very minor nits, but nothing I want to block this on.

v2/lib/accounts_db/rooted.zig
dnut: [10298d5](https://github.com/Syndica/sig/pull/1703/commits/10298d5d8f8c0a14366a5be4ee0dfaabffce93b6)

v2/lib/solana/snapshot.zig
dnut: [b213444](https://github.com/Syndica/sig/pull/1703/commits/b21344400e6359e4fe04bb6e05c4f5ff56b6cdc3)

v2/lib/util.zig
dnut: I see what you mean, but I still see potential value. I figured why not make this type comprehensive now when I already have the context, it's easy enough. But I won't make a fuss over it. We can add it back if needed.

[6ebf113](https://github.com/Syndica/sig/pull/1703/commits/6ebf1134f56f6afd9cb5cea9785965813be86d91)

v2/services/replay.zig
dnut: [6ebf113](https://github.com/Syndica/sig/pull/1703/commits/6ebf1134f56f6afd9cb5cea9785965813be86d91)

v2/lib/util.zig
dnut: [6ebf113](https://github.com/Syndica/sig/pull/1703/commits/6ebf1134f56f6afd9cb5cea9785965813be86d91)

InKryption (dismissed):

Sobeston (dismissed):

kprotty (approved):

Sobeston (approved):

PR #1707, opened 2026-07-10
refactor(runtime): hardcode raise_account_cu_limit

hamza-syndica:
Closes #1653

### DIFF
--- a/shared/core/features.zon
+++ b/shared/core/features.zon
@@ -1530,6 +1530,11 @@
         .name = "raise_account_cu_limit",
         .pubkey = "htsptAwi2yRoZH83SKaUXykeZGtZHgxkS2QwW1pssR8",
         .description = "SIMD-0306: Raise account CU limit to 40% max",
+        // NOTE: The active path is not yet implemented — the feature is hardcoded
+        // because the gate is no longer needed, but the raised CU limit behavior
+        // still needs to be implemented once the cost tracker is added. We use
+        // .hardcoded (rather than .supported) to avoid a feature compatibility
+        // error with solfuzz-agave, which has already hardcoded this feature.
         .status = .hardcoded,
     },
     .{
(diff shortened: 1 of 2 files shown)


shared/core/features.zon
```diff
@@ -1530,7 +1530,7 @@
         .name = "raise_account_cu_limit",
         .pubkey = "htsptAwi2yRoZH83SKaUXykeZGtZHgxkS2QwW1pssR8",
         .description = "SIMD-0306: Raise account CU limit to 40% max",
-        .status = .supported,
+        .status = .hardcoded,
```
yewman, line 1533: Add a note here that this is not actually implemented, it is hardcoded because it no longer needs to be switched on but we never actually implemented the active path. It needs to be either 'supported' or 'hardcoded' since it is hardcoded in solfuzz-agave and would otherwise cause a feature compatibility error. I think 'hardcoded' is the better of the two because we do not need the gate, but still need to implement the active path once cost tracker is added.

yewman (changes requested):

shared/core/features.zon
hamza-syndica: https://github.com/Syndica/sig/pull/1707/changes/58e642696c71622c97a80de5fd0d5dbcd3fc93e5

yewman: `.supported` would be compatible, we can't use `.unsupported` or we would have incompatible feature sets. `hardcoded` is better than supported imo because when we do implement the logic, it will not be conditional on a gate.

hamza-syndica: https://github.com/Syndica/sig/pull/1707/changes/4771e64d72f589c8e3ddb7dd1e27ff9f14a1ce23

yewman (approved):

dnut (approved):

PR #1711, opened 2026-07-10
feat(v2:replay): add zero-copy VersionedTransaction parsing 

prestonsn:
Adds a new `Layout` struct that contains offsets into the transaction wire byte payload (bincoded bytes of `VersionedTransaction`). This allows replay and related components to re-use the calculated offsets for jumping directly to fields of interest instead of repeated bincode de-serialization. This also avoids requiring a stack allocated buffer/FBA setup. 

The core idea here is that we're leveraging the existing parse code in replay to also note down field offsets during block/transaction decoding from shreds. This lets downstream components (account loading, exec, etc.) to use the offsets into the payload rather than perform bincode de-serialization again. 

This lets us remove these bits of code from replay and exec: 

```zig
var deserialised_buf: [16 * 1024]u8 = undefined;
var deserial_fba: std.heap.FixedBufferAllocator = .init(&deserialised_buf);
var reader = std.io.Reader.fixed(ro.replay_transaction_pool.indexToConstPtr(data.tx_idx));
const transaction: lib.solana.transaction.VersionedTransaction =
    try lib.solana.bincode.read(
        &deserial_fba,
        &reader,
        lib.solana.transaction.VersionedTransaction,
);
defer deserial_fba.reset();
```

Note: this does not eliminate the copy out of frag shred storage, just allows for zero-copy access to parsed `VersionedTransaction`s in downstream services/components with access to `TransactionPool`

This is part of the effort in implementing improved account reads for replay (https://github.com/Syndica/sig/issues/1681)


v2/lib/replay.zig
```diff
@@ -1,15 +1,45 @@
...
-pub const TransactionPool = collections.SharedPool([1232]u8, 10_000);
+pub const TransactionPool = collections.SharedPool(TransactionRecord, 10_000);
 
 pub const BlockPool = collections.SharedPool(Node, 1024);
 
+/// Transction bytes plus their validatoes wire layout.
+///
+/// This struct itself is safe to share between processe. Consumers can construct transient
+/// `TransactionView`s locally, avoidng needing to re-parse the transaction bytes.
+///
```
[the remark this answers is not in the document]
prestonsn: https://github.com/Syndica/sig/pull/1711/commits/73b5fe28206025f62f7890c2f181b47e99200862

v2/lib/solana/transaction.zig
```diff
@@ -282,6 +563,18 @@ pub const SliceReader = struct {
...
+        if (len > self.bytes.len - self.pos)
+            return error.EndOfStream;
+
+        const result = self.bytes[self.pos..][0..len];
+        self.pos += len;
+        return result;
+    }
 };
 
 pub const VersionedMessage = union(enum) {
```
kprotty:
```suggestion
pub const VersionedMessage = union(VersionByte) {
```

v2/lib/solana/transaction.zig
```diff
@@ -290,6 +583,22 @@ pub const VersionedMessage = union(enum) {
...
+        legacy = 0xff,
+        v0 = 0,
+    };
+
+    pub fn versionByte(self: *const VersionedMessage) VersionByte {
+        return switch (self.*) {
+            .legacy => .legacy,
+            .v0 => .v0,
+        };
+    }
```
kprotty, line 597: by using VersionByte for the union, this should allow for:
```suggestion
pub fn versionByte(self: *const VersionedMessage) VersionByte {
    return @as(VersionByte, self.*);
}
```

v2/lib/replay.zig
```diff
@@ -5,13 +5,40 @@ const ipc = @import("ipc.zig");
...
+        const payload_len: usize = self.layout.payload_len;
+        std.debug.assert(payload_len <= self.payload.len);
+        return self.payload[0..payload_len];
+    }
+
+    pub fn view(self: *const TransactionRecord) VersionedTransaction.View {
+        return .{
+            .layout = &self.layout,
+            .payload = self.bytes(),
+        };
```
dnut, line 38: Assuming we actually do want to read into the packet layout directly, I think the approach can be simplified. Why do we need all these types: VersionedTransaction, Layout, View, TransactionRecord? It's complicated.

In principle there could be only a single struct, for example:

```zig
pub const Transaction = struct {
    payload: [VersionedTransaction.MAX_BYTES]u8,
    payload_len: u16,
    signatures_off: u16,
    message_off: u16,
    message_len: u16,
    // ... other layout fields

    pub fn header(self: *const Transaction) MessageHeader {
        return .{
            .num_required_signatures = self.signature_count,
            .num_readonly_signed_accounts = self.num_readonly_signed_accounts,
            .num_readonly_unsigned_accounts = self.num_readonly_unsigned_accounts,
        };
    }

    /// ... other View methods
}
```

This could replace all four of the other types, which would simplify the code and make it easier to understand. Now maybe there are reasons for the other types, but it's not obvious to me why they're needed. I'd like to be sure each one is individually justified before committing to this complicated approach.

- why do we need VersionedTransaction if we can read directly into the bytes with methods?
- why use View when we have `*const TransactionRecord`?
- why not embed Layout within `TransactionRecord`?

v2/lib/solana/transaction.zig
prestonsn: Big https://github.com/Syndica/sig/pull/1711/commits/ce8db7283f3e6e591976d06e7311e987670e968e

v2/lib/solana/transaction.zig
prestonsn: https://github.com/Syndica/sig/pull/1711/commits/ce8db7283f3e6e591976d06e7311e987670e968e

PR #1715, opened 2026-07-13
conformance(v2:shred): local conformance for v2 shred parse harness

yewman:
WIP


v2/components/replay/forest.zig
```diff
@@ -86,3 +95,568 @@ pub const ExecResponse = extern struct {
...
+
+/// Represents a deshredded FEC set.
+///
+/// Used as a hashmap value, and a tree node (these are the same memory)
+/// This node is also used for the keys of hashmaps. When doing so, be careful of which adapted
+/// context you use.
+///
+/// NOTE: When used inside the Pool, these may be items in a free list. However such nodes should
+/// not be in either map or the tree.
+pub const MerkleNode = extern struct {
```
dnut, line 25: Once you merge in main you'll get a conflict here. I would put all your additions to this file in `component.zig`

v2/components/shred/receiver.zig
```diff
@@ -170,77 +195,46 @@ pub const Receiver = struct {
...
-                // TODO: once repair is implemented, repaired shreds should skip these checks to
-                // allow conflicting fec sets to be inside the in-progress map. We will need to do
-                // this to reliably repair when equivocation is detected.
-                .mismatching_signature => return error.EquivocationDifferentHashForSameFecSetId,
-            }
-
-            // if we have this FecSetId with a different signature, this means equivocation has occured
-            if (state.in_progress.containsId(fec_set_id)) {
-                // NOTE: see above note.
-                return error.EquivocationMatchingFecSetWithDifferentSignatureAlreadyInProgress;
```
dnut, line 218: i think this is still an equivocation case we should detect.

v2/components/shred/receiver.zig
```diff
@@ -157,11 +169,24 @@ pub const Receiver = struct {
         );
         zone.text(str);
 
-        const fec_set_ctx = if (state.in_progress.getFecSetCtx(
-            &shred.signature,
-        )) |fec_set_ctx| existing_set: {
-            // fec set is already being built. This branch will be taken for 31/64 shreds (assuming
-            // zero packet loss).
+        var shred_merkle_root: Hash = undefined;
+        try shred.merkleRoot(&shred_merkle_root);
```
dnut, line 176: if you're going to unconditionally pre-calculate merkle root and allow different signatures, why not just key by merkle root instead of signature?

i'd rather use the optimal design for the current protocol, rather than use the optimal design for some other protocol that we prefer plus some hacks to make it work with the actual protocol we're stuck with. it feels convoluted with the current approach

v2/lib/shred/receiver.zig
```diff
@@ -513,6 +574,16 @@ const InProgressSets = struct {
     signatures: []Signature, // idx correspond with fecset idxs
 
     signature_map: SignatureMap,
+    /// Authoritative primary `(slot, fec_set_idx) → *FecSetCtx`, one entry
+    /// per live ctx. Absorbs case-A equivocation (agave admits both
+    /// same-root variants because `check_merkle_root_consistency` is
+    /// signature-blind) and sig-collision-across-fec-sets under
+    /// sig-verify-off fuzz. `signature_map` is a fast-path index only
+    /// (see `assertCounts`).
```
dnut: this comment is pretty confusing, i can't really make sense of it as standalone reference. it seems that all the comments in this pr are ai-written. some are better than others

v2/lib/shred/receiver.zig
```diff
@@ -513,6 +574,16 @@ const InProgressSets = struct {
...
+    ///
+    /// TODO: fold away once signature uniqueness at `(slot, fec_set_idx)`
+    /// is a protocol invariant.
```
dnut: that's a big if.

conformance/src/shred_parse.zig
```diff
@@ -257,12 +281,19 @@ const HarnessState = struct {
...
                 .data_complete = completed.data_complete,
                 .slot_complete = completed.slot_complete,
             }) catch @panic("OutOfMemory");
+
+            _ = sig_v2.replay.insertFecSet(
+                sig_v2.telemetry.Logger("main").noop,
+                completed,
+                &self.forest,
+                self.block_pool,
+            ) catch {};
```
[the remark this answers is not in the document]
dnut: this seems reasonable, no?

v2/lib/shred/receiver.zig
yewman: 86d462f6f179f0ee9c4aa3acbc4f10bae35e5154
Yeah I agree this one is a bit dubious, I think this is an improvement lmk if you agree

v2/lib/shred/receiver.zig
yewman: 86d462f6f179f0ee9c4aa3acbc4f10bae35e5154

conformance/src/shred_parse.zig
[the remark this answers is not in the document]
yewman: 86d462f6f179f0ee9c4aa3acbc4f10bae35e5154

v2/components/shred/receiver.zig
yewman: We are still detecting it, thats what the error return on line 223 is doing

v2/components/shred/receiver.zig
yewman: I understand where you are coming from, but that would close the door to a future optimisation which we should do, but is not a priority right now. Added a comment for clarity.  060a521f2f72cde61a2444379359886c9e6eb34f

v2/components/shred/receiver.zig
dnut: I don't see it on line 223 but I see something analogous on line 204:
```
            if (!shred_merkle_root.eql(&fec_set_ctx.merkle_root))
                return error.MismatchedMerkleRoot;
```

It's comparing by merkle root which is all we need right now

PR #1718, opened 2026-07-13
feat(ci): codecov fallback

hamza72x:
Alternative to codecov that works for external contributors.

Closes #1696

### DIFF
--- a/.github/workflows/check_linux.yml
+++ b/.github/workflows/check_linux.yml
@@ -84,7 +84,7 @@ jobs:
             v2/zig-out/kcov/kcov-merged/cobertura.xml \
             --compare-branch="$BASE_REF" \
             --format markdown:diff-coverage.md \
-            --fail-under 0 || true
+            --fail-under 0
 
           # Output to GitHub step summary
           if [ -f diff-coverage.md ]; then
--- a/.github/workflows/check_linux.yml
+++ b/.github/workflows/check_linux.yml
@@ -83,7 +83,7 @@ jobs:
             kcov-merged/kcov-merged/cobertura.xml \
             v2/zig-out/kcov/kcov-merged/cobertura.xml \
             --compare-branch="$BASE_REF" \
-            --markdown-report diff-coverage.md \
+            --format markdown:diff-coverage.md \
             --fail-under 0 || true
 
           # Output to GitHub step summary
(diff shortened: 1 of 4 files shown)


.github/workflows/check_linux.yml
```diff
@@ -70,7 +70,32 @@ jobs:
...
+          BASE_REF="origin/${{ github.base_ref }}"
+          echo "Comparing coverage against $BASE_REF"
+
+          # Run diff-cover on both v1 and v2 cobertura reports
+          diff-cover \
+            kcov-merged/kcov-merged/cobertura.xml \
+            v2/zig-out/kcov/kcov-merged/cobertura.xml \
+            --compare-branch="$BASE_REF" \
+            --format markdown:diff-coverage.md \
+            --fail-under 0 || true
```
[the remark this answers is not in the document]
dnut: this is wrong. it worked already and here's the help output:

```
positional arguments:
  coverage_files        coverage report (XML or lcov.info)
```

note: plural

dnut: Why is this here? This step may not be critical, but I don't expect it to fail, and if it does fail, I would like to know about it so we can either fix it or delete it, rather than assuming everything is working.

```suggestion
            --fail-under 0
```

.github/workflows/check_linux.yml
```diff
@@ -70,7 +70,32 @@ jobs:
           echo "--- v2 filename samples ---"
           grep 'filename=' v2/zig-out/kcov/kcov-merged/cobertura.xml | head -3
 
+      - name: Generate local coverage report
+        if: github.event_name == 'pull_request'
+        run: |
+          pip install --quiet diff-cover
```
dnut: can you pin the version here so it always works even if diff-cover changes its cli? 


```suggestion
          pip install --quiet diff-cover==10.3.0
```

dnut (changes requested):

dnut (approved):

.github/workflows/check_linux.yml
hamza-syndica: https://github.com/Syndica/sig/pull/1718/changes/390d8b71cc57503b1cd8ce58481a35182a7cbd39

.github/workflows/check_linux.yml
hamza-syndica: https://github.com/Syndica/sig/pull/1718/changes/0a8e6f73625ee547fb4fe003df0b626b25c60b4d

jbuckmccready (approved):

PR #1723, opened 2026-07-13
fix(util): properly compare type signatures in assertInterface

jbuckmccready:
Declaration types were not compared correctly which allowed function signatures to drift from interface declared. Fixed it, added doc comment, and fixed two drifting cases in the code. Tested locally behavior is as described now in the doc comment.

### DIFF
--- a/v2/lib/util.zig
+++ b/v2/lib/util.zig
@@ -24,7 +24,7 @@ pub fn fmtSlice(slice: anytype) FmtSlice(@TypeOf(slice[0])) {
 
 /// Verifies that a struct, or pointer to struct, has the contract's declarations and fields with
 /// exact types. Methods must use explicit error sets because inferred error sets (`!T`) from
-/// separate declarations do not compare equal.
+/// separate declarations never compare equal even if they have the same inferred set of errors.
 pub fn assertInterface(comptime InterfaceType: type, comptime ContractStruct: type) void {
     const Contract = ContractStruct;
     const Interface = switch (@typeInfo(InterfaceType)) {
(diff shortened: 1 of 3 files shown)


v2/lib/util.zig
```diff
@@ -39,23 +42,22 @@ pub fn assertInterface(comptime InterfaceType: type, comptime ContractStruct: ty
...
-                @typeName(Decl),
+                @typeName(ContractDeclType),
             }));
         }
 
-        // TODO: support function types with error union returns.
-        const IDecl = @TypeOf(@field(Interface, decl.name));
-        if (@TypeOf(Decl) != @TypeOf(IDecl)) {
+        const InterfaceDeclType = @TypeOf(@field(Interface, decl.name));
+        if (ContractDeclType != InterfaceDeclType) {
```
kprotty, line 55: Good fix. 

Is it worth adding back the TODO here in case of differing inferred error sets? Or should the function decls always explicitly match regardless. Fine to accept either way.

jbuckmccready: I tested current behavior with 0.15.2 and inferred error set signature always fails to compare equal (no silent success) even if:
 - They infer identical errors.
 - Both infer an empty error set.
 - One inferred set has the same errors as an explicit set.


This commit also updates the doc comment to state that inferred error sets will never compare equal so I figured I could remove the TODO comment. But after rereading it I improved the wording a bit in the doc comment: https://github.com/Syndica/sig/pull/1723/changes/7e1b8610365b0dbf88324085b97fb787cc8a6e4f

prestonsn (approved):

kprotty (approved):

PR #1725, opened 2026-07-14
chore(gossip): remove redundant GossipMap.Metadata

0xzrf:
Removes redundant`GossipMap.Metadata` to use `GossipMetadata` instead


dnut (approved):
Looks good to me.

FYI the code in src/ is v1, which we are in the process of decommissioning. Mainly we're focused on v2 right now

PR #1734, opened 2026-07-16
test(snapshot): validate snapshot constructed by agave

dnut:
This adds a unit test that validates our code is able to properly deserialize a snapshot produced by agave.

I made this tool to generate minimal snapshots that are small enough to commit into the git repo and can be used for testing: https://github.com/Syndica/snapshot-generator

I wrote this test to use the outputs of that tool.

Maintenance:
- As more fields are deserialized from the snapshot, add validations in this test case to check that your implementation is correct.
- The inputs for this test should be periodically regenerated using the latest version of agave to ensure ongoing compatibility with any changes to the snapshot format.


v2/lib/solana/snapshot.zig
```diff
@@ -795,3 +795,116 @@ pub const ZstReader = struct {
...
+    }
+
+    const account_entries = accounts.get("entries").?.array.items;
+    try expectEqual(jsonU64(accounts.get("count").?), account_entries.len);
+
+    var account_index: usize = 0;
+    while (try snapshot_iter.next()) |account| : (account_index += 1) {
+        try expect(account_index < account_entries.len);
+        const a = account_entries[account_index].object;
+
```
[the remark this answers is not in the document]
dnut: The generated snapshot and json are produced by the same code which we control, and the easiest thing here is to maintain consistent ordering. If necessary we can modify this to use hashmaps or something like that, but I'd rather do that on an as-needed basis.

v2/lib/solana/snapshot.zig
```diff
@@ -795,3 +795,116 @@ pub const ZstReader = struct {
...
+    const blockhash_queue_json = merged.get("blockhash_queue").?.object;
+    try expectEqual(
+        jsonU64(blockhash_queue_json.get("max_age").?),
+        snapshot_iter.manifest.bank_fields.blockhash_queue.max_age,
+    );
+
+    const json_hashes = blockhash_queue_json.get("hashes").?.array.items;
+    const bhq_hashes = snapshot_iter.manifest.bank_fields.blockhash_queue.hashes;
+    try expectEqual(json_hashes.len, bhq_hashes.count);
+    outer: for (bhq_hashes.array[0..json_hashes.len]) |hash| {
```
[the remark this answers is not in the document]
dnut: This can actually follow the pattern I justified in my response to copilot's prior review comment. I can simplify this code to be a single loop by assuming consistent ordering. It's a stronger assertion about the code which has pros and cons. I'm fine with requiring consistent ordering for now and only adding complexity here if we ever decide to have inconsistent ordering.

[75f09d0](https://github.com/Syndica/sig/pull/1734/commits/75f09d0534cbb1fbf0cb5a07248bb209b40bfc9d)

v2/lib/solana/snapshot.zig
```diff
@@ -795,3 +795,112 @@ pub const ZstReader = struct {
...
+    var hash_buf: [Hash.BASE58_MAX_SIZE]u8 = undefined;
+    var pubkey_buf: [Pubkey.BASE58_MAX_SIZE]u8 = undefined;
+
+    const json_path = lib.test_data_dir ++ "test-snapshot-v4.2.0-beta.1.json";
+    const snapshot_path = lib.test_data_dir ++ "test-snapshot-v4.2.0-beta.1.tar.zst";
+
+    const json_bytes = try std.fs.cwd().readFileAlloc(allocator, json_path, 8 * 1024 * 1024);
+    defer allocator.free(json_bytes);
+
+    const parsed_json: std.json.Parsed(std.json.Value) =
```
kprotty, line 819: nit: does this need type-hint

v2/lib/solana/snapshot.zig
```diff
@@ -795,3 +795,112 @@ pub const ZstReader = struct {
...
+    const accounts = parsed_json.value.object.get("accounts").?.object;
+
+    const compressed = try std.fs.cwd().readFileAlloc(allocator, snapshot_path, 16 * 1024 * 1024);
+    defer allocator.free(compressed);
+
+    var zstd_reader_state = try @import("zstd").Reader.init(compressed);
+    defer zstd_reader_state.deinit();
+    var zstd_reader = zstd_reader_state.reader();
+    const tar_bytes = try zstd_reader.readAllAlloc(allocator, 32 * 1024 * 1024);
+    defer allocator.free(tar_bytes);
```
kprotty: worth testing ZstReader here too?

v2/lib/solana/snapshot.zig
dnut: Not for correctness. But without this, zls is unable to figure out the type, and all the usages of this data throughout the test get no help from the editor interpreting fields and methods.

image

image

v2/lib/solana/snapshot.zig
dnut: 1e06876b34429c148a3b36cd3a9f7fca2a02b687

scripts/snapshot-generator/src/main.rs
```diff
@@ -0,0 +1,570 @@
...
+        buf.extend_from_slice(&1u64.to_le_bytes()); // HashMap len = 1
+        buf.extend_from_slice(&blockhash);
+        buf.extend_from_slice(&fork_count.to_le_bytes());
+        buf.extend_from_slice(&2u64.to_le_bytes()); // inner Vec len = 2
+
+        // Ok entry.
+        buf.extend_from_slice(&ok_key);
+        buf.extend_from_slice(&0u32.to_le_bytes()); // Ok tag
+
+        // Err(InstructionError(3, Custom(9))) entry.
```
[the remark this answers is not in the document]
dnut: dca061760b939cdcae156b87583ca66dd6e5e647

scripts/snapshot-generator/check
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+cargo fmt --all -- --check
+cargo clippy --locked --all-targets -- -D warnings
+cargo test --locked
```
jbuckmccready, line 6:
```suggestion

set -euo pipefail
cargo fmt --all -- --check
cargo clippy --locked --all-targets -- -D warnings
cargo test --locked
```

PR #1739, opened 2026-07-17
test(v2.shred-streamer): integrate as a v2 service

hamza-syndica:


v2/tests/replay_offline/main.zig
```diff
@@ -0,0 +1,135 @@
...
+const topology = lib.topology;
+
+const Region = topology.Region;
+
+const Topology = struct {
+    shred_streamer: topology.ServiceRegions(.from(services.shred_streamer)),
+    shred_receiver: topology.ServiceRegions(.from(services.shred_receiver)),
+    telemetry: topology.ServiceRegions(.from(services.telemetry)),
+};
+
```
dnut, line 102: This should also include replay, exec, accountsdb, and snapshot. It's basically a full validator minus networking. I would look for opportunities to reuse code from the main topology

v2/build.zig
```diff
@@ -331,6 +327,8 @@ const Sig = struct {
                     .{ .name = "start_service", .module = start_service },
                     .{ .name = "tracy", .module = deps.tracy },
                     .{ .name = "services", .module = services_mod },
+                    .{ .name = "rocksdb", .module = deps.rocksdb },
+                    .{ .name = "rocksdb-c", .module = deps.rocksdb_c },
```
dnut: This adds rocksdb as a dependency for every service. We can't have rocksdb in sig, it's a massive bloated dependency that destroys compile times. There needs to be a way to ensure this is only used for shred-stream, and it's only used if shred-stream is actually being used

hamza-syndica: https://github.com/Syndica/sig/pull/1739/commits/f1bec2acfa7aa42b52cac8f4eb2f5e239b0acb6c

v2/tests/replay_offline/main.zig
hamza-syndica: https://github.com/Syndica/sig/pull/1739/commits/e47fc585a3f05fa9fa69e0f351b5529222ee3156

v2/config/shred-stream-replay.zon
```diff
@@ -0,0 +1,26 @@
+.{
+    .sandboxing_mode = .threaded,
+
+    .cluster = .testnet,
+    .leader_schedule_file = "",
+    .gossip = .{
+        .port = 0,
+        .advertise_tvu_port = false,
+    },
```
dnut: Why do we need gossip here?

I would either use the same exact config as the main topology, or if you're going to create a special one just for the shred stream mode, then it should be stripped down to only configure the relevant services.

I lean towards using the same config as the main topology so we don't have to juggle multiple config files, but I could see value in either approach.

v2/services/shred_streamer.zig
```diff
@@ -295,7 +319,263 @@ const Arg = enum {
...
+
+        // Batch commit
+        if (unpublished_packets.* >= producer_publish_packets) {
+            writer.markUsed();
+            unpublished_packets.* = 0;
+        }
+    }
+}
+
+// Legacy main() — kept for reference until all test modes are ported to serviceMain.
```
dnut: why aren't all the test modes supported in serviceMain?

also is this even used anywhere? i wonder if we should continue to allow this tool to be used over the network instead of just inside the test topology. @prestonsn do you think we should keep both?

v2/tests/shred_stream_replay/main.zig
```diff
@@ -0,0 +1,403 @@
...
+                .block_pool = block_pool.finish(),
+            },
+            .rw = .{ .exec_req_response = exec_req_response.finish() },
+        },
+    });
+
+    // Wait for shred_streamer to finish streaming. It signals
+    // completion by closing the writer side of the recv ring.
+    // We can't rely on children.isActive() alone because some
+    // services (e.g., accounts_db) never signal idle.
```
dnut: I don't think we should shutdown if the other services are still doing work. What services do not report idle? We should fix that.

v2/build.zig
```diff
@@ -48,10 +48,6 @@ pub fn build(b: *Build) !void {
     const bb_test_step = b.step("bb-test", "Run black box tests.");
     for (tools.black_box_tests) |bbt| bbt.addToStep(bb_test_step);
 
-    // shred-stream
-    const shred_stream_step = b.step("shred-stream", "Stream shreds from an Agave ledger");
-    tools.shred_stream.addToStep(shred_stream_step);
```
dnut: How are you running it now that this is removed? Are you just running the default install step and then executing the compiled binary? 

I think we should keep this so it can be compiled and run as a standalone entity.

It might also make sense to rename this. The service can be called shred-stream but the custom topology should probably have a new name. In v1 we called this "replay-offline" which I think is more suitable since it encompasses more than just streaming shreds. Maybe there's a better name but I can't think of it. In agave they have something similar called "ledger tool" but I think that name is somewhat misleading and it's not really the same thing that this code is doing.

v2/build.zig
```diff
@@ -430,6 +415,39 @@ const Tools = struct {
             .optimize = .Debug,
         }));
 
+        // shred-stream-replay: standalone tool, not auto-run in CI
```
dnut: Just a thought... maybe in the future we *should* run this in CI. We could run it with a predefined ledger and assert that we end up with a specific bank hash at the end. Not in scope right now but just food for thought.

v2/build.zig
```diff
@@ -430,6 +415,39 @@ const Tools = struct {
...
+                "exec",
+                "snapshot",
+                "accounts_db",
+                "telemetry",
+            }) |service_name| {
+                exe.compile.linkLibrary(for (sig.service_libs) |entry| {
+                    if (std.mem.eql(u8, entry.name, service_name)) break entry.lib;
+                } else std.debug.panic("unknown service '{s}'", .{service_name}));
+            }
+            break :blk exe;
```
dnut: Can't we use the same approach as we do for black box tests? I would have tried reusing the code for creating black box tests. We just don't need to call it "black box test" but maybe something more generic like "custom topology" and then that code is used both for creating bbts and any other custom topologies that are triggered in different contexts.

What do you think, could this work cleanly or is it more trouble than it's worth?

v2/services/shred_streamer.zig
hamza-syndica: Tried to keep the changes simple, I was not sure if we would need everything right away, I should've asked. Anyway: @prestonsn waiting on your input.

v2/config/shred-stream-replay.zon
hamza-syndica: > Why gossip?

So that I can parse the same conf file like main one. Let me see if I can ignore certain fields.

Also, having a dedicated conf file would make things easy to test this specific step it separately (the conf structure is basically same). We can disagree on this though.

v2/tests/shred_stream_replay/main.zig
hamza-syndica: After streaming from start to end, it was stuck. So, added a automatic shutdown. But need to check which one was not reporting idle.

v2/config/shred-stream-replay.zon
hamza-syndica: https://github.com/Syndica/sig/pull/1739/commits/298af3d4b0f295353746158e3893c863fe458c17

v2/tests/shred_stream_replay/main.zig
hamza-syndica: https://github.com/Syndica/sig/pull/1739/commits/c32193a63c7a3dbda346a258f5d00d03a0c5ce01

Can you please help if accounts db shutdown is okay with this change?

v2/build.zig
hamza-syndica: https://github.com/Syndica/sig/pull/1739/commits/0dba7dec41509231b6a31056d50924f1f63eca09

Yes, it was being built via the default install step and then executed directly.

Renamed the topology to `replay-offline`, agreed it better describes what this is. Added a dedicated `zig build replay-offline` step so it can be compiled standalone.

v2/build.zig
hamza-syndica: https://github.com/Syndica/sig/pull/1739/commits/3397b5fe3eecabc22279935a1342deb682a74fcc

v2/config/shred-stream-replay.zon
dnut: The file we commit into the repository is only an example. If the format differs, then committing separate examples makes sense. but if the format is identical, I think only one example is needed.

If you want to maintain separate versions locally, that's fine, but it's not relevant to how many files we commit in the repo. That's something you should manage locally by making separate local files in a gitignored location

My point was actually that we should use the same format as the main topology. I just think it's simpler if the format is identical, so you don't get errors from using the wrong file. I'd rather not have to worry about which validator mode a certain config file is compatible with.

v2/services/shred_streamer.zig
dnut: I must be missing something. Doesn't the shred stream tool already support all the options with a cli parser already implemented? I thought the simplest solution would be to just directly plug into that, and automatically get support for everything it already works with. What constraints are there?

prestonsn: I think we should keep both. It would be nice to have the standalone CLI tool (since I've been using that for testing sig running on remote machines, with a ledger on a local machine streaming out the shreds from this tool). And I think the CLI tool is pretty nice for debugging things that could come up in the future (I can't really think of anything here, but one thing I did do last week was use the shred-stream CLI tool to dump specific FEC sets from a ledger to file to create some tests)

I'm leaning towards maybe something like this: 

```
v2/lib/shred_stream/
    config.zig
    agave_blockstore.zig
    plan.zig
    stream.zig

v2/scripts/shred_stream.zig
v2/services/shred_streamer.zig 
```

Basically splitting out the original shred_stream CLI source into its core components, and creating a folder under `lib/` for it. Then the service for in-topology testing and the CLI tool can import what they need from `lib`. 

This should also help clean up the code a bit (I think the service code here is pretty bloated with CLI related stuff)

prestonsn: @dnut Just saw your response, yeah it has a simple CLI parser. The script uses this struct: 

```zig
const PartialConfig = struct {
    ledger: ?[]const u8 = null,
    target: ?[]const u8 = null,
    start_slot: ?Slot = null,
    end_slot: ?Slot = null,
    rate_hz: ?f64 = null,
    test_mode: TestMode = .linear,
    seed: ?u64 = null,
    selected_count: ?usize = null,
    shred_kind: ?ShredKindFilter = null,
    plan_limit: ?usize = null,
    corrupt_bytes: ?usize = null,
    dry_run: bool = false,

    fn finalize(self: PartialConfig, stdout: *std.Io.Writer) ParseArgsError!Config;

(...)
}
```

So it should be possible to just populate those fields/potentially make it `extern struct` to use directly as a shared memory region to then call `finalize()` on to get the final `Config` (basically bypassing the CLI arg parse stuff, but keeping that code in the CLI's source file under `v2/scripts/shred_stream.zig`)

v2/services/shred_streamer.zig
```diff
@@ -295,7 +319,263 @@ const Arg = enum {
...
+    const args_str = ro.config.getArgs();
+    var arg_ptrs: [64][]const u8 = undefined;
+    var arg_count: usize = 0;
+    var iter = std.mem.splitScalar(u8, args_str, ' ');
+    while (iter.next()) |arg| {
+        if (arg.len == 0) continue;
+        if (arg_count >= arg_ptrs.len) break;
+        arg_ptrs[arg_count] = arg;
+        arg_count += 1;
+    }
```
prestonsn: Could we use the strongly-typed `Config` struct (or likely its sibling `PartialConfig`) to both represent the exact supported args to shred stream but also avoid needing to handle string arg parse here? 

This struct: 

https://github.com/Syndica/sig/blob/4350dc658ea7c67bbb945820363b4514763f59f0/v2/scripts/shred_stream.zig#L129-L143

This ties in a bit to my other comment here: https://github.com/Syndica/sig/pull/1739#discussion_r3639053058

v2/services/shred_streamer.zig
```diff
@@ -295,7 +319,263 @@ const Arg = enum {
...
+
+        stats.recordPacket(kind, packet_data.len);
+        unpublished_packets.* += 1;
+
+        // Batch commit
+        if (unpublished_packets.* >= producer_publish_packets) {
+            writer.markUsed();
+            unpublished_packets.* = 0;
+        }
+    }
```
prestonsn: I think there's an activity signaling issue in these loops. I see that we call `try runner.activity.signalIdleSpinning();` when waiting for a place in the ring (back pressure), while waiting we signal that we're spinning. 

Say we then successfully find room for a packet, then we break out of the loop, write the data, update the ring state, and go to the top of the outer loop. We check `try runner.activity.checkCanceled();` , but I don't see where we `signalActive()` again?

v2/services/accounts_db.zig
```diff
@@ -139,19 +137,27 @@ pub fn serviceMain(runner: lib.runner.Connection, _: ReadOnly, rw: ReadWrite) !n
...
                 result.* = res;
                 _ = replay_out.next();
                 replay_out.markUsed();
+                did_work = true;
             }
         }
+
+        if (!did_work) {
+            try runner.activity.signalIdleSpinning();
+        }
```
prestonsn: I don't think this is an accurate reflection of if the accounts_db service is doing work? From my understanding, `queueRead()` can submit async io_uring reads, and `pollRead()` can return  `null` while those reads are still in flight (making progress under the hood). 

Do you think we should track outstanding lookups and only signal idle when there are no queued/in-flight/ready lookups AND no input requests waiting?

v2/tests/replay_offline/main.zig
prestonsn, line 1: Thoughts on renaming this folder to `replay_offline`? Since this seems to be the main entrypoint to starting replay offline topology.

v2/tests/replay_offline/main.zig
```diff
@@ -0,0 +1,385 @@
...
+        },
+    });
+
+    // Wait for all services to finish their work and go idle.
+    // In offline replay the pipeline drains naturally:
+    //   shred_streamer → shred_receiver → replay → exec
+    // Once every service reports idle, cancel and clean up.
+    while (children.isActive()) {
+        std.atomic.spinLoopHint();
+    }
```
prestonsn, line 428: Maybe we want to strengthen this without solely relying on the topology to signal idle to wait for completion? Could we have some way for the guiding service that's driving the replay offline (`shred_streamer` to signal that it's complete its work, and is not just stalled due to back-pressure and is hence signaling idle?)

Basically some sort of intentional communication that things are done given the params/configuration given to replay-offline mode?

v2/tests/replay_offline/main.zig
```diff
@@ -0,0 +1,385 @@
...
+        const cfg_file = std.fs.cwd().openFile(cfg_path, .{}) catch |err| {
+            std.debug.print(
+                "error: cannot open config file '{s}': {s}\n",
+                .{ cfg_path, @errorName(err) },
+            );
+            return err;
+        };
+        defer cfg_file.close();
+
+        const cfg_str = try cfg_file.readToEndAllocOptions(allocator, 1024 * 1024, null, .@"1", 0);
```
prestonsn: `readToEndAllocOptions` is set for deprecation, I think we should prefer something like this here: 


```suggestion
        const cfg_str = try cfg_file.readToEndAllocOptions(allocator, 1024 * 1024, null, .@"1", 0);
        var file_reader = cfg_file.reader(&.{});
        const file_size = try file_reader.getSize();
        if (file_size > 1024 * 1024) return error.FileTooBig;
        const contents = try allocator.alloc(u8, @intCast(file_size));
        defer allocator.free(contents);
        try file_reader.interface.readSliceAll(contents);
```

v2/tests/shred_stream_replay/main.zig
prestonsn: I didn't see this thread earlier, but related: https://github.com/Syndica/sig/pull/1739#discussion_r3639615456

v2/services/shred_streamer.zig
prestonsn: Hmm I'm for keeping things simple and just wiring up the simple forward test mode/plan as a start, but we should also reject modes that aren't supported by the in-topology setup for the time being. From the code it seems current behavior is to silently accept an unsupported mode, then runs `forward` for it. 

Also, flags like `--dry-run` don't really make sense in the in-topology service configuration for shred-streamer, but make sense for CLI to dump information about a planned test. So I feel that the split mentioned above should help clean this up.

v2/services/shred_streamer.zig
```diff
@@ -295,7 +319,263 @@ const Arg = enum {
...
+        @max(1, @as(u64, @intFromFloat(@ceil(@as(f64, @floatFromInt(std.time.ns_per_s)) / rate))))
+    else
+        null;
+    var base_instant: ?std.time.Instant = null;
+    var next_send_offset_ns: u64 = 0;
+
+    while (try shred_iter.next(&err_data)) |entry| {
+        try runner.activity.checkCanceled();
+
+        const key = parseShredKey(entry[0].data) catch return;
```
prestonsn: In the original code, we had an error log here. Thoughts on keeping it? 

https://github.com/Syndica/sig/blob/2b24ca870687821f349b1d8ba27cec1b78570054/v2/services/shred_streamer.zig#L1998-L2004

v2/services/shred_streamer.zig
```diff
@@ -295,7 +319,263 @@ const Arg = enum {
...
+        if (key.slot != slot) break;
+
+        const packet_data = entry[1].data;
+        if (packet_data.len > max_shred_packet_bytes) continue; // skip oversized
```
prestonsn: I don't think we should skip these, this should be a hard error right? Since it's a totally unexpected case. (maybe this is intentional/something came up when testing these changes?)

Original code: 

https://github.com/Syndica/sig/blob/2b24ca870687821f349b1d8ba27cec1b78570054/v2/services/shred_streamer.zig#L1933

v2/services/shred_streamer.zig
prestonsn: Actually, there's likely another issue with this. This would basically fail if the arg string supplied was something like `--ledger "/mnt/Agave Ledgers/testnet"`, and this parse doesn't seem to support escape characters either. I think we should rely of the struct based parsing, supplying the args via `replay-offline.zon`

v2/services/shred_streamer.zig
```diff
@@ -295,7 +319,263 @@ const Arg = enum {
...
+                now.since(base)
+            else blk: {
+                base_instant = now;
+                break :blk 0;
+            };
+            if (now_offset_ns < next_send_offset_ns) {
+                std.Thread.sleep(next_send_offset_ns - now_offset_ns);
+            }
+            next_send_offset_ns = @max(next_send_offset_ns, now_offset_ns) + interval_ns;
+        }
```
prestonsn: With rate-limiting now operating at a "per-slot" level, I think this would introduce a burst in between slots right?

I think we should inline this logic into serviceMain and keep the original rate limiting logic in place.

v2/lib/shred_streamer/agave_blockstore.zig
```diff
@@ -0,0 +1,208 @@
...
+    } else |_| {}
+    allocator.free(nested_rocksdb_path);
+
+    if (std.fs.cwd().statFile(ledger_path)) |stat| {
+        if (stat.kind == .directory) return try allocator.dupe(u8, ledger_path);
+    } else |_| {}
+
+    std.debug.print("ledger path does not exist or is not a directory: {s}\n", .{ledger_path});
+    return error.InvalidLedgerPath;
+}
```
InKryption: Why do we not simply expect `ledger_path` to already be pointing at `rocksdb`? This path resolution stuff seems like something that should be handled at an earlier point outside of this code.

v2/lib/shred_streamer/agave_blockstore.zig
```diff
@@ -0,0 +1,208 @@
...
+        allocator.free(names);
+    }
+
+    for (names, raw_names[0..count]) |*n, raw_name| {
+        n.* = try allocator.dupe(u8, std.mem.span(raw_name));
+        names_len += 1;
+    }
+
+    return .{ .names = names };
+}
```
InKryption: A nice thing about zig is that we can make our code interop with the C stuff more easily just by using `[*:0]u8` and stuff. I would change `names: []const []const u8` to `names: []const [*:0]const u8`, and eschew the intermediate allocation altogether. Would just mean a few ptr-to-slice conversions with `std.mem.span` during lookup, which is fine since all of this will only be used once at the beginning. Can avoid the dupeZs and allocs in many places just by making the parameters be `[:0]const u8`, whose `.ptr` field is `[*:0]const u8`.

v2/lib/shred_streamer/agave_blockstore.zig
```diff
@@ -0,0 +1,208 @@
...
+        var available_cfs = try listColumnFamilies(allocator, rocksdb_path);
+        defer available_cfs.deinit(allocator);
+
+        try requireColumnFamily(&available_cfs, agave_cf_default);
+        try requireColumnFamily(&available_cfs, agave_cf_meta);
+        try requireColumnFamily(&available_cfs, agave_cf_data_shred);
+
+        const has_code_shred = available_cfs.contains(agave_cf_code_shred);
+        const cfs = try columnFamilyDescriptions(allocator, has_code_shred);
+        defer allocator.free(cfs);
```
InKryption: And also delete `columnFamilyDescriptions`:

```suggestion
        var cfs_buf: [4]rocks.ColumnFamilyDescription = undefined;
        const cfs = cfs: {
            const count: usize = if (has_code_shred) 4 else 3;
            const cfs = cfs_buf[0..count];
            cfs[0..3].* = .{
                .{ .name = agave_cf_default },
                .{ .name = agave_cf_meta },
                .{ .name = agave_cf_data_shred },
            };
            if (has_code_shred) cfs[3] = .{ .name = agave_cf_code_shred };
            break :cfs cfs;
        };
```

v2/lib/shred_streamer/plan.zig
```diff
@@ -0,0 +1,132 @@
...
+    return schedule;
+}
+
+pub fn collectSlotShredRefs(
+    allocator: Allocator,
+    blockstore: *const AgaveBlockstore,
+    slot: Slot,
+    kind: ShredKind,
+    refs: *std.ArrayList(ShredRef),
+    cancel: anytype,
```
InKryption: Is there a situation where this shouldn't just be a `lib.runner.Connection`? It isn't a complex data structure, can be pretty simply used directly in tests.
(Same to all other instances of `cancel: anytype`).

v2/lib/shred_streamer/plan.zig
```diff
@@ -0,0 +1,132 @@
...
+    defer shred_iter.deinit();
+
+    var err_data: ?rocks.Data = null;
+    defer if (err_data) |err| err.deinit();
+
+    while (try shred_iter.next(&err_data)) |entry| {
+        if (cancel.isCanceled()) break;
+
+        const key = parseShredKey(entry[0].data) catch |err| {
+            std.debug.print(
```
InKryption: I'll just say this here now: I think all of these instances of std.debug.print should be replaced with our logger.

v2/lib/shred_streamer/config.zig
```diff
@@ -0,0 +1,1225 @@
+const std = @import("std");
+
+pub const Slot = u64;
```
InKryption: This should use `lib.solana.Slot`.

v2/lib/shred_streamer/config.zig
```diff
@@ -0,0 +1,1225 @@
...
+            .linear => "linear",
+            .reverse => "reverse",
+            .shuffle_global => "shuffle-global",
+            .shuffle_slot => "shuffle-slot",
+            .drop => "drop",
+            .late => "late",
+            .duplicate => "duplicate",
+            .corrupt => "corrupt",
+        };
+    }
```
InKryption, line 31: Feel free to use or not use (same to other instances of this):
```suggestion
    fn kebabify(comptime str: []const u8) []const u8 {
        comptime {
            var kebab = str[0..].*;
            std.mem.replaceScalar(u8, &kebab, '_', '-');
            const copy = kebab;
            return &copy;
        }
    }

    pub fn parse(raw: []const u8) ?TestMode {
        inline for (@typeInfo(TestMode).@"enum".fields) |e_field| {
            if (std.mem.eql(kebabify(e_field.name), raw)) return @enumFromInt(e_field.value);
        }
        return null;
    }

    pub fn modeName(self: TestMode) []const u8 {
        return switch (self) {
            inline else => |itag| kebabify(@tagName(itag)),
        };
    }
```

v2/lib/shred_streamer/config.zig
```diff
@@ -0,0 +1,1225 @@
...
+        return null;
+    }
+
+    pub fn kindName(self: ShredKindFilter) []const u8 {
+        return switch (self) {
+            .any => "any",
+            .data => "data",
+            .code => "code",
+        };
+    }
```
InKryption, line 63:
```suggestion
    pub fn parse(raw: []const u8) ?ShredKindFilter {
        return std.meta.stringToEnum(ShredKindFilter, raw);
    }

    pub fn kindName(self: ShredKindFilter) []const u8 {
        return @tagName(self);
    }
```

v2/lib/shred_streamer/config.zig
```diff
@@ -0,0 +1,1225 @@
...
+        // not by the v2 service which writes directly to the IPC ring.
+        const target = self.target orelse "";
+
+        if (self.start_slot != null and
+            self.end_slot != null and
+            self.end_slot.? < self.start_slot.?)
+        {
+            try stdout.print("--end-slot must be greater than or equal to --start-slot\n", .{});
+            return error.InvalidArguments;
+        }
```
InKryption: Usually would print to `stderr` for errors.

v2/services/shred_streamer.zig
```diff
@@ -0,0 +1,244 @@
...
+                .config => |c| c,
+            };
+        } else {
+            const ipc = ro.config;
+            break :cfg .{
+                .ledger = ipc.getLedger(),
+                .start_slot = if (ipc.has_start_slot) ipc.start_slot else null,
+                .end_slot = if (ipc.has_end_slot) ipc.end_slot else null,
+                .rate_hz = if (ipc.has_rate_hz) ipc.rate_hz else null,
+                .test_mode = @enumFromInt(@intFromEnum(ipc.test_mode)),
```
InKryption: Why is this done this way? Why are they not simply the same type?

v2/services/shred_streamer.zig
```diff
@@ -0,0 +1,244 @@
...
+                .seed = if (ipc.has_seed) ipc.seed else null,
+                .selected_count = ipc.selected_count,
+                .shred_kind = @enumFromInt(@intFromEnum(ipc.shred_kind)),
```
InKryption: Same comment here about not being the same type.

v2/services/shred_streamer.zig
```diff
@@ -0,0 +1,244 @@
...
+    rate_hz: ?f64,
+
+    // Rate limiting state
+    base_instant: ?std.time.Instant = null,
+    next_send_offset_ns: u64 = 0,
+
+    pub fn fillPacket(out: *lib.net.Packet, packet_data: []const u8) void {
+        out.len = @intCast(packet_data.len);
+        @memcpy(out.data[0..packet_data.len], packet_data);
+        out.addr = std.mem.zeroes(std.net.Address);
```
InKryption: `std.mem.zeroes` is a codesmell. I see no reason to use it here.
```suggestion
        out.len = @intCast(packet_data.len);
        @memcpy(out.data[0..packet_data.len], packet_data);
        out.addr = .initIp4(.{ 0, 0, 0, 0 }, 0);
```

v2/services/shred_streamer.zig
```diff
@@ -0,0 +1,244 @@
...
+            if (unpublished_packets.* != 0) {
+                writer.markUsed();
+                unpublished_packets.* = 0;
+                continue;
+            }
+            was_idle = true;
+            try self.runner.activity.signalIdleSpinning();
+        };
+        // Re-signal active after recovering from back-pressure idle.
+        if (was_idle) try self.runner.activity.signalActive();
```
InKryption: You don't have to put this behind a conditional like this, you can just call it.
```suggestion
        const result = while (true) {
            if (writer.peek()) |p| break p;
            // Ring full — flush pending writes so reader can drain
            if (unpublished_packets.* != 0) {
                writer.markUsed();
                unpublished_packets.* = 0;
                continue;
            }
            try self.runner.activity.signalIdleSpinning();
        };
        try self.runner.activity.signalActive();
```

v2/tests/replay_offline/main.zig
```diff
@@ -0,0 +1,385 @@
...
+                "  <config.zon>  Path to validator config file" ++
+                " (same format as main validator).\n" ++
+                "                A snapshot must already exist" ++
+                " in the configured snapshot folder.\n\n" ++
+                "  Remaining arguments are passed to the" ++
+                " shred_streamer service:\n" ++
+                "    --ledger <path>  Path to Agave ledger\n" ++
+                "    --start-slot <N> First slot to stream\n" ++
+                "    --end-slot <N>   Last slot to stream\n" ++
+                "    --rate-hz <F>    Rate limit in Hz\n",
```
InKryption:
```suggestion
            \\usage: replay-offline <config.zon>
            \\ [shred-streamer-args...]
            \\
            \\  <config.zon>  Path to validator config file
            \\ (same format as main validator).
            \\                A snapshot must already exist
            \\ in the configured snapshot folder.
            \\
            \\  Remaining arguments are passed to the
            \\ shred_streamer service:
            \\    --ledger <path>  Path to Agave ledger
            \\    --start-slot <N> First slot to stream
            \\    --end-slot <N>   Last slot to stream
            \\    --rate-hz <F>    Rate limit in Hz
            ,
```

v2/tests/replay_offline/main.zig
InKryption: Shouldn't use deprecated functions.

v2/tests/replay_offline/main.zig
```diff
@@ -0,0 +1,385 @@
...
+
+    children.cancel();
+    try children.wait(5 * std.time.ns_per_s);
+}
+
+fn populateSnapshotConfig(
+    data: *lib.snapshot.SnapshotConfig,
+    cfg: Config.Snapshot,
+    cluster: lib.solana.Cluster,
+) !void {
```
InKryption: Might make sense to extract this into `lib.snapshot.SnapshotConfig` if we're going to need to re-use this function like this.

v2/services/shred_streamer.zig
dnut: My thought was that the main process can simply pass down the entire string of CLI args into the shred stream service and it would do exactly the same parsing that it does as its own standalone tool. My thought was that no changes would be needed to how configuration works within the shred streamer. It would still support the same full set of config options without any extra work. This seemed like the easiest approach in my opinion, but I didn't analyze the code too deeply so maybe I missed some challenges.

I can see how options like dry-run may seem weird when managed by a service, but I don't think it's an actual problem as long as the shred-stream service is able to terminate the validator. Having two separate configs seems like unnecessary complexity at this time.

v2/services/shred_streamer.zig
dnut: I don't think we should use a config file for this unless we also use a config file for the standalone shred streamer process. The shred streamer should always be configured the same way regardless of the context where it runs. If we're going to support cli args in the standalone process, we should support the exact same cli args with the same exact code in the service. I agree this parsing is wrong. The problem is that it's reinventing the wheel instead of reusing the same parsing logic that already existed.

v2/tests/replay_offline/main.zig
dnut: The folder is already called replay_offline. Are you talking about moving it up a level?

v2/services/shred_streamer.zig
prestonsn: That's a fair point, I was mostly trying to keep the tool aspect separated from whatever offline-replay would be making use of. It's been nice to be able to quickly modify the tool/script as needed to debug and create test cases, but I'm not really attached to that, and it's likely better to right something in Rust to directly re-use as much as Agave's code as possible for things like parsing and reading the ledger and dumping shreds (removing more CLI things like the dry-run functionality along the way)

In the meantime I'm taking a look at the other approach in [dnut/replay-offline](https://github.com/Syndica/sig/tree/dnut/replay-offline)

